PR #184, opened 2012-11-25
Extensibility of search improved

srkunze:
First of four pull requests.
required extensibility for personalization extension (ckanext-semantic)

[1] https://github.com/okfn/ckan/pull/184
[2] https://github.com/okfn/ckan/pull/185


albina1231:
Nice thread! Thanks<a href="http://www.sirpriz.com/post62">!</a>

Issue #340, opened 2013-01-31
CKAN 2.0 does not run on Windows


thriuin:
When running the CKAN 2.0 release branch on Windows, there are number of fanstatic resources loading errors when it starts. These errors occur in the helpers.py and fanstatic_resources.py files when resources names are built due to the directory separators not being set correctly for the underlying OS (I think).

PR #422, opened 2013-02-19
[wip] Fanstatic broken on windows

tobes:
due to paths

see #340 #341

### DIFF
--- a/ckan/lib/fanstatic_resources.py
+++ b/ckan/lib/fanstatic_resources.py
@@ -45,10 +45,10 @@ def create_library(name, path, depend_base=True):
         path = os.path.normpath(unix_path)
         if not inline:
             # resource_name is name of the file without the .js/.css
-            rel_path, filename = os.path.split(path)
+            rel_path, filename = os.path.split(unix_path)
             filename = os.path.join(rel_path, filename)
             path_min = min_path(os.path.join(resource_path, filename))
-            if os.path.exists(path_min):
+            if os.path.exists(os.path.normpath(path_min)):
                 kw['minified'] = min_path(filename)
             if filename.endswith('.js'):
                 renderer = core.render_js


kodokpintar:
I have exactly the same issue as @banders, below here was the error
  
File "C:\pyenv\src\ckan\ckan\lib\fanstatic_resources.py", line 214, in create_library
    members.append(getattr(module, fanstatic_name))
AttributeError: 'module' object has no attribute 'vendor/bootstrap/js/bootstrap.js'


I tried to run on Windows machine and almost done with the installation. I got issue at paster db init
I have looked nigelbabu's link, i am not quite understand as it mentioned about css and not js.

Please enlighten me.

Thanks!

Dobedani:
Mmm, I understand why fanstatic_resources.py did not work on Windows. I applied these shortcuts:
line 83:  resource = Resource(library, path.replace('\\', '/'), **kw) 
line 101: setattr(module, fanstatic_name.replace('\\', '/'), resource)
Now I do not get AttributeErrors again ;-)

jakubklimek:
Hi, I used release-v2.3-latest on Windows 7. After running into the AttributeError, I manually patched fantastic_extensions.py according to this pullrequest and replaced fantastic_resources.py and it worked! Could you please merge this pull request?

wardi:
@jakubklimek none of the core developers are developing on windows at the moment. Are you willing to step in and help if there are windows-related support requests for ckan?

jakubklimek:
Well, I am not a python developer, I just installed CKAN on Windows using the patch in this pull request and it works, so I am wondering why it isn't merged yet. I am not sure how I could be of assistance with windows-related support requests for ckan.

tobes:
Have you tried in debug and non debug mode?

It only works in one I think

jakubklimek:
I tried running it as <code>paster serve development.ini</code> where <code>debug=false</code>

tobes:
ok well I'm for merging it as that will shake out any last issues

bplewe:
Trying to run 2.6.0 on Windows 2012 Server. Using the 2.5.2 windows install guide on the github wiki (which includes sbhoek's patches to fanstatic_resources.py), it starts up without error messages, but bundles won't load or flip the slashes, and the font-awesome.min.css slashes don't flip (other css files flip just fine). I then went into fanstatic_extensions.py and found another place to flip them:
58:    return self.renderer('%s/%s' % (library_url, relpath.replace('\\','/')))
now the slashes are flipped on the bundles, but still not for font-awesome.css. And the .js bundles still won't load.

As suggested above, I tried debug=true, and all of this works perfectly! Why is that? Of course, running it in debug mode isn't a solution, but is it promising?

I'm happy to tinker with the code if anyone has suggestions to try.

bplewe:
Fixed! It required flipping the slashes in two more places, but now everything (font-awesome.css, bundles) works great without being in debug mode. So here are all the places I changed (I hope your line numbers are close enough to find them):
```
fanstatic_extensions.py:
58:    return self.renderer('%s/%s' % (library_url, relpath.replace('\\','/')))
fanstatic_resources.py:
50:                kw['minified'] = min_path(filename.replace('\\','/'))
85:        resource = Resource(library, path.replace('\\','/'), **kw)
102:        setattr(module, fanstatic_name.replace('\\','/'), resource)
174:              resource_list.append(filepath.replace('\\','/'))
196:        resource_name.replace('/','\\')
```
This is an inelegant solution to be sure, but it works for me.

PR #964, opened 2013-05-31
[Blocked] 500 Server Error when posting a non-string package extra

seanh:
**This is blocked until https://github.com/okfn/ckan/pull/1117 gets merged so that I can write some tests.**

If you post something that isn't a string (eg. a dict) as the value of an extra when creating or updating a package via the API, SQLAlchemy crashes and CKAN 500s.

In CKAN 1.8, it used to be possible to post non-strings as extra values. In 2.0, they have to be strings.

See this issue on ckanext-harvest for details: https://github.com/okfn/ckanext-harvest/issues/42

The point of _this_ issue is that CKAN should catch this at the schema validation level and sent back a helpful error message.


seanh:
Closing as too old

PR #1000, opened 2013-06-13
Enable CORS for previews

domoritz:
See https://en.wikipedia.org/wiki/Cross-origin_resource_sharing

This avoid usage of the resource proxy where possible.


amercader:
I won't merge this sadly, as this causes more problems that the potential benefits (see eg https://github.com/ckan/ckanext-geoview/commit/fb8b99af3f15bb1548793e095efa749271540c17)

domoritz:
This was my last open PR and I get a bit sentimental. I can't believe it's been almost two years and I learned so much from you guys. Time flies by.

amercader:
@domoritz it's never too late to submit a new one :) 
Time flies indeed, but your contributions are still benefiting the community

PR #1105, opened 2013-07-09
[#1078] package_show performance part three

wardi:
A separate pull request for the third part of work in #1079 plus fixes (8.2x improvement for our site, 5.0x improvement without any plugins)

This PR allows package_show to return a LazyJSONObject when the context includes return_type='LazyJSONObject'. This object behaves like a dict but avoids calling json.loads and json.dumps unless absolutely necessary. A LazyJSONEncoder is added as a fall-back within the api controller for LazyJSONObjects nested anywhere within the return value data.


wardi:
#2207 is merged, @rossjones if you don't have time, should I assign this to someone else?

rossjones:
I'll unassign, but I probably shouldn't just pick someone else to take a look.

wardi:
Can we consider this one for just after the 2.4 release? This is a big performance win for us, but would be great to have thorough testing by other users.

davidread:
Nag ;-)

wardi:
Wooo! thanks Joe, those of us with monstrous dataset metadata thank you.

PR #1349, opened 2013-12-19
Write tests for theming example plugins

seanh:


seanh:
Cleaned this up a bit but some of the tests are failing. Looks like the example is broken probably. Would be worth fixing and merging imo, keep the docs working

amercader:

PR #1641, opened 2014-04-10
resource_create should break when called without URL

vitorbaptista:
The `resource_create` logic action has two required parameters: `package_id`
and `url`. If it's called without `package_id`, it'll complain throwing a
helpful `ValidationError` related to it, as in:

```
ValidationError: {'package_id': [u'Missing value']}
```

But if we call it without an `url`, it'll break for some random reason:

```
...
  File "/usr/lib/ckan/default/src/ckan/ckan/logic/action/create.py", line 286, in resource_create
    upload.upload(context['package'].resources[-1].id,
```

We should have validations in place for `url` as well.

### DIFF
--- a/ckan/public/base/less/layout.less
+++ b/ckan/public/base/less/layout.less
@@ -109,6 +109,7 @@
     margin: 0 0 5px 0;
     font-size: 18px;
     line-height: 1.3;
+    .break-word();
   }
   .info {
     margin-top: 15px;
--- a/ckan/public/base/less/media.less
+++ b/ckan/public/base/less/media.less
@@ -90,6 +90,7 @@
   font-size: 18px;
   line-height: 1.3;
   margin: 5px 0;
+  .break-word();
 }
 
 // Overlay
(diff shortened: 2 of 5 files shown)


joetm:
What's the point of this url parameter? Why is it required when I want to upload a local file with the ckanapi?

         res = ckan.action.resource_create(
             package_id='2ad3c9de-502c-403a-8b03-bfc619697ff2',
             apikey='...',
             user_id='...',
             url='dummy-value-because-you-force-me-to',
             description='my upload with CKANAPI',
             upload=open('./myfile.csv')
         )

wardi:
That's a good question!

CarlQLange:
yeah, this has just come up for me as well. no reason for the url param to be mandatory when a file's being uploaded.

vitorbaptista:
At the time, I sent this pull request because the docs said that the `url` parameter was required, but didn't check it on the code. Looking back now, it's obvious that the docs were wrong, and not the code.

I've created an issue about this to fix this problem: https://github.com/ckan/ckan/issues/2769

Issue #320, opened 2013-01-30
Hitting enter on step two of dataset creation submits form and sends you back to step 1


shevski:
1. Go to create a new dataset
2. Fill in name, click on next
3. type in an example or any URL and click enter
4. Screen takes you back to step one of your dataset, however, resource has been added
5. repeat
6. don't click enter to actually finish process and click on buttons
7. you get a page with multiple resources

![Screen Shot 2013-01-30 at 12 48 54](https://f.cloud.github.com/assets/1260289/110899/79dce1de-6adb-11e2-8716-0854c8ceaa77.png)

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #1677, opened 2014-04-25
[#320] Press enter should submit form just like clicking the primary sub...

vitoravelino:
...mit button

The context is the dataset creation wizard (see #320).

One of the things I did was change the wizard clickable steps from `<button>` to `<a>` because I thought that it would make more sense semantically speaking. I also added "type=button" to the others `<button>` to avoid the submit when the user hits enter because it comes first than the primary (blue) one.

For each of those elements, I added the data-* attribute to store the name and value.
I retrieve the data and add the a input hidden when user clicks on one of them.

Feel free to ask or suggest any modification.

Best,
Vítor Avelino.


amercader:

PR #1679, opened 2014-04-25
dataset updates using jsonpatch

### LINKED ISSUE #1416: Atomic update of objects
For what I see, it seems there is no way to update only _part_ of an object through the API; instead it is necessary to:
- retrieve the whole object
- update the retrieved object
- send back the whole object through the API

While in certain cases it might be useful to update the whole object, I think in most use cases (at least, mine) it would be handier / safer to have a way to update just part of the object, such as a single field / extra item.

This not only saves the burden of getting / updating the object, but also helps avoiding the risk of data loss in case two processes are concurrently performing updates on the same object.

It also removes the need to know which parts of the object must be sent back to the update and which are not necessary (eg. all the related objects such as tags, resources, .. seem not to be affected if missing in an updated object).

deniszgonjanin:
To resolve #1416

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -358,13 +358,13 @@ def package_update(context, data_dict):
 def package_patch(context, data_dict):
     '''Patch a dataset (package).
 
-    The difference between the update and patch methods is that the patch will
-    perform an update of the provided parameters, while leaving all other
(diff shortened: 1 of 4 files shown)


amercader:
I think patch actions were added on another PR. Some interesting discussion about further options but this PR is not going to get merged.

PR #1665, opened 2014-05-01
Group auth for any logged-in user

vitorbaptista:
Before CKAN 2.0 it was possible for a group admin to add group auth for 'any logged in user'.

This functionality should be restored as it would be useful for some use cases, see e.g. 

https://lists.okfn.org/pipermail/ckan-dev/2014-April/007301.html


amercader:

PR #1853, opened 2014-07-21
Add quotes to package ID in Solr query in _bulk_update_dataset to prevent Solr errors with custom dataset IDs.

razz0:
Bulk updating of datasets on Organization page doesn't work for our CKAN instance because of using URNs (which contain ":" characters) as dataset IDs. The dataset IDs should be in quotes in the Solr query to prevent errors.

### DIFF
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -264,7 +264,7 @@ class PackageSearchQuery(SearchQuery):
     def get_index(self,reference):
         query = {
             'rows': 1,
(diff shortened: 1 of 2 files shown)


sivang:
@amercader @razz0 I'm seeing this when trying to search for a dataset that has a colon in its name in 2.8.3, could this fix that as well?

amercader:
@sivang it could be related to quoting the relevant parameters yes. You need to track where the query to Some is made

PR #1855, opened 2014-07-22
Extend IDomainObjectModification

morty:
Currently the notify signal is sent before the database commit has occurred. It
is not possible to listen for modification events and know that the changes
have made it to the database. This can lead to a race condition if you try to
get a record from the database.

For an example see:

https://github.com/ckan/ckanext-archiver/issues/6

This extension of the interface would allow notification after the database
commit has occurred.


davidread:
BTW this went into the CKAN 2.3 and subsequent releases

PR #1886, opened 2014-08-18
Don't dictize packages when dictizing a vocab

seanh:
When dictizing a vocabulary, don't dictize every package of every one of
the vocab's tags. This makes vocabulary_dictize and vocabulary_show much
faster, which makes a big difference when creating and updating datasets
that have some custom fields that use tag vocabularies.


seanh:
@davidread Added a couple of tests as requested (expecting them to pass on Travis)

davidread:
Awesome, cheers Sean

davidread:
Doesn't this change make it impossible to get a list of datasets that use a particular vocab? You've removed the test that does this, so it seems quite intentional, so I'd be interested in the reasoning.

As a compromise, why don't you make it a data_dict option in vocabulary_show? I could do the change, if that suits.

wardi:
@davidread there's always package_search. It's got the benefit of actually performing well with more than a few datasets too.

We are sorely lacking a set of examples of package_search use in our docs, though.

davidread:
I just want to play with this a little more when I get a chance, before merging.

davidread:
BTW I agree with @wardi about the searching.

wardi:
@davidread moved the milestone to 2.4 because I think this is ready. Please assign to me if you don' have time to review

davidread:
Hmmm does this mean you can't find out what packages use a tag any more?

wardi:
If a tag has > 1000 packages you can't find all the packages in a tag with tag_show or vocabulary_show anyway. Also you have to wait a very long time and transfer a huge amount of data just to get a response.

You're right, though, this change should add an `include_datasets=False` parameter added to tag_show that does a package_search like the one in group_show when users request the packages.

IMHO vocabulary_show should never return packages.

davidread:
Sounds good. I'll see if I can knock that out.

wardi:
@davidread that was quick! I was going to try implementing the change but you came up with a cleaner thing than what I was going to do.

I think all this needs is a note in the changelog about the vocabulary_show no longer including datasets and the new tag_show parameters and defaults. Maybe also instructions about how to get the complete list of datasets for a tag using package_search, and a link from the tag_show docs.

davidread:
I've added changelog and docs now - see what you think @wardi  The HDX site is the only one I can find with vocabs, and it's clearly unwieldy returning datasets in the vocabulary_list & vocabulary_show functions, so I've given examples using package_search instead.

ckan/logic/action/get.py
```diff
@@ -1273,6 +1273,15 @@ def tag_show(context, data_dict):
...
     :param id: the name or id of the tag
     :type id: string
+    :param vocabulary_id: the id or name of the tag vocabulary that the tag is
+        in - if it is not specified it will assume it is a free tag.
+        (optional)
+    :type vocabulary_id: string
+    :param include_datasets: include a list of the tag's datasets. (Up to a
+        limit of 1000 - for more flexibility, use package_search - see
+        :py:func:`package_search` for an example.)
+        (optional, default: ``True``)
```
wardi: We've been setting the default on include_datasets parameters to False on the other APIs we added it to. I think we should do that here too

davidread: Got you - #2206

davidread:
Ok hopefully that'll pass. Reassigning to you for review, assuming that's ok.

wardi:
Sure thing. master is broken so fixing the merge won't even help much at the moment :-(

ckan/tests/factories.py
```diff
@@ -393,3 +393,20 @@ def validator_errors_dict():
 
     '''
     return {('other key',): ['other error']}
+
+
+class Vocabulary(factory.Factory):
+    '''A factory class for creating tag vocabularies.'''
+
+    FACTORY_FOR = ckan.model.Vocabulary
+    name = factory.Sequence(lambda n: 'test_vocabulary_{n}'.format(n=n))
```
wardi: \<code golf :golf: please ignore\>
```python
name = factory.Sequence('test_vocabulary_{0}'.format)
```

wardi:
I just removed a change to the legacy test that's no longer necessary (testing locally). If this passes travis I will merge.

wardi:
Thank you @seanh ! Thank you @davidread !

That's one more hack I can remove from my own ckan fork.

davidread:
Phew, nice to get that in :-)

wardi:
Just in time too, @joetsoi was just minutes away from merging *my* PR that does a lot of the same things as this one (shows you how good my memory is)

PR #1921, opened 2014-09-10
resource_show: use package_show to get validated data

wardi:
resource_show currently returns resources straight from resource_dictize, skipping the validation that might have occurred in an IDatasetForm plugin. Call package_show instead to fix this issue.

In my case I need this for ckanext-fluent multilingual text fields appearing in resources.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -987,19 +987,26 @@ def resource_show(context, data_dict):
     id = _get_or_bust(data_dict, 'id')
 
     resource = model.Resource.get(id)
-    context['resource'] = resource
+    resource_context = dict(context, resource=resource)
 
     if not resource:
         raise NotFound
 
-    _check_access('resource_show', context, data_dict)
-    resource_dict = model_dictize.resource_dictize(resource, context)
+    _check_access('resource_show', resource_context, data_dict)
 
-    _add_tracking_summary_to_resource_dict(resource_dict, model)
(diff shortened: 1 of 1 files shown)


wardi:
@TomeCirun what's the reason for `plugin_extras` at the resource level? Resource extras are already very lightweight and stored as a combined json field. Are you implementing a plugin that needs to attach hidden data to a resource?

TomeCirun:
@wardi i read it from here and i  thought i can implement the same for Resource as you have described in the description or i misunderstood? https://github.com/ckan/ckan/issues/4147

wardi:
@TomeCirun you're right I mention Resources in https://github.com/ckan/ckan/issues/4147#issue-308129035 along with Users, Groups, Organizations and Datasets.

There really isn't as compelling a reason to add PluginExtras to Resources as there is to convert the package extras to plugin extras, so I was wondering about your planned use case.

In terms of ordering, I think that some of the other related changes should be done first, e.g.:

1. converting package extras (or group/org extras) would yield a major performance improvement and new private data functionality for extensions.
2. completing IUserForm would add significant new functionality for customizing users and enable ckanext-scheming/other extensions to customize. It would also involve creating the validators we need to store/retrieve fields in `plugin_extras` for Datasets, Groups and Organizations.

Adding `plugin_extras` to resources might make sense once we have a compelling use case, but resource and dataset metadata is often treated as part of the same object at an API level so adding `plugin_extras` to resources might not even be necessary.

PR #1952, opened 2014-09-27
Add dependencies needed by python requests for SNI support

tochev:
The default python requests library for python2 is not able to handle
SNI enabled sites (sites with multiple https configurations per IP, see
https://en.wikipedia.org/wiki/Server_Name_Indication for more info).

To correct this, one needs to install pyOpenSSL, ndg-httpsclient and
pyasn1 (recommended packages for urllib3 in some distros, e.g. debian -
https://packages.debian.org/sid/python-urllib3 )

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,8 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -q -y install \
         apache2 \
         libapache2-mod-wsgi \
         postfix \
-        build-essential
+        build-essential \
+        libffi-dev
 
 # Install CKAN
 RUN virtualenv $CKAN_HOME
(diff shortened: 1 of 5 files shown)


wardi:
I think there are some dependencies here that aren't part of adding SNI support, e.g. 'distribute' is the old fork of setuptools iirc.

tochev:
@wardl You seem correct about 'distribute' - I guess it must have been a dependency of something, but now that I reinstalled the SNI packages in a fresh virtualenv none of them required it, so most likely it was a result of the virtualenv creation in an old ubuntu.

Anyway the versions of the packages initially included in the pull request are now somewhat old, so it would be wise to update them before the merge.

wardi:
yes, please update the versions in requirements.in and regenerate the requirements.txt. I'd like to have this considered along with some other requirements changes that will happen shortly after we fork the 2.4 branch.

tochev:
ok, I can do it in the weekend if that would be ok?

wardi:
for sure, take a look at the merge conflict too, please

amercader:

PR #1953, opened 2014-09-29
Don't require a page reload to expand the catalog search filter options

ydave-reisys:
Fix for Don't require a page reload to expand the catalog search filter options

https://github.com/GSA/data.gov/issues/309


amercader:
@ydave-reisys Sorry, this got forgotten as it was assigned to a developer no longer working with CKAN.
I think the original intent was useful, if you want it to have considered for merging please update it with latest master and resubmit. Cheers.

Issue #1961, opened 2014-10-08
Cannot sort the recline grid on fields containing spaces


aliceh75:
`DatastorePlugin._is_valid_sort` and `DatastorePlugin._sort` both expect field names to be escaped (using quotes). This is obvious from the usage of `shlex.split`.

However the recline grid sends the sort string with un-escaped field names. This causes sorts on fields with spaces in them to fail.

Given that:
- This is an API, not an SQL interface, the intuitive approach is to not quote field names;
- Other parts of the API (for instance when selecting fields to return) expect fields to not be escaped
I think the best in terms of consistency is to not escape the fields for the sort string.

For backward compatibility, the API could handle both.

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #1962, opened 2014-10-08
Allow the 'sort' section of the datastore API to work with quoted and unquoted fields.

aliceh75:
As per #1961:

`DatastorePlugin._is_valid_sort` and `DatastorePlugin._sort` both expect field names to be escaped (using quotes). This is obvious from the usage of `shlex.split`.

However the recline grid sends the sort string with un-escaped field names. This causes sorts on fields with spaces in them to fail.

Given that:
- This is an API, not an SQL interface, the intuitive approach is to not quote field names;
- Other parts of the API (for instance when selecting fields to return) expect fields to not be escaped 

I think the best in terms of consistency is to not escape the fields for the sort string.

For backward compatibility, the API could handle both.

### DIFF
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -347,24 +347,22 @@ class DatastorePlugin(p.SingletonPlugin):
         return data_dict
 
     def _is_valid_sort(self, clause, fields_types):
+        '''Return True if clause is a valid sort expression, False otheriwse.
+
+        A valid sort expression is defined as an (optionally double quoted)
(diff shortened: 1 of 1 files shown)


rossjones:
This PR is 3.5 years old now, and doesn't look like it'll make it in without tests and a rebase.
If nobody is interested in finishing this PR in a week or so, suggest we close it.

rossjones:
@amercader I think you can safely close this one now.

amercader:
If someone wants to revisit this feel free to resubmit.

PR #2002, opened 2014-10-26
[WIP] Remove generic "other" licenses

jpmckinney:
Generic "other" licenses are a bad practice. Over 40% of Sweden's and over 30% of Brazil's datasets are under these generic licenses. The generic license options offer no clarity, as they are not, in fact, real licenses. CKAN should only ship with real licensing options. Publishers can always choose to add vague licensing options if they choose.

### DIFF
--- a/ckan/model/license.py
+++ b/ckan/model/license.py
@@ -239,7 +239,7 @@ class LicenseGNUFreeDocument(DefaultLicense):
     @property
     def title(self):
         return _("GNU Free Documentation License")
-O
+
 class LicenseOpenGovernment(DefaultLicense):
     domain_content = True
     id = "uk-ogl"
--- a/ckan/model/license.py
+++ b/ckan/model/license.py
@@ -57,13 +57,8 @@ class LicenseRegister(object):
                 LicenseCreativeCommonsAttribution(),
                 LicenseCreativeCommonsAttributionShareAlike(),
                 LicenseGNUFreeDocument(),
-                LicenseOtherOpen(),
-                LicenseOtherPublicDomain(),
-                LicenseOtherAttribution(),
                 LicenseOpenGovernment(),
                 LicenseCreativeCommonsNonCommercial(),
-                LicenseOtherNonCommercial(),
-                LicenseOtherClosed(),
                 ]
             self._create_license_list(default_license_list)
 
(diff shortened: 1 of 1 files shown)


davidread:
There's concern about this changing breaking existing instances that update, for datasets already using the 'other' licences.

jpmckinney:
@davidread  Like @seanh proposed, the licenses can be removed from the dropdown; if an old dataset is licensed under one of these generics, its entry can be added to the dropdown just so that it saves correctly.

davidread:
Sorry @jpmckinney I hadn't read this properly. That sounds like a reasonable thing.

And I quite like Sean's idea to push one licence, but there still needs to be a way to select other licences. Having a default value pre-selected is probably a powerful-enough thing. Whilst a CKAN for a city will usually get to choose its licence for all its datasets, and forcing a single licence might well be suitable, it's also quite common (e.g. London Datastore, data.gov.uk, data.gov, new EU-wide aggregation of national portals later this year) to see CKAN used as an aggregation of a number of data owners, each with different licences, so you do want easy selection. data.gov.uk would not want a default value pre-selected, since we'd rather publishers made a conscious decision to select the correct licence, rather than get it wrong. So ideally we'd just have a config option containing the default licence, that data.gov.uk would change to blank.

I agree it would be useful to include the reference the Creative Commons URLs, but really it should be in addition to ones to a standardized licence catalogue http://licenses.opendefinition.org/ The two URLs serve different purposes and are equally valid.
http://licenses.opendefinition.org/licenses/CC-BY-4.0.json returns:
```
{
    "domain_content": true,
    "domain_data": true,
    "domain_software": false,
    "family": "",
    "id": "CC-BY-4.0",
    "od_conformance": "approved",
    "osd_conformance": "not reviewed",
    "maintainer": "Creative Commons",
    "status": "active",
    "title": "Creative Commons Attribution 4.0",
    "url": "https://creativecommons.org/licenses/by/4.0/"
}
```
CKAN should probably just store the title, od_conformance and the two URLs. I've opened a separate ticket for stuff related to this: https://github.com/ckan/ideas-and-roadmap/issues/130

jpmckinney:
@davidread This seems reasonable. I wonder if perhaps the dropdown can be fixed first, and the questions of default licenses, additional license metadata, etc. can be addressed later? I've collected the metadata of all Open Government Partnership, and over 32,000 datasets use one of these "generic" licenses, which is not good for licensing clarity.

davidread:
Good call @jpmckinney I've moved the discussion about defaults to https://github.com/ckan/ideas-and-roadmap/issues/131 This ticket should stick to removing the 'Other' options.

I think that what remains to be addressed on this pull are:

* 'other license' entry is added to the dropdown if the dataset already uses it (for backwards compatibility)
* 5ish broken tests

davidread:
Getting rid of the generic licenses is being done in #2452, so closing this. Thanks for the push @jpmckinney !

jpmckinney:
:+1:

Issue #1976, opened 2014-10-13
Crash when site_url not set


seanh:
Lots of stuff breaks with site_url is not set in the config file. Email notifications, FileStore... Make site_url mandatory and make CKAN crash at startup if it's not set or not valid.

Jan-Marcek:
Hi 
https://github.com/ckan/ckan/commit/4a046aee93e78d69362e3abc02df780ba8890f3d#commitcomment-9821429

I'll be really grateful to hear a answer from you guys.

davidread:
Done

rossjones:
The original problem seems to have been fixed by #1976

PR #2016, opened 2014-10-29
Make sure CKAN complains if site_url is missing

rossjones:
As per #1976, stuff breaks if the site_url is not set. This makes it mandatory with a RuntimeError if it is not set.


amercader:
@wardi @davidread I've added the schema and backslash checks, and improved the tests

wardi:
I bet accepting the hostname from clients could be abused too. I can set any name for a site in my hosts file, if the server accepts it it might send password reset emails with a link to a domain under my control.

Issue #1651, opened 2014-04-17
Ubuntu 14.04 LTS support


amercader:
We should aim to support Ubuntu 14.04, probably by next release.

I'd expect the main issues would be on installation instructions and libraries / requirements compatibility.

Things to check:
http://packages.ubuntu.com/trusty

* [x] Postgres 9.3 - SQL commands (eg datastore)
* [x] Solr 3.6 - Solr 4 has been widely used across CKAN instances, so I wouldn't expect problems here
* [x] Libraries versions (see eg #1635)
* ...

badbruce07:
I have some issues in setting up on ubuntu 14.04 LTS regarding jetty server and schema.xml. How can I resolve this issue?

florianm:
What issues are you running into exactly? I posted a more detailed walk through on [open-data](https://github.com/opendata/CKAN-Multisite/issues/10), hope that helps.

frogsapo:
I'm newbie to CKAN and I'm trying to install it from sources on Ubuntu 14.04 LTS.
I followed the current version of the [CKAN install guide](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) and hit the solr-jetty snag mentioned by @rparrapy and @badbruce07 as well as some [other folks](http://stackoverflow.com/questions/30355839/ckan-local-installation-500-error-on-solr-jsp-support-not-configured) which brought me to this discussion.

As I understand it (and my understanding is very limited) this solr-jetty bug probably has to do with an ill packaged old version (6 apparently) of jetty on Ubuntu 14.04 but alas the tomcat-coyote.jar workaround mentioned by @Hoedic, that did the trick for @pwhipp, didn't work for me. And as newbie, the "install solr manually" path (documented by @florianm ) looks quite steep to me. 

Hence, if want to install from sources while remaining in CKAN's mainstream, what would you guys - as CKAN advanced users and experts - recommend :
1. have my sysadmin install an Ubuntu 12 vm and start from there ?
2. try the "solr manual install" ?
3. go for installing jetty version 8 ? 
4. another path I missed while googling ?
Thanks.

amercader:
Just today I've installed Solr on Ubuntu 14.04 without any problem (`sudo apt-get install solr-jetty` + edit the `/etc/default/jetty` file), so I wonder if there is something weird with old versions.

Here's my server version:

```
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
[...]

PR #2020, opened 2014-10-29
[#1651] Support ubuntu 14.04

seanh:
Get the source install, deployment and package install working on Ubuntu 14.04


opensas:
I also think that support for Ubuntu 14.04 is quite important. Nevertheless, if building a package for it is much too complicated, it would be wise at least to include the detailed instructions to perform a source deployment in production for a 14.04 server.

amercader:
Closed in favour of #2726

PR #2023, opened 2014-10-30
Dataset are included in organization pages that don't need them

Zharktas:
Organization pages like edit ja members don't exclude datasets from api query like they are in read-function

https://github.com/ckan/ckan/blob/master/ckan/controllers/group.py#L186

This creates slow page rendering when there are lot on extra fields.


Zharktas:
This probably can be merged now since `edit` and tests were fixed in #2065 ?

Issue #1851, opened 2014-07-21
Resource views pre-release work


amercader:
The new resource views branch (#1251) ~~is about to be merged~~ has been merged but there are issues
that need to be addressed before being ready for release. The main ones are 
(new issues should be created for each of these):

- [x] Better handling of the default views to create for a particular format.
    Right now each view implements `IPackageController` to listen to creates and edits
    and create a view of its type if the formats are correct. Also the paster command
    has view types hardcoded for the view types to create. This can be handled better
    with a list of default views that need to be created on the ini file and a `ViewCreator` 
    or similar system plugin that takes care of that. Needs a new extension interface to know when the datapusher finished an upload (#1852).

- [x] Related to the previous point, the `views` paster commands on cli.py needs a refactoring, as has view types hardcoded and it won't work for other custom view types. Also strings are not translatable.

- [x] The UI for the views page needs to be improved, there is too much whitespace, and the carousel  on top is not very convenient when having several views.

- [x] Docs. There are none, and they need to be there. How to add new view types, how to write new views (example extension with tests?). Crucially, how to migrate existing instances to the new  views and how to adapt existing `IResourcePreview` plugins.

Other minor stuff but also important (these can be combined in less issues):

- [x] Rename all view plugins consistently to end with `_view`. Rename the existing ones, keeping a plugin with the old name for compatibility (eg `pdf_preview`)

- [x] More guidance in the frontend for why there are no view types available on the views form  (need to load view plugins). Show a "No views available" message if no views could be found.

- [x] More guidance as for why there is not a view on the resource page (plugin view for this format not loaded, need to migrate old views, datastore not ready, ...). It can be located in a "Show more info" div. Try to show that the datapusher is uploading data on the resource page.

- [ ] `datapusher_hook` action needs an extension point so other plugins can do the same as `recline_grid` currently does and create a view automatically when an upload to the datastore is finished.

- [x] Clarify how the `clean` command works. Do we need a command for deleting all views from a particular type?

- [x] Add deprecation log messages in `IResourcePreview`

- [x] Can you embed views? How does it work?

- [x] Keep `recline_preview` (renamed as `recline_full_view` or `data_explorer_view`) as default  for CSV/Excels for now on the next release.

- [ ] ...

PR #1852, opened 2014-11-04
Better handling of default resource views

amercader:
Parent #1851

In order to create a default view object for certain formats (eg an `img` tag for an image, a grid for a CSV, ...) right now each view plugin implements `IPackageController` to listen to create and edits and create a view of its own type if the formats are correct.

This is not optimal as each view needs to replicate this logic. A different approach could be that the views are left to decide whether if they are suitable against a particular condition (generally a certain format), but the logic to decide if they need to be created or not is left to a system plugin that takes care of that depending on the values of a config option. This config option (`ckan.views.default_views` or similar) will contain the names of the view plugins that users want to get created by default (of course it will have a default as well).

Views that don't depend on the datastore, will get created on the `after_create` and `after_update` methods of `IPackageController`, similarly to how the view plugins [do it now](https://github.com/ckan/ckan/blob/1251-resource-view/ckanext/textpreview/plugin.py#L88:L111).

For those that depend on the datastore being ready, we need a new `IDatapusher` interface that allows to hook into [this particular](https://github.com/ckan/ckan/blob/1251-resource-view/ckanext/datapusher/logic/action.py#L157:L167) point in the code, and allows other views to know when the uploading to the datastore was finished (not just the hardcoded `text_grid`). Of course that depends on the datapusher being used, if people POSTs stuff directly to the datastore it won't work.

This will also help make the paster commands more generic and work for custom view types, as right now they have the view types to create hardcoded.

Quick and dirty pseudo-code for the system plugin that creates default views (we might need to discuss the best way of handling datastore/not datastore based):

```
class ViewsCreator

    implements IPackageController
    implements IDatapusher


    after_create
        add_default_views

    after_update
        add_default_views

    after_datastore_upload
        add_default_views


    add_default_views(dtastore=True/False)

        d_v = config.get()

        for v in d_v;
            info

            if v.can_view(resource):
                create_view

            if info.dastore_based and datastore:
               create_view
            else
               create_view
```


wardi:
My treat-no-resource-views-as-default-resource-views suggestion is just a performance optimization and affects code that has already been merged so it shouldn't hold up this PR.

Would someone that's using resource views care to give one last review and merge this PR?

amercader:
Thanks @wardi and @joetsoi, I think that the migration side of things (improving the `paster` stuff) and docs are still needed before 2.3, but we definitely can move this forward so it can be merged and improve the others afterwards in separate PRs.

amercader:
@joetsoi I merged master and fixed travis

amercader:
@joetsoi sorry, merge conflicts, I'll fix them and resubmit

joetsoi:
The dataproxy has been reenabled right? So recline_preview could be added by default again? I'm merging this now anyway.

amercader:
@joetsoi you are right, `recline_view` should be default, I'll add it on a separate PR

Issue #2008, opened 2014-10-28
Large CKAN instance is very slow (activity listing)


mknezevic:
As suggested on the <a href="https://lists.okfn.org/pipermail/ckan-dev/2014-October/008269.html" target="_blank">dev list</a>, I'm posting this as a new issue.

I have a CKAN instance (upgraded to 2.2.1) with ~90,000 datasets. Most of these datasets (~80,000) belong to one organization. With an increase of number of datasets, everything becomes utterly slow. What happens is that postgres becomes very busy for a long period of time. I set up db logging (queries that take more than 2 seconds to execute are logged) and it seems that this query is executed each time I try to access, well, anything, even the About page. It takes more than 7 seconds to execute and problematic are external sort and merge. I don't think I should increase Postgres work_mem that much... Here are the logged query and the result of explain analyze

```
 duration: 7731.616 ms  statement: SELECT anon_1.anon_2_activity_id AS anon_1_anon_2_activity_id, anon_1.anon_2_activity_timestamp AS anon_1_anon_2_activity_timestamp, anon_1.anon_2_activity_user_id AS anon_1_anon_2_activity_user_id, anon_1.anon_2_activity_object_id AS anon_1_anon_2_activity_object_id, anon_1.anon_2_activity_revision_id AS anon_1_anon_2_activity_revision_id, anon_1.anon_2_activity_activity_type AS anon_1_anon_2_activity_activity_type, anon_1.anon_2_activity_data AS anon_1_anon_2_activity_data
        FROM (SELECT anon_2.activity_id AS anon_2_activity_id, anon_2.activity_timestamp AS anon_2_activity_timestamp, anon_2.activity_user_id AS anon_2_activity_user_id, anon_2.activity_object_id AS anon_2_activity_object_id, anon_2.activity_revision_id AS anon_2_activity_revision_id, anon_2.activity_activity_type AS anon_2_activity_activity_type, anon_2.activity_data AS anon_2_activity_data
        FROM (SELECT activity.id AS activity_id, activity.timestamp AS activity_timestamp, activity.user_id AS activity_user_id, activity.object_id AS activity_object_id, activity.revision_id AS activity_revision_id, activity.activity_type AS activity_activity_type, activity.data AS activity_data
        FROM activity
[...]

PR #2034, opened 2014-11-05
activity query performance improvements part 1

wardi:
This is an attempt at a "minimal" fix for #2008 

"minimal" in quotes because this change makes the original query even larger by passing down the limits to each subquery. The effect is to limit the run time of the query in the most common cases without changing the structure of the query, in the hope that this can be treated as a fix and be allowed in to 2.3

Part 2 would be a rewrite of this query, hopefully:
1. not using subqueries
2. using joins instead of passing in lists of values queried separately
3. adding a parameter to limit the activity date ranges that we're interested in (no older than..) so that the every-single-page dashboard query of unseen activities can be made faster

### DIFF
--- a/ckan/model/activity.py
+++ b/ckan/model/activity.py
@@ -93,7 +93,7 @@ def _activities_union_all(*qlist):
     and remove duplicates
     '''
     import ckan.model as model
-    return model.Session.query(model.Activity).select_from(
+    return model.Session.query(model.Activity).select_entity_from(
         union_all(*[q.subquery().select() for q in qlist])
         ).distinct(model.Activity.timestamp)
 


amercader:
@wardi looking at this for backporting. Unfortunately it can not go as it is now because `select_entity_from` was added on SQLAlchemy 0.8 and CKAN 2.2.x uses SQLAlchemy 0.7.

If you can refactor [this snippet](https://github.com/ckan/ckan/pull/2034/files#diff-43183fa02245f85a419fe7bd91d0b482R96) targetting 2.2 (so SA 0.7) I'm happy to backport it.

wardi:
@amercader  Yes, I think the change to make this work on 2.2 is to revert https://github.com/wardi/ckan/commit/a4de57e5838d649f4ce873187cd148c08973427c

amercader:
@wardi great! backport done

PR #2079, opened 2014-11-23
makes changes to solr schema to close #2078

### LINKED ISSUE #2078: Organization missing from Solr Schema
ckan/lib/search/index.py contains the following lines to index organizations in solr:

    if pkg_dict.get('organization'):
           pkg_dict['organization'] = pkg_dict['organization']['name']
        else:
           pkg_dict['organization'] = None

However the schema.xml file has no parameter for organization, therefore organizations are not included in the searchable part of the dataset index.

mbellotti:


amercader:
@mbellotti Just to clarify, you are right in that there is no explicit `organization` field in the Solr schema, however it is indexed as part of the ` <dynamicField name="*" type="string" indexed="true" stored="false"/>` catch all field, so you can do:

https://data.hdx.rwlabs.org/search?q=organization:mapaction

or

https://data.hdx.rwlabs.org/search?organization=mapaction

What it does not happen, and I guess this is what you meant originally, is that the field is not indexed in the general `text` field which aggregates text from various fields for the global search. So without this patch this won't return any results:

https://data.hdx.rwlabs.org/search?q=mapaction

Can you confirm this is the case?

One small thing to note is that this may have effects when indexing the org name in the general field, for instance note the number of results returned on this query:

https://data.hdx.rwlabs.org/search?q=MapAction

This is probably due to Solr tokenizing "Map" and "Action" separately.

wardi:
need this change in multilingual solr schema too?

Issue #2095, opened 2014-11-25
'allow_partial_update' needs documenting for group_update


davidread:

davidread:
Merged, so marking this as closed.

PR #2096, opened 2014-11-25
'allow-partial-update' documented for 2 action funcs

davidread:
Not in docstring to stop it showing over the API where it cannot be set - this option is just for internal callers.

Issue: https://github.com/ckan/ckan/issues/2095


davidread:
Ping @joetsoi - this should be a quick one to review

Issue #1912, opened 2014-09-05
Allow homepage placeholder to be modified in the admin UI


marks:
Currently, you can change the home page image in https://github.com/ckan/ckan/blob/master/ckan/templates/home/snippets/promoted.html while the text can be edited in the Admin UI. 

@wardi and I agree that this should be able to be done from the Admin UI:

> 9:16 AM <SkramX_> whats the rationale for not letting that be changed in the online system like the text right above it?
> 9:25 AM <wardi> SkramX: good question. want to submit a pr to change that?

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #2100, opened 2014-11-26
[#1912] Allow sysadmin to change featured image, image link and text in ...

george-sattler:
...the home page  Introductory area.

With this change, sysadmin will see 3 new fields on the ckan-admin/config page to control
- The image
- The image link
- The short line of text under the feature image.


bluefedora:
Did the changes as outlined and got this error when updating the config.  I'm wondering if there's a place to store the fields which I missed.  In a table?  Somewhere else?

The form contains invalid entries:

message: Configuration option(s) 'ckan.featured_image ckan.featured_text ckan.featured_link' can not be updated

Thanks,
BF

amercader:
@bluefedora you probably need to add the config options that need to be editable to this schema: https://github.com/ckan/ckan/blob/master/ckan/logic/schema.py#L657.

PR #2106, opened 2014-12-01
Ideas 31 resourceproxy max file size configurable

florianm:
Addressing roadmap [issue #31](https://github.com/ckan/ideas-and-roadmap/issues/31), this introduces a config-based resource_proxy.max_file_size.

* The config template has now a new, commented-out, section "Resource Proxy" with setting ckan.resource_proxy.max_file_size showing the default value
* The Resource Proxy controller seeks this setting, or defaults to the previously hard-coded 1MB.

Next steps:
* Write tests
* Determine sane limits for max_file_size
* Handle errors caused by extreme max_file_size settings with informative log messages

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -143,6 +143,10 @@ ckan.feeds.author_link =
 #ckan.datapusher.formats = csv xls xlsx tsv application/csv application/vnd.ms-excel application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
 #ckan.datapusher.url = http://127.0.0.1:8800/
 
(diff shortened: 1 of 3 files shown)


joetsoi:
@florianm, I've merged this in,  but we should probably change the config setting so that 1 mean 1MB instead of having to supply "1 * 1024 * 1024". The filestore (ckan.max_resource_size) and ckan.max_image_size are specified in MB, so it'd probably be best to keep things consistent for the end users.

florianm:
That makes sense, yes! Would you like me to push that as an update to this branch, or create a new PR?

joetsoi:
just create a new pr.

Issue #1753, opened 2014-06-09
Nothing is new_


wardi:
Targets for renaming:

- [x] `templates/package/new_package_form.html` #1659
- [x] `new_authz.py` #1995 
- [x] old `tests` become `tests/legacy` #2221
- [x] `new_tests` become `tests` #2349

wardi:
Thank you @vitorbaptista @amercader @davidread @joetsoi !!!

davidread:
:+1:

PR #2134, opened 2014-12-10
2125 add organization autocomplete

mwahlroos:
Add autocomplete action and tests

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -174,6 +174,8 @@ def make_map():
         m.connect('/util/resource/format_icon',
                   action='format_icon', conditions=GET)
         m.connect('/util/group/autocomplete', action='group_autocomplete')
+        m.connect('/util/organization/autocomplete', action='organization_autocomplete',
+                  conditions=GET)
         m.connect('/util/markdown', action='markdown')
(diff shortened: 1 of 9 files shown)


ckan/config/routing.py
```diff
@@ -174,6 +174,8 @@ def make_map():
         m.connect('/util/resource/format_icon',
                   action='format_icon', conditions=GET)
         m.connect('/util/group/autocomplete', action='group_autocomplete')
+        m.connect('/util/organization/autocomplete', action='organization_autocomplete',
+                  conditions=GET)
```
wardi: This is here just because the /util/group/autocomplete is here, right?

But this action is available by the API, what's the use of the extra controller?

davidread: I imagine consistency is the only thing, and IMO not unreasonable

wardi:
old test dir strikes again, ahem #1753 ... This code likely won't keep working if we rip out the old tests.

@davidread since @mwahlroos isn't still interested in this, what do you think? should I drop it or take it as-is with new hacks on top of old fragile tests?

davidread:
I've taken this and put improved tests, in new PR: https://github.com/ckan/ckan/pull/2226

I've also tried to progress #1753 with a PR for the test dir rename.

PR #2139, opened 2014-12-11
2138-palceholders-not-transatable

EnxEng:
Really quick fix


amercader:
Thanks for the patch @EnxEng

Issue #1963, opened 2014-10-08
Datastore breaks if there's a row with a single whitespace on a numeric field


vitorbaptista:
Say that I submit a CSV file with contents:

| value |
| ------- |
| 52     |
| 41     |
| 32     |
| _...thousands of rows..._      |
| 99     |
| _(empty)_ |
| _(whitespace)_ |
| 102   |

The Datapusher will split this file in smaller chunks and submit to the Datastore. Now imagine that the first chunk have just nicely formatted numbers. Seeing this, the Datastore will create a table with a numeric column `value`. All fine. But then, in the next chunks, a row with just a single whitespace `"  "` appears. We'll try to insert that into the Datastore, and it'll fail.

There's no easy way for the user to know what is going on. The error message sent to the user is very cryptic. Even I spent almost one hour trying to figure out what was happening (and could only discover by looking into the machine's logs).

The problem here is that the user data is messy, and we could argue that we shouldn't try to clean it up (because we don't know what he wants to do). But I guess just stripping all strings that come our way isn't going to hurt.

A long time ago there was a similar discussion on OpenSpending. I vagely remember that we decided to strip strings there as well, but I'm not sure. Do you remember it, @tryggvib?

Aaron-M:
Another case where the datapusher fails in a CSV such as below.  The file in question actually has 33581 records, but only some were imported.  The problem seems to be with the 'Animal code' column - the DataPusher interpreted A8301 as a Date/time, rather than a string, and thus created 2 problems - misrepresented the data it did import, and failed to import all the data (probably when it encounterd data that did not conform to the assumed field format). I got around it by putting the "A8301" inside quotes, but that is not ideal.  

Example csv:
TrackID,Study Site,Animal code,NZDT_datetime,Latitude,Longitude,HDOP,Season
1,NC,A8301,21/09/2009 00:56:36,-42.14529,173.51146,1.8,spring
2,NC,A8301,21/09/2009 02:57:35,-42.14553,173.51102,9.1,spring
3,NC,A8301,21/09/2009 04:57:23,-42.14539,173.51086,2.1,spring
4,NC,A8301,21/09/2009 07:56:34,-42.14533,173.51110,1.5,spring

amercader:
We have improved the information on messages on #2145 for 2.3. On next releases we can try and improve the actual handling of wrong data. Bumping the milestone for now.

davidread:
[...]

PR #2145, opened 2014-12-15
Improve datapusher error messages

seanh:
This is a rewrite of https://github.com/ckan/ckan/pull/2000 based on recent CKAN Service Provider and DataPusher refactoring.

This only works with DataPusher 0.0.2, it's not backwards-compatible with DataPusher 0.0.1 and I don't think we should bother with backwards-compatibility.

The job dicts that DataPusher 0.0.2 sends to CKAN to report the outcome of a job *always* contain an "error" key. The value for this key is either `None` (or an empty dict) if there was no error, or if there was an error then the value is always a dict with a "message" key whose value is a human-readable string. The error dict may also contain other keys specific to the type of error, and DataPusher now uses these extra keys to report specific information about HTTP errors (status code, the URL that errored, etc).

This pr changes CKAN's templates to display all keys from these new error objects:

screenshot from 2014-12-12 21 02 47

screenshot from 2014-12-12 20 51 19

This pr also changes the datapusher page to always show the upload log, even when the job has errored

Note that https://github.com/ckan/ckan/pull/2000 also

### DIFF
--- a/ckan/templates/package/resource_data.html
+++ b/ckan/templates/package/resource_data.html
@@ -20,7 +20,7 @@
     </div>
   {% elif status.task_info and status.task_info.error %}
     <div class="alert alert-error">
-      <strong>{{ _('Error:') }}</strong> {{ status.task_info.error }}
+      <strong>{{ _('Error:') }}</strong> {{ status.task_info.error.message }}
     </div>
   {% endif %}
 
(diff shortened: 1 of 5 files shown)


Aaron-M:
#1963 Added another example of 'invalid' data (value A8301 interpreted as a date/time...)

amercader:
@seanh I tweaked a bit the message returned when inserting wrong data. If you are happy with my two commits (290850554bfff7e22e8cd0410074eb41b68b3563 and 59975ba0a23cd26bc72c096bdbe381feb34ea3b5) I'll merge the whole thing

seanh:
@amercader Those look fine to me. I think our ckan service provider, datapusher and datastore work is far from done, but happy for this to be merged now

Issue #2147, opened 2014-12-17
Datastore doesn't update if we upload file a new but it has same name as the existing file.


tobes:
I have a fix for this to follow

If we change a resource by uploading a new file and it has the same name ckan does not see the resource as having changed and therefore there is no trigger for IResourceUrlChange therefore we do not update the datastore

wardi:
@tobes I've fixed this a different way in #2519, let me know if this works for you.

tobes:
@wardi shall we close this?

reopen if you disagree

Issue #1874, opened 2014-08-07
Modifying resource doesn't update it's last updated timestamp


Zharktas:
Resources show their creation date and last updated date which are the same. If the resource is modified, the last updated date should change but it is not updated.

wardi:
I think we should update this value when users upload files too. It's displayed to end users as the resource's "last updated" date, and users expect that date to change when they upload a new file.

augusto-herrmann:
I think the meaning and presentation of the fields should be more clear and consistent.

last_modified: the date and time when the actual data has been last changed
revision_timestamp: the date and time when the resource metadata has been last changed

If this is the correct meaning for the fields, the visual label on the templates for the resource should be consistent with that:

last_modified -> "Data last updated"
revision_timestamp -> "Metadata last updated"

Currently these fields are used interchangeably to display the same label "Last updated". See

`<td>{{ h.render_datetime(res.last_modified) or h.render_datetime(res.revision_timestamp) or h.render_datetime(res.created) or _('unknown') }}</td>`
https://github.com/ckan/ckan/blob/master/ckan/templates/package/resource_read.html#L158

dmichelakis:
I just noticed that the last modified date does not update when editing or replacing a link in a resource but does change when editing or replacing a file. Is this a bug or is this suppose to happen?

PR #2148, opened 2014-12-17
[#2147] allow files with the same name to replace datastore contents

tobes:
issue #2147

If we change a resource by uploading a new file and it has the same name ckan does not see the resource as having changed and therefore there is no trigger for IResourceUrlChange therefore we do not update the datastore

### DIFF
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -27,6 +27,12 @@ def resource_dict_save(res_dict, context):
     table = class_mapper(model.Resource).mapped_table
     fields = [field.name for field in table.c]
 
+    # url_type __upload__ is a special condition to trigger uploading of a file
+    # to the datastore when the filename has not changed but the file has been
+    # newly updated and so needs to be loaded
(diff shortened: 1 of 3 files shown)


wardi:
Instead of toggling the url_type value back and forth can we add an obvious flag to the resource that we can test for but that doesn't get saved in the db? I'm thinking `resource['__uploaded_file_changed'] = True`

wardi:
This fix should go in. While it's a hack, it does have enough comments to make what's going on clear. I just ask that the '__upload__' value be changed. Something like 'upload_again_trigger' says more to me than dunder-upload in the url_type.

@tobes let me know if you don't have time, I'd like this in 2.4 and for backporting

tobes:
I'll try to make the change in the next day or two

Thanks

wardi:
@tobes can we modify the resource's last_modified value as suggested in #1874 instead? Do you know if there's a reason we're not doing that?

wardi:
Closing due to lack of activity. Please reopen if you'd like to finish this off,

Issue #2057, opened 2014-11-18
Really long file names kill the datapusher


tobes:
if you upload a resource in CKAN and the file name is very long (over 90ish chars) then the datapusher gets called with a url that causes a 500 error  (the url contains a truncated filename).

I am not sure if the url is incorrectly produced or that the datapusher has issues around trying to get the extension for the file type.

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #2149, opened 2014-12-17
[#2057] Fix filename munging

tobes:
Issue #2057 

If we upload a file with a huge name then the extension gets eaten and we end up with the datastore not being able to import the file correctly if the format has not been set for the resource

Also it seems browsers on windows 7 seem to pass the whole path along with the file name so this creates problems for these users.

So we remove the path and try to keep the extension - really long extensions (over 20 chars) get truncated but these are probably erroneous extensions anyway.

### DIFF
--- a/ckan/lib/munge.py
+++ b/ckan/lib/munge.py
@@ -4,6 +4,7 @@
 # improved.
 
 import re
+import os.path
 
 from ckan import model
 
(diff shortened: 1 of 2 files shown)


seanh:
@davidread Looks like you're reviewing this? Shall I assign to you for final review and merge? Probably no point in us both looking at it if you have already

davidread:
Sure, reassigning to me.

davidread:
Would you mind simply renaming `munge_filename_ext_safe` to `munge_filename2` or something? And fix the spelling 'extention'.

tobes:
how about changing `munge_filename` to `munge_filename_legacy` and adding a comment to that code and then have the new function as `munge_filename`?

I don't like `fn_name2` style names

davidread:
Yes sounds good

davidread:
I've made some improvements to this in a new PR: https://github.com/ckan/ckan/pull/2257

PR #2151, opened 2014-12-17
Rewrite ckan.logic.converters.convert_to_extras 

### LINKED ISSUE #2150: Custom dataset fields causes IndexError when arbitrary other "extra" fields are used.
Hi,

I followed the instructions here:

http://docs.ckan.org/en/latest/extensions/adding-custom-fields.html#customizing-dataset-and-resource-metadata-fields-using-idatasetform

Unfortunately, when adding new datasets to the catalogue that have further "extra" fields I get the following error:

```
File '/home/vagrant/ckan/lib/default/src/ckan/ckan/lib/navl/dictization_functions.py', line 399 in unflatten
  current_pos = current_pos[key]
```
To cut a (very) long story short, the unflatten function was getting data that included entries such as the following:

```
 ('extras',): [{'key': 'coverage_end_date', 'value': u'2002-09-01'},
               {'key': 'coverage_start_date', 'value': u'2001-09-01'},
               {'key': 'homepage', 'value': u'http://homepage.com/'},
               {'key': 'origin', 'value': u'Origin'}],
 ('extras', 0, 'key'): u'qwe',
 ('extras', 0, 'value'): u'qwe',
 ('extras', 1, 'key'): u'wer',
 ('extras', 1, 'value'): u'wer',
 ('extras', 2, 'key'): u'etr',
 ('extras', 2, 'value'): u'etr',
 ('extras', 3, 'key'): u'rty',
 ('extras', 3, 'value'): u'rty',
 ('extras', 4, 'key'): u'tyu',
 ('extras', 4, 'value'): u'tyu',
 ('extras', 5, 'key'): u'yui',
 ('extras', 5, 'value'): u'yui',
 ('extras', 6, 'key'): u'uio',
 ('extras', 6, 'value'): u'uio',
```
It turns out that the existence of the `('extras', )` key/value pair was interfering with the unflattening process of the `('extras', n, foo)` items. How did such an item come about? It turns out that `ckan.logic.converters.convert_to_extras` is the cause and doesn't work in what I would say is a safe and unsurprising manner (it generates the list of `dict` objects rather than inserts the appropriate `('extras', n, foo)` items into the data object).

Interestingly enough, the related `convert_from_extras` function *does* work in what I would call the expected manner (handling `('extras', n, foo)` items).

I'll submit a patch once I get the ticket number for this issue.

ntoll:
This branch introduces a single change:
- Rewrite the `convert_to_extras` function so it behaves in a way that doesn't interfere with other extras and complements the related convert_from_extras function.

I notice there is no test coverage for these.

Fixes #2150.

### DIFF
--- a/ckan/logic/converters.py
+++ b/ckan/logic/converters.py
@@ -8,6 +8,8 @@ from ckan.common import _
 
 
 def convert_to_extras(key, data, errors, context):
+    if ('extras', ) in data:
+        del data[('extras', )]
     # There is no tally for the number of fields converted to extras.
     extras = [k for k in data.keys() if k[0] == 'extras' and len(k) > 1]
     new_pos = 0
--- a/ckan/logic/converters.py
+++ b/ckan/logic/converters.py
@@ -8,15 +8,14 @@ from ckan.common import _
 
 
 def convert_to_extras(key, data, errors, context):
-
-    # Get the current extras index
-    current_indexes = [k[1] for k in data.keys()
-                       if len(k) > 1 and k[0] == 'extras']
-
-    new_index = max(current_indexes) + 1 if current_indexes else 0
-
-    data[('extras', new_index, 'key')] = key[-1]
-    data[('extras', new_index, 'value')] = data[key]
+    # There is no tally for the number of fields converted to extras.
+    extras = [k for k in data.keys() if k[0] == 'extras' and len(k) > 1]
+    new_pos = 0
+    if extras:
+        extras.sort()
+        new_pos = extras[-1][-2] + 1  # e.g. ('extras', 5, 'value')
+    data[('extras', new_pos, 'key')] = key[-1]
+    data[('extras', new_pos, 'value')] = data[key]
 
 
 def convert_from_extras(key, data, errors, context):
--- a/ckan/new_tests/logic/test_converters.py
+++ b/ckan/new_tests/logic/test_converters.py
@@ -55,6 +55,46 @@ class TestConvertToExtras(unittest.TestCase):
 
         eq_(errors, {})
 
+    def test_convert_to_extras_with_added_extras_output_unflattened(self):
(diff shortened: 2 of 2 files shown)


amercader:
OK, several things going on here :)

I've been having a look at this and could reproduce the issue using your extension (you get the exception when you provide more free extras than custom fields, otherwise free extras will make custom fields disappear (as described in #1894)).

One thing to note is that on 2.2.1 (the one that your instance targets) you can *not* mix custom fields that get internally transformed into extras with free form extras when using the core converter. As @davidread mentioned in the previous comment the docs said otherwise but unfortunately this was a a mistake left from bad merging (this is now fixed #2259).

AFAICT the issue is fixed when using master or the `release-v2.3` branch. The test you provided passes for me in current master if I comment out the `assert not ('extras',) in data` bit.

Looking at patching this in previous versions, I think we agreed that it was too big a change and broke the API, as it required a change in logic to check that the free extras don't clash with the custom ones (for full context see [1]). I'd suggest that you keep using your custom `convert_to_extras` converter on your extension until you can upgrade to 2.3.

Finally, I'm not saying that this patch is no longer relevant, as you mention it may do things in a safer way ( I need to have another look with more time), but is not directly relevant to fix the original exception you reported.

Hope all of this makes sense.

[1] https://lists.okfn.org/pipermail/ckan-dev/2014-October/008170.html

ntoll:
Hey hey... great stuff and glad to hear the problem will be fixed in the next release. When is that scheduled to be?

wardi:
The free-form vs. converted extras problem will be solvable after #3072 is adapted for dataset extras. We'll have a separate namespace available for each plugin so we could easily have these values live side-by-side (and remove the extras table while we're at it)

rossjones:
Looks like this PR is now unnecessary? Suggest we close it.

PR #2154, opened 2014-12-18
[WIP] [#1852] First approach at db-less views

amercader:
@wardi This is based off #1852 and will correctly render views on the resource page based on
`ckan.views.default_views`.

Problems start when you try to manage those, eg when editing one oft hem, as we don't know the view type.

I'm not super optimistic about this approach, but well worth a try :)


amercader:
Abandoned

Issue #2140, opened 2014-12-11
Resource Views - UX review


johnmartin:
## Major issues
1. Setup is a little fiddly: requires `datastore` + `datapusher` setup and there's not really any documentation saying that. Also there appears to be no helper text within the product to tell you what the different view types do, eg. what's difference between Datagrid + table)
2. It's a little confusing that views don't have defaults. E.g. when I add a new resource I don't just get a best guess view for it (means I have to create a new view for that resource). Is there any way a default view can automatically generated for each resource? Or could that be a config option?
3. Documentation for creating your own resource view - There doesn't appear to be any. How do I add one to the dropdown list? How do I make it so that CKAN can have a preview before saving? How do I add config options? etc. As a side note: the documentation that is there is a little out of date: http://docs.ckan.org/en/latest/maintaining/data-viewer.html (or it will be when 2.3 is released)

## Minor issues
1. It would be nice if the embed modal had a copy button (implementation suggestion: https://github.com/zeroclipboard/zeroclipboard)
2. Delete view does not have confirm dialog (should use standard CKAN `confirm-action` JS module)
3. Plain text view doesn't exist any more (eg `.txt`, `.json` & `.xml` will all now not have a view)
4. Also confusing as to why as a core extension we have all the recline style graphs/tables AND the lightweight extensions (I'd just choose recline, as it does it good enough for now)
5. Sorting views - the drag points, I think were from a different view, so probably needs updating for 2.3 (completely minor thing)

seanh:
> @brew regarding the messages (screenshot below for others), I wouldn't show the detailed info message always, to all users. At most, I'd show them to users who can actually create or edit views on the resource (ie h.check_access('resource_view_create')), that is essentially publishers which will be the ones who'll need to care about the views rendering fine. To the rest of general users I'd show a generic "There are no views available for this resource" only where there are none.
> 
> @wardi @davidread @rossjones @seanh what do you think?

Agree. The long error message with all the bullet points is a great example of a good error message that actually explains what might have gone wrong and suggests to the user what they can do about it. But there's no point in showing this to users who only have read access and can't do the suggested things.

There's a slight worry about causing confusion by showing different error messages to different users, if it can be presented in such a way that it's clear that the sysadmin is seeing the same error plus some extra info, that'll be good.

PR #2179, opened 2014-12-31
Resource view information, warnings and re-enabled DataProxy

brew:
Related to #2140. This PR:

- re-enables the DataProxy for recline_view (Data Explorer). 
- adds an information reveal for resource view creators detailing possible reasons they may not be seeing the views they expect. 
- adds a warning when creating recline_view views in the absence of the DataStore (that would display using the DataProxy), to warn the view may be slow and unreliable.


rossjones:
Looks good. Has been merged.

Issue #2188, opened 2015-01-07
Deleted groups remain visible by direct URL access


jammycakes:
It appears that after deleting a group, you can still access it if you know its exact URL.

To replicate:

 * Create a new group at http://localhost:5000/group/new - call it "test" with a url of localhost:5000/group/test
 * Click "Edit" on the group's home page then "Delete". The group will disappear from the list of groups at http://localhost:5000/group as expected; the "state" value of the corresponding row in the "group" table in the database will be set to "deleted".
 * *Unexpected:* Return to http://localhost:5000/group/test - the group will still be visible. The activity stream will also be visible (showing no record of the deletion) as will the "About" page.
 * *Unexpected:* The deleted group remains visible even to a logged-out user (as can be verified, for example, using a private browsing session in Firefox or Chrome).

CKAN version affected: 2.1.3. I have not yet determined whether other CKAN versions are affected.

Aaron-M:
Same experience on CKAN version 2.2.1

rossjones:
This has been fixed, see #785 for solution to general 403/404 error.

Issue #785, opened 2013-04-18
When a dataset is deleted the package_show action returns 403


teleyinex:
I guess that it should return 404 but I get 403 in datahub.io. Is this the expected behavior?

rossjones:
This issue is being closed as it is more than 12 months old, it is now being addressed in #2805

PR #2190, opened 2015-01-08
[#2188] Do not allow public to view deleted groups. 

davidread:
Fixes: #2188 

Also, show in template when a pkg/group/org is deleted. These pages are only visible by a sysadmin (or the org editor/admin or group member) when they go to the direct URL (or via the trash for packages). But it it is helpful to see they are deleted, to avoid confusion.

### DIFF
--- a/ckan/logic/auth/__init__.py
+++ b/ckan/logic/auth/__init__.py
@@ -16,7 +16,7 @@ def _get_object(context, data_dict, name, class_name):
             data_dict = {}
         id = data_dict.get('id', None)
         if not id:
-            raise logic.ValidationError('Missig id, can not get {0} object'
+            raise logic.ValidationError('Missing id, can not get {0} object'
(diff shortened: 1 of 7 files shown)


jammycakes:
Just one observation here -- it looks like this will show the login page when an unauthorised user requests a deleted group. Is this intentional? I'd personally have thought that it would be better to return 404 Not Found in that case, just as for when the user requests a group that never existed in the first place.

davidread:
It's a complicated issue with a number of obscure use cases. I think I agree overall, of giving 404 to public users at least. But what I've done follows what we already do for deleted datasets.

Feel free to open another ticket for changing them both and working through the use cases.

rossjones:
Looks good.  We've 3 other tickets for showing how to 404 instead of 403 for most users - and a description of how to fix at https://github.com/ckan/ckan/issues/785

PR #2191, opened 2015-01-08
Added commands for removing nginx and apache default sites

mattrm:
If you just follow the guide, it won't work as you'll always have a conflict with apache or nginx default sites.  So they need to be disabled.


mattrm:
It might.  But you have to remove the default site for CKAN to work.  So it can't be helped.

wardi:
@mattrm this looks good, thanks.

PR #2195, opened 2015-01-08
[#2194] Fixes NameError for related page number handling

### LINKED ISSUE #2194: /related gives Internal Server Error
E.g. http://master.ckan.org/related

Looking at the error log, there's a  NameError: name 'request' is not defined in line 38 of related.py.

This seems to have snuck in when #2062 was merged.

jqnatividad:
Fixes #2194


rossjones:
Obviously not caught by any sort of test, but as it is sorta-almost-deprecated happy to take this fix without one (there wasn't one there before either).

PR #2198, opened 2015-01-09
URL encode call_action_api parameters

razz0:
CKAN's call_action_api function is useful for testing custom extensions through API but we came across problems with some parameters that should be URL encoded. 

ckan.tests.call_action_api only converts input data to a JSON string, but does not URL encode it. This results in plus (+) signs being converted to spaces using this function (and very likely other issues). This added 'urllib.quote' should fix the issue.

### DIFF
--- a/ckan/tests/__init__.py
+++ b/ckan/tests/__init__.py
@@ -15,6 +15,7 @@ from unittest import TestCase
 from nose.tools import assert_equal, assert_not_equal, make_decorator
 from nose.plugins.skip import SkipTest
 import time
+import urllib
 
 from pylons import config
 from pylons.test import pylonsapp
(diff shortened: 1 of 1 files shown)


wardi:
@razz0 would you provide an example of some code that's failing?

If you're using this from an extension's tests you might prefer [ckanapi.TestAppCKAN](https://github.com/ckan/ckanapi/#testappckan)

razz0:
@wardi Thanks for pointing out **ckanapi**. Using it solves this problem for our extension.

CKAN tests still rely on the `call_action_api`, which may produce unexpected results for any `**kwargs` input with characters requiring URL encoding. But as the test inputs are not in need of URL encoding now, I wonder if fixing this is necessary and whether the encoding should rather be left for `webtest.TestApp.post` to handle like in **ckanapi**?

wardi:
closing as this is a change in the legacy tests and ckanapi is a better way to call the API

Issue #2199, opened 2015-01-09
organization_list_for_user doesn't use hierarchy


davidread:
Action function `organization_list_for_user` returns a list of organizations that the user has permission for, and is used for the drop-down in the forms when creating a dataset or harvest source. However it doesn't take account that permissions 'cascade' to organizations lower in the hierarchy, and these child organizations have not been returned, when they should have been.

PR #2200, opened 2015-01-09
[#2199] Add permission cascading to the org list.

davidread:
Issue: https://github.com/ckan/ckan/issues/2199


davidread:
Please do review this @joetsoi when you have a chance

PR #2206, opened 2015-01-12
group_show, organization_show: exclude datasets by default

wardi:
Prefer to not include datasets for the common case of just querying the group/org information (performance win). This brings the `include_datasets` parameter default in line with `include_datasets` parameter on user_show and `include_tags`, `include_groups` and `include_extras` in group_list and organization_list.


wardi:
Woo!

PR #2208, opened 2015-01-12
solr index includes two copies of validated_data_dict

wardi:


wardi:
This was fixed in e6bcc7288 but that commit didn't make it to master, so I fixed the test to actually show the regression and cherry picked the fix.

PR #2209, opened 2015-01-12
tag_show, vocabulary_show, vocabulary_list: exclude datasets by default

wardi:
This is a real performance problem for sites using tag vocabularies, especially with index_package calling `vocabulary_show` for each vocabulary tag in the dataset, which calls `tag_dictize` for each tag in the vocabulary, which pulls up to 1000 datasets from solr.

That's (vocab tags in dataset) * (tags in same vocab) * (up to 1000) package dicts generated just to get the vocabulary name. Yeah.

Like my other recent PRs this adds an `include_datasets` parameter to the API call, with a new default of False.


wardi:
#1886 fixes this better

Issue #1551, opened 2014-02-26
Meaningless 'source' field


mwainwright:
The dataset creation process has a 'source' field that has no apparent meaning.

The only clue is that the example text pre-populating it is a URL with a .json extension. This suggests a data file, but in CKAN, these are added as resources.

(Background: this appears to be new in 2.2. A very old version of CKAN had a 'source' field that was intended for a link to a web page about the data, not the actual data. It was rarely used and, because of the bad naming, even more rarely understood.)

rossjones:
This issue is being closed as it is more than 18 months old.  If you are still experiencing this problem and wish to help investigate further, or submit a PR, please feel free to re-open it.

PR #2107, opened 2015-01-14
Source field linkifies even if not a URI scheme

amercader:
À propos to #1551, when the 'Source' field's metadata value is just text (i.e. not a [URI Scheme](http://www.iana.org/assignments/uri-schemes/uri-schemes.xml)), it's created as a link. This isn't exactly ideal.


amercader:
@wardi I added these quick changes based on your #2085 PR, can you have a look when you have a moment?

PR #2214, opened 2015-01-14
Fixes #2210 - group_create should return group_show not group_dictize

### LINKED ISSUE #2210: group_create does not return results of group_show
When calling group_create, the action layer returns the result of calling group_dictize (which ignores the schema). The same is also true for group_list.

When calling package_create it returns the result of package_show.

It feels like it should be more consistent so that group_create definitively returns the same as group_show.

rossjones:
Instead of calling group_dictize and returning the results at the end of group_create this now returns group_show instead for consistency.  This ensures the show schema is used to validate and process the result returned to the user.  It should make it easier to call create, modify the response and then post to update.

Instead of using group_list_dictize the group_list call now uses group_show for each group in the list to return the relevant data. This will be marginally slower as it performs the schema checks which weren't previously checked.

The deprecated use of packages as either a list or a count has been removed as package_count is already returned on each group_show call.

This is more than likely a breaking change and will need close checks to make sure it is accurate.


ckan/logic/action/get.py
```diff
@@ -413,22 +414,16 @@ def _group_or_org_list(context, data_dict, is_org=False):
...
-        include_tags = False
-    # even if we are not going to return all_fields, we need to dictize all the
-    # groups so that we can sort by any field.
-    group_list = model_dictize.group_list_dictize(
-        groups, context,
-        sort_key=lambda x: x[sort_info[0][0]],
-        reverse=sort_info[0][1] == 'desc',
-        with_package_counts=all_fields or
-        sort_info[0][0] in ('packages', 'package_count'),
-        include_groups=asbool(data_dict.get('include_groups', False)),
```
wardi: include_groups is no longer an option, it should be removed from the docstrings

ckan/logic/action/get.py
```diff
@@ -1141,11 +1136,10 @@ def _group_or_org_show(context, data_dict, is_org=False):
     group = model.Group.get(id)
     context['group'] = group
 
-    include_datasets = data_dict.get('include_datasets', True)
+    include_datasets = data_dict.get('include_datasets', False)
```
wardi: This change needs to be in the docstring too

PR #2221, opened 2015-01-15
[#1753] Rename tests -> tests_legacy.

davidread:
I think we can worry about renaming new_tests to tests once this has been merged, to reduce the chance of "noise between branches" mentioned here: https://github.com/ckan/ckan/issues/1753#issuecomment-45594112


wardi:
@davidread to prepare for renaming new_tests how about moving to `tests/legacy` instead of `tests_legacy` so we don't end up with two test dirs in the top level?

davidread:
Sounds good, i'll give it a try.

davidread:
@amercader you mentioned before you'd be up for renaming tests in CKAN 2.3. I think this PR is pretty straightforward, and the longer it is not merged, the harder it will be to merge, since it affects every legacy test. So can this go in the 2.3 milestone?

Once this is merged I'll do a PR for moving new_tests to tests, and ideally that'd go into the same CKAN release. That would mean that tests in extensions will only need updating once.

davidread:
Whoop!

PR #2222, opened 2015-01-16
Resource views formatting lost on master and 2.3

amercader:
The list of available views should look like this:
fzlj8ot

But it looks like this:
jfw6s4l

This was fixed back in October on #1920 but there probably has been a wrong merge on the way.


rossjones:
screenshots look the same to me.

amercader:
sorry, I was being slow, fixed now. Anyway, I pushed the fix if you want to review it :)

rossjones:
Yeah, looks much better :)

PR #2223, opened 2015-01-16
Moves the cli commands to a module

rossjones:
Moves all of the cli commands from a single file to a file-per-command in the ckan.lib.commands module.

Unfortunately removing cli.py from the pep whitelist has made the pep8 command now start checking a lot of files that it didn't before - I suspect a bug in the pep8 test somewhere.


ckan/tests/test_coding_standards.py
```diff
@@ -953,4 +955,4 @@ def test_bad(self):
                ' resolving\nWe should not be translating exceptions in most'
                ' situations.  We need to when the exception message is passed'
                ' to the front end for example validation')
-        show_fails(msg, self.fails)
+        show_fails(msg, self.fails)
```
davidread: Found it - looks like you got rid of the final \n, which is what the PEP8 error is about. I'll commit the fix.

davidread:
Hmmm I suspect this isn't going to be finished off and may not be worth merging all the updates since this was first worked on. Closing

Issue #2125, opened 2014-12-08
Add organization autocomplete action


mwahlroos:
I'd like to see an organization_autocomplete action that can be accessed over the API. I'm not sure if there's a direct use case in CKAN core at the moment but autocompletion could be useful for API users or extensions.

I should be able to contribute.

wardi:
fixed by #2226

PR #2226, opened 2015-01-16
#2125 add organization autocomplete

davidread:
Based on: https://github.com/ckan/ckan/pull/2134 but with new_tests


davidread:
Tests now pass, ready for review @wardi

PR #2227, opened 2015-01-19
Remove user id from redirect to dashboard after logging in.

bellisk:
After logging in to CKAN, the user is redirected to their own dashboard, with a url like this: myckaninstance.org/dashboard?__no_cache__=True&id=admin. This involves sending the user ID as a GET parameter, which was pointed out in a penetration test as a potential security issue.

The user ID is not necessary to render user/dashboard.html, so I have removed it from the redirect.


george-sattler:
I've given this a test and checked it through. Looks good to merge.

PR #2229, opened 2015-01-19
Autocomplete double encodes search terms

Zharktas:
Autocomplete.js double encodes search terms and therefore doesn't find tags that have alphabets that need encoding for example scandivavian å, ä and ö.

This pull request removes the second encoding.


amercader:
thanks for the patch @Zharktas

PR #2235, opened 2015-01-22
1940 check old password on user password update

george-sattler:
Basically, this adds a field to the UI and...
- doesn't attempt to validate anything unless there's at least something entered in the new password field (and the confirmation field so if it's some other user details that are being edited, it won't try to verify
- reports the error to the user
- if old password is incorrect, nothing proceeds to the action layer (regarding Adria's comment on the issue, I think that hopefully was the intention).


ckan/controllers/user.py
```diff
@@ -324,6 +326,15 @@ def _save_edit(self, id, context):
             context['message'] = data_dict.get('log_message', '')
             data_dict['id'] = id
 
+            if data_dict['password1'] and data_dict['password2']:
+                print 'got values in p1 and p2'
```
rossjones: It's normally me that leaves print statements in  :)

george-sattler: d'oh! and Friday was going so well already :)

george-sattler:
The Travis CI build was passing before I removed the print statement that I'd left in there and I can't seem to see in the travis job log what precisely it is that's the problem. @rossjones would you be able to point out to me what the error is?

rossjones:
It's the old datastore_active problem that *occassionally* appears - even though I'm fairly sure it was fixed.  The datastore plugin adds something to the resource on package_show (not as an extra) and it doesn't always fire. 

I've restarted the job

rossjones:
Any chance of a new-style test?  They're the tests in new_tests.

george-sattler:
yes, sure thing. I wasn't aware there were style tests.

rossjones:
I meant the tests done in the new fashion, - https://github.com/ckan/ckan/tree/master/ckan/new_tests - rather than testing styles.

george-sattler:
@rossjones ah, yes it's so obvious now what you meant. 

Could you suggest what the test might look like? Following Adria's note on the issue thread for this one, i've implemented by doing a check in the controller before it gets to calling the action. So this would be a test of the controller? i.e. I would need to create a new file: new_tests/controllers/test_user.py (since there doesn't already appear to be one?

I would also need to log the user in somehow to then test the password update. Is this possible to do in code?

rossjones:
@george-sattler sorry for the late reply - https://github.com/ckan/ckan/blob/master/ckan/new_tests/controllers/test_package.py is probably the most useful, as it shows a user being created with the factories and calling the controller actions.

george-sattler:
Thanks, @rossjones , was able to work it out from some examples in test_package.py

ckan/templates/user/edit_user_form.html
```diff
@@ -1,11 +1,10 @@
 {% import 'macros/form.html' as form %}
 
-<form class="dataset-form form-horizontal" method="post" action="{{ action }}">
+<form id="user-edit" class="dataset-form form-horizontal" method="post" action="{{ action }}">
```
brew: @joetsoi recently submitted [a pull request that adds ids to form tags](https://github.com/ckan/ckan/pull/2330/files) for testing purposes. This should probably follow the same naming convention, i.e. the id should be `user-edit-form`.

ckan/controllers/user.py
```diff
@@ -341,6 +351,10 @@ def _save_edit(self, id, context):
             errors = e.error_dict
             error_summary = e.error_summary
             return self.edit(id, data_dict, errors, error_summary)
+        except UsernamePasswordError:
+            errors = {'oldpassword': [u'Password entered was incorrect']}
+            error_summary = {u'Old Password': u'incorrect password'}
```
brew: Do these strings need to translatable, wrapped in `_( )`?

joetsoi:
@george-sattler do you mind making these small changes that @brew suggested and I'll merge after.

george-sattler:
@joetsoi no probs  - i'm on it

amercader:
Thanks for the patch @george-sattler

Issue #2236, opened 2015-01-22
Package state is not translated in template package/snippets/additional_info.html


andreivicentiuhincu:
Hello, the state of the dataset is not translated, at line 57: <td class="dataset-details">{{ pkg_dict.state }}</td>. It should be <td class="dataset-details">{{ _(pkg_dict.state) }}</td>. I'll make a pull request and fix it.

PR #2237, opened 2015-01-22
Fixed dataset state translation

andreivicentiuhincu:
#2236 - The dataset state was not translated on the dataset details
page....Fixed it.


amercader:
Good catch, thanks @andreivicentiuhincu

Issue #2232, opened 2015-01-20
Connection pool configuration


pstarace:
Hi all

I found that the connection pool for sqlalchemy is instantiated in ckan/config/environment.py, row 355:
```
engine = sqlalchemy.engine_from_config(config, 'sqlalchemy.', **extras)
```
So it should be feasible to inject pooling configuration adding new properties in the ini config file, something like:
```
sqlalchemy.pool_size=10
sqlalchemy.max_overflow=20
```
Can you please confirm?

I'd like also to do the same for datastore, but I found the current code in ckan/ckanext/datastore/db.py doesn't support extra configuration:
```
engine = sqlalchemy.create_engine(connection_url)
```
In this case the pool is created using the default params (pool_size=5, max_overflow=10).
Do you know any drawback adding new properties in the ini file:
```
ckan.datastore.pool_size=10
ckan.datastore.max_overflow=20
```
and loading them using:
```
import pylons
try:
    extras = {'pool_size': int(pylons.config['ckan.datastore.pool_size']), 
        'max_overflow' : int(pylons.config['ckan.datastore.max_overflow'])}
except Exception:
    extras = {'pool_size': 5, 'max_overflow' : 10}
engine = sqlalchemy.create_engine(connection_url, **extras)
```
?

Thanks
Paolo

davidread:
Yes I believe you can pass in parameters to the sqla enginer using config settings like `sqlalchemy.pool_size`. I remember playing with the pool like this before successfully. (The one that didn't work was `poolclass` as you can only provide a string in the config file, rather than an actual class, but it sounds like you're not interested in that).

It sounds like a good plan to do the same for datastore. However I'd suggest adding .sqlalchemy in the keys, to keep these options in their own namespace. So you'd have: `ckan.datastore.sqlalchemy.pool_size`
And rather than catch all exceptions it would be better to catch a specific one. Or you could perhaps just use pylons.config.get and supply the default. Anyway, please do submit a Pull Request.

amercader:
Fixed in #2244

PR #2244, opened 2015-01-28
Update db.py

pstarace:
That change enables the user to provide custom sqlalchemy properties through the config file using the 'ckan.datastore.sqlalchemy.' prefix.
Example:
```
ckan.datastore.sqlalchemy.pool_size=10
ckan.datastore.sqlalchemy.max_overflow=20
```

The 'url' property is passed as param in the 'extras' map to avoid changing the structure of the current config file.

For further details please look at https://github.com/ckan/ckan/issues/2232

### DIFF
--- a/ckanext/datastore/db.py
+++ b/ckanext/datastore/db.py
@@ -104,7 +104,9 @@ def _get_engine(data_dict):
     if not engine:
         import pylons
         extras = {'url': connection_url}
-        engine = sqlalchemy.engine_from_config(pylons.config, 'ckan.datastore.sqlalchemy.', **extras)
+        engine = sqlalchemy.engine_from_config(pylons.config,
+                                               'ckan.datastore.sqlalchemy.', 
+                                               **extras)
         _engines[connection_url] = engine
     return engine
 
--- a/ckanext/datastore/db.py
+++ b/ckanext/datastore/db.py
@@ -105,7 +105,7 @@ def _get_engine(data_dict):
         import pylons
         extras = {'url': connection_url}
         engine = sqlalchemy.engine_from_config(pylons.config,
-                                               'ckan.datastore.sqlalchemy.', 
+                                               'ckan.datastore.sqlalchemy.',
                                                **extras)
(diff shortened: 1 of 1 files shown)


ckanext/datastore/db.py
```diff
@@ -102,7 +102,9 @@ def _get_engine(data_dict):
     engine = _engines.get(connection_url)
 
     if not engine:
-        engine = sqlalchemy.create_engine(connection_url)
+        import pylons
+        extras = {'url': connection_url}
+        engine = sqlalchemy.engine_from_config(pylons.config, 'ckan.datastore.sqlalchemy.', **extras)
```
rossjones: This is causing the pep8 tests to fail :( 

Probably wants to be 
```
        engine = sqlalchemy.engine_from_config(pylons.config, 
                                               'ckan.datastore.sqlalchemy.', 
                                                **extras)
```

but I can't push to your branch to fix it. Sorry.

pstarace: Hi Ross, what's the issue with that line?
Thanks
Paolo

rossjones: The tests run pep8 against the code, which complains about lines longer than 80 chars.  I know, I know, I'm not a fan either :(

amercader:
This looks good @pstarace, can you just document the config option on the configuration section?

Just under this one should be fine:

https://github.com/ckan/ckan/blob/master/doc/maintaining/configuration.rst#ckandatastoreread_url

amercader:
Also there are PEP8 errors that are making the tests fail, can you fix those as well:

```
-----------------------
ckanext/datastore/db.py
-----------------------
W291 ln:108 trailing whitespace
```

pstarace:
Hi @amercader the change to the config file has been submitted.
What's the issue with PEP8? Tried to remove white space at the end of line 108...

amercader:
Thanks for the patch @pstarace. 

Just so you know, if you push new commits to the same branch on your local repo, they will automatically added to this PR, so there's no need to create a new one like #2279

PR #2204, opened 2015-01-28
Improve command line migration / creation of resource views

amercader:
New resource views require a database object to display a particular visualization on the resource page (until we can find an optimization like the ones approached on #2154 ). To manage this, an admin tool is required, for instance to create the db objects when migrating from a previous version.

The [current implementation](https://github.com/ckan/ckan/blob/master/ckan/lib/cli.py#L2155) presents some drawbacks, it only works for a limited number of view plugins and has db queries hardcoded for each particular case. To support changes in #1852 to define default views, the `create` command will need to iterate over all datasets (or a subset of them) to check if the provided view types (or the default ones) are suitable for each of them. This is likely to be time consuming on large instances, but hopefully we can add functionality to help with this.

My proposed implementation is to use the search index to get all datasets rather than the db, because:

1. It is significantly faster
2. It returns the `data_dict`, required by the view plugins to decided if they can render a view for a particular resource, so we don't need to dictize.
3. Allows to filter the datasets to check by different criteria

So, it could work something like this (assuming the general crappiness of `cli.py` allows it):

This will create the views defined in `ckan.views.default_views` on all suitable datasets:

    paster views create

You can also specify view types:

    paster views create image_view recline_view

To limit the amount of datasets to process, you can filter by a particular format:

    paster views create --format pdf pdf_view

Internally this would use `fq=res_format:pdf` on the `package_search` call. We could add further helper options for other useful stuff (eg groups, orgs).

We can also provide support for passing an arbitrary search dict (the same you would pass to `package_search`) so users can tweak the number of datasets to process, but not sure how easy would be this from the command line.

It is not ideal, but this will allow maintainers of really big instances to plan and fine tune the migration process so it is less painful.


amercader:
Summary so far for the reviewer:

* `paster views create` is used to create the resource views objects on the db for those resources / datasets where it's necessary

* All variants of the command use the dataset search to get all necessary datasets and pass the data_dict to the `datapreview.py` functions that check if views need to be created (it is ultimately each plugin which decides it, via the `can_view()` method).

* If no view types are provided (`paster views create`), the default ones are used (these are defined on `ckan.views.default_views` or a hardcoded default.

* You can provide specific view types: `paster views create image_view recline_view pdf_view`

* For certain view types (the ones on core), default filters are applied to the search to only get relevant resources. For instance if `pdf_view` is defined, `fq:res_format:PDF` will be added to the search params. This can be disabled.

* Apart from this, users can provide arbitrary search params like the ones supported by `package_search`, eg:

```
paster views create geojson_view -s '{"fq": "res_format:GEOJSON"}'

paster views create '{"q": "groups:visualization-examples"}'
```

* For convenience, there is an option to create views on a particular dataset or datasets

```
paster views create -d dataset_id

paster views create -d dataset_name -d dataset_name
```

There are pretty comprehensive docs in the help prompt and the source code, the ones in the actual documentation are coming.

amercader:
@wardi I did some benchmarking to test how long it could take to run the command. For my tests I created X datasets with 6 resources, each with a random format (`['CSV', 'PNG', 'PDF', 'TXT', 'GEOJSON', 'XXX']`. CSV, PNG, PDF and TXT are part of core view plugins, so default filters are applied to them. GEOJSON was recognized by an external view plugin, and XXX didn't have any view available.

First column shows checking only for views included in core [1], second one also includes an external view, so default filters are not applied.

number of datasets | core view plugins | external view plugins
-------------------|-------------------|----------------------
10 | 1.7 s | 1.8 s
100 | 6.5 s | 6.7 s
1000 | 1:03 | 1:03
10000 | 20:32 | 27:36

Of course that looks like it could not scale nicely once you get pass the 100k datasets. But then I'd argue that you'd need to plan in advance ([faceting](http://catalog.data.gov/api/action/package_search?facet.field=res_format) [by](http://data.gov.uk/api/action/package_search?facet.field=%5B%22res_format%22%5D) [format](http://open.canada.ca/data/en/api/action/package_search?facet.field=%5B%22res_format%22%5D)) and play a bit with the search parameters filtering on the `paster views create` command to filter out any resources you are not interested in or vice-versa and do some counts beforehand.

I wrote some guidance on how to that on #2205 ([link](https://github.com/ckan/ckan/pull/2205/files?unchanged=expanded#diff-f98f3f854827f0ac3f4e60dcd954c792R246)), which is still WIP.


[1] `time paster views create image_view recline_view pdf_view text_view -y`
[2] `time paster views create image_view recline_view pdf_view text_view geoview -y`

seanh:
If I understand it right, basically what this does is:

* Do a package search to get a list of datasets
* Add the default views to all the resources of those datasets

I guess `paster views create` can't just be an action function, because it might take a long time to execute?

But maybe "add default views to this resource" and "add default views to all of this dataset's resources" should be quick enough that they can be API actions?

Then all you need is an API client script to call the package search API to get a list of datasets, then iterate over them all calling the new "add default views to all of this dataset's resources" on each dataset.

I guess this API client script could be implemented as a paster command. But I sort of wish we could do it in a way that meant it didn't have to be run from the same server that CKAN is on. So just a standalone script. But then if I just want to run this script on my laptop then I don't necessarily want to install CKAN on my laptop just to get the script, so then I think it should be a separate repo like e.g. ckanapi-exporter. Maybe that's going too far and just a paster command is fine.

seanh:
Sorry I can't be more productive, ran out of time. I'll be working on Salford next two days but should be able to talk about this at the same time

amercader:
> But maybe "add default views to this resource" and "add default views to all of this dataset's resources" should be quick enough that they can be API actions?

These are already separate [functions](https://github.com/ckan/ckan/blob/2204-resource-views-cli/ckan/lib/datapreview.py#L339:L390) on `lib/datapreview.py`. They could be moved to actions if necessary though.

> Then all you need is an API client script to call the package search API to get a list of datasets, then iterate over them all calling the new "add default views to all of this dataset's resources" on each dataset.

This is basically what the current command at `cli.py` does. The bulk of the code (`_get_view_plugins()`, `_add_default_filters()`, `_update_search_params()` and `search_datasets()`) are basically helper functions to build the relevant search params for `package_search` depending on user input and paster cli boilerplate. Once we get the list of datasets it's just a [simple iteration](https://github.com/ckan/ckan/pull/2204/files#diff-530e98d83857b6bd57795518ad4c2dbdR2456) over the datasets and calling  "add default views to all of this dataset's resources" for each of them.

Depending on the number of datasets, this command can take a while, so moving everything to an action doesn't sound like a good idea, until we have the generic queue framework for handling async jobs. In this case it'd make sense to just ask the remote CKAN to create all views, get the job token and query in a while whether it has finished or not, but with an API action it will probably time out (just like eg `paster db upgrade` would on a moderately big db).

I think that realistically, for 2.3 CLI is not that bad a place to have this stuff, but I'm happy to refactor if there are suggestions, and definitely make it available to the API when we have an async queue.

@wardi @joetsoi what do you think?

joetsoi:
I think it's ok for this code to live in cli.py, much of the branching just involves trying to make the output easier for the user "do you want to continue yes/no", and I'm always cautious of while(true) loops.

I think some of the if statements could be cleared up to be made easier to read, unless I'm reading the code wrong.

ckan/lib/cli.py
```diff
@@ -2152,38 +2156,332 @@ def command(self):
...
+        n = self._page_size
+
+        search_data_dict = {
+            'q': '',
+            'fq': '',
+            'fq_list': [],
+            'rows': n,
+            'start': n * (page - 1),
+        }
+
```
joetsoi: minor readibility thing, could these 3 if blocks not just be reveresed and made into an if/elif/elif?
```
if self.options.dataset_id
elif self.options.search_params
elif not self.options.no_default_filters
```
unless there is a case that I'm over looking? I just get easily confused when there are many conjunctions and negations to consider in if statements.

ckan/lib/cli.py
```diff
@@ -2152,38 +2156,332 @@ def command(self):
...
+        if (self.options.search_params and
+                not self.options.dataset_id):
+            self._update_search_params(search_data_dict)
+
+        if self.options.dataset_id:
+            search_data_dict['q'] = ' OR '.join(
+                ['id:{0} OR name:"{0}"'.format(dataset_id)
+                 for dataset_id in self.options.dataset_id]
+            )
+
```
joetsoi: minor readibility thing, could these 3 if blocks not just be reveresed and made into an if/elif/elif?
```
if self.options.dataset_id
elif self.options.search_params
elif not self.options.no_default_filters
```
unless there is a case that I'm over looking? I just get easily confused when there are many conjunctions and negations to consider in if statements.

amercader:
@joetsoi you are absolutely right, fixed now.

I'm marking as WIP because I'll add a quick method for clear resource views

seanh:
I'm happy for this to be merged if you guys are happy with it

wardi:
@amercader I really like @seanh 's suggestion of making these actions available by the API. It sounds very useful to be able to reset dataset resources to default views, and run long batch processes remotely or locally.

amercader:
@wardi just to double check, do you mean the "add default views to this resource" and "add default views to all of this dataset's resources" actions or "create default views for (all) datasets"

wardi:
@amercader "add default views to this resource", "add default views to this dataset" only.

ckan/lib/cli.py
```diff
@@ -2152,230 +2156,355 @@ def command(self):
...
-            resource_view = {'title': 'PDF View',
-                             'description': 'PDF view of the resource.',
-                             'resource_id': resource.id,
-                             'view_type': 'pdf'}
+        context = {'user': self.site_user['name']}
 
-            logic.get_action('resource_view_create')(context, resource_view)
+        page = 1
+        while True:
+            query = self._search_datasets(page, loaded_view_plugins)
```
deniszgonjanin: I think because this is going through solr, it fails silently for recline views (ckan/ckanext/reclineview/plugin.py:61), since `data_dict['resource']['datastore_active']` does not exist in the resource dict when it is fetched from solr.

amercader: You are absolutely right @deniszgonjanin. We need to implement #2234 to store `data_dict['resource']['datastore_active']` properly.

amercader:
@joetsoi @wardi shiny new `resource_create_default_resource_views` and
`package_create_default_resource_views` actions now available.

Removed WIP

joetsoi:
A well a everybody's heard about the merge
M-m-m merge, merge, merge, m-merge is the word

ckan/lib/cli.py
amercader: For the record, @deniszgonjanin I think that the issue you mention had more to do with #2159 than #2234

PR #2248, opened 2015-01-28
Allow normal users (non sysadmins) to change and remove owner_org from datasets

### LINKED ISSUE #2218: Enable a logged in user to move dataset to another organization
In #953 this change was introduced:

> User is logged in, and is a member of an organization, and dataset is a
member of any org: Visibility shows but Organization does not show when
updating a dataset.

So, this prevents a logged in user to change dataset's organization.

Why can't a legitimate user to move datasets around organizations?

brew:
This PR allows normal users, who have available organizations, to change the `owner_org` property for datasets they can edit using the web ui. The ability to change `owner_org` was previously only available via the API.

Normal users can also now remove the value of `owner_org` from their datasets (if `ckan.auth.create_unowned_dataset` is `True`), both using the web ui and API. Previously, this was only available to sysadmins.

The commits make the organization dropdown visible to normal users when creating/editing a dataset. And change the `owner_org_validator` method that previously restricted `owner_org` removal to just sysadmins.

Fixes #2218.

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -34,13 +34,9 @@ def owner_org_validator(key, data, errors, context):
     model = context['model']
     user = context['user']
     user = model.User.get(user)
-    if value == '' :
+    if value == '':
         if not new_authz.check_config_permission('create_unowned_dataset'):
             raise Invalid(_('A organization must be supplied'))
-        package = context.get('package')
-        # only sysadmins can remove datasets from org
-        if package and package.owner_org and not user.sysadmin:
-            raise Invalid(_('You cannot remove a dataset from an existing organization'))
         return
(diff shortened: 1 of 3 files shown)


davidread:
BTW I made a mild improvement to the test. We've been bitten in the past with checking for exact strings in the html. By using the WebTest form library for this might help someone one day. https://github.com/ckan/ckan/commit/088638a6ff2f3549da91c1b670272823a0086613

We alluded to this policy here:
https://github.com/ckan/ckan/blob/master/ckan/new_tests/controllers/__init__.py#L19-L20

brew:
:+1:

amercader:
Backported to 2.3

Issue #1647, opened 2014-04-15
Stats extension links to groups


rossjones:
The stats extension has a section called Largest Groups.  It shows organisations, and then links to the /group/NAME url (causing an error).

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #2250, opened 2015-01-29
[#1647] Stats extension. 'Largest Groups' links lead either to organizations or to groups page

serkostyx:
Currently all links under the 'Largest Groups' refer to groups page.

After this change links refer to group page for groups and to organization page for organizations


amercader:
Thanks for the patch @serkostyx

Issue #1649, opened 2014-04-15
Stats extension tag counts are incorrect


rossjones:
http://beta.ckan.org/stats#top-tags shows 'test' having 6 datasets, clicking on the name takes you to search where it appears there are only 2 datasets using that tag.

rossjones:
This seems to have been fixed at some point.

PR #2253, opened 2015-01-29
[#1649] Stats extension. Tag counter works properly

serkostyx:
Currently on 'Top Tags' tab displayed total amount of tags which includes deleted ones, private and deleted packages

After this change, tags with state=deleted, deleted packages, private packages are filtered out


amercader:
Thanks for the patch @serkostyx

Issue #1372, opened 2013-12-09
Make resource name default to file name


mwainwright:
Users are usually not sure what to enter as the 'resource name' and it ends up being blank. It's mildly helpful for a resource to have a name, however.

I suggest the field should be pre-populated with the filename, or (for a linked file) component of the name after the last "/". E.g. for <http://example.com/this/that/theother.csv> the "name" would be "theother.csv", unless the user changed it.

Obviously, if the URL ends with a / like <http://example.com/this/that/theother/>  the default name should be "theother", rather than being blank.

rossjones:
@smotornyuk is this fixed now?

wardi:
No, there's a PR to address this with client side JS, but it overwrites the name even if it has been given so I haven't merged it

rossjones:
Will assign to Sergey then in case he wants to finish it off.

smotornyuk:
@rossjones , i've fixed problem mentioned by Ian and updated pull request.

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #2254, opened 2015-01-29
[#1372] Resource name default to file name

ghost:
Currently user have to manually add resource name after or before uploading/linking resource

After this change name will be automatically filled when resource uploaded or added link to resource


wardi:
@ghost this is good, except that it replaces any name that the user might have entered into the form whenever the url or file is changed. I can imagine some people getting upset when data they manually entered gets replaced with data duplicated from another field.

Can you change this so that in only fills in the name when the name value hasn't been manually set?

wardi:
Looks like #2880 has the change I was looking for.

Issue #2216, opened 2015-01-14
User_show doesn't display  users draft datasets  anymore


benjaminlaot:
After this commit : https://github.com/ckan/ckan/commit/b2dfccc8932947711d4f9b0f2500be4c40208b21  , user_show displays only users active datasets and not users draft datasets anymore. 
So if we delete these filters : filter_by(state='active'), filter_by(private=False), users draft datasets could be displayed again, but deleted ones too...

amercader:
@davidread you did the original work on #2130 so you might know more about this. What do you think about these filters? Considering the authorization should allow only the same user and sysadmins to see these datasets it sounds like at least private datasets and drafts could be visible.

davidread:
@benjaminlaot thanks for raising this. I've done a PR: https://github.com/ckan/ckan/pull/2256 - what do you think?

benjaminlaot:
I think you did a good job doing this and I think we have now the right behaviour.

PR #2256, opened 2015-01-29
[#2216] user can now see his private & draft datasets again

davidread:
My recent commit https://github.com/ckan/ckan/commit/b2dfccc8932947711d4f9b0f2500be4c40208b21 stopped draft and private datasets from being returned by user_show, because other general users should not know about these datasets that they cannot view. However it was rightly pointed out by @benjaminlaot that when a user views his own user page (or a sysadmin views it) it makes sense to show these draft and private datasets to him. So I've added logic for this in this PR.

Similar logic is added to the number_created_packages property (user_show). However user_list's number_created_packages excludes draft&private datasets for everyone (for simplicity) - before it just excluded draft ones. This is now also documented and tested.

### DIFF
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -574,7 +574,9 @@ def user_dictize(user, context):
     result_dict['display_name'] = user.display_name
     result_dict['email_hash'] = user.email_hash
     result_dict['number_of_edits'] = user.number_of_edits()
-    result_dict['number_created_packages'] = user.number_created_packages()
+    result_dict['number_created_packages'] = user.number_created_packages(
+        include_private_and_draft=context.get(
+            'count_private_and_draft_datasets', False))
 
     requester = context.get('user')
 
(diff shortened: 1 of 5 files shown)


davidread:
@amercader this will need to go into CKAN 2.3 - can you label it please?

benjaminlaot:
Great job !

benjaminlaot:
If I may, I think that to fit with the  spirit of #2261 reported by @clementmouchet, it would be great to allow the display of packages with custom states by changing this filter in def user_show(context, data_dict) line 1359: 

dataset_q = dataset_q \
                .filter(_or_(model.Package.state=='active',
                            model.Package.state=='draft'))

by this one : dataset_q = dataset_q \
                       .filter(model.Package.state!='deleted')

It won't break existing behaviour as draft and active datasets will be displayed for a user looking at his own account or a  sysadmin, but it also allows the existence of custom states for more flexibility

If you are agreed with this, these lines might change too (line 205 in ckan/model/user.py)

if include_private_and_draft: 
       q = q.filter(model.Package.state.in_(('active', 'draft')))

replaced by : 
if include_private_and_draft: 
     q = q.filter(model.Package.state!='deleted')

Whad do you think @davidread ?

amercader:
@davidread merged to master and 2.3

PR #2257, opened 2015-01-29
Tobes 2057 fix filename munging

davidread:
Issue #2057

This PR is an improved version of Tobe's PR:
https://github.com/ckan/ckan/pull/2149/files

If we upload a file with a huge name then the extension gets eaten and we end up with the datastore not being able to import the file correctly if the format has not been set for the resource

Also it seems browsers on windows 7 seem to pass the whole path along with the file name so this creates problems for these users.

So we remove the path and try to keep the extension - really long extensions (over 20 chars) get truncated but these are probably erroneous extensions anyway.

### DIFF
--- a/ckan/lib/munge.py
+++ b/ckan/lib/munge.py
@@ -149,7 +149,7 @@ def munge_filename(filename):
         ext = ext[:21]
     # max/min size
     ext_length = len(ext)
-    name = _munge_to_length(name, 3 - ext_length, 100 - ext_length)
+    name = _munge_to_length(name, max(3 - ext_length, 1), 100 - ext_length)
     filename = name + ext
 
     return filename
--- a/ckan/lib/munge.py
+++ b/ckan/lib/munge.py
@@ -116,9 +116,9 @@ def munge_filename_legacy(filename):
     ''' Tidies a filename. NB: deprecated
 
     Unfortunately it mangles any path or filename extension, so is deprecated.
-    It needs to remain for use by model_dictize() because if this routine
-    changes then group images uploaded previous to the change may not be
-    viewable.
+    It needs to remain unchanged for use by group_dictize() and
+    Upload.update_data_dict() because if this routine changes then group images
+    uploaded previous to the change may not be viewable.
     '''
     filename = substitute_ascii_equivalents(filename)
     filename = filename.strip()
@@ -130,10 +130,8 @@ def munge_filename_legacy(filename):
 def munge_filename(filename):
     ''' Tidies a filename
 
-    Keeps the filename extension.
+    Keeps the filename extension (e.g. .csv).
     Strips off any path on the front.
-    It can be useful to keep the extension (e.g. .csv) in a resource when the
-    format field set.
     '''
 
     # just get the filename ignore the path
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -116,8 +116,7 @@ def resource_dictize(res, context):
     ## for_edit is only called at the times when the dataset is to be edited
     ## in the frontend. Without for_edit the whole qualified url is returned.
     if resource.get('url_type') == 'upload' and not context.get('for_edit'):
-        last_part = url.split('/')[-1]
-        cleaned_name = munge.munge_filename(last_part)
+        cleaned_name = munge.munge_filename(url)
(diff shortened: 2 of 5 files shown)


ckan/lib/munge.py
```diff
@@ -119,6 +127,34 @@ def munge_filename(filename):
...
+    filename = filename.lower().strip()
+    filename = re.sub(r'[^a-zA-Z0-9. -]', '', filename).replace(' ', '-')
+    # resize if needed but keep extension
+    name, ext = os.path.splitext(filename)
+    # limit overly long extensions
+    if len(ext) > 21:
+        ext = ext[:21]
+    # max/min size
+    ext_length = len(ext)
+    name = _munge_to_length(name, 3 - ext_length, 100 - ext_length)
```
tobes: This seems wrong `3 - ext_length` can be `-19` a min length of `-19` seems illogical.
`'_' * -19` in `_munge_to_length` will create odd names like `.ext` where as `_.ext` might be better

tobes:
I think this also breaks the existing logic see previous pull request regarding legacy issues and corrupting existing data.

ckan/lib/munge.py
davidread: Good spot on the negatives. It happens to work ok, but just to be sure I've added a min value of 1.

Not sure I can do much about '.ext' without getting rid of splitext:
```
>>> os.path.splitext('.ext')
('.ext', '')
```
And I'm not convinced it is an edge case to necessarily worry about.

davidread:
Good point - I'd used munge_filename_legacy in the wrong place - I'm changed it to the image upload stuff.

brew:
Is there still work to be done on this before review?

ckan/lib/munge.py
```diff
@@ -119,6 +127,34 @@ def munge_filename(filename):
     return filename
 
 
+def munge_filename(filename):
+    ''' Tidies a filename
+
+    Keeps the filename extension.
+    Strips off any path on the front.
+    It can be useful to keep the extension (e.g. .csv) in a resource when the
+    format field set.
```
tobes: The reason we need the extension is for the datapusher to function etc.

It is needed when the format is __not__ set

tobes:
There are the following issues, the second is more cosmetic/code read-ability

https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L475
```
image_url = munge.munge_filename(image_url)
```
should be using legacy munging to keep existing logic

https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L119
```
last_part = url.split('/')[-1]
cleaned_name = munge.munge_filename(last_part)
```

should be

```
cleaned_name = munge.munge_filename(url)
```

If any extensions use `munge_filename` (they shouldn't) they will also break

brew:
Is there a real need to munge in `group_dictize`? 

https://github.com/ckan/ckan/blob/cf63aa0654f4ae583d9cf1437bf08b5507936ad5/ckan/lib/dictization/model_dictize.py#L475

tobes:
>  Is there a real need to munge in group_dictize? 

Did you not read the comment above the line you linked to?  Assume that all user supplied data is evil is a good thing to do so yes we should.  If you remove this then you would also need to 'clean' existing data as well as ensure that 'bad' data cannot be introduced via any routes into ckan so it is probably best to keep it.

Also I'd not be happy changing the logic unless there is a good reason to do so.  The issues here about the need for the legacy function relate to uploads done via windows systems which I don't think any developers actually use so will not notice any breakage until it affects clients.

The image url may need looking at at some poin,t but the issue in this request (and my previous one it is based on) does screw existing clients and prevents the datastore working on windows systems with long paths.

brew:
Are there unsanitised input routes that make this field especially vulnerable? If we're happy the input is sufficiently sanitised, then removing the need to munge the output might mean we don't need to keep the legacy stuff around.

davidread:
@tobes I've done another commit that addresses the two issues you mention https://github.com/ckan/ckan/pull/2257#issuecomment-72650077 - cheers for that.

> Did you not read the comment above the line you linked to?

This is not constructive language.

davidread:
@tobes just checking if you are happy with this PR, as it stands now?

tobes:
Yes it's all good with me

davidread:
Great, over to you @brew.

Issue #1541, opened 2014-02-25
Alt text for widgets


mwainwright:
The widgets for dashboard, settings, etc have no alt text.

They seemingly are inserted with some clever hack using CSS (so in fact the images are visible in Safari or Firefox even if image loading is turned off) but loading the page in a text browser such as Lynx, they don't appear at all, so the user can't see those links.

PR #2258, opened 2015-01-30
add text to account links

mintcanary:
Adds text to the account-masthead links. This addresses https://github.com/ckan/ckan/issues/1541 (text browser support), as well as giving more options when theming (without the need to edit the template).


joetsoi:
I am so overjoyed that we can now say we support elinks, that I might start using elinks for development.

PR #2262, opened 2015-02-04
[#1847] Add middleware that cleans up the pylons response string.

### LINKED ISSUE #1847: Large memory leak in datastore API
When querying the datastore API on a large dataset (eg. querying 10,000 rows at a time) the memory used by the process keeps growing. On our dataset this is about 20MB leaking per request - more than can be reasonably mitigated by restarting the worker after a number of requests.

This happens both when running the application using Paster and when running as a WSGI script under Apache2.

It seems the string containing the response is kept in memory. This is referenced by the response object, which is referenced by the pylon context and (a few steps up) by the WSGI environment. The environment itself is referenced numerous times, so it's hard to track this down.

As a proof of concept, the symptom can be removed by this "amusing" code:

```python
diff --git a/ckan/controllers/api.py b/ckan/controllers/api.py
index 08ba3a8..4b09102 100644
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -19,6 +19,12 @@ import ckan.lib.munge as munge
 
 from ckan.common import _, c, request, response
 
+from threading import Timer
+
+
+def _clean_links(env):
+    env['pylons.controller']._py_object.response._body = ''
+
 
@@ -67,7 +73,9 @@ class ApiController(base.BaseController):
      
         # avoid status_code_redirect intercepting error responses
         environ['pylons.status_code_redirect'] = True
-        return base.BaseController.__call__(self, environ, start_response)
+        result = base.BaseController.__call__(self, environ, start_response)
+        Timer(10, _clean_links, [environ]).start()
+        return result
      
     def _finish(self, status_int, response_data=None,
                 content_type='text'):
```

Obviously this is not a workable solution!

joetsoi:
(fixes #1847)
Once the response string has been served, this middleware replaces the response string with a dummy one, so the original response can be garbage collected. This is lifted from https://code.google.com/p/modwsgi/wiki/RegisteringCleanupCode This middleware is can be switched on by setting 'ckan.use_pylons_response_cleanup_middleware = true' in your development.ini

So I did a comparison of with and without this enabled. Everyone likes graphs right?
memory_usage
Notably, the memory usage still grows with this enabled, but is a bit more stable. I also did some examining of the memory and this memory has been returned to python's free lists but has not been released by the python process. see (http://effbot.org/pyfaq/why-doesnt-python-release-the-memory-when-i-delete-a-large-object.htm). So we'll still use up as much memory as there are threads

So the other way around it is to do something like use gunicorn +  gevent, although this isn't a fair direct comparison as this is only using one worker process, but I mainly suggest this as you can set worker processes to restart after serving a maximum number of requests (see http://docs.gunicorn.org/en/develop/configure.html#max-requests), which would release any memory that the process has been allocated but is not using.

Anyway, this isn't critical or anything as it does just fix an edge case for a very particular workload that most people will not come across. (still needs tests)


ckan/config/middleware.py
```diff
@@ -360,3 +362,43 @@ def __call__(self, environ, start_response):
             self.engine.execute(sql, key, data.get('url'), data.get('type'))
             return []
         return self.app(environ, start_response)
+
+
+class Generator(object):
+    def __init__(self, iterable, callback, environ):
+        self.__iterable = iterable
+        self.__callback = callback
+        self.__environ = environ
```
wardi: ew. private members with name munging

joetsoi:
@aliceh75, might be nice if you could test this, as it's basically your fix without the timer, you might also want to look into using gunicorn + gevent that I suggested to utilize the process restart after serving a number of requests.

aliceh75:
Hi,

I can confirm this stabilizes the memory. The test I was just running (continuous loop fetching 10,000 rows with 4 workers) saw the memory go from 800M to 1.8G and back quite rapidly. With the middleware enabled it is far more stable (It's been oscillating between 793M and 823M for a while now, most of the time staying around 810M).

This is a great improvement :-) Thank you for doing this.

aliceh75:
mod_wsgi can also restart after a number of requests btw.

amercader:
:tada: woo, nice great work @joetsoi

wardi:
@joetsoi I've made a couple small changes: https://github.com/wardi/ckan/tree/1847-pylons-clean-middleware would you take a look and cherry-pick or merge into your branch?

wardi:
@joetsoi this link makes it easier to see the two commits https://github.com/wardi/ckan/commits/1847-pylons-clean-middleware Would you like a PR against your PR? :-)

joetsoi:
@wardi, we should just all give each other edit permissions to each other's ckan forks, that'd be distributed version control in action :p

I've fixed up the tests/changelog and made it switch on by default, no point having it if it's not going to be used.

wardi:
@joetsoi sounds good. is this good to go?

joetsoi:
@wardi yeah it's all ready, it's just coveralls moaning.

PR #2265, opened 2015-02-04
Include type in group controllers index method.

njall:
The index action in the group controller did not include 'type' in data_dict meaning any new custom group types will only have groups of type 'group' returned.


njall:
Submitting again with fix as this had an error. Details below:

_guess_group_type returns None if the group type is 'group' making it try to filter all groups by type None. 
Only include type in data_dict if group_type is a thing 
        if group_type:
            data_dict['type'] = group_type

PR #2266, opened 2015-02-05
Update deployment.ini_tmpl to reflect changes in resource views

amercader:
Basically load relevant plugins in `ckan.plugins` and `ckan.views.default_views` to match the old behaviour as much as possible.


seanh:
I guess for people who already have config files and are upgrading, we need a note in the changelog for what they need to add to their config?

amercader:
@seanh yes, will add it

PR #2268, opened 2015-02-05
Filter by group type

njall:
We can create our own custom groups easily in CKAN as an extension but when loading the index page of our new group using the Group Controller, we do not include type in our request. 
This fix adds group_type to data_dict['type'] which will be the name of our custom group if using it. If it is the default group it will be None so ignores it.


wardi:
This change doesn't have any test coverage. Would you add an example extension that implements a custom group type something like the other ckanext/example_* directories so we can be confident this works and won't break in the future?

njall:
I've added an example_igroupform plugin and controller. I can't spare the time to write a full UI or a test suite for it though. Is this enough? Could it be made a bug for someone else to enhance in the future?

wardi:
Is this complementary to or superceeded by #2354 ?

Also is removing the 404 in the controller correct? Shouldn't we always be using the right controller for the group_type?

wardi:
Closing as duplicate of #2354

PR #2273, opened 2015-02-08
Do not provide email configuration in default config file

nigelbabu:
The default config file provides an email address. This can lead to interesting side effects thanks to how Pylons works. An exception is printed to stdout if and only if there's no email address provided. If the provided one doesn't work, it just leads to an error about failing to send email. See [my post](http://nigelb.me/2014-08-18-arrrrrgh-tracebacks-and-exceptions.html) for more details.


wardi:
@amercader seems like an obvious thing for 2.3 and backporting

Issue #2285, opened 2015-02-13
user.state is not translated in user/read_base.html as well as some placeholders in organization/member_new.html


benjaminlaot:
I will submit a PR for that.

And in the member_new template line 50:
<input id="email" type="text" name="email" placeholder="Email address">

I would like to have a type="email" in order to validate from the client side the email address. Would you accept the PR?

benjaminlaot:
Fixed in https://github.com/ckan/ckan/pull/2287

PR #2287, opened 2015-02-13
#2285 made some fields translatable

benjaminlaot:
user.state in user/read_base.html is now translatable as well as some placeholders in organization/member_new.html


benjaminlaot:
Is this going into ckan 2.3 @amercader ?

amercader:
@benjaminlaot Sadly no, as it involved adding new strings for translation

Issue #2283, opened 2015-02-13
template_legacy resources is missing after an installation - /ckan-admin/trash page is unable


Jan-Marcek:
I'm unable to purge deleted dataset via trash webpage.

I'm using https://github.com/ckan/ckan/tree/ckan-2.2.1

```
[Fri Feb 13 16:14:48 2015] [error] 2015-02-13 16:14:48,485 INFO  [ckan.lib.base]  /api/i18n/en render time 0.001 seconds
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] Error - <class 'genshi.template.loader.TemplateNotFound'>: Template "layout.html" not found
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] URL: http://XXXX/ckan-admin/trash
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] File '/usr/share/python/odn-ckan-shared/lib/python2.7/site-packages/weberror/errormiddleware.py', line 162 in __call__
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1]   app_iter = self.application(environ, sr_checker)
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] File '/usr/share/python/odn-ckan-shared/lib/python2.7/site-packages/webob/dec.py', line 147 in __call__
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1]   resp = self.call_func(req, *args, **self.kwargs)
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] File '/usr/share/python/odn-ckan-shared/lib/python2.7/site-packages/webob/dec.py', line 208 in call_func
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1]   return self.func(req, *args, **kwargs)

[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] File '/usr/share/python/odn-ckan-shared/lib/python2.7/site-packages/webob/request.py', line 1053 in get_response
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1]   application, catch_exc_info=False)
[Fri Feb 13 16:15:03 2015] [error] [client 127.0.0.1] File '/usr/share/python/odn-ckan-shared/lib/python2.7/site-packages/webob/request.py', line 1022 in call_application
[...]

PR #2289, opened 2015-02-16
#2283 pack template legacy resources - fix in v2.2.1

Jan-Marcek:
Currently, there is a bug (#2283- template_legacy resources was missing after an installation) so that when a user runs an installation via command python setup.py install then the user is unable to get to trash page - ckan-admin/trash - http://docs.ckan.org/en/latest/sysadmin-guide.html#permanently-deleting-datasets

After this change, the user can delete the datasets permanently and the resources of template_legacy is exported after an installation via python setup.py install


amercader:
Merged to 2.2.2 in fa7c79

PR #2296, opened 2015-02-17
Featured views

deniszgonjanin:
The CKAN home page can be very bland. This PR uses the new Resource Views to create a new homepage layout, which showcases a number of 'featured' Resource Views on the home page.

A View is Featured via a checkbox in the default Resource View form. This PR:

- Creates a new Homepage layout template
- Adds a `featured` column to the Resource View model
- Fixes an issue with CKAN checkbox validation, whereas an unchecked checkbox was being ignored instead recognized as meaning `False`

A demo is available here: http://featuredviews.dcats.ca/

Primarily I would like to start a discussion as to how to make the Homepage show off CKAN better than it does now, whether people like this approach, and what are some other ways to achieve a similar effect.


wardi:
Closing as I believe that https://github.com/deniszgonjanin/ckanext-featuredviews is now a complete version of this feature implemented as an extension.

Consider registering that extension on http://extensions.ckan.org/ !

Issue #2301, opened 2015-02-19
The title of an email is not translatable. A parenthesis is misplaced in mailer.py


benjaminlaot:
A parenthesis is misplaced in mailer.py line 151
subject = _('Invite for {site_title}'.format(site_title=g.site_title))

It should be : subject = _('Invite for {site_title}').format(site_title=g.site_title)

amercader:
Fixed in #2302

PR #2302, opened 2015-02-19
#2301 Modify the place of the misplaced parenthesis

benjaminlaot:
The title of the email is now translatable as it should be.


amercader:
This looks great @benjaminlaot
Can you send a new PR with just the ef03c2cb8a7c9273bdd5375937819dd042c3c351 change? Just to keep history a bit more tidy.

Thanks

PR #2314, opened 2015-02-23
Upgrade all occurences of Flot to Flot 0.8.3.

mattfullerton:
This seems to (at least) resolve a problem with x/y axes ordering in the preview (which was on Flot 0.7) and harmonizes behaviour between the preview and the Recline Graph View (which was on Flot 0.8.1).

I have not extensively tested this, but am using it intensively myself (although not the stats plugin). So this pull request is as much an encouragement to consider upgrading Flot as it is a concrete request to merge.

### DIFF
--- a/ckan/public/scripts/vendor/flot/0.7/jquery.flot.js
+++ b/ckan/public/scripts/vendor/flot/0.7/jquery.flot.js
@@ -1,16 +1,17 @@
-/*! Javascript plotting library for jQuery, v. 0.7.
- *
- * Released under the MIT license by IOLA, December 2007.
- *
- */
+/* Javascript plotting library for jQuery, version 0.8.3.
+
+Copyright (c) 2007-2014 IOLA and Ole Laursen.
+Licensed under the MIT license.
+
+*/
 
 // first an inline dependency, jquery.colorhelpers.js, we inline it here
 // for convenience
 
 /* Plugin for jQuery for working with colors.
- * 
+ *
  * Version 1.1.
- * 
+ *
  * Inspiration from jQuery color animation plugin by John Resig.
  *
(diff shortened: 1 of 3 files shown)


amercader:
> This seems to (at least) resolve a problem with x/y axes ordering in the preview (which was on Flot 0.7) and harmonizes behaviour between the preview and the Recline Graph View (which was on Flot 0.8.1).

@mattfullerton can you expand on this? What it the issue exactly?

mattfullerton:
Is there a CKAN instance where I can create Recline Graph views in addition to the Dataexplorer preview? demo.ckan.org doesn't seem to support it.

amercader:
http://master.ckan.org (running master) and http://beta.ckan.org (running 2.3)

mattfullerton:
I can't reproduce it :( But I think on principle we should harmonise the versions in use. Isn't there a way to have the plugins depend on the flot (and perhaps other things) already included in CKAN core?

https://github.com/ckan/ckan/blob/master/ckanext/stats/public/ckanext/stats/vendor/jquery.flot.js#L1
https://github.com/ckan/ckan/blob/master/ckanext/reclineview/theme/public/vendor/flot/jquery.flot.js#L1
https://github.com/ckan/ckan/blob/master/ckan/public/scripts/vendor/flot/0.7/jquery.flot.js#L1

If accepted, the directory name should also be changed from 0.7 to 0.8

amercader:
@mattfullerton I've removed the backport labels because this is not actually fixing any issue (unless you can reproduce it). The recline view on Master is running flot 0.8, and I'm not sure why the one on /public is used for, so I'm closing this for now until there is a specific issue that needs to be fixed.

PR #2316, opened 2015-02-23
Make it possible to add a dataset without adding data.

mattfullerton:
This has probably been discussed before somewhere so I expect this will be rejected. But one use case is when a user is unsure whether the data upload will be successful and doesn't want to risk creating an orphaned dataset and hence lose the name (slug). I'm sure there are a few edge CKAN use cases where a user does not want to add any resources.

This code is probably not the best, I did this in a hurry. If the feature is actually desired I can work on it some more.

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -935,6 +935,14 @@ class PackageController(base.BaseController):
                         url = h.url_for(controller='package',
                                         action='new_metadata',
                                         id=pkg_dict['name'])
+                    elif request.params['save'] == 'finish':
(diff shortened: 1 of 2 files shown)


amercader:
Migration from previous versions: custom python code that gets all resource ids from the datastore database and adds an extra manually to the resource table.

mattfullerton:
@amercader Should I understand that comment?

amercader:
@mattfullerton sorry, wrong PR

amercader:
Closing in favour of #2327

PR #2234, opened 2015-02-24
datastore sets datastore_active on resource at runtime.

amercader:
The datastore seems to intercept every time a resource is going to be shown, to insert a datastore_active field - by querying the database.  With datasets containing a large number of resources, it isn't clear why this is preferable to setting it in datastore_create and removing it in datastore_delete.


amercader:
Migration from previous versions: custom python code that gets all resource ids from the datastore database and adds an extra manually to the resource table.

rgrp:
@amercader can this get merged?

amercader:
@rgrp no, there is the migration bit missing

ckan/migration/versions/081_set_datastore_active.py
```diff
@@ -0,0 +1,48 @@
...
+        resources = migrate_engine.execute('''
+            SELECT id, extras
+            FROM resource
+            WHERE id IN ({0}) AND extras IS NOT NULL
+        '''.format(
+            ','.join(['\'{0}\''.format(_id[0])
+                      for _id
+                      in resources_in_datastore])
+            )
+        )
```
amercader: @wardi @joetsoi any ideas on how to use parameters binding for this (in an IN clause)?

wardi: can't you just pass the list?

ckanext/datastore/logic/action.py
```diff
@@ -141,6 +142,14 @@ def datastore_create(context, data_dict):
     except db.InvalidDataError as err:
         raise p.toolkit.ValidationError(str(err))
 
+    # Set the datastore_active flag on the resource if necessary
+    if not resource_dict:
+        resource_dict = p.toolkit.get_action('resource_show')(
+            context, {'id': data_dict['resource_id']})
+    if not resource_dict.get('datastore_active'):
+        resource_dict['datastore_active'] = True
+        p.toolkit.get_action('resource_update')(context, resource_dict)
```
wardi: Could use `resource_patch(id=data_dict['resource_id'], datastore_active=True)` here and save the extra action call. Same for datastore_delete below

ckan/migration/versions/081_set_datastore_active.py
```diff
@@ -0,0 +1,54 @@
...
+
+    datastore_engine = create_engine(datastore_connection_url)
+
+    try:
+
+        resources_in_datastore = datastore_engine.execute('''
+            SELECT table_name
+            FROM information_schema.tables
+            WHERE table_schema = 'public'
+            AND table_name != '_table_metadata'
```
wardi: Should we get the list of tables from _table_metadata in case someone was clever and added their own tables to this db?

wardi:
@rossjones neither of my comments are a big deal, if you don't feel like changing this PR I'm happy to merge as-is

ckan/migration/versions/081_set_datastore_active.py
amercader: I'm not sure is worth it, the statement on line 31 will check that the table names actually match resource ids, so if we get their own tables here they will be ignored. `_table_metadata` contains aliases for actual tables (ie views), so those would also need to be ignored

amercader:
@wardi I always forget about patch!

ckan/migration/versions/081_set_datastore_active.py
wardi: :+1: Makes sense to me, leave it as-is.

metaodi:
@wardi @amercader I think my Travis CI job for ckanext-harvest failed because this PR has been merged: https://travis-ci.org/ckan/ckanext-harvest/jobs/90579736

I guess in the DB setup the additinal `datastore_default` user is missing (or even more?). This issue might occur on other extensions as well. This is just quick note about this, just a reminder, that there might be additional cleanup work needed.

wardi:
We should check that the datastore has actually been set up instead of relying on the configuration value. Maybe a simple try: except: in this migration will suffice.

ckanext/datastore/plugin.py
```diff
@@ -282,21 +282,9 @@ def before_show(self, resource_dict):
                 controller='ckanext.datastore.controller:DatastoreController',
                 action='dump', resource_id=resource_dict['id'])
 
-        connection = None
+        if 'datastore_active' not in resource_dict:
+            resource_dict[u'datastore_active'] = False
```
wardi: this part makes me sad because it kills the speedup we get from serving the validated data dict straight from solr

rossjones: I seem to recall this is just a catch all for when datastore_active isn't set at all on the resource for some reason. It can probably go as long as we definitely set datastore_active: False on all existing resources.

amercader: sorry, I don't follow. Why does it kill it?

wardi: Correction: It's not this part, it's simply the fact that we have an IResourceController plugin enabled. https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L998 just looping over the resources to pass them to IResourceController.before_show() causes our LazyJSONObject to not be lazy any more.

Since so many sites enable the datastore maybe it's not worth having the LazyJSONObject at all. Or we could only apply the before_show calls on unvalidated data_dicts (and cache the value along with the rest of the metadata).

PR #2307, opened 2015-02-25
My Datasets view on dashboard is empty

amercader:
Beta 2.3 site pre existing data (loaded up during previous version beta testing) is not appearing in users dashboard | my dataset view, but IS showing up if I click on my user name at the top.
missing dataset list

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -640,7 +640,7 @@ class UserController(base.BaseController):
     def dashboard_datasets(self):
         context = {'for_view': True, 'user': c.user or c.author,
                    'auth_user_obj': c.userobj}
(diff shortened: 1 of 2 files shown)


amercader:
@wardi @davidread An easy fix (probably left out on #2133). Can you review and merge if you are happy? This should go into 2.3

davidread:
Thanks for spotting his and fixing.

Just wondering if someone else should in fact be allowed to see your
datasets. Isn't that the previous behaviour? Sorry, am away and can't check.

amercader:
@davidread the dashboard in general is only visible to the own logged in user. Another user's datasets can be seen on the user page (eg http://beta.ckan.org/user/okfn)

davidread:
Got it, cheers

PR #2326, opened 2015-02-26
give default ini value to celery when config not specified

deniszgonjanin:
Celery daemon fails unless config file is explicitly identified. This change just makes the celery paster command in line with how other ckan paster commands behave.

Hoping this might make it into 2.3 somehow cc/ @amercader


amercader:
@deniszgonjanin this looks good. Should the command fail if no config was provided and development.ini is not present on the current directory?

deniszgonjanin:
@amercader yes, there. an appropriate message is better than a random fail message

PR #2329, opened 2015-03-02
State that before_show is AFTER validation in the comment

mattfullerton:


mattfullerton:
Sorry, I see this was already done

PR #2330, opened 2015-03-03
add id to search form in search_form.html snippet

joetsoi:
http://webtest.readthedocs.org/en/latest/forms.html

minor change to search_form.html. I want to be able to select forms in webtests by id
`my_form = forms['my-search-box']`
instead of either hoping `forms[0]` is my form or iterating and checking it contains correct text

### DIFF
--- a/ckan/templates/group/index.html
+++ b/ckan/templates/group/index.html
@@ -17,7 +17,7 @@
 {% block primary_content_inner %}
   <h1 class="hide-heading">{{ _('Groups') }}</h1>
   {% block groups_search_form %}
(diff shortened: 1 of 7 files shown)


joetsoi:
I should fix the dataset/org/group templates to use this as well

brew:
The only references I could find in the docs to id attribute use are in the [CSS coding standards](http://docs.ckan.org/en/latest/contributing/css.html):

> Also ids should not be used in selectors as it makes it far too difficult to override later in the cascade.

> ```/* Cannot override this button style without including an id */```
> ```.btn#download {}```

and in the [HTML coding standards](http://docs.ckan.org/en/latest/contributing/html.html#forms), where id use is encouraged within form fields to benefit accessibility. 

I don't either of these conflict with your intension here. A few thoughtfully named id attributes to help target within the DOM should be fine.

Are you intending to add similar to all relevant forms?

Issue #2238, opened 2015-01-22
recline_view.js assumes that window.parent.ckan is defined


clementmouchet:
There is an assumption in `recline_view.js` that means views won't load when embedded in a different website.

l.65
``` javascript
      if (window.parent.ckan.views && window.parent.ckan.views.filters) {
```
This is assumes `window.parent.ckan` is defined, but it's not always defined.
Changing this to:

``` javascript
      if (window.parent.ckan && window.parent.ckan.views && window.parent.ckan.views.filters) {
```
fixes the problem.

I'll submit a pull request asap.

amercader:
Good catch, can you send a PR so we can backport it to a patch release?

clementmouchet:
PR: https://github.com/ckan/ckan/pull/2335

Thanks @amercader

amercader:
Fixed in #2423

PR #2335, opened 2015-03-05
2238 Fix an assumption in recline_view.js 

clementmouchet:
Fix for issue https://github.com/ckan/ckan/issues/2238


amercader:
@clementmouchet Actually the check won't work at all if you are embeding the view on an external site. I don't think the browser will allow you to access stuff like `window.parent.ckan` unless protocols, domains and ports match. Can we detect somehow that we are on an embeded iframe (checking parent and self domains) and bypass the check all together if that's the case?

amercader:
Or wrap it in a try / catch

PR #2337, opened 2015-03-06
Fix #2328 by enabling URL quoting

### LINKED ISSUE #2328: Requesting page with non-ASCII URL fails in Jinja template rendering
[Jinja only accepts unicode or ASCII](http://jinja.pocoo.org/docs/dev/api/#unicode)

`ckan/lib/base.py` relays requested URLs in raw format to Jinja, causing `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 20: ordinal not in range(128)` if URL or its parts are rendered to page.

This issue is visible, for example, when searching for datasets using a browser that does not auto-escape non-ASCII characters. For example, IE dies not escape while Firefox does. This bug is easily reproduced by searching datasets with "ä" as the search term using IE, or by using curl: `curl -v "http://demo.ckan.org/dataset?q=ä"`

URLs and any other data used in Jinja templates should be decoded and encoded into unicode before rendering. This is not a trivial fix, since correctly identifying an unknown encoding in all cases is borderline, if not, impossible.

One possibility would be to assume requests come in UTF-8 or ASCII and just decode everything as if it was UTF-8 (since only-ASCII UTF-8 is the same as the same text in ASCII). This would not work for other encodings like latin-1, but would be a start.

I can create a patch and pull request for the UTF-8/ASCII-only solution if it's considered acceptable. Otherwise I'd like to request comments on other options. [BeautifulSoup](http://www.crummy.com/software/BeautifulSoup/bs3/documentation.html#Beautiful%20Soup%20Gives%20You%20Unicode,%20Dammit) has one solution for a few encodings.

bzar:
https://github.com/ckan/ckan/issues/2328

### DIFF
--- a/ckan/config/middleware.py
+++ b/ckan/config/middleware.py
@@ -218,7 +218,7 @@ class I18nMiddleware(object):
 
             if qs:
                 # sort out weird encodings
-                #qs = urllib.quote(qs, '')
+                qs = urllib.quote(qs, '')
                 environ['CKAN_CURRENT_URL'] = '%s?%s' % (path_info, qs)
             else:
                 environ['CKAN_CURRENT_URL'] = path_info


wardi:
Thank you @bzar !

Issue #2255, opened 2015-01-29
Docker installation fails to build due to apparent Solr schema version conflict


beechnut:
I followed the [Docker installation instructions](http://docs.ckan.org/en/latest/maintaining/installing/install-using-docker.html#installing-ckan), and while the Solr and Postgresql containers are running, CKAN fails to build.

```
$ docker ps -a
CONTAINER ID        IMAGE                    COMMAND                CREATED             STATUS                              PORTS               NAMES
596bbfb621f2        ckan/solr:latest         "java -jar start.jar   5 days ago          Up 38 minutes                       8983/tcp            solr
b3af69f91938        ckan/postgresql:latest   "/usr/local/bin/run"   5 days ago          Up 42 minutes                       5432/tcp            db
```
Supporting services are running, but then (running it interactively to see logs):

```
$ docker run -it --rm -p 8000:8000 --link db:db --link solr:solr ckan/ckan
*** Running /etc/my_init.d/50_configure...
Distribution already installed:
  ckan 2.4a from /usr/lib/ckan/default/src/ckan
Creating /etc/ckan/default/ckan.ini
Now you should edit the config files
  /etc/ckan/default/ckan.ini
Edited option sqlalchemy.url = "postgresql://ckan_default:pass@localhost/ckan_default"->"postgresql://ckan:ckan@172.17.0.40:5432/ckan" (section "app:main")
Option uncommented and set solr_url = "http://172.17.0.41:8983/solr/ckan" (section "app:main")
Option uncommented and set ckan.storage_path = "/var/lib/ckan" (section "app:main")
Edited option email_to = "you@yourdomain.com"->"disabled@example.com" (section "app:main")
Edited option error_email_from = "paste@localhost"->"ckan@5c6d3e2d141c" (section "app:main")
*** Running /etc/my_init.d/70_initdb...
Traceback (most recent call last):
  File "/usr/lib/ckan/default/bin/paster", line 9, in <module>
    load_entry_point('PasteScript==1.7.5', 'console_scripts', 'paster')()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 104, in run
    invoke(command, command_name, options, args[1:])
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 143, in invoke
    exit_code = runner.run(args)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 238, in run
[...]

PR #2338, opened 2015-03-06
Docker compose

milafrerichs:
### What does this PR do?
Adds docker-compose (replacement for fig) capability. 

It uses the image from the [docker-hub](https://registry.hub.docker.com/u/ckan/ckan/) and depends on the other [images](https://registry.hub.docker.com/repos/ckan/) from the hub.
But each of these can be easily replaced. 

#### How can this help?
Instead of using three commands on the command line
```shell
$ docker run -d --name db ckan/postgresql
$ docker run -d --name solr ckan/solr
$ docker run -d -p 80:80 --link db:db --link solr:solr ckan/ckan
```
you will only run one command
```shell
$ docker-compose up -d
```

## Important
This PR should only be merged after the ckan solr image on the hub is updated. 
See related #2255

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,14 @@
+ckan:
+  image: ckan/ckan
+  links:
+    - db
+    - solr
+  ports:
+    - 80:80
+
+db:
+  image: ckan/postgresql
+
+solr:
+  image: ckan/solr
+


joetsoi:
thanks for the pr, the docker stuff is migrating to https://github.com/ckan/ckan-docker

mattfullerton:
Can the docker stuff from ckan/ckan be removed ASAP then? It's very confusing for newcomers.

filipefigcorreia:
Can someone provide a pointer on the motivation to move to https://github.com/ckan/ckan-docker ? I've skimmed through the README there, and the process seems to be much more complex than before :\

wardi:
The reason the separate ckan-docker repo was created is that our review process was slowing down its development because most core developers are not yet using docker.

Now there are at least two projects that are using ckan with docker. I think we should be generating a common ckan base image from this repo and automate the builds on the docker hub. I don't think we should put the fig/compose configuration here because the core devs aren't going to be able to support it.

westonplatter:
@wardi agree with both of you points. I would prefer the base docker image approach.

Anyone else - thoughts?

wardi:
Closing due to lack of activity. I think we're going to need to remove the Docker config in this repo unless we can start using it as part of our circleci runs.

PR #2349, opened 2015-03-16
rename new_tests -> tests

wardi:
Follow-up to #2221


wardi:
@joetsoi this is good to review now

ckan/new_authz.py
```diff
@@ -431,12 +431,13 @@ def auth_is_anon_user(context):
 
         See ckan/lib/base.py:232 for pylons context object logic
     '''
```
joetsoi: @amercader, this change is a change for the better (avoiding pylons c object), but can you think of any reason not to do this? Do you happen to know of any cases where this would break anything?

amercader: I think it makes sense to rely on the user in the context as this is what the function is passed as param and the controllers are doing the `context = {'user': c.user or c.author}` thing all over the place.

Part of me is slightly anxious about someone having a custom validator on user names that allows dots and this causing problems, but I'm probably just being too much paranoid :)

@wardi I'm curious as to why this was needed when moving tests around (again, I'm happy with it)

wardi: Do we support custom validators on user fields yet? Yes, I'm also terrified about mixing IP addresses and usernames, which is why I want to fix that, in a separate PR.

The tests broke because the new tests relied on getting a TypeError because no context had been set up yet. With the new test order an context now does exist (more reason to avoid using c) but is empty so `c.user == ''` and `c.author == ''` which made `is_anon_user == False`. This breaks the test.

amercader: Makes sense

@joetsoi :shipit:

wardi:
@joetsoi I'm working on resolving the merge conflict

wardi:
good to go, I think

davidread:
Nice one!

Issue #2344, opened 2015-03-11
Error on redirection after creating custom group


KrzysztofMadejski:
CKAN 2.3

Error: `GenerationException: url_for can only return a string, got unicode instead: category_read?id=nowa-grupa2`

Because in `controllers/group` (llok for `# !!`):
```python
def _save_new(self, context, group_type=None):
        try:
            data_dict = clean_dict(dict_fns.unflatten(
                tuplize_dict(parse_params(request.params))))
            data_dict['type'] = group_type or 'group'
            context['message'] = data_dict.get('log_message', '')
            data_dict['users'] = [{'name': c.user, 'capacity': 'admin'}]

            # !! data_dict['type'] is in string
            group = self._action('group_create')(context, data_dict) 
            # !! group['type'] is in unicode

            # Redirect to the appropriate _read route for the type of group
            # !! redirect_to doesn't like unicode
            h.redirect_to(group['type'] + '_read', id=group['name'])
```

KrzysztofMadejski:
I have problem when displaying custom group as well beacuse of strict-check in read:
```python
    def read(self, id, limit=20):
        group_type = self._get_group_type(id.split('@')[0])
        if group_type != self.group_type: # does this check is really needed here?
            abort(404, _('Incorrect group type'))
```

Does one has to apart from implementing IGroupForm also override GroupController and mappings to put to that Controller. Shouldn't default GroupController handle all group types?

How do you support custom groups?

davidread:
The GenerationException is probably due to the name of the route not being registered properly in https://github.com/ckan/ckan/blob/master/ckan/lib/plugins.py#L161 . url_for (called by redirect_to) will first try the parameter (u"category_read") as a route name, and I it should be ok casting that unicode to a str string. But since "category_read" has not been registered it assumes that param is actually a URL, and it doesn't like it as unicode.

So check you have enabled your plugin in your ckan.plugins, and also that your plugin is implements IGroupForm and has a group_types method like this:
```
     def group_types(self):
         return ('category',)
```

davidread:
The second issue, with the `group_type != self.group_type` does seem to be a problem with CKAN. 

There seems to be a contradiction - when you have an IGroupForm it registers the routes for it, connecting it up with either the default GroupController or a custom one:
https://github.com/ckan/ckan/blob/master/ckan/lib/plugins.py#L157-L165
[...]

PR #2354, opened 2015-03-17
IGroupForm improvements

davidread:
Derived from https://github.com/ckan/ckan/issues/2344


ckan/controllers/group.py
```diff
@@ -88,8 +85,6 @@ def _bulk_process_template(self, group_type):
     ## end hooks
     def _replace_group_org(self, string):
         ''' substitute organization for group if this is an org'''
-        if self.group_type == 'organization':
-            string = re.sub('^group', 'organization', string)
```
davidread: Only the OrganizationController has customized logic functions - it is not to do with the group_type. Therefore this code has been moved into OrganizationController.

ckan/controllers/group.py
```diff
@@ -104,22 +99,25 @@ def _render_template(self, template_name):
...
         return render(self._replace_group_org(template_name))
 
-    def _redirect_to(self, *args, **kw):
-        ''' wrapper to ensue the correct controller is used '''
-        if self.group_type == 'organization' and 'controller' in kw:
-            kw['controller'] = 'organization'
+    def _redirect_to__this_controller(self, *args, **kw):
+        ''' wrapper around redirect_to but it adds in this request's controller
+        (so that it works for Organization or other derived controllers)'''
+        kw['controller'] = request.environ['pylons.routes_dict']['controller']
```
davidread: redirects and url_for should keep the same controller as the current request. That is better than trying to determine it from the group_type, which may be a custom one.

ckan/controllers/group.py
```diff
@@ -133,6 +131,21 @@ def _guess_group_type(self, expecting_name=False):
 
         return gt
 
+    def _ensure_controller_matches_group_type(self, id):
+        group = model.Group.get(id)
+        if group is None:
+            abort(404, _('Group not found'))
+        if group.type not in self.group_types:
+            abort(404, _('Incorrect group type'))
+        return group.type
```
davidread: This method is called at the start of most controller methods, to stop it working if the user tries using /organization/mygroup instead of /group/mygroup etc.

ckan/controllers/group.py
```diff
@@ -295,15 +305,17 @@ def pager_url(q=None, page=None):
                     facets[facet] = facet
 
             # Facet titles
-            for plugin in plugins.PluginImplementations(plugins.IFacets):
-                if self.group_type == 'organization':
+            from ckan.controllers.organization import OrganizationController
+            if isinstance(self, OrganizationController):
```
davidread: The choice between group_facets and organization_facets is down to whether we're in GroupController or OrganizationController. Ideally this would be a method which OrganizationController overrides, but it is neater to keep it inline. So we simply use isinstance.

ckan/controllers/group.py
```diff
@@ -395,7 +405,7 @@ def bulk_process(self, id):
         if not actions:
             # unicode format (decoded from utf8)
             limit = 500
-            self._read(id, limit)
+            self._read(id, limit, group_type)
```
davidread: Pass in group_type, instead of working it out again in _read.

davidread:
@wardi Do you have a chance to cast an eye over this and see if it is generally ok? I'm not sure if this completes IGroupForm, but it certainly now works in the basic cases of editing and viewing them.

Since people will use the default GroupController, I guess it is worth changing the group templates so that all the links are to /<group_type>/new rather than /group/new? Otherwise they'd have to provide custom templates for everything.

wardi:
@davidread yes, I'll take a look.

@deniszgonjanin have you started on some core changes to ckan for making scheming work with groups and orgs? This is surely related.

deniszgonjanin:
@wardi there was really only one change needed to make scheming work, and that was for IGroupForm to pass the group_type to the plugin. However that change was backwards incompatible the way I did it for my immediate purposes.

davidread:
@deniszgonjanin Could you be more specific about which function to/from
would helpfully pass group_type?

deniszgonjanin:
@davidread I am fairly sure it was https://github.com/ckan/ckan/blob/master/ckan/controllers/group.py#L59

`group_type` needs to be passed along to implementing plugins, for scheming to know which schema it is supposed to use.

davidread:
I see, thanks.

When you edit an existing group then the type is available as data_dict['type']. However if it is a new group then data_dict is {}. How about we add in the type into data_dict at this point? That would save changing the variables passed to setup_template_variables, and makes it consistent between new & edit forms. It just means the setup_template_variables method can't tell if it is a new or edit by seeing if data_dict == {}.

The alternative is to try calling setup_template_variables(context, data_dict, group_type) and if that raises an exception, try it without the group_type param. That keeps it backward compatible, but is perhaps not as neat. Thoughts?

wardi:
As much as possible I hope these changes will bring IGroupForm use more in-line with the recent IDatasetForm changes. In particular I resisted using setup_template_variables because it's really tricky to make sure that function gets called (and nothing else messes with a value in the context) before you need to read the value out for all cases.

Instead I explicitly pass dataset_type to the jinja2 render calls. It's a larger change but it feels less magical and easier to understand. https://github.com/ckan/ckan/pull/1795

davidread:
Thanks @wardi - I've now added group_type to all the render calls.
https://github.com/ckan/ckan/commit/70e51adb8195d27c1f9714ca70c71133e8fc906b

davidread:
It would be great to merge this at some point if it is ok @wardi

wardi:
@davidread yes, thanks for the reminder! I'll look at it now

ckan/controllers/group.py
wardi: I've never seen a double underscore in the middle of a method name. Is there some significance I'm missing?

ckan/controllers/group.py
```diff
@@ -227,15 +240,15 @@ def _read(self, id, limit):
         def search_url(params):
             if group_type == 'organization':
```
wardi: @davidread shouldn't this be checking is_organization instead? I ask because I see you fixed some code near this

ckan/controllers/group.py
wardi: then we have three ways testing if this is related to an organization. that doesn't seem good.

ckan/controllers/group.py
```diff
@@ -606,19 +635,19 @@ def delete(self, id):
         try:
             if request.method == 'POST':
                 self._action('group_delete')(context, {'id': id})
-                if self.group_type == 'organization':
-                    h.flash_notice(_('Organization has been deleted.'))
-                else:
-                    h.flash_notice(_('Group has been deleted.'))
-                self._redirect_to(controller='group', action='index')
+                h.flash_notice(_('%s has been deleted.')
+                               % group_type.capitalize())
```
wardi: should the parameter be in its own gettext call? if so that's a hidden string for the translators to find. if not the type is not translatable.

Or how about attaching a `deleted_flash_notice_message` object to the class and let the OrganizationController override it?

ckan/controllers/organization.py
```diff
@@ -1,19 +1,23 @@
...
 import ckan.controllers.group as group
 
 
 class OrganizationController(group.GroupController):
-    ''' The organization controller is pretty much just the group
-    controller. It has a few templates defined that are different and sets
-    the group_type to organization so that the group controller knows that
-    it is in fact the organization controller.  All the main logical
-    differences are therefore in the group controller.
+    ''' The organization controller is for Groups of type 'organization'. It
```
wardi: `''' The organization controller is for Organizations, which are implemented as ckan groups with is_organization=True`

ckan/controllers/organization.py
```diff
@@ -1,19 +1,23 @@
...
+    works the same as the group controller apart from:
+    * templates and logic action/auth functions are sometimes customized
+      (switched using _replace_group_org)
+    * 'bulk_process' action only works for organizations
 
-    The main differences the group controller provides for organizations are
-    a few wrapper functions that swap organization for group when rendering
-    templates, redirecting or calling logic actions '''
+    Nearly all the code for both is in the GroupController (for simplicity?).
```
wardi: for historical reasons, I hope

wardi:
Other than the comments I added this looks great @davidread 

I'm just curious what custom org types you have added that use this.

ckan/controllers/group.py
davidread: Mild significance... just a separator, like two spaces. This method is a variation on the familiar func 'redirect_to'. Rather than reading 'redirect to this controller', it more means 'redirect to whatever the *args/**kw say, as long as it is in this controller'. I'll remove the two if you prefer or can think of a better name.

ckan/controllers/group.py
davidread: I think group['is_organization'] and group_type=='organization' are interchangeable. I don't see a situation where you'd have one and not the other - or can you?

ckan/controllers/group.py
davidread: On the face of it these three are the same. I think I've been having half an eye on the situation where someone wants to inherit and customize OrganizationController. But really, rather than hypothecate wildly, I think we should just leave these alone until we have a proper use case and test, and then improve these. Otherwise my head will explode and will get it wrong anyway. What do you think?

ckan/controllers/group.py
davidread: > should the parameter be in its own gettext call? if so that's a hidden string for the translators to find. if not the type is not translatable.

Hmm I have an idea.

> Or how about attaching a deleted_flash_notice_message object to the class and let the OrganizationController override it?

If you attach the message to the class, then you can't customize it for a different group_type unless you create a subclass GroupController, which would not normally be necessary for an IGroupForm - as you can see from ckanext/example_igroupform/plugin.py.

ckan/controllers/group.py
```diff
@@ -608,17 +637,22 @@ def delete(self, id):
                 self._action('group_delete')(context, {'id': id})
                 if self.group_type == 'organization':
                     h.flash_notice(_('Organization has been deleted.'))
-                else:
+                elif self.group_type == 'group':
                     h.flash_notice(_('Group has been deleted.'))
-                self._redirect_to(controller='group', action='index')
+                else:
+                    h.flash_notice(_('%s has been deleted.')
+                                   % _(group_type.capitalize()))
```
davidread: I've put back the original flash messages, to avoid disrupting the default group_types. And although it is a hidden string here for a custom group_type, it is translatable if people spot it. Am not sure what else we can do.

davidread:
I've also got rid of the double underscores, in case that confuses any more.

davidread:
Ok that's built again now. What do you think?

ckan/controllers/group.py
wardi: IGroupForm.group_types lets plugins define different types for their orgs. I take that to mean is_organization is the definitive test. I haven't actually tried creating different org types, but that's what I hope to do with scheming in the future: make orgs as customizable as datasets, and one day maybe users too.

ckan/controllers/group.py
wardi: No, I'm confusing cases where we're looking at a specific org/group vs this controller search method. I agree with your comment that this would be nicer as a method overridden in OrganizationController, though. Hiding an isinstance checks in here makes it hard to discover the special case.

davidread:
Ok, I've addressed the points now I think. I've not tried it with any custom org types. Merely following up on the issue raised here: https://github.com/ckan/ckan/issues/2344 and added a v. basic test for custom group type. It's all a bit theoretical use, but at least it is functional now...

davidread:
Please do review @wardi when you have a chance

wardi:
@davidread thanks for the ping, will do

wardi:
@davidread looks great! I'm looking forward to using this. Please resolve the conflict and I will merge.

davidread:
Cheers @wardi

PR #2355, opened 2015-03-18
allow disabling of repoze.who middleware during tests

joetsoi:
repoze.who middleware intercepts 401s and changes it to a 302 redirect to the login page. This is good behaviour in production.

In tests I want to know that a 401 was correctly raised.

e: note this occurs if i'm filling a form in with web test and submitting it. I can just directly post using webtest to generate the 401


joetsoi:
closing this as a I can generate the 401 in tests by posting directly, if I'm expecting a 401, then arguably the form should not be appearing to the user in the first place.

Issue #2187, opened 2015-01-06
RFE: include hidden form fields when confirming actions


ThrawnCA:
We would like to implement token-based CSRF protection on POST requests; however, actions that involve confirmation (like deleting a resource) use JavaScript to assemble a POST without incorporating the token.

Would it be reasonable for confirmation requests to incorporate all hidden fields from the form containing the link that was clicked?

Suggested implementation is to add the following line to performAction function in confirm-action.js, immediately after constructing the form (line 69 as at 2015-01-07):

    $(this.el).closest('form').find('input[type=hidden]').appendTo(form);

ThrawnCA:
The JQuery snippet appears to need an update:

 $('form[method=POST]').find('input[type=hidden]').appendTo(form);

ThrawnCA:
Won't be implemented

ThrawnCA:
If anyone is interested, our CSRF filter now works without this patch, and is available at https://github.com/ThrawnCA/ckanext-qgov/blob/master/ckanext/qgov/common/anti_csrf.py

saurabhjangir:
@ThrawnCA I was looking for CSRF protection in CKAN code and I came to know about this issue. I was just wondering why this isn't implemented and not a part of core CKAN code. Could you please help me out regarding this.

ThrawnCA:
@saurabhjangir The discussion at https://github.com/ckan/ckan/pull/2359 concluded with the core developers deciding that this would conflict with their future CSRF protection plans.

It should be possible to use the extension with core code in its current state; however, if I recall correctly, it needs a bugfix to make the 'remove member from organisation' links work.

PR #2359, opened 2015-03-23
Include hidden fields when confirming actions in JavaScript

ThrawnCA:
See #2187


wardi:
My concern is this change is too broad: will every hidden field from every form get sucked in to every post? That sounds dangerous.

I'd feel better about this change if this had a very specific selector. For example on field with id='form-csrf-token' or similar.

ThrawnCA:
Well, it will only affect forms with method POST. But yes, as it stands, it will pull every hidden field from those forms, when doing a popup confirmation. I don't think that confirm-action.js is used in many places, though: dataset/resource/user deletion is all I can find.

The original JQuery snippet on #2187 went looking for the form closest to the popup, which would have been better, and it seemed to work in 2.0, but not 2.2, unfortunately. If anyone knows a way to make it work in the latest CKAN, then please go ahead.

It would be possible, of course, to target a specific form field (the CSRF token). But I thought, since the confirmation dialog is assembling a form that mimics the original, it makes sense to include any hidden fields that the server had inserted. Much more future-proof.

brew:
> since the confirmation dialog is assembling a form that mimics the original, it makes sense to include any hidden fields that the server had inserted

I agree, and don't think there's a problem to include hidden fields as a general case for these forms. @wardi, can you explain what your concern is? If there's a field present in the original form, I don't see a problem with including it in the submit for the confirm as well.

However, it looks like the jquery snippet as currently presented will add the hidden fields from _all_ forms (with a post method) in the document, not just the one being confirmed. @ThrawnCA, can you make this more specific.

wardi:
@brew that's just it. I'm worried that this would pull fields out of unrelated forms. And would cause lots of confusion for someone trying to debug.

ThrawnCA:
OK, it looks like the original snippet not working was an anomaly; it's working OK on my local machine. I've committed it now. It will pull fields only from the form containing the link originally clicked.

brew:
@ThrawnCA After further discussion, I think I agree with @wardi that we shouldn't include unrelated fields in the confirm form submit, and that we should be specific with adding the token to the confirm form.

Instead of using a hidden field, I would suggest passing the csrf token to the confirm-action module as a `data-module-*` attribute on the `a` tag that sets up the confirm. Something like: 

```jinja
<a class="btn btn-danger pull-left" href="{% url_for controller='package', action='delete', id=data.id %}" data-module="confirm-action" data-module-csrf-token="{{ csrf_token }}" data-module-i18n="{{ locale }}">{% block delete_button_text %}{{ _('Delete') }}{% endblock %}</a>
```

Then use that value as an option in confirm-action.js when the form is constructed in the `performAction` function.

ThrawnCA:
Well, that would work, of course.

It feels more intrusive to me, because any form making use of the confirm-action module would need to specify the token, which means getting access to it when rendering the page...all in all, it means touching a lot of files, and it's guaranteed to break any extensions that aren't aware of it.

I've been working on a filter that inserts tokens into POST forms on the fly, so that it's not necessary to alter the static source of every page. I suppose, in theory, it could perform a similar insertion of the csrf-token option...though it would be harder to avoid double insertions.

Why do you consider hidden fields in the same form to be "unrelated"? They haven't previously been *needed*, because the confirmation action is currently only used for deletions - but I don't see that as the same thing. If someone wanted to confirm whenever dataset metadata was changed, or perhaps whenever a specific field was changed, then suddenly every field in the form (hidden or not, in fact) would indeed be relevant.

ThrawnCA:
How about this - instead of looking for a specific field name, the snippet could look for a CSS class. So, any inputs with class 'confirm-action' would be included. That restricts it to fields that were purposely included, without imposing a specific name (and being theoretically usable for multiple fields).

ThrawnCA:
Latest commit changes it to pull in only fields with the 'confirm-action' class.

ThrawnCA:
Bump

brew:
Sorry, I'll try and look at this later in the week.

ThrawnCA:
It would be feasible, if you want, to use a data- attribute, instead of a CSS class, to flag the field for inclusion. Providing the whole token in the attribute is more invasive, since it means that the token must be available to the renderer, but for setting a flag, a data- attribute would be fine.

brew:
Yes. I'd prefer the token was passed into the module with a data- attribute. If the token is being passed to templates so it can be used as the value of a hidden field, then it can also be used as the value of a `data-module-csrf-token` attribute. This explicitly give the module the data it needs within the link that triggers the confirm dialog without adding unrelated classes to other elements.

ThrawnCA:
No, as previously mentioned, I don't want to provide the actual token value as an attribute, because I don't want to have to make the token available in the rendering context.

But I'm happy to use a data- attribute to flag input fields for inclusion.

ThrawnCA:
Sorry about the huge merge commit; I'm not sure of the best way to avoid that.

brew:
I don't understand. Aren't you intending to render the token as a value of a hidden field?

Sorry to keep pushing on this, but my concern here is that we want to have a mechanism for CSRF protection in core, and changes here will likely define how we go about doing that. I'm not hard against what you're proposing, I just want to understand how you plan to use it.

ThrawnCA:
I plan to use it by intercepting the rendering function and doing a regex-based HTML rewrite. It may sound ugly, but it's actually very non-intrusive, and it should work even if other extensions generate their own forms.

Filter code looks like this:

    import ckan.lib.base as base
    import re
    from re import DOTALL, IGNORECASE, MULTILINE
    from genshi.template import MarkupTemplate
    from logging import getLogger
    from pylons import request, response

    LOG = getLogger(__name__)

    RAW_RENDER = base.render
    RAW_RENDER_JINJA = base.render_jinja2
    RAW_BEFORE = base.BaseController.__before__

    POST_FORM = re.compile(r'(<form [^>]*method=["\']post["\'][^>]*>)([^<]*\s<)', IGNORECASE | MULTILINE)
    TOKEN_PATTERN = r'<input type="hidden" data-module-confirm-action name="token" value="{token}"/>'
    TOKEN_SEARCH_PATTERN = re.compile(TOKEN_PATTERN.format(token=r'([0-9a-f]+)'))
    API_URL = re.compile(r'^/api\b.*')

    def is_logged_in():
        return request.cookies.get("auth_tkt")

    # Insert token into applicable responses

    def anti_csrf_render(template_name, extra_vars=None, cache_key=None, cache_type=None, cache_expire=None, method='xhtml', loader_class=MarkupTemplate, cache_force=None, renderer=None):
        html = apply_token(RAW_RENDER(template_name, extra_vars, cache_key, cache_type, cache_expire, method, loader_class, cache_force, renderer))
        return html

    def anti_csrf_render_jinja2(template_name, extra_vars=None):
        html = apply_token(RAW_RENDER_JINJA(template_name, extra_vars))
        return html

    def apply_token(html):
        if not is_logged_in() or not POST_FORM.search(html):
            return html

        token_match = TOKEN_SEARCH_PATTERN.search(html)
        if token_match:
            token = token_match.group(1)
        else:
            token = get_server_token()

        def insert_form_token(form_match):
            return form_match.group(1) + TOKEN_PATTERN.format(token=token) + form_match.group(2)

        return POST_FORM.sub(insert_form_token, html)

    def get_server_token():
        if request.environ['webob.adhoc_attrs'].has_key('server_token'):
            token = request.server_token
        elif request.cookies.has_key("token"):
            token = request.cookies.pop("token")
        else:
            import binascii, os
            token = binascii.hexlify(os.urandom(32))
            response.set_cookie("token", token, max_age=600, httponly=True)

        if token is None or token.strip() == "":
            csrf_fail("Server token is blank")

        request.server_token = token
        return token

    # Check token on applicable requests

    def is_request_exempt():
        return not is_logged_in() or API_URL.match(request.path) or request.method in {'GET', 'HEAD', 'OPTIONS'}

    def anti_csrf_before(obj, action, **params):
        if not is_request_exempt() and get_server_token() != get_post_token():
            csrf_fail("Could not match session token with form token")

        RAW_BEFORE(obj, action)

    def csrf_fail(message):
        from pylons.controllers.util import abort
        LOG.error(message)
        abort(403, "Your form submission could not be validated")

    def get_post_token():
        if request.environ['webob.adhoc_attrs'].has_key('token'):
            return request.token

        postTokens = request.POST.getall('token')
        if not postTokens:
            csrf_fail("Missing CSRF token in form submission")
        elif len(postTokens) > 1:
            csrf_fail("More than one CSRF token in form submission")
        else:
            request.token = postTokens[0]

        # drop token from request so it doesn't populate resource extras
        del request.POST['token']

        return request.token

    def intercept_csrf():
        base.render = anti_csrf_render
        base.render_jinja2 = anti_csrf_render_jinja2
        base.BaseController.__before__ = anti_csrf_before

brew:
I've discussed this with the CKAN development team and the approach you're proposing will conflict with future plans for implementing CSRF protection in CKAN core.

If you want to pursue this direction, I'd suggest providing your own confirm-action module in an extension that supports your implementation.

ThrawnCA:
Well, I've found a way to do this in a self-contained fashion, anyway, by re-writing the confirm-action links at the same time I re-write POST forms. So never mind.

PR #2361, opened 2015-03-26
Missing position argument causes broken test in Python 2.6

brew:
In datastore/logic/action[1] there's a missing positional argument for the format string that's causing a test (ckanext.datastore.tests.test_info.TestDatastoreInfo.test_info_success) to break in python 2.6.

Not sure why this is passing on Travis, but it's broken on my local machine.

[1] https://github.com/ckan/ckan/blob/066c450be9b10629921db445f1309bd264fe6aeb/ckanext/datastore/logic/action.py#L273

### DIFF
--- a/ckanext/datastore/logic/action.py
+++ b/ckanext/datastore/logic/action.py
@@ -270,7 +270,7 @@ def datastore_info(context, data_dict):
         # We need to make sure the resource_id is a valid resource_id before we use it like
         # this, we have done that above.
         meta_sql = sqlalchemy.text(u'''
-            SELECT count(_id) FROM "{}";
+            SELECT count(_id) FROM "{0}";
         '''.format(resource_id))
         meta_results = db._get_engine(data_dict).execute(meta_sql, resource_id=resource_id)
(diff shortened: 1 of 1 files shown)


joetsoi:
we should really use sqlalchemy's [text]()(http://docs.sqlalchemy.org/en/latest/core/tutorial.html#using-text) properly. Even if resource goes through a schema and it's checked that resource exists.

wardi:
worrying that travis didn't catch this

joetsoi:
https://coveralls.io/builds/2245255/source?filename=ckanext%2Fdatastore%2Flogic%2Faction.py#L219

looks like it's not covered in the postgres 8.4 tests, https://github.com/ckan/ckan/blob/master/ckanext/datastore/tests/test_info.py#L29 as it's in legacy mode.

joetsoi:
Ah, I remember, we turned off the postgres 9/python26 and postgres 8/python27 combos in travis, because having 4 travis builds takes ages and the common combo for linux distros is to have postgres 8/python 2.6 and postgres 9 and python 2.7.

So i'm going to merge and not worry so much about the tests

PR #2363, opened 2015-03-27
Allow extension writers to add tabs to ckan-admin navigation

### LINKED ISSUE #2351: No clear way to add a tab to ckan-admin page
If you go to /ckan-admin (as a sysadmin) there are some tabs for config, trash etc.  As a plugin author, there's no obvious way for me to provide info about a tab that I would like adding to that page. For instance if I need a config page for my plugin, it seems like ckan-admin would be the perfect place - but I will need to override the admin template just to insert a link.

Perhaps it would be better if there was a way of doing this without copying/overriding templates?

brew:
Introduces a new method to allow extension writers to add their own tabs to the ckan-admin page:

`plugins.toolkit.add_ckan_admin_tab(<route name>, <tab label>)`: an extension can use this within update_config to register a new ckan-admin tab.

An example iconfigurer extension has been added to ckanext as an example and for testing.

Closes #2351.


amercader:
@brew I merged this as it looked great. At some point it'd would be nice if this was documented in the Extending docs, similar to this page or part of it:

http://docs.ckan.org/en/latest/extensions/custom-config-settings.html

PR #2371, opened 2015-03-30
user_show include_datasets should only include datasets of type dataset by default

brew:
Restricting user_show to only include datasets of type `dataset` by default will prevent inappropriate datasets with custom type from showing up in user_show when `include_datasets` is True.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1306,6 +1306,11 @@ def user_show(context, data_dict):
         that are draft or private.
         (optional, default:``False``, limit:50)
     :type include_datasets: boolean
+    :param include_all_dataset_types: If include_datasets is True, this will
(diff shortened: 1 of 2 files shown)


brew:
I've included a new option for user_show's data dict: `include_all_dataset_types` will not restrict `include_datasets` by type. The default is False.

brew:
After discussion in the dev call, concerns were raised that some users specify a different type as their standard type, rather than `dataset`. Or they have more than one type that they consider to be their main dataset type.

So, instead of filtering _in_ `dataset`, extensions should instead filter themselves _out_ somehow. The suggestion is to have some sort of 'don't think of me as a dataset' option in the IDatasetForm interface, a bit like `is_fallback`.

brew:
Will create a new PR.

Issue #2374, opened 2015-04-02
Add num_followers to ckan.logic.action.get.package_show


u10313335:
Should we add num_followers to ckan.logic.action.get.package_show to prevent from using deprecated h.get_action() in ckan/templates/package/snippets/info.html? I will send a pull request later.

wardi:
see https://github.com/ckan/ckan/pull/2375#issuecomment-88950394

u10313335:
Just wondering if it is okay to change:
h.SI_number_span(h.get_action('dataset_follower_count', {'id': pkg.id})) }} in ckan/templates/package/snippets/info.html
to something like
h.SI_number_span(h.follow_count('dataset', pkg.id)) to get the number of followers of a dataset (the h.follow_count is a built-in helper function).

wardi:
@u10313335 sounds perfect

u10313335:
Thanks @wardi, I will create a new issue and work on it.

PR #2375, opened 2015-04-02
2374 Add num_followers to ckan.logic.action.get.package_show

u10313335:
Fix for issue #2374


wardi:
@u10313335 thanks. These "include_..." parameters are introduced to allow us to slowly remove features that have been slowing down our API calls. They aren't meant to be a model for adding more optional information to package_show etc.

Instead would you create a new helper function to get the num_followers and any other information that the info snippet needs? Adding a helper is something any extension developer can do without needing to change core, so it's a better example for people extending templates.

u10313335:
Thanks @wardi, I will try to do that!

Issue #2370, opened 2015-03-30
Password reset token are not invalidated [security]


KrzysztofMadejski:
I can use generated token multiple times to reset password. It should be invalidated after first successful password change!

Concerns: CKAN 2.3

wardi:
@KrzysztofMadejski absolutely, you're right. Do you have time to submit a fix?

For security related issues like this one please send your future reports to security@ckan.org

KrzysztofMadejski:
sorryy, I will use it in the future!

I see it's already PRed.

tino097:
Looks like this was fixed, also its targeting old release. Closing

PR #2379, opened 2015-04-02
add reset for reset_key on successful password change

nateprewitt:
and associated test.

Fix for #2370

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -462,8 +462,6 @@ class UserController(base.BaseController):
 
     def perform_reset(self, id):
         # FIXME 403 error for invalid key is a non helpful page
-        # FIXME We should reset the reset key when it is used to prevent
-        # reuse of the url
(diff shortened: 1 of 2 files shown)


wardi:
@nateprewitt nice! Thanks.

wardi:
We're not adding new tests to tests/legacy at this time, would you move the test to tests/controllers/test_user.py instead?

nateprewitt:
Yep, wasn't sure where to put it so I grouped it with the other perform_reset tests. I'll move it over now.

nateprewitt:
I had to import create_reset_key from [mailer.py](https://github.com/ckan/ckan/blob/master/ckan/lib/mailer.py#L154) to initialize the reset_key in the model.

If there's a cleaner interface to do this, let me know and I'll make the fix.

wardi:
@amercader how far can we backport this one?

amercader:
@wardi all the way down to 2.0!

PR #2380, opened 2015-04-03
Refactor user_show to use package_search instead of package_show

brew:
`user_show` to use `package_search` instead of `package_show` when including datasets in results. This means a change to `package_search` to return draft datasets (when a flag in the context allows drafts, similar to how `ignore_capacity_check` works to allow private datasets). Solr will also now index drafts.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1696,7 +1696,7 @@ def package_search(context, data_dict):
                           if 'capacity:' not in p)
             data_dict['fq'] = fq + ' capacity:"public"'
 
-        # If present, Solr doesn't need 'include_drafts', so pop it.
+        # Solr doesn't need 'include_drafts`, so pop it.
         include_drafts = data_dict.pop('include_drafts', False)
         fq = data_dict.get('fq', '')
         if include_drafts:
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1358,17 +1358,13 @@ def user_show(context, data_dict):
 
         fq = "+creator_user_id:{0}".format(user_dict['id'])
 
+        search_dict = {'rows': 50}
+
         if include_private_and_draft_datasets:
             context['ignore_capacity_check'] = True
-            context['allow_drafts'] = True
-            fq += " +state:(active OR draft)"
-        else:
-            fq += " +state:active +capacity:public"
+            search_dict.update({'include_drafts': True})
 
-        search_dict = {
-            'fq': fq,
-            'rows': 50
-        }
+        search_dict.update({'fq': fq})
 
         user_dict['datasets'] = \
             logic.get_action('package_search')(context=context,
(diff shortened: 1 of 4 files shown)


ckan/tests/logic/action/test_get.py
```diff
@@ -41,6 +41,15 @@ def foo(key, data, errors, context):
...
         eq(dataset2['new_field'], 'foo')
 
+
+class TestGroupList(object):
+
+    def setup(self):
+        helpers.reset_db()
+
+        # Clear the search index
+        search.clear()
```
wardi: I think we have a FunctionalTestBase or similar that does the reset_db etc. (so you can save a few extra lines on all these classes)

brew: Good point!

davidread:
You have test_package_search_includes_drafts_with_allow_drafts_in_context but you should also test that drafts owned by other users aren't visible to this user even if allow_drafts=True.

btw wouldn't it be better as include_drafts rather than allow_drafts?

wardi:
+1 include_foo seems common in other actions

brew:
As discussed in ckan dev call:
Rather than add another special context flag to allow drafts, instead allow authorized users and sysadmins to pass a flag in the data_dict that forms part of the action request and handle it solely in `package_search`. This syncs up the ability to return drafts between the web and api.

@wardi @davidread: The reason for 'allow' rather than 'include' is because the `fq` value passed to `package_search` still needs to include a `state` key that explicitly includes drafts in the result. Otherwise `query.run` will automatically add `+state:active`, regardless of the value of `allow_drafts`. That way of thinking about it may change anyway if the implementation changes, as discussed above.

ckan/logic/action/get.py
```diff
@@ -1355,25 +1355,21 @@ def user_show(context, data_dict):
...
-                continue
-            user_dict['datasets'].append(dataset_dict)
+
+        fq = "+creator_user_id:{0}".format(user_dict['id'])
+
+        search_dict = {'rows': 50}
+
+        if include_private_and_draft_datasets:
+            context['ignore_capacity_check'] = True
+            search_dict.update({'include_drafts': True})
```
wardi: seems like a long way to write `search_dict['include_drafts'] = True`, but, ok

ckan/logic/action/get.py
```diff
@@ -1692,9 +1693,25 @@ def package_search(context, data_dict):
...
+            user_id = authz.get_user_id_for_username(user, allow_none=True)
+            if authz.is_sysadmin(user):
+                data_dict['fq'] = fq + ' +state:(active OR draft)'
+            elif user_id:
+                # Query to return all active datasets, and all draft datasets
+                # for this user.
+                data_dict['fq'] = fq + \
+                    ' ((creator_user_id:{0} AND +state:(draft OR active))' \
+                    ' OR state:active)'.format(user_id)
+        elif not authz.is_sysadmin(user):
```
wardi: doesn't this mean sysadmins get all drafts even if they have set include_drafts=False?

wardi:
@brew except for the sysadmin with include_draft=False case this all looks good.

ckan/logic/action/get.py
brew: If `+state` hasn't been already been set, it will be set in `PackageSearchQuery.run()` to `+state:active`[1], thereby excluding drafts by default. That `elif` on L1712 allows only sysadmins to add a state to their queries explicitly if they want. I've added a test for sysadmins with `include_drafts=False` that passes as expected.

[1] https://github.com/ckan/ckan/blob/b5c43f0ece24f8ff2fd4dd05651a11c1f8e9e23d/ckan/lib/search/query.py#L329

wardi:
Thanks @brew I got confused by some of the checks happening in the action and some in the lib function.

wardi:
maybe i'll update the action and PackageSearchQuery.run docstrings to describe these features at some point.

brew:
Thanks for checking this out!

davidread:
nice work @brew

Issue #2383, opened 2015-04-03
Change the way of getting the number of followers of a dataset in the package info snippet


u10313335:
Since h.get_action() is deprecated in the current CKAN release, I proposed a new way to get the number of followers of a dataset in ckan/templates/package/snippets/info.html:

```diff
diff --git a/ckan/templates/package/snippets/info.html b/ckan/templates/package/snippets/info.html
index 4a17a56..32e2edb 100644
--- a/ckan/templates/package/snippets/info.html
+++ b/ckan/templates/package/snippets/info.html
@@ -21,7 +21,7 @@ Example:
               <div class="nums">
                 <dl>
                   <dt>{{ _('Followers') }}</dt>
-                  <dd>{{ h.SI_number_span(h.get_action('dataset_follower_count', {'id': pkg.id})) }}</dd>
+                  <dd>{{ h.SI_number_span(h.follow_count('dataset', pkg.id)) }}</dd>
                 </dl>
               </div>
             {% endblock %}
```

Will submit a PR later. Thanks for @wardi's advice in #2375.

joetsoi:
thanks again

PR #2384, opened 2015-04-04
2383 Change the way of getting the number of followers of a dataset in the package info snippet

u10313335:
Fix for issue #2383


joetsoi:
looks good!

PR #2382, opened 2015-04-08
package_search performance

wardi:
Let's turn the repeated sql queries that verify packages returned by solr into a single query with something like  `.filter(model.Package.id.in_(package['id'] for package in query.results))` and see if it makes package_search faster.

Same idea as #1924 but with a different implementation

update: Some performance numbers on my test machine (lowest of 10 requests):

| version | time | vs. master |
| --- | --- | --- |
| master | 0.291 | 100% |
| #2382 | 0.222 | 76% |


wardi:
coverage decreased because I deleted code :-P

joetsoi:
... did I not merge this already? @wardi sorry could you fix tests and I'll merge

wardi:
@joetsoi done!

Issue #2110, opened 2014-12-03
Add not existing License


EnxEng:
I have tried to add a new license to CKAN.
I followed the documentation:
http://docs.ckan.org/en/943-writing-extensions-tutorial/configuration.html#licenses-group-url

and so, created a new json, by merging it with the one online:
http://licenses.opendefinition.org/licenses/groups/ckan.json

But a have an error, when creating a new dataset:
KeyError: 'is_okd_compliant'

In fact I have noticed from the source code, you changaed somethin in the revisiotns, and in order to make License, transatable, you have diffrent keys, fomr the ones officialy used by ODK.
I notices that all keys are the same, except for:
            'is_generic',
            'is_okd_compliant',
            'is_osi_compliant',

that you use in ckan/model/license.py
where, Open Definition, uses:
  "od_conformance": 
  "osd_conformance".

I think that keys in the source code should be updated.

kapucko:
I got the same problem. Do you have a solutions @EnxEng?

kapucko:
I added keys 'is_okd_compliant' and 'is_osi_compliant' to each license in json...

Conzar:
I have found this error when preparing to demonstrate a different error (with spatial extension).  This error is reproducible here: 
https://bitbucket.org/landcareresearch/vagrant-ckan-spatial-error

joetsoi:
yeah the opendefinition license.json is missing is_okd_compliant which causes ckan to break, I will bring it up in tomorrows dev meeting to decided whether we should fix license.json or remove the checks for is_okd_compliant references from ckan.

joetsoi:
@Conzar I recommend for now you create a new license.json and set your license_group_url to that you can generate the right one using the ridiculous oneliner

`curl http://licenses.opendefinition.org/licenses/groups/ckan.json | jq 'map(.+{is_okd_compliant: (if .od_conformance == "approved" then true else false end)})' | jq 'map(.+{is_osi_compliant: (if .osd_conformance == "approved" then true else false end)})'`

although I'm sure you could do it in one jq command

davidread:
I've done a pull request for moving to the new keys: https://github.com/ckan/ckan/pull/2391
I can't see any reason to be left behind licenses.opendefinition.org

metaodi:
@joetsoi @davidread 
[...]

PR #2391, opened 2015-04-13
[#2110] Update license keys to match opendefinition.org.

davidread:
It is backward compatible with old licence files.

The only mild concern is that in extensions, if monkeying with this, people have used license.isopen() as it is designed (rather than license.is_okd_compliant which is more esoteric). It's a bit of pain to smoothe this, since there are three ways that the license keys are presented ( license.is_okd_compliant, license['is_okd_compliant'] & license.as_dict()['is_okd_compliant']) and you'd have to fix it in all three. Unless you store both keys, and I'd rather not see the old keys when the licence is shown in the CKAN API.


davidread:
Backwards compatibility now sorted. Ready for review @joetsoi

davidread:
Ping @joetsoi Hopefully this is straightforward to review, with clear tests.

Issue #2350, opened 2015-03-16
User schema improperly validated


KrzysztofMadejski:
Commit with correction will follow. UserController.edit is calling completly wrong function to validate schema

KrzysztofMadejski:
Forked from stable CKAN 2.3

KrzysztofMadejski:
:)

PR #2393, opened 2015-04-14
[#2350] Fix validation of user schema.

davidread:
See issue: https://github.com/ckan/ckan/issues/2350

I've not written any tests for this as it is a reasonably big job to setup a test extension, inherit UserController and setup routes. I don't think we're very happy with how the user schema is customized - it would make sense to convert to an IUserForm, so investing in this old way seems pointless. So I suggest we just fix this as @KrzysztofMadejski suggests and not worry about a test until this area is looked at properly by someone.

BTW @KrzysztofMadejski inherits UserController and adds an extra validator and a couple of new fields:
https://github.com/DanePubliczneGovPl/ckanext-danepubliczne/blob/develop/ckanext/danepubliczne/controllers/user.py#L79

### DIFF
--- a/ckan/templates/user/edit_user_form.html
+++ b/ckan/templates/user/edit_user_form.html
@@ -1,6 +1,6 @@
 {% import 'macros/form.html' as form %}
 
-<form class="dataset-form form-horizontal" method="post" action="{{ action }}">
+<form id="user-edit" class="user-form form-horizontal" method="post" action="{{ action }}">
   {{ form.errors(error_summary) }}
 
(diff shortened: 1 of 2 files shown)


KrzysztofMadejski:
+1 for *IUserForm* :)

davidread:
I'll just add a basic functional test for editing a user.

davidread:
I've added the test, so ready for review.

davidread:
This is a one-line fix with a test, ready for review, ping @amercader ;-)

PR #2327, opened 2015-02-27
Allows creating dataset without the need to specify resource

Jan-Marcek:
Now in the second stage of creating dataset, it's needed to specify at least one resource. Otherwise it's not possible to continue to the next stage of creating dataset.

This enhancement adds new property "ckan.dataset.create.require.resource". If its not specified or has value "True", CKAN behaves normally (requires resource). If it's value is "False", you can continue to the next stage.

PR #2392, opened 2015-04-15
Refering to non-existing action when creating new dataset

wardi:
This causes 404 Not Found http error

When creating new dataset you can replicate this error with following steps:
1. Click create new dataset
2. go to second "new resource" stage, input the info for the resource and click "Save & add another"
3. click "Finish" button

The cause is commit https://github.com/ckan/ckan/commit/beef24f7e0956ee5a5a00db0237da37ea8a52aa4
in ckan/controllers/package.py removes the new_metadata action, but its still referenced in new_resource method [here](https://github.com/ckan/ckan/blob/b5c43f0ece24f8ff2fd4dd05651a11c1f8e9e23d/ckan/controllers/package.py#L673-L700) (line 700)

its also referenced in the [_save_new](https://github.com/ckan/ckan/blob/b5c43f0ece24f8ff2fd4dd05651a11c1f8e9e23d/ckan/controllers/package.py#L933-L937) method, but I couldn't confirm the steps to cause this error.

and in the [ckan/config/routing.py](https://github.com/ckan/ckan/blob/b5c43f0ece24f8ff2fd4dd05651a11c1f8e9e23d/ckan/config/routing.py#L225) it wasn't removed

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -222,7 +222,6 @@ def make_map():
                   ])))
         m.connect('/dataset/{action}/{id}',
                   requirements=dict(action='|'.join([
-                      'new_metadata',
                       'new_resource',
                       'history',
                       'read_ajax',
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -695,9 +695,13 @@ class PackageController(base.BaseController):
                         errors = {}
                         error_summary = {_('Error'): msg}
                         return self.new_resource(id, data, errors, error_summary)
-                # we have a resource so let them add metadata
+                # XXX race condition if another user edits/deletes
+                data_dict = get_action('package_show')(context, {'id': id})
+                get_action('package_update')(
+                    dict(context, allow_state_change=True),
+                    dict(data_dict, state='active'))
                 redirect(h.url_for(controller='package',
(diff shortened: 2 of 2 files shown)


wardi:
converted to PR with the fix referenced by @mvi-eea-sk

wardi:
@mvi-eea-sk instead of copying the go-metadata code from below can't we just remove the redirect?

mvi-eea-sk:
@wardi no, the creation process needs to be completed there

current (should be) state:
you add resource through the "Save & add another", but you change your mind and you don't want to add another one so you click "Finish" and dataset is created.

if the redirect is removed:
you add resource through the "Save & add another", but you change your mind and you don't want to add another one so you click "Finish" and ... it won't let you finish the process. You would have to create another resource and delete it afterwards.

Removing the redirect won't cause any errors (checked it) but isn't user friendly at all.

Also, in the #2327 it is made so, that you won't need to add the resource at all and it's counting on the redirect.

wardi:
@amercader what's the protocol, do I need to ping you for backports or is adding some tags enough?

amercader:
@wardi labels should be all right.

I think it's time to prepare patch releases, there is enough stuff already.

PR #2406, opened 2015-04-24
Major revision/update/completion of Greek translation of CKAN 2.2.x

kalxas:
Full Greek translation of CKAN 2.2 as it will appear in geodata.gov.gr


amercader:
This is great @kalxas !

Rather than merge the translation file directly into the repo, can you upload the po file to Transifex?

https://www.transifex.com/projects/p/ckan/resource/2-2/

This way it will get pulled from Transifex with the rest of the languages before releasing 2.2.3. I believe you have permissions to edit the translations there but let me know if not.

Σας ευχαριστώ πολύ!

kalxas:
Hi @amercader 
This is a major edit, not a review, so it would be better to merge here and then copy from here to transifex.

amercader:
@kalxas that sounds good, as long as Transifex is updated asap after the merge to avoid conflicts

kalxas:
@amercader yes, I will update transifex right after the merge. Thanks!

amercader:
@kalxas 2 formatting errors:
```
paster check-po-files ckan/i18n/el/LC_MESSAGES/ckan.po         

Checking file ckan/i18n/el/LC_MESSAGES/ckan.po
    Format specifiers don't match:
    Show More {facet_type} -> Εμφάνιση Περισσοτέρων
    Format specifiers don't match:
    Show Only Popular {facet_type} -> Εμφάνιση Μόνο των Δημοφιλέστερων
```

kalxas:
@amercader actually this is an intentional fix: 
In order to use {facet_type} in this case in a Greek translation, we would need to conjugate each facet type according to gender for the whole sentence to make sense. 
Nouns in Greek have cases (nominative, genitive etc) and also they have genders (a group is a female and an organization is a male), so those messages are not generic if we retain the facet type. What you see is the Greek plural genitive which includes all.

amercader:
@kalxas cool, we should probably just update the string to "Show more" and "Show less" in future versions.

Merged now, I'll compile the mo file before the release, please update the translations on Transifex

kalxas:
Thanks @amercader I will update transifex now

kalxas:
done

PR #2407, opened 2015-04-27
Support for new reCAPTCHA v2.0

AdzRivers:
Support for the new reCAPTCHA on the registration form as seen here: https://www.google.com/recaptcha/intro/index.html.

newrecaptchasupport

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -115,6 +115,7 @@ ckan.preview.loadable = html htm rdf+xml owl+xml xml n3 n-triples turtle plain a
 # package_hide_extras = for_search_index_only
 #package_edit_return_url = http://another.frontend/dataset/<NAME>
 #package_new_return_url = http://another.frontend/dataset/<NAME>
+#ckan.recaptcha.version = 1
 #ckan.recaptcha.publickey =
 #ckan.recaptcha.privatekey =
 #licenses_group_url = http://licenses.opendefinition.org/licenses/groups/ckan.json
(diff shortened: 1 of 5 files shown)


amercader:
Many thanks for this @AdzRivers, seems like a useful addition.
AFAICT v2.0 is still a relatively recent technology and some people may want to keep their existing implementation on their sites. Perhaps we could keep support for the previous version adding a `ckan.recaptcha.version` config option that allows you to choose which version to load, and the code to handle the request differently.

What do you think?

AdzRivers:
@amercader I've added a config option called `ckan.recaptcha.version`, which changes the version of the recaptcha api used.

AdzRivers:
@TkTech Well spotted, thanks for that! They're now fixed.

astatke:
I was trying to hide the "Privacy and Terms" using this css but to no avail:
.rc-anchor-pt {display: none;}

Could you advise as to how to achieve that?

mattfullerton:
Try visibility: hidden;

astatke:
mattfullerton - thank you for your prompt response.

Unfortunately, it still doesn't work with visibility: hidden;

mattfullerton:
Do you have an instance of the change running publicly? Depending on what order different CSS files are getting loaded, your change might be getting overridden.

astatke:
It is still in development. Below is a screenshot on my followup findings/question:
styling-recaptcha

mattfullerton:
Why exactly do you want to remove privacy and terms? ;-)

Assuming the method you suggest is possible, maybe you can use a different kind of selector ($("iframe")) with jquery to get a hold of the iframe instead of the id.

astatke:
I guess leaving the labels as is should be OK too :)

I will give it a shot and let you know my findings.

~cheers and have a great day

Issue #2387, opened 2015-04-04
recline_map_view does not recognize GeoJSON fileds


u10313335:
If we try to create a recline_map_view (CKAN 2.3+) for data stored in the DataStore, the GeoJSON fields can not be listed since they are "text" fields rather than "numeric" fields.

It can be fixed by adding "text" type to the ReclineMapView extension (tested in my CKAN instance):

```diff
diff --git a/ckanext/reclineview/plugin.py b/ckanext/reclineview/plugin.py
index 20e204b..cc59163 100644
--- a/ckanext/reclineview/plugin.py
+++ b/ckanext/reclineview/plugin.py
@@ -178,7 +178,7 @@ class ReclineMapView(ReclineViewBase):

     datastore_fields = []

-    datastore_field_types = ['numeric']
+    datastore_field_types = ['numeric', 'text']

     def list_map_field_types(self):
         return [t['value'] for t in self.map_field_types]
```

If the above solution is right, I will send a PR then.

amercader:
Fix in #2537

u10313335:
Confirmed. Bug fixed. Thank you all!

PR #2411, opened 2015-04-30
Fixes https://github.com/ckan/ckan/issues/2387

mattfullerton:
As proposed in https://github.com/ckan/ckan/issues/2387. It would have been nice to differentiate between the geojson field and the lat/long fields, but it's not possible without big changes elsewhere. I know this is assigned to someone else, but I need it now! ;-)


amercader:
Closing this in favour of #2537, as that fix does not mix types between lat/lon and geojson.

PR #2413, opened 2015-05-02
Fixes #2410 by upgrading leaflet cluster plugin

### LINKED ISSUE #2410: Clustering markers in CKAN 2.3
Specific markers are not shown when zooming in, while when clicking the area they are shown ok. 

After zooming in and out clusters don't come back to theri original position.

Problem:
http://dp-dev.epf.p5.tiktalik.io/dataset/testowa-baza/resource/e5a1e2ef-49e8-463a-bf9b-44070d5085d2

While it work ok in http://demo.ckan.org/dataset/mazowieckie-jst/resource/e06d2626-a5c7-420f-8238-570f0dc21b36

I see that Leaflet was updated from 0.4.4 to 0.7.3. Maybe some code around it was not updated?

markers_ok
markers_problem

mattfullerton:
This has also been done in the central recline repo


brew:
This is good to go, @amercader.

amercader:
Thanks for the patch @mattfullerton

PR #2414, opened 2015-05-03
Clean up unused storage code

nigelbabu:
We don't use the upload bits and bobs in the old storage code. All we have left is the code that will let you still download the old files. I suspect this can be left in for another release so that there is an upgrade path from 2.1 to 2.4.

I'd be strongly in favor of dumping the whole thing after the next release. I need to test this. A bit complicated because I need to use the 2.1 datastore and upgrade to master and confirm that migration works. Bah.


nigelbabu:
More code needs to be removed, but it doesn't have to happen all at once.

davidread:
Forgive me for being clueless, but where do file uploads get stored now, since this part is being removed? Perhaps this should be explained in the changelog?

amercader:
The code being removed in this PR is the old OFS based used up until 2.1. There are more details [here](http://docs.ckan.org/en/latest/maintaining/filestore.html#filestore-21-to-22-migration).

The "new" code (since 2.2) that handles uploads is at [`lib/uploader.py`](https://github.com/ckan/ckan/blob/master/ckan/lib/uploader.py). Files still get stored to the server but with a more sane layout and URLs, and with permissions checks.

davidread:
Ah good, thanks for explaining. I see the migration info is in the changelog for 2.2 too.

PR #2164, opened 2015-05-06
Default configurations allow anyone to "pretend they hacked the site"

amercader:
I just noticed that someone used the default configurations to pretend they hacked a site by just creating a new dataset which does not look good for those who installed CKAN without configuring the authentication.

I propose we change the default configurations to require users to be in organisations to create datasets (and disallow unowned datasets).

That is change the ``ckan.auth.create_unowned_dataset`` and ``ckan.auth.create_dataset_if_not_in_organization`` defaults to ``False`` instead of ``True``.

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -62,10 +62,10 @@ ckan.site_url =
 ## Authorization Settings
 
 ckan.auth.anon_create_dataset = false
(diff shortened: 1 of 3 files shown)


amercader:
@davidread we got bitten by this again so I went and update the settings and docs.
Just unassigning you so it's flagged on the meeting, but if you want to have a look and merge if happy that would be great

PR #2420, opened 2015-05-08
Cast string to unicode to allow resources that include a BOM

### LINKED ISSUE #2401: 500 Error on view_form_filters.html with some utf-8 files with BOM
We get an internal server error when loading the "edit view" page (in our case to delete a view that was created).

Based on the server log, the error is coming from view_form_filters.html, the template/feature used to add predefined filters in the view.

The specific issue seem to be with a special character 

```
 [remote 172.17.42.1:25449] File '/project/ckan/ckan/templates/package/snippets/
 [remote 172.17.42.1:25449]   data-module-template-filter-inputs = "{{ h.
 [remote 172.17.42.1:25449] File '/project/ckan/ckan/lib/helpers.py', line 1806 in 
 [remote 172.17.42.1:25449]   return str(string).replace('\\n', '')
 [remote 172.17.42.1:25449] UnicodeEncodeError: 'ascii' codec can't encode character u'\\ufeff' in position 851: ordinal not in range(128)
```
(Full log available here: http://pastebin.com/uSaTUhjZ)

After playing with the files, it seems that the error appears on some utf-8 encoded files with BOM. I shared 2 files (https://drive.google.com/folderview?id=0B7gAVbY0-eHrflZ6a0dUbVNEVzZKNkJDdFZCN2t0ak9YdWs0UzJHSUR6RUJDOGN3X1lVbm8&usp=sharing), one with BOM that triggers the error, another without BOM that works fine.

brew:
In the template helper `remove_linebreaks `, casting the string to `str` causes an error for characters outside of the ascii range. Notably, BOM at the beginning of documents. Casting to unicode fixes this. Fixes #2401.

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1832,7 +1832,7 @@ def resource_view_full_page(resource_view):
 
 def remove_linebreaks(string):
     '''Remove linebreaks from string to make it usable in JavaScript'''
-    return str(string).replace('\n', '')
+    return unicode(string).replace('\n', '')
 
 
 def list_dict_filter(list_, search_field, output_field, value):
--- a/ckan/tests/lib/test_helpers.py
+++ b/ckan/tests/lib/test_helpers.py
@@ -91,13 +91,13 @@ class TestHelpersRemoveLineBreaks(object):
         assert result.find('\n') == -1,\
             '"remove_linebreaks" should remove line breaks'
 
-    def test_remove_linebreaks_casts_into_str(self):
-        class StringLike(str):
+    def test_remove_linebreaks_casts_into_unicode(self):
+        class UnicodeLike(unicode):
             pass
 
-        test_string = StringLike('foo')
+        test_string = UnicodeLike('foo')
         result = h.remove_linebreaks(test_string)
 
-        strType = ''.__class__
+        strType = u''.__class__
         assert result.__class__ == strType,\
-            '"remove_linebreaks" casts into str()'
+            '"remove_linebreaks" casts into unicode()'


amercader:
This is all very well, but maybe we should ask ourselves why is the BOM there in the first place :)

Importing this [file](https://dl.dropboxusercontent.com/u/3109637/DISTRICTELECTORAUX_with_bom2.csv) via the DataPusher creates a field in Postgres which is actually named `\ufeffENTITEID`, although when inspecting the API output with JSONView or the DB table with psql that might not be apparent:

```
datastore_default=# \d 188a71a4-cc02-481c-9d99-aed95126e440
                             Table "public.188a71a4-cc02-481c-9d99-aed95126e440"
   Column   |   Type   |                                      Modifiers                                       
------------+----------+--------------------------------------------------------------------------------------
 _id        | integer  | not null default nextval('"188a71a4-cc02-481c-9d99-aed95126e440__id_seq"'::regclass)
 _full_text | tsvector | 
 ﻿ENTITEID   | text     | 
 MUNID      | numeric  | 
...

datastore_default=# select "ENTITEID" from "188a71a4-cc02-481c-9d99-aed95126e440";
ERROR:  column "ENTITEID" does not exist
LINE 1: select "ENTITEID" from "188a71a4-cc02-481c-9d99-aed95126e440...

```

This is where headers are [guessed](https://github.com/ckan/datapusher/blob/master/datapusher/jobs.py#L344) so it might be that we need to go down to the messytables level. [This comment](https://github.com/okfn/messytables/blob/master/messytables/commas.py#L26) might relevant.

brew:
@amercader: Yes - it looks like that code in messytables will remove the BOM from UTF-16{LE/BE} files, but not from UTF-8 files that contain a BOM.

wardi:
a BOM in a UTF-8 file.. uh, what?

brew:
I'm submitted a PR for messytables to [remove the UTF8 BOM](https://github.com/okfn/messytables/pull/130).

brew:
Awaiting messytables merge, then will update requirements.txt with new messy tables version.

amercader:
I guess this means we need to do a release in messytables.

The messytables requirement is from DataPusher not CKAN, and I think it will be ok because it's listed as >= xx

amercader:
Also this is still useful for preventing failures on existing instances so I'll merge it as well.

amercader:
New messytables version 0.15.0 includes a fix to prevent the BOM from ending up in the field title.

Issue #2421, opened 2015-05-08
Package schema erroneously passed to create_view


pduchesne:
During the call to 'package_create_default_resource_views' in https://github.com/ckan/ckan/blob/master/ckan/logic/action/create.py#l211-212, the context is passed including the validation schema of the package.
That schema then ends up in resource_view_create (https://github.com/ckan/ckan/blob/master/ckan/logic/action/create.py#l356) and causes validation of the view data to fail.

Error encountered during a spatial harvest.

pduchesne:
here's a related PR : https://github.com/ckan/ckan/pull/2422

PR #2422, opened 2015-05-08
[#2421] pop package schema from context before creating views

pduchesne:


amercader:
Thanks for the patch @pduchesne

Issue #2426, opened 2015-05-11
IMapper methods apparently not being called.


knirirr:
I've mentioned an issue both on the mailing list, and also in a StackOverflow post here:

http://stackoverflow.com/questions/30163162/ckan-imapper-plugin-appears-to-do-nothing

Basically, IMapper's methods such as before_insert() don't seem to be called despite being used as the documentation specifies. 
No-one seems to know the answer to this, so I wonder if there is a bug in CKAN or the documentation is missing something. I'm using version 2.3b.

davidread:
@knirirr There is a test for this situation which does pass and before_insert() of IMapper is firing.
```
nosetests --ckan --with-pylons=test-core.ini ckan/tests/legacy/test_plugins.py:TestPlugins.test_mapper_plugin_fired
```
I can't see anything wrong with your plugin code, but you don't show how you are triggering the insert. I suggest you compare your code with the test and bisect to find the problem.

knirirr:
I've been using the web interface to add/delete records.
Thanks for mentioning the test - I'll investigate as soon as I have a moment.

knirirr:
OK, so the test passes.

On my installation before_insert and after_insert work for datasets but before_delete and after_delete do not. None of these work with groups (which are my main concern at the moment). As mentioned, this is using the web interface.

Adding datasets via the API triggers before_insert and after_insert, but there's nothing for groups.

davidread:
Great to acknowledge that before_insert is working after all.

It looks like before_delete and after_delete trigger when you 'purge' an object (which is almost unheard of in ckan outside tests), rather than 'delete' (state='deleted' - the normal way). You could work-around by listening for a change event and checking the state. Or you could fix CKAN - that would be appreciated.

As for groups, please do dig in and find out what's going on.

In the meantime, I'll leave this ticket open as I will add a note in the docs about this.

davidread:
Docs are now correct. Cheers @knirirr

PR #2431, opened 2015-05-15
[#2426] Acknowledge that before_delete only fires on PURGE not DELETE

davidread:
Discussion in issue: #2426 

Adjusted docs and added test.


davidread:
@amercader 2 months waiting for review - perhaps we can merge this now?

PR #2436, opened 2015-05-19
[#2435] Raise ValidationError on package_update if no id or name

### LINKED ISSUE #2435: Server error if you don't provide an id or name on package_update
Should only raise a Validation Error

amercader:
Instead of just crashing

Fixes #2435


wardi:
right, package_show is the one that actually takes a 'name_or_id' parameter, but not a 'name'? I'd be super happy if we were using the same conventions in all the actions.

kmbn:

PR #2443, opened 2015-05-22
Fix broken boolean validator

deniszgonjanin:
The boolean validator treats missing values as True. They should be treated as False, both semantically and because unchecked  HTML checkboxes in forms transmit empty values on form submits.

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -103,11 +103,12 @@ def is_positive_integer(value, context):
 def boolean_validator(value, context):
     '''
     Return a boolean for value.
-    Return value when value is a python bool type.  
+    Return value when value is a python bool type.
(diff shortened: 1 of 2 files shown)


wardi:
We're adding docstrings to these validators as we work on them something like https://github.com/ckan/ckan/blob/master/ckan/logic/validators.py#L61-L69

Also please update the tests to cover this new case. e.g. https://github.com/ckan/ckan/blob/master/ckan/tests/logic/test_validators.py#L427

deniszgonjanin:
@wardi added tests and docstrings

wardi:
@deniszgonjanin just a pep8 fix and this is good to go

deniszgonjanin:
@wardi is good now

PR #2450, opened 2015-05-26
 #2145 Improve datapusher error messages

smotornyuk:
Add label near the resourche title on the dataset detail list
Add decorator to send messages to user in case of api call's crash

Notifying is turned on by setting directive <notify.crash_mail_sent_to_guilty> in config file to true
decorator can take named argument <exception> which defines type(or types if tuple used) of exceptions which will be provided with email to user who make something that raise error
Example:
        from ckan.logic import in_case_of_crash_notify_user_from_context
        @in_case_of_crash_notify_user_from_context(exception=EXCEPTION_CLASS)
        def create_table(context, data_dict):
            ...
Datastore functions 'alter_table' and 'create_table' wrapped by new decorator

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -115,6 +115,29 @@ def get_reset_link_body(user):
         }
     return reset_link_message.format(**d)
 
+def get_crash_body(user, error, action, data):
+    crash_message = _(
+    "Something wrong happened on {site_title} during execution of API call to {action}.\n"
+    "\n"
+    "Error message:\n"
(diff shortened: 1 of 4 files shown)


maxious:
error_label
Here's a screenshot of the UI label

brew:
Could these be separated out into two different Pull Requests; one for the notification decorator, and one for the label?

mattfullerton:
@amercader Is this a no-go until separated? It would reduce the pain barrier for new setups a lot.

PR #2452, opened 2015-05-27
Store license list in database - ckan/ideas-and-roadmap#130

iaroslav13:
Currently licenses are hardcoded, CKAN sysadmin cannot administer License types using web interface, also there is no API action that allow manipulating with licenses.

After this change an ability to store licenses in database is added. Also the API actions are added. It is still possible to use licenses from external JSON file if 'licenses_group_url' is specified in conf file.

API actions:
- license_item: returns license type by ID
- license_create: allows to add licenses to database
- license_update: allows to edit licenses
- license_delete: allows to delete licenses from database. The record will be deleted if license is not used in any dataset or marked as deleted if it is used.
- license_reinstate: allows to reinstate deleted licenses

Web interface allows sysadmin to:
- view a list of licenses
- add new licenses to DB
- edit licenses
- delete licenses
- reinstate licenses

Also migration script has been added. This script will add to database all hardcoded licenses that connected to datasets as well as all custom licenses attached to datasets.


iaroslav13:
Tests fails are related to "other-open" license that is not available by default now. Only "cc-by" and "odc-by" are created by default if no other licenses are used in Datasets.

I am not sure if I can alter "ckan.tests.legacy.lib.test_dictization" tests.

iaroslav13:
I added other-open license as test data as it is used in test Datasets.

davidread:
I like this, and would like to merge it with bits of #2485. I propose:

* store licences in the database - as you have it, using state=deleted for deprecated licenses rather than 'exclude' config option in my code. Call option 'include_deleted' in licenses_list action (rather than 'all').
* rather than have both license.url and license.home_url, just store the CC (or whatever) url in license.url.
* use the logic layer - as you have it, and get rid of the logic in my helper function.
* default licenses are the open CC ones (cc0, cc-by, cc-by-sa), initialized from JSON file (from my code) into the database. The ODC licenses seem not to have taken off so I see no reason to have them as defaults.
* migration puts into the db the default licenses PLUS any legacy ones that are used by datasets.

Do let me know how this strikes you @iaroslav13

iaroslav13:
@davidread thanks for your feedback.

I like the points you propose. Let's go through all of them one by one:

# store licences in the database - as you have it, using state=deleted for deprecated licenses rather than 'exclude' config option in my code. Call option 'include_deleted' in licenses_list action (rather than 'all').

davidread:
All good.

>> get rid of the logic in my helper function.

> Do you mean "license_options" helper function?

Yes, exactly

> OK, I will change default licenses.

That's great that you're happy with these as default. Would you be happy for these to be stored in JSON file, rather than code? This seems better for people to copy from if they have their own custom license list in JSON. I can merge this change in from my PR.

> > migration puts into the db the default licenses PLUS any legacy ones that are used by datasets.

> this is how it works now.

Ah, got you now - perfect.

iaroslav13:
That's great that you're happy with these as default. Would you be happy for these to be stored in JSON file, rather than code? This seems better for people to copy from if they have their own custom license list in JSON. I can merge this change in from my PR.
>> I am not quite sure about it because I made so that it is possible to store licenses either in DB or in external files (in this case they are not exported to DB and not editable via web interface or API). But I am happy to move the license list that was stored  in code to a separate JSON file.

kmbn:

Issue #2451, opened 2015-05-27
In query.py the connection to Solr must only created once to aviod "Cannot assign requested address" in certain Linux distros


qlands:
Since SolrConnection() is persistent by default it is not advisable to create a new connection to Solr each time a package is queried (query.py line 258). This could result in Error 99 "Cannot assign requested address" in  some Linux distros particularly by calling the API package_create several times.

PR #2453, opened 2015-05-28
Use only one Solr connection

qlands:
[#2451]

This fixes  Error 99 "Cannot assign requested address" in some Linux distros particularly by calling the API package_create several times. The fix creates only one connection (_solr_connection) in the pylons global environment and uses it both in index and query files.

Tested against master and Release 2.3 with over a 1000 package_create calls

### DIFF
--- a/ckan/lib/app_globals.py
+++ b/ckan/lib/app_globals.py
@@ -172,6 +172,7 @@ class _Globals(object):
         self._init()
         self._config_update = None
         self._mutex = Lock()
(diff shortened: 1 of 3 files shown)


wardi:
@qlands The CI failures suggest we need one connection per thread. Would you take a look at that?

qlands:
ok. Let me have a look.

wardi:
Closing due to lack of activity. If you'd like to finish this work off please reopen.

PR #2455, opened 2015-06-03
Implement additional page for resource so that user will be able to play with api

smotornyuk:
Added page with url /dataset/{id}/resource/{resource_id}/api-sandbox/ which also available through new button on resource view page "API Sandbox", that appears on page after admin enables 'Developer Mode' at config page.

On this page user can chose one of predefined actions. After that, this action's help will be shown and form with fields for this action's parameters will be generated
User can fill few(or all) fields and submit form. In that case below the form will be displayed server's response and few examples of request (using CURL and JavaScript for the moment)
Also user allowed to check option "Sandbox mode". In that case response will be shown, but no changes for actions like resource_create, resource_delete, etc. will be commited

This ability implemented by adding param 'sandbox'. Such actions check presence of this parameter and don't make Session.commit() or repo.commit() in that case

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -276,6 +276,9 @@ def make_map():
                   action='resource_view')
         m.connect('/dataset/{id}/resource/{resource_id}/view/',
                   action='resource_view')
+        m.connect('resource_api_sandbox',
+                  '/dataset/{id}/resource/{resource_id}/api-sandbox/',
+                  action='resource_api_sandbox')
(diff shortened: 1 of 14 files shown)


wardi:
Have you seen https://github.com/ckan/ckanext-apihelper ?

I feel that this sort of change isn't suited to core because many people will want to customize which APIs and examples are shown. See if you can contribute your work to that extension or if not, start your own.

PR #2457, opened 2015-06-04
Fix getting available organizations for a particular user

amercader:
As described in [this note](https://github.com/ckan/ckan/blob/078cb28f4b2433bd5a82b8edac90b5f3517225a1/ckan/logic/action/get.py#L627), the default permission that is checked when calling `organization_list_for_user` is `edit_group`. This permission is not listed on the [roles definition](https://github.com/ckan/ckan/blob/078cb28f4b2433bd5a82b8edac90b5f3517225a1/ckan/authz.py#L182) in `authz.py` (where `manage_group` is used).

This effectively means that only organizations where the user is Admin are returned (because of how [`get_roles_with_permission`](https://github.com/ckan/ckan/blob/078cb28f4b2433bd5a82b8edac90b5f3517225a1/ckan/authz.py#L222) does the check).

I think this is a bug and very unintuitive. One of the consequences is that on the Organizations tab for the user's dashboard, only orgs where the user is an admin appear. Considering a user which is admin, editor and member on 'Admin org', 'Editor org' and 'Member org' respectively, this is what he or she will see:

snclfsr

Which is not very helpful.

My proposed changes are:

1. Change the default permission checked on `organization_list_for_user` to `manage_group` so all organizations that the user belongs to are returned, regardless of role

2. Add the user role on each particular organization to the output of  the `organization_list_for_user` action, otherwise is impossible to tell which role has the user without a call to `organization_show` for each of them (we are already doing something similar on other parts of the code using `context['with_capacity']=True`. Eg output will be:

    ```
    {
        "name": "org1",
        "title": "Org 1",
        "capacity": "admin",
        ...
    }
    ```

3. Display the role that the user has on each organization on the Dashboard.

After all these changes, the dashboard looks like this:

fskkpuf

I propose to backport 1 and 2 as this is something that would be nice to fix and the changes are backward compatible.


davidread:
Suggest adding tests

amercader:
Return all orgs, not just admin

davidread:
Mention in the changelog the change of default for this api

davidread:
@amercader says he still needs to add tests

davidread:
I don't understand why you picked 'read' for the default permission checked. That suggests that you return all the groups which the user can 'read', which includes all public groups, not just the ones it is admin/editor/member of.

I can see from the tests it does work, but only because I guess the auth logic for reading a group is hard-wired, rather than a logged-out person having the 'read' permission.

I think it would make more sense to change the default to 'manage_group'.

amercader:
```python
ROLE_PERMISSIONS = OrderedDict([
    ('admin', ['admin']),
    ('editor', ['read', 'delete_dataset', 'create_dataset', 'update_dataset', 'manage_group']),
    ('member', ['read', 'manage_group']),
])
```

In any case, given how the permissions work using `manage_group` does not make any difference as we are calling `roles = authz.get_roles_with_permission(permission)` and in both cases `member` will be included. But I agree that `read` is ambiguous so let's use `manage_group` (which should be really named `read_private_datasets` :) )

davidread:
Great! Interesting about it only being for reading private datasets - I'd have thought it was for changing the name and description of the group or adding members, but maybe they are just implicit in the 'admin' permission.

ckan/lib/helpers.py
```diff
@@ -1652,7 +1652,7 @@ def groups_available(am_member=False):
 
 
 @core_helper
-def organizations_available(permission='edit_group'):
+def organizations_available(permission='read'):
```
davidread: was this going to change too?

ckan/logic/action/get.py
```diff
@@ -615,17 +615,28 @@ def organization_list_for_user(context, data_dict):
     datasets in. This takes account of when permissions cascade down an
     organization hierarchy.
 
+    :param id: the name or id of the user to get the organization list for
+        (optional, defaults to the currently authrized user (logged in or via
```
davidread: typo authrized

amercader:
@davidread done

PR #2460, opened 2015-06-06
Fix dataset-purge Paster command

drmalex07:
The `dataset purge` command leaves dangling references to purged datasets in "member"
table (so this fix takes care to remove those membership records before).

I came across this issue when trying to purge an organization (whose datasets were supposed to be purged before) and received a 5xx server error. This was happening because the process of purging an organization includes purging of membership records (but of course, the actual member datasets were now absent).

I have only tested this against 2.2.2, but from a quick look i think this also applies to 2.2.3.

### DIFF
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -930,7 +930,7 @@ class DatasetCmd(CkanCommand):
         # Purge membership to several groups
         q_member = model.Session.query(model.Member).filter(and_(
             model.Member.table_name == 'package',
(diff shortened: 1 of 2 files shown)


wardi:
whoops, this was for release-v2.2.2 will reset

wardi:
@drmalex07 closing this PR, please resubmit against master and we can backport it.

drmalex07:
@wardi , ok i will try it on master. 
As long as the membership model is not changed, i do not expect any significant differences.

PR #2461, opened 2015-06-10
add ITranslations interface

### LINKED ISSUE #959: Translations for extensions need resolving
How do we do translations for extensions?

we could do some nasty merge of po/mo files using polib but that would need write permissions so could be problematic on a server.

joetsoi:
This allows extensions to merge in their own translations with the main
ckan strings at runtime. Each extension should have an 'i18n' directory in ckanext/myplugin/ of their repo containing its translations. The gettext domain
(and hence filename) of the .mo files should be ckanext-{extname}

see (http://babel.pocoo.org/docs/support/#extended-translations-class)

fixes #959


ckan/lib/plugins.py
```diff
@@ -524,3 +524,31 @@ def activity_template(self):
...
+
+class DefaultTranslation(object):
+    def directory(self):
+        '''Change the directory of the *.mo translation files
+
+        The default implementation assumes the plugin is
+        ckanext/myplugin/plugin.py and the translations are stored in
+        i18n/
+        '''
+        import os
```
joetsoi: currently looking at the wrong directory here, we want the extension's plugin.py

ckan/lib/plugins.py
```diff
@@ -524,3 +524,31 @@ def activity_template(self):
...
+        return os.path.abspath(os.path.join(os.path.dirname(__file__),
```
joetsoi: currently looking at the wrong directory here, we want the extension's plugin.py

wardi: don't know if this is useful here https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/plugins.py#L301-L315

That's how I go from a python import path like ckanext.myplugin to a file system path

wardi:
Super excited about this! :surfer:

joetsoi:
:( 

I'm writing an example_tranlation plugin and I've run into a minor problem. Either I'm going to have to write custom plugin implementations (not testing the DefaultTranslation) or we're going to have to change the default location of the i18n directory. The current default is in the root directory of the repo (eg. https://github.com/ckan/ckanext-harvest/tree/master/i18n)

but ckanext in the main repo https://github.com/ckan/ckan/tree/master/ckanext would put the i18n directory somewhere in the main tree, which I really don't want.

I could create the plugin nested a bit further
```
/ckanext/example_translation/ckanext/example_translation
/ckanext/example_translation/i18n
```
but this is different from the layout of all the other example plugins and might confuse users.

ckan/lib/plugins.py
joetsoi: `return os.path.join(os.path.dirname(inspect.getfile(self.__class__)), '../../', 'i18n')`

should give me what I want, as self will still refer to the plugin class

wardi: hmm. seems magical, and requires that users put their plugin in a specific directory. Also, won't that break if an extension has been "pip install"ed and not "setup.py develop"ed?

joetsoi: yeah it unfortunately does, but this is just the default implementation that most people will have if they ran a `paster --plugin=ckan create`, but to the people who move their plugin to a different location they will just have to tolerate overriding `directory()` with this implementation.

wardi: i18n is going to have to be under ckanext/myplugin/i18n and included in the manifest to work with installed packages

joetsoi: fair enough, I've never really messed with the manifest stuff, but it sounds like it'll have to go in that location, it'll fix my test problem as well.

ckan/lib/plugins.py
```diff
@@ -524,3 +526,38 @@ def activity_template(self):
...
+class DefaultTranslation(object):
+    def directory(self):
+        '''Change the directory of the *.mo translation files
+
+        The default implementation assumes the plugin is
+        ckanext/myplugin/plugin.py and the translations are stored in
+        i18n/
+        '''
+        # assume plugin is called ckanext.<myplugin>.<...>.PluginClass
+        extension_module_name = '.'.join(self.__module__.split('.')[0:2])
```
joetsoi: @wardi, is this less offensive? I take the module name 'ckanext.whatever.blah.plugin' and assume the first two are 'ckanext.pluginname'. This feels a bit fragile.

ckan/lib/i18n.py
```diff
@@ -131,6 +138,27 @@ def handle_request(request, tmpl_context):
...
+            except AttributeError:
+                # this occurs when an extension has 'en' translations that
+                # replace the default strings. As set_lang has not been run,
+                # pylons.translation is the NullTranslation, so we have to
+                # replace the StackedObjectProxy ourselves manually.
+                environ = pylons.request.environ
+                environ['pylons.pylons'].translator = translator
+                if 'paste.registry' in environ:
+                    environ['paste.registry'].replace(pylons.translator,
+                                                      translator)
```
joetsoi: the other option to this would be to remove all the `if lang != 'en'` statements, this would probably also mean maintaining an empty i18n/en/LC_MESSAGES/ckan.mo/po file

joetsoi:
hum, should add a changelog entry under a new v2.5 for this as well.

joetsoi:
with each extension inside its own domain, it's probably worth investigating at some point whether we can use gettext.binddomain to restrict translations, so extensions can specify that they don't want specific strings overwritten by other translations.

metaodi:
@joetsoi how does that work with the domains? Do you have to specify from which domain you want to load a translation in a template? Or would a plugin have to override the `i18n_domain()` method and return `ckan` to replace translations from the CKAN core?

joetsoi:
by default you will not have to specify a domain at at all and the strings will be overriden, I want investigate if we can prevent extensions overwriting each others strings when they happen to clash.

ckan/lib/plugins.py
wardi: I'm ok with this.

wardi:
@joetsoi do you have a minute to resolve the conflicts?

joetsoi:
@wardi, just some of the docs conflicting, all done now

joetsoi:
Just as a note, if a message is overriden say from one plugin to another you'd have to run `gettext.dgettext(domain, message)` https://docs.python.org/2/library/gettext.html#gettext.dgettext to get the strings from specific extensions

so an extension would probably have to use dgettext instead of _ if it wanted specific strings, I don't think jinja has tags for dgettext which is unfortunate.

wardi:
I'm sure we could import dgettext into a place that makes it possible to `{{ dgettext(...) }}` in templates

Issue #1527, opened 2014-02-19
Invite to organization has hard-coded email text


rossjones:
Would be useful if the email text in get_invite_body() (from mailer.py) loaded a template with the text so that extensions can override it.  When working around https://github.com/ckan/ckan/issues/1506 it is likely that the email template will need to be changed, and/or other users might want to change the text of the email.

hbunke:
What's the status of this issue? The code has been in master for some time now (https://github.com/ckan/ckan/commit/66549991119191fb00a7aaf907e4027036df4561), but nothing seems to have made it into a release so far.

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #2465, opened 2015-06-15
[#1527] Allow to customize emails sent from CKAN

amercader:
The reset password and user invite emails sent from CKAN have their subjects and bodies hardcoded into `ckan/lib/mailer.py`. This PR includes changes to make the subjects configurable on the ini file, and the bodies via overriding templates from extensions.
It also adds a bit more detail to the user invite email, including which organization/group you've been actually added to, and with which role.

Both subjects and bodies are translated.

It also fixes a couple of bugs on the `user_invite` action and updates and refactors all mailing legacy tests,


wardi:
How about a couple 1-line templates for the subjects instead:
* ckan/templates/emails/invite_user_subject.txt
* ckan/templates/emails/reset_password_subject.txt

amercader:
@wardi you got it: 422df75

wardi:
@amercader thanks! :beers:

amercader:
One thing to note (outside the scope of this PR) is that the message extractors that we rely on for txt files and Javascript files are based on Genshi:

https://github.com/ckan/ckan/blob/60a7edf78ebc4d980c8caa25d5dbd6a5b7c294b6/setup.py#L183:L191

Would be good to get rid of these and use Jinja2 so we can drop the Genshi requirement

wardi:
I can't think of a reason to keep genshi for text and js templates, We already removed the legacy templates, so extensions that depend on them must be really broken already.

amercader:
Sure, as long as we extract strings on txt and js files using jinja2 (which I didn't manage with a 2 min try) let's get rid of it for good.

On other news, I hate CKAN legacy tests...

amercader:
@wardi All tests should be green now in case you want to squeeze this on 2.5

wardi:
Dang, didn't get to this one in time

PR #2469, opened 2015-06-16
Also notify observers of packages involved in PackageRelationship

filipefigcorreia:
When relations exist between packages (via PackageRelationship) the related packages are not being included in the notification of what is changing. A consequence is that when viewing the details of a package, the related packages are not shown.
 
This pull request fixes this issue by including the related packages in the list of changed objects.


wardi:
@filipefigcorreia sorry for the slow reply. we're not taking PRs against release branches, please resubmit against master.

Issue #2470, opened 2015-06-18
Update requirements.txt for requests and six to avoid version conflict with ckan-galleries


florianm:
When installing [ckan-galleries](https://github.com/DataShades/ckan-galleries), ckan's requirements of python packages requests and six collide with flickrapi's requirements (via ckan-galleries), also reported [here](https://github.com/DataShades/ckan-galleries/issues/2).

Would anything speak against increasing the versions to the following? I haven't got a sane test environment atm so will have to send a PR to let Travis do the testing.
```
requests==2.7.0
six==1.9.0
```

davidread:
requests upgrades need coordination, as several common extensions use it and need to be upgraded in step. So someone need to go through them carefully and make any necessary changes to include support for the later requests version.

florianm:
What would be the best procedure for this, @davidread ? 

I'll see which of the 133 extensions from http://extensions.ckan.org/ depend on requests and/or six, but I'll probably need help to test them all against requests 2.7.0 and six 1.9.0. 

I can't see any breaking changes in python-request's [release notes](http://docs.python-requests.org/en/latest/community/updates/#software-updates), all the listed behavioural changes look more like additional features and bug fixes but I could be wrong.
There seem to be no release notes for six so I've got no clear idea of possible breaking changes.

joetsoi:
I wouldn't go through and test all the extensions, as david said we should probably test the more common extensions.

wardi:
Also, the version in ckan's requirements.txt doesn't prevent you from requiring a newer version in one of your extensions. That's the reason we pin versions in requirements.txt and not in install_requires

davidread:
I imagine six is purely for ckan - if it passes the tests then that's probably enough.

ckanext-archiver was my particular concern, although it's only used by uk, us and some old OKF project afaict. But if it is still at 1.1 then it's not going to be much worse at 2.7.

I'd have thought doing some basic checks with ckan including datapusher, ckanext-harvest, ckanext-spatial and ckanapi is probably sufficient.

This should probably miss the 2.4 branching next week to let it bed in.

florianm:
@joetsoi I'll go through the list of extensions to at least identify which ones depend on requests and six so we can choose which ones to test
@wardi I had the opposite experience - ckan-galleries's newer versions of requests and six collided with ckan's older versions (in my local datacats install)
[...]

PR #2471, opened 2015-06-18
updated requirements for requests and six

florianm:
Addressing #2470 

This accommodates for extensions using the flickrapi, which requires higher versions of requests and six than ckan master provides.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,9 +31,9 @@ pyutilib.component.core==4.5.3
 repoze.lru==0.6
 repoze.who==2.0
 repoze.who-friendlyform==1.0.8
-requests==2.3.0
+requests==2.7.0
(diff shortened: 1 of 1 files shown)


joetsoi:
closing this for now, also we'd want to make the changes to requirements.in as  well as they are the direct requirements.

florianm:
No problem. @joetsoi how can I get this merged eventually?

joetsoi:
I think we should push for it after we branch for 2.4 which should be some time in the next week or so, i think requests 2.3 -> 2.7 should be reasonable and it can push us to take a look at the other requirements. But as said in the other issue #2470 if we can just do some testing on the main extensions (harvesting spatial which ever other regularly installed ones use requests) then I think we'd be satisfied to update it.

PR #2477, opened 2015-06-18
Remove old useless scripts in /bin

davidread:


wardi:
big +1 from me

wardi:
This would let us remove the ckanclient dev dependency too. The only other thing I see using it is a couple tests of questionable value in ckan/tests/legacy/wsgi_ckanclient.py.

davidread:
Sounds good, I'll do it.

wardi:
@joetsoi can I steal the assignment from you? I'm itching to merge this

wardi:
love it

PR #2481, opened 2015-06-21
update to newest

roadlit:


roadlit:
rtest

joetsoi:
These look like changes that need to be either made in your own extensions, or if you're just doing translations, you should do them on transifex.

PR #2415, opened 2015-06-22
Allow uppercase letters in local part of email when sending user invitations

amercader:
* usernames must be lower case
* user invitations generate a random username from local part of the submitted email address and some random number.
* we shouldn't be assuming  that the localpart of an email is always lower case. https://en.wikipedia.org/wiki/Email_address

just needs a  .lower() in [`_get_random_username_from_email`](https://github.com/ckan/ckan/blob/9ab53fd540869570275547c9aa2741d013a8ae97/ckan/logic/action/create.py#L1093) and a relevant test.


amercader:
@joetsoi I took this off your hands as I was looking at the same code. Can you merge if you are happy?

Issue #2482, opened 2015-06-22
Package schema erroneously passed to resource_view_create when updating


metaodi:
The same issue as described in #2421, happens when updating a package here:
https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L364-366

I guess a similar fix should do the trick:
- Move this https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L375-376
- up some lines before calling `package_create_default_resource_views`

I didn't look at this in depth, but it failed for me in a custom harvester in CKAN 2.3, so a backport to 2.3 just like #2421 would be nice.

If I find the time, I can help with the PR:

wardi:
/me mutters something about https://github.com/ckan/ideas-and-roadmap/issues/53

metaodi:
Okay, here is the PR: #2484

rossjones:
This appears to have been fixed by https://github.com/ckan/ckan/pull/2503 
If this is not the case, please re-open this ticket.

PR #2484, opened 2015-06-22
[#2482] Remove schema before creating views

metaodi:


wardi:
This change doesn't make the current package_update code any worse, but it still relies on modifying the context dict passed. @amercader is there any reason we cant' only includes the keys actually needed by the package_create_default_resource_views and package_show? e.g.

```python
if data.get('resources'):
    logic.get_action('package_create_default_resource_views')(
        {'model': context.model, 'user': context.user},
        {'package': data})
```

amercader:
@wardi I've created https://github.com/ckan/ckan/pull/2503 with your suggested fix. Closing this one

PR #2429, opened 2015-05-22
Remote CKAN config

amercader:
See https://github.com/ckan/ideas-and-roadmap/pull/145 for context (now a bit outdated, will update it next week)

Tl;DR of this PR:

```
http POST localhost:5000/api/action/config_option_update Authorization:XXX ckan.homepage_style=2 ckan.site_title="My site title" ckan.site_description="Nice portal" ckanext.example_iconfigurer.test_conf="test_value" ckan.datasets_per_page=5

{
    "help": "http://localhost:5000/api/3/action/help_show?name=config_option_update", 
    "result": {
        "ckan.datasets_per_page": 5, 
        "ckan.homepage_style": 2, 
        "ckan.site_description": "Nice portal", 
        "ckan.site_title": "My site title", 
        "ckanext.example_iconfigurer.test_conf": "test_value"
    }, 
    "success": true
}

```

And also:

zraunid

A bit more detail:

* Refactored the code that handles storing settings in the db (and on the `app_globals` object) to be a bit more sane, to use actions and check which config options can be updated using a schema (which is also used to validate them)
* Extensions can modify this schema to add (or remove) config options to it, making them available to update ([example](https://github.com/ckan/ckan/pull/2429/files#diff-8d8bfa5ed17aef6869bd49c39aad1e3e))
* The `config_option_update` action can be used to edit any of the allowed options in bulk passing config key and values, as in the example above.
* `config_option_list`will return the available options, and `config_option_show` the current value for one of them
* Extensions can also extend the admin config form to allow editing the configuration options from the frontend, as in the screenshot above.

PR #2499, opened 2015-06-25
[#2498] Fixes the "None" seen in the corner

davidread:
Issue: #2498

### DIFF
--- a/ckan/tests/lib/test_app_globals.py
+++ b/ckan/tests/lib/test_app_globals.py
@@ -1,0 +1,17 @@
+from ckan.lib.app_globals import app_globals as g
+
+
+class TestGlobals(object):
+    def test_config_not_set(self):
+        # ckan.site_about has not been configured.
+        # Behaviour has always been to return an empty string.
+        assert g.site_about == ''
+
+    def test_config_set_to_blank(self):
+        # ckan.site_description is configured but with no value.
+        # Behaviour has always been to return an empty string.
+        assert g.site_description == ''
+
+    def test_set_from_ini(self):
+        # ckan.template_head_end is configured in test-core.ini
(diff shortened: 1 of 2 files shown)


davidread:
@amercader perhaps you could review this since it relates to your PR #2429 ?

ckan/tests/lib/test_app_globals.py
```diff
@@ -0,0 +1,17 @@
+from ckan.lib.app_globals import app_globals as g
+
+
+class TestGlobals(object):
+    def test_config_not_set(self):
+        # ckan.site_about has not been configured.
+        # Behaviour has always been to return an empty string.
+        assert g.site_about == ''
+
+    def test_config_set_to_blank(self):
```
amercader: Maybe we should use the change_config decorator to avoid relying on actual configuration values on `test-core.ini`?

```
@helpers.change_config('ckan.site_description', '')
def test_config_set_to_blank(self):
    ...
```

ckan/tests/lib/test_app_globals.py
```diff
@@ -0,0 +1,17 @@
...
+        # ckan.site_description is configured but with no value.
+        # Behaviour has always been to return an empty string.
+        assert g.site_description == ''
+
+    def test_set_from_ini(self):
```
amercader: Same here:

Maybe we should use the change_config decorator to avoid relying on actual configuration values on `test-core.ini`?

```
@helpers.change_config('ckan.template_head_end', '<link href="example.com"'')
def test_config_set_in_config(self):
    ...
```

ckan/tests/lib/test_app_globals.py
davidread: change_config doesn't rerun the app_globals stuff unfortunately, so g is unaffected by it.

Issue #2493, opened 2015-06-23
Arabic translation format string issue


davidread:
There is a problem with the arabic translation:

```
(ckan)vagrant@precise64:/vagrant/src/ckan$ tx pull -f -r ckan.2-3 -l ar
(ckan)vagrant@precise64:/vagrant/src/ckan$ paster check-po-files ckan/i18n/*/LC_MESSAGES/ckan.po
Checking file ckan/i18n/ar/LC_MESSAGES/ckan.po
    Format specifiers don't match:
    Please <a href="{link}">update your profile</a> and add your email address and your full name. {site} uses your email address if you need to reset your password. ->   <"{a href="{link>    <a/>         . {site}
```

In Transifex I can toggle the "RTL" (right-to-left) button and it goes right/wrong, but I'm not sure if it is correct or wrong. I don't support you could take a look @amercader ?

davidread:
Adria says we can fix this by changing the way we specify the params in the string.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #2500, opened 2015-06-25
[#2493] Stats extension clean-up

davidread:
Issue: #2493 

* removed templates_legacy - no-one uses these. Removed associated
  style.css.
* leaderboard goes with it as there is no jinja equiv and no interest.
  Removed associated controller, demo.html and app.js.


davidread:
These are the last two Genshi HTML files going!

davidread:
Recent merge conflicts fixed. Ready to eliminate our last genshi templates!

PR #2502, opened 2015-06-26
Remove open-id related nag messages. 

davidread:
Moved home controller tests from legacy.


davidread:
Ping @joetsoi ;-) This is a great time in the lifecycle post-release to merge things!

Issue #2494, opened 2015-06-24
Timezone setting


davidread:
Add a config setting for the timezone for the site. So times that appear correctly in the Last Updated / Created fields, Activity Stream and extensions such as ckanext-harvest's harvest errors. Until now they all appear in UTC, as it is in the database, I believe.

Perhaps it should default to the timezone of the server? Although that's probably not helpful for cloud servers.

wardi:
JS will save us all. let the user's browser decide how to display times

brew:
Something like: http://momentjs.com/timezone/

davidread:
If not JS, then perhaps by default we can have "UTC". And maybe we display the timezone after the time?

metaodi:
I'll look into that. As far as I know, there is no other way but JavaScript to determine a users timezone. But apart from that, there should be a server side setting to either explicitly set the timezone or use the default timezone of the server.

metaodi:
PR for this issue: #2505

metaodi:
@davidread  #2505 got merged, so I think we can close this issue.

PR #2505, opened 2015-06-26
[#2494] New option to change the timezone of displayed datetimes

metaodi:
This PR introduces a new configuration option called `ckan.timezone`. This option accepts all valid timezone strings, that are accepted by [pytz](http://pytz.sourceforge.net/).

The dates are kept as UTC internally, and there is no need to change that. Using pytz, the given date is converted from a "naive" to an "aware" object (see the [datetime docs for details](https://docs.python.org/2/library/datetime.html)). After this conversion, the date has a `tzinfo` attribute. This attribute can be read and changed.

Feedback would be much appreciated.


doc/maintaining/configuration.rst
```diff
@@ -1556,6 +1556,21 @@ Default value: (none)
 
 By default, the locales are searched for in the ``ckan/i18n`` directory. Use this option if you want to use another folder.
 
+.. _ckan.timezone:
+
+ckan.timzone
```
davidread: typo "timzone"

requirements.txt
```diff
@@ -41,3 +41,4 @@ unicodecsv==0.9.4
 vdm==0.13
 wsgiref==0.1.2
 zope.interface==4.1.1
+pytz==pytz==2012j
```
davidread: Repeat looks like a mistake

davidread:
Needs a test.

I guess the main question is whether this is the best approach, to set a time-zone for a site, or is it better to use JS to detect the user's timezone? I see that gmail uses my computer's timezone. That seems pretty reliable, and makes more sense for a user in a different timezone than the site's configuration. What do you think?

metaodi:
@davidread I'll look into the browser-based approach. Maybe it would be good to have both, i.e. to be able to set the `ckan.timezone` option to something like `browser` which would use the users setting. This could even be the default.

metaodi:
@davidread if you have time, maybe you can have another look at my approach.

I added a new API endpoint to save the timezone offset from JavaScript in the beaker session. Like that I can use then this value to render the datetimes correctly and print the timezone name as something like "UTC+3". I don't know if this is the way to go, any feedback would be appreciated.

wardi:
tying time zone to the site or even the user's session seems complicated to me. Why not always render UTC but replace it with local time using JS?

metaodi:
@wardi this would lead to a "flickering" page, where content gets replaced after it is rendered. So I would definitely do this on the server side before it gets rendered.

davidread:
I'm not very keen on using a session which would mean sending a cookie to the 99% of users who aren't logged-in.

I think the archiver is the only place where the time makes most sense displayed absolutely, whereas the activity stream and 'last updated/created' fields can be relative (e.g. "2 weeks ago"). The great thing about displaying relative time, is that you don't get the flicker, since there is time for the JS to fix the timezone before the user gets a chance to hover over it to see the absolute date. I see that's what the Twitter feed does. I'd be interested to see what big sites do that display an absolute time, without you being logged in, but can't think of any examples.

metaodi:
@davidread and what is bad about that? If they don't want cookies, the times will stay as they are. Or why are cookies inherently bad? Please note, that `render_datetime` is also used in extensions (like ckanext-harvest), so this is not something we can only solve for the CKAN frontend alone.

davidread:
Yes, cookies are problematic for lots of reasons, such as preventing caching, EU cookie laws and generally being suspect of invasion of privacy.

I think my preferred solution right now is using the config setting for a server timezone and use JS to adjust to the user's timezone, without cookies. That would have flicker but only on the harvest page when viewed by admins, or we could just change those times to be relative until you hover.

But again, I'd like to see what what other sites do about showing times for non-logged in users, as there may be another way we've not thought of here.

metaodi:
@davidread the standard solution to this issue is to simply ask the user for his timezone. So for logged in users, we could just add an extra field to the user model to save the timezone. This way, at least logged in users would see the "their" time, all the others would see whatever the server setting is.

davidread:
Suggestion - on this one page (harvest log) where we want to display absolute time, can we have in-line JS that fixes it? That would get rid of the flicker. What do you think?

And for relative time we'd use post-load JS to fix the absolute time shown on hover. This could be done at a later date.

Ian an I agree that the session saving is just a pain and not something we'd want to maintain in CKAN. And we're not hugely keen on the site having a configured time-zone for all users - big countries like US, Canada, Russia, China have multiple timezones per country and some sites (datahub.io) have little emphasis on being in a particular place. I feel bad that I've been encouraging this and now you've put the effort in to write all this code - so sorry about that.

metaodi:
@davidread I'll look into this, how "flickerish" it'll be if we do it with JS only, maybe it's not that bad after all.

About the countries with multiple timezones: yes sure it exists, but that's not a good reason to ignore them. Such a site could still decide to leave everything in UTC (the current default), but the server setting would give all those small country and city level portals the possibility to at least set "their" timezone. I suggest to keep this part, but I'm fine with removing the session part. The server setting is something additional, with no privacy concerns, and nothing changes, unless an admin actively sets a timezone. Do you agree?

And don't worry about "my" code/effort. It's all about finding a good solution together. I like to code and argue about it. It's all good from my side :+1: It's part of making open source software :smiley:

davidread:
Sounds good. I think your addition basic setting is only a few code lines ( https://github.com/metaodi/ckan/commit/88218c2944b34a8d0ef84001ad10a92106cc9076 ), plus the pytz dependency (which btw should be in requirements.in) and is a reasonable basic level. However it might make the JS more complicated. @wardi are you happy to leave just this bit in?

To summarize:

Config option "ckan.timezone" can be:
* particular timezone e.g. 'Australia/Sydney'
* 'server' (i.e. 'same TZ as the server')
* 'UTC' - which is the default if not specified (for backwards compatibility)

How about 'server' is set for new CKAN installs? i.e. set it in the config file template.

And you add:
* in-line JS in harvest page to fix render_datetime() output (no flicker) (although this would have to work 

You or someone else later could add:
* "32 minutes ago" widget (localised_nice_date) has the hover value (absolute time) fixed in normal ckan.js (flicker if the user manages to hovers before it gets changed).

metaodi:
@davidread @wardi I finally got the time to work more on this PR. I kept the `ckan.timezone` config option to influence the rendering of the datetimes on the server already. By using the the "special" value `server`, the local timezone of the server is used to render the dates.

In order to get the JavaScript part to work, I added a `<span>` with a data attribute around the rendered dates, containing the same date in the ISO-format. This makes it very easy to parse with [moment.js](http://momentjs.com/). I added the multi-language version of moment.js, in order to support as many users as possible. I then used the [localized date format](http://momentjs.com/docs/#/displaying/format/) to display the dates to the user.

To prevent the "flickering", all the dates are hidden when the page is loaded (using the `.js` CSS class). As soon as the dates are loaded, the are displayed. Of course, it's still not the same, as when it is rendered on the server-side, but it's very much acceptable. Btw: if JavaScript is disabled, the original server-rendered-value is displayed.

As always: any feedback is very welcome!

PS: I already created a PR for ckanext-harvest, to add the `<span>`s there as well: https://github.com/ckan/ckanext-harvest/pull/142

ckan/templates/package/snippets/additional_info.html
```diff
@@ -60,13 +60,22 @@
         {% if pkg_dict.metadata_modified %}
           <tr>
             <th scope="row" class="dataset-label">{{ _("Last Updated") }}</th>
-            <td class="dataset-details">{{ h.render_datetime(pkg_dict.metadata_modified, with_hours=True) }}</td>
+            <td class="dataset-details">
+                <span class="datetime" data-datetime="{{ h.render_datetime(pkg_dict.metadata_modified, date_format='%Y-%m-%dT%H:%M:%S%z') }}">
```
wardi: data-datetime gets set to a non-UTC time when the timezone config is set, meaning we have pytz doing UTC -> tz_server and then moment.js doing tz_server -> tz_local. That *should* work but it feels fragile to me. How about putting UTC in data-datetime instead?

ckan/templates/package/snippets/additional_info.html
```diff
@@ -60,13 +60,22 @@
...
         {% endif %}
         {% if pkg_dict.metadata_created %}
           <tr>
             <th scope="row" class="dataset-label">{{ _("Created") }}</th>
-            <td class="dataset-details">{{ h.render_datetime(pkg_dict.metadata_created, with_hours=True) }}</td>
+
+            <td class="dataset-details">
+                <span class="datetime" data-datetime="{{ h.render_datetime(pkg_dict.metadata_created, date_format='%Y-%m-%dT%H:%M:%S%z') }}">
+                    {{ h.render_datetime(pkg_dict.metadata_created, with_hours=True) }}
+                </span>
```
wardi: this span is repeated but looks useful for all sorts of dates. How about creating a "templates/snippets/local_friendly_datetime.html" (or similarl) that can be reused?

ckan/public/base/css/main.css
```diff
@@ -5560,6 +5560,9 @@ a.tag:hover {
 .js .tab-content.active {
   display: block;
 }
+.js .datetime {
```
wardi: this class has an innocuous name that's also hard-to-grep-for, but it really should be shouting about how it does some spiffy JS magic and be easy to find in the code. Consider ".automatic-local-datetime" or similar

doc/maintaining/configuration.rst
```diff
@@ -1556,6 +1556,21 @@ Default value: (none)
 
 By default, the locales are searched for in the ``ckan/i18n`` directory. Use this option if you want to use another folder.
 
+.. _ckan.timezone:
+
+ckan.timezone
```
wardi: another naming comment :smile: how about "ckan.display_timezone" or similar to make clear how this value is used.

ckan/templates/package/snippets/additional_info.html
metaodi: This should always work, but I yes, it seems a _bit_ complicated. If I should change that, I guess I would have to introduce a new argument for `render_datetime` called `use_utc=True` or similar.

ckan/public/base/css/main.css
metaodi: Good input, I changed that.

ckan/templates/package/snippets/additional_info.html
metaodi: That makes sense, I added such a snippet.

doc/maintaining/configuration.rst
metaodi: @wardi well, [naming is considered to be hard](http://martinfowler.com/bliki/TwoHardThings.html) :smiley:  Thanks for this suggestion as well, I changed my code accordingly.

metaodi:
@wardi @davidread from my POV, I consider this PR done and mergable.

Chealer:
Thank you for this change.
Just a small remark, but I find the resulting localised_nice_date() a bit inconsistent. If it displays the timezone when displaying hours, it seems to me it should also display it when not displaying hours.

Issue #1749, opened 2014-06-09
Remove instances of h.get_action


nigelbabu:
Now that `h.get_action()` is deprecated, we need to remove instances of it in our example code and especially from the documentation.

Quick grep shows at least files need modification:
```
ckan/templates/package/snippets/info.html
ckanext/example_theme/v19_02_error_handling/templates/ajax_snippets/example_theme_popover.html
ckanext/example_theme/v21_custom_jquery_plugin/templates/ajax_snippets/example_theme_popover.html
ckanext/example_theme/v20_pubsub/templates/ajax_snippets/example_theme_popover.html
ckanext/example_theme/v18_snippet_api/templates/ajax_snippets/example_theme_popover.html
ckanext/example_theme/v19_01_error/templates/ajax_snippets/example_theme_popover.html
```

wardi:
fixed by #2508

PR #2508, opened 2015-06-29
[#1749] Remove the last usage of "h.get_action"

egrajeda:
This removes the last usage of `h.get_action()` I could find in the code. #2384 previously removed another one.

I believe #1749 can be closed if this PR is accepted and merged.


wardi:
@egrajeda thanks!

PR #2510, opened 2015-06-30
An IUploader interface for extensions

brew:
A way to override the default `lib.uploader.ResourceUploader` class from extensions, enabling other file storage solutions, e.g. integration with cloud providers.


brew:
Example use: https://github.com/okfn/ckanext-s3filestore/blob/0870c5ecc46438cf98e56c3e55ec715ed9dc326f/ckanext/s3filestore/plugin.py

PR #2512, opened 2015-06-30
delete organizations causes 500 errror

joetsoi:
current on master and release-v2.4.0

didn't update the delete controller for groups when group_types were changed in https://github.com/ckan/ckan/pull/2354 . self.group_type is now self.group_types and is a list of the group types. this fixes the check in the controller.

split from the organization tests so there is less to review


davidread:
Great work finding this @joetsoi 
Merged to master and cherry-picked to release-v2.4.0.

PR #2517, opened 2015-07-03
show toc on front of docs, use rtd redirects

wardi:


wardi:
proposed front page here: http://ckan.readthedocs.org/en/2517-rtd-toc-redirects/

This PR also removes the api.rst document to allow rtd redirects to work.

PR #2519, opened 2015-07-03
update resource last_modified on file uploads

### LINKED ISSUE #1874: Modifying resource doesn't update it's last updated timestamp

wardi:
This fixes #1874 and #2147


tobes:
This looks good to me and is a better fix than the one i made I think it should be merged and backported

wardi:
@tobes thanks! We can't backport this because it changes the behaviour of last_modified. Maybe we can just fix the bug in 2.3 with something like your #2148

PR #2521, opened 2015-07-06
Add limit to revision search

davidread:
It's a legacy API, but it should have a limit anyway.


davidread:
2 line fix with lots of tests improved from legacy. Merge?

PR #2526, opened 2015-07-07
Remove noisy log statement

davidread:
Creates reams of logging on the first request - not needed unless working on this specific code.

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -81,7 +81,7 @@ class AuthFunctions:
                             resolved_auth_function_plugins[name]
                         )
                     )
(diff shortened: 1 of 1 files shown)


wardi:
this is a log.debug, isn't it only noisy if you're logging at the debug level? Even then you can choose to disable logging from ckan.authz in your logging configuration.

davidread:
Yes I find debug logging helpful for both development and in production. We should keep the useful stuff and delete the dross like this. I'm frequently debugging authz issues so don't want to lose the other debug in this module.

wardi:
ok, then remove the line completely and I'll merge it. I'm not a fan of commented-out code

davidread:
I agree! Done now.

PR #2529, opened 2015-07-09
Add mimetypes to resource_formats.json

davidread:
* Correct the MS Word mimetype.
* Add alternative mimetypes for recognizing Excel files. Being used in a DGU harvester.


davidread:
That's the merge conflict sorted now, hopefully this only needs a cursory glance before merging.

PR #2537, opened 2015-07-16
[#2387] Fix GeoJSON field option in map view

amercader:
Because the datastore fields were limited to numeric ones (to
accommodate the lat/lon option), it was not possible to select a text
field for the geojson option.

This patch separates the types used on both, while keeping the
validation in place and not mixing types between options.

### DIFF
--- a/ckanext/reclineview/plugin.py
+++ b/ckanext/reclineview/plugin.py
@@ -178,7 +178,9 @@ class ReclineMapView(ReclineViewBase):
 
     datastore_fields = []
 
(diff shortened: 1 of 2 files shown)


amercader:
2.4 if possible

amercader:
The issue can be tested with [this CSV](https://dl.dropboxusercontent.com/u/3109637/Bus_Stops_reduced_geojson.csv), trying to create a Recline map view with the GeoJSON option. The GEOM field can not be selected.

PR #2543, opened 2015-07-20
user logout test

brew:
Front-end test of log out redirect

### DIFF
--- a/ckan/tests/controllers/test_user.py
+++ b/ckan/tests/controllers/test_user.py
@@ -123,6 +123,19 @@ class TestLoginView(helpers.FunctionalTestBase):
                                    .format(user['fullname']))
 
 
+class TestLogout(helpers.FunctionalTestBase):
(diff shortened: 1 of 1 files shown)


ckan/tests/controllers/test_user.py
```diff
@@ -123,6 +123,19 @@ def test_registered_user_login_bad_password(self):
                                    .format(user['fullname']))
 
 
+class TestLogout(helpers.FunctionalTestBase):
+
+    def test_user_logout(self):
```
davidread: You're not testing the actually logging out (cos you are not logged in) just the redirect associated, so can you set the test title to reflect that? Otherwise this is great.

joetsoi:
would be nice to somehow check the current user in the request but this works fine i think

davidread:
Perfect, cheers @joetsoi and @brew

PR #2545, opened 2015-07-21
resource controller tests

joetsoi:


joetsoi:
hum for whatever reason github is looking at the wrong commit :/ the changes in 0671f97  to the package controller are needed to catch the NotAuthorized exception and not 500

amercader:
Woo, tests! :tada:

PR #2549, opened 2015-07-22
Handle special characters in column names to fix recline view

### LINKED ISSUE #2490: ReclineView: Bug: Column name with special chars (star, braces, etc.)
On https://danepubliczne.gov.pl/dataset/baza_teleadresowa_administracji_zespolonej/resource/f051b7ab-2780-441c-a39b-ce02d7b8d8fa

in console you can see error: `bootstrap.js:3 Uncaught Error: Syntax error, unrecognized expression: GEOGRAFICZNA*)` which is caused by column `SZEROKOŚĆ GEOGRAFICZNA*)` while invoking: 

https://github.com/ckan/ckan/blob/0f145a54629287a6f7764b8fd73963bda62260ef/ckanext/reclineview/theme/public/vendor/recline/recline.js#L3225

    function sanitizeFieldName(name) {
      var sanitized = $(name).text();
      return (name !== sanitized && sanitized !== '') ? sanitized : name;
    }

BTW: `$(name)` never seems to return anything.

Problem arises in CKAN 2.3 (reclineview).
This work correctly in CKAN 2.2.1 (reclinepreview) as you can see in demo: http://demo.ckan.org/dataset/lubuskie/resource/42f569fc-6da1-4eea-9ab7-1b1f43a836b2

maxious:
Some special characters/unicode characters are causing recline view
sanitization for HTML to fail with a JS error:
```
"bootstrap.js:3 Uncaught Error: Syntax error, unrecognized expression:"
```

Fixes #2490 by catching JS exceptions

### DIFF
--- a/ckanext/reclineview/theme/public/vendor/recline/recline.js
+++ b/ckanext/reclineview/theme/public/vendor/recline/recline.js
@@ -3222,7 +3222,12 @@ my.SlickGrid = Backbone.View.extend({
     }
 
     function sanitizeFieldName(name) {
-      var sanitized = $(name).text();
+      var sanitized;
+      try{
+        sanitized = $(name).text();
+      } catch(e) {
+        sanitized = '';
+      }
       return (name !== sanitized && sanitized !== '') ? sanitized : name;
     }
 


amercader:
Thanks for the patch @maxious

Issue #2551, opened 2015-07-22
Widget.recordCount.js doesn't handle singularity. Wrong label "records"


benjaminlaot:
It doesn't handle the result's singularity.

Ex : 1200 records  -> OK
        1records -> KO

![capture](https://cloud.githubusercontent.com/assets/9968236/8822869/bcc6023c-3068-11e5-9a06-475d5525f845.PNG)

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #2552, opened 2015-07-22
2551 Allow a good orthograph for the word 'record' when there is only one result.

benjaminlaot:
Fix issue #2551 

It would be great if 'record' and 'Unknown number of' were translatable in widget.recordCount.js .


rossjones:
@amercader apart from a small spacing issue, this one seems reasonably straight-forward. Wasn't translatable before, so not sure anything is lost from this.

kmbn:
@amercader Any objection to this PR?

PR #2553, opened 2015-07-23
Wrong format autodetected for TSVs prevents DataPusher from being called

amercader:
When you upload or link to a TSV file (eg http://example.com/test.tsv) and don't specify a resource format, the format that you end up with is `text/tab-separated-values`. This is not recognized by the DataPusher and the resource is not uploaded to the DataStore.

This value is returned by `mimetypes.guess_type()` on the [`if_empty_guess_fomat`](https://github.com/ckan/ckan/blob/master/ckan/logic/validators.py#L770) validator. The [`clean_format`](https://github.com/ckan/ckan/blob/master/ckan/logic/validators.py#L781) validator should transform this to `TSV` but it doesn't because the miemtype is missing in the alternative representations for this format in [resource_formats.json](https://github.com/ckan/ckan/blob/master/ckan/config/resource_formats.json#L21).

This patch adds this alternative representation.

### DIFF
--- a/ckan/config/resource_formats.json
+++ b/ckan/config/resource_formats.json
@@ -18,7 +18,7 @@
   ["MDB", "Access Database", "application/x-msaccess", []],
   ["NetCDF", "NetCDF File", "application/netcdf", []],
   ["ArcGIS Map Service", "ArcGIS Map Service", "ArcGIS Map Service", ["arcgis map service"]],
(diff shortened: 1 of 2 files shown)


davidread:
It would be marginally better to swap 'text/tab-separated-values' and 'text/tsv' here, because the official one is 'text/tab-separated-values'.

amercader:
@davidread done

PR #2561, opened 2015-07-28
Upgrade libraries

davidread:
Just by running pip-compile of the top-level requirements.in this upgrades some of the libraries. Tests don't pass yet.

It reorders the libraries, to see the changes more easily, see this commit: https://github.com/ckan/ckan/commit/0f1c8582d38089da5bddce93028bd5249f170e51#diff-b4ef698db8ca845e5845c4618278f29aL1


davidread:
@wardi LazyJSONObject doesn't appear to work with the new version of simplejson (3.8.0 instead of 3.3.1) :-(. I tracked it down to this bit of code that sees that the JSONString isn't actually an int, so does str() and returns -1.

    def _encode_int(value):
        skip_quoting = (
            _int_as_string_bitcount is None
            or
            _int_as_string_bitcount < 1
        )
        if type(value) not in integer_types:
            # See #118, do not trust custom str/repr
            value = int(value)

I've added a test to this PR that shows the problem: ckan/tests/lib/test_lazyjson.py
I don't suppose you could take a look?

davidread:
@wardi I notice that even with older simplejson, LazyJSONObject needs to be dumped with the for_json param, or it returns just ```'{}'```. So since it can't be dumped without knowing it is a special case, then why don't we change this code to instead work out if the object has been converted into a dict at all, and if not just use the string stored inside it. So all the object needs to do is make a note if it was used as a dict, and we can get rid of the JSONString bit.

davidread:
I'm thinking this, as I can't see a way for JSONString to inherit from any types now, since type() is used to check it, whether it is a string, float, None, True/False etc.

wardi:
@davidread yes, the current LazyJSONObject needs a better repr().

But if there's no way of making it work with new simplejson, do we drop the optimization or keep the older dependency?

When writing this hack I should have talked to the simplejson people to get this sort of hook implemented nicely on their side. I'll try to take a look this week.

davidread:
Good call, I'll keep the old version on this PR to try and get it merged, and split off this issue: https://github.com/ckan/ckan/issues/2681

dev-requirements.txt
```diff
@@ -11,3 +11,4 @@ factory-boy==2.1.1
 coveralls==0.4.1
 sphinx-rtd-theme==0.1.6
 beautifulsoup4==4.3.2
+pip-tools==1.1.2
```
davidread: which includes `pip-compile`, used to generate requirements.txt

Issue #2564, opened 2015-07-28
datastore_search should use a 'read-only' postgres user


davidread:
Currently the datastore's search uses the 'write_url' postgres user, rather than the 'read_url' so that its search can include private datasets. However that gives it more power than is needed - it is the only one to use a write user of all the actions available withing the appropriate authentication. It seems safer to use:
- for searching public datasets use the 'read-only' user
- for searching private datasets use a new postgres user that can only read, but can access private datasets too

davidread:
So we don't need a third user once #2562 is done

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

Issue #2555, opened 2015-07-23
Minimum Postgresql version increased to 9.2


davidread:
Following proposal here: https://github.com/ckan/ideas-and-roadmap/issues/117
we think Postgres 9.2 is a good compromise, bringing in basic JSON column type, and it is supported by recent linux distributions.

- [ ] update source install instructions
- [ ] update package install instructions
- [ ] Travis - remove 8.4 test and make it test on 9.2 and 9.4?
- [ ] remove postgres 8.x code in datastore

metaodi:
Is this change planned for CKAN 2.5? We currently have some installations of CentOS where PostgreSQL 9.2 is not available because of internal restrictions.

I agree with changing it, just wanted to know if there is a known timeframe for it, so we can plan accordingly.

davidread:
Yes, planned for CKAN 2.5 in 3 months. We're aware that upgrading will
ruffle some feathers with bigger orgs, but 9.2 is almost 3 years old so
reasonably conservative.

Dave

davidread:
Update: I've had all positive feedback so far, so I'm still planning to do this for CKAN 2.5

davidread:
This change depends on us moving to Ubuntu 14.04 (which comes with Postgres 9.3, whereas Ubuntu 12.04 comes with 9.1). @amercader are you still gunning to get that (https://github.com/ckan/ckan/issues/1651) in for the CKAN 2.5 branch in early October?

And I guess we should have some upgrade instructions to cover upgrading Postgres, that covers Ubuntu 12.04 & 14.04, source & package installs. I've added that to the checkbox list above.

davidread:
Update: I did much of the docs to show how to upgrade one's postgres database from 8.4:

https://github.com/ckan/ckan/compare/postgres-version

however I didn't get the postgis_restore.pl script to work. It's not something that is a priority for my work, so unassigning myself.

NicolaiMogensen:
I get this error when trying to execute this command `sudo -u postgres psql --cluster 9.5/main -f backup_ckan.sql`

error: 
`The input is a PostgreSQL custom-format dump.
Use the pg_restore command-line client to restore this dump to a database.`

PR #2570, opened 2015-08-06
[WIP] Auth fixes for datastore_search and datastore_search_sql 

rossjones:
This PR removes the use of the write_url in the datastore_search action, as the permissions are already checked later on against the auth.datastore_search function (which delegates to resource_show).

Looks like the non-removal of write_url in the action method was an accidental oversight.

Tests (including the privacy tests for TestDatastoreSearch.test_search_private_dataset) continue to pass.

Should fix tickets #2564 and #2562

### DIFF
--- a/ckanext/datastore/logic/action.py
+++ b/ckanext/datastore/logic/action.py
@@ -399,7 +399,7 @@ def datastore_search(context, data_dict):
         raise p.toolkit.ValidationError(errors)
 
     res_id = data_dict['resource_id']
-    data_dict['connection_url'] = pylons.config['ckan.datastore.write_url']
+    data_dict['connection_url'] = pylons.config['ckan.datastore.read_url']
 
     resources_sql = sqlalchemy.text(u'''SELECT alias_of FROM "_table_metadata"
                                         WHERE name = :id''')


rossjones:
Test failure was for Python 2.6 but doesn't seem to make any sense :(

rossjones:
Does this PR also need removal/modification of datastore_make_private ?

amercader:
We definitely need to take the datastore_make_private stuff into account if we are to merge this.

What essentially happens now is that datastore_search (and _insert, _upsert, etc) will use the "write" connection. Whenever a dataset is made public or private, the DataStore plugin will actually GRANT or REVOKE permissions on the actual table to the "read" user. This was to avoid the table being accessible on `datastore_search_sql`.

Forcing `datastore_search` to use the "read" user shouldn't conflict with this, if the permissions are properly set up. So your changes here add another level of authorization higher up at the logic layer. The question is whether we want to keep the database level check as a safety net. Something to discuss on the dev meeting.

The Travis issue are caused by 2 issues

* On Postgresql 8.4 the DataStore runs on [legacy_mode](http://docs.ckan.org/en/latest/maintaining/datastore.html#legacy-mode) and it does not use `ckan.datastore.read_url`, so all tests involving `datastore_search` are failing:

```
 File "/home/travis/build/ckan/ckan/ckanext/datastore/logic/action.py", line 402, in datastore_search
    data_dict['connection_url'] = pylons.config['ckan.datastore.read_url']
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/paste/registry.py", line 146, in __getitem__
    return self._current_obj()[key]

KeyError: 'ckan.datastore.read_url'
```
As agreed in #2555 we are moving away from 8.4 on CKAN 2.5, so perhaps we can skip these tests until we remove all the 8.4 stuff properly.

* And of course there are also PEP8 failures :) :

```
-----------------------
ckanext/datastore/db.py
-----------------------
E501 ln:1227 line too long (82 > 79 characters)
E501 ln:1228 line too long (83 > 79 characters)

-------------------------------
ckanext/datastore/logic/auth.py
-------------------------------
E302 ln:50 expected 2 blank lines, found 1
```

BTW, please prepend your branches and commits with the issue number, otherwise the CHANGELOG will miss them.

rossjones:
This is for two separate issues, but it's pretty much the same code.  I might split into two PRs.

wardi:
We need a command to reset the permissions on the private datastore tables so the read user can access them. Also suggested to include a dummy migration that triggers this command to make it easy for users to not forget it

davidread:
Users need to migrate the datastore database, to ensure people keep rights to search private datasets. So we discussed adding some sort of datastore paster command e.g.:

    paster datastore upgrade

rossjones:
We decided on the call to not integrate with the core migrations. Just a command and a huge warning in next release.

davidread:
What's the remaining WIP?

Any idea about these test failures? e.g.
```
======================================================================

ERROR: ckanext.datastore.tests.test_interface.TestInterfaces.test_datastore_delete_insecure_filter

  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/home/travis/build/ckan/ckan/ckanext/datastore/tests/test_interface.py", line 161, in test_datastore_delete_insecure_filter

    resource_id=resource['id'])

  File "/home/travis/build/ckan/ckan/ckan/tests/helpers.py", line 104, in call_action

    return logic.get_action(action_name)(context=context, data_dict=kwargs)

  File "/home/travis/build/ckan/ckan/ckan/logic/__init__.py", line 416, in wrapped

  File "/home/travis/build/ckan/ckan/ckan/logic/__init__.py", line 532, in wrapper

    return action(context, data_dict)

  File "/home/travis/build/ckan/ckan/ckanext/datastore/logic/action.py", line 402, in datastore_search

    data_dict['connection_url'] = pylons.config['ckan.datastore.read_url']

  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/paste/registry.py", line 146, in __getitem__

    return self._current_obj()[key]

KeyError: 'ckan.datastore.read_url'
```

rossjones:
Missing config options?

amercader:
As mentioned above on postgres 8.x the DataStore does not use `ckan.datastore.read_url`, so this PR is incompatible with the DataStore legacy mode. We agreed to get rid of it (#2555) so we either wait for that or we skip the tests on legacy mode (but then we really commit to #2555 before 2.5)

rossjones:
Can this be discussed at the tech-team-call today so that we know one way or the other?

davidread:
Code to skip tests for legacy mode is here:
https://github.com/ckan/ckan/pull/2572#issuecomment-132544087

amercader:
I've fixed the failure on the view plugin.
We still need the migration command for previously private tables.

Should it be a matter of running the following?

```
GRANT SELECT ON ALL TABLES IN SCHEMA public TO "datastore_readonly_user";
```

PR #2571, opened 2015-08-06
Introduces IDataPusher plugin interface

rossjones:
This new plugin allows users to reject resources for submission,
and also to be notified once a datapush is complete.

The two methods in the interface are:

can_upload: which will abort datapusher_submit calls if they return True.
after_upload: Called with the context, resource dict and package dict after the upload into the datastore is complete.

The check for can_upload is performed in the datapusher_submit action as doing it anywhere else (such as the plugin that catches resource changes) would mean that the plugins would be ignored when it is triggered via an API call.

This should provide enough functionality to close https://github.com/ckan/ideas-and-roadmap/issues/155

### DIFF
--- a/ckanext/datapusher/logic/action.py
+++ b/ckanext/datapusher/logic/action.py
@@ -10,6 +10,7 @@ import ckan.lib.navl.dictization_functions
 import ckan.logic as logic
 import ckan.plugins as p
 import ckanext.datapusher.logic.schema as dpschema
+import ckanext.datapusher.interfaces as interfaces
 
 log = logging.getLogger(__name__)
 _get_or_bust = logic.get_or_bust
@@ -52,6 +53,14 @@ def datapusher_submit(context, data_dict):
     callback_url = site_url.rstrip('/') + '/api/3/action/datapusher_hook'
 
     user = p.toolkit.get_action('user_show')(context, {'id': context['user']})
+
+    for plugin in p.PluginImplementations(interfaces.IDataPusher):
+        upload = plugin.can_upload(res_id)
+        if not upload:
+            msg = "Plugin {0} rejected resource {1}"\
+                .format(plugin.__class__.__name__, res_id)
+            log.info(msg)
+            return False
 
     task = {
         'entity_id': res_id,
@@ -165,6 +174,9 @@ def datapusher_hook(context, data_dict):
 
         dataset_dict = p.toolkit.get_action('package_show')(
             context, {'id': resource_dict['package_id']})
+
+        for plugin in p.PluginImplementations(interfaces.IDataPusher):
+            plugin.after_upload(context, resource_dict, dataset_dict)
 
         logic.get_action('resource_create_default_resource_views')(
             context,
--- a/ckanext/datapusher/tests/test_interfaces.py
+++ b/ckanext/datapusher/tests/test_interfaces.py
@@ -1,0 +1,151 @@
+import json
+import httpretty
+import nose
+import sys
+import datetime
+
+from nose.tools import raises
+from pylons import config
+import sqlalchemy.orm as orm
+import paste.fixture
+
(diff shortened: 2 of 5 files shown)


mattfullerton:
Are there plans to prevent previewing until the datapush is complete? A confusing issue for users is that garbled previews of data being directly previewed in javascript pop up after the upload and the proper preview out of the datastore, sanitized with messytables, comes some time later after a refresh.

rossjones:
The new plugin certainly allows that, as one of the callbacks is to notify you when the upload is complete.   - so you could flag resources by default as non-previewable-yet or something - sounds like it is worth opening an issue.

ckanext/datapusher/logic/action.py
```diff
@@ -158,6 +168,9 @@ def datapusher_hook(context, data_dict):
     task['state'] = status
     task['last_updated'] = str(datetime.datetime.now())
     if status == 'complete':
+        for plugin in p.PluginImplementations(interfaces.IDataPusher):
```
amercader: I would move the hook a few lines below and pass the `resource _dict` and `dataset_dict` as params to the plugins, to save them the calls

ckanext/datapusher/logic/action.py
```diff
@@ -53,6 +54,15 @@ def datapusher_submit(context, data_dict):
 
     user = p.toolkit.get_action('user_show')(context, {'id': context['user']})
 
+    for plugin in p.PluginImplementations(interfaces.IDataPusher):
+        upload = plugin.can_upload(res_id)
+        if not upload:
+            msg = p.toolkit._("Plugin {} rejected resource {}"\
```
amercader: The {} formatters won't work on Python 2.6 (see Travis failure). You need to use {0}.

amercader: And we no longer translate internal messages

ckanext/datapusher/logic/action.py
amercader: Also pass the context as the first param

rossjones: to the plugin?

amercader: Yes, something like `plugin.after_upload(context, resource_dict, dataset_dict)`

ckanext/datapusher/tests/test_interfaces.py
```diff
@@ -0,0 +1,157 @@
...
+    def setup_class(cls):
+        wsgiapp = middleware.make_app(config['global_conf'], **config)
+        cls.app = paste.fixture.TestApp(wsgiapp)
+        if not tests.is_datastore_supported():
+            raise nose.SkipTest("Datastore not supported")
+        p.load('datastore')
+        p.load('datapusher')
+        p.load('test_datapusher_plugin')
+
+        ctd.CreateTestData.create()
```
amercader: CreateTestData iis frowned upon on the new tests. Better to call factories.Sydadmin/User/Dataset ... on the relevant tests

rossjones:
Have fixed up according to comments so far. Just waiting for travis to be happy.

amercader:
Great stuff @rossjones, also check the PEP8 failures, I know they are your favourite ones :)

rossjones:
Done. With PEP8. Let's see what Travis thinks ...

amercader:
This all looks good. One thing I think users might find confusing is that before `can_upload` is called there still is a prior check on the [`notify()`](https://github.com/ckan/ckan/blob/cebb92bd7676569f67e84c2e582c624aa905869d/ckanext/datapusher/plugin.py#L107:L109) method, that checks the resource format and the `ckan.datapusher.formats` config value. So people implementing this interface might expect their method to be called and it might not happen because of this format check.
I don't think we can remove the check, but maybe mention it on the interface `can_view` method docstring?

Issue #2578, opened 2015-08-12
markdown_extract only expects strings


metaodi:
When using [ckanext-fluent](https://github.com/open-data/ckanext-fluent) to define a multilingual markdown field, the output currently fails with the following error:

    Error - <type 'exceptions.AttributeError'>: 'dict' object has no attribute 'strip'

I could trace this error back to the [following line of code](https://github.com/ckan/ckan/blob/13b8befcd8a8eed566509179a09f3d31d1a2b60d/ckan/lib/helpers.py#L810):

    if (text is None) or (text.strip() == ''):

PR follows to make sure the site doesn't break because of this.

rossjones:
I haven't checked that overriding helpers is possible, but perhaps it might be better to override ```markdown_extract``` in ckanext-fluent instead?

metaodi:
@rossjones well monkey patching could work yes. I'll check if this is possible. But maybe it would be good to have a "proper" way of doing this.

rossjones:
I don't think you need to monkey-patch, you could use ITemplateHelpers - https://github.com/ckan/ckan/blob/master/ckan/plugins/interfaces.py#L888-L910  but it appears that it does stop you overwriting  helpers at https://github.com/ckan/ckan/blob/9bf61b4b2707ff943d38a4f6035114272638a42b/ckan/config/environment.py#L70 :(

wardi:
You can use an IPackageController before_show method to replace the any core fluent fields with a single-language version. I'm planning to add one of these to ckanext-fluent when I can make it work well enough to handle fluent with core dataset fields like title and notes.

@metaodi I've tried to pitch the idea of core ckan adopting the ckanext-fluent behaviour for dicts in place of text but haven't been able to sell it to the others yet :-) Maybe move this issue to ckanext-fluent for now.

metaodi:
@wardi Okay I'll look into that to replace the dict with a single language version. For now, I'll try to implement that in my own extension, if I can make it work, I'll try to integrate that in fluent as well.

wardi:
closing in favour of #2678

PR #2579, opened 2015-08-12
[#2578] Make sure markdown_extract doesn't break if a dict is passed

metaodi:


metaodi:
@rossjones No the function does not handle this case correctly and no the output won't the perfect either. But I think this is probably something that needs to be handled either in ckanext-fluent or an other custom extensions. This check is only to make sure, that the CKAN site does not break (i.e. throw an internal server error).
It's not perfect, but at least you can still use the site.

In my special case I only need to access this data via API, which works perfectly :wink:

rossjones:
I'll mention during tomorrow's call that it's a "stop ckan crashing because of an accidental external extension".  Might be worth mentioning on the ckanext-fluent github too though?

davidread:
Can you be clearer about the case that this helps please? Surely any dict will cause an exception on the re.sub step straight afterwards?

metaodi:
I will close this for now, as it is clearly not a CKAN core issue, but rather a problem of ckanext-scheming or ckanext-fluent.

PR #2580, opened 2015-08-12
Improve group_list speed

### LINKED ISSUE #2554: group_list performance issues: timeouts on homepage
Problems:

1. `group_list` / `organization_list` are calling `group_show` / `orgsanization_show` for all items even when you are just requesting ids/names (the default behaviour)

2. The widgets on the homepage and their associated helpers are calling `group_list` / `organization_list` when they only need to show 1 of each.

This basically means timeouts on the homepage and pages that call `group_list` or `organization_list` on instances that have many groups / orgs (like demo.ckan.org).

demo.ckan.org was running 2.2.x and worked fine, so something along the way went wrong.

brew:
Only perform group_show when all_fields is present.

This improves loading of the homepage (fixes #2554) when the ckan instance has a lot of groups or organizations.

With an instance that has 200 groups, the average time taken (in seconds) of calling group_list 10 times is:
```
before: 5.106
after:  0.042
```

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -392,13 +392,28 @@ def _group_or_org_list(context, data_dict, is_org=False):
 
     all_fields = data_dict.get('all_fields', None)
     include_extras = all_fields and \
-                     asbool(data_dict.get('include_extras', False))
+        asbool(data_dict.get('include_extras', False))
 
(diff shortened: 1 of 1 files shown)


davidread:
For reference, these changes were merged into PR #2554

PR #2582, opened 2015-08-18
Remove horrible code from home controller

amercader:
It forces a `group_list` call on every home page load which is not actually used any more (and hasn't been for a while)


davidread:
For reference, this change was included in PR #2554

PR #2583, opened 2015-08-18
Update plugin.py

alantygel:
The original "packages" sort option was broken ; changed to package_count


amercader:
Thanks @alantygel

PR #2584, opened 2015-08-18
group_list incorrectly including all fields by default

amercader:
#2214 replaced the `organization/group_list` call to `group_list_dictize` by a `organization/group_show` call for each group. This was bound to be slower as it includes validation, followers, etc but it was made worse by the `include_*` params added in #1840 not being passed to `group_show` and `group_dictize`, effectively running all the extra calls for each group on the CKAN instance. This was part of 2.4.0


davidread:
For reference, the changes in this PR were merged into PR #2554

Issue #2581, opened 2015-08-13
Uploading a resource trips datapusher_hook - sometimes


florianm:
Running a CKAN source install 2.5a current as of today.
My users report that loading exactly the same file on multiple pages has worked on some dataset pages and not on others.

This has happened with CSV, PDF, and TXT. I'm getting the error trace below.
It starts with the datapusher hook running, and ultimately `ckan/lib/lazyjson.py` tripping over a type mismatch.
How is this possible with some datasets and not others?

Has anyone else run into this problem so far?

```
[Thu Aug 13 13:09:15.817497 2015] [:error] [pid 10327] [remote 10.6.20.100:59513] URL: http://internal-data.dpaw.wa.gov.au/api/3/action/datapusher_hook
[Thu Aug 13 13:09:15.817508 2015] [:error] [pid 10327] [remote 10.6.20.100:59513] File '/mnt/ckan/default/lib/python2.7/site-packages/weberror/errormiddleware.py', line 162 in __call__
[Thu Aug 13 13:09:15.817511 2015] [:error] [pid 10327] [remote 10.6.20.100:59513]   app_iter = self.application(environ, sr_checker)
[Thu Aug 13 13:09:15.817514 2015] [:error] [pid 10327] [remote 10.6.20.100:59513] File '/mnt/ckan/default/lib/python2.7/site-packages/webob/dec.py', line 147 in __call__
[Thu Aug 13 13:09:15.817517 2015] [:error] [pid 10327] [remote 10.6.20.100:59513]   resp = self.call_func(req, *args, **self.kwargs)
[Thu Aug 13 13:09:15.817530 2015] [:error] [pid 10327] [remote 10.6.20.100:59513] File '/mnt/ckan/default/lib/python2.7/site-packages/webob/dec.py', line 208 in call_func
[Thu Aug 13 13:09:15.817533 2015] [:error] [pid 10327] [remote 10.6.20.100:59513]   return self.func(req, *args, **kwargs)
[Thu Aug 13 13:09:15.817536 2015] [:error] [pid 10327] [remote 10.6.20.100:59513] File '/mnt/ckan/default/lib/python2.7/site-packages/fanstatic/publisher.py', line 234 in __call__
[Thu Aug 13 13:09:15.817539 2015] [:error] [pid 10327] [remote 10.6.20.100:59513]   return request.get_response(self.app)
[...]

PR #2585, opened 2015-08-18
[#2581] fixes for lazyjson module

wardi:


wardi:
This fix is pretty serious. without it package_show crashes when there are no plugins with type IPackageController or IResourceController enabled.

wardi:
But first I'll add a test to this to make sure it keeps working

wardi:
@amercader working better now, please take a look when you have a minute

PR #2587, opened 2015-08-19
Add tests for ckan version checkers

davidread:


amercader:
That looks good! Do you want to throw some letters in there for good measure? IIRC beta and alpha should also be supported, eg `2.1.3b`, 2.5.0a`

davidread:
Whilst I agree that would be a useful addition, and quick, it's probably worth merging this without further delay and someone can add that next time they're in the area.

PR #2593, opened 2015-08-24
Always use `ckan.site_url` to generate urls of resources

### LINKED ISSUE #2592: Download URL has wrong domain name
Based on the mails form the ckan-dev mailinglist (https://lists.okfn.org/pipermail/ckan-dev/2015-August/009236.html), there seems to be a problem when a file is uploaded manually and if the external address of CKAN is different from the internal one.

In our setup the admin users use an internal-only domain to manage the open data portal. The external address is different and set correctly in the configuration (`ckan.site_url`). But when generating the absolute download URL of a resource, this setting is not considered, thus the download gets an internal URL, which an external user can not download.

@amercader already pointed me in the right direction, as where the URL is generated, the code is here: https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L118.

PR follows.

metaodi:
This PR fixes #2592. This might be worth for other branches as well, I'm happy to create another PR for `master` if needed.

### DIFF
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -143,6 +143,16 @@ def _unified_resource_format(format_):
         format_new = format_.lower()
     return format_new
 
+def get_site_url_and_protocol():
+    site_url = config.get('ckan.site_url', None)
+    if site_url is not None:
+        parsed_url = urlparse.urlparse(site_url)
+        return (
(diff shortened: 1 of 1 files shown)


wardi:
duplicated by #2599 closing this one since it's older

wardi:
oh, I see, this is against a release branch.  We merge to master and cherry pick to stable releases as appropriate.

PR #2595, opened 2015-08-24
Don't use cached values when rendering notes.

TkTech:
Currently, a cached field with the rendered markdown is used in the
package/read template.

This value is set very early in the process (and also results in
rendering markdown when it is not needed).

Since it is set so early, it bypasses all of the hooks designed to allow
these values to be manipulated, such as `before_view`. This breaks
extension on core fields that depend on being able to modify these
fields, like the multi-lingual support extension ckanext-fluent.

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -261,8 +261,6 @@ class DefaultDatasetForm(object):
         return ckan.logic.schema.default_show_package_schema()
 
     def setup_template_variables(self, context, data_dict):
-        from ckan.lib.helpers import render_markdown
-
         authz_fn = logic.get_action('group_list_authz')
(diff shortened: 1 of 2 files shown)


wardi:
At the meeting this morning we decided to just include this change on master (the fix has already been merged to scheming for those of us hacking around with fluent notes fields)

@TkTech would you add the change that removes c.pkg_notes_formatted completely? Fewer magical context variables is our real goal.

TkTech:
@wardi I have removed it completely. It is likely the other value set there (and many times in packages.py, which seems redundant) `c.related_count` can also be removed. Comments indicate the value only exists for the disqus plugin, and I can find no usage in core ckan.

PR #2599, opened 2015-08-25
Use `ckan.site_url` to generate urls of resources

### LINKED ISSUE #2592: Download URL has wrong domain name

metaodi:
Fixes #2592 on master

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -110,6 +110,17 @@ def url(*args, **kw):
     return _add_i18n_to_url(my_url, locale=locale, **kw)
 
 
+def get_site_url_and_protocol():
+    site_url = config.get('ckan.site_url', None)
+    if site_url is not None:
+        parsed_url = urlparse.urlparse(site_url)
+        return (
+            parsed_url.scheme.encode('utf-8'),
+            parsed_url.netloc.encode('utf-8')
+        )
+    return (None, None)
+
+
 def url_for(*args, **kw):
     '''Return the URL for the given controller, action, id, etc.
 
@@ -139,6 +150,8 @@ def url_for(*args, **kw):
             raise Exception('api calls must specify the version! e.g. ver=3')
         # fix ver to include the slash
         kw['ver'] = '/%s' % ver
+    if kw.get('qualified', False):
+        kw['protocol'], kw['host'] = get_site_url_and_protocol()
     my_url = _routes_default_url_for(*args, **kw)
     kw['__ckan_no_root'] = no_root
     return _add_i18n_to_url(my_url, locale=locale, **kw)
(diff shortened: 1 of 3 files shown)


wardi:
@metaodi this is great.

I would prefer if the fix was one level lower though, inside ckan's `url_for` helper. Any time a caller passes qualified=True ckan's helper should pass the protocol and host to the routes url_for function.

metaodi:
@wardi good point, I'll change that :+1:

ckan/lib/helpers.py
```diff
@@ -110,6 +110,17 @@ def url(*args, **kw):
     return _add_i18n_to_url(my_url, locale=locale, **kw)
 
 
+def get_site_url_and_protocol():
```
wardi: This returns (protocol, host). How about renaming this function `get_site_protocol_and_host()`

Also please add a docstring that shows that when for e.g. `site_url = http://example.com/prefix/` this helper returns `('http', 'example.com')`

wardi:
I'll need to test this on a site that has ckan running with a prefix on site_url

metaodi:
@wardi I updated the name and the docstring. About the prefix -> if I'm not wrong, this would not be part of the [site_url](http://docs.ckan.org/en/943-writing-extensions-tutorial/configuration.html#ckan-site-url), but rather the [root_path](http://docs.ckan.org/en/943-writing-extensions-tutorial/configuration.html#ckan-root-path).

wardi:
@metaodi oh, good call. Recent directions on ckan-dev have been saying to use site_url, but root_path does look like the right place for that.

I'll need to test that root_path still works then

metaodi:
@wardi I added 3 tests to make sure `url_for` works as expected.

wardi:
@metaodi would you add one more with a non-default root_path set?

metaodi:
@wardi I added the tests, there seems to something broken, I'll look into that.

metaodi:
@wardi I think I could fix it. The previous implementation of the `root_path` thingy seems broken, there was a misuse of the variables `root` and `root_path`. Please check if it makes sense now. My tests pass, but maybe there are other cases that I don't check.

wardi:
@metaodi very nice! Thanks!

PR #2600, opened 2015-08-25
issue 2597 allow underscores in filenames

mbellotti:


TkTech:
These changes are very minimal and a test update is provided. I'm okay with buttoning this in.

PR #2606, opened 2015-08-31
Removes dataset history template/controller and test

### LINKED ISSUE #2605: Delete history templates + controller
The history page is not now accessible, and hasn't been for at least 2 versions - it seems to have been demoted in favour of activity.  We should remove it.

rossjones:
These are no longer accessible via a link in CKAN, and seem to have been replaced by activity.

Should close #2605


wardi:
I'm concerned that we don't have a replacement for this functionality. Activities don't store the actual history, and have some performance problems.

Fixing these pages would also be really useful for some people. @mattleduc what do you think?

That said, +1 to deleting old code, and this would make it easier to eventually remove vdm for packages.

rossjones:
I just wanted to start a conversation with a possible solution ;)

Aaron-M:
History is a seldom used but important/valuable feature in our use case as a research organisation.  Generally a dataset gets deposited and won't need to be changed, but if we spot an error and do make a change then it is good to be able to see what was done by whom when (and even better to access/roll back to the unchanged version).  Similarly if we don't think we have made a change, but someone points out that something is wrong with the data, we can check that nobody as been changing things...  When we publish a paper we also publish the data to allow others to reuse the data, and if they choose, to attempt to replicate our analysis (validate our findings).  If someone attempting to validate our analysis could not replicate our result, our first check would be if anything had occurred with the data we published.

mattleduc:
I think that this code should be removed, but we've had a huge number of requests to include similar functionality in our extensions.

amercader:
Maybe I'm biased because I'm and old timer but I tend to use the history pages quite often. They occasionally give more useful information than the activity streams ([example](http://demo.ckan.org/revision/diff/afghanistan-election-data?diff_entity=package&oldid=5b7c84d6-ce91-4298-8591-7482e22191c2&diff=2b72bf91-9644-4ce7-af47-d254e6b51387)). Having said that, I agree that the actual functionality is quite broken and heavily dependant on vdm so a replacement would be ideal (and costly :) )

Perhaps we can draft a small spec for requirements and possible implementation to make easier to find someone willing to spend time or resources on it?

davidread:
We should defo discuss it in the tech meeting.

Would it be a useful compromise to just make history an extension, just to get this deleted in core? Then those that care can use it and develop it away from the main platform.

rossjones:
It would be great if we had a better idea of who and how many people actually use it.  Perhaps this might be simpler/easier/less-to-maintain as a paster command for managing history?

davidread:
@amercader says there is a lot of requests for a feature to seeing versions and diffs. No-one willing to pay to improve it.

@wardi says it would be valuable to have it in a good state, and it's better coming from where it is now, rather than starting from scratch.

We agreed it is worth leaving the code in there for now.

Concerns about it at the moment:
* templates are awful - need updating and made friendlier. e.g. diff for custom schemas
* too much code is in the controller, not logic layer. Difficult to understand the vdm/revisions and diff code.
* getting old versions of datasets from the logic layer requires setting context so is not accessible from the API. Is this dataset_show?
* diff function misses fields? (@amercader says yes, I've not seen that though)

davidread:
Sorry for all the work :-/ but we decided to keep this code in there after all.

PR #2607, opened 2015-08-31
option for disabling related items

deniszgonjanin:
Now that we have [ckanext-showcase](https://github.com/ckan/ckanext-showcase), I wanted to have a simple way of disabling (or rather, hiding) the related items feature. This PR hides it when a config variable is set, otherwise it doesn't touch the related items by default.

Related to https://github.com/ckan/ckanext-showcase/pull/16

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1158,6 +1158,8 @@ def related_item_link(related_item_dict):
                   id=related_item_dict['id'])
     return link_to(text, url)
 
+def related_enabled():
+    return asbool(config.get('ckan.related_enabled', True))
 
 def tag_link(tag):
     url = url_for(controller='tag', action='read', id=tag['name'])
@@ -2090,6 +2092,7 @@ __allowed_functions__ = [
     'resource_display_name',
     'resource_link',
     'related_item_link',
+    'related_enabled',
     'tag_link',
     'group_link',
     'dump_json',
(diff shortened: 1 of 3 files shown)


davidread:
Should we cut out related items full stop - does ckanext-showcase replace it well enough for most people?

@rossjones What's DGU's position on this? Do we use it?

rossjones:
@davidread we use the DB and logic layer from related to import apps from the Drupal component - but as there's a migration path from related -> showcase, it probably won't be too painful.

I'm +1 for disabling related items, its inclusion in that form was a mistake. I'm +5 for deleting the code entirely.

davidread:
Great, perhaps you could do a PR to delete related items then?

deniszgonjanin:
+1 for deleting entirely. the new extension does a great job, and it even has a migration path from related items.

rossjones:
I've got a local branch without related items now, but I think being able to disable first, so that people have the opportunity to migrate to showcase might be sensible.Otherwise it'll be hard to have the migration that deletes the tables before users have had a chance to migrate.  

So, what I'm suggesting is:

1. Merge the patch to disable through config.
2. Document the required migration during next release.
3. Then a PR to delete related items.

davidread:
That seems sensible - leave the data for now and do the delete migration in
a subsequent release.

wardi:
see also #2610

davidread:
Tech team thinks:

*    We alert people now that it is deprecated
*    2.5 release the front-end is removed
*    2.6 release we remove the rest

@rossjones would you mind doing a PR for just the front-end to be removed?
@brew have you got the list of sites using related?

davidread:
Closing this, in favour of a PR to remove the related items front-end in ckan 2.5.

PR #2610, opened 2015-09-02
Removes the related item feature

rossjones:
Once users are migrated over to ckanext-showcase, this PR will remove all traces
of the related-items feature from core.

Assuming it passes muster it probably shouldn't be merged until after the next release to ensure users have a chance to migrate.


davidread:
This is great.

Tech team thinks:
* We alert people now that it is deprecated
* 2.5 release the front-end is removed
* 2.6 release we remove the rest

davidread:
@rossjones would you mind doing a PR for just the front-end to be removed?
@brew have you got the list of sites using related?

brew:
@davidread Here's the survey doc: https://github.com/ckan/ideas-and-roadmap/blob/64-showcase-spec/specs/showcase/ckan-instances-related-items-survey.md

davidread:
We intend to merge this for CKAN 2.6.

It doesn't have a migration to delete the table, but maybe it should have one that prompts the user to confirm, if it still has data?

I suggest we do no more work on this PR until after CKAN 2.5 is released.

rossjones:
This is now up to date.

I've changed the migration to emit a warning, and instructions in cases where data exists.  The reason for this is so that people have a last chance to install ckanext-showcase if they have not yet migrated their data.

brew:
A few odds and ends left over, I think:

* related_items.html https://github.com/ckan/ckan/blob/e225985c82e007746aed67ab3504d52d958f86dd/ckan/templates/package/related_list.html (I think this snuck back it during the merge of this commit https://github.com/ckan/ckan/commit/3735d9011b45b0e0264af1697a3e7a174eef0e40)
* related-item.js (https://github.com/ckan/ckan/blob/e225985c82e007746aed67ab3504d52d958f86dd/ckan/public/base/javascript/modules/related-item.js)
* related-item.spec.js (https://github.com/ckan/ckan/blob/e225985c82e007746aed67ab3504d52d958f86dd/ckan/public/base/test/spec/modules/related-item.spec.js)

And a reference to related-item.js in resource.config (https://github.com/ckan/ckan/blob/e225985c82e007746aed67ab3504d52d958f86dd/ckan/public/base/javascript/resource.config#L30)

Apart from that, looks good to me!

rossjones:
@brew Should be good now.

PR #2615, opened 2015-09-04
Fix for ckanext-showcase

davidread:
Cope with the fact that user is optional in the context (judging by the prevelance of similar lines throughout).

Fix needed for: https://github.com/ckan/ckanext-showcase/issues/15


davidread:
@brew maybe you'd like to review this?

brew:
Thanks!

PR #2616, opened 2015-09-04
PEP8 etc

davidread:
Tidy a few key files:
* PEP8
* pylint
* log string substitution refinement (i.e. using commas instead of percent saves the substitution if log level is high enough to avoid the log being done)


rossjones:
LGTM.

Issue #2609, opened 2015-09-02
Delete related items code in favour of ckanext-showcase


rossjones:
The related items code is no longer necessary now that the ckanext-showcase extension exists.  We should delete the related-items code.

There is a migration from related-items to using showcase (in ckanext-showcase) but we probably wouldn't want to deleted related-items before people have a chance to perform the migration.

Perhaps the best approach would be to mark it as deprecated in the next release, and apply any PRs afterwards so that people have a chance to migrate.

wardi:
implemented in #2610

davidread:
I've let people know on ckan-dev that related-items is deprecated now and will be removed in the next release. Had one response - positive.

The plan is:
*  CKAN 2.5 release the front-end is removed #2618 
*  CKAN 2.6 release we remove the rest #2610

rossjones:
The associated PRs are all merged now, closing this ticket.

PR #2618, opened 2015-09-07
Removes the front-end parts of related items.

rossjones:
This is implemented as a partial fix for #2609 as requested by @davidread in #2610


davidread:
We intend to merge this for CKAN 2.6.

It doesn't have a migration to delete the table, but maybe it should have one that prompts the user to confirm, if it still has data?

I suggest we do no more work on this PR until after CKAN 2.5 is released.

PR #2617, opened 2015-09-07
IGroupForm: new groups (of type 'group') use group_form

brew:
Extensions can implement IGroupForm's `group_form` method to define the form template to use for new and editing forms. This isn't working for new forms because the group controller method `_guess_group_types` returns `None` for the default group type [1].

This commit allows `_guess_group_types` to return 'group' for the default group type and adds some tests to ExampleIGroupForm extension tests.

[1] https://github.com/ckan/ckan/blob/9026c30f5d7024623f1512c926acdcc0c46a0490/ckan/controllers/group.py#L131

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -439,7 +439,7 @@ class DefaultGroupForm(object):
         into a format suitable for the form (optional)'''
 
     def db_to_form_schema_options(self, options):
-        '''This allows the selectino of different schemas for different
+        '''This allows the selection of different schemas for different
         purposes.  It is optional and if not available, ``db_to_form_schema``
         should be used.
         If a context is provided, and it contains a schema, it will be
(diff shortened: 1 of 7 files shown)


ckan/lib/plugins.py
```diff
@@ -439,7 +439,7 @@ def db_to_form_schema(self):
         into a format suitable for the form (optional)'''
 
     def db_to_form_schema_options(self, options):
-        '''This allows the selectino of different schemas for different
+        '''This allows the selection of different schemas for different
```
davidread: :+1:

wardi:
@brew was this completely broken before, or did it work to make an IGroupForm plugin that defined a package_form method?

brew:
I think it would work for a custom group type, but not for the default type 'group'.

PR #2622, opened 2015-09-08
Fix api documentation parameter name typo for config_option_show

trickvi:
A minor documentation fix. The function ``logic.action.get.config_option_show`` needs a parameter to identify what config option value to show. That parameter is called ``key`` but the documentation stated that it's called ``id`` which therefore isn't correct.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -3508,8 +3508,8 @@ def config_option_show(context, data_dict):
     :py:func:`~ckan.logic.action.get.config_option_list`), which can be updated with the
     :py:func:`~ckan.logic.action.update.config_option_update` action.
 
-    :param id: The configuration option key
-    :type id: string
+    :param key: The configuration option key
+    :type key: string
 
     :returns: The value of the config option from either the system_info table
         or ini file.


wardi:
`id` would be more consistent with other actions. We should probably change the parameter instead of the docs. We can continue to accept `key` for backwards compatibility

trickvi:
Doesn't matter to me, I can make that change instead. I was only looking at the docs. Will change my pull request to use ``id``. Thanks @wardi

trickvi:
@wardi, the function uses ``_get_or_bust`` which I obviously don't want if I want to also look for ``key`` for backwards compatibility. Does ckan have a similar function to ``_get_or_bust`` that allows me to name ``id`` OR ``key``?

Should I maybe be wrapping it all in a ``try/except KeyError``. I think that will be slower than just checking ``if 'key' in data_dict:`` (but I'd rather want to do that with something like ``_get_or_bust`` so the function won't have to know how ``data_dict`` behaves).

wardi:
@trickvi try/except aren't as slow in Python as they are in other languages. Just do that if it makes the code clearer.

PR #2623, opened 2015-09-08
Stop reraising exceptions caused by notifications.

davidread:
They are secondary errors and should not prevent the change to the dataset occurring. e.g. I am harvesting some data, but because datapusher has an exception, every time the harvester tries to write it has an exception and fails. Admins should look at their logs for these exceptions, and the primary functions should carry on working.

### DIFF
--- a/ckan/model/modification.py
+++ b/ckan/model/modification.py
@@ -1,5 +1,7 @@
 import logging
 
+from ckan.lib.search import SearchIndexError
+
 import ckan.plugins as plugins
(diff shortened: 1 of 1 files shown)


rossjones:
I get a lot of test errors with this around search, will investigate.

rossjones:
It turns out that CKAN as a whole relies on SearchIndexErrors that are raised in notify(), so I have re-instated raising *just* for those errors.

PR #2625, opened 2015-09-08
 Also notify observers of packages involved in a PackageRelationship.

filipefigcorreia:
When relations exist between packages (via PackageRelationship) the related packages are not being included in the notification of what is changing. A consequence is that when viewing the details of a package, the related packages are not shown.

This pull request fixes this issue by including the related packages in the list of changed objects.

Note that this is a second try of [this](https://github.com/ckan/ckan/pull/2469) PR.

### DIFF
--- a/ckan/model/modification.py
+++ b/ckan/model/modification.py
@@ -4,6 +4,7 @@ import ckan.plugins as plugins
 import domain_object
 import package as _package
 import resource
+import package_relationship
 
 log = logging.getLogger(__name__)
 
@@ -63,6 +64,9 @@ class DomainObjectModificationExtension(plugins.SingletonPlugin):
         changed_pkgs = set(obj for obj in changed if isinstance(obj, _package.Package))
 
         for obj in new | changed | deleted:
+            if isinstance(obj, package_relationship.PackageRelationship):
+                changed_pkgs.add(obj.subject)
+                changed_pkgs.add(obj.object)
             if not isinstance(obj, _package.Package):
                 try:
                     related_packages = obj.related_packages()


wardi:
self-assigning because I had #2469

wardi:
This change suggests you're extending CKAN in a very low-level way. Would you tell me more about your use case? The package relationship code has no UI and is in danger of being removed in a future release. Hearing about how some users are using it would be very useful.

filipefigcorreia:
There are non-obvious relations between our datasets that we would like to capture explicitly in CKAN. We're using package relationships to store them, and we have created a custom UI to show and edit them. More specifically, there are two kinds of relationships that we need to express: 
 * Derived from. E.g., dataset ``X`` has derived from dataset ``Y`` (includes [derivative works](https://en.wikipedia.org/wiki/Derivative_work)); 
 * Composed by. E.g., dataset ``Z`` is composed by  datasets ``J`` and ``K``. In this example, ``Z`` probably has no resources of its own, it's just a way to group datasets that have a special meaning when put together.

Our UI allows to see/edit these relationships from both ends. We may be able to generalize tis and contribute it back to CKAN, in case you're interested, but right now it makes an assumption or two about our specific data.

If you choose to remove the package relationship code, we will probably start storing these relations as an extra field, but we would prefer that you didn't. Except for this bugfix, it's working pretty well as it is for us... :-)

wardi:
@filipefigcorreia that sounds very useful. I hope you do find time to try and contribute that work because I've had it requested a number of times.

filipefigcorreia:
Cool. We should be publishing our sources during the next month. After that, we may be able to find a bit of time to clean that code up for a PR.

wardi:
For this PR would you add a test to show that this code works properly?

wardi:
Closing due to lack of activity. Let me know if you'd like to finish this and just need some more direction on writing tests.

PR #2629, opened 2015-09-08
Try travis with docker

davidread:
Travis has a new system running on Docker, but you need to avoid sudo:
http://docs.travis-ci.com/user/migrating-from-legacy/

I'm trying just getting rid of the sudos and seeing if it works...

I also sacrificed multiple postgres versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,13 @@ language: python
 python:
   - "2.6"
   - "2.7"
-env:
-  - PGVERSION=9.1
-  - PGVERSION=8.4
-matrix:
-  exclude:
-    - python: "2.7"
-      env: PGVERSION=8.4
-    - python: "2.6"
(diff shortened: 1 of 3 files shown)


davidread:
I failed to get Travis to approve the apt package libcommons-fileupload-java:amd64=1.2.2-1 (see attempts: https://github.com/travis-ci/apt-package-whitelist/issues?utf8=%E2%9C%93&q=is%3Aissue+author%3Adavidread+ ) so it's not straightforward. Now we have circleci anyway, so closing this work.

amercader:
I'm not sure if we need this package any more, it was needed to work around a Jetty bug. Was that the only blocker? As you say no big deal.

davidread:
Getting different postgres versions to work was another thing that would have needed redoing and I didn't fancy that.

Issue #1572, opened 2014-03-04
API lacks a package_purge function


Vilsepi:
The CKAN API has delete and purge functions for organizations and groups, but dataset only has package_delete. Deleting a dataset using package_delete leaves it in the system with a "deleted" state. If one tries to to create a new dataset with the same dataset URL, the operation fails, claiming that the URL is already in use.

There should be a __package_purge__ function in the API to completely remove a dataset from the database.

Context: For development and testing purposes, I want to be able to create thousands of datasets, delete them, and later on create more datasets (with the same names). This pollutes the database as the old datasets are left in the database.

wardi:
Looks like I might need to add this, so I've assigned myself

wardi:
@mattleduc has done some work on this but ran into a performance issue we'll need to resolve that before adding this to the API.

Somewhat related, I've added an issue to discuss renaming deleted datasets when they conflict with a new one being created: #2438

rossjones:
I think we can close this - @davidread has implemented in b5b8cc5daeb23d432d67ae9defa08bbce06c47ac and it's fast enough (paster takes longer to load than the purge takes).

PR #2634, opened 2015-09-10
dataset_purge

davidread:
Adds the `dataset_purge` action.

Issue: #1572

### DIFF
--- a/ckan/model/group.py
+++ b/ckan/model/group.py
@@ -104,11 +104,11 @@ class Member(vdm.sqlalchemy.RevisionedObjectMixin,
     def __unicode__(self):
         # refer to objects by name, not ID, to help debugging
         if self.table_name == 'package':
-            table_info = 'package=%s' % meta.Session.query(_package.Package).\
-                get(self.table_id).name
+            pkg = meta.Session.query(_package.Package).get(self.table_id)
+            table_info = 'package=%s' % pkg.name if pkg else 'None'
         elif self.table_name == 'group':
-            table_info = 'group=%s' % meta.Session.query(Group).\
-                get(self.table_id).name
+            group = meta.Session.query(Group).get(self.table_id)
+            table_info = 'group=%s' % group.name if group else 'None'
         else:
             table_info = 'table_name=%s table_id=%s' % (self.table_name,
                                                         self.table_id)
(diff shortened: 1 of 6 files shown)


ckan/model/group.py
```diff
@@ -104,11 +104,11 @@ def related_packages(self):
     def __unicode__(self):
         # refer to objects by name, not ID, to help debugging
         if self.table_name == 'package':
-            table_info = 'package=%s' % meta.Session.query(_package.Package).\
-                get(self.table_id).name
+            pkg = meta.Session.query(_package.Package).get(self.table_id)
+            table_info = 'package=%s' % pkg.name if pkg else 'None'
```
davidread: In the thick of a purge, one of my tests caused an exception here. The Package was purged, yet in the bowels of SQLAlchemy it was doing str(member), so `meta.Session.query(_package.Package).get(self.table_id)` was None and therefore calling `.name` on it caused an Exception. So this just adds an 'if' to cope with that situation.

davidread:
On the issue @wardi noted:
> @mattleduc has done some work on this but ran into a performance issue we'll need to resolve that before adding this to the API.

I've tried this code on DGU. It was taking 1 or 2 minutes to purge each dataset typically, and sometimes for many more minutes. Watching the SQL log, the problem was identifying resource_revisions that it was purging - each took about 1.5 seconds. So I've added a postgres index for resource_revision.continuity_id (in a migration) and now it is down to about typically 10s to purge a dataset and half of that is the COMMIT. I managed to get it to take 21s for a datasets with 2000 resource revisions, and 30s for a monster with 93 resources. There are just lots of SQL calls that SQLAlchemy makes for the purge and cascades etc. I think 10s typical is fine.

mattleduc:
@davidread 10s would be significantly more reasonable, we were up to about 4 minutes for some datasets (with lots of extras and a couple resources)

rossjones:
Tests pass, cli works - on my non-DGU ckan paster took longer to run than the purge.

davidread:
It's worth adding that another thing that slowed purge down in the past was deleting associated user_object_role objects. Migration 78 sorted that out very recently: https://github.com/ckan/ckan/blob/master/ckan/migration/versions/078_remove_old_authz_model.py

PR #2635, opened 2015-09-11
Solr CKAN schema version check fix

ojsdude:
This changes the way that the schema is requested to handle the fact
that when a managed schema is used, solr 5.3 renames the schema.xml
file.

### DIFF
--- a/ckan/lib/search/__init__.py
+++ b/ckan/lib/search/__init__.py
@@ -56,7 +56,7 @@ _QUERIES = {
     'package': PackageSearchQuery
 }
 
-SOLR_SCHEMA_FILE_OFFSET = '/admin/file/?file=schema.xml'
+SOLR_SCHEMA_FILE_OFFSET = '/schema?wt=xml'
 
 if SIMPLE_SEARCH:
     import sql as sql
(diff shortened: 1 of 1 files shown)


ojsdude:
The Reason that I uploaded this change was that, on Solr 5.3.0, CKAN is unable to run the check_solr_schema_version() method. I found that this seemed to be because the solr instance needed the schema to be manageable, and solr therefore renamed the file, meaning that CKAN was unable to see it. The other change was due to the fact that the alternate filename given, has a different format.
The checks above fail due to the fact that they run on solr 1.4.0 which does not have this same behaviour

amercader:
@ojsdude If that's the case in Solr 5.3 then it makes sense to change the check, but we need to keep compatibility with Solr 1.x for a while. Perhaps you can get the Solr version somehow before checking the schema version and do one call or the other.

ojsdude:
@amercader   I tried on many occasions to produce something that would be able to do this, however it still failed the tests every time I made the slightest change

rossjones:
There hasn't been any activity on this PR for a few years now.  Time to close it?

rossjones:
@amercader presumably nobody needs/wants support for Solr 1 anymore?

kmbn:

PR #2637, opened 2015-09-11
dataset_purge improvements

davidread:
Issue: #1572

* no revision is needed when purging the package. The only thing being created is a Activity, which is not revisioned.
* no need to explicitly delete an object's revision objects. Despite the very old comment, it is clear that this does happen naturally through cascades. This is verified in the test and by reading the SQL produced.
* more indexes to improve speed further.
* Added a resource to the test for fullness.
* Added helpful comments to the test-core.ini to show people how to use it to see generated SQL commands.


rossjones:
LGTM and purge quite snappy now for me.

PR #2638, opened 2015-09-11
Removes deprecation warning.

rossjones:
There's no ticket for this because it's cleanup.

Tests were logging ```DeprecationWarning: BaseException.message has been deprecated as of Python 2.6``` every time something tried to extract a message from an exception extending
ActionError.  This PR fixes that and also does PEP8 on the __init__.py file.

### DIFF
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -25,7 +25,14 @@ class UsernamePasswordError(Exception):
 
 
 class ActionError(Exception):
-    pass
+
+    def __init__(self, message=''):
+        self.message = message
+        super(ActionError, self).__init__(message)
+
+    def __str__(self):
+        return self.message
+
 
 class NotFound(ActionError):
     '''Exception raised by logic functions when a given object is not found.
@@ -44,7 +51,6 @@ class NotAuthorized(ActionError):
     For example :py:func:`~ckan.logic.action.create.package_create` raises
     :py:exc:`~ckan.plugins.toolkit.NotAuthorized` if the user is not authorized
     to create packages.
-
     '''
     pass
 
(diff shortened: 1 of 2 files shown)


davidread:
What about instead we change occurrences of `e.messages` to `str(e)`? I assume that's more pythonic going forward?

wardi:
is str(e) safe? what happens if e.messages was turned into a unicode object?  Just trying to avoid another #2596

davidread:
There are only four places .message is done, according to the tests:
```
/home/travis/build/ckan/ckan/ckan/controllers/package.py:1249: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  error_message = (e.message or e.error_summary

/home/travis/build/ckan/ckan/ckan/controllers/package.py:1253: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  error_message = e.message

/home/travis/build/ckan/ckan/ckan/controllers/user.py:687: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  error_message = (e.message or e.error_summary

/home/travis/build/ckan/ckan/ckan/controllers/user.py:707: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  error_message = e.message
```
All of which just do:

    h.flash_error(error_message)

so I think calling unicode(e) would better and safer than overriding Exception.message.

wardi:
@davidread :+1:

ckan/logic/__init__.py
```diff
@@ -25,7 +25,14 @@ class UsernamePasswordError(Exception):
 
 
 class ActionError(Exception):
-    pass
+
+    def __init__(self, message=''):
+        self.message = message
+        super(ActionError, self).__init__(message)
+
+    def __str__(self):
```
wardi: this should be `__unicode__` instead, for the cases where the message passed was a unicode object, right?

PR #2639, opened 2015-09-11
[WIP] Fix encoding of setup.py for non-ascii names

### LINKED ISSUE #2636: Unicode decode error when creating an extension
I have a weird name. It has a weird letter: *ö* in it. Trying to create an extension:

    paster --plugin=ckan create -t ckanext ckanext-example

and providing my real name when asked for it ends up with a ``UnicodeDecodeError``:

    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 10: ordinal not in range(128)

This could be solved with me changing my name. But that's not going to happen for everyone with weird characters in their names.

rossjones:
Sets the default-encoding so that what the user enters is utf8 and
this stops the paster templates for exploding.  Also sets the encoding
on the setup.py template so that it can be run if it has utf8 chars in it

This should fix #2636

### DIFF
--- a/ckan/pastertemplates/__init__.py
+++ b/ckan/pastertemplates/__init__.py
@@ -51,6 +51,9 @@ class CkanextTemplate(Template):
     def check_vars(self, vars, cmd):
         vars = Template.check_vars(self, vars, cmd)
 
+        reload(sys)
+        sys.setdefaultencoding('utf-8')
+
         if not vars['project'].startswith('ckanext-'):
             print "\nError: Project name must start with 'ckanext-'"
             sys.exit(1)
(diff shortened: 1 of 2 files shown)


ckan/pastertemplates/__init__.py
```diff
@@ -51,6 +51,9 @@ class CkanextTemplate(Template):
     def check_vars(self, vars, cmd):
         vars = Template.check_vars(self, vars, cmd)
 
+        reload(sys)
+        sys.setdefaultencoding('utf-8')
```
wardi: yikes, this could have really surprising side effects. I think we should fix our bytes/unicode confusion issues where they lie instead of with a global flag like this.

rossjones: I think it's inside paster

PR #2640, opened 2015-09-14
Exception when viewing Organizations extended with IGroupForm

brew:
Viewing an organization that's been extended with IGroupForm causes an exception:

`Exception: menu item 'organization_activity' need parameter 'offset'`

The default organization_activity route [1] doesn't include an `offset` parameter, but a new route is set up for custom IGroupForm plugins in `lib/plugins.py` [2], which does include the `offset` param. The `group_activity` route includes the offset param, so it being missing from organization is likely an oversight.

Suggest:

1) Add the offset parameter to the organization read_base.html template in the same way it is in the group read_base.html

2) Adding the `{offset}` param to the default organization_activity route.

---

[1] https://github.com/ckan/ckan/blob/ed5fe6e86e520800edcf84d37e5f0c1d8a6ef485/ckan/config/routing.py#L319

[2] https://github.com/ckan/ckan/blob/ed5fe6e86e520800edcf84d37e5f0c1d8a6ef485/ckan/lib/plugins.py#L191

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -316,7 +316,7 @@ def make_map():
                       'member_delete',
                       'history'
                   ])))
-        m.connect('organization_activity', '/organization/activity/{id}',
+        m.connect('organization_activity', '/organization/activity/{id}/{offset}',
                   action='activity', ckan_icon='time')
         m.connect('organization_read', '/organization/{id}', action='read')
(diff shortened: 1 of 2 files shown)


davidread:
Added an issue for this: #2765

wardi:
@amercader can we backport this to 2.4? I'll add the tag

amercader:
done @wardi

PR #2641, opened 2015-09-14
Search index rebuild takes a while; let's give it a progress counter

deniszgonjanin:
Search index rebuild can take an hour or two when you have many thousands of datasets. This PR shows a simple counter, so that the user can get a little bit of feedback on the progress, and know that the rebuild isn't stuck somewhere.

### DIFF
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -402,14 +402,14 @@ class SearchIndexCommand(CkanCommand):
     '''Creates a search index for all datasets
 
     Usage:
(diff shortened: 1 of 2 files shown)


davidread:
This is neat! Could you add a --quiet option for when this is run in build scripts?

Ian says there's an off-by-one error too ;-)

deniszgonjanin:
Heh, yes there is

deniszgonjanin:
Should be good to go

davidread:
This is great. Creating new PR with a minor test fix: #2645

PR #2643, opened 2015-09-15
ITranslation docs

joetsoi:
#2461 should be merged first

I've removed some old genshi stuff from setup.py
The default plugin template will now generate a babel config and setup the message extractors so plugin writers can run a `python setup.py extract_messages` and all the other translation stuff without thinking about it too much, although I did not include the ITranslation interface into the default extension. Which I perhaps could do.


doc/extensions/translating-extensions.rst
```diff
@@ -0,0 +1,160 @@
...
+.. tip::
+
+   If you have generated a new extension whilst following this tutorial the
+   default template will have generated these files for you and you can simply
+   run the ``extract_messages`` command immediately.
+
+Check your ``setup.py`` file in your extension for the following lines
+
+.. code-block:: python
+    :emphasize-lines: 5-6, 12-15
```
wardi: ooh. I learned something new

joetsoi:
http://docs.ckan.org/en/latest/extensions/translating-extensions.html

crap apparently I screwed up the docs links, i'll fix them in a new pr

PR #2647, opened 2015-09-16
allow exporting and importing users, with their passwords

wardi:


wardi:
I'm going to add api key importing to this too, as it seems to fit

wardi:
@joetsoi I've changed my mind, I'll do the password hash setting in a separate PR. This one is controversial enough as it is :-)

wardi:
This is better, I'm not importing authz into logic modules any more.

PR #2653, opened 2015-09-22
Separates search.clear() into two functions

### LINKED ISSUE #2652: ckan.lib.search.clear is dangerous and should be more explicit.
The call to clear() in ckan.lib.search takes an optional package id.  If no package_id is provided (the default parameter is None) then it will clear the entire search index.

This is likely to make programmer error much more pronounced when they accidentally pass in None, or an empty string and their search index disappears - ask me how I know?

rossjones:
Fixes #2652 by separating search.clear() into search.clear(pkg) and search.clear_all() to make it harder to accidentally pass a falsey value to clear and wipe your search index.

### DIFF
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -222,7 +222,7 @@ class ManageDb(CkanCommand):
                 os.remove(f)
 
             model.repo.clean_db()
-            search.clear()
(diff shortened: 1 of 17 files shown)


amercader:
LGTM

PR #2668, opened 2015-10-01
Add a before_edit method to the IPackageController

LaurentGoderre:
The reason why we need this is that we use Fluent on datasets titles. For the view pages, we use the `before_view` method to resolve the fluent fields to the current UI language but we can't on the edit page which cause fluent field to be rendered as dictionaries.

### DIFF
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -591,6 +591,13 @@ class IPackageController(Interface):
         '''
         return pkg_dict
 
+    def before_edit(self, pkg_dict):
+        '''
+             Extensions will recieve this before the dataset gets
+             displayed. The dictionary passed will be the one that gets
+             sent to the template.
+        '''
+        return pkg_dict
 
 class IResourceController(Interface):
     """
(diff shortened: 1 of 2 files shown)


ckan/plugins/interfaces.py
```diff
@@ -591,6 +591,13 @@ def before_view(self, pkg_dict):
         '''
         return pkg_dict
 
+    def before_edit(self, pkg_dict):
+        '''
+             Extensions will recieve this before the dataset gets
```
davidread: spelling 'receive'

wardi:
The general response from the meeting is that we would prefer to have the "view" version of the data separate from the data sent to the form. I can't think of a way to do that while maintaining compatibility with the existing breadcrumb templates and edit form, though.

If ckan were to grow a standard way of storing multilingual titles so that the breadcrumbs could always have the correct version displayed we wouldn't need a before_edit or before_view at all, right?

@LaurentGoderre would you confirm that the only thing you need this for is to fix the breadcrumbs?

@amercader did we settle on title_lang as the place to store multilingual titles?

LaurentGoderre:
I believe that is correct isn't it @TkTech ?

amercader:
@wardi Sorry, I was meant to create an issue to discuss this. I'm happy with it but let's confirm with the rest of the team: https://github.com/ckan/ckan/issues/2678

wardi:
My preference is to fix the display of multilingual titles the way we're discussing in #2678 For now multilingual titles can already be handled by overriding templates in a number of places.

There are also unresolved issues with how this approach could affect data edited in the form in potentially surprising ways.

PR #2670, opened 2015-10-06
PostgreSQL version 9.2

davidread:
Issue: #2555


doc/maintaining/upgrading/upgrade-postgres.rst
```diff
@@ -0,0 +1,272 @@
+===========================================
+Upgrading Postgres for the CKAN 2.5 release
```
wardi: Update this to say 2.6 and I think we can merge this as it is.

davidread:
I didn't get the postgis_restore.pl script to work sadly.

wardi:
Can we not link to the instructions on http://postgis.net/docs/postgis_installation.html#upgrading and leave it at that?

@amercader Our bounding box search is based on solr now for most users isn't it? How many people need postgis in their ckan database?

davidread:
Tbh it was a lot of work putting the doc together, based on:

* those postgis instructions
* some guy's elaboration of some of the commands cos they aren't clear in the official instructions
* a bit of ckan knowledge of what databases need migrating and which don't
* your suggestion about having both versions of postgres installed at the same time. It's a great plan for minimal downtime, but really adds to the difficulty as all the pg commands need directing at a particular postgres version and that proved tricky to work out.

I believe that even though ckanext-spatial uses solr for the geo-search, ckanext-spatial still requires postgis to store the datasets' bounding boxes as a vector, rather than a blob that solr can index.

wardi:
@davidread thanks, I wasn't aware. So we should find a way to get this done, I'll see if we have resources on our side.

doc/maintaining/upgrading/upgrade-postgres.rst
```diff
@@ -0,0 +1,272 @@
...
+
+     sudo -u postgres psql --cluster 9.4/main -f backup_roles.sql
+
+   Expect there will be one error::
+
+     psql:backup_roles.sql:22: ERROR:  role "postgres" already exists
+
+   which you can ignore - it should carry on regardless and finish ok.
+
+#. Create the databases and enable PostGIS on them:
```
amercader: No need to enable PostGIS on the DataStore database

doc/maintaining/upgrading/upgrade-postgres.rst
```diff
@@ -0,0 +1,272 @@
...
+
+   And you should read through the differences in postgresql.conf. This is a handy way to do this whilst ignoring changes in the comment lines::
+
+     diff -u -B <(grep -vE '^\s*(#|$)' /etc/postgresql/9.1/main/postgresql.conf)  <(grep -vE '^\s*(#|$)' /etc/postgresql/9.4/main/postgresql.conf)
+
+   Once you've finished your changes, restart both versions of |postgres|::
+
+     sudo /etc/init.d/postgresql restart 9.4
+
+#. REALLY? Follow the instructions in :ref:`postgres-setup` to setup |postgres| with a user and database. Ensure your username, password and database name match those in your connection settings (see previous section.)
```
davidread: It's a note to self to check this line. So yes it should go.

amercader:
These instructions look great, I think that @davidread has gone far beyond the detail expected for an external requirement upgrade, and PostGIS is only relevant if you are using the spatial extension. I'd add a note linking to the official PostGIS upgrade docs when postgis_upgrade.pl is mentioned referring users there if they find an error.

Apart from a couple minor corrections, which I'm happy to make, I think this can be merged as is.

@wardi it's true that PostGIS is still required even if using the Solr back end, but there is no internal reason, mainly lack of time on my side to decouple the two things. I'd love to revisit this specially with relation to the different search back ends support being done by @tktech

TkTech:
@amercader Anything preventing this from going in now (other than the doc conflict)?

amercader:
Just some minor tweaks, I'll push them now

amercader:
@TkTech ready to go. I've marked for cherry-picking into dev-v2.6

PR #2674, opened 2015-10-06
Deleting routes

davidread:
Fixes: #2673 
- [ ] Test
- [ ] Changing routes?


davidread:
Closing as this is to do with pylons which we're leaving.

PR #2675, opened 2015-10-06
Fix title munge resulting in multiple dashes

### LINKED ISSUE #2669: title munger allows repeated dashes
If you munge title ```Flood    data``` then those three spaces get converted to 2 dashes: ```flood--data```, when it should really just be one ```flood-data```.

There are hacks in several harvesters to fix this, but it's better fixed at source in core ckan.

davidread:
Fixes #2669


davidread:
This function is used in 2 places:
* In util API "munge_title_to_package_name" used by the dataset form to fill in the title field automatically.
* In harvesters that choose a new name for a dataset.

So I think it is perfectly safe to make a slight change to the algorithm like this.

PR #2680, opened 2015-10-09
 Upgrade requests version pin

davidread:
Issue: #2470


davidread:
Merging myself, as it was not me that suggested this change.

PR #2682, opened 2015-10-13
explain and fix for jetty HTTP 500 error

yanikkoval:
I have this error every time, while following docs and using clean ubuntu 12.04 x64

### DIFF
--- a/doc/maintaining/installing/install-from-source.rst
+++ b/doc/maintaining/installing/install-from-source.rst
@@ -287,6 +287,24 @@ installed, we need to install and configure Solr.
     or::
 
         JAVA_HOME=/usr/lib/jvm/java-6-openjdk-i386/
+    
+        
+    .. note::
+
+    Very common error I'm got, when trying Install CKAN on clean Ubuntu 12.04 x64::
+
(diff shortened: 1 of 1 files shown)


davidread:
This problem is covered lower down the document: https://github.com/wide-awake/ckan/blob/patch-1/doc/maintaining/installing/install-from-source.rst#unable-to-find-a-javac-compiler

However it doesn't mention the possibility of the JRE being installed. In that circumstance, do we need to mention add about changing JAVA_HOME, or is that done anyway by ```sudo apt-get install openjdk-6-jdk```?

rossjones:
I think the documentation has since been updated, and it's reasonably clear under http://docs.ckan.org/en/2.8/maintaining/installing/install-from-source.html#setup-solr how to resolve this error.

davidread:
Closing as there are a number of unresolved questions in the comments above and:
* Ubuntu 12.02 is not supported any more, would need to test this on 14.04 and 16.04
* "Very common error I'm got" would need improving
Happy to look at this again if we can resolve some of these things or other people are experiencing this issue still.

PR #2684, opened 2015-10-13
Do not crash on AttributeError in dict_list_reduce helper function

dlax:
I got the following error on a freshly installed CKAN 2.4.1 (source install) after adding a few datasets on a request to the home page.

```
[...]
File '/usr/local/src/ckan/ckan/templates/snippets/package_item.html', line 66 in block "resources_inner"
  {% for resource in h.dict_list_reduce(package.resources, 'format') %}
File '/usr/local/src/ckan/ckan/lib/helpers.py', line 872 in dict_list_reduce
  value = item.get(key)
AttributeError: 'int' object has no attribute 'get'
```

Not sure why this occurs... Looking closer, I saw `0` and `1` integers values as `item` (`key` was `'format'`).

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -868,7 +868,11 @@ def dict_list_reduce(list_, key, unique=True):
     values for the key with unique values if requested. '''
     new_list = []
     for item in list_:
-        value = item.get(key)
+        try:
+            value = item.get(key)
+        except AttributeError:
+            # item might not be a dict.
+            continue
         if not value or (unique and value in new_list):
             continue
         new_list.append(value)


ckan/lib/helpers.py
```diff
@@ -868,7 +868,11 @@ def dict_list_reduce(list_, key, unique=True):
     values for the key with unique values if requested. '''
     new_list = []
     for item in list_:
-        value = item.get(key)
+        try:
+            value = item.get(key)
+        except AttributeError:
+            # item might not be a dict.
+            continue
```
rossjones: As the function is documented as taking a list of dicts, it's possibly not the right thing to do, to just silently ignore the bad data passed in.  Perhaps if you raise an error that points out that it has been passed bad data (e.g. not a list of dicts) may help in tracking down the real cause of the problem (possibly in a schema somewhere?).

dlax: Ross Jones wrote:
> As the function is documented as taking a list of dicts, it's possibly
> not the right thing to do, to just silently ignore the bad data passed
> in. Perhaps if you raise an error that points out that it has been
> passed bad data (e.g. not a list of dicts) may help in tracking down the
> real cause of the problem (possibly in a schema somewhere?).


I agree this is not the right thing to do (just did not read the 
docstring :P). I can debug this live if you point me to something to 
look for. The full traceback does not look very helpful from my point of 
view...

dlax: It appears that the `package.resources` object passed to `dict_list_reduce` is actually not a `list` but a `dict` (corresponding to one resource). But the package at stake actually has two resources.

rossjones: I think the thing to investigate is why when calling ```dict_list_reduce``` with 

```h.dict_list_reduce(package.resources, 'format')```

we get things that are not resource dicts.  package.resources should *definitely* be a list of dicts.

why one of the items is an int and not a dict.

Perhaps just dumping ```{{package.resources}}``` in the package_item template might show what it is being passed, I can only imagine that a plugin/custom-schema is interfering somehow.

Comment (dlax) [inline ckan/lib/helpers.py]: Yes, I'm using ckanext-dcat and this dataset comes from a DCAT harvest.

Comment (dlax) [inline ckan/lib/helpers.py]: ``package.resources`` is just one dict which looks like:
```
{0: {u'cache_last_updated': None, 
       u'issued': u'2015-10-13T15:00:17.332357', 
       u'uri': u'http://localhost:8080/distribution/1301',
       u'package_id': u'6b0af45b-81c5-425c-bcb2-42f291937ae7',
      ...}
}
```

Comment (rossjones) [inline ckan/lib/helpers.py]: This is weird, perhaps @amercader might know if this is a dcat extension specific weirdness.

As a temporary workaround, you can probably get away with (obviously untested)..

```
h.dict_list_reduce(package.resources.values(), 'format')
```

but it should only be a temporary workaround.

kmbn:

PR #2685, opened 2015-10-13
Exception handling

neon-ninja:
These two commits resolve issues I was having while trying to run the ckanext-oaipmh harvester.

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -11,6 +11,8 @@ from ckan.common import OrderedDict, _, c
 
 import ckan.lib.maintain as maintain
 
+from sqlalchemy.orm.exc import DetachedInstanceError
+
 log = getLogger(__name__)
 
 
@@ -118,6 +120,8 @@ def _get_user(username):
             return c.userobj
     except TypeError:
         # c is not available
+        pass
+    except DetachedInstanceError:
         pass
     # Get user from the DB
     return model.User.get(username)
--- a/ckan/lib/search/index.py
+++ b/ckan/lib/search/index.py
@@ -245,6 +245,8 @@ class PackageSearchIndex(SearchIndex):
                         value = None
                 except ValueError:
                     continue
+                except TypeError:
+                    continue
             new_dict[key] = value
         pkg_dict = new_dict
 


ckan/authz.py
```diff
@@ -119,6 +121,8 @@ def _get_user(username):
     except TypeError:
         # c is not available
         pass
+    except DetachedInstanceError:
```
rossjones: I can't imagine an instance where catching and passing silently on DetachedInstanceError is the right thing to do.  This is likely to be masking the real error elsewhere. Do you have a deeper stack trace from when this fixed your problem?  I suspect that another query somewhere has left the c.userobj hanging without being connected to a session (somehow). This hasn't shown up in core before, is it possible it's an issue in your extension?

I know we silently pass on the TypeError above, we shouldn't and I'll raise a separate issue for that.

ckan/lib/search/index.py
```diff
@@ -245,6 +245,8 @@ def index_package(self, pkg_dict, defer_commit=False):
                         value = None
                 except ValueError:
                     continue
+                except TypeError:
```
rossjones: Again I think this may be masking a problem elsewhere.  I'm not suggesting that we shouldn't handle all cases - but given the previous comment - perhaps it would be more straight-forward to find the timezone on the date being passed in oaipmh.harvester?

neon-ninja: Stack trace:

2015-10-16 10:28:00,776 ERROR [ckan.lib.search] can't compare offset-naive and offset-aware datetimes
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/search/__init__.py", line 103, in dispatch_by_operation
    index.update_dict(entity)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/search/index.py", line 103, in update_dict
    self.index_package(pkg_dict, defer_commit)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/search/index.py", line 240, in index_package
    if date != bogus_date:
TypeError: can't compare offset-naive and offset-aware datetimes

These lines:

if date != bogus_date:
                        value = date.isoformat() + 'Z'

Attempt to parse a date, and if the parse succeeds, converts it to isoformat and adds a Z. However, if the date is already in isoformat with a trailing Z it throws an exception. Perhaps instead of using bogus_date it may be wiser to simply check if the string is empty or None?

ckan/authz.py
neon-ninja: The stack trace wasn't particularly useful, but the Travis CI linter revealed this variable that was assigned but never used. I think I resolved this with this commit - https://github.com/neon-ninja/ckanext-oaipmh/commit/2a835955b45c558ff3872a7e19ddca56f76f2c52

rossjones: Yes, that looks promising. Passing the username in the context is how most action functions expect to find it, no need to fetch the user yourself.

ckan/lib/search/index.py
rossjones: There's definitely a problem there - could you add this as an issue please?

neon-ninja: No problem, done https://github.com/ckan/ckan/issues/2689

rossjones:
Hope you don't mind if I close this as one issue was fixed at the source, and the other is better handled by proper tz handling in the search index.

PR #2589, opened 2015-10-15
Package_search does not return more than a 1000 rows

amercader:
Package_search allows to define the amount of results through the rows parameter, but apparently values more than a 1000 don't have effect. At least in our instance at https://www.opendata.fi/en and data.gov.uk.

If this is by design, it should at least be documented since it is used in quite many extensions for example: 
https://github.com/ckan/ckanapi-exporter/blob/master/ckanapi_exporter/exporter.py#L18

which returns only 1000 results.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1768,7 +1768,8 @@ def package_search(context, data_dict):
         documentation, this is a comma-separated string of field names and
         sort-orderings.
     :type sort: string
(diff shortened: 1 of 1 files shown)


amercader:
@Zharktas There is indeed a hard limit of 1000 datasets per query, to prevent performance and DOS issues

https://github.com/ckan/ckan/blob/master/ckan/lib/search/query.py#L314

I've submitted a quick patch to update the docs to reflect this.

PR #2694, opened 2015-10-20
Do not crash when indexing fails on a dataset with unicode.

TkTech:
If an exception is raised while indexing, and that dataset contains unicode, a second exception will be raised trampling the first.


wardi:
duplicate of #2596

Hey @brew I think @TkTech is getting impatient :-)

TkTech:
Whoops, my apologies didn't notice I'd already run into this issue :)

Issue #2697, opened 2015-10-21
package_autocomplete only find packages that start with the given search query


metaodi:
The [current implementation of package_autocomplete](https://github.com/ckan/ckan/blob/2f8d75c4da1ce5bf1fe9c59ab0f9aa8e3cd32270/ckan/logic/action/get.py#L1611) looks like that:

```python
like_q = u"%s%%" % q
```

So only packages that start with the given string will be returned.
I expected something like `%searchterm%'` to get better results. E.g. `format_autocomplete` uses this format as well.

Or was this intentional?

metaodi:
PR for this issue: #2698

rossjones:
I merged this and then reverted it when wardi pointed out that it doesn't have the right type of index.  As punishment I'll make it work with Solr tomorrow and send a PR.  Sorry again.

wardi:
I don't think this action is used anywhere in core CKAN at the moment, maybe we should consider deprecating it instead of updating it?

@metaodi if you use this action and want it to behave differently why not add it to your own extension?

metaodi:
@wardi Ah sure I can put it in my extension, I just thought this was a bug. I'm fine with deprecating it.

PR #2698, opened 2015-10-21
[#2697] Make sure package_autocomplete uses LIKE with wildcards

metaodi:
The current implementation uses 'q%' to search for packages starting
with a certain search term. This commit changes this to use the
following format: '%q%', thus search for the term anywhere in the title
and name field.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1608,7 +1608,7 @@ def package_autocomplete(context, data_dict):
     limit = data_dict.get('limit', 10)
     q = data_dict['q']
 
-    like_q = u"%s%%" % q
+    like_q = u"%%%s%%" % q
 
     query = model.Session.query(model.Package)
     query = query.filter(model.Package.state == 'active')


rossjones:
Makes a lot more sense, thanks for the PR.

wardi:
No, this will kill performance for many datasets because it won't use the index in postgres any more. We should be using solr instead for returning these results.

rossjones:
@metaodi I jumped the gun a bit merging this too quickly :( Sorry, I'll have to revert it.

metaodi:
Wait, I explicitely want to use this *instead* of Solr. If I want to use Solr, I can use `package_search`, but this doesn't work. If for your instance this would be a performance issue, just switch to to package_search. But for most instances with a couple of thousand datasets, this shouldn't be a problem. And is the performance of the old version so much better than this?

Maybe I don't understand Solr, but I don't think it's made for these kind of queries. Or do you have something else in mind then using `package_search`?

rossjones:
So looking at it, we'd have had solr setup to index word-parts, not just words, so I think you're right, solr won't really work for this.  

Perhaps the solution is to modify the index to include text_pattern_ops?

PR #2705, opened 2015-10-23
Read both resource proxy constants from config file

mordae:
Hi, using the `1 * 1024 * 1024` value in `ckan.resource_proxy.max_file_size` configuration option has been failing with integer format exception, so I took the liberty to change it to an actual integer. I have also added the second parameter `ckan.resource_proxy.chunk_size`, which originally had a rather small value of `512`. New default is `4096`, which is the PAGESIZE on most platforms.

Please consider merging in.

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -151,7 +151,9 @@ ckan.feeds.author_link =
 
 # Resource Proxy settings
 # Preview size limit, default: 1MB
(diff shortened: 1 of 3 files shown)


amercader:
This looks great, only one minor PEP8 error and it can get merged:

-----------------------------------
ckanext/resourceproxy/controller.py
-----------------------------------
E501 ln:15 line too long (91 > 79 characters)

mordae:
It should be fine now.

amercader:
Thanks for the patch @mordae

PR #2706, opened 2015-10-23
[WIP] Initial implementation of providing async calls via RQ

rossjones:
Hi,

I’ve started this PR so that we can have a discussion about approach for solving https://github.com/ckan/ideas-and-roadmap/issues/66 - suspect this will be post 2.5 now.

My initial approach is to provide a function that we (and plugin developers) can call to run a task asynchronously.  This task should be enqueued pretty quickly and return as soon as it can.

Currently that code looks like 

```
from ckan.lib.task import async 
from myckanextention.tasks import my_async_task
async(my_async_task, [arg1, arg2], priority=‘low’)
```

This will then get picked up by rqworker running as a separate process and executed. 

There are lots of outstanding questions:
1. Currently I’ve implemented low, medium and high priority, but rqworker is flexible enough to listen to several, or only one queue, so this can be configured as necessary.  Enough levels? Too many priority levels? 
2. Importing the function you want to call is potentially onerous, should we have core and plugins register the functions they want to call against a name (via the ITask plugin), which can be used rather than the logic layer having to import them - like ITemplateHelpers and IAction? Sort of like `async(‘harvest_run’, [])`. 
3. Should rqworker be allowed to pre-import the CKAN runtime, or should tasks be force to run independently (only communicating via the API)?
4. Is this approach sane? Should we leave it to those who want to use it to do all the work with setup and call rq directly? Should CKAN be providing the entry-point (I obviously think that yes, it should).
5. We should handle the case where someone doesn’t have redis installed (and therefore async tasks fail).  Should this be fatal, i.e. if you try and use async() and redis isn’t installed - it should crash?

### DIFF
--- a/bin/travis-install-dependencies
+++ b/bin/travis-install-dependencies
@@ -14,7 +14,7 @@ fi
 
 # Install postgres and solr
 sudo apt-get update -qq
-sudo apt-get install postgresql-$PGVERSION solr-jetty libcommons-fileupload-java:amd64=1.2.2-1
+sudo apt-get install postgresql-$PGVERSION solr-jetty libcommons-fileupload-java:amd64=1.2.2-1 redis-server
 
 if [ $PGVERSION == '8.4' ]
 then
--- a/ckan/lib/queue.py
+++ b/ckan/lib/queue.py
@@ -1,0 +1,107 @@
+from pylons import config
+
+from rq import Queue
+from redis import Redis
+
+from logging import getLogger
+log = getLogger(__name__)
+
+
+queues = None
+
+
+def init_queues():
+    '''
+    Attempt to initialise the three queues on first use. There's no need
+    to create these until we need them (we might not). This will explode
+    if redis is not reachable
+    '''
+    global queues
+
+    connection_url = config.get(u'ckan.queues.url',
+                                u'redis://localhost:6379/0')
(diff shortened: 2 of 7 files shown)


ckan/lib/task.py
```diff
@@ -0,0 +1,101 @@
+from rq import Queue
+from redis import Redis
+
+from logging import getLogger
+log = getLogger(__name__)
+
+redis_conn = Redis()
```
rossjones: We should late-bind this (assuming lib doesn't already late-bind) and move it into init_queues()

TkTech: Definitely don't think this should be setup like this (or here at all). Here's an example from one of my tools:

```python
if args['worker']:
        with app.app_context():
            with Connection(Redis.from_url(app.config['BROKER_URL'])):
                qs = [Queue(n) for n in args['--names']] or [Queue()]
                w = Worker(qs)
                w.work()
```

Just add this as a paster command.

Benefits:

1. Properly maintains the Connection stack
2. No connection state or details in tasks.py
3. Far more control over the queus than what you can get with the rqworker command (which is really just a dev tool)
4. Uses the config from Flask in this case, but would use the development.ini.

rossjones:
Closing this PR, I think it's unnecessary now.

PR #2707, opened 2015-10-23
fixes error message when using GET on POST-only actions

perceptron-XYZ:
When request_data is None and try_url_params, it basically means you are calling a POST-only action with GET method, the original error message is not quite clear. I bumped into the error message when I was calling organization_purge in the url bar in my browser, and it took me some time to figure out why, so I think the modification would give users a better idea what exactly went wrong

### DIFF
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -893,7 +893,7 @@ class ApiController(base.BaseController):
             cls.log.debug('Retrieved request body: %r', request.body)
             if not request_data:
                 if not try_url_params:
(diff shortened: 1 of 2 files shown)


wardi:
@perceptron-XYZ would you fix the pep8 error? This isn't perfect but it's better than what we have, so I'll merge it once the tests are passing.

perceptron-XYZ:
Yes sure. I'm on my way to office.I'll take a look when I come in.

perceptron-XYZ:
@wardi I got the pep8 and test case fixed

PR #2709, opened 2015-10-26
resource_edit incorrectly setting action=new instead of action=edit (2.3-latest backport)

TkTech:
resource_edit incorrectly setting `action='new'` instead of `action='edit'`. This value being set properly is required for our extension to behave properly when setting field defaults (we only want to set a default value if action='new').


amercader:
Will be backported when #2710 is merged

PR #2710, opened 2015-10-26
resource_edit incorrectly setting action to new instead of edit.

TkTech:
resource_edit incorrectly setting `action='new'` instead of `action='edit'`. This value being set properly is required for our extension to behave properly when setting field defaults (we only want to set a default value if action='new').

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -644,7 +644,7 @@ class PackageController(base.BaseController):
         errors = errors or {}
         error_summary = error_summary or {}
         vars = {'data': data, 'errors': errors,
-                'error_summary': error_summary, 'action': 'new',
+                'error_summary': error_summary, 'action': 'edit',
                 'resource_form_snippet': self._resource_form(package_type),
                 'dataset_type': package_type}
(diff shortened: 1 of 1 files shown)


ckan/controllers/package.py
```diff
@@ -590,7 +590,8 @@ def resource_edit(self, id, resource_id, data=None, errors=None,
 
             context = {'model': model, 'session': model.Session,
                        'api_version': 3, 'for_edit': True,
-                       'user': c.user or c.author, 'auth_user_obj': c.userobj}
+                       'user': c.user or c.author, 'auth_user_obj': c.userobj,
+                       'action': 'edit'}
```
wardi: What are the context changes needed for?

TkTech: Simple consistency. If a resource is being edited, the action is edit.

wardi: But `'action'` didn't exist in the context dict before. Is there a reason you're adding it here and below?

TkTech: Simple consistency. If a resource is being edited, the action is edit. O-O

PR #2711, opened 2015-10-26
Use old style Travis builds

palcu:
Travis has migrated to the new [container infrastructure](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) for every new build. We cannot use this because we use some commands with `sudo`.

I've modified the `.travis.yml` template so that new extension build on the old infrastructure.

### DIFF
--- a/ckan/pastertemplates/template/+dot+travis.yml_tmpl
+++ b/ckan/pastertemplates/template/+dot+travis.yml_tmpl
@@ -1,4 +1,5 @@
 language: python
+sudo: required
 python:
     - "2.6"
     - "2.7"


palcu:
Also, the CKAN version for building the extension against is [hardcoded](https://github.com/ckan/ckan/blob/30a3c133fc033ed946d2301ee9299395a314fc68/ckan/pastertemplates/template/bin/travis-build.bash_tmpl#L13). I would suggest building against the latest master by default or somehow writing the current release branch.

amercader:
@davidread you've been playing with Travis new infrastructure, what do you think?

davidread:
This is not necessary because it defaults to `sudo:required` and travis has been building fine on its 'legacy infrastructure' since it introduced the new ones, but it doesn't harm to be explicit, so I'll merge it.

PR #2717, opened 2015-10-29
Enable multiple select for form.select

hvwaldow:
The Jinja2 macro `form.select` can be used to create a multiple-selection element by passing `attrs={"multiple"="multiple"}`. A small change is necessary to make a multiple default selection work (by assigning a list to the parameter `selected`).


hvwaldow:
pull req went wrong re-submitting

PR #2718, opened 2015-10-29
Enable multiple for form.select

hvwaldow:
The Jinja2 macro `form.select` can be used to create a multiple-selection element by passing `attrs={"multiple"="multiple"}`. A small change is necessary to make a multiple default selection work (by assigning a list to the parameter `selected`).

### DIFF
--- a/ckan/templates/macros/form.html
+++ b/ckan/templates/macros/form.html
@@ -71,7 +71,7 @@ name        - The name of the form parameter.
 id          - The id to use on the input and label. Convention is to prefix with 'field-'.
 label       - The human readable label.
 options     - A list/tuple of fields to be used as <options>.
-selected    - The value of the selected <option>.
+selected    - The value(s) of the selected <option>s. A list of strings for multiple selected <option>s.
 error       - A list of error strings for the field or just true to highlight the field.
 classes     - An array of classes to apply to the control-group.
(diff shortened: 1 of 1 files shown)


wardi:
This will change the behaviour for the non-multiple-selection case. `in` on strings will check for the presence of a substring.

hvwaldow:
I think you overlooked the new line 88.

wardi:
You're right, I did!

wardi:
I still see some problems with this change. This change is to support extension writers' use of the ckan select macro for multi-select. 

1. Extensions can already add their own form elements (e.g. scheming adds multiple select with a [snippet](https://github.com/open-data/ckanext-scheming/blob/master/ckanext/scheming/templates/scheming/form_snippets/multiple_select.html) ).

2. The documentation you've added doesn't indicate to users that they need to add `attrs={"multiple"="multiple"}` to the call for this to work properly, but even if it did, that's not a very nice interface.

3. This introduces type detection into the interface which may create bugs (e.g. I believe the old code worked with integer values and this won't)

I'd be happier with a separate `select_multiple` macro that always took a sequence for its selected items, but even then, I'm leery about adding features that aren't used by or tested in ckan core.

hvwaldow:
Ok. Considering your last sentence I'll then just add something like that to my extension as snippet. Thanks for the feedback anyway.

Issue #2678, opened 2015-10-07
Core model multilingual fields


amercader:
The consensus for handling multilingual metadata fields seems to be to follow the way ckanext-fluent handles them.

For custom fields this looks uncontroversial:

```
{
...
  "favourite_fruit": {
    "en": "Peach",
    "ca": "Préssec",
    "es": "Melocotón"
  }
...
}
```

There is an issue though with using this approach on existing core fields like `title` or `notes`. Up until this point these have always had a single string value, and changing this to return a dict on some occasions would break things, both in core (eg in templates) or in extensions.

1. For dealing with this **in the current API version** I'm proposing that for these particular core fields that need to be translated we add a separate field:

   ```
{
...
  "title_lang": {
    "en": "Annual Budget",
    "ca": "Pressupost Anual",
    "es": "Presupuesto Anual"
  }
...
}
```
  I have no strong preference for what the `_lang` suffix actually is called (`_trans`, `_i18n` ...)

  The list of fields that would need these is:

  * Dataset title
  * Dataset notes
  * Resource name
  * Resource description
  * Resource format
  * Tag name (these are likely to be handled separately)

  Just to clarify, for now it would be ckanext-fluent (or any other extension) who will create these fields, no changes on ckan core would be made.

2. Moving forward, **on a new API version**, we can decide whether the same pattern can be applied directly to core fields as well:

  ```
{
...
  "title": {
    "en": "Annual Budget",
    "ca": "Pressupost Anual",
    "es": "Presupuesto Anual"
  }
...
}
```
The main issue I see is what happens with those CKAN instances (the majority) that don't handle multilingual metadata. Do we allow string and dict values on the same field? (that doesn't sound like a good idea tbh). Do we always enforce a language key, even if there is only one for the instance default locale? eg:

  ```
{
...
  "title": {
    "en": "Annual Budget"
  }
...
}
```

  or

  ```
{
...
  "title": {
    "es": "Presupuesto Anual"
  }
...
}
```

amercader:
@davidread @brew @rossjones @joetsoi @wardi let me know what you think, specially about point 1 as it relates to #2668

wardi:
:+1: and I'm fine with the `_lang` suffix

For a field like frequency I'd rather use a controlled list that stores a single string value with the translations in the schema, like how the scheming 'select' preset works, not store all the languages in every record, so that might not be a great example.

Is resource format a multilingual field? I thought those were file extensions, essentially

[...]

PR #2729, opened 2015-11-12
[#2678] Add core translation get_translated helper

joetsoi:
Change package templates to use helper, need to fix some snippets

### DIFF
--- a/ckan/templates/package/base.html
+++ b/ckan/templates/package/base.html
@@ -10,7 +10,7 @@
   {% if pkg %}
     {% set dataset = h.dataset_display_name(pkg) %}
     {% if pkg.organization %}
-      {% set organization = h.get_translated(pkg.organization, 'title') or h.get_translated(pkg.organization, 'name') %}
+      {% set organization = h.get_translated(pkg.organization, 'title') or pkg.organization.name %}
       <li>{% link_for _('Organizations'), controller='organization', action='index' %}</li>
       <li>{% link_for organization|truncate(30), controller='organization', action='read', id=pkg.organization.name %}</li>
     {% else %}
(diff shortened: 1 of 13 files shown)


ckan/templates/package/confirm_delete.html
```diff
@@ -8,9 +8,10 @@
   <section class="module span6 offset3">
     <div class="module-content">
       {% block form %}
-        <p>{{ _('Are you sure you want to delete dataset - {name}?').format(name=c.pkg_dict.name) }}</p>
+        {% set dataset = h.get_translated(c.pkg_dict, 'name') %}
+        <p>{{ _('Are you sure you want to delete dataset - {name}?').format(name=dataset) }}</p>
         <p class="form-actions">
-          <form id='confirm-dataset-delete-form' action="{% url_for controller='package', action='delete', id=c.pkg_dict.name %}" method="post">
+          <form id='confirm-dataset-delete-form' action="{% url_for controller='package', action='delete', id=dataset %}" method="post">
```
wardi: That's not going to work, is it?

ckan/templates/package/base.html
```diff
@@ -8,9 +8,9 @@
 
 {% block breadcrumb_content %}
   {% if pkg %}
-    {% set dataset = pkg.title or pkg.name %}
+    {% set dataset = h.dataset_display_name(pkg) %}
     {% if pkg.organization %}
-      {% set organization = pkg.organization.title or pkg.organization.name %}
+      {% set organization = h.get_translated(pkg.organization, 'title') or h.get_translated(pkg.organization, 'name') %}
```
wardi: hmm. use h.dataset_display_name again? and maybe rename it to h.translated_display_name?

wardi: or, `h.get_translated` could accept fall-back items to try, then we can use `h.get_translated(pkg.organization, 'title', 'name')`

PR #2730, opened 2015-11-12
SQL 'execute' audit

davidread:
I grepped for all SQL execute statements and where possible I changed them so params are passed to execute, rather than substituted into the string beforehand (for security reasons). Pleased to see there are only a couple of very minor things that could be improved.

### DIFF
--- a/ckanext/datastore/tests/test_create.py
+++ b/ckanext/datastore/tests/test_create.py
@@ -505,9 +505,9 @@ class TestDatastoreCreate(tests.WsgiAppCase):
 
             assert results == results_alias
 
-            sql = (u"select * from _table_metadata "
-                "where alias_of='{0}' and name='{1}'").format(resource.id, alias)
-            results = c.execute(sql)
+            sql = u"select * from _table_metadata " \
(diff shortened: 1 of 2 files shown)


ckanext/datastore/tests/test_create.py
```diff
@@ -617,13 +617,13 @@ def test_create_basic(self):
...
-                "where alias_of='{0}' and name='{1}'").format(resource.id, alias)
-            results = c.execute(sql)
+            sql = u"select * from _table_metadata " \
+                   "where alias_of=%s and name=%s"
+            results = c.execute(sql, resource.id, alias)
             assert results.rowcount == 0
 
         sql = (u"select * from _table_metadata "
-            "where alias_of='{0}' and name='{1}'").format(resource.id, 'another_alias')
+               "where alias_of='{0}' and name='{1}'").format(resource.id, 'another_alias')
```
morty: Did you miss this one?

davidread: Yes! thanks, I'll do that too

PR #2733, opened 2015-11-15
snippet for automatic connection recovery after invalidation

earthquakesan:
Prevents user to see 500 error on database restart.

In our use case we have a firewall inside of network, which is not really accessible by our CKAN management team (read unforeseen forces, which mess the network :)) Firewall discards connections to database after some time. The result is that user sees 500 error, lowering usability and leading to frustration of the user. This connection_ping function fixes the issue.

The code is taken from: http://docs.sqlalchemy.org/en/latest/core/pooling.html
I put it into model package, not sure if it is appropriate place, but it is related to Pool management.


rossjones:
@earthquakesan Thanks for the PR! It was discussed during the team meeting, and I think the general concensus was that the solution to your problem might actually be pgbouncer (which will use keepalive to stop your connection being killed), and not to add this to core.  On further reflection though I'm torn because this does seem like a sensible thing to do (reconnect when a query fails on pool checkout).

I hope you don't mind me not merging this straight away, I want to think about it a little more and discuss at the next team meeting again.  

Plus, a PEP8 error causing the tests to fail - I'm glad I am not the only one that gets these :)

wardi:
My concern about this change is it adds an extra round trip to the DB every time a connection is checked out. Won't this affect performance for the site as a whole?

earthquakesan:
@wardi yes, indeed. However, in our case usability is more important than overhead for a quick connection check.
The overhead is running ```SELECT 1``` query one time in case if everything is good and two times in case when connection was invalidated.

The better solution would be to listen to "invalidate" event and fire additional select queries when it occurs: http://docs.sqlalchemy.org/en/latest/core/events.html#sqlalchemy.events.PoolEvents.invalidate

This would be more precise solution, I checked the event (it fires indeed when DB is restarted), but didn't manage to recover from it. Then I found the second solution and just used it. Combining both of them would be better approach.

This is pseudocode (in sense, that I didn't try to run it yet), worth trying out. Basically, event should be switched from "checkout" to "invalidate".
```
from sqlalchemy import event
from sqlalchemy import exc
from sqlalchemy.pool import Pool

# standard decorator style
@event.listens_for(Pool, 'invalidate')
def receive_invalidate(dbapi_connection, connection_record, exception):
    "listen for the 'invalidate' event"
cursor = dbapi_connection.cursor()
    try:
        cursor.execute("SELECT 1")
    except:
        # raise DisconnectionError - pool will try
        # connecting again up to three times before raising.
        raise exc.DisconnectionError()
    cursor.close()
```

earthquakesan:
The snippet I proposed above does not work as I thought. dbapi_connection is closed by the moment of "invalidate" event occurs. The solution using "invalidate" event probably requires using connection_record instance to recreate connection, however my knowledge is not good enough atm to implement it.

P.S. Performance should definitely be checked. Are there any automated benchmarks?

rossjones:
@earthquakesan it's been a while since there was any activity, is this something you're still interested in or should the PR be closed?

earthquakesan:
@rossjones I have incorporated these changes into our deployment. If you can give me pointers how to finalize this pull request, I can do that.

There is a valid concern raised by @wardi, this might affect performance in a bad way.

All in all, that was the only deployment where I have encountered this problem. It is ok to close this pull request without merging.

kmbn:

PR #2738, opened 2015-11-17
[#2234] Improve datastore connection checks on migration

amercader:
If the datastore connection is set on the config file but the user or
database don't actually exist, skip the migration.

Note that this only happens when the config options are there but the
datastore plugin is not loaded (like in many extensions that extend
their test.ini with the core test-core.ini). Otherwise the connection
will crash anyway during the loading of the plugin itself.


amercader:
@wardi extra checks for the #2234 migration

PR #2739, opened 2015-11-17
Adds metadata_created as a proper DB field.

### LINKED ISSUE #2735: package.metadata_created() should be a field
Currently metadata_created is a property that calls the metadata_created function - which in turn queries PackageRevision to find the earliest one.  This means that modifying it becomes a chore (for instance, when harvesting).

This should be turned into a column on the model (similar to metadata_modified).  Will need a migration.

rossjones:
Currently the metadata_created is calculated on each access by querying
the package_revision table and finding the oldest revision (it is a property 
function).  There's little advantage to this over setting the field to be a 
column on the package table and defaulting to now().  This means it can 
be overwritten by those components, like harvesting, that require the ability 
to do so- without meddling with revisions.

This should fix #2735

### DIFF
--- a/ckan/migration/versions/082_add_metadata_created.py
+++ b/ckan/migration/versions/082_add_metadata_created.py
@@ -1,0 +1,15 @@
+def upgrade(migrate_engine):
+    migrate_engine.execute('''
+        ALTER TABLE package_revision
+            ADD COLUMN metadata_created timestamp without time zone;
+        ALTER TABLE package
+            ADD COLUMN metadata_created timestamp without time zone
+            DEFAULT now();
+
+        UPDATE package SET metadata_created=
+            (SELECT revision_timestamp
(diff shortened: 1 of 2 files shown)


ckan/migration/versions/082_add_metadata_created.py
```diff
@@ -0,0 +1,15 @@
+def upgrade(migrate_engine):
+    migrate_engine.execute('''
+        ALTER TABLE package_revision
+            ADD COLUMN metadata_created timestamp without time zone;
+        ALTER TABLE package
+            ADD COLUMN metadata_created timestamp without time zone
+            DEFAULT now();
```
wardi: Let these be null in the ALTER TABLE then postgres won't have to walk the whole package table on this statement only to overwrite it in the update below.

PR #2742, opened 2015-11-17
Remove old RDF rendering of dataset page

### LINKED ISSUE #2741: Remove old RDF cruft.
The old template-style RDF templates are no longer present, which means there is a chunk of code (around simple conneg) that can be removed from ckan/lib/accept.py and from the read() method of the package controller.  Currently it is running code that is never used on each request.

Make sure it is obvious that users should install ckanext-dcat instead.

rossjones:
As ckanext-dcat is now a thing, and the old templates are removed, we no
longer need to support .rdf extensions or conneg - and so this PR
removes them.

This should fix #2741


davidread:
Just checking with @amercader that this removal ties in ok with the work on ckanext-dcat?

Issue #1868, opened 2014-08-04
Organization custom fields removing bug


razz0:
An organization can have custom fields and in the organization admin page there is a "Remove" checkbox next to each custom field. However the checkbox often doesn't seem to do what's expected.

For example when only the uppermost "Remove" is checked, and you are not adding more custom fields, the field doesn't get removed. If you add a new custom field at the same time, it is removed. Also when you have checked "Remove" for all your custom fields, nothing gets removed. With some combinations of checked and unchecked checkboxes they do function as expected.

The problem exists atleast in CKAN versions 2.1.2 and 2.2 (demo.ckan.org).

rossjones:
This PR just looks like it needs test fixing to account for the change.

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

PR #2748, opened 2015-11-22
Fixing issue with proper deletion/creation of extra fields for organizations

earthquakesan:
https://github.com/ckan/ckan/issues/1868

Remove button now puts empty strings to key/value fields
Extras with empty fields are delete from data_dict
Remove button is rendered with additional macros not to break any existing code

### DIFF
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -420,6 +420,10 @@ def group_dict_save(group_dict, context, prevent_packages_update=False):
         for key in new_extras:
             group.extras[key] = extras[key]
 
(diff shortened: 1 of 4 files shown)


wardi:
@earthquakesan looks like this has caused a number of test failures. Do you have time to investigate?

earthquakesan:
@wardi Aha, I didn't run tests before the pull request. Is there a doc with guidelines for testing?

wardi:
@earthquakesan there is http://docs.ckan.org/en/latest/contributing/test.html

rossjones:
If there's no activity on this PR in the next 2 weeks, suggest we close it.

PR #2750, opened 2015-11-23
Quieten activity logs

davidread:
This logging clouds out other much more useful debug logging.

When working on this feature you can enable the logging again easily, temporarily.


davidread:
Decided to just delete the log statements instead

PR #2554, opened 2015-08-19
group_list performance issues: timeouts on homepage

amercader:
Problems:

1. `group_list` / `organization_list` are calling `group_show` / `orgsanization_show` for all items even when you are just requesting ids/names (the default behaviour)

2. The widgets on the homepage and their associated helpers are calling `group_list` / `organization_list` when they only need to show 1 of each.

This basically means timeouts on the homepage and pages that call `group_list` or `organization_list` on instances that have many groups / orgs (like demo.ckan.org).

demo.ckan.org was running 2.2.x and worked fine, so something along the way went wrong.

PR #2752, opened 2015-11-24
user_list performance enhancement

amercader:
Much is the same vein to other actions like group_list (#2554) this endpoint needs to be improved.

Context: Moving datahub.io to a new platform setup, DB servers maxing out CPU no matter how many resources thrown at it, with queries averaging 20 minutes to run. Datahub had spam problems in the past so it has around 300k users (yes, this neds to be cleaned up at some point). Problems were pinpointed to `/user` calls.

Culprits: 
* [This query](https://github.com/ckan/ckan/blob/34e1e3f05f52b9ec5b2c2ff79af9822e6d01dd0a/ckan/logic/action/get.py#L900) (here in all its [SQL glory](https://gist.github.com/amercader/29ecaf2a3ecd00215d97)) + 
* Bad indexes on `package.creator_owner_id` +
* Requiring number of revisions and owned datasets without needing it + 
* Querying for all users regardless of pagination and offset

### DIFF
--- a/ckan/migration/versions/082_create_index_creator_user_id.py
+++ b/ckan/migration/versions/082_create_index_creator_user_id.py
@@ -1,0 +1,7 @@
+def upgrade(migrate_engine):
+    migrate_engine.execute(
+        '''
+        CREATE INDEX idx_package_creator_user_id ON package
+        USING BTREE (creator_user_id);
+        '''
+    )


amercader:
The rest of the action/controller is painfully inefficient, but at least the index will make things bearable. On Datahub `/user` goes from 20 *minutes* + to load to 5-6 seconds

davidread:
I tried it on 18k users locally. Just running that query in package_list on its own drops from 234 to 2 seconds with this PR.

PR #2766, opened 2015-12-01
Replace markdown library with python-markdown

rossjones:
Replaces the previous markdown renderer that was problematic (and
unmaintained) with the more widely used python-markdown.

As python-markdown has deprecated the safe_mode for removing HTML, this
PR also contains bleach to strip out any HTML supplied to the render_markdown 
calls where allow_html is False - as recommended by the python-markdown docs.

The only noticeable change in rendering is that the previous library inserted 
a newline before a closing p tag, which python-markdown doesn't do.  This 
is unlikely to change how the content is rendered, although it does require 
changes to the tests in several places (to remove the expected newline).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,12 +7,14 @@
 argparse==1.4.0           # via ofs
 babel==0.9.6
 beaker==1.7.0
+bleach==1.4.2
 decorator==4.0.4          # via pylons, sqlalchemy-migrate
 fanstatic==0.12
 formencode==1.3.0         # via pylons
 Genshi==0.6
 Jinja2==2.6
 mako==1.0.2               # via pylons
+Markdown==2.1
 markupsafe==0.23          # via mako, webhelpers
 nose==1.3.7               # via pylons
 ofs==0.4.1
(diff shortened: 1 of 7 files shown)


rossjones:
Python-markdown doesn't **appear** to support Python 2.6.  This has irrevocably ruined my day.

requirements.txt
```diff
@@ -7,15 +7,18 @@
 argparse==1.4.0           # via ofs
 babel==0.9.6
 beaker==1.7.0
+bleach==1.4.2
```
amercader: These should be added to requirements.in as well

requirements.txt
```diff
@@ -7,15 +7,18 @@
...
 decorator==4.0.4          # via pylons, sqlalchemy-migrate
 fanstatic==0.12
 formencode==1.3.0         # via pylons
 Genshi==0.6
 Jinja2==2.6
 mako==1.0.2               # via pylons
+Markdown==2.1
```
amercader: Why 2.1? AFAICT Markdown 2.4 still supports Python 2.6

rossjones: Claimed to, didn't work for me.  Perhaps i got myself muddled with the missing ordereddict.  Will try 2.4 again, although we should remember to update to 2.6.5 when we drop Py26 support.

PR #2777, opened 2015-12-08
Expose config and translator via toolkit.

### LINKED ISSUE #2776: Add config and translator to plugin toolkit
So that developers don't need to directly import pylons config and translator, let's make it available in the toolkit for them.

rossjones:
To avoid importing of pylons in extensions, we should include the most
commonly imported pylons modules in the plugin toolkit.

This PR does that in order to fix #2776

### DIFF
--- a/ckan/plugins/toolkit.py
+++ b/ckan/plugins/toolkit.py
@@ -20,6 +20,8 @@ class _Toolkit(object):
         '_',                    # i18n translation
         'ungettext',            # i18n translation (plural forms)
         'c',                    # template context
+        'config',               # The pylons configuration
+        'translator',           # Pylons translator, often used in plugins when create commands
         'request',              # http request object
         'render',               # template render function
         'render_text',          # Genshi NewTextTemplate render function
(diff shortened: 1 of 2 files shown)


ckan/plugins/toolkit.py
```diff
@@ -138,6 +140,20 @@ def _initialize(self):
...
 '''
+        t['config'] = pylons.config
+        self.docstring_overrides['config'] = '''The Pylons configuration object.
+
+The configuration object containing information about this request as well as
+options loaded from the .ini file.
+'''
+
+        t['translator'] = pylons.translator
+        self.docstring_overrides['translator'] = '''The Pylons translator object.
```
wardi: What does translator give that we don't get with `_` and `ungettext` already present?

rossjones: I found it in use in an extension, can't remember which, I'll see if I can find it again.  In all likelihood we (I) should probably submit a PR to the extension.

amercader:
Superseded by #3163

PR #2782, opened 2015-12-10
Pin third party requs in requirements.txt, not setup.py

florianm:
Add a section to Docs >  Extensions > Best practices about third party requirements as advocated by @davidread in https://github.com/ckan/ckan/issues/2470


davidread:
Thanks!

PR #2785, opened 2015-12-10
[WIP] Allow extensions to contain migrations

### LINKED ISSUE #2784: Migrations are not available to extensions
It would be good to be able to have migrations run from extensions, so that they don't have to:

a. Have a separate paster command to create the database tables they need
b. Have workarounds when the db schema changes
c. Try and autoload tables at startup time.

Integrating with `db upgrade` appears to be the favoured approach.

rossjones:
As currently managing database tables in extensions is rather painful we
end up with developers either checking and creating the tables at
startup time, or creating paster command specific to their extension
which creates tables. Running actual migrations is often problematic.

These changes provide a new interface called IMigration that allows an
extension to specify where it stored its migrations.  These migrations
are then run whenever the `db upgrade` command is executed.

The paster templates for creating a new extension are also updated to
automatically created the required files for running migrations.

This should resolve #2784 but is currently missing:
- [ ] Tests
- [x] Documentation

### DIFF
--- a/ckan/migration/migrate.cfg
+++ b/ckan/migration/migrate.cfg
@@ -5,16 +5,16 @@ repository_id=Ckan
 
 # The name of the database table used to track the schema version.
 # This name shouldn't already be used by your project.
-# If this is changed once a database is under version control, you'll need to 
-# change the table name in each database too. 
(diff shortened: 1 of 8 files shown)


kmbn:

PR #2791, opened 2015-12-16
Update release process docs with deb packaging instructions

amercader:


amercader:
@davidread as agreed, these are the deb packages / S3 instructions

Issue #2799, opened 2015-12-17
authz.auth_is_anon_user() doesn't work inside tests


vitorbaptista:
A user is considered anonymous if it's either falsy or an IP address (https://github.com/ckan/ckan/blob/master/ckan/authz.py#L426-L442). The problem is that `BaseController` sets `c.author` to be either the `c.user` or her IP address, and when there's no IP address (as when the requests comes from testing), it is set to the default `Unknown IP Address` string (https://github.com/ckan/ckan/blob/master/ckan/lib/base.py#L260-L263).

In that case, `authz.auth_is_anon_user()` will return false, as if the user isn't anonymous. This creates issues in (for example) `package_create` (https://github.com/ckan/ckan/blob/master/ckan/logic/auth/create.py#L7-L37).

vitorbaptista:
The test to check if the user is an IP address is also quite fragile: it simply checks if there's a `.` (dot) in the username.

wardi:
even better would be to never store an IP address in this author field.

vitorbaptista:
@wardi Definitely. While that doesn't happen, I've submitted a PR to fix this and also add tests to this method (https://github.com/ckan/ckan/pull/2801)

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #2801, opened 2015-12-17
[#2799] Fix authz.auth_is_anon_user() when couldn't get the user's IP

vitorbaptista:
Also improve the method for checking if the username is an IP.

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -430,12 +430,15 @@ def auth_is_anon_user(context):
 
         See ckan/lib/base.py:232 for pylons context object logic
     '''
+    is_ip = lambda x: re.match('\d+\.\d+\.\d+\.\d+', x)
+    is_unknown_ip = lambda x: x == 'Unknown IP Address'
+
     context_user = context.get('user')
     # FIXME: our current pattern is to set context['user'] to
     # the IP address in our controller code. Detect and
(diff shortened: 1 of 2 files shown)


ckan/authz.py
```diff
@@ -430,12 +430,15 @@ def auth_is_anon_user(context):
 
         See ckan/lib/base.py:232 for pylons context object logic
     '''
+    is_ip = lambda x: re.match('\d+\.\d+\.\d+\.\d+', x)
+    is_unknown_ip = lambda x: x == 'Unknown IP Address'
```
davidread: This magic value comes from https://github.com/ckan/ckan/blob/master/ckan/lib/base.py#L261 so I'd encourage you to reference it in a comment.

davidread:
Otherwise, this looks great.

wardi:
I feel strongly that we should remove the madness of mixing IP addresses and user names in the same variable instead of further cementing this behaviour into core.

wardi:
I'll work on a PR that replaces all the `'user': c.user or c.author` code with `'user': c.user` in our controllers

davidread:
Good plan. The IP is still always available in the request object if we should ever want it.

amercader:
This is probably superseded by #2817

amercader:
No longer relevant after #2817

PR #2817, opened 2015-12-31
make context['user'] always username or None

wardi:
This removes the need to detect IP addresses and other non-username strings in `context['user']`. This is important because this value is used to determine authorization to perform actions.

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -431,12 +431,6 @@ def auth_is_anon_user(context):
         See ckan/lib/base.py:232 for pylons context object logic
     '''
     context_user = context.get('user')
-    # FIXME: our current pattern is to set context['user'] to
-    # the IP address in our controller code. Detect and
-    # ignore that case for now. Stop putting the IP address
-    # in context['user'] in a future ckan version.
-    if context_user and '.' in context_user:
-        context_user = None
     is_anon_user = not bool(context_user)
 
     return is_anon_user
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -52,7 +52,7 @@ class ApiController(base.BaseController):
 
         self._identify_user()
         try:
-            context = {'model': model, 'user': c.user or c.author,
+            context = {'model': model, 'user': c.user,
                        'auth_user_obj': c.userobj}
             logic.check_access('site_read', context)
         except NotAuthorized:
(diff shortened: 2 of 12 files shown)


wardi:
@vitorbaptista does this change fix the tests for you, like #2799 ?

davidread:
:+1:

vitorbaptista:
@wardi Unfortunately it doesn't. The test is simply:

```python
    @helpers.change_config('ckan.auth.anon_create_dataset', False)
    def test_new_requires_user_to_be_able_to_create_packages(self):
        url = toolkit.url_for('import_datapackage')
        response = self.app.get(url)
        assert_equals(302, response.status_int)
```

Where the `import_datapackage` page requires that the user is able to `package_create`. Even though the user is `u'Unknown IP Address'` at that moment (using your branch), `toolkit.check_access('package_create', context)` still returns `True`. The issue appears to be on:

```python
# ckan/authz.py
def auth_is_anon_user(context):
    ''' Is this an anonymous user?
        eg Not logged in if a web request and not user defined in context
        if logic functions called directly

    '''
    context_user = context.get('user')  # context_user == u"Unknown IP Address"
    is_anon_user = not bool(context_user)  # is_anon_user == False (should be True)

    return is_anon_user
```

wardi:
Hmm. I wonder how we're getting "Unknown IP Address" in context['user'].  @vitorbaptista  Is the import_datapackage controller code putting `'user': c.user or c.author` in the context?

vitorbaptista:
@wardi the "Unknown IP Address" comes from https://github.com/ckan/ckan/blob/master/ckan/lib/base.py#L261

wardi:
@vitorbaptista I see. That's being stored in c.remote_addr, I'm trying to figure out how that value ends up in context['user']. The code I changed is trying to prevent it ending up there via c.author

wardi:
@vitorbaptista I think this could be your problem https://github.com/ckan/ckanext-datapackager/blob/master/ckanext/datapackager/controllers/datapackage.py#L38
Try dropping the `or toolkit.c.author` part.

vitorbaptista:
@wardi you're right! Removing the `toolkit.c.author` fixed the tests :+1:

I don't know the implications of this PR but, as far as I'm concerned, it's good to go.

Issue #2820, opened 2016-01-05
Docker install just links to datacats


davidread:
Rather than supply 'official' ckan docker images and install instructions, we should simply link to datacats one as it works and is maintained.

There are serious issues with ckan's documented docker install (e.g. #2255) due to the images being long unmaintained.

The conclusion of the discussion:
https://github.com/ckan/ckan/issues/2255#issuecomment-146168103
and subsequent discussion at the tech team meeting was to simply guide people towards the datacats docker install.

davidread:
@gamesbook [said](https://github.com/ckan/ckan/issues/2255#issuecomment-168991411):
> @davidread Thanks! Unfortunately, the datacats approach does not work either.

They've had a great track record looking after user issues. However progress has slowed recently - we'll see how your issue fares - https://github.com/datacats/datacats/issues/372

gamesbook:
@davidread It could be support is "down" over the year end period.  I think this might be a Docker issue as that project is moving pretty fast.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #2821, opened 2016-01-05
Docker install archived

davidread:
Fixes issue #2820


wardi:
@davidread there's no longer a company supporting datacats. I still use it every day, but I don't have time to keep up with docker API breakage, OSX support etc. Let's just say the existing docker images are not maintained and docker installation is not supported by us for now.

davidread:
Ok, how's this?

dersteppenwolf:
Big trouble. 
Probably datacats it is the only way for new users to run ckan. 
(BTW, awesome work...)

And looking to the official docs...http://docs.ckan.org/en/latest/maintaining/installing/index.html?highlight=docker
> There used to be an ‘official’ Docker install of CKAN. There are legacy docker images at https://hub.docker.com/u/ckan/ which are not maintained ...

Therefore... we are on our own... 
I tried to build the  images using datacats https://github.com/datacats/datacats/blob/master/docker/build_images.sh  but it fails.
 the "latest" (ckan 2.5?)  tag doesn't work.

Looks like a perfect time for a fork.

davidread:
> Probably datacats it is the only way for new users to run ckan. 

Nope, there are excellent instructions for source install and also ubuntu package.

> Looks like a perfect time for a fork [of datacats]

Yeah the norm is for each developer to create a personal fork, do each fix on a branch and create Pull Requests to get it merged into the main-line for everyone's benefit.

PR #2828, opened 2016-01-11
Disable forced reindexing on group/organization.

TkTech:
Basically reverting commit dbe25d8b8d7fa.

Currently, if a group or organization changes, all datasets belonging to that group/organization are reindexed _in the webserver process_. If you have many datasets in your group/organization, this effectively never finishes and blocks indefinitely.

When a proper background queue worker is available, this should be enabled and utilize the queue. For now, it would be best to let sysadmins know they should be scheduling nightly/periodic reindexes.


wardi:
@TkTech as discussed, this is just needs the change to index group ids in solr instead of the names.

kmbn:

PR #2830, opened 2016-01-12
Output JUnit XML output for tests on CircleCI

amercader:
This allows to list failures directly on top of the report so you don't have to scroll through the nose output.

0rrxkjo

Also list them on the emails:

t2faxup


amercader:
@wardi I know you are a CircleCI fan :)

PR #2831, opened 2016-01-12
Show 404 when user has no permission for deleted package

### LINKED ISSUE #2805: Opening a deleted dataset or organization logs you out
Steps to reproduce:

1. Create a dataset as user A
1. Remove the dataset
1. Log out
1. Log in as user B (regular user)
1. Open the removed dataset via address bar
1. You are now logged out at the login screen

Redirecting to login screen is OK since the user does not have sufficient privileges to view the removed dataset, but logging them out is, in my opinion, unnecessary.

rossjones:
When a user currently has no permissions to view it, a deleted package
will send them to the login page and log them out.

This PR will instead, when the user has no permission, send a 404.  

~~I am not 100% happy with this solution (catch the notauthorized and check state) and think it may be better for check_access to throw NotFound when it is a deleted object (although this will probably still end up catching the exception from the auth check).~~

Should fix #2805

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -361,10 +361,8 @@ class PackageController(base.BaseController):
         try:
             c.pkg_dict = get_action('package_show')(context, data_dict)
             c.pkg = context['package']
-        except NotFound:
+        except (NotFound, NotAuthorized):
             abort(404, _('Dataset not found'))
-        except NotAuthorized:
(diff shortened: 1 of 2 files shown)


rossjones:
After discussion we have decided to do this in the PackageController so that APIs always return 403 when appropriate. That way if a user finds a page they don't have access to, they will get a 404.

PR #2834, opened 2016-01-13
Removes all of the code for processing genshi templates

### LINKED ISSUE #2833: Remove Genshi support
This has been deprecated for 2.5 years now, time to remove it.

rossjones:
As genshi was deprecated 2.5 years ago, it's about time we removed it.
This PR removes genshi.

This should fix #2833

### DIFF
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -87,8 +87,7 @@ def render_jinja2(template_name, extra_vars):
 
 
 def render(template_name, extra_vars=None, cache_key=None, cache_type=None,
-           cache_expire=None, method='xhtml',
-           cache_force=None, renderer=None):
+           cache_expire=None, cache_force=None, renderer=None):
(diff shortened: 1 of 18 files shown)


ckan/lib/base.py
```diff
@@ -105,7 +87,7 @@ def render_jinja2(template_name, extra_vars):
 
 
 def render(template_name, extra_vars=None, cache_key=None, cache_type=None,
-           cache_expire=None, method='xhtml', loader_class=MarkupTemplate,
+           cache_expire=None, method='xhtml',
```
amercader: We can probably get rid of `method` as well, as the Jinja `render` function won't use it

amercader:
@rossjones this looks good, only one minor comment. I also added a small commit updating the config options docs.

rossjones:
@amercader Great, let me get rid the the method param as well and should be gtg.

PR #2835, opened 2016-01-14
toolkit: add h, StopOnError, DefaultOrganizationForm

wardi:
These changes will let me remove all deep ckan imports from ckanext-scheming.

For `h` I've chosen to export the same object that the templates see (with all the same side-effects) I think that's better than just exporting `ckan.lib.helpers` but I would kind of like to have proper AttributeErrors when I mistype a helper name.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -40,6 +40,7 @@ class _Helpers(object):
     def __init__(self, helpers):
         self.helpers = helpers
         self._setup()
+        self.no_magic = _HelpersNoMagic(self)
 
     def _setup(self):
         helpers = self.helpers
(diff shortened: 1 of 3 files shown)


wardi:
sadly, this doesn't let me `import ckan.plugins.toolkit.test_helpers as helpers` yet. Need to dig further

wardi:
These changes will let me remove all deep ckan imports from ckanext-scheming

wardi:
This now has proper AttributeErrors for missing helpers on `toolkit.h` (unlike `h` in templates)

As a bonus we can now use `h.no_magic.myhelper` instead of `h.myhelper` in templates if we want an exception when this helper does not exist.

ckan/config/environment.py
```diff
@@ -96,7 +90,29 @@ def __getattr__(self, name):
...
                 self.log.critical(msg)
-            return self.null_function
+            return _null_function
+
+
+def _null_function(*args, **kw):
+    ''' This function is returned if no helper is found. The idea is
+    to try to allow templates to be rendered even if helpers are
+    missing.  Returning the empty string seems to work well.'''
+    return ''
```
wardi: @brew I moved this out because there's no reason for it to be a `classmethod` (it never uses `cls`), and because as a normal function I can do the `is _null_function` check below

PR #2836, opened 2016-01-14
Update datastore.rst

hvwaldow:
I believe this was a typo.


amercader:
Thanks @hvwaldow, I've backported the fix in 2dfa2a4

PR #2840, opened 2016-01-19
Added the SVG version of the logo

LaurentGoderre:


wardi:
@amercader is this the latest version of the logo? It looks different than the one I see on ckan.org and default ckan sites.

LaurentGoderre:
How is it different?

wardi:
in this one compared to the one on ckan.org the lettering extends above the graphic on the left, there is a black outline around the large triangle and black text is used.

The logo I see on default ckan skins is monochrome white but otherwise just like the one on ckan.org.

Where are you using this logo in your design?

LaurentGoderre:
Through CSS I modify it to the white treatment and it gives this (the top one):

http://laurentgoderre.github.io/ckan-bootstrap-static/

LaurentGoderre:
Hmmm, I see what you saying about the alignment.....maybe there was an issue with the eps and how it imported in Inkscape

LaurentGoderre:
Ckan.org is not even the actual logo because it's made out of real text and it's not even using the right font

LaurentGoderre:
That is the version that I got from the EPS https://upload.wikimedia.org/wikipedia/en/e/e2/CKAN_Logo_full_color.png

amercader:
@smth is the mastermind behind this and worldwide expert on CKAN logos. I'll defer to his judgement.

smth:
That logo looks correct to me, assuming it is going on a white background.

wardi:
@smth We need one for a coloured background http://laurentgoderre.github.io/ckan-bootstrap-static/

smth:
OK, I just put one here https://bitbucket.org/okfn/assets.okfn.org/src/a6617cc2c34b3d1c9f291ab1a408d46aa72d9565/p/ckan/img/ckan-logo-white.svg?at=master&fileviewer=file-view-default

LaurentGoderre:
You don't need this though.....it's SVG. I am using the same svg for the white one and making it white with CSS alone. That is the beautify of SVG.

LaurentGoderre:
Here is a pen to demonstrate it, both using the same SVG

http://codepen.io/anon/pen/YwaPpm

smth:
Indeed. Though if you *only* need a single colour version, you might be able to optimise the SVG code further. (Looks like your demo is failing in Chrome)

LaurentGoderre:
Yeah....I just found that out....in theory this should be the ideal solution, it's how SVG was intended but I guess we can rely on it until browsers fix this bug (only Firefox has it right).

And when it comes to optimizing the single color, treament, I think the difference is negligeable when gzip is turned on the http connection.

LaurentGoderre:
I added the white treatment one, the difference between them (gzipped) is 31 bytes

LaurentGoderre:
This article is great!! http://tympanus.net/codrops/2015/07/16/styling-svg-use-content-css/

LaurentGoderre:
Well nevermind, Chrome is still bugged as hell when it comes to SVG styling.

ckan/public/base/images/ckan-logo-white.svg
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg height="100%" version="1.1" width="100%" xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1757.17 578.39502" id="ckan-logo">
+    <style>
+        * {fill:#fff;}
+    </style>
```
wardi: did you just hack this style in with a text editor?

LaurentGoderre: No, why?

Issue #2842, opened 2016-01-25
Decouple configuration from Pylons


amercader:
The configuration object is a good candidate for decoupling from Pylons. It is used all over the place but at the same time is a relatively simple implementation.

Right now the config object is imported directly from Pylons (`from pylons import config`) almost everywhere, both in CKAN core and extensions (specially since for some mysterious reason config is not part of the toolkit). This is obviously not great if we want to decouple from Pylons. 

Both Pylons and Flask configuration objects are dict-like objects with a few custom methods:

```
Pylons

    - add_template_engine
    - current
    - current_conf
    - defaults
    - init_app
    - load_environment
    - pop_process_config
    - pop_thread_config
    - push_process_config
    - push_thread_config
    - set_defaults

Flask

    - from_envvar
    - from_object
    - from_pyfile
    - root_path
```

Flask configuration is accessed via the `config` attribute of the Flash object (the app).

Both Pylons and Flask add internal keys to the configuration object, eg things like `pylons.config['pylons.app_globals']` in Pylons or `app.config['PROPAGATE_EXCEPTIONS']` in Flash. It seems dangerous to try to combine configurations for both apps in a single object in case there might be conflicts

Proposed implementation:

* There will be a dict-like CKAN config object available, which will be imported via:

    - Core: `from ckan.common import config`
    - Extensions: `toolkit.config` (if needed we can also add it to `ckantoolkit`)

* This object will be accessed in exactly the same way as the current one, both for setting, getting and deleting keys, so current code should work if the above imports are updated.

* When setting or deleting a key, the CKAN config object will take of proxying the changes internally to the Pylons and Flask config objects in case they are needed at that level.
* If really needed, users can access the actual Pylons or Flask configuration objects via `config.pylons_config` or `config.flask_config`.

amercader:
Proof of Concept: #2843

amercader:
See #3163

PR #2843, opened 2016-01-25
[POC] Decoupling config from Pylons

amercader:
See #2842


amercader:
Closed in favour of #3163

PR #2844, opened 2016-01-25
make resource urls optional

wardi:
When uploading a file or when creating a datastore resource, CKAN requires a dummy value entered in to the url field simply to pass validation.

Some users may want to create resources that aren't accessible publicly yet, again requiring some kind of dummy value to be entered.

### DIFF
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -113,7 +113,7 @@ def resource_dictize(res, context):
                                     resource_id=res.id,
                                     filename=cleaned_name,
                                     qualified=True)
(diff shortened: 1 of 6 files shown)


wardi:
This still needs a review of the front end to warn users about creating resources without urls. (Maybe allow creation of datasets without resources with similar language? might be another issue for that)

I also need to add tests to see that the API behaves in a sensible way: returning an empty string when url not provided. Not prefixing an empty string with 'http://' etc.

wardi:
@amercader @yingqiaogit is working on a test for this so hold off reviewing just yet

wardi:
@amercader added the replacement test

PR #2846, opened 2016-01-26
Use 403 when actions are forbidden, not 401 

wardi:
401 is for when a user can't authenticate. All our current 401 errors are causing users to be logged out by repoze (which is reasonable). Instead we want to say "yes, I know who you are, you just aren't allowed to access this page"

### DIFF
--- a/ckan/config/middleware.py
+++ b/ckan/config/middleware.py
@@ -108,12 +108,12 @@ def make_app(conf, full_stack=True, static_files=True, **app_conf):
         # Handle Python exceptions
         app = ErrorHandler(app, conf, **config['pylons.errorware'])
 
-        # Display error documents for 401, 403, 404 status codes (and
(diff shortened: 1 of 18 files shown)


wardi:
This change won't fix extensions that have copied ckan controller code and still use `abort(401)` Also, this might break extensions that rely on getting 401 errors sent to IAuthenticator plugin abort() methods.

It's not clear to me what the IAuthenticator abort() method it intended for. Does it allow a redirect or some other action? If so, because a number of _show calls for unauthorized users are now 404's instead of 401's that could cause more breakage

bzar:
This PR seems to fix any issues I've had regarding spontaneous logouts.

wardi:
@nigelbabu thanks! I'll look into the test failures, possibly some new tests added since I created this PR

PR #2848, opened 2016-01-27
Param use_default_schema does not work on package_search

amercader:
#1327 made both `package_search` and `package_show` return the validated data_dict instead of the default one. In both cases passing `use_default_schema=True` should allow to retrieve the unvalidated data_dict, but on `package_search` is returning an exception:

```
    response = fn(context, data)
  File "/home/adria/dev/pyenvs/flask/src/ckan/ckan/logic/__init__.py", line 416, in wrapped
  File "/home/adria/dev/pyenvs/flask/src/ckan/ckan/logic/action/get.py", line 1852, in package_search
    query.run(data_dict)
  File "/home/adria/dev/pyenvs/flask/src/ckan/ckan/lib/search/query.py", line 306, in run
    raise SearchQueryError("Invalid search parameters: %s" % invalid_params)
SearchQueryError: Invalid search parameters: ['use_default_schema']
```

That's because the param is not removed before being sent to Solr

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1713,6 +1713,10 @@ def package_search(context, data_dict):
         sysadmin will be returned all draft datasets. Optional, the default is
         ``False``.
     :type include_drafts: boolean
+    :param use_default_schema: use default package schema instead of
+        a custom schema defined with an IDatasetForm plugin (default: False)
+    :type use_default_schema: bool
+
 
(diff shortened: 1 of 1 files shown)


wardi:
This looks obviously correct, but should we add a little test to make sure it stays working?

wardi:
@amercader Ok I'll let the tests slide this time because it's a fix we need to backport...

Issue #2841, opened 2016-01-24
Encode EXPLAIN SQL string before sending to datastore


drmalex07:
Hello CKAN team.

Shouldn't the SQL unicode string (the EXPLAIN query) be properly encoded before executing it as a raw SQL query in the postresql backend ? (see https://github.com/ckan/ckan/blob/master/ckanext/datastore/helpers.py#L82). I suggest something like:

```
'EXPLAIN (FORMAT JSON) {0}'.format(sql.encode("utf-8"))).fetchone()
```

We have encountered `UnicodeDecodeError` exceptions in our production deployment at `www.geodata.gov.gr` from requests on the SQL-friendly datastore API call (containing non-ascii characters).

Αn example of a CSV resource (ingested by datastorer) can be found at:
http://labs.geodata.gov.gr/en/dataset/simeia-polisis/resource/fd798f29-44a4-4ea9-a1ef-ed5450e9a412

The corresponding failing request (500) is:
http://labs.geodata.gov.gr/api/action/datastore_search_sql?sql=SELECT%20*%20from%20%22fd798f29-44a4-4ea9-a1ef-ed5450e9a412%22%20WHERE%20%22%CE%A0%CE%95%CE%A1%CE%99%CE%9F%CE%A7%CE%97%22%20LIKE%20%27jones%27

amercader:
@drmalex07 This sounds really sensible, do you want to submit a PR with the fix and a small test case here: https://github.com/ckan/ckan/blob/master/ckanext/datastore/tests/test_helpers.py#L90

drmalex07:
@amercader Yes, of course. 
I suppose i should create a table with a column named with non-ascii characters (at  https://github.com/ckan/ckan/blob/master/ckanext/datastore/tests/test_helpers.py#L78), so i will also need a proper Python header for the entire file (`# -*- encoding utf-8 -*-`). Is this ok?

amercader:
Yes, sounds good!

drmalex07:
@amercader I think this is ready

amercader:
Fixed in #2849

PR #2849, opened 2016-01-27
Encode EXPLAIN SQL before sending to datastore

drmalex07:
This PR is related to #2841


amercader:
Thanks for the patch @drmalex07

Issue #2832, opened 2016-01-13
Incorrect resource url when CKAN is in a subpath


AdrianMBarrera:
Dear all,

We have a CKAN 2.5.1 instance which is installed in a subpath.
As stated here: http://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-site-url
our ckan.site_url reflects that ( ```ckan.site_url = http://taro-des.fg.ull.es/datos``` ).

We are facing a problem with the url of the resources that happens when SOLR indexes the dataset. For example:

- original url: http://taro-des.fg.ull.es/datos/dataset/0faa5f08-dd15-41f7-a51e-8648174c35de/resource/0c475489-2251-4564-8d4c-2485ac3462d9/download/areas.txt

- indexed url: http://taro-des.fg.ull.es/dataset/0faa5f08-dd15-41f7-a51e-8648174c35de/resource/0c475489-2251-4564-8d4c-2485ac3462d9/download/areas.txt

To us it seems that url_for is not returning the correct url of the resource in resource_dictize function in model_dictize.py, so is causing this problem. 

We have checked the database and there the urls of the resources are stored ok.

Any help would be appreciated. Thanks a lot.

wardi:
The docs seem to be wrong here :-( Your subpath should be specified in http://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-root-path not site_url

AdrianMBarrera:
Thank you @wardi 

We have set the ckan.root_path setting but it results in: 
```TypeError: object of type 'Locale' has no len()```

See the full stack-trace below:

```
[Thu Jan 14 09:30:41.288960 2016] [:error] [pid 14659:tid 140120053233408] [client 127.0.0.1:38713] mod_wsgi (pid=14659): Exception occurred processing WSGI script '/etc/ckan/default/apache.wsgi'.
[Thu Jan 14 09:30:41.289020 2016] [:error] [pid 14659:tid 140120053233408] [client 127.0.0.1:38713] Traceback (most recent call last):
[Thu Jan 14 09:30:41.289057 2016] [:error] [pid 14659:tid 140120053233408] [client 127.0.0.1:38713]   File "/usr/lib/ckan/default/lib/python2.7/site-packages/paste/cascade.py", line 130, in __call__
[Thu Jan 14 09:30:41.289111 2016] [:error] [pid 14659:tid 140120053233408] [client 127.0.0.1:38713]     return self.apps[-1](environ, start_response)
[...]

PR #2851, opened 2016-01-27
fix ckan.root_path

marcfor:
The language selector is passing a babel Locale object but
_add_i18n_to_url expects a str.


marcfor:
Should fix this issue: https://github.com/ckan/ckan/issues/2832

Issue #2853, opened 2016-01-27
FQ params are incorrectly wrapped in list when passed to Solr.


rossjones:
Just before 'fq' parameters are passed to Solr, they are wrapped in a list. This will break any usage of tagged fq parameters (such as {!tag name}name:something) as they don't like being sent as a single long string inside a list.

Solr is happy to either get a string for fq, with all parameters in, or a list where each item is a single fq term.  Currently fq in the code is treated as a string and dangerously splits on spaces (which will break with terms that contain a space).  We should change the fq parameter to be a list of filters to apply, but would need special considering for API calls to ensure it doesn't break anything.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #2854, opened 2016-01-27
[WIP] Switch fq from string to list

morty:
For ticket #2853 

WIP because this is a port of the code I did for data.gov.uk but I'm not sure it doesn't break anything. 

Currently only switches fq of package search. Should all the other times fq is sent to Solr be converted too?

### DIFF
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -571,7 +571,7 @@ class ApiController(base.BaseController):
                     # public
                     if 'fq' in params:
                         del params['fq']
-                    params['fq'] = '+capacity:public'
+                    params['fq'] = ['+capacity:public']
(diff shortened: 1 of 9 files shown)


davidread:
@morty what's the plan with this now?

morty:
@rossjones is looking at using fq_list for the required purpose that might mean that this refactor wouldn't be required.

davidread:
ok thanks

rossjones:
@morty Any objections to me closing this PR now?  I don't think it's going to get finished this way.

morty:
@rossjones I guess not. There are a number of conflicts as the codebase has moved on. This PR will still be referenced in the issue if someone wants to try again.

PR #2855, opened 2016-01-28
Name field in error message

raphaelstolt:
To avoid quests which field is the one already present the name of the field is attached to the error message.

Please also merge this back into the 2.3 branch.

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -823,7 +823,8 @@ def extra_key_not_in_root_schema(key, data, errors, context):
 
     for schema_key in context.get('schema_keys', []):
         if schema_key == data[key]:
-            raise Invalid(_('There is a schema field with the same name'))
+            error_message = _('There is a schema field with the same name')
+            raise Invalid(error_message + ":field_name=" + schema_key)
 
 
 def empty_if_not_sysadmin(key, data, errors, context):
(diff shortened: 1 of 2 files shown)


ckan/logic/validators.py
```diff
@@ -823,7 +823,8 @@ def extra_key_not_in_root_schema(key, data, errors, context):
 
     for schema_key in context.get('schema_keys', []):
         if schema_key == data[key]:
-            raise Invalid(_('There is a schema field with the same name'))
+            error_message = _('There is a schema field with the same name')
+            raise Invalid(error_message + ":field_name=" + schema_key)
```
wardi: This is a message that will be displayed on the dataset editing form. ":field_name=" looks too much like code. How about something like
```python
raise Invalid(_('Field name {{field_name}} is reserved. Please choose a different name.')
    .format(field_name=schema_key))
```

wardi:
This isn't a bug fix, so it wouldn't normally be back-ported. If you need it for your site I recommend installing ckan from git and cherry-picking the change yourself.

ckan/logic/validators.py
raphaelstolt: Wouldn't that require a change of all related i18n files because of introducing a new language variable?

wardi: Yes, but it's a better message, isn't it?

raphaelstolt: Yes it is.

k-nut:
@wardi should the original change still be merged? It seems like it is "feature complete", right? Or is this still missing anything?

wardi:
No, this doesn't fit with the style of our error messages and it introduces text that can't be overridden by translations

PR #2860, opened 2016-01-31
Fix runtime error when solr does not return a package_dict

tuffnatty:
See also my comment at https://github.com/ckan/ckan/commit/34e1e3f05f52b9ec5b2c2ff79af9822e6d01dd0a#commitcomment-15781370


davidread:
Closed in favour of #2861

PR #2863, opened 2016-02-02
Allow Sysadmins to create users with the form

CarlQLange:
(Even if create_user_from_web is false).

I've come across several workflows where people wanted a centralised administrator to be the **only** person able to create accounts. This change allows that. Everything else works perfectly. I can offer more detail later on - I'm just in a rush at the moment.

Cheers,
Carl

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -166,7 +166,7 @@ class UserController(base.BaseController):
         if context['save'] and not data:
             return self._save_new(context)
 
-        if c.user and not data:
+        if c.user and not data and not authz.is_sysadmin(c.user):
             # #1799 Don't offer the registration form if already logged in
             return render('user/logout_first.html')
 


wardi:
This is good, but would you add a small test so that the behaviour doesn't change accidentally in the future.

Also, redirecting to the user just created would be better than rendering the logout_first page when a sysadmin finishes creating the user.

CarlQLange:
Yep, will do. And I agree, that makes sense. Will add those changes tomorrow. Thanks for getting back to be so fast!

k-nut:
@CarlQLange did you ever get around to doing this? I just created a small test case that makes sure this works. I could provide it unless you already have anything?

CarlQLange:
@k-nut Sorry! I didn't get around to it. Please do!

k-nut:
This is now implemented in PR #3082

PR #2864, opened 2016-02-03
Changed the date formatter to display the timezone abbreviation instead of full name

LaurentGoderre:
Displays `February 3, 20016 (EST)` instead of `February 3, 20016 (America/Toronto)` or `February 3, 20016 (America/New York)`

### DIFF
--- a/ckan/lib/formatters.py
+++ b/ckan/lib/formatters.py
@@ -137,7 +137,7 @@ def localised_nice_date(datetime_, show_date=False, with_hours=False):
         'day': datetime_.day,
         'year': datetime_.year,
         'month': _MONTH_FUNCTIONS[datetime_.month - 1](),
-        'timezone': datetime_.tzinfo.zone,
+        'timezone': datetime_.tzname(),
     }
 
     if with_hours:


ckan/lib/formatters.py
```diff
@@ -137,7 +137,7 @@ def months_between(date1, date2):
         'day': datetime_.day,
         'year': datetime_.year,
         'month': _MONTH_FUNCTIONS[datetime_.month - 1](),
-        'timezone': datetime_.tzinfo.zone,
+        'timezone': datetime_.tzinfo._tzname,
```
amercader: Not saying we can't use it but this looks like an internal property, is it safe to use?

And BTW what happens if the `datetime` object provided does not have a `tzinfo` available? My understanding is that these are not set by default (eg if creating a date like `datetime(2016, 2, 4)`). (I know this is out of the scope of this PR)

LaurentGoderre: The issue of dates not having a tzinfo existed before this PR though

amercader: > (I know this is out of the scope of this PR)
I know :)

What about the private property? Is there a public method on datetime / tzinfo that does something similar?

LaurentGoderre: Found a non private member!

PR #2868, opened 2016-02-08
Added a display_timezone helper

LaurentGoderre:


ckan/config/environment.py
```diff
@@ -224,6 +225,16 @@ def update_config():
             'ckan.site_url should be a full URL, including the schema '
             '(http or https)')
 
+    display_timezone = config.get('ckan.display_timezone', '')
+    if display_timezone:
+        if (
+            display_timezone != 'server' and
+            display_timezone not in pytz.all_timezones
+        ):
```
wardi: this should have failed our pep8 tests! how about a more normal-looking indentation?

ckan/config/environment.py
```diff
@@ -224,6 +225,16 @@ def update_config():
...
+            raise RuntimeError(
+                'ckan.display_timezone is empty or not a valid timezone'
```
wardi: `ckan.display_timezone is not 'server' or a valid timezone`

Also, RuntimeError is a poor exception to use here. Why not add a ConfigurationError exception to ckan/exceptions.py and use that instead?

ckan/lib/helpers.py
```diff
@@ -1010,6 +1007,14 @@ def _range(self, regexp_match):
         return re.sub(current_page_span, current_page_link, html)
 
 
+def get_display_timezone():
```
wardi: missing a docstring. what sort of values are returned from this function? Can we rely on this helper returning the time zone the user will see by default (no, so you should warn about that)

Would it be more useful to return the timezone object here instead of the name? That would be more flexible, but there's no other code that uses this method so it's hard to judge.

wardi:
Please add a test for this method so that we can rely on it continuing to work.

ckan/lib/helpers.py
```diff
@@ -1010,6 +1007,14 @@ def _range(self, regexp_match):
...
+    timezone_name = config.get('ckan.display_timezone', '')
+    if timezone_name == 'server':
+        timezone_name = tzlocal.get_localzone().zone
+
+    return timezone_name
+
+
 def render_datetime(datetime_, date_format=None, with_hours=False):
     '''Render a datetime object or timestamp string as a localised date or
     in the requested format.
```
wardi: The UnknownTimeZoneError handling below should be dead code now, please remove it too

ckan/config/environment.py
LaurentGoderre: What is a "normal" looking indendation?

ckan/config/environment.py
LaurentGoderre: Just trying to follow what other similar errors use buit I don't mind adding a new exception

wardi:
Looks like `change_config` is broken. Seems like it doesn't properly clear your setting after the tests run. @TkTech you noticed this before, did you see any promising ways to fix this?

LaurentGoderre:
@wardi change_config worked..'...my helper did not handle getting an empty string so I fixed it. There were other test issues as well

Issue #2870, opened 2016-02-10
User email can be changed without verifying password


Zharktas:
Since user passwords can be changed using email addresses through reset links, changing email address should be verified against current password. 

We will fix this in our environment so will probably supply a PR.

rossjones:
You should be aware of the work mentioned in #2838 as well.  Perhaps it might also get covered there?

Zharktas:
It probably should be also covered in there, but the change was quite simple so I made PR #2872 for it also.

davidread:
@rossjones might you add @Zharktas to this github org so that we can assign him this ticket?

Zharktas:
Fixed in https://github.com/ckan/ckan/pull/2872

PR #2872, opened 2016-02-10
#2870: require password when changing email

Zharktas:


rossjones:
Should be assigned to @smotornyuk once ready.

smotornyuk:
@Zharktas , in case of applying this PR, user become unable to change any information without password verification. Probably, you should check if email had been changed in addition?

Zharktas:
@smotornyuk True, I thought I checked that it would only used with changing email, but apparently not. Will change it to check for modifications of email instead.

smotornyuk:
Thanks, i think that completed. @rossjones , can it be merged now?

rossjones:
@smotornyuk Ideally we'd have a test, but if you or @Zharktas don't have time to do that, I'd rather merge it than have the PR sat around for ever.  What do you think?

Zharktas:
@rossjones added tests for both trying to change email with wrong password and with right password.

rossjones:
Thanks very much for doing that, it is much appreciated.

PR #2880, opened 2016-02-16
[#1372] Make resource name default to file name

smotornyuk:
Currently user have to manually add resource name after or before
uploading/linking resource

After this change name will be automatically filled when resource
uploaded or added link to resource, but only if name has not been
previously modified by user and was empty from the very begining.
i.e. this won't work for resource update

### DIFF
--- a/ckan/public/base/javascript/modules/image-upload.js
+++ b/ckan/public/base/javascript/modules/image-upload.js
@@ -1,6 +1,6 @@
 /* Image Upload
- * 
- */  
+ *
+ */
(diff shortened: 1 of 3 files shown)


amercader:
This looks great @smotornyuk!
Can you add a small JS test here and then it's good to merge:

https://github.com/ckan/ckan/blob/master/ckan/public/base/test/spec/modules/resource-upload-field.spec.js

smotornyuk:
I created new file with few tests. And in future it will be good idea form me or someone else to create tests for  the whole image-upload module

Issue #2871, opened 2016-02-10
Admin urls load empty templates outside of controller


Zharktas:
Since admin urls are mapped behind ckan-admin/someurl, templates are loaded through admin/trash because ckan (or pylons) tries to match folder structure to unmapped url.

For example: 
http://demo.ckan.org/admin/trash 
http://demo.ckan.org/admin/confirm_reset

These won't actually do anything but urls should match to controllers, will provide a PR when I have change to make it.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #2884, opened 2016-02-23
#2871: rename the admin template directory

Zharktas:
Decided to rename the template directory instead of changing the urls, so that the existing urls won't break.


amercader:
@Zharktas This would break any extensions that have customized the admin templates. Other options of the top of my head:

1. Rename the controller from `admin` to ckan-admin` and update the routing accordingly
2. Add a permanent redirect from `/admin/` to `/ckan-admin/`

I'm not sure about the implications of both though, although either sounds like very low impact.

WDYT?

rossjones:
@Zharktas Is this still a thing you're interesting in pursuing, or can we safely close the PR?

Zharktas:
This is still a problem, but it's not that big of issue anyway. At least I don't have time for it right now, especially since this hasn't had any commits in the past two years. If i have time for it the future, I'll create a new PR and I close this one for now.

Issue #2681, opened 2015-10-12
LazyJSON incompatible with latest simplejson


davidread:
LazyJSONObject doesn't appear to work with the new version of simplejson (3.8.0 instead of 3.3.1) :-(. I tracked it down to this bit of code (simplejson/encoder.py, but normally it uses a C version of this code...) that detects that the JSONString isn't actually an int, so does str() and returns -1.

    def _encode_int(value):
        skip_quoting = (
            _int_as_string_bitcount is None
            or
            _int_as_string_bitcount < 1
        )
        if type(value) not in integer_types:
            # See #118, do not trust custom str/repr
            value = int(value)

I've added a test to https://github.com/ckan/ckan/pull/2561 that shows the problem: ckan/tests/lib/test_lazyjson.py

davidread:
I've moved the failing test to a new branch: 2681-lazy-json

wardi:
opened an issue on simplejson to discuss the best approach https://github.com/simplejson/simplejson/issues/125

amercader:
I was [bitten](https://circleci.com/gh/ckan/ckan/1017) by this while working on the Flask migration and took me a while to understand why the API was returning -1 instead of a dict :) 
Basically while upgrading the requirements for supporting flask, wsgiparty etc, simplejson got also [updated](https://github.com/ckan/ckan/commit/3e56b2bc493c3d2abe39b6b6df5a79ef665fe1be#diff-b4ef698db8ca845e5845c4618278f29aL38). If we really need to stick with an old simplejson version for now then we should probably add it to the `requirements.in` file and not just `requirements.txt` as this is a hard requirement.

@wardi what are your feelings on the approach suggested on the simplejson issue? Is it worth pursuing that option or should we think of something on our side?

wardi:
This work is certainly better done as a well designed simplejson improvement. I'm fine with removing LazyJSON and eating the performance hit instead of having to keep the old dependency.

davidread:
@amercader I've done a PR for you

davidread:
@wardi is there something in particular we need from newer versions of simplejson? I quite like good performance...

wardi:
@davidread no, but it's never fun to have to maintain an old fork of an external library. It would take some work to make this thing general enough for simplejson and get it accepted, though.

wardi:
Also, are you sure you're getting a performance improvement from LazyJSON? Many plugins defeat the performance hack because they modify or check metadata fields returned by package_show

[...]

PR #2886, opened 2016-02-25
lazy json test

davidread:
Doesn't fix anything, but it's helpful

### Proposed fixes:

* fixes simplejson version better in requirements
* test for lazy json problem with newer simplejson versions

### Features:

- [ ] includes API changes


wardi:
as mentioned in #2681 I think the better approach is to drop LazyJSON and allow use of a newer version of simplejson.

davidread:
Ok, agreed

PR #2894, opened 2016-03-03
Fix #2893: Issues with extension's Python namespace packages

### LINKED ISSUE #2893: Issues with extension's Python namespace packages
I'm developing an extension for CKAN 2.5.1 and I'm running into problems with Python namespace packages.

The first problem I encountered was a `KeyError` once I enabled my plugin after installing it into CKAN's virtualenv usign `pip`:
```
$ paster serve /etc/ckan/default/production.ini
  File "/usr/lib/ckan/default/lib/python2.7/site.py", line 703, in <module>
    main()
  File "/usr/lib/ckan/default/lib/python2.7/site.py", line 683, in main
    paths_in_sys = addsitepackages(paths_in_sys)
  File "/usr/lib/ckan/default/lib/python2.7/site.py", line 282, in addsitepackages
    addsitedir(sitedir, known_paths)
  File "/usr/lib/ckan/default/lib/python2.7/site.py", line 204, in addsitedir
    addpackage(sitedir, name, known_paths)
  File "/usr/lib/ckan/default/lib/python2.7/site.py", line 173, in addpackage
    exec(line)
  File "<string>", line 1, in <module>
KeyError: 'ckanext'
```
This error has been reported previously ([1](https://github.com/ckan/ckanext-pdfview/issues/17), [2](https://lists.okfn.org/pipermail/ckan-dev/2015-November/009478.html)). It is rather severe because the virtualenv is now broken, for example trying to uninstall the extension via `pip` leads to the same error. The only remedy I found was to completely remove the virtualenv.

As discussed in the previous bug reports, the error is related to Python namespace packages: The CKAN base installation declares `ckanext` to be a namespace package. The error above is caused when an extension is installed which contains a top-level `ckanext` package which is *not* declared as a namespace package.

The solution therefore seems to be to make sure that my plugin declares `ckanext` as a namespace package. I thought that this was already the case, since I had created the extension using the appropriate `paster` command provided by CKAN. However, it turns out that the relevant declarations are missing from the [`setup.py` template used by paster](https://github.com/ckan/ckan/blob/91e41b2e68faa3df5296a632f4862f5a55e69e62/ckan/pastertemplates/template/setup.py_tmpl). I therefore added the following parameter manually:

    namespace_packages=['ckanext', 'ckanext.my_extension']

In addition, the appropriate code has to be added to `ckanext/my_extension/__init__.py`, which by default is empty. The code is the same as the one in `ckanext/__init__.py`.

With these additions I was able to successfully install and uninstall my extension using `pip`. Running `paster serve ...` also works, but only until a request is served:

```
Exception happened during processing of request from ('193.197.165.44', 32898)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/httpserver.py", line 1068, in process_request_in_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/httpserver.py", line 442, in handle
    BaseHTTPRequestHandler.handle(self)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/httpserver.py", line 437, in handle_one_request
    self.wsgi_execute()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/httpserver.py", line 287, in wsgi_execute
    self.wsgi_start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/cascade.py", line 130, in __call__
    return self.apps[-1](environ, start_response)
  File "/usr/lib/ckan/default/src/ckan/ckan/config/middleware.py", line 228, in __call__
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/registry.py", line 379, in __call__
    app_iter = self.application(environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/repoze/who/middleware.py", line 87, in __call__
    app_iter = app(environ, wrapper.wrap_start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/middleware.py", line 214, in __call__
    self.app, new_environ, catch_exc_info=True)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/util.py", line 94, in call_wsgi_application
    app_iter = application(environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/weberror/errormiddleware.py", line 165, in __call__
    return self.application(environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/dec.py", line 147, in __call__
    resp = self.call_func(req, *args, **self.kwargs)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/dec.py", line 208, in call_func
    return self.func(req, *args, **kwargs)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/fanstatic/publisher.py", line 234, in __call__
    return request.get_response(self.app)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/request.py", line 1053, in get_response
    application, catch_exc_info=False)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/request.py", line 1022, in call_application
    app_iter = application(self.environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/dec.py", line 147, in __call__
    resp = self.call_func(req, *args, **self.kwargs)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/dec.py", line 208, in call_func
    return self.func(req, *args, **kwargs)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/fanstatic/injector.py", line 54, in __call__
    response = request.get_response(self.app)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/request.py", line 1053, in get_response
    application, catch_exc_info=False)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/request.py", line 1022, in call_application
    app_iter = application(self.environ, start_response)
  File "/usr/lib/ckan/default/src/ckan/ckan/config/middleware.py", line 389, in inner
    result = application(environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/beaker/middleware.py", line 73, in __call__
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/beaker/middleware.py", line 155, in __call__
    return self.wrap_app(environ, session_start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/routes/middleware.py", line 131, in __call__
    response = self.app(environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/wsgiapp.py", line 125, in __call__
    response = self.dispatch(controller, environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/wsgiapp.py", line 324, in dispatch
    return controller(environ, start_response)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/base.py", line 337, in __call__
    res = WSGIController.__call__(self, environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/controllers/core.py", line 217, in __call__
    response = self._inspect_call(self.__before__)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/controllers/core.py", line 107, in _inspect_call
    result = self._perform_call(func, args)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/pylons/controllers/core.py", line 60, in _perform_call
    return func(**args)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/base.py", line 243, in __before__
    i18n.handle_request(request, c)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/i18n.py", line 145, in handle_request
    if lang in plugin.i18n_locales():
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/plugins.py", line 549, in i18n_locales
    directory = self.i18n_directory()
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/plugins.py", line 540, in i18n_directory
    return os.path.join(os.path.dirname(module.__file__), 'i18n')
AttributeError: 'module' object has no attribute '__file__'
```
This error also seems to be related to namespace packages, because namespace packages, in contrast to regular packages, do not have a `__file__` attribute. However, the `DefaultTranslation` class tries to construct the location of an extension's `i18n` directory using that attribute. The offending lines in [`plugin.py`](https://github.com/ckan/ckan/blob/master/ckan/lib/plugins.py#L542) are

```
'''Change the directory of the *.mo translation files

The default implementation assumes the plugin is
ckanext/myplugin/plugin.py and the translations are stored in
i18n/
'''
# assume plugin is called ckanext.<myplugin>.<...>.PluginClass
extension_module_name = '.'.join(self.__module__.split('.')[0:2])
module = sys.modules[extension_module_name]
return os.path.join(os.path.dirname(module.__file__), 'i18n')
```

It is clear why that fails: By assumption, `self.__module__` is `ckanext.my_extension.plugin`, so taking the first two components yields `ckanext.my_extension`, which has been declared a namespace package and therefore has no `__file__` attribute. The solution is simple: Take the first 3 components (`self.__module__.split('.')[:3]`) so that the `plugin` module is included, which does have a `__file__` attribute (and is in the same directory as `__init__`, so the final path will be correct).

**TL;DR:**
* The templates for extensions should include the necessary declarations and code to make `ckanext` and `ckanext.extension` namespace packages.
* `DefaultTranslation.i18n_directory` has to be fixed to work with namespace packages.

I will prepare a corresponding PR.

torfsen:
Please see #2893 for details.

Tell me if we need tests for this.

### DIFF
--- a/ckan/pastertemplates/template/setup.py_tmpl
+++ b/ckan/pastertemplates/template/setup.py_tmpl
@@ -54,6 +54,7 @@ setup(
     # You can just specify the packages manually here if your project is
     # simple. Or you can use find_packages().
     packages=find_packages(exclude=['contrib', 'docs', 'tests*']),
+    namespace_packages=['ckanext'],
 
     # List run-time dependencies here.  These will be installed by pip when your
     # project is installed. For an analysis of "install_requires" vs pip's


ckan/pastertemplates/template/setup.py_tmpl
```diff
@@ -54,6 +54,7 @@ setup(
     # You can just specify the packages manually here if your project is
     # simple. Or you can use find_packages().
     packages=find_packages(exclude=['contrib', 'docs', 'tests*']),
+    namespace_packages=['ckanext', 'ckanext.{{ project_shortname }}'],
```
wardi: Yes, `ckanext` should be a namespace package, but `ckanext.{{ project_shortname }}` should not.

torfsen: @wardi: I was unsure about that, too. I checked some existing CKAN extensions and in most cases both `ckanext` and the extension's sub-package were declared as namespace packages. For example, [`ckanext.harvest`](https://github.com/ckan/ckanext-harvest/blob/master/setup.py#L19), [`ckanext.geoview`](https://github.com/ckan/ckanext-geoview/blob/master/setup.py#L17), and [`ckanext.dcat`](https://github.com/ckan/ckanext-dcat/blob/master/setup.py#L18) are namespace packages. Of course there are also counter-examples (e.g. [`ckanext.pdfview`](https://github.com/ckan/ckanext-pdfview/blob/master/setup.py#L56)). So I decided to use the more popular approach.

The benefit of registering `ckanext.my_extension` as a namespace package would be that one could distribute sub-extensions for `my_extension`. While I don't think that this is a common use case I also don't see any disadvantages of declaring `ckanext.my_extension` as a namespace package.

In the end I think both approaches are valid and I'd be happy to change it according to your preferences.

wardi: As I see it declaring ckanext.harvest, ckanext.geoview, and ckanext.dcat as namespace package is wrong and should be fixed. They aren't designed to be extended as namespace packages, and not being able to find a single concrete path for those packages (namespace packages exist in many locations) is a problem for static assets that are resolved relative to the extension path.

torfsen: @wardi: OK, that's fine with me. I will update my PR so that only `ckanext` is declared as a namespace package.

torfsen:
As discussed with @wardi I've removed the namespace declaration of `ckanext.my_package`. Hence only `ckanext` is declared as a namespace package.

PR #2896, opened 2016-03-03
Fix #2895: Problems with translations in plugins

### LINKED ISSUE #2895: Problems with translations in plugins
(This came up together with #2893, see there for more details)

I found two issues regarding translations in plugins:

1. `ckan.lib.plugins.DefaultTranslation.i18n_directory` relies on the extension package's `__file__` attribute (`ckanext.my_extension.__file__`). However, many extensions are distributed using namespace packages, and namespace packages don't have a `__file__` attribute.

2. It seems that `.mo` compiled translation files are not automatically picked up when an extension is packaged using `setup.py`.

I will prepare a PR for this.

torfsen:
See #2895 for details.

Tell me if we need tests for this.

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -537,7 +537,7 @@ class DefaultTranslation(object):
         i18n/
         '''
         # assume plugin is called ckanext.<myplugin>.<...>.PluginClass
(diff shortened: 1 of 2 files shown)


amercader:
Thanks for the patch @torfsen

torfsen:
Glad I could help, @amercader!

Issue #2474, opened 2015-06-18
Datastore: internal error when creating indexes


mvi-eea-sk:
When calling ```/api/action/datastore_create``` Internal server error is raised when creating indexes for that resource if there is the same field in parameter primary_key and indexes:

```
{
  ...
  "fields":[{"id": "ID", "type": "int"},...],
  "primary_key":["ID"],
  "indexes":["ID"],
  ...
}
```

Datastore automatically creates indexes for each primary key, so if the same columns/field is in primary_key and indexes parameter, datastore tries to create index twice for the same column and causes sqlalchemy error:
```
Error - <class 'sqlalchemy.exc.ProgrammingError'>: (ProgrammingError) relation "7b400931c05eb6cb3407e2b5e9cf0de69380ca3f" already exists
'CREATE unique INDEX "7b400931c05eb6cb3407e2b5e9cf0de69380ca3f" ON "9c2f2024-5b51-43a1-8528-d4a32b8b9872" ("ID")' {}
...
```
The problem is here: https://github.com/ckan/ckan/blob/ckan-2.3/ckanext/datastore/db.py#L426
It doesn't check if the added primary key is already in the indexes list and adds ```[u'ID']``` list
so ```indexes = [u'ID', [u'ID']]```

This bug is not only in 2.3 but in 2.2.2 too. But in the 2.2.2 version the index names are generated randomly so no Internal error is caused, but index for the same column is created twice.

joetsoi:
could you provide a pull request to fix this?

PR #2902, opened 2016-03-09
[#2474]add-index-check

bwanglzu:
Fixes #
### Proposed fixes:

Fix issue [#2474](https://github.com/ckan/ckan/issues/2474), add primary key check before append index into indexlist
### Features:
- [ ] includes API changes
- [ X] includes bugfix for possible backport

### DIFF
--- a/ckanext/datastore/db.py
+++ b/ckanext/datastore/db.py
@@ -430,7 +430,8 @@ def create_indexes(context, data_dict):
 
     if primary_key is not None:
         _drop_indexes(context, data_dict, True)
-        indexes.append(primary_key)
+        if primary_key not in indexes:
+            indexes.append(primary_key)
 
     for index in indexes:
         if not index:


ckanext/datastore/db.py
```diff
@@ -430,7 +430,8 @@ def create_indexes(context, data_dict):
 
     if primary_key is not None:
         _drop_indexes(context, data_dict, True)
-        indexes.append(primary_key)
+        if primary_key not in indexes:
+            indexes.append(primary_key)
```
wardi: primary key and indexes might be given as strings or lists, so this won't catch the case of the same index being specified in different ways. Maybe move this fix below after the call to `datastore_helpers.get_list(index)`

bwanglzu:
Hi, wardi. I'm a bit confused. 
1. Do you mean that both `primary_key` and `indexes` could be list?
2. If `primary_key` could be list, if I move the fix blow after `datastore_helpers.get_list(index) `, we can not ensure `primary_key` contains no duplicate elements, and we can only ensure current `primary_key` not in variable `index_fields`, rather than the whole possible indexes.

wardi:
What I mean is that you could have: `indexes=['pka,pkb', 'another']` and `primary_key=['pka', 'pkb']` and you'll still end up with a duplicate because the `datastore_helpers.get_list` normalization happens inside the loop below. Better to check for a repeated index inside that loop.

kmbn:

PR #2904, opened 2016-03-10
datastore: Prevent unicode/ascii conversion errors

mordae:
Some error messages were causing the errors due to format conversion of unicode fields into str templates.

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/db.py
+++ b/ckanext/datastore/db.py
@@ -264,12 +264,12 @@ def check_fields(context, fields):
     for field in fields:
         if field.get('type') and not _is_valid_pg_type(context, field['type']):
             raise ValidationError({
-                'fields': ['"{0}" is not a valid field type'.format(
+                'fields': [u'"{0}" is not a valid field type'.format(
                     field['type'])]
             })
         elif not _is_valid_field_name(field['id']):
             raise ValidationError({
-                'fields': ['"{0}" is not a valid field name'.format(
+                'fields': [u'"{0}" is not a valid field name'.format(
                     field['id'])]
             })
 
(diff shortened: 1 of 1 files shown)


ckanext/datastore/db.py
```diff
@@ -123,7 +123,7 @@ def _cache_types(context):
         if 'nested' not in _type_names:
             native_json = _pg_version_is_at_least(connection, '9.2')
 
-            log.info("Create nested type. Native JSON: {0}".format(
+            log.info(u"Create nested type. Native JSON: {0}".format(
```
wardi: How about using repr() for user-supplied values sent to logs. That seems safer.

mordae:
You are right, it's usually not a good idea.

TkTech:
@mordae `{0!r}` is the same as `%r/repr(x)`

mordae:
OK, let's make it simpler. These are just the `ValidationError` formats.

ckanext/datastore/db.py
```diff
@@ -1310,8 +1313,7 @@ def make_private(context, data_dict):
 
 
 def make_public(context, data_dict):
-    log.info('Making resource {0} public'.format(
-        data_dict['resource_id']))
+    log.info('Making resource {resource_id!r} public'.format(**data_dict))
```
wardi: have you been using something other than UUIDs for your resource_ids?

mordae: Nope, just being consistent and quoting all of them. It's `UnicodeText` in the model...

wardi:
@amercader This puts python reprs in our logs now, do you think that's a concern for backporting?

mordae:
As for the `ValidationError` fixes; yes, the unicode-related crashes are easily reproducible. As for the other commit I frankly do not know. Thanks for the merge, by the way. :-)

Issue #2845, opened 2016-01-26
Decouple Pylons-specific routing


brew:
Currently, routing requests to Pylons actions uses the `Routing` library. CKAN mostly adds new route map rules in the `config/routing.py` module. There are points here for extensions to add additional routing rules to the map before and after the main map is defined.

A move to Flask will require that existing extension routes should still work, so any move to Flask's own routing system will need to take this into account.

Pylons routing discovers the controller module and calls it with the specified action method. This partly dictates how the app is structured around controllers. In Flask, there is a little more flexibility, and code structure is left largely up to the developer, and its routing system reflects that.

As we transition from Pylons to Flask, my suggested approach is to route requests using Flask's routing, and if no route is found there, to fallback to the Pylons app. As work progresses more requests will be handled by Flask's routing. Eventually we can replace Pylon's dispatcher with a light-weight WSGI app to handle extensions that use 'Routes' style map connections to call Flask actions.

brew:
Determining which application (Flask or Pylons) to route to is potentially tricky. Especially when considering that extensions can add endpoints that override actions/view_funcs in core. I'd like to avoid maintaining a third routing table just for routing to one app or the other. 

One potentially solution is to ask each app whether it can handle a given url during dispatch. This could be achieved with DispatchingMiddleware that queries each app with a predefined handler and selects the appropriate app to handle the request based on the results. Something like http://wsgi-party.readthedocs.org/en/latest/ maybe able to provide the app-to-app communication needed.

The DispatchingMiddleware would also have to determine what to do if both apps are able to handle a request. For example, we would usually prefer Flask over Pylons, but if the Pylons route has been provided by an extension, we would likely prefer that over a core Flask action. So we would need to know if a route is core, or extension provided.

Order of preference: Flask Extension, Pylons Extension, Flask Core, Pylons Core.

PR #2905, opened 2016-03-10
[#2845] WSGI Dispatcher

amercader:
See #2845 for previous discussion

TL;DR:

A top level middleware named `AskAppDispatcherMiddleware`, implemented using [WSGIParty](http://wsgi-party.readthedocs.org/en/latest/). This wraps the two CKAN apps, the Flask one and the Pylons one. One each request it asks both applications if they can handle it, and the apps answer yes or no and whether core or an extension can handle it. The request is then forwarded to the relevant app following this order of precedence:

Flask Extension > Pylons Extension > Flask Core > Pylons Core

### How it works

All the magic happens on `ckan.config.middleware`. 

1. The main WSGI entry point, `make_app` creates the Flask and Pylons stacks separately and passes the two apps to the `AskAppDispatcherMiddleware` constructor. 
2. This stores a reference to the apps and sends them the "invites", which essentially means calling the `join_party` methods of the `CKANFlask` app and the Pylons `PartylineController` to register their `can_handle_request` methods as the ones the dispatcher middleware will call on each request.
3. When a request comes in, `AskAppDispatcherMiddleware` asks both the Pylons and Flask apps if they can handle it (on its `__call__` method). It calls the `can_handle_request` method of each app passing the WSGI environment.
4. Each application performs the check in their own way, eg on Flask is something like:
    ```
          urls = self.url_map.bind_to_environ(environ)
          try:
              endpoint, args = urls.match()
              ...
    ```
    and on Pylons:
    ```
        pylons_mapper = config['routes.map']
        match_route = pylons_mapper.routematch(environ=environ)
        if match_route:
        ...
    ```
5. Each `can_handle_method` returns a tuple in the form `(True, app_name, origin)`. `app_name` can be either `flask_app` or `pylons_app` and `origin` can be `core` or `extension` depending on where this route was defined. (Note that this is not yet implemented for Flask, as we don't know how extensions will handle this).
6. Based on that information, `AskAppDispatcherMiddleware` forwards the request to relevant app, enforcing the order of precedence mentioned above.


amercader:
@davidread @rossjones would love your views on this

davidread:
Looks good to me. So this will allow us to start replacing controller actions one-by-one, pretty much.

amercader:
@davidread exactly, that's the idea

wardi:
This looks good. It's some extra complexity but it's written in a way that's easy to remove once we drop pylons support.

rossjones:
Yup, removability is probably the biggest concern for me, but looks like you've got it covered. Any idea how much work would be involved in getting it beyond WIP?

amercader:
@rossjones I have to write a few more tests, and maybe move the CKAN Flask app somewhere else. I don't want to add more stuff on this particular PR

amercader:
This is now ready for review

davidread:
Looks great to me, merging.

PR #2908, opened 2016-03-10
Start rewriting helpers.py

TkTech:
~~This PR is up for comments, but not ready for merging.~~

One of the primary reasons for this change is the difficulty of using static analyzers on this file. Unused methods could not be detected because of the wildcard imports and usage/call stats were smudged.

- [X] Remove unused method
- [X] Remove wildcard imports
- [X] PEP8
- [ ] 100% docstring coverage
- [x] Fix any ckan usages of helpers.py that depend on imports-of-imports (ex: h.asbool in controllers/users.py)


TkTech:
This PR is now merge-ready.

amercader:
Some questions before having looked at the code in depth:

* Does this affect extensions using helpers via the toolkit? (I guess no, because the toolkit just proxies to pylons.h)
* Do extensions need to use the `@builtin_helper` decorator to register their own helpers or do they still need to use the `ITemplateHelpers` interface?

ckan/plugins/toolkit.py
```diff
@@ -141,7 +141,7 @@ def _initialize(self):
 current request.
 
 '''
-        t['h'] = getattr(pylons.config['pylons.h'], 'no_magic', None)
+        t['h'] = pylons.config['pylons.h']
```
wardi: can we just use helpers.helper_functions now?  The code as-is has import order issues (configuration must be run before a module can import ckan.toolkit.h) but referencing the helper_functions AttributeDict might be safer.

wardi:
builtin_helper is not for extensions. Maybe just change the name a bit?  `_builtin_helper` or `_core_helper`?

wardi:
There are two extension-developer-visible changes to current behaviour here that need to be considered. I support both.

1. AttributeError on missing helpers in templates
2. helpers can now be overridden by extensions

ckan/lib/helpers.py
```diff
@@ -2121,125 +2262,35 @@ def get_translated(data_dict, field):
...
+def load_plugin_helpers():
+    """
+    (Re)loads the list of helpers provided by plugins.
+    """
+    global helper_functions
+
+    helper_functions.clear()
+    helper_functions.update(_builtin_functions)
+
+    for plugin in p.PluginImplementations(p.ITemplateHelpers):
```
wardi: reversed() would be good here. The first plugin to define a helper should win.

amercader:
> There are two extension-developer-visible changes to current behaviour here that need to be considered. I support both.
> 1.  AttributeError on missing helpers in templates
> 2.  helpers can now be overridden by extensions

As long as the AttributeError makes clear what the issue is (a helper is not defined) I'm happy with these.
They both should be mentioned in the Changelog.

ckan/plugins/toolkit.py
TkTech: Yes, we can do this since it's only ever modified in-place. I will make this change in the morning.

ckan/lib/helpers.py
TkTech: Are you sure? It seems more intuitive that the latest one overrides the earlier one, as this is how I'd assume it would work. This is how the current, existing _Helper behavior works.

TkTech:
> builtin_helper is not for extensions. Maybe just change the name a bit? _builtin_helper or _core_helper?

I intend to split helpers.py into many topic-based files later. Accessing an `_` prefixed member from an external module or class is a style violation.

> As long as the AttributeError makes clear what the issue is (a helper is not defined) I'm happy with these.

This is currently how the existing `no_magic` helper works, and is what Jinja expects. Jinja2 *already* does not consider accessing a non-existent member an error and will hide it. This original behavior, where an empty string is returned, I assumed to be a relic of the old templating.

wardi:
> I intend to split helpers.py into many topic-based files later. Its current complexity score is unacceptable and non-maintainable. Accessing an _ prefixed member from an external module or class is a style violation, and it is intended to be used from other internal files defining helpers.

Ok, `core_helper` is fine with me too. `core` is generally the word we use when we mean "not part of an extension"

>> helpers can now be overridden by extensions

> This is already the case.

No. In the current code a plugin can't override another plugin's helpers.

>> As long as the AttributeError makes clear what the issue is (a helper is not defined) I'm happy with these.

> This is currently how the existing no_magic helper works, and is what Jinja expects. Jinja2 already does not consider accessing a non-existent member an error and will hide it. This original behavior, where an empty string is returned, I assumed to be a relic of the old templating.

Maybe I don't understand you here. The existing behaviour is to return `lambda *args,**kwargs: ''` for missing attributes. That's not a default Jinja2 behaviour. `no_magic` is brand new, and not part of any stable CKAN release yet.

I think @amercader is asking for something like this to be added to the helpers AttributeDict. I support this idea.
```python
def __getattr__(self, name):
    raise AttributeError('Helper {name} has not been defined'.format(name=name))
```

ckan/lib/helpers.py
wardi: Template directories are the precedent here. A theme plugin that needs to override all other templates must appear earlier in the plugins list. It wouldn't make sense for the later plugins to override the theme's helpers.

TkTech:
> > > helpers can now be overridden by extensions
> > This is already the case.
> No. In the current code a plugin can't override another plugin's helper

Extras cannot override other extras, but they can override builtin helpers which is what I thought you meant.

Jinja's normal behavior is to suppress attribute errors. From the docs:

> If a variable or attribute does not exist, you will get back an undefined value. What you can do with that kind of value depends on the application configuration: the default behavior is to evaluate to an empty string if printed or iterated over, and to fail for every other operation.

> For the sake of convenience, foo.bar in Jinja2 does the following things on the Python layer:
> 
> check for an attribute called bar on foo (getattr(foo, 'bar'))
> if there is not, check for an item 'bar' in foo (foo.__getitem__('bar'))
> if there is not, return an undefined object.

Thus, `{{ h.doesn't_exist }}` does not cause an error even *without* the magic in _Helper.

wardi:
> Extras cannot override other extras, but they can override builtin helpers which is what I thought you meant.

You're right, I described the behaviour the wrong way initially.

> Thus, {{ h.doesn't_exist }} does not cause an error even without the magic in _Helper.

I agree, but the magic in _Helper also prevents {{ h.doesnt_exist() }} from raising an error, which is the behaviour we're changing.

Oh, unless you're saying raising an AttributeError won't help us because it's swallowed by Jinja2. So we'll need to raise some other exception, or raise the exception on the call not on getattr.

wardi:
@TkTech There are a few things left to be done: Renaming to core_helper, the plugin override order fix https://github.com/ckan/ckan/pull/2908#discussion_r56159587, the helpful exception error message and updating the changelog. Please let me know if you're going to have time to finish these.

davidread:
Needed also for: fixes https://github.com/ckan/ckan/issues/2941

The work all looks great. I just want to encourage @TkTech to complete this to get it merged sooner rather than later.

ckan/lib/helpers.py
```diff
@@ -50,18 +50,37 @@
...
 
+class HelperError(Exception):
+    """Raised if an attempt to access an undefined helper is made.
+
+    Normally, this would be a subclass of AttributeError, but Jinja2 will
+    catch and ignore them. We want this to be an explicit failure re #2908.
+    """
+
+
 class AttributeDict(dict):
```
wardi: name this HelperDict? We have a custom Helper-specific error now (and I'd rather not encourage the use of AttributeDict in other parts of the code... python is not JS!)

ckan/lib/helpers.py
```diff
@@ -50,18 +50,37 @@
 log = logging.getLogger(__name__)
 
 
+class HelperError(Exception):
+    """Raised if an attempt to access an undefined helper is made.
+
+    Normally, this would be a subclass of AttributeError, but Jinja2 will
+    catch and ignore them. We want this to be an explicit failure re #2908.
+    """
```
wardi: @amercader @TkTech how do you feel about moving HelperError to ckan.exceptions? CKAN tends to have all its exceptions buried in the modules that use them, but in projects I've built recently I like to have my exceptions defined in a tiny shared module. It's helped me to prevent import loops and Exceptions feel like part of the public interface to the code.

At a minimum we should add it to ckan.plugins.toolkit so extensions can check for the existence of helpers in a safe way.

amercader: That sounds good to me (ckan.extensions and toolkit)

ckan/lib/helpers.py
```diff
@@ -44,17 +40,53 @@
...
 
-from ckan.common import (
-    _, ungettext, g, c, request, session, json, OrderedDict
-)
+from ckan.common import _, ungettext, g, c, request, session, json
 
-get_available_locales = i18n.get_available_locales
-get_locales_dict = i18n.get_locales_dict
+# get_available_locales = i18n.get_available_locales
+# get_locales_dict = i18n.get_locales_dict
```
wardi: if we're not using this let's just remove it

davidread:
I just want to flag that this change appears to get rid of h.nav_named_link. However it's probably ok because it was deprecated a couple of years ago, and extensions should have been fixed earlier (ckanext-harvest).

PR #2909, opened 2016-03-11
Added the ability to specify translations for the default language

LaurentGoderre:
Allows to override the default language without modifying the templates

### DIFF
--- a/ckan/lib/i18n.py
+++ b/ckan/lib/i18n.py
@@ -127,19 +127,19 @@ def _set_lang(lang):
     if config.get('ckan.i18n_directory'):
         fake_config = {'pylons.paths': {'root': config['ckan.i18n_directory']},
                        'pylons.package': config['pylons.package']}
-        i18n.set_lang(lang, pylons_config=fake_config, class_=Translations)
-    else:
+
+        try:
+            i18n.set_lang(lang, pylons_config=fake_config, class_=Translations)
+        except i18n.LanguageError:
+            pass
+    elif lang != 'en':
         i18n.set_lang(lang, class_=Translations)
-
-
 
 def handle_request(request, tmpl_context):
     ''' Set the language for the request '''
     lang = request.environ.get('CKAN_LANG') or \
         config.get('ckan.locale_default', 'en')
-    if lang != 'en':
-        set_lang(lang)
-
+    set_lang(lang)
 
     for plugin in PluginImplementations(ITranslation):
         if lang in plugin.i18n_locales():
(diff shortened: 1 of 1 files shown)


amercader:
@LaurentGoderre could you give more context about why this is needed and what is the effect of this PR?

LaurentGoderre:
This allows changing english ckan strings without modifying all the templates where they are defined. For example, it was requested that we change all instance of the word dataset to the word record. Instead of changing all the 200+ templates where it occurs, this takes care of that requirement.

Most people probably won't need this but it's a neat feature to have.

ckan/lib/i18n.py
```diff
@@ -127,19 +127,19 @@ def _set_lang(lang):
...
         fake_config = {'pylons.paths': {'root': config['ckan.i18n_directory']},
                        'pylons.package': config['pylons.package']}
-        i18n.set_lang(lang, pylons_config=fake_config, class_=Translations)
-    else:
-        i18n.set_lang(lang, class_=Translations)
-
 
+        try:
+            i18n.set_lang(lang, pylons_config=fake_config, class_=Translations)
+        except i18n.LanguageError:
```
amercader: We were not capturing and ignoring these before, is this needed by the new behaviour (supporting `en`)?

LaurentGoderre: Basically, if you don't have an english override, it triggers that exception saying 'english is not found'. So by trapping it this way, if there is no english override or if they use a missing language, it will revert to the default language

joetsoi:
@LaurentGoderre, changing "Dataset" to "Record" should have already been possible by using [ckan.i18n.extra_directory](http://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-i18n-extra-directory) and dropping an extra translation file in. Or by adding an english translation in your extension [there's an example one in ckanext folder](https://github.com/ckan/ckan/tree/master/ckanext/example_itranslation)

What is your default language set as, it's possible there is a slight different case which I didn't account for when I wrote this.

LaurentGoderre:
The english translation file are ignored because they are the default without this changed

joetsoi:
If english is default, then there should be no translation object in pylons , this case should be handled by
https://github.com/ckan/ckan/blob/master/ckan/lib/i18n.py#L164-L173 which adds the the extra translations you provide as the default.

I'll take a look again as it's been a while since I looked at this.

LaurentGoderre:
If you can show me some code that makes it work it would be appreciated. I couldn't get it to work

khayo36:
Can anyone tell me how to add new language(a translation not in ckan.locale) in to ckan, please ????

tino097:
@khayo36 please check the docs http://docs.ckan.org/en/2.8/contributing/i18n.html

kmbn:

PR #2913, opened 2016-03-14
Add some new badges to README

amercader:
Added license, docs and support badges. This was a requirement of the
FIWARE project which is funding some of the current core work.
They seem to be generally useful but if someone is not happy we can
definitely drop one.


wardi:
office-space

Well, okay. Fifteen is the minimum, okay?

Issue #2524, opened 2015-07-07
Schema.xml not compatible with Solr 5.2.1


FabiApfelkern:
I set up CKAN 2.3 with Solr 5.2.1. Everything works fine and new datasets are indexed. But when you restart Solr then the core seems to be corrupted and is not loading with this error [1]. It cannot be recovered. It's very likely to be a bug in Solr, or the CKAN Schema is in a way not compatible with this Solr Version. 

[1]
```
java.lang.NullPointerException
	at org.apache.lucene.util.NumericUtils.prefixCodedToLong(NumericUtils.java:226)
	at org.apache.lucene.util.NumericUtils.getMaxLong(NumericUtils.java:582)
	at org.apache.solr.update.VersionInfo.getMaxVersionFromIndex(VersionInfo.java:234)
	at org.apache.solr.update.UpdateLog.seedBucketsWithHighestVersion(UpdateLog.java:1580)
	at org.apache.solr.update.UpdateLog.seedBucketsWithHighestVersion(UpdateLog.java:1610)
	at org.apache.solr.core.SolrCore.seedVersionBuckets(SolrCore.java:861)
	at org.apache.solr.core.SolrCore.<init>(SolrCore.java:843)
	at org.apache.solr.core.SolrCore.<init>(SolrCore.java:658)
	at org.apache.solr.core.CoreContainer.create(CoreContainer.java:637)
	at org.apache.solr.core.CoreContainer$1.call(CoreContainer.java:381)
	at org.apache.solr.core.CoreContainer$1.call(CoreContainer.java:375)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at org.apache.solr.common.util.ExecutorUtil$MDCAwareThreadPoolExecutor$1.run(ExecutorUtil.java:148)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

mordae:
Confirming.

wardi:
We had to make this change to our solr config (not schema.xml) https://github.com/open-data/ckanext-stcndm/commit/0b1f4b423 does that help?

marinajungblut:
Same problem here.

mordae:
Weird. Changing type of the `_version_` field to `long` apparently fixed the issue for us.

[...]

PR #2915, opened 2016-03-14
[#2914] Solr5 schema.xml compatiblity, resolves #2914

### LINKED ISSUE #2914: Default Schema Incompatible With New Solr Cores [5+]

All

schema.xml, once moved into the core folder and Solr's managed schema's disabled, should "just work".

Solr raises errors when reloading the core due to types used in sorlconfig.xml but not present in schema.xml. These are a set of new "example" multivalue variants of the default types (booleans, tlongs, tints). Several stackoverflow posts exist already from users encountering this.

This "fix", while not a technical fix, should lower user entry into CKAN should be fully backwards compatible.

Follow CKAN setup instructions with Solr 5+.

TkTech:
Fixes #2914

### DIFF
--- a/ckan/config/solr/schema.xml
+++ b/ckan/config/solr/schema.xml
@@ -41,6 +41,13 @@ schema. In this case the version should be set to the next CKAN version number.
     <fieldType name="date" class="solr.TrieDateField" omitNorms="true" precisionStep="0" positionIncrementGap="0"/>
     <fieldType name="tdate" class="solr.TrieDateField" omitNorms="true" precisionStep="6" positionIncrementGap="0"/>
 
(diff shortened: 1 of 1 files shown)


amercader:
@TkTech Did you find any of the issues described in #2524? There were mentions there of other fields that needed to be added to the Solr schema

TkTech:
@amercader That particular issue can most likely be closed - it affects only one very specific point release of Solr where a change to the version reader was made which broke when reading strings. It was fixed in r1694389 and is only broken between 5.1 and 5.3. This was a bug in Solr, not CKAN's schema, and the change to `long` from `string` was just a workaround.

PR #2898, opened 2016-03-08
Some links in CKAN is not reachable

amercader:
### CKAN Version if known (or site URL)

for example:

Configuration File:

http://docs.ckan.org/en/latest/maintaining/CKANconfigurationfile

Environmental Variables:

http://docs.ckan.org/en/latest/maintaining/env-vars-wikipedia

### Please describe the expected behaviour

Expect link to related documents.

### Please describe the actual behaviour

Page not found.

PR #2918, opened 2016-03-15
[#2898] Fix documentation links

smotornyuk:


amercader:
Thanks for the patch @smotornyuk

PR #2921, opened 2016-03-17
[#2856] Limit calls between DataPusher -> DataStore -> CKAN

### LINKED ISSUE #2856: DataPusher called multiple times when creating a dataset
When you upload a CSV the DataPusher [gets called](https://github.com/ckan/ckan/blob/master/ckanext/datapusher/plugin.py#L98) to do its thing.
On https://github.com/ckan/ckan/pull/2234 we added a couple of calls to `resource_patch` on `datastore_create` and `datastore_delete` to set the `datastore_active` extra on the resource.

This is all very well but this update on the resource triggers again the DataPusher, which pings `datastore_delete` and `datastore_create`, which trigger...

Not sure how we can handle this on the `notify` extension point, as we only get the model object there.

amercader:
Fixes #2856

* Don't send resources to DataPusher if there is an ongoing job
* Only set datastore_active on resources when necessary


ckanext/datastore/logic/action.py
```diff
@@ -143,8 +143,13 @@ def datastore_create(context, data_dict):
         raise p.toolkit.ValidationError(str(err))
 
     # Set the datastore_active flag on the resource if necessary
-    p.toolkit.get_action('resource_patch')(
-        context, {'id': data_dict['resource_id'], 'datastore_active': True})
+    if not resource.extras.get('datastore_active') is True:
```
TkTech: "is not"

amercader:
@TkTech @wardi What do you think about getting this merged for now even with the caveat that @TkTech mentioned on IRC?

> 12:03 PM <tktech> Update CSV -> datapusher job starts
> 12:03 PM <tktech> Realize it's wrong, quickly update CSV again -> datapusher status check shows a running job, nothing starts/queues.

There is actually a way to resubmit a file before the first import is finished. The pending check added on this patch only applies to the , but does not apply to manually reimporting the resource to the DataStore (Resource page > Manage > DataStore > Upload to DataStore)
This would avoid

TkTech:
I'm okay with it as long as the caveat is made very, very clear. There are many portals that deal with potentially sensitive material that may need updating/redaction, and even though it's unlikely this gotcha can result in stale/incorrect data.

amercader:
@TkTech @wardi 83b963c (and 31b607f) should address the scenario where an update is done while a DataPusher job is ongoing. We checked if a new file was uploaded (via `resource['last_modified']`) or if a new URL was set.

I'll try and come up with some tests for this

amercader:
@TkTech @wardi I pushed your suggested tweaks. I've also added a bunch of tests. They use a new decorator that helps mock action functions (f63a972). Some tests are a bit long-winded but that's what you get when you deal with the async stuff from DataPusher. Let me know if you need some clarification.

amercader:
bump

wardi:
Tests could use FunctionalTestBase instead of calling reset_db on setup and teardown of whole class. That's a small thing, though. LGTM

amercader:
Cheers @wardi

PR #2925, opened 2016-03-21
Explicitly create a ckan core for tests

LaurentGoderre:
This will make it easier in the future to upgrade to a newer version of Solr, where there is no default core.

### DIFF
--- a/bin/solr_init/create_core.sh
+++ b/bin/solr_init/create_core.sh
@@ -1,7 +1,11 @@
 #!/bin/sh
+SOLR_HOME=${SOLR_HOME:-/opt/solr}
+SOLR_URL=${SOLR_HOST:-http://localhost:8983}
+SOLR_CORE=${SOLR_CORE:-ckan}
+SOLR_SCHEMA=${SOLR_SCHEMA:-$HOME/ckan/ckan/config/solr/schema.xml}
 
-INSTANCE_DIR=$SOLR_HOME/example/solr/$CKAN_SOLR_CORE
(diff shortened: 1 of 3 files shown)


bin/solr_init/create_core.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+INSTANCE_DIR=$SOLR_HOME/example/solr/$CKAN_SOLR_CORE
+cp -R $SOLR_HOME/example/solr/collection1 $SOLR_HOME/example/solr/$CKAN_SOLR_CORE
+cp $CKAN_SOLR_SCHEMA $SOLR_HOME/example/solr/$CKAN_SOLR_CORE/conf
+
+curl "http://localhost:8983/solr/admin/cores?action=CREATE&name=$CKAN_SOLR_CORE&instanceDir=$INSTANCE_DIR"
```
wardi: you've got `localhost:8983` hard-coded here. Add another variable?

What do you think about using bash defaults at the top of these scripts for the variables they need? That sort of documents them and will make our circle.yml much smaller.

e.g.
```bash
CKAN_SOLR_CORE=${CKAN_SOLR_CORE:-ckan}
CKAN_SOLR_SCHEMA=${CKAN_SOLR_SCHEMA:-$HOME/ckan/ckan/config/solr/schema.xml}
SOLR_HOME=${SOLR_HOME:-$HOME/solr}
...
```

LaurentGoderre: I wasn't sure what the syntax was for that but I would gladly do it

LaurentGoderre: Done. Perhaps we can do the same with the postgres script too

PR #2927, opened 2016-03-22
Fix #2926: Only include snippet HTML comment in debug mode.

### LINKED ISSUE #2926: Snippet HTML comment should only be inserted when debugging is enabled
As of CKAN 2.5.1, the `{% snippet %}` Jinja-directive always places HTML-comments around the snippet content:

```html
<!-- Snippet snippets/tag_list.html start -->
  <ul class="tag-list">
      <li>
        <a class="tag" href="/data/dataset?tags=geodaten">geodaten</a>    
      <li>
        <a class="tag" href="/data/dataset?tags=markt">markt</a>    
  </ul>
<!-- Snippet snippets/tag_list.html end -->
```

In my opinion, this should only be done if debug mode is enabled.

torfsen:
This fixes #2926.

Previously, the `{% snippet %}` Jinja-directive always surrounded the snippet's content with HTML comments. Now this is only done if the `debug` configuration variable is true.

### DIFF
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -75,8 +75,9 @@ def render_snippet(template_name, **kw):
     cache_force = kw.pop('cache_force', None)
     output = render(template_name, extra_vars=kw, cache_force=cache_force,
                     renderer='snippet')
-    output = '\n<!-- Snippet %s start -->\n%s\n<!-- Snippet %s end -->\n' % (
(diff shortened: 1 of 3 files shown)


amercader:
Thanks for the patch @torfsen

torfsen:
Glad I could help :wink:

PR #2929, opened 2016-03-23
Made the user list limit configurable

LaurentGoderre:
Allows the user list limit to be set by ini setting or querystring


amercader:
This looks good @LaurentGoderre, can you add the new option to the config option docs? http://docs.ckan.org/en/latest/maintaining/configuration.html

LaurentGoderre:
@amercader done!

Issue #2873, opened 2016-02-10
Incorrect 'download resource' hyperlink when a resource is unable to upload to datastore 


Mbrownshoes:
When a dataset cannot be previewed in the Data Explorer - e.g. when the dataset it too large, or has not been loaded to the DataStore (e.g. http://demo.ckan.org/dataset/bad-hyperlink-example/resource/36a112c0-7a4e-40da-aa0d-8e0c67e64957) the Data Explorer is providing an incorrectly addressed "Download resource" hyperlinked button
![image](https://cloud.githubusercontent.com/assets/1154765/12952293/d5c9a624-cfca-11e5-9a18-754d1f0e928a.png)
 
The hyperlink URL  http://demo.ckan.org/dataset/bad-hyperlink-example/resource/36a112c0-7a4e-40da-aa0d-8e0c67e64957 is missing the terminating text /download/

amercader:
@smotornyuk @Engerrs can either of you have a look at this one? It should be easy to track where the URL is coming from. The template used is this one:

https://github.com/ckan/ckan/blob/master/ckan/templates/package/snippets/resource_view.html

Thanks!

smotornyuk:
I will take that one

amercader:
Thanks @smotornyuk!

wardi:
Was this not fixed by https://github.com/ckan/ckan/pull/2599

smotornyuk:
No, problem here in variable that not passed into template

amercader:
Fixed by #2930

PR #2930, opened 2016-03-24
[#2873] Use `resource.url` as raw_resource_url

smotornyuk:
`resource_view` template uses raw_resouce_url in order to create
download link if resource can't be displayed, but this variable is not
passed into snippet. To make that simpler, use resource.url property
for download button.


amercader:
Thanks for the patch @smotornyuk!

PR #2931, opened 2016-03-28
Fix response code test failures

wardi:


wardi:
@nigelbabu do you have time to look at this too? It fixes the tests I broke in #2846

nigelbabu:
Merged, sorry about causing bitrot thanks to my slowness!

PR #2928, opened 2016-03-22
Added a test for the url_for_static helper

LaurentGoderre:

PR #2932, opened 2016-03-29
Fixed change_config not restoring config when a test fails

LaurentGoderre:


wardi:
This one is clearly correct and independent from #2928

PR #2933, opened 2016-03-29
Split tests across circle CI containers

wardi:


wardi:
@amercader so, yeah, --segments is revealing some broken tests, and caching the nvm directory is making the restore hugely slower and *still* not helping to prevent the re-download.

What do you think about me switching back to the version before using --segments and adding a hack that will skip the npm install command when there is a cached version found (like http://stackoverflow.com/questions/31766930/circleci-not-caching-my-globally-installed-node-module )

wardi:
#2937 has the phantomjs fix. I'll rebase this PR on to that code

amercader:
The 3 segments failures look easy to fix (just duplicated package names because of tests not cleaning up) and I really like the idea of not having to explicitly define the tests to run and forget about it. The nvm caching hack (which sounds great) is unrelated to using segments or not right?

wardi:
So how do I figure out which tests aren't cleaning up after themselves?

Yes #2937 is the npm caching hack, unrelated to this PR

amercader:
Can we change the dataset name on the 3 failing tests?

wardi:
Aha! here's what I need: https://circle-artifacts.com/gh/ckan/ckan/1171/artifacts/2/tmp/circle-junit.eVfC8te/nose/junit.xml

wardi:
@amercader I think I got it

amercader:
6 minutes! wooo :bullettrain_side: :dash: :dash:

amercader:
Hats off to @kindly for the original segments implementation!

wardi:
@kindly :tophat: :womans_hat:

We could do a little better if we run the JS tests on just one of the containers (number 0 seems to be a good target). I'll have to think about the best way to do that since we need circle's `background` setting

wardi:
@TkTech 's work on #2916 was also huge in getting us down to this number. These tests were running around 10 minutes before I rebased.

PR #2934, opened 2016-03-30
Better stacktrace for broken SOLR

davidread:
### Proposed fixes:
* The benefit of "raise" on its own is that the original stacktrace is made available, rather than creating a new stacktrace with "raise search_error". So its easier to debug.

### Features:

- [ ] includes API changes


wardi:
yes, please!

PR #2954, opened 2016-04-12
Fix #2953: Strange facet item order.

### LINKED ISSUE #2953: Strange facet item order
As of CKAN 2.5.2, the search facet items are listed in a strange order: They are first sorted by the number of matching datasets (that's good) and then by their name in *reverse* alphabetic order, with lower case before upper case (that's strange). Here's a screenshot from [demo.ckan.org](http://demo.ckan.org):

facet_order

I would have expected the following order: First search by count and then by name in alphabetic order, without distinguishing between upper and lower case.

torfsen:
Fixes #2953:

Previously, facet items (e.g. tags) were first sorted descendingly by
number of matching datasets and then descendingly by case-sensitive
display name. The latter is unexpected and makes browsing the facets
unnecessarily difficult.

With this PR, facets are first sorted descendingly by dataset count
(as before) and then ascendingly by case-insensitive display name.

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -670,7 +670,8 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
             facets.append(dict(active=False, **facet_item))
         elif not exclude_active:
             facets.append(dict(active=True, **facet_item))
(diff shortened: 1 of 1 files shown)


amercader:
thanks for the patch @torfsen

PR #2956, opened 2016-04-12
Adjust copyright messages to reference all contributors, not just OKF

davidread:
It's more appropriate to reference CKAN Association and contributors than OKF, now that CKAN is not just the product of OKF any more.

Copyright messages should reference the owners, and for an open source project it's ok to say it is the contributors (and by implication their organizations). See: http://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/

First CKAN commit was 2007 not 2006 - see https://github.com/ckan/ckan/commits?author=rgrp&page=28

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -2,7 +2,7 @@ License
 +++++++
 
 CKAN - Data Catalogue Software
-Copyright (C) 2007 Open Knowledge Foundation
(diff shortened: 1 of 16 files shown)


amercader:
While changing the OK references to the CKAN Association and the wider contributors sounds good, copyright notices might have wider implications so let's confirm that bit before merging

PR #2957, opened 2016-04-13
[#2946] "Unsupported" warning banner shown for old docs

### LINKED ISSUE #2946: Old docs misguiding people
I think we should stop showing old CKAN docs on readthedocs - just have the latest (or at least only the last couple of versions that we are releasing patch releases for). I think it confuses more people than it helps.

It's not too bad if you search for something that is still current:

e.g. https://www.google.co.uk/?q=ckan%20install%20source returns results:
- docs.ckan.org/en/latest/maintaining/installing/install-from-source.html
- docs.ckan.org/en/ckan-2.2/install-from-source.html
- docs.ckan.org/en/ckan-1.3.3/maintaining/installing/index.html
- docs.ckan.org/en/ckan-1.8/install-from-source.html
- docs.ckan.org/en/latest/maintaining/installing/install-from-package.html
- docs.ckan.org/en/ckan-2.0/install-from-source.html
- docs.ckan.org/en/latest/maintaining/upgrading/upgrade-source.html

although you just hope people will click the top answer and ignore the rest.

If they search for something CKAN doesn't do any more then it's a problem:
- https://lists.okfn.org/pipermail/ckan-dev/2016-April/009891.html
- https://lists.okfn.org/pipermail/ckan-dev/2016-April/009879.html

I think the latest docs are plenty tbh. It's usually the best source of info, even for older versions, since descriptions are improved over time. We've always written docs with that in mind. And it's rare that we drop features or break anything - backward compatibility rules. But if a feature _has_ been changed/dropped, and someone still wants to read about how it was documented back then, then they have the .rst on their disk to refer to. And I think that is better balance than having people ask about dropped features forever because they came across it on google.

Let's hear views/votes @brew @amercader @wardi:
a) keep all the versions
b) keep latest & last 2 point releases (about what we support with patch releases)
c) keep only the latest

davidread:
Fixes #2946

Shows "Unsupported" banner over docs older than the latest 3 point releases. e.g. currently docs for 2.5.2, 2.4.3 and 2.3.4 would show as normal, but 2.2.3, 2.1.5 etc would show as "unsupported".

banner

It's a bit annoying that it clips some of the page content, but that requires CSS changes to the sphinx template. I think it is good enough.

To test this:
```
pip install -r dev-requirements.txt
vim ckan/__init__.py # and set the version temporarily to 2.1.0
python setup.py build_sphinx
# open ckan/build/sphinx/html/api/index.html
```

### DIFF
--- a/ckan/__init__.py
+++ b/ckan/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '2.1.0a'
+__version__ = '2.6.0a'
 
 __description__ = 'CKAN Software'
 __long_description__ = \
--- a/ckan/__init__.py
+++ b/ckan/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '2.6.0a'
+__version__ = '2.1.0a'
 
 __description__ = 'CKAN Software'
 __long_description__ = \
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -9,6 +9,8 @@ polib==1.0.3
 mock==1.0.1
 factory-boy==2.1.1
 coveralls
-sphinx-rtd-theme==0.1.9
+# sphinx-rtd-theme - we need commit eeff564 which missed the 0.1.9 release
+# (which is the latest at time of writing)
+-e git://github.com/snide/sphinx_rtd_theme.git@eeff5645dc4eeb7a5c4910d39c17131961045dce#egg=sphinx_rtd_theme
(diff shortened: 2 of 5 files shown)


wardi:
If this sphinx theme was a separate package we could update it to apply theme changes across all the versions we build, instead of backporting this to every version of ckan.

davidread:
Good idea.

PR #2959, opened 2016-04-13
Rewrite activity handling.

TkTech:
This PR is not yet ready for merging and is posted for review only.

The primary goal of this PR is to fix unacceptable performance of activity streams. It's secondary goal is general code quality improvements.

### DIFF
--- a/ckan/lib/activity_streams_session_extension.py
+++ b/ckan/lib/activity_streams_session_extension.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from pylons import config
 from sqlalchemy.orm.session import SessionExtension
 from paste.deploy.converters import asbool
@@ -130,9 +131,9 @@ class DatasetActivitySessionExtension(SessionExtension):
                     activity_detail = activity_stream_detail(
                         obj, activity.id, activity_type)
                     if activity_detail is not None:
-                        if not package.id in activities:
+                        if package.id not in activities:
                             activities[package.id] = activity
-                        if activity_details.has_key(activity.id):
+                        if activity.id in activity_details:
                             activity_details[activity.id].append(
                                 activity_detail)
                         else:
--- a/ckan/lib/activity_streams_session_extension.py
+++ b/ckan/lib/activity_streams_session_extension.py
@@ -44,6 +44,8 @@ class DatasetActivitySessionExtension(SessionExtension):
         if not asbool(config.get('ckan.activity_streams_enabled', 'true')):
             return
 
+        from ckan.model.activity import Activity
+
         session.flush()
 
         try:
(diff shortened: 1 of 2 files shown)


ckan/lib/activity_streams_session_extension.py
```diff
@@ -142,6 +144,24 @@ def before_commit(self, session):
...
+                except KeyError:
+                    continue
+
+                data = activity.data.copy()
+                data['real_object_id'] = activity.object_id
+
+                session.add(
+                    Activity(
+                        user_id=activity.user_id,
+                        object_id=obj['owner_org'],
```
amercader: Depending on your auth configuration, not all datasets always belong to an organization so it may be worth checking if `obj.get('owner_org')`

amercader:
@TkTech this is great, activities are indeed often a bottleneck.

Looks good so far. My only question (and I don't know the activities code that well) is if 592127b won't increase too much the size of the activities table, is that essentially duplicating all dataset activities?

TkTech:
@amercader It duplicates new, changed, and deleted events for datasets. Of these three, "CHANGED" is the only one that has a chance of generating any quantity of results.

This should have been done in the first place. Organization history must be a property of the organization, not of individual data sets that have mutable membership.

That said, these individual entries are tiny - the ActivityDetails records are not duplicated but referenced from both the data set Activity and the organization Activity. You can realistically store billions of these records before running into issues. Only the first 30 on a sorted index are recalled in the current code. If this size becomes an issue, we can optimize them through a number of easy paths + migrations. One of these is changing activity_type from a large string to an ENUM type. Another is changing the `id` from a UUID to an incremental BIGINT (there is no reason for these to have UUIDs).

Additionally, history length can be a configurable option. Once background workers are implemented, we can easily add caps. Alternatively we can make the logic and schema more complex using a `SEQUENCE` with `CYCLE` behavior to implement a capped table.

Long story short, this is a change that must be made now. It is currently possible to use this as a denial-of-service vector on any existing CKAN instance that does not have activity disabled and has more than 1000 records in a single organization.

ckan/lib/activity_streams_session_extension.py
TkTech: Good point! I will make this change.

amercader:
@TkTech makes perfect sense, happy to go dow that route

Let me know when this is ready to review & merge

ckan/model/activity.py
```diff
@@ -180,14 +211,12 @@ def _group_activity_query(group_id):
         # Return a query with no results.
         return model.Session.query(model.Activity).filter("0=1")
 
-    dataset_ids = [dataset.id for dataset in group.packages()]
+    q = model.Session.query(
+        model.Activity
+    ).filter(
+        model.Activity.object_id == group_id
+    )
```
wardi: Would you separate out this performance fix into its own PR? We could merge this part quickly.

TkTech: Nope, because the behaviour of the two is not the same. Just this change would leave you with no dataset activity.

PR #2962, opened 2016-04-15
Allow RSS feed to be extended

martinburchell:
### Features:

We've made a couple of small changes to allow a different webhelpers RSS class to be used for RSS feeds. For [our plugin](https://github.com/aptivate/ckanext-mapactiongeorss) we wanted to have GeoRSS feed support  and populate the georss fields from custom fields in the dataset metadata.

Code behaviour should be the same as before this change ie existing tests should still pass.

- [X] includes API changes

Allow RSS feed classes in webhelpers other than Atom1.
Allow additional values to be set on a feed item.


martinburchell:
Is there anything I can do to help move this one along?

amercader:
Many apologies @martinburchell for the late reply, this slipped under my radar.

The canonical approach for extending CKAN core functionalities is using [interfaces](https://github.com/ckan/ckan/blob/master/ckan/plugins/interfaces.py) from plugins, rather than sub-classing core classes.

In your case, we need a new `IFeed` interface with the two new methods you added in this PR. Something like this:

```diff
diff --git a/ckan/controllers/feed.py b/ckan/controllers/feed.py
index 3b41af9..2cdac43 100644
--- a/ckan/controllers/feed.py
+++ b/ckan/controllers/feed.py
@@ -31,6 +31,7 @@ import ckan.model as model
 import ckan.lib.base as base
 import ckan.lib.helpers as h
 import ckan.logic as logic
+import ckan.plugins as plugins
 
 from ckan.common import _, g, c, request, response, json
 
@@ -356,7 +357,13 @@ class FeedController(base.BaseController):
             config.get('ckan.site_url', '').strip()
 
         # TODO language
-        feed = _FixedAtom1Feed(
+        feed_class = None
+        for plugin in plugins.PluginImplementations(plugins.IFeed):
+            feed_class = plugin.get_feed_class()
+        if not feed_class:
+            feed_class = _FixedAtom1Feed
+
+        feed = feed_class(
             title=feed_title,
             link=feed_link,
             description=feed_description,
@@ -372,6 +379,9 @@ class FeedController(base.BaseController):
         )
 
         for pkg in results:
+            for plugin in plugins.PluginImplementations(plugins.IFeed):
+                extras = plugin.get_item_extras(pkg)
+
             feed.add_item(
                 title=pkg.get('title', ''),
                 link=self.base_url + h.url_for(controller='package',
@@ -391,7 +401,8 @@ class FeedController(base.BaseController):
                                               id=pkg['name'],
                                               ver='2'),
                     unicode(len(json.dumps(pkg))),   # TODO fix this
-                    u'application/json')
+                    u'application/json'),
+                **extras
             )
         response.content_type = feed.mime_type
         return feed.writeString('utf-8')
diff --git a/ckan/plugins/interfaces.py b/ckan/plugins/interfaces.py
index b2907c8..a49a78d 100644
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -199,6 +199,32 @@ class IDomainObjectModification(Interface):
         pass
 
 
+class IFeed(Interface):
+    """
+    Allows to extend the default Atom feeds
+    """
+
+    def get_feed_class(self):
+        """
+        Allows plugins to provide a custom class to generate feed items.
+
+        The feed item generator class should accept the following parameters
+        on the constructor method:
+        TODO
+        """
+
+        pass
+
+    def get_item_extras(self, dataset_dict):
+        """
+        Allows plugins to set additional fields on a feed item.
+
+        :param dataset_dict, a dict with the dataset metadata
+        """
+        pass
+
+
+
 class IResourceUrlChange(Interface):
     """
     Receives notification of changed urls.
```

And then from your plugin you would do something like:

```python

import ckan.plugins as p


class MyPlugin(p.SingletonPlugin):

    p.implements(p.IFeed)

    def get_feed_class():

         return MyCustomFeedClass

   def get_item_extras(dataset_dict):

       # ....

```


Does this make sense?

martinburchell:
Yes that makes sense. I'll see what I can do.

ckan/plugins/interfaces.py
```diff
@@ -200,6 +202,42 @@ def notify_after_commit(self, entity, operation):
...
+    """
+    Allows extension of the default Atom feeds
+    """
+
+    def get_feed_class(self):
+        """
+        Allows plugins to provide a custom class to generate feed items.
+
+        The feed item generator class should accept the following parameters
+        on the constructor method:
```
amercader: I guess not all of these are mandatory. Can you indicate which are (if any) and if not mention that they are all optional.

ckan/tests/controllers/test_feed.py
```diff
@@ -81,3 +84,56 @@ def test_custom_atom_feed_works(self):
         assert '<title>{0}</title>'.format(dataset1['title']) in res.body
 
         assert '<title>{0}</title>'.format(dataset2['title']) not in res.body
+
+
+class TestFeedInterface(helpers.FunctionalTestBase):
+    @classmethod
+    def setup_class(cls):
+        super(TestFeedInterface, cls).setup_class()
+        plugins.load('test_feed_plugin')
```
amercader: Do 

```python
    if not plugins.plugin_loaded('test_feed_plugin'):
        plugins.load('test_feed_plugin')
```

This prevents problems when tests fail and plugins are not unloaded properly

amercader:
@martinburchell let's solve these conflicts with master and we are good to go

martinburchell:
@amercader I've resolved the conflicts and I think the failures are still unrelated

torfsen:
Hm. I'm wondering if that failure is related to #3243, where we've seen similar errors. That issue should have been fixed in #3247 but perhaps there's still something wrong.

amercader:
Thanks for the patch @martinburchell and sorry it took a while to get merged. We'll sort these failures elsewhere.

PR #2963, opened 2016-04-15
OKF->CKAN Association/contributors

davidread:
Based on https://github.com/ckan/ckan/pull/2956 but without the changes to copyright messages, for which a decision from the steering board may be required, which delays things. Once this PR is merged it can simplify that other PR too, making it easier for the steering board to read that PR without the extraneous changes.


davidread:
Sounds like @amercader is happy with this part - merge?

PR #2968, opened 2016-04-18
Fix redirect not working with non-root

LaurentGoderre:


wardi:
@LaurentGoderre looks pretty good. Would you be able to write a test to show that the redirects with a root_path are working now?

PR #2971, opened 2016-04-20
[POC] Flask views

amercader:
Goal: Have a working example of two controllers/views able to handle requests completely via Flask, with the same current functionality as the Pylons based ones:

* an API one (v3 action API)
* a web based one that includes template rendering

During the process we need to identify issues that will need addressing or discussion, for example:

* Routing
* Authorization (integration with repoze)
* Translations
* Template rendering
...

Specific questions that need to be addressed as part of this work

* Settle on a preferred way of defining routes
* Best option for extension to modify this routes (ie support for `IRoutes` interface)

Prior work done by @rossjones : https://github.com/ckan/ckan/tree/flask-poc


amercader:
Issues identified so far:

* The new Flask views have been created in the `ckan.views` folder to clearly separate them from the Pylons controllers

* To structure the view, I used [Flask-Classy](https://pythonhosted.org/Flask-Classy/). That seemed to provide more flexibility and allow a structure closer to the existing one with Classes and methods for controllers and actions. Of course that implies that we need to use whatever routing options it offers. Right now routes are defined via a decorator of the action method. This is a departure from the current "all routes defined on routing.py" approach. Maybe there is a way of follow a similar approach or maybe it's actually a good thing to have the routes next to the code they execute.

* The pattern in the Pylons middleware is to pass the `app` to the next middleware, which will store a reference to the previous app, eg:

```
     app = PylonsApp()
     ...
     app = RegistryManager(app)
     app = I18nMiddleware(app, config)
     ...

     app = TrackingMiddleware(app, config)
     return app
```

   So in this case, app is an instance of `TrackingMiddleware`. Some current tests require access to the Flask application (eg to add a new route). We can not access this with the output of `make_flask_stack()` so we might need a way to reference the Flask app object (via [current_app](http://flask.pocoo.org/docs/0.10/api/#flask.Flask.app_context)?)

* The way Flask handles request params parsing is making a lot of tests fail because params are not extracted from the payload (eg `ckan.tests.controllers.test_api.TestApiController.test_config_option_list_access_sysadmin`). This is most probably a problem in how we are doing requests in the tests, but we need to support it in Flask for backwards compatibility

* One of the first things that was needed from Pylons was the `c` object which is used everywhere. The simplest solution was to create a wrapper in `lib/common.py` that proxied to `pylons.c` or `flask.g` depending on if we were in the context of a Flask request or a Pylons request (checking for `flask.request`)

* Translations via [flask-babel](https://pythonhosted.org/Flask-Babel/) don't appear to work exactly as in Pylons. For some reasone the Flask gettext changes 'String %s' to 'String {}' (maybe it's the babel version?)

brew:
With reference to the routing prescribed by Flask-Classy, and looking at other options:

Flask-Classy:
* Allows us to group related actions together as class methods, which is most similar to our current pylons controller structure
* Performs some magic in `FlaskView.register()` to automatically set up routing, which can be overridden by a route decorator
* Can probably add rules with Flask's `add_url_rule`, but automatic rules will still exist, with potentially unforeseen consequences
* Common code can be run before a request at the app level with `Flask.before_request`[1], at the class level with `before` and at the view level with `before_<view>`.

Flask Views:
* Class based views supporting one view per class (MyView.as_view()), similar to Django's CBVs
* Can add routes with add_url_rule
* MethodView is good for CRUD APIs with GET, POST, etc verbs, but doesn't fit with CKAN's current controller structure

Flask view functions:
* Structure with modules and functions
* Perhaps use something like Flask-Via[2] to centralise routing?
* Common code to run before a request can be run with `Flask.before_request`[1]

I'm leaning toward Flash-Classy, and not being too hung up on centralising the routing into a `routing.py` file.
	

[1] http://flask.pocoo.org/docs/0.10/api/#flask.Flask.before_request
[2] Flask-Via: http://flask-via.thisissoon.com/

amercader:
Regarding Flask and translations, unfortunately the migration won't be straightforward.

The default i18n solution for Flask is [Flask-Babel](https://pythonhosted.org/Flask-Babel/), but it has two main issues.
Annoyingly, it expects the translations to be on a `translations` folder in a specific location and there is no way to configure it. 

The main problem though is how it handles strings marked for translation that have parameters, more particularly the ones that have the `%s` string formatters, for example:

```
_('Action name not known: %s') % action_name
_('Tag "%s" length is less than minimum %s') % (value, MIN_TAG_LENGTH)

```


Frustatingly, contrary to Pylons, Django and Python's own [gettext](https://docs.python.org/2/library/gettext.html#gettext.gettext) implementations, Flask-Babel performs a string substitution when calling the `_` function, so you always have to pass the substitution parameters when calling the function. It expects something like:

```
_('Action name not known: %(action_name)s', action_name=action_name)
_('Tag "%(tag_name)s" length is less than minimum %(tag_len)s', tag_name='value, tag_len=MIN_TAG_LENGTH)

```

The good news is that the way we generally handle params substitution works fine:


```
_('{actor} deleted the extra {extra} from the dataset {dataset}')
```

There are around [100](https://gist.github.com/amercader/77a14df30ea12ee6b67ba0bd0e662a22) "%s-style" strings left on the code base. Note that these are discouraged anyway as they don't allow different languages to move around the parameters depending on their syntax. 

So what should we do?

1. Update all the above strings to have named parameters (as in `{actor} deleted`) + move i18n files to `translations` + use Flask-Babel

2. Port Pylons' [i18n](https://github.com/Pylons/pylons/blob/master/pylons/i18n/translation.py) module to CKAN code base

3. Roll our own based on Python gettext

4. ?

I think even if it's a bit of a pain, option 1 is probably the best one long term.

There is also the slight concern that Flask-babel doesn't seem to be very [actively maintained](https://github.com/python-babel/flask-babel/issues/87), but there's not much we can do about this and in any case it will be defintely better that Pylons 0.9.7.

Any thoughts?

TkTech:
@amercader Funny enough, I will probably be flask-babel's new maintainer. I ran into issues with it and Notifico so I gave mitsuhiko a shout. You can just pull [this PR](https://github.com/python-babel/flask-babel/pull/88) to get up to date.

My vote is for option #1 :)

wardi:
@TkTech sounds great. Are you willing to remove the `%s` munging?

TkTech:
@wardi Most likely not, you really shouldn't be doing it. Don't want to make it *easier* to support bad practices.

@amercader Regarding your first complaint, I've just pushed a commit that adds a new config option, "BABEL_TRANSLATION_DIRECTORIES", which is a `;` separated list of directories to load translations from.

wardi:
@TkTech I don't see a problem with using `%s` in a translation string if there's only one parameter. But the real issue IIUC is that flask-babel is swapping `{}` in for `%s` even when you haven't asked it to do any parameter substitution. That just seems like a broken gettext implementation to me.

Otherwise we have to convert every `%s` to `{}` in every translated string (multiplied by every language translated), convert every `%` to `.format()` in our code base. And ask every extension to do the same.

TkTech:
@wardi Sorry, I see what you mean now. I misunderstood your comments.
@wardi / @amercader This behaviour is now fixed in https://github.com/python-babel/flask-babel/pull/88/commits/fec2ef76893b5686814ee4434c9f11833e6dc071 - please let me know if it's working for you now.

That said, within CKAN itself and extensions under its umbrella, switching to named parameters should probably be a goal somewhere along the line.

wardi:
@TkTech awesome! Also your plan to update the `%s` values in our code manually to meaningful names and on transifex with some code using its API sounds great. I hope I overestimated the effort involved.

TkTech:
@wardi / @amercader flask-babel version 0.10 is now on pypi and includes 4 years of fixes :)

amercader:
@TkTech Amazing to hear this and incredible good timing! I'll pull the latest version and recheck the issues but I'm now pretty much convinced about going with option 1.

amercader:
Down to 4 test failures!

* `ckan.tests.legacy.functional.api.test_dashboard.TestDashboard.test_10_dashboard_activity_list_html_does_not_crash`
  Crashes because it internally calls `lib.base.render`, which is Pylons only

* `ckanext.example_itranslation.tests.test_plugin.TestExampleITranslationPlugin.test_translated_string_in_extensions_templates`
  Fails because translations don't work (yet) on Flask

* `ckanext.datastore.tests.test_search.TestDatastoreSQL.test_not_authorized_to_access_system_tables` and `ckanext.datastore.tests.test_search.TestDatastoreSQL.test_self_join`
  Fail because Flask has trouble parsing arguments which include `=`, eg http://demo.ckan.org/api/action/datastore_search_sql?sql=SELECT * FROM "af807528-6c72-4cc7-a276-28a3752f8383" WHERE "Estacion" = 'xx'

brew:
Added a separate poc branch (https://github.com/ckan/ckan/compare/poc-flask-views...poc-flask-views.blueprint) demonstrating a way to register plugins as Flask Blueprints. This includes tests for adding a unique url rule, a url rule to override one defined in pylons core, and a rule to override one defined in flask core.

amercader:
4a12fc41d gets rid of Flask-classy and implements the API view as a blueprint. I think it's a nice and logic approach to have a blueprint per Pylons controller, as these are quite big (essentially mini-apps) and have their own endpoint (eg `/dataset`, `/user`, `/api`, etc)

For now it follows a very basic structure:

```
# Imports and config vars
...

# Blueprint definition

api = Blueprint('api', __name__, url_prefix='/api')

# Private methods

def _identify_user():
   ...

# View functions

def action(logic_function, ver=API_DEFAULT_VERSION):
    ...

def get_api(ver=1):
    ...

# Routing

api.add_url_rule('/', view_func=get_api, strict_slashes=False)
api.add_url_rule('/action/<logic_function>', methods=['GET', 'POST'],
                 view_func=action)
api.add_url_rule('/<int(min=3, max={0}):ver>/action/<logic_function>'.format(
                 API_MAX_VERSION),
                 methods=['GET', 'POST'],
                 view_func=action)

```

I like the idea of having all routing defined in a single place, but of course ideally it'd need to be on top of the file so you get a quick overview of it when you open it. This does not work as it stands now because view functions are not yet defined.

Also some of the current private methods will need to be shared among different view functions from different  blueprints. We'll focus on that next using the `before_request`, etc functions.

amercader:
Apologies for the long text!

One major thing we need to work around is URL generation with `url_for`.

There are several things to consider. The most obvious is that Pylons (or Routes) and Flask `url_for` accept different parameters.

eg this:


```python
url_for(controller='api',
          action='action',
          logic_function='help_show',
          ver=ver,
          name=logic_function,
          qualified=True,
)
```

Needs to be changed to this (assuming we use blueprints):

```python
url_for('api.action',
        logic_function='help_show',
        ver=ver,
        name=logic_function,
        _external=True,
)
```

Up until f877337 tests were passing because even though the API controller had moved to Flask the old Pylons routes and controller hadn't been removed, so the "old" URL generation still worked. Once these are removed Pylons can not generate routes to the new view functions.

We can:

1. Keep the old routes and controllers (maybe empty the functions to make clear they are deprecated)
2. Update all `url_for` imports to a CKAN wrapper (there are a lot of places that do `from routes import url_for`) and in there try to convert the Routes params to something that Flask understands, try if Flask can generate the URL and fallback to Pylons if not
3. ...?

I had a quick go with option 2, modifying the `url_for` that we already customize in helpers: https://gist.github.com/amercader/15c50fd81c3067d1b4dfbd24d4c7beca
This almost certainly does not cover all cases but eventually I guess we could make it work.

The other main problem is using `url_for` in tests. Right now the Routes `url_for` is used all over the place, both in legacy and new tests. If you try to use Flask's `url_for` (or our helpers wrapper that calls it) from the tests you will get `RuntimeError: Attempted to generate a URL without the application context being pushed. This has to be executed when application context is available.`

In order to get it working you need to change this:

```python
    def test_config_option_list_access_sysadmin_jsonp(self):
        ...
        url = url_for(
            controller='api',
            action='action',
            logic_function='config_option_list',
            ver='/3')
        app = self._get_test_app()

        app.get(
            url=url,
            ...
```

To this:

```python
    def test_config_option_list_access_sysadmin_jsonp(self):
        ...
        app = self._get_test_app()
        flask_app = self._find_flask_app(app)
        with flask_app.test_request_context():
            url = url_for(
                controller='api',
                action='action',
                logic_function='config_option_list',
                ver='/3')

            app.get(
                url=url,
                ...
            )

            app.get('/about')

```

(This assumes that we are using the wrapper I mentioned before that translates Routes params to Flask)

Note that all further calls to the app need to be inside that context otherwise we run into the `RuntimeError`.

I'm not particularly looking forward to rewrite hundreds of tests in this way, so any pointers to possible solutions would be appreciated.

TkTech:
@amercader instead of:

```
api.add_url_rule('/', view_func=get_api, strict_slashes=False)
```

Use:

```
@api.route('/', strict_slashes=False)
def get_api(ver=1):
```

amercader:
@TkTech I know about the route decorators, the idea was to have all routes defined in central place, ideally on top of the file so it was easy to get an overview of the endpoints, but I agree the the route next to the view function might be preferable to all routes defined on the bottom of the file.

ckan/common.py
```diff
@@ -6,12 +6,107 @@
 # NOTE:  This file is specificaly created for
 # from ckan.common import x, y, z to be allowed
 
+import flask
+import pylons
 
-from pylons.i18n import _, ungettext
-from pylons import g, c, request, session, response
+from flask.ext.babel import gettext as flask_gettext
```
k-nut: Importing in the style of `flask.ext.babel` is deprecated in flask 0.11[1] in favor of `flask_babel`. Maybe we should use the new format directly so that we are not a bad example and do not have problems later?! 
[1] https://github.com/pallets/flask/issues/1135

amercader:
All work done on this branch has been cleaned up and improved in separate PRs: https://github.com/ckan/ckan/issues/3196

PR #2984, opened 2016-04-27
[#2952] Fix issue when password contain special characters that need escaping

### LINKED ISSUE #2952: Escape special characters in password while db loading
CKAN doesn't escape special symbols within password.

Restore procedure:

```
paster db load -c /etc/ckan/default/production.ini my_database_dump.sql
```

in my case produces the following invalid command:

```
PGPASSWORD=Ab1239idap&u && psql -U ckan_default -h localhost -d ckan_default -f /mnt/portal/tmp/ckan.pg_dump
[1] 26628
-bash: u: command not found
```

dancastellani:
Fixes #2952

### Proposed fixes:
- Changed the method `_run_cmd()` to instead of run the cmd with `shell=True` that needs scaping, runed with `shell=False` (default) so [`subprocess.call`](https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module) takes care of scaping what is needed.
- instead of passing the password to pg_dump and load in the command passed as environmental variable.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -255,36 +255,43 @@ class ManageDb(CkanCommand):
             raise AssertionError('Expected postgres database - not %r' % self.db_details.get('db_type'))
         pg_cmd = command
         pg_cmd += ' -U %(db_user)s' % self.db_details
+        env = {}
         if self.db_details.get('db_pass') not in (None, ''):
-            pg_cmd = 'export PGPASSWORD=%(db_pass)s && ' % self.db_details + pg_cmd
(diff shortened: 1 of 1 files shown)


wardi:
as I said in #2952 I'd rather see these commands removed and direct users to use the postgres tools directly. This won't be the last time we need to fix or extend this command if we keep it in ckan.

TkTech:
Likewise, this adds nothing on top of what psql can already do for you. Just more to maintain.

PR #2986, opened 2016-04-28
avert sqlalchemy warning on IN clause operating on empty set

### LINKED ISSUE #2740: Sqlalchemy warning during tests (expensive non-query).
I'm seeing the following warning when running the tests..

```
SAWarning: The IN-predicate on "group.name" was invoked with an empty sequence. 
This results in a contradiction, which nonetheless can be expensive to evaluate.  
Consider alternative strategies for improved performance.
```

which doesn't seem like a big issue, and may just be test specific.  But it might be worth tracking down to ensure we don't have an unnecessary expensive query that isn't guarded against in the action layer.

Maybe good-for-contribution?

sklump:
Fixes #2740 

### Proposed fixes:
On package search, only look for matching groups in database if there are any to match, averting sqlalchemy warning 'The IN-predicate on ... was invoked with an empty sequence. This results in a contradiction, which nonethless can be expensive to evaluate.  Consider alternative strategies for improved performance.'

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1895,9 +1895,11 @@ def package_search(context, data_dict):
     for field_name in ('groups', 'organization'):
         group_names.extend(facets.get(field_name, {}).keys())
 
-    groups = session.query(model.Group.name, model.Group.title) \
-                    .filter(model.Group.name.in_(group_names)) \
-                    .all()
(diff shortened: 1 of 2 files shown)


sklump:
Still learning git. Must resubmit to separate from upcoming doc pull request. Thanks for your patience.

PR #2995, opened 2016-04-30
Fix + enable automatic docker image builds

deinok:
Little fix for docker Files
Please enable automatic building on Docker Hub.

It uses the official Postgres image, fixes the issue with Solr Image

Tested


contrib/docker/solr/schema.xml
```diff
@@ -1 +0,0 @@
-../../../ckan/config/solr/schema.xml
```
wardi: why are you duplicating this file now? For the solr schema we need a way to keep this file up to date with the current ckan solr schema, but also allow users to customize their solr schema.

How about using a volume shared between the ckan docker image and the solr one?

deinok:
@wardi for now, the best way is download the schema from GitHub.
On next pulls i will try to use official images of Solr, and try to simplyficate the code.
Thanks for your attention

deinok:
Try it on docker with deinok/postgresql deinok/solr and deinok/ckan

wardi:
@deinok it sounds like you're interested in helping us maintain these files, that would be really appreciated as none of the current core developers are using them.

deinok:
@wardi Okey, in our work we are interested in CKAN, and we are going to use it in a project. It would be fantastic mantain the Docker files and improve them with more options.

PR #2998, opened 2016-05-03
fix redirect with nonroot take 2

wardi:
#2968 with fixes

### DIFF
--- a/ckan/config/middleware.py
+++ b/ckan/config/middleware.py
@@ -198,6 +198,8 @@ def make_pylons_stack(conf, full_stack=True, static_files=True, **app_conf):
     if asbool(config.get('ckan.tracking_enabled', 'false')):
         app = TrackingMiddleware(app, config)
 
+    app = RootPathMiddleware(app, config)
+
     return app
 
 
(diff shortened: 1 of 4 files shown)


MrkGrgsn:
@wardi Is it safe to alter `SCRIPT_NAME`? I would think changing an environment variable could break other middleware?

wardi:
@MrkGrgsn you're right this change inserts the middleware around all the other middleware. I thought we were inserting the middleware at the other end of the stack. We should fix that.

I also agree in general this is a hack, but the alternative was to fork routes so that url_for would work properly with a root_path setting

wardi:
for backporting: related changes include
- https://github.com/ckan/ckan/commit/bf106b0
- https://github.com/ckan/ckan/commit/a76da0c
- https://github.com/ckan/ckan/commit/9cf0369
- ... these commits ...
- https://github.com/ckan/ckan/commit/16a81926b

amercader:
@wardi I know we've gone over this, but just want to double check if you think this is safe for backporting, seeing as it involves changes down to the middleware level.

Also if that's the case that all the commits from different PRs in the above comment should be backported.

wardi:
well, now that we're on 2.5 + these changes I'm still seeing multiple different redirect issues, it's probably better to leave it out until I can track down the new problems.

amercader:
cool, I won't backport it then

PR #3000, opened 2016-05-05
Slow page rendering when using lots of snippets

wardi:
update the the debug snippet to not render all variables passed to every snippet.

This is especially painful when passing a complete package data_dict through to scheming display snippets. The complete dictionary is rendered as HTML on every single snippet rendered, adding massively to the size of the page rendered.


wardi:
@amercader thanks! it seems that the variables I removed aren't mentioned in the docs, so no change there is necessary

PR #3004, opened 2016-05-09
Upgrade Babel version

amercader:
Another prehistoric library version that we are using (0.9.6 vs 2.3.4). There are issues around the language selector and how URLs are built. Essentially what happens is that comparing `Locale('ca') == 'ca'` used to work, but not any more (probably for the best). 

We will need to upgrade in order to use Flask-babel anyway (see discussion [here](https://github.com/ckan/ckan/pull/2971#issuecomment-215400251))


amercader:
I also tested manually the whole translation file update process from Transifex and everything seems to work fine.

amercader:
@TkTech can you have a look at this as it relates to flask-babel?

amercader:
@TkTech ping, would be really useful or the Flask views POC branch to get this reviewed and merged if you are happy (this is not related to https://github.com/python-babel/flask-babel/pull/91 btw)

PR #3008, opened 2016-05-10
retrieve the title of a package in a selected language rather than in …

yingqiaogit:
Fixes #

### Features:

- [ ] includes API changes

…the form of a dictionary when a multi-language plugin is used. With our release-v2.4-ndm branch using the multi-language plugin, the UI front displays the title of a dataset as a dictionary in the page of package edit or history. The page of package read displays the title correctly in the selected language. In this update, the multi-language plugin in "package_show" is enabled by setting the context['for_view'] as true for the cases of package edit and history. 
We have tested the update in our release-v2.5-ndm branch. The title is displayed correctly. We are propagating the change to the master branch now. At the UI front, we do not see any change. 
There are other functions, for example, new resource, follow/unfollow, followers, groups, calling the package_show without setting the context['for_view'] field. Since the package edit function calls package_show with context['for_edit'], should we differentiate the edit function with others in the package_show instead of adding context['for_view'] in lots of places? For instance, the multi-language plugin is called in the case that context['for_edit'] is false.

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -420,7 +420,8 @@ class PackageController(base.BaseController):
                 h.redirect_to(controller='revision', action='diff', **params)
 
         context = {'model': model, 'session': model.Session,
-                   'user': c.user, 'auth_user_obj': c.userobj}
+                   'user': c.user, 'auth_user_obj': c.userobj,
+                   'for_view': True}
         data_dict = {'id': id}
         try:
             c.pkg_dict = get_action('package_show')(context, data_dict)
@@ -747,7 +748,9 @@ class PackageController(base.BaseController):
         if context['save'] and not data:
             return self._save_edit(id, context, package_type=package_type)
         try:
-            c.pkg_dict = get_action('package_show')(context, {'id': id})
+            c.pkg_dict = get_action('package_show')(dict(context,
+                                                         for_view=True),
+                                                    {'id': id})
             context['for_edit'] = True
             old_data = get_action('package_show')(context, {'id': id})
             # old data is from the database and data is passed from the


amercader:
@yingqiaogit Thanks for this patch. I was able to reproduce this. Just to be clear about the context of this:

* This happens on all CKAN versions including master when using the multilingual plugin included in CKAN core (If you are dealing with multilingual metadata I strongly suggest you consider using [ckanext-fluent](https://github.com/ckan/ckanext-fluent), I'm sure @wardi can help you get started :) )

* The symptom is that the multilingual fields are not translated on all pages. For instance, the title on the breadcrumbs appears translated on the dataset read page:
screenshot from 2016-05-12 16 32 49

  but not on the dataset edit page (and history, and others):
 screenshot from 2016-05-12 16 32 38

* The cause is that on these particulars controller calls, the `for_view` key is not passed to the context of the `package_show` call, and so the [`before_view`](https://github.com/ckan/ckan/blob/master/ckanext/multilingual/plugin.py#L317) hook of the multilingual plugin is not called.

This `for_view` behaviour is not great and is probably something that we won't be replicating on newer versions of the controllers so rather than adding some extra logic around `for_edit` like you suggest I'd prefer that we explicitly passed the `for_view` key to `package_show` on the necessary methods.

Let me know if you want to add the `for_view` key to more methods on this PR, otherwise I'll merge it as is.

yingqiaogit:
@amercader Thank you for your detail explanation. I have discussed with @wardi about your suggestion. And, we agree that we will add for_view key to more cases for the front UI later. Processing only the for_edit case in package_show might not be a good idea since some API's controllers may call package_show as well. 
Yes, please merge the update now since it includes a bug fix for the package edit case. I will continue to work on this problem later.

PR #3010, opened 2016-05-11
Allow filtering on group type when listing orgs

Psykar:
Updating actions so organizations can have a group type that is not 'organization'.
This allows you to use IGroupForm to implement custom organization types.

The documentation already implies this is possible (and it is mostly, this fixes some actions which had organization hardcoded, regardless of what is passed in)

### Features:

- [x] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

PR #3015, opened 2016-05-12
Add missing routes to IGroupForm

Psykar:
IGroupForm is missing two routes:
_about
_bulk_process

This adds them (and adds a new example plugin with some tests around IGroupForm's with organizations)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/group.py
+++ b/ckan/controllers/group.py
@@ -380,10 +380,6 @@ class GroupController(base.BaseController):
         group_type = self._ensure_controller_matches_group_type(
             id.split('@')[0])
 
-        if group_type != 'organization':
-            # FIXME: better error
-            raise Exception('Must be an organization')
-
         # check we are org admin
 
         context = {'model': model, 'session': model.Session,
@@ -400,6 +396,10 @@ class GroupController(base.BaseController):
             c.group = context['group']
         except (NotFound, NotAuthorized):
             abort(404, _('Group not found'))
+
+        if not c.group_dict['is_organization']:
+            # FIXME: better error
+            raise Exception('Must be an organization')
 
         #use different form names so that ie7 can be detected
         form_names = set(["bulk_action.public", "bulk_action.delete",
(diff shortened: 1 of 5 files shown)


Psykar:
Hrm two of my unit tests depend on #3010 

Worth combining the two PRs? I feel that #3010 will be much easier to accept in it's current state, where as this has some larger changes.

amercader:
The bulk_process and about stuff makes perfect sense, but why do you need to extend the `GroupController` and `OrganizationController` from your plugin? You should only need to extend the `DefaultGroupForm` and `DefaultOrganizationForm` which are already available on the toolkit, and set the appropriate value on their [`group_type`](https://github.com/ckan/ckan/blob/master/ckanext/example_igroupform/plugin.py#L30) method.

This should be of course documented or at least the method should be added to `DefaultOrganizationForm`. I'm not saying that this is working now, but it's how it should be, even if we need to do changes on the controllers.

Psykar:
Kind of?

However, I have needed them myself when adding these custom group types as top level menu items, as `_link_active` in helpers ( https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L483 ) checks if the controller instance and action is active, but if you have multiple group types using the same controller, then both will be active even though you're only viewing one.
One option would be to fix this, but the simpler is to just have a separate controller for each menu item / custom organization / group type (which makes sense)

If you'd prefer to not have these controllers added to the plugin toolkit though, I'll happily remove them (and update the example extension to call `.add_group_type` on the controller) - let me know and I'll update the PR.

amercader:
To be honest, adding these controllers to the toolkit would probably open a can of worms (why not add X as well?). If the `_link_active` issue is not easy to fix, please update the example as you suggested.

Psykar:
No worries, I'll remove them then (extension can always just import the controllers directly even if it's not best practise)
I'll get this done tomorrow sometime :)

Psykar:
@amercader should be good to go.

Psykar:
Missed delete - added in now.

PR #3016, opened 2016-05-12
#2373: Do not validate packages or resources from database to views

### LINKED ISSUE #2373: Dataset modified timestamp gets updated even if nothing was changed in dataset
If user clicks update dataset without actually changing anything, it will update the modified timestamp. This itself is not a big problem, but since it will not update solr index, the timestamps won't match in package_show https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L948

This then triggers package_dictize which won't use custom schemas and before_view for datasets gets extras as key value pairs without the schema. 

For example multilingual gets different dataset_dict and makes ignoring specified attributes difficult.

Zharktas:
Fixes #2373 

### Proposed fixes:

Package read and resource read expect unvalidated dicts for templates. This context change skips validation and hence solves this problem with certain extensions for example scheming and fluent since these produce dicts instead of just strings if validated.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -344,7 +344,7 @@ class PackageController(base.BaseController):
     def read(self, id):
         context = {'model': model, 'session': model.Session,
                    'user': c.user, 'for_view': True,
-                   'auth_user_obj': c.userobj}
+                   'auth_user_obj': c.userobj, 'validate': False}
         data_dict = {'id': id, 'include_tracking': True}
 
         # interpret @<revision_id> or @<date> suffix
@@ -1060,7 +1060,8 @@ class PackageController(base.BaseController):
         context = {'model': model, 'session': model.Session,
                    'user': c.user,
                    'auth_user_obj': c.userobj,
-                   'for_view': True}
+                   'for_view': True,
+                   'validate': False}
 
         try:
             c.package = get_action('package_show')(context, {'id': id})


wardi:
@Zharktas I don't understand. The scheming and fluent templates do expect the validated dict on my sites. Also can you tell me how this relates to #2373 ?

Zharktas:
@wardi If package or resource is updated without modifying anything, the timestamp gets updated in model but not in solr as per #2373. This results to that the package or resource is fetched from the database and validated in [here](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L971) and [here](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L997), instead of just returning it from solr.

This in turn results into that package read and resource read templates get multilingual dicts with scheming and fluent instead of strings with some language. 

This gives error 500 at least in our instance with scheming and fluent since [this](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L1833) tries to strip whitespace from markdown field (for example description) but fails since the field is a dict instead of string.

Ourdataset schema is [here](https://github.com/vrk-kpa/api-catalog/blob/master/ckanext/ckanext-apicatalog_scheming/ckanext/apicatalog_scheming/schemas/dataset.json)  if its any help to understand the issue since last time we discussed this, it was quite hard to verify the problem itself.

wardi:
Sounds like there are a bunch of things going on. For one, we've decided to not officially support using fluent on core fields, see the discussion in #2678 

#2373 is certainly a problem we should fix.

I don't quite understand how you're getting the non-validated version of resources sent to your templates normally. That's not what I expect at all.

The lines that you point to look like they would cause the unvalidated data to be passed through when the data exists in solr but is out of date. That's a bug, but not the one you're describing.

Zharktas:
Did not actually know that fluent is not officially supported in core fields since someone else implemented before me in this project.

So if the validated package and resource is desired outcome for read templates, even though it will break our instance for now, somewhere at least partly non-validated package gets stored in solr. SInce our instance should break in all cases, instead of just this one. In our instance when the package if fetched from solr, core fields are always strings which appers to be non-validated.

I have to try this PR in our other project, where the #2373 was originally created. We don't use scheming there and it was more of cosmetic issue but this might fix it or reveal bunch of problems.

Zharktas:
Tested this in our other project which only uses multilingual for translations, if validation is skipped, the whole translation in multilingual is skipped, so it appears to be fluke in the other instance that this actually works. 

So this PR should not be in the core, so I'm closing this.

PR #3018, opened 2016-05-13
[WIP/RFC] Add group type to search facets.

Psykar:
This is somewhat of a Request For Comment / WIP.

I'm looking at separating search facets by group type. This appears to be the easiest way, and shouldn't much if any overhead. I can't see an easy way to do the split on the results page unfortunately, but perhaps adding this info to the facet won't be frowned on, even if it's only so the info can be available for extensions / helpers / templates.

Currently I'm doing the display of the group type facets in an extension, manually splitting the organizations facet returned by solr based on the type this PR adds, but it requires hardcoding the group types (by overriding `dataset_facets`, `before_search` and the `get_facet_items_dict` helper)

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1885,17 +1885,17 @@ def package_search(context, data_dict):
         'sort': data_dict['sort']
     }
 
-    # create a lookup table of group name to title for all the groups and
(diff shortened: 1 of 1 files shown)


kmbn:

PR #3020, opened 2016-05-13
[#3012] Persist datastore_active on resource updates

### LINKED ISSUE #3012: "Data API" button disappears on resource page after empty update
ckan_version: 2.5.2

When updating a resource without changing any fields, url or description, the green "Data API" button should still be there, because the data is still in the datastore (And API calls of this type confirms that it still is). Not sure if it is expected of the datapusher to trigger its update or not. 

The green "Data API" button disappears, as if data is now removed from the datastore for the particular resource. It would seem like CKAN thinks this triggers the datapusher hook, even though it does not (logs confirm it nothing happens).

Push .CSV file to the datastore using the web interface, then update the resource using either the `resource_update` API call or the web interface, without changing anything.

I'm curious to see if i'm the only one with this issue, i can reproduce over several CKAN installations running 2.5.2. If i need to post this in the datapusher github issue tracker instead, close this and i will post it there instead.

amercader:
Fixes #3012

As datastore_active is stored as an extra, if it's not provided in the
incoming data_dict on resource_update it will get dropped. This happens
most noticeably when udpating via the frontend, causing the green Data
API button to disappear.

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -80,6 +80,11 @@ def resource_update(context, data_dict):
         log.error('Could not find resource %s after all', id)
         raise NotFound(_('Resource was not found.'))
 
+    # Persist the datastore_active extra if already present and not provided
+    if ('datastore_active' in resource.extras and
+            'datastore_active' not in data_dict):
+        data_dict['datastore_active'] = resource.extras['datastore_active']
+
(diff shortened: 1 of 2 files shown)


wardi:
Feels like datastore_active should be a core field now that we have a special case for it. Fix is fine as is, though.

PR #3022, opened 2016-05-13
2844 optional resource url

yingqiaogit:
Fixes #

### Features:

- [ ] includes API changes

The update contains a test case that tests the resource_show API on a resource with empty url. We tested the update by running the CKAN tests. All of the test cases passed.


wardi:
@yingqiaogit the legacy test directory contains old tests we're trying to get rid of, would you resubmit with a test in the style of https://github.com/ckan/ckan/blob/master/ckan/tests/logic/action/test_create.py#L371-L414 (maybe just add a method to that class)

PR #3024, opened 2016-05-16
Add a submit_all command to datapusher.

CarlQLange:
### Proposed fixes:
This allows you to add every resource of every package to the datastore.
this is useful if you're setting up datastore for a a ckan that's
already got datasets. Also renames the existing submit_all function to resubmit_all, because
it would not act on resources that had not already been submitted to
datastore.

### Features:

- [ ] includes API changes


CarlQLange:
Unfortunately I'd need some guidance on how to write a test for this, if it's necessary to do so.

wardi:
Don't think this needs a test, but you should fix the pep8 error you've created.

PR #3027, opened 2016-05-17
Fix #2331

### LINKED ISSUE #2331: KeyError 'resources' in error handler
https://github.com/ckan/ckan/blob/master/ckan/logic/action/create.py#L305

Extension-code dependent, it can happen that a perfectly normal validation error leads to this block and CKAN throws an exception concerning a missing resources key when there is no resources key in the dictionary. This can lead to serious confusion for a developer.

CarlQLange:
Fixes #2331
### Proposed fixes:

No longer causes a 500 when trying to get the validation errors. If there's no 'resources' key, just prints the base error_dict .
### Features:
- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -300,8 +300,10 @@ def resource_create(context, data_dict):
         _get_action('package_update')(context, pkg_dict)
         context.pop('defer_commit')
     except ValidationError, e:
-        errors = e.error_dict['resources'][-1]
-        raise ValidationError(errors)
+        try:
+            raise ValidationError(e.error_dict['resources'][-1])
+        except (KeyError, IndexError):
+            raise ValidationError(e.error_dict)
 
     ## Get out resource_id resource from model as it will not appear in
     ## package_show until after commit
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -98,8 +98,10 @@ def resource_update(context, data_dict):
         updated_pkg_dict = _get_action('package_update')(context, pkg_dict)
         context.pop('defer_commit')
     except ValidationError, e:
-        errors = e.error_dict['resources'][n]
-        raise ValidationError(errors)
+        try:
+            raise ValidationError(e.error_dict['resources'][-1])
+        except (KeyError, IndexError):
+            raise ValidationError(e.error_dict)
 
     upload.upload(id, uploader.get_max_resource_size())
     model.repo.commit()


ckan/logic/action/create.py
```diff
@@ -300,8 +300,11 @@ def resource_create(context, data_dict):
         _get_action('package_update')(context, pkg_dict)
         context.pop('defer_commit')
     except ValidationError, e:
-        errors = e.error_dict['resources'][-1]
-        raise ValidationError(errors)
+        if ('resources' in e.error_dict):
+            errors = e.error_dict['resources'][n]
+            raise ValidationError(errors)
```
wardi: how about
```python
try:
    raise ValidationError(e.error_dict['resources'][-1])
except (KeyError, IndexError):
    raise ValidationError(e.error_dict)
```

that will catch the case of an error in a different resource than the one submitted too

ckanext/datapusher/cli.py
```diff
@@ -14,6 +14,9 @@ class DatapusherCommand(cli.CkanCommand):
                     ignoring if their files haven't changed.
         submit <pkgname> - Submits all resources from the package
                          identified by pkgname (either the short name or ID).
+        submit_all  - Submit every package to the datastore.
+                      This is useful if you're setting up datastore
+                      for a ckan that already has datasets.
```
wardi: unrelated change slipped in here?

CarlQLange:
Oh, shit, sorry! Based off the wrong commit :) I'll make the changes you suggested.

CarlQLange:
Fixes made, and removed hanging commit from another branch :)

amercader:
@CarlQLange this is great but we need a test, specially with a confusing issue with this one.
Also might be worth checking if this affects the displaying of errors on the UI? (eg if an error is related to a dataset field but we are on the resource form).

amercader:
@k-nut do you have commits related to this on your fork?

k-nut:
@amercader actually not. For some reason I added the number of this pull in my commit messages even though I was referring to another one. I probably got confused because both PRs came from @CarlQLange . The commits that are shown here were just merged yesterday with this commit: https://github.com/ckan/ckan/pull/3082

So it seems like that I do not have any chances related to this.

wardi:
@amercader how do you feel about merging this without the test?

For me it comes up when a validation rule has changed and existing package metadata is no longer valid, that's a hard thing to test for or even properly direct the user about how to fix. These sorts of errors might only be fixable with an API call e.g. if new rules break both package and resource metadata in the same dataset.

amercader:
@wardi yes, getting a test for this will be difficult. Do you think it's something worth backporting further back (2.6, 2.5, etc)?

wardi:
Yes let's backport it. It's a simple fix for 500 errors in some cases

PR #3029, opened 2016-05-17
Solr Docker Image Updated

deinok:
Updates the Solr Docker Image to 6.0 and CKAN Image requires less space in HD.

@wardi Please, enable Automatic-Build on Docker Hub as we talk. If you need help, please, contact me. This will enable the automatic update of the image every time the Dockerfiles are updated.


deinok:
@wardi ping

Dockerfile
```diff
@@ -1,61 +1,61 @@
...
-ADD ./contrib/docker/svc /etc/service
-CMD ["/sbin/my_init"]
-
-# Volumes
-VOLUME ["/etc/ckan/default"]
-VOLUME ["/var/lib/ckan"]
-EXPOSE 80
-
-RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+FROM phusion/baseimage:latest
```
wardi: let's pin a version instead. Otherwise our build won't be reproduceable

contrib/docker/solr/Dockerfile
```diff
@@ -1,23 +1,28 @@
...
-# Install CKAN Solr core
-RUN cp -R $SOLR_HOME/collection1/ $SOLR_HOME/ckan/
-RUN echo name=ckan > $SOLR_HOME/ckan/core.properties
-ADD https://raw.githubusercontent.com/ckan/ckan/master/ckan/config/solr/schema.xml $SOLR_HOME/ckan/conf/schema.xml
-
-EXPOSE 8983
-WORKDIR /opt/solr/example
-CMD ["java", "-jar", "start.jar"]
\ No newline at end of file
+FROM solr:latest
```
wardi: here too

contrib/docker/solr/solrconfig.xml
```diff
@@ -0,0 +1,1462 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
wardi: 1400 lines of solr config? I guess this is because the new default is now incompatible with ckan?

wardi:
I've deleted and recreated the ckan/solr and ckan/postgresql images as automated builds. It seems like the ckan/ckan image isn't building automatically even though it's set to. I might have to delete and recreate it as well.

Please pin the docker images we use for postgresql and solr should be to some specific release version and I'll merge. It's ok with me if our docker dependencies are ahead of the versions listed in the source install, because this is going to be bleeding edge for a while I'm sure.

contrib/docker/solr/solrconfig.xml
deinok: Yes, about 80% is comments, default solrconfig for solr 6.0 don't use the schema.xml, it uses a managed-schema
This modification enables the schema.xml

deinok:
@wardi 
Postgres  pined to 9.5
Solr pined to 6.0
Base image temporary pined  to 0.9.15 (probably we shoud move to ubuntu base images, phusion/baseimage was set when ubuntu images were not very stable for docker)

contrib/docker/solr/Dockerfile
```diff
@@ -1,23 +1,28 @@
...
+# User
+USER root
+
+# Create Directories
+RUN mkdir -p /opt/solr/server/solr/$SOLR_CORE/conf
+RUN mkdir -p /opt/solr/server/solr/$SOLR_CORE/data
+
+# Adding Files
+ADD ./solrconfig.xml \
+https://raw.githubusercontent.com/deinok/ckan/master/ckan/config/solr/schema.xml \
```
wardi: ckan/ckan ?

contrib/docker/solr/Dockerfile
```diff
@@ -1,23 +1,28 @@
...
+https://raw.githubusercontent.com/apache/lucene-solr/master/solr/server/solr/configsets/basic_configs/conf/currency.xml \
+https://raw.githubusercontent.com/apache/lucene-solr/master/solr/server/solr/configsets/basic_configs/conf/synonyms.txt \
+https://raw.githubusercontent.com/apache/lucene-solr/master/solr/server/solr/configsets/basic_configs/conf/stopwords.txt \
+https://raw.githubusercontent.com/apache/lucene-solr/master/solr/server/solr/configsets/basic_configs/conf/protwords.txt \
+https://raw.githubusercontent.com/apache/lucene-solr/master/solr/server/solr/configsets/data_driven_schema_configs/conf/elevate.xml \
```
wardi: can we pull from a tag corresponding to the pinned version instead of master?

contrib/docker/solr/Dockerfile
deinok: I forget do this change before PR

contrib/docker/solr/Dockerfile
deinok: yes, i will pine it to 6.0 also, i think we can delete the coments from solrconfig.xml

deinok:
@wardi Tested images, seems ok to merge, take a look please

Dockerfile
```diff
@@ -1,61 +1,61 @@
...
+ADD ./contrib/docker/my_init.d /etc/my_init.d
+ADD ./contrib/docker/svc /etc/service
+CMD ["/sbin/my_init"]
+
+# Volumes
+VOLUME ["/etc/ckan/default"]
+VOLUME ["/var/lib/ckan"]
+EXPOSE 80
+
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
wardi: You've switched this whole file to use DOS line endings?

wardi: contrib/docker/solr/Dockerfile and contrib/docker/docker-cloud.yml too.

deinok:
@wardi yes, i think i was modified, but its an workin version, I will full modify CKAN image in a next PR, it will enable more configuration using ENV and use Debian or Ubuntu images, now that images work better than phusion/baseimage.

Also i see that there is a Docker repo in CKAN, but is dead.

wardi:
Would you use unix line endings instead? Changing the line endings makes it look like you changed the whole files

deinok:
Okey, will make the changes when i get home

deinok:
@wardi take a look ;)

PR #3031, opened 2016-05-19
Move code path for group into explicit else

k-nut:
Fixes  -
### Proposed fixes:
When going over `ckan/controllers/feed.py` I noticed a small part of the code where readability could be easily improved. This might also make the code *slightly* faster since if `is_org` is `True` we save four assingments that would later be overwritten.

Here are my comments from the commit message:
- the functionality in the code remains the same
- The branch for the case that `is_org` is False was just moved into
  an explicit `else` for better readibility
- The code in the else branch was also slightly reindented to be more
  simliar to the if branch

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/feed.py
+++ b/ckan/controllers/feed.py
@@ -188,22 +188,24 @@ class FeedController(base.BaseController):
                                   action=group_type,
                                   id=obj_dict['name'])
 
-        guid = _create_atom_id(u'/feeds/group/%s.atom' %
-                               obj_dict['name'])
-        alternate_url = self._alternate_url(params, groups=obj_dict['name'])
-        desc = u'Recently created or updated datasets on %s by group: "%s"' %\
-            (g.site_title, obj_dict['title'])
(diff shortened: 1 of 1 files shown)


ckan/controllers/feed.py
```diff
@@ -188,24 +188,26 @@ def _group_or_organization(self, obj_dict, is_org):
...
                                    obj_dict['name'])
             alternate_url = self._alternate_url(params,
                                                 organization=obj_dict['name'])
-            desc = u'Recently created or  updated datasets on %s '\
+            desc = u'Recently created or updated datasets on %s '\
                 'by organization: "%s"' % (g.site_title, obj_dict['title'])
             title = u'%s - Organization: "%s"' %\
                 (g.site_title, obj_dict['title'])
 
+        else:  # is organization
```
brew: Should the comment here be `# is group`?

k-nut:
@brew I updated the comment. Well spotted.

PR #3033, opened 2016-05-19
Build Status badge in README should point to CircleCI, not Travis

brew:
We now use CircleCI for continuous integration, rather than Travis. We should change the Build Status badge in the README to point to the appropriate ckan project on CircleCI.


k-nut:
You could also give it `?style=shield` so that it has the same look as the other badges. The file format would also svg.
Build Status

PR #3037, opened 2016-05-20
Fix resource downloads to call IUploader plugins

Psykar:
This covers cases where IUploader changes the save path of an upload - it needs to use the same uploader to generate the path for downloading!

I tried to write a unit test, but it's a bit beyond me - requiring either mocking of call_application or some fancy handling of creating temporary files, which is probably not ideal.

I did notice the only unit test around resource uploading is actually commented out at the moment anyway...

### DIFF
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -1492,7 +1492,7 @@ class IUploader(Interface):
     upload resources and group images.
     '''
 
-    def get_uploader(self):
+    def get_uploader(self, upload_to, old_filename):
         '''Return an uploader object to upload general files that must
         implement the following methods:
 
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -1145,7 +1145,7 @@ class PackageController(base.BaseController):
             abort(404, _('Resource not found'))
 
         if rsc.get('url_type') == 'upload':
-            upload = uploader.ResourceUpload(rsc)
+            upload = uploader.get_resource_uploader(rsc)
             filepath = upload.get_path(rsc['id'])
             fileapp = paste.fileapp.FileApp(filepath)
             try:
--- a/ckanext/example_iuploader/plugin.py
+++ b/ckanext/example_iuploader/plugin.py
@@ -1,0 +1,21 @@
+import os
+
+from ckan import plugins
+from ckan.lib.uploader import ResourceUpload as DefaultResourceUpload
+
+
+class ExampleIUploader(plugins.SingletonPlugin):
+    plugins.implements(plugins.IUploader, inherit=True)
+
+    def get_resource_uploader(self, data_dict):
+        return ResourceUpload(data_dict)
+
+
+class ResourceUpload(DefaultResourceUpload):
+    path_prefix = 'filename_prefix_'
+
+    def get_path(self, id):
+        directory = self.get_directory(id)
+        filepath = os.path.join(
+            directory, '{}_{}'.format(self.path_prefix, id[6:]))
+        return filepath
--- a/ckanext/example_iuploader/test/test_plugin.py
+++ b/ckanext/example_iuploader/test/test_plugin.py
@@ -1,0 +1,121 @@
+import __builtin__ as builtins
+
+import paste.fileapp
+from mock import patch
+from nose.tools import (
+    assert_equal,
+    assert_is_instance
+)
+from pyfakefs import fake_filesystem
+from pylons import config
+from routes import url_for
+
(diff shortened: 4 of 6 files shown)


Psykar:
(I should mention, if a unit test is required to get this merged I can throw more time at it, but thought I'd throw this up first given how simple it is)

brew:
A test would be great, if possible. You can see I didn't have time to get fully into that resource upload test, which is why it's currently commented out. I'd love to get to grips with that, and assume it would involve a lot of mocking.

Psykar:
The main problem I have with writing a unit test is ckan.storage_path is not set in `test-core.ini`

A few possible solutions:

1. Update test-core.ini to have a storage_path set
  This option is currently the HEAD of this PR - but perhaps is not an ideal solution as it happens to break a few tests (which is fixable) due to the upload field now appearing on group and organization controllers. It sets the storage_path to be local to the working directory, and ignores it with gitignore

2. Import `ckan.lib.uploader._storage_path` and set it in the tests `_apply_config_changes`
  This option is rather ugly, but would be required as get_storage_path is called from the middleware ( see https://github.com/Psykar/ckan/blob/bugfix/resource-download-iuploader/ckan/config/middleware.py#L170 ) meaning that overriding the config isn't enough to update the _storage_path var stored in the uploader module ( see https://github.com/Psykar/ckan/blob/bugfix/resource-download-iuploader/ckan/lib/uploader.py#L48 )
  It also would require adding the check that exists in config.middleware to make the directory if it doesn't exist.

3. Add a new config var `ckan.uploads_enabled` so `lib.helpers.uploads_enabled` (which is what shows / hides the upload button in various places) doesn't have to rely on `uploader.get_storage_path` meaning we can have `ckan.storage_path` set in `test-core.ini` all the time, and turn uploads on / off by overriding `ckan.uploads_enabled` in tests `_apply_config_changes` 
  Would still need to handle creation of the test storage directory as well.

I lean towards 3, but would require adding a new config var... Thoughts @brew ?

brew:
I had a quick look at the commented out Package test yesterday, and got some of the way there by mocking the `uploads_enabled` call (which has to be done on the config object):

```python
@mock.patch.object(config['pylons.h'], 'uploads_enabled', return_value=True)
```

I like 1 or 3, but I'm not massively keen on having tests write files to disk, if we can work around it. Perhaps we can't without a lot of mocking.

brew:
Perhaps another option is to use pyfakefs to mock calls the a fake filesystem in memory: https://github.com/jmcgeheeiv/pyfakefs ?

Psykar:
Nice find with pyfakefs.

I think the solution then is
* Mock lib.helpers.uploads_enabled
* Inherit from pyfakefs.fake_filesystem_unittest.TestCase (and call setUpPyfakefs() in setup()
* Set config['storage_path'] in _apply_config_changes()

Will hopefully have this up shortly.

Psykar:
Probably a bit of cleanup left, but thought I'd put this up for a lookover. Seems to work OK.
I really dislike that you have to mock open across all modules at once, I couldn't find a workaround for this though.

If it looks OK I'll clean it all up properly.

(Note I couldn't use fake_filesystem_unittest.TestCase due to ckan not actually using TestCase, and the fact it mocks all filesystem calls, whereas we need some to hit the real filesystem for controllers / templates etc.

ckan/plugins/interfaces.py
```diff
@@ -1492,7 +1492,7 @@ class IUploader(Interface):
     upload resources and group images.
     '''
 
-    def get_uploader(self):
```
wardi: what's this interface change for?

ckan/plugins/interfaces.py
```diff
@@ -1492,7 +1492,7 @@ class IUploader(Interface):
...
+    def get_uploader(self, upload_to, old_filename):
```
amercader: Why are these new params needed? This will break existing extensions

brew: Is this a necessary change?

Psykar: The signature here was actually wrong.
When get_uploader is called on plugins, it's actually passed one arguments
see
https://github.com/ckan/ckan/blob/ed2196b31421365fc559eb8e0fa9141fa17a057f/ckan/lib/uploader.py#L24

Psykar:
Cleaned and rebased, should be good to go.

ckan/plugins/interfaces.py
Psykar: Hell, if you read the docstring it even mentions the arguments!

PR #3039, opened 2016-05-20
[#2981] Preserve sort order in groups on pagination

### LINKED ISSUE #2981: Group pagination does not preserve sort order
2.5.2

On group page, eg http://demo.ckan.org/group the pagination links at the foot of the page should preserve the sort order

The sort order is set to the default

Create 22 groups (currently available at http://demo.ckan.org/group)
Go to the group page
Change the sort order to "Name Descending"
Go to page 2

The group on page 2 should be the first alphabetically - instead it's the last

This has already been fixed for organizations - see https://github.com/ckan/ckan/pull/2141 and https://github.com/ckan/ckan/pull/2153

k-nut:
Fixes #2981

### Proposed fixes:
- Uses the same code that was introduced in #2153 for organizations
- Changes the tests to sort descending so that they actually make sure
  that the sort order is not changed
  - I did this because when I removed the sorting from the templates the tests would still pass. This is because they did not set the sort order to descending so they were just testing the old behavior but not actually the new one which should be preserving the order. This way we can make sure that is actually happening

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/group/index.html
+++ b/ckan/templates/group/index.html
@@ -34,7 +34,7 @@
     {% endif %}
   {% endblock %}
   {% block page_pagination %}
-    {{ c.page.pager() }}
+    {{ c.page.pager(q=c.q or '', sort=c.sort_by_selected or '') }}
   {% endblock %}
 {% endblock %}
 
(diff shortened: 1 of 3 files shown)


k-nut:
I just changed one of the tests to also include the empty `sort` and `q` url params that are now put into the pagination urls. All the tests are passing now.

amercader:
Thanks for the patch @k-nut

PR #3044, opened 2016-05-23
[#2760] change old password label if user is sysadmin

### LINKED ISSUE #2760: Changing user password in web interface
We are running CKAN 2.4.1.  There is a discrepancy between the documentation, and reality, and this discrepancy prevents a sysadmin changing/resting a users password via the web interface. 
  
Documentation http://docs.ckan.org/en/latest/sysadmin-guide.html#managing-users suggests that sysdmin just needs to enter a new password twice, and the users password will be changed.  
  
However, in CKAN 2.4.1 we are presented with an additional box, requesting the Old Password, as well as setting the new password. This box is redundant, and prevents the sysadmin maintaining the users record.  ie. if the user has forgotten their old password then they cannot tell it to the sysdmin to allow them to change the password, and if they remembered their old password then the user would not need to ask the sysadmin to change the password for them.  
  
Can the Old password box be removed from the manage user screen.  If the user can get to that screen, then clearly they know their own password as they are logged in already, and the sysadmin should not need to know a users old password.

k-nut:
Fixes #2760

### Proposed fixes:
The label on the edit user page now reads `Sysadmin password` if the user is a sysadmin or `Old password` if the user is not. This should clarify the situation in which the sysadmin wants to change another user's password. 

I am not sure if this is your standard way to change labels programatically. I am open for pointers on how to do it in a more fitting way.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -315,6 +315,10 @@ class UserController(base.BaseController):
         c.is_myself = True
         c.show_email_notifications = asbool(
             config.get('ckan.activity_streams_email_notifications'))
+        if authz.is_sysadmin(c.user):
+            vars['oldpassword_label'] = 'Sysadmin password'
+        else:
+            vars['oldpassword_label'] = 'Old password'
         c.form = render(self.edit_user_form, extra_vars=vars)
 
         return render('user/edit.html')
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -315,10 +315,6 @@ class UserController(base.BaseController):
         c.is_myself = True
         c.show_email_notifications = asbool(
             config.get('ckan.activity_streams_email_notifications'))
-        if authz.is_sysadmin(c.user):
-            vars['oldpassword_label'] = 'Sysadmin password'
-        else:
-            vars['oldpassword_label'] = 'Old password'
         c.form = render(self.edit_user_form, extra_vars=vars)
 
         return render('user/edit.html')
--- a/ckan/templates/user/edit_user_form.html
+++ b/ckan/templates/user/edit_user_form.html
@@ -24,7 +24,7 @@
 
   <fieldset>
     <legend>{{ _('Change password') }}</legend>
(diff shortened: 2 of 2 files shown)


ckan/controllers/user.py
```diff
@@ -315,6 +315,10 @@ def edit(self, id=None, data=None, errors=None, error_summary=None):
         c.is_myself = True
         c.show_email_notifications = asbool(
             config.get('ckan.activity_streams_email_notifications'))
+        if authz.is_sysadmin(c.user):
+            vars['oldpassword_label'] = 'Sysadmin password'
```
amercader: Please translate them here with ` _( )`  otherwise they won't be extracted.

k-nut:
@amercader I changed this so that it changes the label in the template. Would you merge it like this or should I rebase my three commits into one?

ckan/templates/user/edit_user_form.html
```diff
@@ -24,7 +24,7 @@
 
   <fieldset>
     <legend>{{ _('Change password') }}</legend>
-    {{ form.input('old_password', type='password', label=_('Old Password'), id='field-password', value=data.oldpassword, error=errors.oldpassword, classes=['control-medium'], attrs={'autocomplete': 'off'} ) }}
+    {{ form.input('old_password', type='password', label=_('Sysadmin Password') if c.is_sysadmin else _('Old Password'), id='field-password', value=data.oldpassword, error=errors.oldpassword, classes=['control-medium'], attrs={'autocomplete': 'off'} ) }}
```
wardi: this line is getting really long, I would split it up (jinja2 allows newlines between tokens in {{ }} blocks)

k-nut: Funny you say that. I actually had done that but then saw that it was not done in any of the other lines so reverted it. But I agree that it would be much more readable. Changing it back now.

k-nut:
@wardi this is split into multiple lines now.

PR #2948, opened 2016-04-11
add icons and css adoption

hknochi:
Fixes #
add a Icon to ckan/public/base/images/sprite-resource-icons.png (in ckan-style) to get shown an icon for image resources
add  rules in main.css to get shown the icon in UI (resource-view)

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [x ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

Issue #2910, opened 2016-03-13
Extending sprite resource icons with JPG/JPEG icons


earthquakesan:
### CKAN Version if known (or site URL)

2.5.0a
### Please describe the expected behaviour

Resources JPEG/JPG mimetype/format should have a specific icon on the dataset view (i.e. http://example.com/dataset/my-dataset) similar to XML/JSON/HTML icons from https://github.com/ckan/ckan/blob/master/ckan/public/base/images/sprite-resource-icons.png 
### Please describe the actual behaviour

Right now the icon is DATA (the first one on the sprite-resource-icons.png), which is the icon for generic resources.
### What steps can be taken to reproduce the issue?
- Install CKAN with datastore extension enabled
- Create example dataset
- Create a resource and upload JPG image to the resource

amercader:
@earthquakesan That sounds good, the problem is that the original Photoshop or GIMP file is not in the repo. Perhaps @aron still has it somewhere lying around. 
If any new formats can be added to the sprite somehow then it's just a matter of editing the relevant bits of the less file: https://github.com/ckan/ckan/blob/91e41b2e68faa3df5296a632f4862f5a55e69e62/ckan/public/base/less/dataset.less

aron:
@amercader just checked through my computer but couldn't find any design assets. I have a feeling we kept them on Google Drive or something similar...

earthquakesan:
@aron @amercader would it be possible to find and share them?

seebi:
@amercader What about the source of this? Did you find it?

amercader:
@seebi @earthquakesan Found them! @smth keeps a repo with them over here: https://github.com/mintcanary/file-types

hknochi:
this problem will be fixed by pull-request #2948

AnjaStemme:
Hi,

would it be possible to have more icons? For example as well for .doc .odt and separate image icons for png jpg gif?

smth:
I just added a the types suggested here (over at https://github.com/mintcanary/file-types), in case that's of any use to anyone.

PR #3048, opened 2016-05-23
Include icon for image data

k-nut:
Fixes #2948, #2910 

### Proposed fixes:
In order to clean up Pull Requests a little I just took the changes from #2948 and applied the to the lastest master. Of course this still does not include the original vector files as I do no have them.

### Features:

- [ ] includes API changes


k-nut:
In #2910 @amercader pointed out that the original files seem to be in this repository: https://github.com/mintcanary/file-types 
That repository does not include the new image-type icon though. Maybe @hknochi can tell us where he got it from or provide the original?

k-nut:
@amercader I think we should merge this rather soon instead of waiting for the original vector files. Waiting for them led to the last PR becoming conflicting with the current master. I could create a separate issue for also including the vector files?

hknochi:
@k-nut : i think to add the vector files in the repo won't be the best idea (may these should be in a other/single one)

k-nut:
Yes. We might want to store them in a separate repo. But at least having something like a link to the original somewhere in the ckan project could prove useful already.

hknochi:
okay i'll add the source . you can move them to a custom path afterwards

hknochi:
please take a look here: https://drive.eccenca.com/d/73541a96f2/

k-nut:
@hknochi cool, thank you! How are the icons licensed?

k-nut:
@amercader do you think we should create a new repository for design assets like this? Or is there something already where we could put this? Or would you just include it in this repository?

seebi:
Fromm eccenca's POV this should not be decided by us but by @smth who originally pushed these files (https://github.com/mintcanary/file-types/commit/1740cd625afc8f0d0f0c9506d41a1caaa3000cae). The work of @hknochi can be licensed under any terms suitable for you.

smth:
I just added a license [here](https://github.com/mintcanary/file-types/blob/master/LICENSE.txt).

k-nut:
Created a first [pull request](https://github.com/ckan/ckan-assets/pull/1) in the new `ckan-assets` repository. Once that is merged this PR should also be good to merge.

Issue #3006, opened 2016-05-10
str vs. unicode


torfsen:
In Python 2, `__repr__` and `__str__` must return instances of `str`. Returning `unicode` instances fails:
```
Python 2.7.6 (default, Jun 22 2015, 18:00:18) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> class Foo(object):
...     def __str__(self): return u'ö'
...     def __repr__(self): return u'ö'
... 
>>> str(Foo())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 0: ordinal not in range(128)
>>> repr(Foo())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 0: ordinal not in range(128)
```
I ran into this problem on my CKAN 2.5.2 installation due to tags containing non-ASCII characters, because `model.Tag.__repr__` [looks like this](https://github.com/ckan/ckan/blob/master/ckan/model/tag.py#L215):
```
def __repr__(self):
    return '<Tag %s>' % self.name
```
That expression will return either a `str` or a `unicode` instance, depending on the type of `self.name`.

I tried to fix that (and similar) errors by making sure that all implementations of `__repr__` and `__str__` always return `str` and `__unicode__` always returns `unicode`. However, while they do enforce the correct behavior, these fixes break a lot of other pieces of code which (implicitly) assumed the old behavior.

The problem is that there seems to be no rule whether `str` or `unicode` are to be used in a given context. Therefore, each part of the code which does string processing and object-to-string conversion needs to make sure it handles both `str` and `unicode` correctly. The errors that pop up after I fixed `__repr__`, etc. show that this is often not done correctly.

Before spending a lot of time fixing each of these little issues I want to make sure that I'm on the right track concerning how Unicode is handled in CKAN. Correct handling of Unicode in Python 2 is a non-trivial issue, and in my opinion CKAN would benefit greatly from a clear strategy. This is especially true with regard to an eventual migration to Python 3. In particular:

- What encoding should CKAN Python source files use (see [PEP 263](https://www.python.org/dev/peps/pep-0263/))?
- Do we want to use `unicode_literals` from [`__future__`](https://docs.python.org/2/library/__future__.html)?
- Do we use `str` or `unicode` internally?
[...]

PR #3049, opened 2016-05-23
[WIP] Consistent Unicode handling.

torfsen:
This is WIP for making CKAN's Unicode handling more consistent. See #3006 for background information.

Work on this PR is ongoing, but comments and feedback are always appreciated.

### DIFF
--- a/doc/contributing/unicode.rst
+++ b/doc/contributing/unicode.rst
@@ -1,0 +1,128 @@
+================
+Unicode handling
+================
+This document explains how Unicode and related issues are handled in CKAN.
+For a general introduction to Unicode and Unicode handling in Python 2 please
+read the `Python 2 Unicode HOWTO`_. Since Unicode handling differs greatly
+between Python 2 and Python 3 you might also be interested in the
+`Python 3 Unicode HOWTO`_.
+
+.. _Python 2 Unicode HOWTO: https://docs.python.org/2/howto/unicode.html
+.. _Python 3 Unicode HOWTO: https://docs.python.org/3/howto/unicode.html
+
+.. note::
+
+    This document describes the intended future state of Unicode handling in
+    CKAN. For historic reasons, some existing code does not yet follow the
+    rules described here.
+
+    *New code should always comply with the rules in this document. Exceptions
+    must be documented.*
+
+
+Overall Strategy
+----------------
+CKAN only uses Unicode internally (``unicode`` on Python 2). Conversion to/from
+ASCII strings happens on the boundary to other systems/libaries if necessary.
+
+
+Encoding of Python files
+------------------------
+Files containing Python source code (``*.py``) must be encoded using UTF-8, and
+the encoding must be declared using the following header::
+
+    # encoding: utf-8
+
+This line must be the first or second line in the file. See `PEP 263`_ for
+details.
+
+.. _PEP 263: https://www.python.org/dev/peps/pep-0263/
+
+
+String literals
+---------------
+String literals are string values given directly in the source code (as opposed
+to strings variables read from a file, received via argument, etc.). In
+Python 2, string literals by default have type ``str``. This has the risk of
+accidentially introducing ASCII strings and makes a future support of Python 3
+difficult (since string literals are interpreted as Unicode in Python 3).
+
+Therefore, the ``unicode_literals`` feature is used to make Python 2 treat
+string literals as Unicode::
+
+    from __future__ import unicode_literals
+
+This line should be the first code line in a Python file (after comments and
+blank lines). See the documentation on `future statements`_ for details.
+
+.. _future statements: https://docs.python.org/2/reference/simple_stmts.html#future
+
+Once ``unicode_literals`` has been imported, any string literal in the module
+will be converted to a ``unicode`` instance. However, some cases might require
+a ``str`` literal (for example for a filename). This can be achieved using the
+``b`` prefix (see `PEP 3112`_)::
+
+    from __future__ import unicode_literals
+
+    x = 'This is unicode'
+    y = b'This is str'
+
+.. _PEP 3112: https://www.python.org/dev/peps/pep-3112/
+
+
+Best Practices
+--------------
+
+Use ``io.open`` to open text files
+```````````````````````````````````
+When opening text (not binary) files you should use `io.open`_ instead of
+``open``. This allows you to specify the file's encoding and reads will return
+``unicode`` instead of ``str``::
+
+    import io
+
+    with io.open('my_file.txt', 'r', encoding='utf-8') as f:
+        text = f.read()  # contents is automatically decoded
+                         # to unicode using UTF-8
+
+.. _io.open: https://docs.python.org/2/library/io.html#io.open
+
+Text files should be encoded using UTF-8 if possible.
+
+
+Normalize strings before comparing them
+```````````````````````````````````````
+For many characters, Unicode offers descriptions. For example, a small latin
+``e`` with an acute accent (``é``) can either be specified using its dedicated
+code point (`U+00E9`_) or by combining the code points for ``e`` (`U+0065`_)
+and the accent (`U+0301`_). Both variants will look the same but are different
+from a numerical point of view::
+
+    >>> x = u'\N{LATIN SMALL LETTER E WITH ACUTE}'
+    >>> y = u'\N{LATIN SMALL LETTER E}\N{COMBINING ACUTE ACCENT}'
+    >>> print x, y
+    é é
+    >>> print repr(x), repr(y)
+    u'\xe9' u'e\u0301'
+    >>> x == y
+    False
+
+.. _U+00E9: http://www.fileformat.info/info/unicode/char/e9
+.. _U+0065: http://www.fileformat.info/info/unicode/char/0065
+.. _U+0301: http://www.fileformat.info/info/unicode/char/0301
+
+Therefore, if you want to compare two Unicode strings based on their characters
+you need to normalize them first using `unicodedata.normalize`_::
+
+    >>> from unicodedata import normalize
+    >>> x_norm = normalize('NFC', x)
+    >>> y_norm = normalize('NFC', y)
+    >>> print x_norm, y_norm
+    é é
+    >>> print repr(x_norm), repr(y_norm)
+    u'\xe9' u'\xe9'
+    >>> x_norm == y_norm
+    True
+
+.. _unicodedata.normalize: https://docs.python.org/2/library/unicodedata.html#unicodedata.normalize
+
--- a/ckan/__init__.py
+++ b/ckan/__init__.py
@@ -1,3 +1,5 @@
+# encoding: utf-8
+
 __version__ = '2.6.0a'
 
 __description__ = 'CKAN Software'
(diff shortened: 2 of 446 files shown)


torfsen:
@wardi I think now would be a good time for a first merge (assuming you're OK with the changes): The two commits don't change any functionality, but the latter touches a lot of files with a very basic change each (addition of the encoding specification). Keeping it around unmerged would make reviews of future changes for #3006 much harder.

wardi:
These changes are unlikely to cause conflicts, so I'm happy to see them go in soon. Would you also add the check to our pep8/coding standards test that the encoding is declared as utf8 on all source file?

torfsen:
@wardi Great! Although it's hard to see among all the changes there's a dedicated test [`test_source_files_specify_encoding`](https://github.com/torfsen/ckan/blob/3006-unicode/ckan/tests/test_coding_standards.py#L128) which checks that the encoding is specified to UTF-8 in all Python source files. Is that what you had in mind?

I've rebased onto master to ease merging.

doc/contributing/unicode.rst
```diff
@@ -0,0 +1,128 @@
...
+String literals are string values given directly in the source code (as opposed
+to strings variables read from a file, received via argument, etc.). In
+Python 2, string literals by default have type ``str``. This has the risk of
+accidentially introducing ASCII strings and makes a future support of Python 3
+difficult (since string literals are interpreted as Unicode in Python 3).
+
+Therefore, the ``unicode_literals`` feature is used to make Python 2 treat
+string literals as Unicode::
+
+    from __future__ import unicode_literals
```
wardi: I don't think we agreed to this, but this might be option that involves fewer changes scattered all over the source. The cost is that moving code between modules that have differently declared unicode_literals may cause hidden breakage.

@amercader, @brew, @TkTech, @davidread, @rossjones your thoughts?

torfsen: There are many arguments for and against using `unicode_literals`, and the discussion in #3006 didn't result in a clear decision. This part is therefore my suggestion of how to deal with the issue, but of course I'm open to other opinions. Basically the two options are

1. Import `unicode_literals` and mark ASCII literals using the `b` suffix. This does the right thing most of the time. On the other hand, moving code between modules might break things, and you might end up with Unicode strings where you don't want to have them (e.g. filenames).
2. Don't import `unicode_literals` and use the `u` prefix for Unicode literals. Code can then be moved easily (because the meaning of a string literal only depends on its local definition) and you don't get Unicode strings by accident. However, you might get ASCII strings by accident (which I think is more problematic if you're trying to Unicode wherever it's possible). Since most string literals in the future should probably be Unicode, string literals without a prefix would be suspicious (intended ASCII literal or forgotten `u` prefix?), so they would either need a comment or (preferrably) a `b` prefix. This means that every single string literal in the source (including docstrings!) needs a `u` or a `b` prefix.

As I wrote in #3006, it's about choosing the lesser of two evils.

deinok:
@torfsen I have take a look at [PEP-263](https://www.python.org/dev/peps/pep-0263/).
Seems like correct way is -->    # coding=<encoding name>
and you are doing this -->    # encoding: utf-8

Correct me if im wrong. Anyway, i think its a good idea follow PEP 263

TkTech:
@deinok both are perfectly valid and are styles originally used by different editors. Although I prefer:

```
# -*- coding: utf-8 -*-
```

... it's completely personal preference. Python doesn't really care what you write there, it's pretty flexible. As long as it matches:

```
"^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)"
```

amercader:
> 1. Import unicode_literals and mark ASCII literals using the b suffix. This does the right thing most of the time. On the other hand, moving code between modules might break things, and you might end up with Unicode strings where you don't want to have them (e.g. filenames).
> 2. Don't import unicode_literals and use the u prefix for Unicode literals. Code can then be moved easily (because the meaning of a string literal only depends on its local definition) and you don't get Unicode strings by accident. However, you might get ASCII strings by accident (which I think is more problematic since we're trying to use Unicode wherever it's possible). Since most string literals in the future should probably be Unicode, string literals without a prefix would be suspicious (intended ASCII literal or forgotten u prefix?), so they would either need a comment or (preferrably) a b prefix. This means that every single string literal in the source (including docstrings!) needs a u or a b prefix.

Both seem to have drawbacks and I'm happy to go with whatever option prefer but I'm perhaps +0 on 1 just because it seems to imply less changes all over the place and (if I understood correctly) aligns better with the Python 3 behaviour.

I've always used the emacs based `# -*- coding: <encoding name> -*-` but just because I didn't realize that `# coding: <encoding name>` was also supported :)

@torfsen I found your Unicode guidance documentation page great is that no longer included on this PR?

wardi:
We have a third option: Leave the literals the way they are but fix the few broken places in the code where strings enter and leave CKAN. When we move to Python 3 we can address the cases where using unicode internally breaks something (I doubt there will be many)

I'm leery of unicode_literals. It means I can't move code from one module (e.g. an external extension) to another and expect it to behave the same.

TkTech:
I might be missing something here, why would we not just be explicit and use `u` when needed? It works in py2 and is a NOP in py3.3+. In other places, `from six import text_type, binary_type`.

wardi:
@TkTech yes, I'm in favour of adding `u` where needed, not doing a giant change that adds a `u` to every literal string in the source.

torfsen:
@amercader: The overview page is still there, but GitHub doesn't show all the changed files anymore because there are so many. See [the first commit of the PR](https://github.com/ckan/ckan/pull/3049/commits/a1ab007efdad3b549dff1858fa87b18f289ca422).

@wardi, @TkTech: I don't think it makes sense to just "add `u` where it's needed". The point is that you need to know, for every string variable at any time, whether it contains `str` or `unicode`. Otherwise you end up either doing a lot of type checking or you'll run into problems where you encode/decode data twice. If you don't want to have `u` prefixes all over the place but want to use Unicode internally then using `unicode_literals` is the only sane choice.

The problem of moving code only exists in the transition phase where only some modules import `unicode_literals`. Does moving code happens that often, anyway? And if it does, why not use these opportunities to fix the code w.r.t Unicode handling? We're doing something similar with the legacy tests -- they're still there, and the policy is to de-legacy them whenever you work on code related to them.

TkTech:
@torfsen I think you might have misunderstood me. If it's needed almost everywhere, it's needed almost everywhere. `unicode_literals` is a terrible kludge and I am definite :-1: on using it. It only serves a purpose when targeting py3.0-3.2.2. It's an easy way to get out of accrued technical debt but it is not the correct one.

We do not need a lot of type checking. There are very few places in CKAN where binary literal strings make sense and those paths should never be mixing `unicode` and `str`.

wardi:
@torfsen unicode_literals and prefixing every literal in the source have costs that are too high. Let me be more concrete. Here are a couple of pretend examples that we would never need to change, and are the sort of thing we do all over the place:

```python
dataset_name = json.loads(my_json)['name']
```
`'name'` never needs to be changed to `u'name'` because it matches `u'name'` in Python2 and is already correct in Python3

```python
data_dict = json.loads(my_json)
return '%s already exists' % data_dict['title']
```

The string formatting never needs to be changed because it will always be promoted to a unicode string in Python 2 and is already correct in Python 3.

I agree with @TkTech unicode_literals is a hack and will cause hidden breakage in code that generally works pretty well. Changing every string literal will touch the source code all over the place, breaking our ability to backport fixes and not improve anything.

torfsen:
@wardi: Okay, I think I'm beginning to understand what you're aiming at. So let's see what string literals are actually used for:

**dict keys**: `{'foo': x}` and `my_dict['my_key']`
Lookup is not a problem as long as the key is ASCII-only:
```
>>> {'o': 1}[u'o']
1
>>> {u'o': 1}['o']
1
```
If the key contains non-ASCII characters then it depends on whether the key inside the dict is Unicode or not:
```
>>> {'ö': 1}['ö']
1
>>> {'ö': 1}[u'ö']
  File "<stdin>", line 1, in <module>
KeyError: u'\xf6'
>>> {u'ö': 1}['ö']
  File "<stdin>", line 1, in <module>
KeyError: '\xc3\xb6'
>>> {u'ö': 1}[u'ö']
1
```
We're probably OK here, since the keys which we refer to using literals are ASCII-only.

**String formatting**: `'hello %s' % name` and `'hello {}'.format(name)`
Percent-formatting (`a % b`) promotes the result to Unicode if `a` is `str` but `b` is Unicode:
```
>>> 'hello %s' % 'world'
'hello world'
>>> 'hello %s' % u'world'
u'hello world'
```
That is, we're save if we are sure that `b` is always `unicode`.

`a.format(b)`, however, doesn't promote to Unicode. You either get a `str` instance (if `b` is `unicode` but contains only ASCII data) or a `UnicodeEncodeError`:
```
>>> 'hello {}'.format('world')
'hello world'
>>> 'hello {}'.format(u'world')
'hello world'
>>> 'hello {}'.format(u'wörld')
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 1: ordinal not in range(128)
```
[We tell people to use `.format` instead of `%`](http://docs.ckan.org/en/latest/contributing/python.html#string-formatting) so we need to make sure that we use Unicode literals for `a`.

**Translation lookup**: `_('hello')`
`_(a)` returns `unicode`, even if `a` is `str`. If `a` is a `str` with non-ASCII characters then you'll get a `UnicodeDecodeError`, but since we're talking about literals here this would only be a problem if we had literals with non-ASCII data.

**Docstrings**
Docstrings are normal string literals, so `x.__doc__` is `str` by default. We're actually using `__doc__` for [several things](https://github.com/ckan/ckan/search?utf8=%E2%9C%93&q=__doc__), for example for generating API and CLI help messages. Hence docstrings should be `unicode`.

**Function parameter defaults**: `def foobar(x='')`
Since these default values often end up as part of the result value they need to be `unicode`.

**General function arguments**: `f('hello')`
Most of the time, these need to be `unicode`, since the target function expects it like that. There will be some exceptions at the boundary to other libraries or to I/O.

There are probably more uses of string literals, but I think this list covers the most important ones.

What I get from this is: In many places, ASCII literals are fine for ASCII-only strings. However, we still need Unicode literals in many places (`.format`, docstrings, function arguments, function defaults). So without `unicode_literals` you would need a detailed list of where non-prefixed literals are OK, and ideally some code that verifies that these rules are followed. Personally, I'd rather use `unicode_literals`, fix the errors that introduces once and then care only about the few places where you actually need ASCII strings.

wardi:
We'll discuss this again at the dev meeting tomorrow.

Would you make this PR only cover the encoding declaration and test for them? The documentation change and other changes are important but we can't see them on github because too many files are being modified. I'd like to merge the encoding declaration changes first.

For similar reasons please submit upcoming PRs a group of files at a time (e.g. a directory and its subdirectories at a time). That lets us use github's code review tools.

torfsen:
@wardi Sure. I wasn't aware of the limitations of GitHub's review system before I created PR, I'll keep it in mind for the future.

Instead of this PR, there are now
* #3068, which contains only the documentation.
* #3069, which contains only the encoding specifications and the corresponding test case.

PR #3051, opened 2016-05-25
[#2856] Create DataPusher task timestamps using UTC

amercader:
When an upload to the DataPusher finishes we check `resource['last_modified']` (in UTC) against `task['last_updated']` to check whether it changed during the previous upload and resubmit it.
As `task['last_updated']` was created using `datetime.now()` this could lead to infinite upload loops.

Discussion is here:

https://github.com/ckan/ckan/issues/2856#issuecomment-220383735


amercader:
thanks @davidread, that was fast!

Issue #2767, opened 2015-12-03
Improve warning messages when user tries to create a dataset but there're no organizations


vitorbaptista:
When installing CKAN, there're no organizations created by default. In that case, when I go and try to add a new dataset, it'll complaing with a somewhat cryptic `Owner org: A organization must be supplied` (see screenshot):

![ckan-without-orgs](https://cloud.githubusercontent.com/assets/76945/11559386/3e0a07b6-99b0-11e5-9a65-acfe22746910.png)

Notice there isn't even a field for setting the organization (because there're no organizations at all). If the user isn't assigned for an organization and CKAN is configured to require one, it would be good to let the user know that as soon as he goes to the dataset creation page.

wardi:
consensus at the meeting is we need somethg like a flash message to appear above this form when no orgs are available but the configuration requires them. Even better if it includes a link to the new org page since this page will only ever be visible for sysadmins

k-nut:
This is duplicated by #3046 an will hopefully be fixed by #3054

wardi:
fixed by #3054

PR #3054, opened 2016-05-25
Display warning when creating dataset without organizations

### LINKED ISSUE #3046: Creation of datasets/harvest sources with no organization specified
2.5.2

Alternative 1: When adding a new dataset/harvest source and no organization has been created so far, the user should be promoted with a note that this has to be done first. Maybe the form could be grayed out.

Alternative 2: When adding a new dataset/harvest source and no organization has been created so far, the form should still include the organization-field. When the user puts focus on it, a dialogue opens assisting the user in creating a new organization.

When adding a new dataset/harvest source and no organization has been created so far, the form field for organization is not displayed. The form can be filled out. After saving an error message reads
> The form contains invalid entries:
> Owner org: A organization must be supplied

This is confusing since no such field is provided.

1. Create a new ckan instance or delete all organizations. 2. Create a new dataset/set up a new harvest source.

k-nut:
Fixes #3046, #2767

### Proposed fixes:
As discussed in #3046 already this now checks if there are organizations to which the dataset can be added or not. If there are non it checks if the user is allowed to create on or not and depending on the answer provides them with a link.

The tests are failing right now unfortunately. When either one of them is commented out they both pass but the first one fails when run together. Any idea what this could be? Also I feel bad for having to introduce `patch` in those tests again, maybe there is a better way to do it?

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -29,7 +29,7 @@ def owner_org_validator(key, data, errors, context):
 
     if value is missing or value is None:
         if not authz.check_config_permission('create_unowned_dataset'):
-            raise Invalid(_('A organization must be supplied'))
+            raise Invalid(_('An organization must be provided'))
         data.pop(key, None)
         raise df.StopOnError
 
@@ -38,7 +38,7 @@ def owner_org_validator(key, data, errors, context):
     user = model.User.get(user)
     if value == '':
         if not authz.check_config_permission('create_unowned_dataset'):
-            raise Invalid(_('A organization must be supplied'))
+            raise Invalid(_('An organization must be provided'))
         return
 
     group = model.Group.get(value)
--- a/ckan/tests/controllers/test_package.py
+++ b/ckan/tests/controllers/test_package.py
@@ -8,7 +8,7 @@ from nose.tools import (
     assert_true,
 )
 
-from mock import patch
+from mock import patch, MagicMock
 from routes import url_for
 
 import ckan.model as model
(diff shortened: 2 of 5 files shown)


ckan/logic/validators.py
```diff
@@ -27,7 +27,7 @@ def owner_org_validator(key, data, errors, context):
 
     if value is missing or value is None:
         if not authz.check_config_permission('create_unowned_dataset'):
-            raise Invalid(_('A organization must be supplied'))
+            raise Invalid(_('An organization must be provided'))
```
Psykar: This will break a bunch of existing translations. Just flagging it :)

ckan/tests/controllers/test_package.py
```diff
@@ -38,6 +39,43 @@ def test_form_renders(self):
...
+    @patch('ckan.logic.auth.create.organization_create',
+           return_value={'success': False})
+    @helpers.change_config('ckan.auth.create_unowned_dataset', 'false')
+    def test_needs_organization_but_no_organizations_no_button(self, *args):
+        ''' Scenario: The settings say every dataset needs an organization
+        but there are no organizations. If the user is not allowed to create an
+        organization they should be told to ask the admin but no link should be
+        presented'''
+        app = self._get_test_app()
+        sysadmin = factories.Sysadmin()
```
Psykar: Change this to factories.User() and you shouldn't need the patch?

ckan/logic/validators.py
k-nut: That is something that I though about to. But there is no way to avoid it, is there?

ckan/tests/controllers/test_package.py
k-nut: The problem with that is that the default `User` who is created does not have the `package_create` privilege so gets a 403 page. Is there a way to create a user who is allowed to do that? I would love to get rid of the `patch`

Psykar: Take a look at `package_create` in ckan.logic.auth.create

You'll need to set a config var for the test, probably `anon_create_dataset` and `create_dataset_if_not_in_organization` You might also need `organization_create` which needs `user_create_organizations`

k-nut: So we just came to the conclusion on irc that we are testing a scenario here that is not possible with the default ckan `package_create` function. That is because it checks if a) we are allowed to create unowned datsets or b) we are in an organization for which we can create datasets. And what we want to test is a ckan without any organizations where the settings are so that every dataset needs to have an organization. 

@Psykar proposed writing an extension for this test instead of patching this because we would be patching ckan into a situation that would not happen in non-extenstion mode. @amercader what do you think?

amercader: You definitely can't get into this situation without a custom `package_create` auth logic. I think it's fine to patch `package_create` to mimic the overriding from an extension as that would be essentially what it will cause, just add a comment explaining why you are doing it.

So mock `package_create` to return True + `@helpers.change_config('ckan.auth.create_unowned_dataset', 'false')` + `@helpers.change_config('ckan.auth.user_create_organizations', 'false')`

k-nut: When I change the test to this configuration it works. But the original problem remains. When only this test is run it works but if the previous test (`test_needs_organization_but_no_organizations_has_button`)is run before the test fails because a 403 status is returned...

k-nut: Ok, found the issue. Apparently `authz` caches the auth functions. So if we want to change `package_create` we need to clear the cache to make sure that this new code is loaded...

wardi: Another case of change_config not actually reloading the config?

k-nut:
@amercader The problem with the tests that make each other fail is now fixed.

 I documented the reasoning for this rather uncommon patching / cache invalidation in the code so that future readers understand what is going on and so that we have a "complex" tests that we could potentially copy from in the future if similar things come up.

k-nut:
I'll make this into a decorator.

k-nut:
@wardi see 71ee7c1 where I created a new `helpers.mock_auth` function similar to the existing `mock_action` method. This works as a decorator and replaces the setup and teardown at begging and end of the test itself.

amercader:
Great stuff @k-nut, love `mock_auth` :+1:

PR #3053, opened 2016-05-25
Remove references to old session data

brew:
Locale is persisted by i18n middleware in the url, and no longer in the session. This PR just removes a couple of stray references to the old session code.


TkTech:
:+1:

PR #3058, opened 2016-05-25
Cleanup schema.py

### LINKED ISSUE #3057: ckan.logic.schema file is messy
There are a couple of pep8/style issues in [ckan/logic/schema.py](https://github.com/ckan/ckan/blob/master/ckan/logic/schema.py)
They are mainly
- [x] unsued imports
- [x] not 2 lines between functions
- [x] dictionaries with duplicate keys

I will clean those in three separate commits and then tick them off here.

The last checkbox probably needs some explanation: [here](
https://github.com/ckan/ckan/blob/master/ckan/logic/schema.py#L210) we call `update` on a dictionary and pass it a new dictionary. The new one contains the keys `cache_last_updated` and `last_modified` twice so that the second empty declarations overwrite the first one. I assume that the values that actually contain something are the ones that are wanted and will remove the empty ones in favor of them.

k-nut:
Fixes #3057

### Proposed fixes:
This should not change anything for the user. It only makes the code a little more readable.

### Features:
- [ ] includes API changes


k-nut:
So the reason the test is failing is because of commit cece50b. Should I revert that and use the empty values in the dictionary or do you think the test needs to be changed?

k-nut:
@wardi the test are passing again now. https://circleci.com/gh/k-nut/ckan/12

wardi:
Thanks. Since this is clean-up can you remove https://github.com/ckan/ckan/blob/master/ckan/tests/legacy/test_coding_standards.py#L417 and fix whatever pep8 issues are present too?

k-nut:
@wardi done! Also interesting to see that github is now reporting the results from my circle.ci results again :)

PR #3059, opened 2016-05-25
2604 file upload ux

### LINKED ISSUE #2604: File Upload UX issue
Not related to #2590. We are upgrading a client from 2.2 to 2.3 and they raised an issue about how the File Upload/Link has changed in 2.3:
screen shot 2015-08-28 at 4 29 55 pm
screen shot 2015-08-28 at 4 29 01 pm

The reasoning is that the 2.2 interface is more user friendly, especially for non-technical users. The 2.3 interface turns the file label into 'URL', and then also sets that URL to the file name. What could confuse editors is that the file didn't turn into a URL, and even once it does the filename that the input field gets populated with is not the URL that the file will be available on. It also replaces the large and unambiguous Remove button with an 'X' at the end of the field.

In CKAN 2.2, this interface is more intuitive.

I know these aren't the kind of issues CKAN usually deals with, but I thought it'd be worth raising. I'd love to get a couple of opinions on this issue, and depending on the discussion I'm happy to submit a patch to revert the interface back to 2.2.

jrods:
Fixes #2604 

### Proposed fixes:

I've changed the label from 'File' to 'Data' when there's no file/link.

When uploading a file from a local machine to a resource, the label will
change to 'File'. When adding a link to a resource, the label will
change to 'URL'.

When editing a resource, if the data is an upload, the data field will
only display the filename, rather than the url. The label for an upload
will also display as 'File'.

The remove data icon now uses text 'Remove' instead of an icon.

screen shot 2016-05-25 at 4 14 44 pm
screen shot 2016-05-25 at 4 14 34 pm
screen shot 2016-05-25 at 4 13 29 pm

---
In a separate commit, I've included the updated main.css file

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/data-upload.js
+++ b/ckan/public/base/javascript/modules/data-upload.js
@@ -1,0 +1,202 @@
+/**
+ * Data-upload module
+ *
+ * This module provides the user experience for adding data to a resource
+ *
+ * Based on image-upload.js
+ *
+ * Author: Jared Smith @ Highway Three Solutions
+ */
+this.ckan.module('data-upload', function($, _) {
+  return {
+    is_url: false,
+    is_upload: false,
+    field_url: '',
+    field_upload: '',
+    field_clear: '',
+    options: {
+      i18n: {
+        upload: _('Upload'),
+        url: _('Link'),
+        remove: _('Remove'),
+        label: _('Data'),
+        label_for_url: _('URL'),
+        label_for_upload: _('File'),
+        upload_tooltip: _('Upload a file on your computer'),
+        url_tooltip: _('Link to a URL on the internet (you can also link to an API)')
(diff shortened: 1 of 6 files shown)


ckan/public/base/javascript/modules/data-upload.js
```diff
@@ -0,0 +1,202 @@
...
+        this._showOnlyButtons();
+      }
+    },
+
+    /* Quick way of getting just the filename in the uri of the resource data
+     *
+     * Returns String.
+     */
+    _fileNameFromUpload: function(url) {
+      url = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
```
k-nut: you probably want to use `===` here?!

PR #3061, opened 2016-05-26
Do not create default views on update

### LINKED ISSUE #3011: Default view is re-added when removed before DataStore upload is complete
I'm doing the following via the API on a CKAN 2.5.2 installation that has DataStore + DataPusher enabled:

1. Create a CSV resource
2. Remove the default `recline_view`
3. Add a new `recline_grid_view` view

I'd expect the resource to end up with a single view (the one I added in step 3). However, it ends up with *two* views: the one I created and a *new* default view (ID is different from the original default view).

It seems that the default view is added again once the data has been uploaded to the DataStore by the DataPusher. If steps 1-3 are done quickly enough (before the upload to the DataStore has completed) then the resource has only the intended `recline_grid_view`, which at that point shows an error (because the data is not yet available). As soon as the data is available (i.e. as soon as the error is gone after a reload) the resource has gotten an additional `recline_view`.

Here's a small script to demonstrate the problem:
```python
ckan = ckanapi.RemoteCKAN(CKAN_URL, apikey=API_KEY)

res_dict = ckan.action.resource_create(package_id=PKG_ID, url=CSV_URL,
                                       name='Test', format='CSV')
view_dicts = ckan.action.resource_view_list(id=res_dict['id'])
for view_dict in view_dicts:
    ckan.action.resource_view_delete(id=view_dict['id'])
ckan.action.resource_view_create(resource_id=res_dict['id'], title='My View',
                                 view_type='recline_grid_view')

# Uncomment to make assertion fail
#import time
#time.sleep(10)

assert len(ckan.action.resource_view_list(id=res_dict['id'])) == 1
```
If you uncomment the `sleep` lines (to wait until the data is in the DataStore) then the assertion fails.

k-nut:
Fixes #3011

### Proposed fixes:
With this change the views are created on `resource_create` but not on `package_update` The tests that made sure that views were created on update were removed. The `test_datepreview.py` file contains a `test_default_views_created_on_resource_create` test that assures that the default views are created when a new resource is created.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -317,14 +317,14 @@ def resource_create(context, data_dict):
     resource = updated_pkg_dict['resources'][-1]
 
     ##  Add the default views to the new resource
-    create_datastore_views = paste.deploy.converters.asbool(
-        data_dict.get('create_datastore_views', False))
-    ckan.lib.datapreview.add_views_to_resource(context,
-                                               resource,
-                                               None,
-                                               view_types=[],
-                                               create_datastore_views=
-                                               create_datastore_views)
+    logic.get_action('resource_create_default_resource_views')(
+        {'model': context['model'],
+         'user': context['user'],
+         'ignore_auth': True
+         },
+        {'resource': resource,
+         'package': updated_pkg_dict
+         })
 
     for plugin in plugins.PluginImplementations(plugins.IResourceController):
         plugin.after_create(context, resource)
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -316,6 +316,16 @@ def resource_create(context, data_dict):
     updated_pkg_dict = _get_action('package_show')(context, {'id': package_id})
     resource = updated_pkg_dict['resources'][-1]
 
+    ##  Add the default views to the new resource
+    create_datastore_views = paste.deploy.converters.asbool(
+        data_dict.get('create_datastore_views', False))
+    ckan.lib.datapreview.add_views_to_resource(context,
+                                               resource,
+                                               None,
+                                               view_types=[],
+                                               create_datastore_views=
+                                               create_datastore_views)
+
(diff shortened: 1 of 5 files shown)


amercader:
@k-nut The view was created before this change because `datastore_create` was calling `package_update` at some point where `datastore_active` was True.

With this change, when `datastore_create` calls `resource_create` , `datastore_active` is still False. We can get the view though by fixing this check in the `can_view` method:

```diff
diff --git a/ckanext/reclineview/plugin.py b/ckanext/reclineview/plugin.py
index 2a22a88..26610c0 100644
--- a/ckanext/reclineview/plugin.py
+++ b/ckanext/reclineview/plugin.py
@@ -90,7 +90,7 @@ class ReclineView(ReclineViewBase):
         resource = data_dict['resource']
 
         if (resource.get('datastore_active') or
-                resource.get('url') == '_datastore_only_resource'):
+                '_datastore_only_resource' in resource.get('url', '')):
             return True
         resource_format = resource.get('format', None)
         if resource_format:
```

This will make the test pass without having to provide a format.

k-nut:
@amercader Changed that and it works fine without `format`. Thank you for the hint! 

I'm wondering if we should also apply the same change to the `can_view` function in `ReclineViewBase` a couple of lines above?

amercader:
@k-nut yes, we should. Good catch.

k-nut:
@amercader just pushed a change that does that.

amercader:
@torfsen can you confirm if this fixes your issues in #3011?

torfsen:
@amercader I currently don't have a test setup for 2.6, but it seems that @k-nut [has verified that his changes make my example script work](https://github.com/ckan/ckan/issues/3011#issuecomment-220948594). That's fine for me. If it's important for you I can test this separately but it would have to wait until I find time for it.

amercader:
@torfsen This should be good to backport as is to 2.5 (and 2.4?) but that's fine, I'm happy to merge it as is

PR #3064, opened 2016-05-27
Fix some auth functions not using IAuthFunctions

Psykar:
### Proposed fixes:

Various auth functions are purely wrappers around other auth functions.
Unfortunately this means when overriding one of auth functions (a parent) that is used by others (children),
the children will still call the original version, not the overridden version.

This commit fixes this for most cases.

Auth functions that are purely a wrapper around a parent auth
function are simplified by using a wrapper function.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/auth/create.py
+++ b/ckan/logic/auth/create.py
@@ -78,11 +78,11 @@ def resource_create(context, data_dict):
 
 
 def resource_view_create(context, data_dict):
-    return resource_create(context, {'id': data_dict['resource_id']})
+    return authz.is_authorized('resource_create', context, {'id': data_dict['resource_id']})
 
 
 def resource_create_default_resource_views(context, data_dict):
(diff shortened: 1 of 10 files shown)


amercader:
@Psykar This is great, but I think that the code would be much easier to understand to newcomers if we keep the same function signatures in all cases (ie not use your wrapper).

So instead of this:

```python
package_delete = logic.is_authorized_wrapper('package_update')
```

keep doing this:

```python
def package_delete(context, data_dict):
    return authz.is_authorized('package_update', context, data_dict)
```

Psykar:
Fair - I got the idea from the patch actions, but have swapped them as well to using the expanded style (simplifies the unit test then too)

amercader:
Thanks @Psykar, great stuff! :100:

PR #3066, opened 2016-05-27
Pad object names from factories with zeros

k-nut:
Fixes #

### Proposed fixes:
The test in ckan/tests/controllers/test_group.py would sometimes fail when run in an unfortunate order (note: never on circle.ci as that splits up the test in way where this does not happen). The `test_page_thru_list_of_orgs_preserves_sort_order` and `test_page_thru_list_of_groups_preserves_sort_order` would fail because some organizations had already been created by other tests and the numbering would start a 5 instead of 1. And because the sorting was done via strings (so `Group 5` came after `Group 31` the test would fail. By always padding them with zeros (so it is `Group 05` now). this should not happen anymore and make the tests more stable.

This does not include any new tests and only changes code in the test folder. Since the test are still passing this should be save to merge.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/factories.py
+++ b/ckan/tests/factories.py
@@ -109,7 +109,7 @@ class User(factory.Factory):
     about = 'Just another test user.'
 
     # Generate a different user name param for each user that gets created.
-    name = factory.Sequence(lambda n: 'test_user_{n}'.format(n=n))
+    name = factory.Sequence(lambda n: 'test_user_{0:02d}'.format(n))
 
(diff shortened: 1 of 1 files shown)


k-nut:
@amercader as discussed in the call yesterday all things that come out of factories are padded with zeros now. Fortunately this did not break any tests. 

Do you think this is good to go?

wardi:
Now we're all set until test number 101. ;-)

PR #3068, opened 2016-05-30
Document CKAN Unicode handling.

torfsen:
This is work for #3006. It was originally submitted via #3049 but has been put into this separate PR to ease reviewing.

The PR updates the documentation with details on how Unicode is handled in CKAN.

### DIFF
--- a/doc/contributing/unicode.rst
+++ b/doc/contributing/unicode.rst
@@ -1,0 +1,160 @@
+================
+Unicode handling
+================
+This document explains how Unicode and related issues are handled in CKAN.
+For a general introduction to Unicode and Unicode handling in Python 2 please
+read the `Python 2 Unicode HOWTO`_. Since Unicode handling differs greatly
+between Python 2 and Python 3 you might also be interested in the
+`Python 3 Unicode HOWTO`_.
+
+.. _Python 2 Unicode HOWTO: https://docs.python.org/2/howto/unicode.html
+.. _Python 3 Unicode HOWTO: https://docs.python.org/3/howto/unicode.html
+
+.. note::
+
+    This document describes the intended future state of Unicode handling in
+    CKAN. For historic reasons, some existing code does not yet follow the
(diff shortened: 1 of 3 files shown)


doc/contributing/unicode.rst
```diff
@@ -0,0 +1,128 @@
...
+String literals
+---------------
+String literals are string values given directly in the source code (as opposed
+to strings variables read from a file, received via argument, etc.). In
+Python 2, string literals by default have type ``str``. This has the risk of
+accidentially introducing ASCII strings and makes a future support of Python 3
+difficult (since string literals are interpreted as Unicode in Python 3).
+
+Therefore, the ``unicode_literals`` feature is used to make Python 2 treat
+string literals as Unicode::
```
wardi: We discussed this again just now. @amercader, @TkTech and I are against using `unicode_literals` and we would prefer to change *all* our literal strings to explicit unicode.

We know that's a big change, and it would need to be done in a number of smaller PRs. Are you willing to help with that?

torfsen: OK, good to see that a decision was made! I'm absolutely willing to help with that, but as you've already said, the work needs to be done in multiple smaller PRs anyway, hence other people can work on it, too.

I will update the docs in this PR so that it can be merged and used as a reference for further work. Before I do that, however, I want to be sure that we're on the same page:

* `unicode_literals` is not used.
* *Every* string literal (including docstrings) gets a type prefix. For most literals that will be `u`, but ASCII-string literals are also explicitly marked using `b` (that is redundant in Python 2 but makes the intention clear and helps with migration to Python 3).

wardi:
There are a few strings that won't have a prefix, such as format strings used in `__repr__` methods IIRC. but yes, I was outvoted :-) all the literal strings that should always be unicode or binary will get a prefix.

torfsen:
@wardi I've updated the PR and it should be fine to merge it now.

PR #3069, opened 2016-05-30
Add and test PEP 263 source file encoding specifications.

torfsen:
This is work for #3006. It was originally submitted via #3049 but has been put into this separate PR to ease reviewing.

This PR adds PEP 263 encoding specifications to all Python source files. It also adds a test which checks that every Python source file does have an encoding specification.

### DIFF
--- a/ckan/__init__.py
+++ b/ckan/__init__.py
@@ -1,3 +1,5 @@
+# encoding: utf-8
+
 __version__ = '2.6.0a'
 
 __description__ = 'CKAN Software'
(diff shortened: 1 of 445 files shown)


amercader:
When #3068 is merged can you add a reference to the unicode docs on this page? http://docs.ckan.org/en/latest/contributing/python.html

torfsen:
@amercader That's already part of #3068: [python.rst](https://raw.githubusercontent.com/torfsen/ckan/3006-document-unicode-handling/doc/contributing/python.rst).

wardi:
@torfsen thanks! you're now the latest author of every file in ckan :-)

Issue #2649, opened 2015-09-18
resource extras stripped by package controller


wardi:
https://github.com/ckan/ckan/blob/5c6ab31/ckan/controllers/package.py#L620

The resource_edit method currently filters fields passed to the resource form for draft datasets, breaking custom schemas with resource extras defined.

torfsen:
Just ran into this. In addition to the problem pointed out by @wardi, there's also the issue that the POST data returned by the form does not include the extras. Hence, when `resource_update` is called with the data, the extras are lost.

As a simple fix it would probably be enough to use `resource_path` instead of `resource_update` [when storing the POST data](https://github.com/ckan/ckan/blob/9de1f467d852d2fac75217cfd36b400d779e718e/ckan/controllers/package.py#L567). That doesn't fix the original problem mentioned by @wardi but it does fix mine (at least according to my quick experiments).

torfsen:
PR #3288 contains the fix I suggested above (using `resource_patch` instead of `resource_update`).

wardi:
fixed by #3480

torfsen:
Reminder for myself: Need to check whether the fix in #3480 also solves the problem I described above regarding the form's POST data not containing the extras.

PR #3072, opened 2016-05-30
User extras

TkTech:
https://github.com/ckan/ideas-and-roadmap/issues/16

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -898,7 +898,7 @@ def user_create(context, data_dict):
 
     You must be authorized to create users.
 
-    :param name: the name of the new user, a string between 2 and 100
+    :param name: the name of the new user, a string between 3 and 255
         characters in length, containing only lowercase alphanumeric
         characters, ``-`` and ``_``
     :type name: string
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -915,6 +915,8 @@ def user_create(context, data_dict):
     :type about: string
     :param openid: (optional)
     :type openid: string
+    :param extras: A dict of extra fields attached to this user. (optional)
+    :type extras: dict
 
     :returns: the newly created user
     :rtype: dictionary
--- a/ckan/logic/schema.py
+++ b/ckan/logic/schema.py
@@ -411,6 +411,7 @@ def default_user_schema():
         'reset_key': [ignore],
         'activity_streams_email_notifications': [ignore_missing],
         'state': [ignore_missing],
+        'extras': [ignore_missing, dict]
     }
     return schema
 
--- a/ckan/model/user.py
+++ b/ckan/model/user.py
@@ -1,6 +1,5 @@
 import datetime
 import re
-import os
 from hashlib import sha1, md5
 
 import passlib.utils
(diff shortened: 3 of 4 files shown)


Psykar:
You need to remove `ckan/model/user.py` from the pep_8 blacklist as you've fixed it's issues.
See `ckan-repo/ckan/tests/legacy/test_coding_standards.py` line 518 or so.

ckan/logic/schema.py
```diff
@@ -411,6 +411,7 @@ def default_user_schema():
         'reset_key': [ignore],
         'activity_streams_email_notifications': [ignore_missing],
         'state': [ignore_missing],
+        'extras': [ignore_missing, dict]
```
wardi: dict won't work as a validator because ValueErrors aren't caught.  Maybe a string_string_dict function  to check that we were passed a dict with unicode keys and values.

TkTech: We accept all json-serializable values not just strings

wardi: Ok, if you'd like to accept any JSON as values (that aligns with how Resource extras works and is more flexible in general, so I'm :+1: ) then something like this instead of `dict`:

```python
def json_object(value):
    "make sure value can be serialized as a json object"
    try:
        if not json.dumps(value).startswith('{'):
            raise Invalid('Not a JSON object')
    except ValueError as e:
        raise Invalid(e... )
    return value
```

TkTech: The JSON is already validated when the POST request is parsed.

It's not a string, it's a literal JSON dict, deserialized into a python dict by the normal `json.loads()` in the API. Checking `startswith('{')` wouldn't be sufficient anyways.

wardi: data arrives at validators from a number of directions. The API controller is one, but it's also passed arbitrary python objects from any ckan code or plugin calling call_action. This is why we have converters that handle dates being passed to some fields.

The code above doesn't assume a string, it just tests that it's something we can serialize as JSON with dumps, and the topmost thing is an object

wardi:
This is great and will enable all sorts of nice features in simple extensions that won't need to create their own tables any more. For the future this can pattern can be repeated to replace package, group, org and resource extras.

This feature creates a genuinely new concept: extras are now namespaced and multiple plugins can share the same JSON object to store their plugin-specific information. So it's bigger than our current extras. If we use the field name 'extras' in the user object that conflicts with the current, more limited free-form extras that exist for groups and organizations, so I'd like to find a different name for this data.

And just to get it out of the way, no, I'm not ok with `new_extras`. :stuck_out_tongue:

wardi:
`plugin_extras`? 
`plugin_data`?

TkTech:
`extra_extras_for_extras`?

`plugin_extras` is probably the best nomination so far :)

ckan/logic/action/create.py
```diff
@@ -915,6 +915,8 @@ def user_create(context, data_dict):
     :type about: string
     :param openid: (optional)
     :type openid: string
+    :param extras: A dict of extra fields attached to this user. (optional)
+    :type extras: dict
```
wardi: `plugin_extras`: private extra user data belonging to plugins. Only sysadmin users may set this value.

wardi:
This needs some tests including:
* plugin_extras can be set with user_create/user_update by sysadmin
* plugin_extras can't be modified or retrieved by non-sysadmin with user_create/user_update/user_show
* plugin_extras are returned only when user_show is passed include_plugin_extras=true and the user requesting is a sysadmin (following the pattern for password_hash)

ckan/model/user.py
```diff
@@ -1,37 +1,53 @@
...
+        types.Boolean,
+        default=False
+    ),
+    Column('sysadmin', types.Boolean, default=False),
+    # Internal user extras. This column can store any type that
+    # can be serialized to JSON. This isn't intended to be used
+    # for end-user key/values, but rather for extensions to store
+    # internal data, like a twitter username which it is then
+    # in charge of displaying.
+    Column('extras', MutableDict.as_mutable(JSON))
```
davidread: Doesn't this need postgres 9.1+? The install instructions still say "v8.4 or newer". So this ticket rather depends on this being done: https://github.com/ckan/ckan/issues/2555

TkTech: 9.2+ is what I'm targeting, same with the new `PostgresIndexer` for search. It's been years, we can't be tied to old versions on `master`.

wardi: #2670 is the PR that needs to get finished. Looks like we included a note in the readme on 2.5 but never included instructions for users upgrading.

torfsen:
It would be really nice to have this kind of extras supported for all CKAN entities in a common way. For example, resource extras currently have issues (#2649) and other entities don't support extras at all (e.g. resource views).

TkTech:
@torfsen That was the idea behind this at least, copy the same mostly-hands-off interface on to everything else to exploit new postgres features. We've never had a concept of user extras so this is the easiest to implement, where with the rest we'll have to offer deprecation and/or migration paths.

TkTech:
Closing in favor of an eventual solution from #4147

PR #3077, opened 2016-06-01
Uncaught email exceptions on user invite

amercader:
When inviting a user to a group or an organization, either via the UI or the `user_invite` API call if there was a problem sending the invite email the exception is not caught, resulting in a 500 error. Common exceptions include `[Errno 111] Connection refused` and `MailerException: SMTPAuthenticationError(435, '4.7.8 Error: authentication failed:')`.

I think we should delete the pending user (which is already created at this point).

Error returned:

5mdnzcw

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -3,6 +3,7 @@
 import logging
 import random
 import re
+from socket import error as socket_error
 
 from pylons import config
 import paste.deploy.converters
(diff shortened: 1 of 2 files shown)


amercader:
Frustratingly the members page will still list the deleted user that was just created even if the email wasn't send. This is because of #3078

k-nut:
@amercader looks good to me 👍

PR #3078, opened 2016-06-01
member_list returns deleted datasets and users

amercader:
Visiting a Group or Org Members page will show deleted users.
Also calling `member_list` on a Group or Org will return deleted datasets (although this is not used on the frontend).

This is because the members table is stateful, ie it has a `state=active/deleted` field which is not updated when deleting a dataset or user.

This PR sets the state of all user/dataset memberships to `deleted` and adds some tests

### DIFF
--- a/ckan/logic/action/delete.py
+++ b/ckan/logic/action/delete.py
@@ -37,10 +37,22 @@ def user_delete(context, data_dict):
     user_id = _get_or_bust(data_dict, 'id')
     user = model.User.get(user_id)
 
+    # New revision, needed by the member table
+    rev = model.repo.new_revision()
+    rev.author = context['user']
+    rev.message = _(u' Delete User: {0}').format(user.name)
+
     if user is None:
         raise NotFound('User "{id}" was not found.'.format(id=user_id))
 
     user.delete()
+
+    user_memberships = model.Session.query(model.Member).filter(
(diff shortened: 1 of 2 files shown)


ckan/logic/action/delete.py
```diff
@@ -37,10 +37,22 @@ def user_delete(context, data_dict):
...
     if user is None:
         raise NotFound('User "{id}" was not found.'.format(id=user_id))
 
     user.delete()
+
+    user_memberships = model.Session.query(model.Member).filter(
+        model.Member.table_id == user_id).all()
+
+    for membership in user_memberships:
+        membership.delete()
```
TkTech: Is this `delete` doing something special in one of the three superclasses? Otherwise we could just use:

```
user_memberships = model.Session.query(
    model.Member
).filter(
        model.Member.table_id == user_id
).delete()
```

amercader: Yes, the member table is versioned with VDM (the custom revision library CKAN uses), so we need to call the domain object `delete` method to create the revision on `member_revision` and set the state automatically (plus any other voodoo that VDM does)

k-nut:
@amercader looks good to me! Do you want me to merge this?

k-nut:
Also how do you backport changes? I would be very much interested to learn how that process works.

PR #3079, opened 2016-06-01
2604 file upload ux

### LINKED ISSUE #2604: File Upload UX issue

jrods:
Fixes #2604 

### Proposed fixes:
Note: These changes only apply to data upload/linking.

I've changed the label from 'File' to 'Data' when there's no file/link.

When uploading a file from a local machine to a resource, the label will
change to 'File'. When adding a link to a resource, the label will
change to 'URL'.

When editing a resource, if the data is an upload, the data field will
only display the filename, rather than the url. The label for an upload
will also display as 'File'.

The remove data icon now uses text 'Remove' instead of an icon.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/image-upload.js
+++ b/ckan/public/base/javascript/modules/image-upload.js
@@ -40,19 +40,19 @@ this.ckan.module('image-upload', function($, _) {
 
       // firstly setup the fields
       var field_upload = 'input[name="' + options.field_upload + '"]';
-      var field_url    = 'input[name="' + options.field_url + '"]';
-      var field_clear  = 'input[name="' + options.field_clear + '"]';
-      var field_name   = 'input[name="' + options.field_name + '"]';
+      var field_url = 'input[name="' + options.field_url + '"]';
+      var field_clear = 'input[name="' + options.field_clear + '"]';
+      var field_name = 'input[name="' + options.field_name + '"]';
 
-      this.input            = $(field_upload, this.el);
-      this.field_url        = $(field_url, this.el).parents('.control-group');
-      this.field_image      = this.input.parents('.control-group');
-      this.field_url_input  = $('input', this.field_url);
-      this.field_name       = this.el.parents('form').find(field_name);
+      this.input = $(field_upload, this.el);
+      this.field_url = $(field_url, this.el).parents('.control-group');
+      this.field_image = this.input.parents('.control-group');
+      this.field_url_input = $('input', this.field_url);
+      this.field_name = this.el.parents('form').find(field_name);
       // this is the location for the upload/link data/image label
-      this.label_location   = $('label[for="field-image-url"]');
+      this.label_location = $('label[for="field-image-url"]');
       // determines if the resource is a data resource
-      this.is_data_resource = (this.options.field_url == 'url') && (this.options.field_upload == 'upload');
+      this.is_data_resource = (this.options.field_url === 'url') && (this.options.field_upload === 'upload');
 
       // Is there a clear checkbox on the form already?
       var checkbox = $(field_clear, this.el);
--- a/ckan/public/base/javascript/modules/image-upload.js
+++ b/ckan/public/base/javascript/modules/image-upload.js
@@ -5,7 +5,7 @@ this.ckan.module('image-upload', function($, _) {
   return {
     /* options object can be extended using data-module-* attributes */
     options: {
-      is_url: true,
+      is_url: false,
       is_upload: false,
       field_upload: 'image_upload',
       field_url: 'image_url',
@@ -17,6 +17,8 @@ this.ckan.module('image-upload', function($, _) {
         url: _('Link'),
         remove: _('Remove'),
         upload_label: _('Image'),
+        label_for_url: _('URL'),
+        label_for_upload: _('File'),
         upload_tooltip: _('Upload a file on your computer'),
         url_tooltip: _('Link to a URL on the internet (you can also link to an API)')
       }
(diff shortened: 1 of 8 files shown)


k-nut:
I think the changes that you made to the css files directly should also be included in the `.less` file as they are going to be overwritten otherwise the next time someone runs `$ ./bin/less`.

More information on this in the [documentation](http://docs.ckan.org/en/latest/contributing/frontend/index.html#stylesheets)

I am not sure if the different styles are also created with less though... (`red.css`, `fuchsia.css` etc.)

ckan/public/base/javascript/modules/image-upload.js
```diff
@@ -38,20 +40,23 @@ this.ckan.module('image-upload', function($, _) {
 
       // firstly setup the fields
       var field_upload = 'input[name="' + options.field_upload + '"]';
-      var field_url = 'input[name="' + options.field_url + '"]';
-      var field_clear = 'input[name="' + options.field_clear + '"]';
-      var field_name = 'input[name="' + options.field_name + '"]';
+      var field_url    = 'input[name="' + options.field_url + '"]';
+      var field_clear  = 'input[name="' + options.field_clear + '"]';
+      var field_name   = 'input[name="' + options.field_name + '"]';
```
k-nut: I am in favor of not adding additional whitespace here as it makes the changes seem much bigger then they actually are. It also seems like aligning equal signs is not done in the other parts of the javascript so I would not change that here.

k-nut:
If you click the `upload` button, then click `remove` and then `upload` again the form does not change to a textfield anymore but the two buttons remain there. I think this should be changed before merging.

jrods:
> If you click the upload button, then click remove and then upload again the form does not change to a textfield anymore but the two buttons remain there. I think this should be changed before merging.

I don't remember this happening, but i will check it out.

> I am in favor of not adding additional whitespace here as it makes the changes seem much bigger then they actually are. It also seems like aligning equal signs is not done in the other parts of the javascript so I would not change that here. 

I will remove the whitespace. This was just my personal preference, and in the javascript style guide, it gave the impression it was fine, regardless I will undo the change.

k-nut:
Hey @jrods just wondering if you had time to look at this again?

jrods:
sorry @k-nut , I have been busy with other work and haven't had the opportunity to work on this.
I recently looked at the issue you were having with the upload textfield and I don't have this problem. I will try a new env and pull in the changes to see if I can get the same issue.

jrods:
@k-nut so i tested in a new env with ckan 2.4.3, and i can't reproduce the issue you were having with click upload, remove, then uploading again. The textfield always appears for me.

k-nut:
OK, I was checking this with the latest master, not with 2.4.3. but I'll take another look at that. Might have been because of some other plugins that I had configured.

ckan/public/base/javascript/modules/image-upload.js
```diff
@@ -38,20 +40,23 @@ this.ckan.module('image-upload', function($, _) {
...
-      this.field_name = this.el.parents('form').find(field_name);
+      this.input            = $(field_upload, this.el);
+      this.field_url        = $(field_url, this.el).parents('.control-group');
+      this.field_image      = this.input.parents('.control-group');
+      this.field_url_input  = $('input', this.field_url);
+      this.field_name       = this.el.parents('form').find(field_name);
+      // this is the location for the upload/link data/image label
+      this.label_location   = $('label[for="field-image-url"]');
+      // determines if the resource is a data resource
+      this.is_data_resource = (this.options.field_url == 'url') && (this.options.field_upload == 'upload');
```
k-nut: Can we use strict equal (`===`) here?

k-nut:
Just checked this again and could not reproduce the issue anymore. So apart from the one occurence of `==` instead of `===` this looks good to me @jrods 👍

ckan/public/base/javascript/modules/image-upload.js
```diff
@@ -105,25 +110,71 @@ this.ckan.module('image-upload', function($, _) {
...
       }
     },
 
+    /* Quick way of getting just the filename from the uri of the resource data
+     *
+     * url - The url of the uploaded data file
+     *
+     * Returns String.
+     */
+    _fileNameFromUpload: function(url) {
```
k-nut: We could replace this function with this regex `/[^\/?#]+(?=$|[?#])/` taken from [stackoverflow](http://stackoverflow.com/a/6543380/4099396). It would make the code shorter but as it is regex also harder to read. Your call @jrods.

jrods: @k-nut i think the function is pretty clear on what it's doing, i personally don't mind the extra lines of code if it's explicit.

k-nut: yep, totally fine with me :)

amercader:
Thanks @jrods, that looks great. Sorry it took a while to get merged.

PR #3080, opened 2016-06-02
Enable chinese locale

### LINKED ISSUE #3073: Chinese (zh_CN and zh_TW) can not be used on a source install
CKAN Version if known: master

Setting up `ckan.locale_default = zh_CN` fails with `AssertionError: default language "zh_CN" not available`.

`zh_CN` and `zh_TW` are the locale codes used on Transifex.

[This check](https://github.com/ckan/ckan/blob/master/ckan/lib/i18n.py#L51:L53) fails. The folder where babel keeps the locale data files (`<virtualenv>/lib/python2.7/site-packages/babel/locale-data`) has different codes for the Chinese locales. Bizarrely this works on the package install (2.5.x), ie the `zh_CN.dat` files are there.

Not sure if it's related to the recent upgrade to Babel (#3004) or something specific with the package install (seems unlikely)

Probably related to https://github.com/python-babel/babel/pull/25

k-nut:
Fixes #3073 

### Proposed fixes:
This Pull Request enables users to start ckan with a default locale of `zh_CH` or `zh_TN`. The first commit in this pull request is responsible for this. It basically uses the higher level `Locale.parse` from `Babel` instead of using the low level `localedata.exists`. This makes it more flexible when it comes to script details such as `zh_Hans_CN` vs `zh_CN` which is not comparable with the lower level `localedata.exists`.

The second commit makes sure that the language picker continues to work by also adding the `short_name` (e.g. `zn_CH`) to the locales that can be selected. The long form (`zn_Hans_CH`) broke the `url_for` here.

I tried to document the reasoning for this in the comments in the code.

### Features:

- [ ] includes API changes


amercader:
Merged! :cn:

TkTech:
👍

PR #3082, opened 2016-06-02
Allow Sysadmins to create users with the form (and test)

k-nut:
Fixes #2863

### Proposed fixes:
Takes the changes from #2863 and adds a test for them.

Additionally the sysadmin is redirected to the newly created user's activity feed now as proposed by @wardi. This is also checked in the test.

### Features:

- [ ] includes API changes


k-nut:
@wardi do you think we can merge this? Right now only quantifiedcode is complaining because of a new undocumented (test) function.

k-nut:
thank you @wardi! 👍

Issue #3086, opened 2016-06-03
ImportError: No module named vdm.sqlalchemy


brylie:
### CKAN Version if known (or site URL)
CKAN 2.5

### Please describe the expected behaviour
When configuring Jetty, with the following command `paster db init -c /etc/ckan/default/development.ini` the command should succeed.

### Please describe the actual behaviour
When configuring Jetty, the `paster db init -c /etc/ckan/default/development.ini` command produces the following error:

```
ImportError: No module named vdm.sqlalchemy
```

### What steps can be taken to reproduce the issue? 
I am following the [Setting up Solf](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html#setting-up-solr) instructions.

brylie:
I ran the following command: `easy_install vdm` and get a new error:

```
ImportError: No module named sqlalchemy
```

k-nut:
Are you still inside the virtual environment and did you `pip install` the `requirements.txt`?

brylie:
I installed SQAlchemy with `sudo pip install SQAlchemy`, and get the following error:

```
ValueError: VDM only works with SQLAlchemy versions 0.4 through 0.8, not: 1.0.13
```

brylie:
Thanks for your reply @k-nut. The installation instructions do not mention to do `pip install` on the `requirements.txt`.

k-nut:
That step should have been done with [step 2](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html#install-ckan-into-a-python-virtual-environment) part `c)` of the documentation. ( `pip install -r /usr/lib/ckan/default/src/ckan/requirements.txt` )

Maybe you missed it? Or do we need to point this out in another way?

brylie:
OK, basically the [install CKAN from package](http://docs.ckan.org/en/latest/maintaining/installing/install-from-package.html) instructions are incomplete, since they rely heavily on steps in the [install CKAN from source](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) instructions.

brylie:
In effect, I cannot install CKAN from package without reading through the installation from source instructions.

k-nut:
Yes, the `install CKAN from package` documentation sometimes redirects you to shared parts that are only documented in the `install CKAN from source` tutorial...

brylie:
OK, where does it direct me to step two part c of the installation from source instructions?

brylie:
[...]

Issue #3085, opened 2016-06-03
FATAL:  password authentication failed for user "ckan_default"


brylie:
### CKAN Version if known (or site URL)
CKAN 2.5

### Please describe the expected behaviour
When running the commmand `sudo ckan db init`,  the database should be successfully initialized.

### Please describe the actual behaviour
When running the command `sudo ckan db init` the following error occurs:

```
FATAL:  password authentication failed for user "ckan_default"
```

### What steps can be taken to reproduce the issue? 
I am following the [CKAN installation instructions](http://docs.ckan.org/en/latest/maintaining/installing/install-from-package.html).

k-nut:
In your `production.ini`/`development.ini` did you put the password into the `sqlalchemy.url` line that you set up when creating the user? As described in [step 4 of the tutorial](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html#create-a-ckan-config-file)

brylie:
Ah, I am reading through the [installing CKAN from package](http://docs.ckan.org/en/latest/maintaining/installing/install-from-package.html) instructions. This is not a simple installation process, it seems.

k-nut:
It is a little confusing that [step 2](http://docs.ckan.org/en/latest/maintaining/installing/install-from-package.html#install-postgresql-and-solr) in the `installing CKAN from package` documentation redirects to setting up postgres as described in the `installing CKAN from source` documentation. But when reading eveything very carefully you should end up with a working installation. 

I followed the tutorial this morning and it took about 45 minutes in total (I do have some more knowledge about how ckan works though...)

brylie:
OK, thanks @k-nut. 

I am trying to read the [Installing CKAN from package](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) instructions carefully, but there seems to be some missing steps in that document. E.g. there does not seem to be any mention of following [step 2](http://docs.ckan.org/en/latest/maintaining/installing/install-from-package.html#install-postgresql-and-solr) of the [installing CKAN from source](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) document when reading through the *installing CKAN from package*. Correct me if I am wrong, but I simply do not see any reference.

amercader:
@brylie See https://github.com/ckan/ckan/issues/3086#issuecomment-223541300 

Closing for now.

tubulabi:
Hiii, Everyone ..
[...]

PR #3087, opened 2016-06-03
Consolidate Postgres and Solr install docs

amercader:
Right now people following the package install docs need to jump to the source install docs for setting up Postgres and Solr and are expected to come back to the package install docs once finished. This causes confusion, as in #3085 and #3086.

This changes moves the Postgres and Solr install instructions to separate file which are then included from the package and source install pages, so you don't need to leave them during the install process.


k-nut:
Seeing how confusing this seems to be I very much like this change!

TkTech:
I think we can start removing any non-multicore solr docs as well, as it's now more than three years obsolete.

amercader:
@TkTech I've removed the old multicore docs but I don't have many experience with the modern approach and the managed schema stuff. Perhaps you can help with this? We would need to take into account the version of Solr that you can install via package on Ubuntu 14.04 (the one that target our packages).

In any case I think this can go into a separate PR. Do you want to merge this one?

TkTech:
@amercader I think you might have gotten that backwards? The multicore docs stick around, the single-core docs disappear (single core mode has been obsolete for 3+ years)

amercader:
@TkTech the multicore docs I removed were meant to for Solr 1.x and involved a lot of manual editing of files, so I think they can be safely removed. As for the new multicore docs, as I said I'm happy with anything that can be done out of the box on the Ubuntu 14.04 packaged Solr version.

amercader:
@TkTech can we get this merged? These are just structural changes and we can deal with the new Solr versions on a separate PR

amercader:
@TkTech bump, would be really good to merge this. I know it's says Solr, but there's nothing controversial here, just structure changes.

TkTech:
Been awhile since I looked at this one so I just gave it another once over. 👍

PR #3090, opened 2016-06-04
[#3089] move RootPathMiddleware up the WSGI stack so it doesn't impact…

### LINKED ISSUE #3089: Login fails with 404 when using root_path
On master under Apache 2.2.4 and mod_wsgi with CKAN mounted at /data (`WSGIScriptAlias /data ...` and `ckan.root_path = /data`) on submitting the login form user is redirected to `/user/logged_in` (i.e., without the root_path) and Apache returns a 404.

MrkGrgsn:
… other middleware. Fixes #3089 

This PR moves the RootPathMiddleware up the WSGI stack so it doesn't impact other middleware including repoze.who plugins, which is the cause of #3089.

I figured this should only change the environment for CKAN and CKAN plugins so I've moved it up above routes.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware.py
+++ b/ckan/config/middleware.py
@@ -88,6 +88,8 @@ def make_pylons_stack(conf, full_stack=True, static_files=True, **app_conf):
     for plugin in PluginImplementations(IMiddleware):
         app = plugin.make_middleware(app, config)
 
+    app = RootPathMiddleware(app, config)
+
     # Routing/Session/Cache Middleware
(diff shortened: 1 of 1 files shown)


wardi:
Thanks @MrkGrgsn this was the way the middleware hack should have been added in the first place. It seems to not affect us on some installations and we traced it to a difference in in headers sent by the web server in the front. This should make the hack work better until we've got a proper solution.

PR #3092, opened 2016-06-06
Dockerfile Update

deinok:
-Upgrade of base image
-Not repository Independent (Now its not required to Docker source be in the same git)
-Pinned by default to 2.5.2
-Overriding CKAN_GIT_BRANCH gives the feature to install any branch in the main GitHub. Overriding to master gives an easy way to develop

### DIFF
--- a/ckan/config/middleware.py
+++ b/ckan/config/middleware.py
@@ -88,6 +88,8 @@ def make_pylons_stack(conf, full_stack=True, static_files=True, **app_conf):
     for plugin in PluginImplementations(IMiddleware):
         app = plugin.make_middleware(app, config)
 
+    app = RootPathMiddleware(app, config)
+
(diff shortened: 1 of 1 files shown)


wardi:
I'm not sure what's up with the circleci failure. Is your circle configured to use a different base image or something?

Also, do you mind squashing all your empty merge commits?

deinok:
@wardi CicleCi is using the dockerfile?
I only nodified the dockerfile.

If CircleCi uses the dockerfile, by default it gets the release-v2.5.2 and not the branch trying to test

wardi:
No, our tests don't depend on any of our docker images. Your circle CI configuration might be using a different base image than ours for running the tests, though. Would you check that?

deinok:
I will take a look and see what why is failing the solr

PR #3100, opened 2016-06-08
Fix #3081: Improve docs of `ckan.site_url` and `ckan.root_path`.

### LINKED ISSUE #3081: site_url vs. root_path vs. WSGIScriptAlias
I'm running a CKAN instance in a sub-"directory" of my host (at ``/data``). Therefore, I have set ``ckan.site_url`` to ``https://example.com/data``, as [described by the docs for `site_url`](http://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-site-url):

> If you want to mount CKAN at a path other than /, then this setting should reflect that, but the URL you mount it at is determined by your apache config (your WSGIScriptAlias path) (or equivalent for other servers).

That worked perfectly for quite some time, until I started uploading resource files via a script that called the action functions directly via `toolkit.get_action` (instead of using web requests). The generated download URLs would end up missing the `data` part of the URL. These URLs are constructed using `ckan.lib.helpers.url_for`, which only uses the host, port and scheme from `site_url` but not the path:

    from pylons import config
    from ckan.lib.helpers import url_for
    config['ckan.site_url'] = 'http://example.com/data'
    print(url_for(controller='package', action='search', qualified=True))
    # prints 'http://example.com/dataset'

Instead, `url_for` uses the [`ckan.root_path`](http://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-root-path) setting for that:

    config['ckan.site_url'] = 'http://example.com'
    config['ckan.root_path'] = '/data/{{LANG}}'
    print(url_for(controller='package', action='search', qualified=True))
    # prints 'http://example.com/data/dataset'

Unfortunately, the documentation of `root_path` is rather fuzzy and mostly talks about where the locale code is placed in the URL. In addition, since `site_url` is a much more prominent setting, it's easy to miss `root_path`.

After moving ``/data`` from `site_url` to `root_path`, my resource URLs were indeed generated correctly. However, no everything else is broken, because `url_for` seems to also take the WSGI root into account: If I set

    WSGIScriptAlias /my-wsgi-prefix /etc/ckan/default/apache.wsgi

in my Apache configuration for illustration (in reality it's `/data`) then I end up with URLs like `/my-wsgi-prefix/data/dataset`, which contain both the WSGI prefix and `root_path`. This only happens when the code runs via WSGI. which explains why my resources had URLs without the prefix while all other URLs were working: I hadn't set `root_path`, so the web UI only took the WSGI prefix plus the scheme, host and port from `site_url`. My resource creation script did not run under WSGI, so it only used scheme, host and port from `site_url`, resulting in the missing `data` prefix. Once I set `root_path`, the script starts working (no WSGI, so no double prefix), but the web UI breaks.

So something won't work, no matter how I'm doing things. Here's a suggestion on how things should work. What information do we need?

* For constructing fully qualified URLs (e.g. download URLs for uploaded resource data) we need the scheme, host, and port. On a web server, these values could be inferred from the environment, but since the code should also work standalone (e.g. for tests, paster scripts, etc.) they should be fixed in the configuration.
* In addition, we need to know how to construct URLs on that host. That includes a potentially non-root `WSGIScriptAlias` and a place for the locale code.

The important point here is that we should ignore values that come from the environment, since these may change depending on how the code is run.

There are multiple ways on how to map these values to configuration options. We could continue to use the existing approach of separate options for the host and the path, or we could combine them into a single setting, e.g.

    ckan.site_url = http://example.com/my-wsgi-prefix/{{LANG}}

The main goal should be a consistent behaviour across all environments and clear documentation.

Note that this and similar issues have been encountered many times before, see #2866, #2592, #2599, #2832, #2644, #3001, #2260, #3035.

torfsen:
Fixes #3081.

### DIFF
--- a/doc/maintaining/configuration.rst
+++ b/doc/maintaining/configuration.rst
@@ -287,15 +287,18 @@ Default value:  (an explicit value is mandatory)
 Set this to the URL of your CKAN site. Many CKAN features that need an absolute URL to your
 site use this setting.
 
-.. important:: It is mandatory to complete this setting
+This setting should only contain the protocol (e.g. ``http://``), host (e.g.
+``www.example.com``) and (optionally) the port (e.g. ``:8080``). In particular,
+if you have mounted CKAN at a path other than ``/`` then the mount point must
+*not* be included in ``ckan.site_url``. Instead, you need to set
+:ref:`ckan.root_path`.
 
-.. note:: If you want to mount CKAN at a path other than /, then this setting
-  should reflect that, but the URL you mount it at is determined by your
-  apache config (your WSGIScriptAlias path) (or equivalent for other servers).
+.. important:: It is mandatory to complete this setting
 
 .. warning::
 
   This setting should not have a trailing / on the end.
+
 
 .. _apikey_header_name:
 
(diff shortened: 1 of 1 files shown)


doc/maintaining/configuration.rst
```diff
@@ -1661,10 +1664,27 @@ Example::
...
-lets you change this. You can use any path that you want, adding ``{{LANG}}``
-where you want the locale code to go.
+This setting is used to construct URLs inside CKAN. It specifies two things:
+
+* *At which path CKAN is mounted:* By default it is assumed that CKAN is mounted
+  at ``/``, i.e. at the root of your web server. If you have configured your
+  web server to serve CKAN from a different mount point then you need to
+  duplicate that setting here.
+
+* *Where to locale is added to an URL:* By default, the URLs are formatted as
```
wardi: where *the* locale

torfsen: Fixed.

PR #3102, opened 2016-06-10
Fixes repr(DictizationError) returning ""

davidread:
I came across this when debugging. When you print a DictizationError (e.g. during package_update) it shows as blank, when it has a string inside.

### DIFF
--- a/ckan/lib/navl/dictization_functions.py
+++ b/ckan/lib/navl/dictization_functions.py
@@ -37,6 +37,9 @@ class DictizationError(Exception):
             return repr(self.error)
         return ''
 
+    def __unicode__(self):
+        return unicode(str(self))
+
 class Invalid(DictizationError):
     '''Exception raised by some validator, converter and dictization functions
     when the given value is invalid.


ckan/lib/navl/dictization_functions.py
```diff
@@ -37,6 +37,9 @@ def __str__(self):
             return repr(self.error)
         return ''
 
+    def __unicode__(self):
+        return unicode(str(self))
```
wardi: how about `unicode(self.error)` instead?

davidread: Because you'd need to do the `if hasattr(self, 'error') and self.error:` as well.

wardi: I tried this change but I get the same behaviour before and after when using `print myerror`. How about we return something other than `''` when self.error isn't set? `super(DictizationError, self).__str__()` is one option.

davidread:
Closing in favour of #3551

PR #3103, opened 2016-06-10
ckan.__version__ available as template helper

brew:
The BaseController in [lib/base.py](https://github.com/ckan/ckan/blob/997b06cf5627660ec7b9b1d8dca2f2463e901d1a/ckan/lib/base.py#L200) adds `__version__` to `c` so it's available to the [base.html](https://github.com/ckan/ckan/blob/997b06cf5627660ec7b9b1d8dca2f2463e901d1a/ckan/templates/base.html#L29) template. 

Instead of having this added to `c`, I suggest removing the line from base.py and add a new template helper, `ckan_version`, that returns `ckan.__version__`, for use in templates.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,11 @@
 Changelog
 ---------
 
+v2.6.0 TBA
+=================
+API changes and deprecations:
+ * Replace `c.__version__` with new helper `h.ckan_version()` (#3103)
+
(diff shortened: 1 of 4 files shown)


k-nut:
I'll update the changelog on this.

k-nut:
@brew just created a new PR that wants to merge into this branch and updates the changelog with this new change.

PR #3105, opened 2016-06-11
Updated DockerFile

deinok:
-Upgrade of base image
-Not repository Independent (Now its not required to Docker source be in the same git)
-Pinned by default to 2.5.2
-Overriding CKAN_GIT_BRANCH gives the feature to install any branch in the main GitHub. Overriding to master gives an easy way to develop

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,61 +1,59 @@
-FROM phusion/baseimage:0.9.15
+FROM phusion/baseimage:0.9.18
 MAINTAINER Open Knowledge
 
-# Disable SSH
-RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
-
-ENV HOME /root
+# Environment
 ENV CKAN_HOME /usr/lib/ckan/default
 ENV CKAN_CONFIG /etc/ckan/default
 ENV CKAN_DATA /var/lib/ckan
+ENV CKAN_GIT_BRANCH release-v2.5.2
 
 # Install required packages
-RUN apt-get -q -y update
-RUN DEBIAN_FRONTEND=noninteractive apt-get -q -y install \
-        python-minimal \
+RUN apt-get -q=2 -y update && DEBIAN_FRONTEND=noninteractive apt-get -q=2 -y install \
         python-dev \
+        python-pip \
         python-virtualenv \
-        libevent-dev \
         libpq-dev \
-        nginx-light \
         apache2 \
         libapache2-mod-wsgi \
+        nginx \
         postfix \
-        build-essential
-
-# Install CKAN
-RUN virtualenv $CKAN_HOME
-RUN mkdir -p $CKAN_HOME $CKAN_CONFIG $CKAN_DATA
+        && apt-get -q=2 -y clean
+        
+# Install Virtual Environment
+RUN mkdir -p $CKAN_HOME/src/ $CKAN_CONFIG $CKAN_DATA
+RUN virtualenv --no-site-packages $CKAN_HOME
 RUN chown www-data:www-data $CKAN_DATA
 
-ADD ./requirements.txt $CKAN_HOME/src/ckan/requirements.txt
+# Download CKAN
+ADD https://codeload.github.com/ckan/ckan/tar.gz/$CKAN_GIT_BRANCH $CKAN_HOME/ckan-git.tar.gz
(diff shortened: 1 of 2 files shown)


deinok:
@wardi Problem was that CircleCi - Ubuntu 14.04 don't have SOLR by default

Dockerfile
```diff
@@ -1,61 +1,59 @@
-FROM phusion/baseimage:0.9.15
+FROM phusion/baseimage:0.9.18
 MAINTAINER Open Knowledge
 
-# Disable SSH
-RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
-
-ENV HOME /root
```
wardi: newer phusion/baseimage has removed ssh?  eventual plan is to get off phusion/baseimage regardless, right?

Dockerfile
```diff
@@ -1,61 +1,59 @@
...
 ENV CKAN_HOME /usr/lib/ckan/default
 ENV CKAN_CONFIG /etc/ckan/default
 ENV CKAN_DATA /var/lib/ckan
+ENV CKAN_GIT_BRANCH release-v2.5.2
 
 # Install required packages
-RUN apt-get -q -y update
-RUN DEBIAN_FRONTEND=noninteractive apt-get -q -y install \
-        python-minimal \
+RUN apt-get -q=2 -y update && DEBIAN_FRONTEND=noninteractive apt-get -q=2 -y install \
```
wardi: extra -q=2 just to be extra quiet? :smile:

Dockerfile
```diff
@@ -1,61 +1,59 @@
...
 
-# Disable SSH
-RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
-
-ENV HOME /root
+# Environment
 ENV CKAN_HOME /usr/lib/ckan/default
 ENV CKAN_CONFIG /etc/ckan/default
 ENV CKAN_DATA /var/lib/ckan
+ENV CKAN_GIT_BRANCH release-v2.5.2
```
wardi: This will push a release branch instead of master to ckan/ckan:latest, won't it?

Dockerfile
deinok: Plan is use httpd (Apache2) - phusion/baseimage is not maintained

Dockerfile
deinok: yes hahaha people only want to see where is the problem, not if something works well

Dockerfile
deinok: The idea, is give the feature of change the deployed branch. This makes the docker image ready for production. If somebody wants to change to another release(branch of ckan/ckan, master included), they only have to change that ENV.

But there are a know issue:
If you build or pull the docker file and ENV is release-v2.5.2, and then you change that ENV to for example "master", there will be no changes, because docker cache every layer[https://github.com/docker/docker/issues/3199(https://github.com/docker/docker/issues/3199)

The solution, is to use "build --no-cache"

I think I should add documentation about how to use it, this is safe for production

Dockerfile
k-nut: For the lazy people from the future looking at this (took me some time to see what wardi was referring to):
```
apt-get -q=2 -y update && DEBIAN_FRONTEND=noninteractive apt-get -q=2 -y install
         ^^^                                                      ^^^
```

wardi: @k-nut yes, I think we only need one of those

Dockerfile
wardi: I'd rather not build the 2.5 image from the master branch, and I don't like downloading the source as a tarball from the Dockerfile.

Docker hub automated builds should build the branch that triggered them. CKAN 2.6 will be the first release with this docker configuration, so we can start docker hub automated builds for stable releases after 2.6 is out by just adding that branch to docker hub's automated builds. If you'd like to build 2.5 (and earlier) then we can consider back-porting just the docker configuration once you're happy with it.

I don't think it would ever make sense for a stable docker image to change just because the docker configuration on master changed. If there is a fix necessary we can backport it along with our other bug fixes.

deinok: Okey, this was only an idea to solve the release of old CKAN releases. If it is going to be mantained and trigered by the push of branch, I can work on the release of 2.6.

Then, should I start developing the 2.6.X  DockerImage and only the 2.6 in mind?

wardi: Develop against master and when we fork for 2.6 it should just work for the release too.

deinok:
@wardi I Close this

Issue #2795, opened 2015-12-17
Travis error when submitting to coveralls after tests run


brew:
Travis sends the coverage report to coveralls.io after a successful test run with the `coveralls` python package. This is silently broken with the error:

```sh
$ coveralls
Submitting coverage to coveralls.io...
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.9/bin/coveralls", line 11, in <module>
    sys.exit(main())
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/coveralls/cli.py", line 49, in main
    result = coverallz.wear()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/coveralls/api.py", line 73, in wear
    data = self.create_data()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/coveralls/api.py", line 122, in create_data
    self._data = {'source_files': self.get_coverage()}
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/coveralls/api.py", line 130, in get_coverage
    workman._harvest_data()
AttributeError: 'Coverage' object has no attribute '_harvest_data'
```

This is due to a change in coverage[1] that removes the private _harvest_data method.

We pin coveralls, but not coverage, so it's likely we're running an older version of coveralls that uses the private method that no longer exists in the newer, unpinned coverage.

I suggest unpinning coveralls, so we're always using the newest versions of both coveralls and coverage. Usually, I'm against adding unpinned package versions to dependencies, but in this case the libraries are used only for development, testing and automated testing, and don't affect the functionality of CKAN itself.

[1] https://github.com/nedbat/coveragepy/blob/master/CHANGES.rst#version-40b1-----2015-08-02

PR #3109, opened 2016-06-13
Dev requirements upgrade

deinok:
Upgrades all dev-requirements.txt


brew:
We unpinned Coveralls so it's always up-to-date with the webservice. We had a problem in the past with changes that prevented coverage submission because we were using an outdated lib (https://github.com/ckan/ckan/issues/2795).

deinok:
@brew Okey, I added a comment on coveralls

PR #3110, opened 2016-06-13
Remove tag validators

### LINKED ISSUE #2780: Tags' "name" attribute documentation (or code) is wrong
Although docs says that the name has to be between 2 and 100 characters, and contain only alphanumeric characters, `-`, `_`, and `.`, the validator also accepts whitespace:

``` python
# https://github.com/ckan/ckan/blob/master/ckan/logic/validators.py#L447-L453
def tag_name_validator(value, context):

    tagname_match = re.compile('[\w \-.]*$', re.UNICODE)
    if not tagname_match.match(value):
        raise Invalid(_('Tag "%s" must be alphanumeric '
                        'characters or symbols: -_.') % (value))
    return value
```

At first I thought whitespace could be part of "alphanumeric", but that's not the case. Python's `\w` definition is to match alphanumeric, and it doesn't match whitespace.

I'm not sure if the issue is with the documentation or the actual code

hamhands:
Fixes #2780 

### Proposed fixes:
Removes `tag_name_validator` and `tag_length_validator` as per @amercader's context in #2780

### DIFF
--- a/ckan/logic/schema.py
+++ b/ckan/logic/schema.py
@@ -28,8 +28,6 @@ from ckan.logic.validators import (
     package_name_validator,
     package_version_validator,
     group_name_validator,
-    tag_length_validator,
-    tag_name_validator,
     tag_string_convert,
     duplicate_extras_key,
     ignore_not_package_admin,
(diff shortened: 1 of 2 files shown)


wardi:
No validation is bad. How about something like what we're using here:

https://github.com/open-data/ckanext-canada/blob/wet4-scheming/ckanext/canada/validators.py#L52-L92

This restricts whitespace, unicode characters that don't have glyphs and sets a reasonable size limit of 2-140 characters

hamhands:
Definitely agree @wardi and thanks for the example. This PR is actually premature as I meant to merge this on my own fork. Will open a new one with proper tests once I'm done developing.

amercader:
@hamhands It's perfectly fine to muddy the PR history, that's what they are for :) Thanks for looking into this

PR #3118, opened 2016-06-16
[WIP] Death To Solr; Long Live Solr

TkTech:
This is a WIP PR so that progress, comments and review can be done.

This PR has multiple goals:
- [x] Remove Old Search
- [x] Implement the `IIndexer` interface allowing plug-and-play search engines
- [x] Implement `NoIndexer`,  a no-op indexer used for testing.
- [ ] Implement `SolrIndexer`, an indexer compatible with the existing Solr setup.
- [ ] Implement `PostgresIndexer`, an indexer that only requires modern postgres and implements faceting (this is what I want to be the new default indexer in new CKAN installs)
- [x] Implement `SolrCloudIndexer`, a solr indexer that supports multi-server Solr.

Additional stretch goals for this are:
- [ ] `ElasticSearchIndexer`
- [ ] `AWSCloudSearchIndexer`

All indexers included in the core MUST support at a minimum, to reproduce existing behaviour:
- [ ] Basic search query, limit, paging, sort.
- [ ] Single-facet search

The interface should also be improved to support:
- [x] Multi-facet search
- [ ] Column declaration (provided to support [Solr managed schemas](https://cwiki.apache.org/confluence/display/solr/Schemaless+Mode), [ElasticSearch dynamic mapping](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/mapping.html))


deinok:
Should we upgrade Solr to 5.X or 6.X ?

torfsen:
@deinok I would definitely welcome an update to a recent version.

TkTech:
I am targeting 5+, but it should also continue to work on older versions if you keep using a manual `schema.xml`. With column declaration, I hope to make schema.xml obsolete and allow the schema to be managed within CKAN.

deinok:
@torfsen @TkTech  Docker solr image of CKAN is using Solr 6.0 and uses schema.xml.
If we move to a newer version of Solr, we should also update CircleCI (It only supports 4.X on Ubuntu 12.04)

ckan/lib/search/__init__.py
```diff
@@ -1,318 +1,23 @@
...
 
-from pylons import config
-from paste.deploy.converters import asbool
 
-import ckan.model as model
-import ckan.plugins as p
-import ckan.logic as logic
+def index(package):
+    for plugin in plugins.PluginImplementations(plugins.IIndexer):
+        return plugin.index(package)
```
k-nut: Just stumbled over this and was wondering if this is the standard way to check for plugins or if there is another way? It seems a little confusing to me that we are looping over the plugins here when what we are actually doing is taking the first implementer and returning its value. Meaning that the loop implicates that we are doing something for each plugin but we already return after the first one...

TkTech: Simply the pattern I copied from the existing code. If there's a better way I'm happily open to it.

k-nut: That's what I thought. I don't know if there is a cleaner way to do this but maybe @wardi or @amercader do?!

torfsen: If we only expect that a single `IIndexer` implementation is active at a given time then I'd prefer
```python
indexer = list(plugins.PluginImplementations(plugins.IIndexer))[0]
return indexer.index(package)
```
This will at least throw an error if no indexer is registered.

If multiple indexers are to be supported then I'd suggest
```python
for plugin in plugins.PluginImplementations(plugins.IIndexer):
    plugin.index(package)
```
altough it's not clear what the correct return value would be in that case.

wardi: Multiple indexers would be useful to live-migrate a site to a new indexer.

1. add second indexer to end of plugins

   `warning: multiple IIndexer plugins enabled. Only the first will be used for search.`

2. search-index rebuild --only-missing to populate the new index. new and changed datasets will be updated in both indexes.

3. reorder the indexes to search with the new plugin (possibly just on one front-end)

4. if everything looks good, remove the old plugin. if not, switch back. Both plugins could be enabled for as long as it takes to make sure all your users are happy

I'd like to loop over each one when indexing, with a shallow copy because I'm sure some plugin is going to modify the dict in-place:
```python
for plugin in plugins.PluginImplementations(plugins.IIndexer):
    plugin.index(dict(package))
```

TkTech:
I should have time to work on this again with the upcoming weekend, I appreciate the feedback!

TkTech:
Just an update since I realize this got more interest from the ckan gov group. This PR is very out of date and is *nothing* like the final implementation, please do not waste effort looking at this as-is. I will be updating this over the weekend.

sammcj:
+100 for moving to Elasticsearch

amercader:
@TkTech How far would be this (or whatever most recent version you've been working on) for a first review / version? Is it worth writing a short spec of what's missing to see if someone else can pick it up? (in case you don't have availability right now)

TkTech:
@amercader it's pending completion of https://github.com/TkTech/sachi which is currently as-time-permits project. As I mentioned this entire PR can be thrown out, as doing it inside CKAN is far less desirable then a re-usable, easily tested external module.

PR #3120, opened 2016-06-17
Tidy-up BaseController pagination code

brew:
In the past, the BaseController used to deal with a lot of pagination logic used by subclassing controllers. Most of this has since been moved to the controllers themselves that make use of a Page helper in `lib/helpers.py`. There's still a `_get_page_number` method on the BaseController that gets and validates a page number from the request params. This can also be moved to the helpers module, providing for better reuse by controller classes and, as part of the Flask migration, by flask views.

There are tests in [tests/controllers/test_feed.py](https://github.com/ckan/ckan/blob/9f59e91113a7e129bd4ce801f3fae5835d0baef4/ckan/tests/controllers/test_feed.py) that provide coverage for this method.

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1097,7 +1097,8 @@ def get_page_number(params, key='page', default=1):
             raise ValueError("Negative number not allowed")
     except ValueError:
         import ckan.lib.base as base
-        base.abort(400, ('"{key}" parameter must be a positive integer'))
+        base.abort(400, ('"{key}" parameter must be a positive integer'
+                   .format(key=key)))
 
     return p
 
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1082,6 +1082,27 @@ class Page(paginate.Page):
 
 
 @core_helper
+def get_page_number(params, key='page', default=1):
+    '''
+    Return the page number from the provided params after verifying that it is
+    an positive integer.
+
+    If it fails it will abort the request with a 400 error.
+    '''
+    p = params.get(key, default)
+
+    try:
+        p = int(p)
+        if p < 1:
+            raise ValueError("Negative number not allowed")
+    except ValueError:
+        import ckan.lib.base as base
+        base.abort(400, ('"page" parameter must be a positive integer'))
+
+    return p
+
+
+@core_helper
 def get_display_timezone():
     ''' Returns a pytz timezone for the display_timezone setting in the
     configuration file or UTC if not specified.
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1097,7 +1097,7 @@ def get_page_number(params, key='page', default=1):
             raise ValueError("Negative number not allowed")
     except ValueError:
         import ckan.lib.base as base
-        base.abort(400, ('"page" parameter must be a positive integer'))
+        base.abort(400, ('"{key}" parameter must be a positive integer'))
 
     return p
 
--- a/ckan/controllers/feed.py
+++ b/ckan/controllers/feed.py
@@ -310,7 +310,7 @@ class FeedController(base.BaseController):
                 search_params[param] = value
                 fq += ' %s:"%s"' % (param, value)
 
-        page = self._get_page_number(request.params)
+        page = h.get_page_number(request.params)
 
         limit = ITEMS_LIMIT
         data_dict = {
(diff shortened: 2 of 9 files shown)


brew:
The last two commits remove a couple of unused module variables.

TkTech:
Every time you remove a line from `BaseController`, an orphan and an abandoned puppy get adopted together.

ckan/lib/helpers.py
```diff
@@ -1082,6 +1082,27 @@ def _range(self, regexp_match):
...
+    '''
+    p = params.get(key, default)
+
+    try:
+        p = int(p)
+        if p < 1:
+            raise ValueError("Negative number not allowed")
+    except ValueError:
+        import ckan.lib.base as base
+        base.abort(400, ('"page" parameter must be a positive integer'))
```
wardi: `'"{key}" parameter must be...`

k-nut:
Hey @brew would you rebase this or just push a new empty commit so that we can trigger QuantifiedcCode again and make sure that the reporeted issues go away with the new configuration?

brew:
Oof - I rebased it and now it's a mess.

k-nut:
hm, thats not good.. Surprisingly the Quantifiedcode check also went away completely apparently?! I'm still not sure when that runs and when it does not...

torfsen:
@brew That doesn't look like a rebase to me -- your branch now includes a lot of unrelated commits from `master`, so this looks more like a merge from `master` onto your branch. Do you recall the exact steps you took?

brew:
@torfsen 

> git checkout master
> git checkout 3120-get-page-number 
> git rebase master
> git pull
> git push

I've now hard reset this branch to make is sane again.

torfsen:
@brew Looks OK to me, although I don't whether the `pull` after the `rebase` makes much sense. My usual PR rebase workflow is
```
git checkout master
git pull
git checkout pr-branch
git rebase master
git push --force
```

brew:
@torfsen The pull after the rebase is where it when wrong, I think. I tried to push after the rebase, but git complained:

```sh
$ git push
To git@github.com:ckan/ckan.git
 ! [rejected]        3120-get-page-number -> 3120-get-page-number (non-fast-forward)
error: failed to push some refs to 'git@github.com:ckan/ckan.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

So I pulled at this point, when I should have force pushed, like you suggest.

k-nut:
@brew @torfsen yep, that is the thing with rebasing. You do always need to force push after it...

Well so we all learned something about git but quantifiedcode still complains for no apparent reason 😕 

I guess the only option then is to just ignore that its complaining. If this problem persists for other PRs we might have to remove that integration again...

wardi:
Let's try to avoid rebasing and push --force on ckan/ckan.

This is our primary repo and lots of people base their work on it. rebasing can create a huge mess for those downstream developers. Also, rebasing can create a whole lot of "fake" history: repo states that never actually existed or were tested against. This is a problem when we try to use bisect to locate the source of a regression, because the whole set of rebased commits might be broken making it really hard to pin down the change responsible.

Finally, if quantified code needs a useless commit or change because it's missing a "re-run test" option, I'd rather ignore it warnings or remove its integration.

torfsen:
@wardi I think rebasing and force-pushing a PR is a valueable tool to get clean merges and a useful commit history. I don't see any risk there, since people usually don't base other work on a PR. Rebasing or force-pushing ckan/ckan's `master` is of course something completely different and should be avoided if necessary.

wardi:
@torfsen we should decide on a policy for rebasing. It would be great to have you join our dev meetings if you have time (they are usually quite short), we could talk about this on Tuesday http://pad.okfn.org/p/ckan-meeting .

torfsen:
@wardi Sounds like a good idea. The Tuesday meeting works for me.

ckan/lib/helpers.py
```diff
@@ -1082,6 +1082,27 @@ def _range(self, regexp_match):
...
+    '''
+    p = params.get(key, default)
+
+    try:
+        p = int(p)
+        if p < 1:
+            raise ValueError("Negative number not allowed")
+    except ValueError:
+        import ckan.lib.base as base
+        base.abort(400, ('"{key}" parameter must be a positive integer'))
```
amercader: Should we not format `{key}` with the actual value here?

brew: Yes. I'm an idiot. 😞

PR #3122, opened 2016-06-17
Test that string literals have an `b`, `u` or `ur` prefix

torfsen:
This PR adds a test that verifies that each string literal has an `b`, `u`, or `ur` prefix (see the [CKAN docs on Unicode handling](http://docs.ckan.org/en/latest/contributing/unicode.html#string-literals) and #3068).

All existing Python source files are explicitly white-listed in the test. Over time, these files should be fixed and delisted. New files should not be added to the white-list but should directly fixed instead.

### DIFF
--- a/bin/running_stats.py
+++ b/bin/running_stats.py
@@ -1,3 +1,5 @@
+# encoding: utf-8
+
 '''Tool for a script to keep track changes performed on a large number
 of objects.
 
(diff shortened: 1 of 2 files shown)


wardi:
@torfsen this looks reasonable, would you resolve the merge conflict that has appeared?

torfsen:
@wardi: Fixed and rebased.

PR #3127, opened 2016-06-17
Remove Python 2.6 compability

### LINKED ISSUE #3126: Drop Python 2.6 support
It was announced to the mailing list by @amercader in January that support for python 2.6 was going to be dropped with ckan 2.6 (what a wonderful coincidence).

https://lists.okfn.org/pipermail/ckan-dev/2016-January/009608.html

> Starting from the next CKAN 2.6 version, the minimum Python version
> required for CKAN will be Python 2.7 (we recommend using the latest
> Python patch release).

There still are some backports and workaraounds for 2.6 in the code and the documentation needs upgrading. 

### Python files
```
ckan/lib/util.py
13:# subprocess.check_output() isn't in Python 2.6.
15:# Python 2.6.

ckan/model/extension.py
17:        "Replaces stdlib operator.methodcaller in python <2.6"

ckan/tests/helpers.py
36:    # Python 2.6 doesn't have these, so define them here

ckanext/datapusher/tests/test.py
87:        # Httpretty crashes with Solr on Python 2.6,

ckanext/resourceproxy/tests/test_proxy.py
63:        # Httpretty crashes with Solr on Python 2.6,
``` 

### Documentation:
```
doc/contributing/python.rst
77:We use `the Python standard library's logging module <http://docs.python.org/2.6/library/logging.html>`_

doc/contributing/string-i18n.rst
339:  Worse (doesn't work in Python 2.6):

doc/maintaining/installing/install-from-source.rst
36:Python                 `The Python programming language, v2.6 or 2.7 <http://www.python.org/getit/>`_
```

I'll prepare a PR to remove all of those things.

k-nut:
Fixes #3126 

Work in progress


k-nut:
This is good for review now. I removed everything mentioned in #3126 with the exception of 
```
doc/contributing/string-i18n.rst
339:  Worse (doesn't work in Python 2.6):
```
as this is still a valid example on how to properly write formattable strings.

There were quite a lot of files changed but it should be relatively easy to review this on a commit by commit basis as they are all independent of each other and limited on one thing each. 

My approach for this was simply to grep for `2.6` and `2.7` and see if that is mentioned anywhere. There might of course be other places where no one bothered to write a comment about a workaround/backport that could also be changed now.

Very much looking forward to your thoughts!

amercader:
Well done @k-nut! :fire: Python 2.6 :fire:

PR #3134, opened 2016-06-24
Bower - Added Support

deinok:
This adds support for Bower for the JS/CSS libs

---
This change is [Reviewable](https://reviewable.io/reviews/ckan/ckan/3134)


k-nut:
So the idea is that we use bower to track dependencies and also track the files in git, correct?

deinok:
well, I prefer to use the command `bower install` but for now this don't add any more changes.
But yes, my idea is delete the vendor folder, and install it with bower

k-nut:
I would also prefer to install with bower but I also see that that would add another step in the `from source` readme (and a dependency for node) that might confuse new users. So we should probably discuss that some more.

But 👍  for this change already!

deinok:
Yes, i think that we need to discuss it more, this is why I add the files in git ;)

TkTech:
100% against this. Absolutely no reason to add node.js, npm, and bower to the ever-growing list of dev dependencies.

These resources almost never change and updating them takes a few minutes at most.

deinok:
For now, this dont break anything, it only adds the support, even without bower everthing will work well. But we should discuss if bower must be mandatory or not

deinok:
Any update about this?

wardi:
We missed our usual meetings last week, but we'll talk about this tomorrow. You're welcome to join if you have time! http://pad.okfn.org/p/ckan-meeting

amercader:
@TkTech I believe you had to add your thoughts about this

TkTech:
Thank you for the contribution! However....

1. We're getting rid of Node.js as a dependence, which only exists due to Less.
2. This adds 1.7 *thousand* files to the CKAN repo. This is ridiculous. It's unacceptable just from this point.
3. These files are almost never modified internally and version bumping them is only a few minutes of effort. It isn't worth the additions.
4. Nothing prevents you from using this on your own fork/branch/custom theme.

deinok:
I only add the 1.7 thousand files to don't break anything for now, but the idea is delate the vendor folder and use `bower install`
Also i think it would be great to specify the version of every vendor library

PR #3136, opened 2016-06-27
Document Unicode filename handling

torfsen:
This is work for #3006.

The PR updates the documentation with information on how to deal with Unicode filenames. It also adds two helper functions (`decode_path` and `encode_path`) for easy handling of byte string filenames.

Since our Unicode documentation has a normative character this PR also is a good time to discuss whether we want to handle Unicode filenames that way.

### DIFF
--- a/ckan/lib/io.py
+++ b/ckan/lib/io.py
@@ -1,0 +1,52 @@
+# encoding: utf-8
+
+u'''
+Utility functions for I/O.
+'''
+
+import sys
+
+
+_FILESYSTEM_ENCODING = unicode(sys.getfilesystemencoding()
+                               or sys.getdefaultencoding())
+
+
+def encode_path(p):
+    u'''
+    Convert a Unicode path string to a byte string.
+
+    Intended to be used for encoding paths that are known to be
+    compatible with the filesystem, for example paths of existing files
+    that were previously decoded using :py:func:`decode_path`. If you're
+    dynamically constructing names for new files using unknown inputs
+    then pass them through :py:func:`ckan.lib.munge.munge_filename`
+    before encoding them.
+
+    Raises a ``UnicodeEncodeError`` if the path cannot be encoded using
+    the filesystem's encoding. That will never happen for paths returned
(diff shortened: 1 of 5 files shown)


wardi:
ckan.lib.helpers are primarily template helpers. I can't think of a reason to use `h.encode_filename()` inside a jinja template, so we should find a better place for them, maybe just a small new module under ckan/lib.

These functions should also either document what exceptions they raise when encoding or decoding fails , or decide on a munging scheme that will work with any filesystem encoding.

torfsen:
@wardi Yeah, I was skeptical about `ckan.lib.helpers`, too, but didn't want to introduce a new module. I've now put the functions into `ckan.lib.io` instead.

Regarding failures during de-/encoding: If your environment is set up correctly and you're using paths returned by Python's I/O functions then `decode_path` and `encode_path` should never fail. The situations in which they can fail are:

1. You have existing files with filenames that are not representable using your filesystem's encoding. This can happen, for example, if you set the environment variable `LC_CTYPE` to a wrong value. This is not a problem that can be fixed inside CKAN, so `decode_path` will throw a `UnicodeDecodeError`. In my opinion that is better than the behavior of, for example, `os.listdir` (which will return such filenames as byte strings even if you request Unicode output).
2. You're generating filenames for new files from input that contains characters that are not representable using your filesystem's encoding. In that case you should use `ckan.lib.munge.munge_filename` before encoding the filename.

I've updated the docstrings and the documentation with information about that.

I think it's important to keep two issues distinct:

1. De-/encoding paths of existing files so that they can be internally stored as Unicode (that's what `decode_path` and `encode_path` are for)
2. Make sure that new filenames can be encoded using the filesystem's encoding (that's what `munge_filename` does by removing/replacing anything non-ASCII)

Because they're intended for existing filenames, `decode_path` and `encode_path` must not munge these filenames, hence they fail hard upon errors.

I've updated `munge_filename` to make sure that it always returns Unicode, even if passed a byte string.

wardi:
@torfsen great work as usual. Thank you!

PR #3139, opened 2016-06-28
Change all example uses of domain names to be RFC compliant special-use names

fincham:
Hi CKAN devs,

We discovered a local installation of your software clogging up our mail queues with undeliverable messages, due apparently to a live domain being used in example configurations and documentation.

This pull request updates all instances we spotted in the sample configuration and documentation where live example domain names are given, and replaces them with "example.com", a domain reserved explicitly for this purpose.

As described in RFC 2606 and RFC 6761, a number of domains such as example.com and example.org are maintained for documentation use, without prior coordination with the IANA.

-Michael

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -166,11 +166,11 @@ ckan.hide_activity_from_users = %(ckan.site_id)s
 
 ## Email settings
 
-#email_to = you@yourdomain.com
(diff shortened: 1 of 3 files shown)


amercader:
Thanks @fincham, sorry for clogging your email queue

fincham:
Thanks for merging super quickly! To be fair, the people who deployed the code at our end probably deserve at least some of the blame :)

PR #3148, opened 2016-07-01
Remove WSGIParty requirement

amercader:
When the App Dispatcher middleware was implemented on #2905 we used [WSGI Party](http://wsgi-party.readthedocs.org/en/latest/) to allow communication between the Pylons and Flask applications. After having worked extensively on the Flask POC branches this has turned out to be unnecessary and to add quite a lot of complexity, specially with the language used around parties, invites, etc.

To set it up there were two internal requests that were done while the app was being initialized, which were really confusing when trying to debug things. We didn't find any use case for having the two applications actually talk to each other, and the asking each app if they can handle the incoming request can be done just by calling a method directly.

This PR removes all WSGI Party logic and its requirement, and extends the `Flask` and `PylonsApp` app objects with our own `can_handle_request(environ)` method, which is directly called by `AskAppDispatcherMiddleware` on each request.


ckan/config/middleware/flask_app.py
```diff
@@ -27,38 +22,19 @@ def hello_world():
     def hello_world_post():
         return 'Hello World, this was posted to Flask'
 
+    # Add a reference to the actual Flask app so it's easier to access
+    setattr(app, '_wsgi_app', flask_app)
```
wardi: why not `app._wsgi_app = flask_app`?

ckan/tests/config/test_middleware.py
```diff
@@ -246,12 +236,14 @@ def test_ask_around_pylons_extension_route_get_before_map(self):
...
-        eq_(len(answers), 1)
-        eq_(answers[0][0], True)
-        eq_(answers[0][1], 'pylons_app')
-        eq_(answers[0][2], 'extension')
+        eq_(len(answers), 2)
+        eq_(answers[0][0], False)
+        eq_(answers[0][1], 'flask_app')
+        eq_(answers[1][0], True)
+        eq_(answers[1][1], 'pylons_app')
+        eq_(answers[1][2], 'extension')
```
wardi: why not just `eq_(answers, [(False, 'flask_app'), (True, 'pylons_app', 'extension')])`?  I think you get pretty-printed diffs of the lists when they don't match that way (and it's a shorter test)

PR #3152, opened 2016-07-05
#2772 - Remove LESS + Node.js Requirements

### LINKED ISSUE #2772: Switch the css pre-processor from LESS to SASS
As part of the upgrade to Bootstrap3 (#2602) I would like to propose changing the CSS pre-processor from LESS to SCSS (SASS). The reasoning is that future version of Bootstrap will be using SASS/SCSS as it's preporcessor.

TkTech:
Fixes #2772.

This transcompiles our base less assets to scss, which makes upcoming work to transition to newer versions of bootstrap simpler (bootstrap migrated from less to scss).

Additionally, it removes node.js and less as requirements in favour of pyScss, which also makes future Flask migration work easier. We can deprecate the current (loathed) asset management in favour of flask-webassets/webassets, which integrates with pyScss.


TkTech:
The scope of this ticket is expanding slightly. I am taking this time to cleanup theming and templating issues that are either directly or closely related.

For example, I have removed the old "color schema generation" feature from the CLI as well as the respective options from the admin panel. This feature was never completed and a cause of many stackoverflow questions. For example, manually editing the list of available themes in a dict inside admin.py was required to make it available. This is not acceptable.

I still think this might be a neat feature, but it should be an extension and should be finished.

deinok:
I like the idea of move to SASS, but i think that node should not be removed

TkTech:
One thing to note is that the theming can guide has always completely ignored these features and suggested just editing the template to include your CSS, so we do not need to update the documentation.

kmbn:

PR #3153, opened 2016-07-05
Improve JavaScript i18n documentation.

torfsen:
Reorganizes the existing documentation in a single place that is linked to from related topics. Also adds more details, for example regarding `paster trans js`.


amercader:
Great stuff @torfsen

PR #3158, opened 2016-07-08
[#2952] Depcreate `db load` and `db dump`

### LINKED ISSUE #2952: Escape special characters in password while db loading

torfsen:
Fixes #2952.

Deprecates the paster commands `db load` and `db dump` in favor of PostgreSQL's `pg_dump` and `pg_restore`. `db load` and `db dump` are kept but print a warning, and their documentation is replaced with a documentation of how to dump/import the database using `pg_dump` and `pg_restore`.

I've taken the chance and re-organized the database management documentation a bit. It was distributed all over the docs, with the same commands being documented in multiple places. There's now a central "Database Management" document to which the other places link.


wardi:
@torfsen thank you! this is great.

PR #3163, opened 2016-07-13
[#2842] New CKAN config object, decoupled from Pylons

amercader:
This PR includes a bunch of changes that provide a common configuration object for CKAN that works both on Flask and Pylons.

Summary:

* Changes in the way the environment is set up to make it more generic  and pass the config options from the ini file to Flask (048632d)
* A new dict-like, thread-local safe (using Werkzeug's Local object) CKAN config object, that apart from holding the config values proxies any changes to the Flask and Pylons own config objects (ad46dce)
* Update *all* core imports of `pylons.config` to `ckan.common.config` (These are the bulk of the changes in a77ff9c)
* Add `config` to the plugins toolkit (37ecbe8)

Discussion:

Both Flask and Pylons config objects are dict-like objects with key/value configuration options. But the Flask one is bound to the application and so it's only available when there is an application context available. This means that it can not support the way in which we were using the Pylons config, which was essentially available everywhere, on the CLI, tests, etc. In order to not change completely the way the application is structured and how the configuration is accessed in all parts of the code it seemed a good approach to have our own dict-like config object that was available in the same way the Pylons config object is.

At the same time we don't want to have different sets of configurations for stuff relevant to CKAN, Flask or Pylons (eg `ckan.site_url`, `PROPAGATE_EXCEPTIONS` or `pylons.app_globals` respectively). The CKAN config objects proxies any changes to the Flask and Pylons config objects (if they are available) so you can still configure all three applications using the same ini file (or env vars).

Moving forward this should be the only configuration object used in all code unless we really
need to access the underlying Flask or Pylons objects for some reason.

Extensions should be encouraged to use `toolkit.config` (or `ckantoolkit.config`, PR incoming).


ckan/plugins/toolkit.py
```diff
@@ -356,6 +366,7 @@ def _add_ckan_admin_tabs(cls, config, route_name, tab_label,
         # update the config with the updated admin_tabs dict
         config.update({config_var: admin_tabs_dict})
 
+#        import ipdb; ipdb.set_trace()
```
davidread: oops

davidread:
Looks excellent to me.

wardi:
Looks good to me too.

amercader:
See also https://github.com/ckan/ckantoolkit/pull/5 for extensions that want to support multiple CKAN versions

PR #3164, opened 2016-07-14
[iar#178] Allow customization of robots.txt

torfsen:
Serve `/robots.txt` from a template instead of from a static file to allow for customization by extensions and site owners. See https://github.com/ckan/ideas-and-roadmap/issues/178.

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -428,6 +428,9 @@ def make_map():
         m.connect('/testing/primer', action='primer')
         m.connect('/testing/markup', action='markup')
 
+    # robots.txt
+    map.connect('/(robots.txt)', controller='template', action='view')
+
     # Mark all unmarked routes added up until now as core routes
     for route in map.matchlist:
         if not hasattr(route, '_ckan_core'):
--- a/ckan/controllers/template.py
+++ b/ckan/controllers/template.py
@@ -2,12 +2,13 @@
 
 import ckan.lib.base as base
 import ckan.lib.render
+from ckan.common import response
 
 
 class TemplateController(base.BaseController):
 
     def view(self, url):
-        """By default, the final controller tried to fulfill the request
+        u"""By default, the final controller tried to fulfill the request
         when no other routes match. It may be used to display a template
         when all else fails, e.g.::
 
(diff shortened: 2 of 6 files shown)


wardi:
You've added the txt handling to the code that magically renders files it finds in the template directory. We've talked about removing this code or making it only match files in a specific template subdirectory. Would you add a simple route/controller for this file instead of extending this magic?

torfsen:
@wardi Sure, a separate controller would work for me. That's what I had planned in the beginning, but then I discovered the template controller :wink:

torfsen:
On a second thought a separate controller just for serving a single file seemed like overkill. So I've now added an explicit route to `/robots.txt` which uses `TemplateController`. That way, the "magic" part of `TemplateController` (i.e. its use as a fallback handler) can be removed in the future.

TkTech:
I'm okay with this if no one has additional comments I'll merge it tomorrow.

torfsen:
Is there anything I can do to help this move forward?

TkTech:
Feel through the cracks, apologies.

torfsen:
No problem, thanks for the merge!

stefina:
Alright, of course I do find a solution as soon as I post my question :P But I would be happy if somebody could explain me if what I did is actually a fix or an ugly hack..

I changed this line https://github.com/ckan/ckan/pull/3164/files#diff-959e9c429a9af0fc9baf8ecdc3465e50R33
to this:
```        
if url.endswith(u'.txt'):
     response.headers[b'Content-Type'] = b'text/plain; charset=utf-8'
```

Maybe this is a bug after all?

stefina:
Ah, never mind my conversation with myself. Just found out, that in the most current version, this is indeed 'the fix' :)
No worries, and sorry for bothering <3

amercader:
@stefina no problem, so you are saying that this fix wasn't backported to 2.7.x?

stefina:
@amercader at least it does not look like it:
https://github.com/ckan/ckan/blob/ckan-2.7.4/ckan/controllers/template.py#L33

amercader:
Thanks, we'll add it to 2.7.5

amercader:
Note: backport the headers fix mentioned above, not the whole PR

Issue #2977, opened 2016-04-22
Provide support for background tasks


rossjones:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=33227756)](https://www.bountysource.com/issues/33227756-provide-support-for-background-tasks?utm_source=33227756&utm_medium=shield&utm_campaign=ISSUE_BADGE)

**If you are interested in the bounty, please mark yourself as 'working on it' at [https://www.bountysource.com/issues/33227756-provide-support-for-background-tasks](https://www.bountysource.com/issues/33227756-provide-support-for-background-tasks)**

Currently running background tasks (sending email and so forth) is a bit too complicated.  

There is an incomplete PR at https://github.com/ckan/ckan/pull/2706 which introduces support for this using python-rq. There is also more discussion at https://github.com/ckan/ideas-and-roadmap/issues/66
### CKAN Version if known (or site URL)

Based off `master`
### Please describe the expected behaviour

Support as per Ideas issue and PR 
### Please describe the actual behaviour

PR to be completed, with basic documentation and tests.

GildedHonour:
@rossjones , how can I work on it, should I just start and then push a PR? What if some else will start working on it at the same, will only one of us get the bounty?

rossjones:
I think you need to go to BountySource ( https://www.bountysource.com/issues/33227756-provide-support-for-background-tasks ) and mark yourself as working on it ... will update the original message to warn people.

torfsen:
@GildedHonour Are you still working on this? BountySource [says so](https://www.bountysource.com/issues/33227756-provide-support-for-background-tasks) but it also says you expected to be done by end of May.

GildedHonour:
Sorry, no.

Sent from my ASUS

-------- Mensaje original --------
De:Florian Brucker <notifications@github.com>
Enviado:Mon, 11 Jul 2016 21:18:07 +0800
Para:ckan/ckan <ckan@noreply.github.com>
Cc:Alex Maslakov <me@gildedhonour.com>,Mention <mention@noreply.github.com>
Asunto:Re: [ckan/ckan] Provide support for background tasks (#2977)

torfsen:
I've started working on this, see #3165. Note that I'm not interested in the bounty.

rossjones:
@torfsen It'd be cool if you did take the bounty, you could always give it to charity, or attach it as a bounty elsewhere?

torfsen:
[...]

PR #3165, opened 2016-07-14
[#2977] Basic implementation of background jobs via python-rq

torfsen:
This PR is WIP for #2977, a new implementation of background jobs using python-rq.

(Note that I'm not interested in the bounty)

It is loosely based on  #2706 by @rossjones.

The implementation has a minimal feature set on purpose: I'm fine with adding more features but I wanted to avoid introducing [things that nobody needs](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). So if you feel that this PR is missing an important feature please speak up.

Likewise, documentation and tests will not be written until there is an agreement regarding the general architecture.

There is now the `ckan.lib.jobs` module which most importantly provides the `init_queue` and `enqueue` functions. The former is called during CKAN startup and establishes the connection to Redis. I wanted to make sure that we fail early when Redis is not available (and not later at some random time when a job is enqueued). However I'm not sure whether `ckan/websetup.py` is the right place for that -- feedback appreciated!

Once you have enqueued a job using `ckan.lib.jobs.enqueue` you can use the `job_*` API functions to manage the queue:

- `job_list`: List all enqueued jobs
- `job_show`: Show details for a certain job
- `job_cancel`: Cancel a certain enqueued job
- `job_clear`: Clear the whole queue

Similar functionality is provided by `paster jobs`, see its help text. In particular, `paster jobs worker` starts a worker that fetches jobs from the queue and executes them. If you want to play around you can use `paster jobs test` to enqueue test jobs (which you then can list, cancel, execute, etc.).


amercader:
You rock @torfsen :beers: 

Don't put anything on `websetup.py`, I don't even know what that file is used for. What about putting the queue init code in `environment.load_environment()`. That's what's first called when creating the web app and the CLI (Right now the queue is not initialised when starting up CKAN as websetup is not called).

Are you envisaging Redis being a hard requirement?

wardi:
I think we all agreed to accept Redis as a hard requirement when this work is merged

TkTech:
Yup. It will also allow us to use Flask-Cache/plain redis as a general cache in core CKAN instead of stuffing everything into Solr.

torfsen:
@amercader I've moved the connection test to `load_environment` so that a warning is shown upon server start when Redis is not available. In addition I've made the queue initialization lazy, see `jobs.get_queue`.

torfsen:
To Do:
- [ ] Tests for `ckan.lib.jobs`
- [ ] Tests for `paster jobs`
- [ ] Tests for API functions
- [ ] Update [from-source-installation instructions](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) to include Redis
- [ ] Update [documentation of background tasks](http://docs.ckan.org/en/latest/maintaining/background-tasks.html) to include instructions for using the new system and for transitioning from Celery.
- [ ] Add deprecation notices to Celery stuff (`ckan.lib.celery_app`, `paster celeryd`, `resource_status_show`)

torfsen:
@TkTech You mentioned that you want to use Redis for other parts of CKAN, too. How should we let the user configure the Redis connection?

Since we're going to use 3rd-party code that uses Redis we probably want to put each distinct usage (background jobs, Flask cache) into a separate Redis database to make sure that they don't interfere with each other. Hence I suggest that we only add a single configuration option, `redis_url` (similar to `solr_url`) which contains Redis' URL without specifying a database (e.g. `redis://localhost:6379`). Our different use cases can then pick a separate database each. Redis [offers 16 databases by default](https://raw.githubusercontent.com/antirez/redis/3.0/redis.conf) which should be plenty.

TkTech:
We need to support it all in a single database. People using hosted redis, or odd hosting setups may not be able to use `SELECT`. Allowing them to specify the connection URI is fine, since the database can be selected in that URI.

There's no problem allowing all our uses to live in one database so long as everything is prefixed.

torfsen:
@TkTech Good point regarding hosted Redis.

RQ [prefixes its keys](http://python-rq.org/contrib/), so interference should be avoidable. It also [seems to empty queues intelligently](https://github.com/nvie/rq/blob/766bb600060c85e8a39579541798930bbb508ec8/rq/queue.py#L82). However, it seems that the prefix is not configurable, so we cannot add the CKAN site ID. This means that two CKAN instances can share a Redis instance but must use separate databases.

We still have to be cautious, though: For example, ckanext-harvest [simply flushes the whole Redis database](https://github.com/ckan/ckanext-harvest/blob/7f506913f8e78988899af9c1dd518dc76e2c3e62/ckanext/harvest/queue.py#L101) to empty its queues. This is a trivial example, since we control that code, but we must be careful not to miss something like that in 3rd-party code.

TkTech:
https://github.com/nvie/rq/blob/766bb600060c85e8a39579541798930bbb508ec8/rq/queue.py#L31

Looks like we can "configure" the prefix this way.

We should definitely update the plugins we control (no good plugin should be wiping the entire keyspace! *Always* use prefixes in redis), but for 3rd party plugins we shouldn't consider that our problem. Document it and assume good behavior.

torfsen:
@TkTech Unfortunately there doesn't seem to be such a setting for the prefix of job keys, that [seems to be hard-coded](https://github.com/nvie/rq/blob/766bb600060c85e8a39579541798930bbb508ec8/rq/queue.py#L85).

torfsen:
It just occurred to me that we can also simply use the CKAN site ID as the queue name for RQ.

torfsen:
Regarding compatibility with ckanext-harvest: See ckan/ckanext-harvest#257 and the corresponding PR ckan/ckanext-harvest#258.

torfsen:
Do we need support for multiple queues? #2706 had, and ckanext-harvest also uses two separate queues (*gather* and *fetch*). This PR currently only supports a single queue but it would be easy to support an arbitrary number of named queues instead. Workers could then listen on all or a certain subset of the queues.

@rossjones and @amercader, what was the rationale behind the multiple queue-functionality in #2706 and ckanext-harvest, respectively?

rossjones:
I had multiple queues in that PR because on data.gov.uk we use different queues for different priority tasks. That way we can have a busy low-priority queue, but still get out the small tasks pretty quickly.  FWIW I'd support multiple queues if it wasn't much work, but I wouldn't let it hold you up.

torfsen:
Multiple queues are now supported. By default there is just a single queue (`default`) which is used whenever no other queue is specified. Queues can be created by simply adding tasks to them.

Queue names are automatically prefixed with the site ID so multiple CKAN instances can now share a Redis database.

torfsen:
I've updated the documentation. Previously we used the term "background task", but I've now used "job" instead of "task" both in code and in documentation. The reason is that I'd like to reserve "task" for a more high-level functionality that supports things like scheduling, failure tracking, etc. (see ckan/ideas-and-roadmap#180).

However, I haven't changed the URL of the [main background job documentation page](http://docs.ckan.org/en/latest/maintaining/background-tasks.html) so that links will still be valid. If we do introduce a high-level task system then it can be documented on the same page.

torfsen:
From my point of view this PR is ready for review.

I'll be on holiday until end of August and probably won't be able to work on this during that time. Feel free to fix things yourselves if you need it sooner (I'd love to see this in 2.6 but I'm not sure about the current release timeline).

requirements.txt
```diff
@@ -1,61 +1,64 @@
...
-WebError==0.13.1
-WebHelpers==1.3
-WebOb==1.0.8
-WebTest==1.4.3
-Werkzeug==0.11.10
+WebError==0.13.1          # via pylons
+WebHelpers==1.3           # via pylons
+WebOb==1.0.8              # via fanstatic, pylons, repoze.who, repoze.who-friendlyform, weberror, webtest
+WebTest==1.4.3            # via pylons
+Werkzeug==0.11.10         # via flask
```
TkTech: Not related to this PR, but there has to be a better way of tracking dependency sources? Any ideas?

test-core.ini
```diff
@@ -31,6 +31,9 @@ ckan.datapusher.formats = csv xls xlsx tsv application/csv application/vnd.ms-ex
 ## Solr support
 solr_url = http://127.0.0.1:8983/solr/ckan
 
+# Redis URL. Use a separate Redis database for testing.
+redis_url = redis://localhost:6379/1
```
TkTech: Do we want to namespace this like newer config options?

Ex: `ckan.db.redis_uri = `?

ckan/tests/helpers.py
```diff
@@ -468,3 +544,149 @@ def wrapper(*args, **kwargs):
...
+
+    def clear(self):
+        u'''
+        Clear all captured log messages.
+        '''
+        self.messages = collections.defaultdict(list)
+
+
+@contextlib.contextmanager
+def temp_file(*args, **kwargs):
```
TkTech: Is this adding something that `tempfile.NamedTemporaryFile` doesn't already provide? You can use it as a context manager as-is without needing to wrap it.

ckan/tests/helpers.py
```diff
@@ -468,3 +544,149 @@ def wrapper(*args, **kwargs):
...
+    manager_level = logger.manager.disable
+    disabled = logger.disabled
+    logger.addHandler(handler)
+    try:
+        if level is not None:
+            logger.setLevel(level)
+        if override_disabled:
+            logger.disabled = False
+        if override_global_level:
+            if (level is None) and (manager_level > old_level):
```
TkTech: Why the brackets?

ckan/tests/helpers.py
```diff
@@ -468,3 +544,149 @@ def wrapper(*args, **kwargs):
...
+                logger.manager.disable = level
+        yield handler
+    finally:
+        logger.handlers.remove(handler)
+        logger.setLevel(old_level)
+        logger.disabled = disabled
+        logger.manager.disable = manager_level
+
+
+class RecordingLogHandler(logging.Handler):
```
TkTech: This seems very useful for testing, but I can see it getting quite large quickly. What do you think of using a Redis key since we're using it already? You can easily implement a bounded queue with PUSH and TRIM.

TkTech:
Still lots to review, but not finding much to nitpick about so far, looking good!

test-core.ini
torfsen: I chose the name similar to `solr_url`, but namespacing it could make sense. However, I'd go with something like `ckan.redis.url` so that we can add more Redis-related settings later on if the need arises. I'd rather use URL than URI, although both would be correct here: All the other config options use URL, so let's keep that consistent.

ckan/tests/helpers.py
torfsen: For me they improve readability, that's all :wink:

test-core.ini
TkTech: Sounds good to me!

ckan/tests/helpers.py
torfsen: `RecordingLogHandler` is intended to be used with the `recorded_logs` context manager, so it's usually only active during a short period of time. In addition you usually attach it only to the logger that you're interested in. I'm therefore not sure whether the effort of storing the logs in Redis would be worth it.

Perhaps the documentation of `RecordingLogHandler` should be updated to reflect this, in my experience it's rather useless on its own -- see all the work that `recorded_logs` has to do in order to make sure that logs are actually recorded during tests.

ckan/tests/helpers.py
torfsen: I think you're right. Usually my main problem with `NamedTemporaryFile` is that it deletes the file when it's being closed instead of when the context manager is left (which would often be more useful for me). So I end up passing `delete=False` to `NamedTemporaryFile` and do the cleanup myself -- that's what `temp_file` does (I've copied it from other projects where it's been useful).

However, I just checked my usages of `temp_file` in this PR and it seems that I don't need that modified deletion behavior at all. So, yes, I think I can remove it.

requirements.txt
torfsen: These comments are automatically added by [pip-compile](https://github.com/nvie/pip-tools) based on `requirements.in`, so I don't think that's a problem.

TkTech: Neat, haven't seen it pip-compile before.

ckan/tests/helpers.py
torfsen: I've updated the docstring of `RecordingLogHandler` to suggest using `recorded_logs`.

ckan/tests/helpers.py
torfsen: Fixed.

test-core.ini
torfsen: Fixed.

torfsen:
I'm back from holiday, so if there's anything I can do to help move this forward let me know.

ckan/lib/cli.py
```diff
@@ -43,6 +46,16 @@ def deprecation_warning(message=None):
...
 
 
+def error(msg):
+    '''
+    Print an error message to STDOUT and exit with return code 1.
+    '''
+    sys.stderr.write(msg)
+    if not msg.endswith('\n'):
+        sys.stderr.write('\n')
+    sys.exit(1)
```
wardi: it's the little things :+1:

ckan/lib/cli.py
```diff
@@ -2551,15 +2553,161 @@ def command(self):
...
+            error(u'There is no job with ID "{}"'.format(id))
+        print(u'Cancelled job {}'.format(id))
+
+    def clear(self):
+        data_dict = {
+            u'queues': self.args,
+        }
+        queues = p.toolkit.get_action(u'job_clear')({}, data_dict)
+        queues = (u'"{}"'.format(q) for q in queues)
+        print(u'Cleared queue(s) {}'.format(u', '.join(queues)))
```
wardi: All of our CLIs should be thin wrappers around action calls. :heart: it.

ckan/lib/redis.py
```diff
@@ -0,0 +1,63 @@
...
+
+from pylons import config
+from redis import ConnectionPool, Redis
+
+
+log = logging.getLogger(__name__)
+
+REDIS_URL_SETTING_NAME = u'ckan.redis.url'
+
+REDIS_URL_DEFAULT_VALUE = u'redis://localhost:6379/0'
```
wardi: If we're requiring redis now is it safe to assume we can connect to a redis running on localhost? How about we fail loudly if the value isn't set instead?

ckan/tests/helpers.py
```diff
@@ -316,21 +371,41 @@ def test_ckan_site_title(self):
...
-                return_value = func(*args, **kwargs)
-            finally:
-                config.clear()
-                config.update(_original_config)
 
-            return return_value
-        return nose.tools.make_decorator(func)(wrapper)
-    return decorator
+@contextlib.contextmanager
+def changed_config(key, value):
```
wardi: This looks a lot like https://github.com/ckan/ckan/blob/a77ff9c021c80eee949d2ed11cbafd743ee5c04a/ckan/tests/helpers.py#L301-L333

doc/extensions/best-practices.rst
```diff
@@ -37,17 +37,61 @@ Implement each plugin class in a separate Python module
...
+* The names of *API action functions* introduced by your extension should begin
+  with the name of your extension. For example
+  ``my_extension_foobarize_everything``.
+
+* The names of *background job queues* introduced by your extension should
+  begin with the name of your extension. For example
+  ``my_extension:super-special-job-queue``.
+
+In some situations, resources like databases may even be shared between
+multiple CKAN *instances*, which requires an even higher degree of uniqueness
```
wardi: Pretty sure you can't share a CKAN database, though solr might work with a different site id. In any case it doesn't seem like a great idea to me.

wardi:
One last change I'd suggest is adding `enqueue` to to `ckan.plugins.toolkit` (possbly as `enqueue_job` if you think it's name isn't clear enough) because it's an interface we want to maintain for extension authors.

Great work @torfsen!

ckan/tests/helpers.py
torfsen: It does indeed :wink:

`changed_config` is a context-manager-variant of `change_config` (which is a decorator). I added it because I needed to temporarily change the config [in the middle of a test case](https://github.com/torfsen/ckan/blob/2977-rq-background-tasks/ckan/tests/lib/test_jobs.py#L76). That's cumbersome to do with a decorator, so I added a context manager. Note that both variants mention each other in their docstrings and that `change_config` (the decorator) now uses `changed_config` (the context manager) internally, so there's no duplicate code.

I think it's a useful tool, but if you prefer I can drop it (at the cost of making that one test case a bit uglier). In hindsight, the better approach would be to add context manager-capability to the existing `change_config` -- starting from Python 3.2 one can use [`contextlib.ContextDecorator`](https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator) for that, on Python 2 we'd have to do it manually (or backport `ContextDecorator`).

So our options are (a) to keep it, (b) to drop it, or (c) to upgrade `change_config`.

wardi: Right, I missed that.:smile:

I'd say leave it as is for now, the naming is a bit awkward but I can't think of something better.

ckan/lib/redis.py
torfsen: [We've previously settled on a warning during startup](https://github.com/ckan/ckan/pull/3165#issuecomment-232898542) in case Redis is not available, but I'm open for other solutions, including a hard fail and no default value. As far as I know, the only reason we're currently failing hard for during startup is if PostgreSQL isn't available. I've seen the Redis dependency similar to Solr, i.e. required for production but not essential. Hence a warning, but no error (just like for Solr).

doc/extensions/best-practices.rst
torfsen: The choice of the word "database" was motivated by the discussion in this PR during which it became clear that sharing a *Redis* database between instances could be useful. But you're probably right that most people will think of the *PostgreSQL* database first when they read "database". I've improved the wording to make things clearer, let me know if that helps.

ckan/lib/redis.py
wardi: We've added a hard failure for when site_url isn't configured now instead of having a site up but sometimes generating broken links. Most of the logic functions that return datasets now depend on package_search (and solr) working so I'd argue we should fail in that case too.

This is a bit different. As written the code assumes a local redis install and starts using it. That comes with a (small) chance of corrupting data from another program. Not a nice thing to do. My vote is for hard failure, what do the rest of you think?

torfsen:
Thanks for the review and good idea regarding `ckan.plugins.toolkit.enqueue_job`, @wardi, that's done.

I've also restored `ckan/config/celery-supervisor.conf`: I had previously removed it while deprecating the old Celery system. However, the old documentation told people to copy *or link* the file, so removing it could break existing installations. I've now restored the file, we should keep it around until we remove Celery support completely.

doc/extensions/best-practices.rst
```diff
@@ -37,17 +39,61 @@ Implement each plugin class in a separate Python module
...
+------------------
+Avoid name clashes
+------------------
+Many of the names you pick for your identifiers and files must be unique in
+relation to the names used by core CKAN and other extensions. To avoid
+conflicts you should prefix any public name that your extension introduces with
+the name of your extension. For example:
+
+* The names of *configuration settings* introduced by your extension should
+  have the form ``ckan.my_extension.my_config_setting``.
```
wardi: The two patterns I've seen for configuration options are
```
ckanext.my_extension.my_config_setting = ...
```

and

```
my_extension.my_config_setting = ...
```

I choose the latter for my extensions because I find the configuration options are really long with the former.
Anything starting with `ckan.` should belong to ckan core.

torfsen: I've changed it to `my_extension.my_config_setting`.

ckan/lib/jobs.py
```diff
@@ -0,0 +1,251 @@
...
+
+Internally, RQ queue names are prefixed with a string containing the
+CKAN site ID to avoid key collisions when the same Redis database is
+used for multiple CKAN instances. The functions of this module expect
+unprefixed queue names (e.g. ``'default'``) unless noted otherwise. The
+raw RQ objects (e.g. a queue returned by ``get_queue``) use the full,
+prefixed names. Use the functions ``add_queue_name_prefix`` and
+``remove_queue_name_prefix`` to manage queue name prefixes.
+
+.. versionadded:: 2.6
```
wardi: please update these to 2.7

ckan/lib/jobs.py
```diff
@@ -0,0 +1,251 @@
...
+'''
+
+import logging
+
+from pylons import config
```
amercader: Since #3163 was merged we should never import `config` directly from Pylons anymore. Use `from ckan.common import config`.

My fault for not documenting / announcing it

ckan/lib/redis.py
amercader: > Internally, RQ queue names are prefixed with a string containing the CKAN site ID to avoid key collisions when the name Redis database is used for multiple CKAN instances.

So if understand correctly we only might run into problems if we have two CKAN instances sharing a Redis database without having changed their site ids. I think that's an unlikely scenario (if you have multiple CKANs sharing resources with the same site id you are going to get into trouble anyway, starting with Solr) and that a hard failure will add an extra step to deployment for people that want to get up and running fast, so I'd be happy with the current default option.

doc/extensions/best-practices.rst
```diff
@@ -37,17 +39,61 @@ Implement each plugin class in a separate Python module
...
+*instances*, which requires an even higher degree of uniqueness for the
+corresponding names. In that case, you should also prefix your identifiers with
+the CKAN site ID, which is available via
+
+::
 
------------------------------------------------------------------
-Names of config settings should include the name of the extension
------------------------------------------------------------------
+    from pylons import config
```
amercader: See comment about `from ckan.common import config`. As this is only available on CKAN>=2.7 perhaps is best to show the import or show two examples, one for CKAN<2.7 and one for CKAN>=2.7

doc/maintaining/background-tasks.rst
```diff
@@ -1,175 +1,356 @@
...
 * Performing processing or validation or on data (as done by the Archiver and
   DataStorer Extensions)
 
+Basically, any piece of work that takes too long to perform while the main
+application is waiting is a good candidate for a background job.
+
+.. note::
+
+    The current background job system is based on RQ_ and was introduced in
+    CKAN 2.6. See :ref:`background jobs migration` for details on how to
```
amercader: 2.7

amercader:
@torfsen This looks really, really good. Thanks for being so thorough with the tests and docs. Apart from the minor comments added it looks good to me.

Can't wait to try it with ckanext-harvest!

ckan/lib/redis.py
wardi: Sounds good to me, leave it as is

ckan/lib/jobs.py
torfsen: Fixed.

ckan/lib/jobs.py
torfsen: No problem, fixed.

doc/extensions/best-practices.rst
torfsen: I've changed the code so that it works on >=2.7 and <2.7.

doc/maintaining/background-tasks.rst
torfsen: Fixed.

wardi:
any objections to merging this now?

amercader:
No, let's get it in!

torfsen:
Yay, great work everybody!

wardi:
@torfsen I wrote a tiny stub of an extension that uses this code https://github.com/wardi/ckanext-geocodejob but it seems rq's forking is interfering with ckan's sqlalchemy state. My first job works but jobs after that fail with `DatabaseError: (DatabaseError) SSL error: decryption failed or bad record mac`

Can we wrap the job running with code that cleanly closes our database connections?

torfsen:
@wardi Good catch! I've created a separate issue (#3243) and I'm working on it.

PR #3169, opened 2016-07-18
Update helpers.py

gsueur:
Fixes #
### Features:
- [ ] includes API changes

Quick fix on a nasty bug (which only really had effect after I pip install --upgrade Babel actually, I don't really understand why). 
As Python considers that boolean values are instances of Int (because before boolean types were introduced they were actuel integers) a test should be done on the specific boolean type BEFORE it is tested against integer type, to avoid sending a boolean value like 'True' to a decimal formatter. I don't know which formatter to use instead, so I just "continue" here, without using any localized representation of boolean values.

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1880,6 +1880,8 @@ def format_resource_items(items):
                 value = formatters.localised_number(float(value))
             elif re.search(reg_ex_int, value):
                 value = formatters.localised_number(int(value))
+        elif isinstance(value, bool):
+            continue
         elif ((isinstance(value, int) or isinstance(value, float))
                 and value not in (True, False)):
             value = formatters.localised_number(value)


wardi:
This looks fine, but would you add a test so this doesn't get broken again in the future.

What is the exact traceback you see without this fix?

gsueur:
Sorry but I don't know CKAN code enough to be able to do that. I guess that this part of the code formats any resource attribute in the users locale (for date, decimal separator...). But if any attribute's value is boolean, it will cause an error. here is the backtrace I got in my apache's log :
` {% for key, value in h.format_resource_items(res.items()) %}
File '/usr/lib/ckan/default/src/ckan/ckan/lib/helpers.py', line 1767 in format_resource_items
value = formatters.localised_number(value)
File '/usr/lib/ckan/default/src/ckan/ckan/lib/formatters.py', line 156 in localised_number
  return numbers.format_number(number, locale=i18n.get_lang())
File '/usr/lib/ckan/default/lib/python2.7/site-packages/babel/numbers.py', line 223 in format_number
  return format_decimal(number, locale=locale)
File '/usr/lib/ckan/default/lib/python2.7/site-packages/babel/numbers.py', line 254 in format_decimal
  return pattern.apply(number, locale)
File '/usr/lib/ckan/default/lib/python2.7/site-packages/babel/numbers.py', line 570 in apply
  value = Decimal(str(value))
File '/usr/lib/python2.7/decimal.py', line 547 in __new__
  "Invalid literal for Decimal: %r" % value)
File '/usr/lib/python2.7/decimal.py', line 3873 in _raise_error
  raise error(explanation)
  InvalidOperation: Invalid literal for Decimal: 'True'`

TkTech:
@gsueur you can use 3 ```s to start and end a code block, I've gone ahead and reformatted your traceback for readability.

wardi:
It seems there was already a check for boolean values two lines below: `value not in (True, False)`, are you sure the latest version of CKAN has the issue you're reporting?

rossjones:
I tried this in 2.8

```
>>> import ckan.lib.helpers as h
>>> items = [('key', True)]
>>> h.format_resource_items(items)
[('key', True)]
```
Looks fine, is this safe to close? must have been a much older version.

kmbn:

PR #3171, opened 2016-07-21
Docs, spelling, pep8

davidread:
Added docs to explain flatten_schema().

No real code changes - just renamed variables.


torfsen:
While you're at it, could you please fix the remaining PEP8 issues and remove the file from the blacklist in `ckan/tests/legacy/test_coding_standards.py`?

davidread:
Ok good plan, I did some more PEP8.

I also spotted validate_flattened could be moved to the tests as it's only used by them (and I checked no-one else on github was using it apart from in forks of the test).

PR #3173, opened 2016-07-25
Fix - better exception for when bad smtp settings.

davidread:
Improve error message when smtp server is not configured correctly. Before this, they just see an uncaught socket exception merely saying: "[Errno 111] Connection refused" when an email tries to be sent.

This was seen on by a user of the issues extension: https://github.com/ckan/ckanext-issues/issues/63#issuecomment-234525736

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -1,6 +1,7 @@
 # encoding: utf-8
 
 import smtplib
+import socket
 import logging
 import uuid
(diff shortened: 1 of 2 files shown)


TkTech:
Sigh, this is one of the things that absolutely has to be moved into the background workers once they're available.

amercader:
Merged as I wanted this in 2.6

PR #3181, opened 2016-08-01
CKAN 2.5 tests and source install fail due to unpinned html5lib

wardi:
html5lib in 2.5 was not pinned in requirements.txt for some reason. This backports the version pinned on master at the moment


metaodi:
:+1: for merging this

amercader:
I updated bleach rather than html5lib on 2.5.3 as suggested on #3214

PR #3182, opened 2016-08-02
Optimize css improve vn lang

nhatdx:
- Update VN language

- Optimize fuchsia.css file (duplicate at class .page-header)

### Features:

- [ ] includes API changes


TkTech:
Any chance we could find a second to double-check the translations in this PR?

wardi:
Translations are done through transifex then imported as part of the release process, we don't usually take them as pull requests.

The css files are generated from less files, so making changes to them directly is also not the right approach. I think we're getting rid of the colour themes too as part of your scss work, right @TkTech ?

amercader:
@nhatdx please register to Transifex and request access to the Vietnamese team. Then you can do the changes there so they don't get lost on the next release (you can upload the updated ckan.po file, you don't need to reenter them manually). Thanks!

And yes, the changes in the css files should be done on the relevant less files (although I suspect that these particular ones would better off in a custom css file on your extension)

PR #3174, opened 2016-07-26
replace MapQuest by Stamen terrain

nibecker:
Fixing #3162 
Stamen terrain has been chosen due to its open CC license and clean look

PR #3184, opened 2016-08-03
Update Recline version

amercader:
The one shipped in core is quite old, and upgrading will probably help with #3174


amercader:
@davidread perhaps you can have a quick look at this one? This should go to 2.6 so would be good to merge it before branching.

davidread:
Seems to work 👍

Issue #3176, opened 2016-07-28
A user can't list all private datasets via the API


davidread:
### CKAN Version if known (or site URL)
Master

### Please describe the expected behaviour
A user should be able to get a list of all the datasets available to him/her - including private ones via the API.

### Please describe the actual behaviour
The only way I can see to get private ones is by using user_show include_datasets=True. However that only includes private ones created by that user. It excludes other private ones created in the organization by other users.

### What steps can be taken to reproduce the issue? 
I was just answering this question on SO by reading the code. http://stackoverflow.com/questions/38625786/how-to-obtain-a-given-users-dateset-listboth-public-and-private-through-ckan/38629242#38629242

LinJianping:
In my opinion, the organization administrator should have privileges to decide whether the members can access all the datasets or those created by themselves.

wardi:
package_search should return private datasets when you specify an organzation you're a member of like the organization pages do, except that our org controller is using a context parameter to make that work. We should fix that. https://github.com/ckan/ckan/blob/master/ckan/controllers/group.py#L297-L303

wardi:
fixed by #3191

PR #3191, opened 2016-08-05
package_search include_private flag

wardi:
Proposed fix for #3176 

Stop using a context variable in package_show to return private datasets (only works from inside ckan), instead allow users to pass `include_private=True` to have all datasets they have permission to see returned.

This is essentially what is already allowed when users browse to the organzation search page *unless* there's a customised package_show auth function. In that case we might be exposing more information than was available earlier (users can get the whole dataset metadata not just what's seen in search results). We'll have to change the way permissions on datasets are defined to fix that issue, see: #3192 for next steps.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1826,23 +1826,30 @@ def package_search(context, data_dict):
         # we should remove any mention of capacity from the fq and
         # instead set it to only retrieve public datasets
         fq = data_dict.get('fq', '')
+
+        # Remove before these hit solr FIXME: whitelist instead
         include_private = asbool(data_dict.pop('include_private', False))
-        if not include_private or not user:
-            fq = ' '.join(p for p in fq.split() if 'capacity:' not in p)
-            data_dict['fq'] = fq + ' capacity:"public"'
-        elif not authz.is_sysadmin(user):
-            fq = ' '.join(p for p in fq.split() if 'capacity:' not in p)
+        include_drafts = asbool(data_dict.pop('include_drafts', False))
+
+        capacity_fq = 'capacity:"public"'
+        if include_private and authz.is_sysadmin(user):
+            capacity_fq = None
+        elif include_private and user:
             orgs = logic.get_action('organization_list_for_user')(
                 {'user': user}, {'permission': 'member'})
             if orgs:
-                data_dict['fq'] = (fq + ' (capactiy:"public"'
-                    ' OR owner_org:({0}))'.format(' OR '.join(
-                        org['id'] for org in orgs)))
-            else:
-                data_dict['fq'] = fq + ' capacity:"public"'
+                capacity_fq = '({0} OR owner_org:({1}))'.format(
+                    capacity_fq,
+                    ' OR '.join(org['id'] for org in orgs))
+            if include_drafts:
+                capacity_fq = '({0} OR creator_user_id:({1}))'.format(
+                    capacity_fq,
+                    authz.get_user_id_for_username(user))
 
-        # Solr doesn't need 'include_drafts`, so pop it.
-        include_drafts = asbool(data_dict.pop('include_drafts', False))
+        if capacity_fq:
+            fq = ' '.join(p for p in fq.split() if 'capacity:' not in p)
+            data_dict['fq'] = fq + ' ' + capacity_fq
+
         fq = data_dict.get('fq', '')
         if include_drafts:
             user_id = authz.get_user_id_for_username(user, allow_none=True)
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1711,6 +1711,10 @@ def package_search(context, data_dict):
         sysadmin will be returned all draft datasets. Optional, the default is
         ``False``.
     :type include_drafts: boolean
+    :param include_private: if ``True``, private datasets will be included in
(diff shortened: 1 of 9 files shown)


wardi:
This needs some more tests covering include_private as a normal parameter. We could also add a config option to default to showing private datasets in the normal search page here too.

wardi:
@TkTech this is sure to conflict with #3118. It feels like lots of solr-specific code in this action should be moved into lib/search

TkTech:
@wardi there are a number of actions and methods relying on Solr syntax queries that I'm going to have to work through one-by-one. The changes proposed and upcoming in #3118 will take months to finish, so we shouldn't let that be a blocker or concern for this PR right now.

wardi:
Spoke to @amercader about this on Tuesday and we decided to enable display of private datasets on the default search page as part of this PR. There will be a new option `ckan.search.default_include_private = False` to revert to the old CKAN behaviour if you really want to.

Aaron-M:
Question - would having the search results set to (true) include_private datasets affect if private datasets are indexed by google... (currently they are not)?

wardi:
anonymous users (like google) shouldn't be able to see private datasets. If you can see private datasets without logging in with these changes please let me know!

ckan/tests/controllers/test_package.py
```diff
@@ -418,12 +418,6 @@ def test_unauthed_user_creating_dataset(self):
 
 
 class TestPackageEdit(helpers.FunctionalTestBase):
-    @classmethod
-    def setup_class(cls):
-        super(cls, cls).setup_class()
-        helpers.reset_db()
-        search.clear_all()
-
```
amercader: Why getting rid of these class setup methods?

doc/maintaining/configuration.rst
```diff
@@ -655,6 +655,21 @@ Default value:  ``false``
...
+
+Example::
+
+ ckan.search.defalt_include_private = false
+
+Default value:  ``true``
+
+Controls whether the default search page (``/dataset``) should include
+private datasets visible to the current user or only public datasets
+visible to everyone.
```
amercader: Shall we add a note here about the behaviour on the API not being the same (ie you don't get private datasets by default and having to pass `include_private`)? I can see this being confusing

ckan/logic/action/get.py
```diff
@@ -1823,16 +1823,33 @@ def package_search(context, data_dict):
...
-        # If this query hasn't come from a controller that has set this flag
-        # then we should remove any mention of capacity from the fq and
+        # we should remove any mention of capacity from the fq and
         # instead set it to only retrieve public datasets
         fq = data_dict.get('fq', '')
-        if not context.get('ignore_capacity_check', False):
+
+        # Remove before these hit solr FIXME: whitelist instead
+        include_private = asbool(data_dict.pop('include_private', False))
+        include_drafts = asbool(data_dict.pop('include_drafts', False))
```
amercader: We need to document these options on the docstring

amercader:
@wardi looks great, some minor comments

nibecker:
Thanks @wardi. We should make sure that users with customised package_show will be informed about this change in advance as it might cause trouble for them. This includes of course careful documentation in the release notes. Any other suggestion where to place a notice?

wardi:
@nibecker yes, we'll need to include the new parameters to package_show in the changelog, however If you've customised package_show you should know you're going to have trouble with upgrades. Overriding a core action is pretty much the same as monkey-patching CKAN code.

doc/maintaining/configuration.rst
wardi: The behaviour of the API stays the same unless you're relying on `context['ignore_capacity_check'] = True` somewhere in your own code. I'd suggest replacing that with `context['user'] = site_user; data_dict['include_private'] = True`

ckan/tests/controllers/test_package.py
wardi: They're duplicating code already in the base class

doc/maintaining/configuration.rst
amercader: I meant that the behaviour is not the same in the frontend than in the API. If I understand correctly by default `package_search` won't show private datasets, as opposed to the frontend.

wardi: Could this be written more clearly? I say "the default search page (``/dataset``)", nothing about the API. Is that confusing?

amercader: Fair enough, let's leave it like this

amercader:
@wardi is that ready to go or did you want to add more tests?

wardi:
@amercader I was just looking at that. Are there cases that aren't covered by the current tests?

amercader:
Maybe "sysadmin can search private datasets" and "user from org1 can not see private datasets from org2"? I know auth testing of the underlying actions should cover that but just in case :)

wardi:
@amercader sure, but I'd like to do those tests at the action level because there's no custom code for either of those cases in the controller

amercader:
But this should be just a matter of setting the appropriate `REMOTE_USER` (a sysadmin and an org1 user) and check the output as you do in the other tests, right?

wardi:
@amercader ok, added those too

amercader:
PEEEEEEEEEEP 8 :8ball: :skull:

amercader:
@wardi fixed that myself, merged now

PR #3194, opened 2016-08-08
Fix redirects in tests to use the host on `ckan.site_url`

amercader:
Whenever we are testing a redirect, or a Location header in the tests we are happy to test against an expected location that uses `localhost`, eg (from [here](https://github.com/ckan/ckan/blob/60134e844f3a68083c5aaf956872823745a3b02a/ckan/tests/controllers/test_util.py#L20:L21)):

```
        assert_equal(response.headers.get('Location'),
                     'http://localhost/dataset')
```

When running the tests we are using `http://test.ckan.net` as `ckan.site_url` so that's the value that one would expect to be used in a redirect.

The underlying problem is letting WebOb know the host to use when raising a redirect exception.
This PR fixes this in a couple of different places, ensuring that we redirect to the full URL using the value in `ckan.site_url` and setting the `HTTP_HOST` key in the WSGI environ when redirecting in the tests.


amercader:
BTW all this is relevant to the new URL generation logic with Pylons/Flask support

ckan/tests/helpers.py
```diff
@@ -213,6 +214,12 @@ def submit_and_follow(app, form, extra_environ=None, name=None,
     Call webtest_submit with name/value passed expecting a redirect
     and return the response from following that redirect.
     '''
+    if extra_environ is None:
+        extra_environ = {}
+    if not extra_environ.get('HTTP_HOST'):
+        extra_environ['HTTP_HOST'] = str(
+            urlparse.urlparse(config['ckan.site_url']).netloc)
+
```
wardi: @amercader When I remove this change the tests are still passing for me, what am I missing?

amercader:
@wardi thanks for merging this. I think the HTTP_HOST stuff was needed for the new Flask/Pylons URL generation to work but I'll add it to the relevant PR #3228 if necessary

Issue #3196, opened 2016-08-08
Foundation work for common Flask / Pylons objects and functions


amercader:
Similarly to the work started with #3163 with the `config`, there is a whole lot of "foundation" work that needs to be done to allow requests to be served both under Flask and under Pylons. These work is necessary before starting the actual migration of the controllers,

There's a bunch of Flask PRs coming over and it's getting more difficult to isolate single features or changes, so I'll keep a map here of what should be reviewed / merged first. Every individual PR also says which ones it depends on as well. I'll update this as we hopefully make progress.

There will surely be more stuff that will need to be added to this common logic, but that can be done when we start to migrate the controllers as needed.

```
#3194 (redirects) +----------------------------------------------------------------+
                                                                                   |
                                                                                   v
#3197 (requests)  +-----> #3203 (g/c) +--------> #3207 (blueprints) +---------->  #3212 (common requests code) +---------> #3213 (i18n) +---------> #3228 (url_for) +-------> API Blueprint (#3239)
                                                            ^                      ^
                                                            |                      |
#3206 (debug) +---------------------------------------------+                      +
                                                                                  #3208 (sessions)


```
- ~~[#3194](https://github.com/ckan/ckan/pull/3194) - Redirects~~
- ~~[#3206](https://github.com/ckan/ckan/pull/3206) - Debug and testing settings for Flask~~
- ~~[#3197](https://github.com/ckan/ckan/pull/3197) - Common request object~~
- ~~[#3203](https://github.com/ckan/ckan/pull/3203) - Common global object (g/c)~~
- ~~[#3207](https://github.com/ckan/ckan/pull/3207) - Blueprints registration~~
- ~~[#3212](https://github.com/ckan/ckan/pull/3212) - Common logic before and after requests~~
-  ~~[#3208](https://github.com/ckan/ckan/pull/3208) - Common session object~~
- ~~[#3213](https://github.com/ckan/ckan/pull/3213) - I18n~~
- ~~[#3228](https://github.com/ckan/ckan/pull/3228) - url_for~~
- ~~[#3239](https://github.com/ckan/ckan/pull/3239) - API Blueprint~~

amercader:
There's a bunch of Flask PRs coming over and it's getting more difficult to isolate single features or changes, so I'll keep a map here of what should be reviewed / merged first. Every individual PR also says which ones it depends on as well. I'll update this as we hopefully make progress.

```
#3194 (redirects)


[...]

PR #3197, opened 2016-08-08
[#3196] Common request object for Flask and Pylons

amercader:
A new CKAN Request object based on Werkzeug's LocalProxy that will forward to Flask or Pylons own request objects depending on the output of `_get_request` (which essentially calls `is_flask_request`), and at the same time provide all objects methods to be able to interact with them transparently.

We don't use LocalProxy directly so we can handle the special case of accessing query string parameters via `request.params` on Flask (which uses `request.args`). All new Flask-only code should use `args` but for backwards compatibility we will support `params` for a while.


ckan/common.py
```diff
@@ -25,6 +25,25 @@
...
+        pylons_request_available = True
+    except TypeError:
+        pylons_request_available = False
+
+    if (flask.request and
+            (flask.request.environ.get(u'ckan.app') == u'flask_app' or
+             not pylons_request_available)):
+        return True
+    else:
+        return False
```
wardi: you can `return (flask.req....` here, no need for an if/else :smile:

ckan/common.py
```diff
@@ -91,10 +110,46 @@ def __delitem__(self, key):
...
+        All new code meant to be run just in Flask (eg views) should always
+        use request.args
+        '''
+        try:
+            return super(CKANRequest, self).__getattr__(name)
+        except AttributeError:
+            if name == u'params':
+                return super(CKANRequest, self).__getattr__(u'args')
+            else:
+                raise
```
wardi: would this work?
```python
    @property
    def params(self):
        try:
            return super(CKANRequest, self).params
        except AttributeError:
            return self.args
```
Maybe just code golf, but it seems less broad to me.

amercader:
Please don't merge this until 2.6 is branched

amercader:
Merged as part of #3212

PR #3199, opened 2016-08-09
Bulk operations don't update the package_revision table

amercader:
`bulk_update_private`, `bulk_update_public` and ``bulk_update_delete` change the `package` table and the search index but not the revision table, which causes the changes to not be reflected everywhere.

This fixes the model query and adds a bunch of tests.

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -1093,6 +1093,7 @@ def _bulk_update_dataset(context, data_dict, update_dict):
     model.Session.query(model.package_revision_table) \
         .filter(model.PackageRevision.id.in_(datasets)) \
         .filter(model.PackageRevision.owner_org == org_id) \
+        .filter(model.PackageRevision.current is True) \
         .update(update_dict, synchronize_session=False)
 
     model.Session.commit()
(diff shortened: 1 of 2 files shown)


ckan/logic/action/update.py
```diff
@@ -1084,16 +1084,15 @@ def _bulk_update_dataset(context, data_dict, update_dict):
...
         .update(update_dict, synchronize_session=False)
 
     # revisions
-    model.Session.query(model.package_table) \
-        .filter(model.Package.id.in_(datasets)) \
-        .filter(model.Package.owner_org == org_id) \
+    model.Session.query(model.package_revision_table) \
+        .filter(model.PackageRevision.id.in_(datasets)) \
+        .filter(model.PackageRevision.owner_org == org_id) \
         .update(update_dict, synchronize_session=False)
```
wardi: pretty sure we need to restrict to the latest revision here, don't we?

This code was changed in https://github.com/ckan/ckan/commit/0413e259ce2c80925c5aadca0896f264122567e4#diff-4b95e302501a681a72c099ecc904de3dL1293 there should be a `current=True` in this query

PR #3200, opened 2016-08-10
Issue 3162 recline view

### LINKED ISSUE #3162: DataStore Map and Explorer not displaying map tiles since 11 July 2016
Direct tile access to MapQuest maps has been discontinued as of 11 July 2016 and the DataStore Map and Explorer previews no longer display map tiles. 

The issue actually lies with recline.js and it has been logged https://github.com/okfn/recline/issues/500  and there is a referenced patch to replace MapQuest with Open Street Map frodrigo/recline@3df0c2a2bb8897124bdbbca715b2be1fd99cb08f

Thought it would be useful to have a record here and request the packaged recline.js should be updated.

jrods:
Fixes #3162 

### Proposed fixes:
This provides a way to swap the map tiling via values in your config file. 
It's the same configuration as ckanext-spatial, except that the config properties are prefixed with ckanext.reclineview.mapview .

Example:
```
ckanext.reclineview.mapview.type = 
ckanext.reclineview.mapview.mapbox.map_id = <your id>
ckanext.reclineview.mapview.mapbox.access_token = <your token>
ckanext.reclineview.mapview.attribution = <copyright, link, etc. >
```

### Features:

- [ ] includes API changes


amercader:
Hi @jrods, many thanks for this. This is essentially the last thing that was missing from the list I made on https://github.com/ckan/ckan/pull/3174#issuecomment-237216080.

I think we can improve a couple of things:

1. I've merged changes in Recline (okfn/recline#502) to allow customization of the tiles without having to extend `recline.View.Map`. We just need to pass the relevant options when initialising the view on `recline_view.js`:

    ```js
    view = new recline.View.Map({
        model: dataset,
        mapTilesURL: '//{s}.tiles.mapbox.com/v4/mapbox.mapbox-streets-v7/{z}/{x}/{y}.png?access_token=pk.XXXX',
        mapTilesAttribution: '&copy; MapBox etc..',
        mapTilesSubdomains: 'ab'
   })
    ```
   Just move the logic [here](https://github.com/ckan/ckan/pull/3200/files#diff-1b3440d60f9e497cc2e38c42029c8083R39) to `recline_view.js` and pass the relevant url, attribution and domains functions when creating the view (so no need for `mapview_ext.js`).

   This should be available in master as soon as #3184 is merged (it is now ready to go)

2. It would be great if users could control the behaviour of all map widgets with a single set of config options. Right now the widgets on ckanext-spatial and ckanext-geoview use `ckanext.spatial.common_map.*` so I think it makes sense to use those as well for the recline map

3. If you can mention these options somewhere on the [docs](http://docs.ckan.org/en/latest/maintaining/data-viewer.html) so people is aware of them that would be awesome.

jrods:
@amercader I've committed the suggestions you made in your comment to the PR

kfishwick:
Would it be possible to have this change backported to version 2.5.x?  That is the release we are using for bcgov.

amercader:
@jrods Great stuff, many thanks for the patch!

@kfishwick as this requires a whole upgrade of recline I'm afraid it will not make it to the 2.5.x line. We will replace the failing tiles directly with Stamen ones as in https://github.com/ckan/ckan/pull/3174.
If you don't want to patch your local fork of CKAN I suggest you follow your original approach and override the relevant recline view methods with ones in a script loaded from your own extension.

PR #3203, opened 2016-08-12
[#3196] Common global context object for Flask and Pylons

amercader:
This relates to `c` and `g`, so bear with me... It's based off #3197 so perhaps better to review and merge that one first.

For request-wide, global variables access, we are currently using:
- Pylons `c` (`tmpl_context`) - Used all over the place, both in code and templates
- CKAN's `app_globals` - CKAN's own logic related to the database table and actions for config options. This is linked to Pylons `g`. These are used mostly in templates with some minor [other places](https://gist.github.com/amercader/9447df4be83cd90f7e5d8561f1e9876f)

The plan moving forward would be: 

A single common `g` object (aliased to `c` for backwards compatibility), which proxies to Flask `g` or Pylons `c` under the hood (again using Werkzeug's `LocalProxy`).

To achieve this we need to ensure that:
- New Flask code only uses Flask's `g`, which has a fallback to CKAN's `app_globals`.
- Current Pylons code only uses `c` (ie Remove all usages of `g` in Pylons non-template code)

This is implemented in this PR (and also some removal of old `c` related functions).

Questions:
1. How far do we go replacing `c` occurrences with `g`? With these changes using `g` or `c` would have the same effect but we obviously want to avoid confusion so I'd suggest to bite the bullet and start changing as many occurrences as we can, certainly when migrating a controller from Pylons to Flask, both on code and templates. I reiterate that extensions using `c` in their templates will still work, even after migrating the controller. 
   
   Ideally we want to minimize the stuff on `g` and pass the variables explicitly to the templates but changing this will break the templates, so for a few versions we still need to set the current `c.x` vars, even if the actual code is refactored not to use them, eg changing this:
   
   ``` python
   
         c.search_facets = query['search_facets']
         c.groups = group_names
   
         render('package/search.html')
   
   ```
   
   to this:
   
   ``` python
   
       # TODO: remove
       c.search_facets = query['search_facets']
       c.groups = group_names
   
       render('package/search.html', extra_vars={
           search_facets: query['search_facets'], groups: group_names})
   ```
2. There is an important change of behaviour between Flask's `g` and Pylons `c`. _In non-template code_ Pylons will return a empty string when accessing a missing key. Flask's `g` will raise an `AttributeError`, eg:

``` python

   # Pylons
   x = c.not_here   # x is ''

   # Flask
   x = g.not_here   # raises AttributeError


```

   Returning an empty string is a trivial change on our implementation, but IMO we should keep the Flask behaviour (of course documenting it on the CHANGELOG). This will bring the `g` behaviour more in line with the default Flask one and will force more explicit code (eg you can do `g.get('not_here')`). This requires a couple of changes on the current code but they are easy to spot.

Any thoughts on 1 and 2?


amercader:
Merged as part of #3212

PR #3205, opened 2016-08-17
Storing the size and mimetypes from uploads

### LINKED ISSUE #1517: Empty mime type and size for uploaded files
Are these supposed to be filled?

jrods:
Fixes #1517 

### Proposed fixes:
I've added values for size and mimetype for resource metadata.

With mimetype I've added some redundancy to determining the resource file's mimetype.
One thing to note, I've added a dependency, `python-magic`. I'd figure this be ok because Datapusher installs `messytables` which requires `python-magic`.

### Features:

- [x] includes API changes

size is stored as bytes

there's a bit of redundancy with getting the mimetypes of files, by
default, mimetype is determined by the url, fallback is done by using
`python-magic` module

### DIFF
--- a/ckan/lib/dictization/__init__.py
+++ b/ckan/lib/dictization/__init__.py
@@ -45,6 +45,8 @@ def table_dictize(obj, context, **kw):
             result_dict[name] = value
         elif isinstance(value, int):
             result_dict[name] = value
+        elif isinstance(value, long):
+            result_dict[name] = value
         elif isinstance(value, datetime.datetime):
             result_dict[name] = value.isoformat()
         elif isinstance(value, list):
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -3,6 +3,7 @@
 '''API functions for adding data to CKAN.'''
 
 import logging
+import mimetypes
 import random
 import re
 from socket import error as socket_error
@@ -295,6 +296,13 @@ def resource_create(context, data_dict):
         pkg_dict['resources'] = []
 
     upload = uploader.get_resource_uploader(data_dict)
+    ## Get out resource_id resource from model as it will not appear in
+    ## package_show until after commit
+    upload.upload(context['package'].resources[-1].id,
+                  uploader.get_max_resource_size())
+
+    data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0] or upload.mimetype
+    data_dict['size'] = upload.filesize
 
     pkg_dict['resources'].append(data_dict)
 
@@ -307,10 +315,6 @@ def resource_create(context, data_dict):
         errors = e.error_dict['resources'][-1]
         raise ValidationError(errors)
 
-    ## Get out resource_id resource from model as it will not appear in
-    ## package_show until after commit
-    upload.upload(context['package'].resources[-1].id,
-                  uploader.get_max_resource_size())
     model.repo.commit()
 
     ##  Run package show again to get out actual last_resource
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -301,8 +301,11 @@ def resource_create(context, data_dict):
     upload.upload(context['package'].resources[-1].id,
                   uploader.get_max_resource_size())
 
-    data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0] or upload.mimetype
-    data_dict['size'] = upload.filesize
+    if not 'mimetype' in data_dict:
+        data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0] or upload.mimetype
+
+    if not 'size' in data_dict:
+        data_dict['size'] = upload.filesize
 
     pkg_dict['resources'].append(data_dict)
 
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -298,9 +298,7 @@ def resource_create(context, data_dict):
     upload = uploader.get_resource_uploader(data_dict)
 
     if not 'mimetype' in data_dict:
-        if 'url' in data_dict:
-            data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0]
-        elif hasattr(upload, 'mimetype'):
+        if hasattr(upload, 'mimetype'):
             data_dict['mimetype'] = upload.mimetype
 
     if not 'size' in data_dict:
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -296,16 +296,16 @@ def resource_create(context, data_dict):
         pkg_dict['resources'] = []
 
     upload = uploader.get_resource_uploader(data_dict)
-    ## Get out resource_id resource from model as it will not appear in
-    ## package_show until after commit
-    upload.upload(context['package'].resources[-1].id,
-                  uploader.get_max_resource_size())
 
     if not 'mimetype' in data_dict:
-        data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0] or upload.mimetype
+        if 'url' in data_dict:
+            data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0]
+        elif hasattr(upload, 'mimetype'):
+            data_dict['mimetype'] = upload.mimetype
 
     if not 'size' in data_dict:
-        data_dict['size'] = upload.filesize
+        if hasattr(upload, 'filesize'):
+            data_dict['size'] = upload.filesize
 
     pkg_dict['resources'].append(data_dict)
 
@@ -317,6 +317,11 @@ def resource_create(context, data_dict):
     except ValidationError, e:
         errors = e.error_dict['resources'][-1]
         raise ValidationError(errors)
+
+    ## Get out resource_id resource from model as it will not appear in
+    ## package_show until after commit
+    upload.upload(context['package'].resources[-1].id,
+                  uploader.get_max_resource_size())
 
     model.repo.commit()
 
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -292,16 +292,16 @@ def resource_create(context, data_dict):
     for plugin in plugins.PluginImplementations(plugins.IResourceController):
         plugin.before_create(context, data_dict)
 
-    if not 'resources' in pkg_dict:
+    if 'resources' not in pkg_dict:
         pkg_dict['resources'] = []
 
     upload = uploader.get_resource_uploader(data_dict)
 
-    if not 'mimetype' in data_dict:
+    if 'mimetype' not in data_dict:
         if hasattr(upload, 'mimetype'):
             data_dict['mimetype'] = upload.mimetype
 
-    if not 'size' in data_dict:
+    if 'size' not in data_dict:
         if hasattr(upload, 'filesize'):
             data_dict['size'] = upload.filesize
 
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -100,7 +100,7 @@ def resource_update(context, data_dict):
         if hasattr(upload, 'mimetype'):
             data_dict['mimetype'] = upload.mimetype
 
-    if not 'size' in data_dict:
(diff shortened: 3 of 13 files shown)


ckan/logic/action/create.py
```diff
@@ -295,6 +296,13 @@ def resource_create(context, data_dict):
         pkg_dict['resources'] = []
 
     upload = uploader.get_resource_uploader(data_dict)
+    ## Get out resource_id resource from model as it will not appear in
+    ## package_show until after commit
+    upload.upload(context['package'].resources[-1].id,
+                  uploader.get_max_resource_size())
+
+    data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0] or upload.mimetype
```
amercader: We need to check if the user has provided a `mimetype` or `size` in the data_dict before overriding them

ckan/logic/action/update.py
```diff
@@ -93,6 +94,10 @@ def resource_update(context, data_dict):
         plugin.before_update(context, pkg_dict['resources'][n], data_dict)
 
     upload = uploader.get_resource_uploader(data_dict)
+    upload.upload(id, uploader.get_max_resource_size())
+
+    data_dict['mimetype'] = mimetypes.guess_type(data_dict['url'])[0] or upload.mimetype
```
amercader: Same here, if the user provides a value for `mimetype` or `size`, don't override it

ckan/tests/logic/action/test_create.py
```diff
@@ -443,6 +443,64 @@ def test_doesnt_require_url(self):
...
+
+    def test_size_of_resource_by_upload(self):
+        pass
+        '''
+        context = {}
+        params = {
+            'package_id': factories.Dataset()['id'],
+            'url': 'http://data',
+            'name': 'A nice resource',
+            'upload': file('/home/vagrant/test.txt', 'wb+')
```
jrods: @amercader how can i test for real file uploads? 
i know that the ResourceUpload expects `cgi.FieldStorage` obj to store the file that was uploaded, is there another way?

TkTech: Not @amercader, but cgi.FieldStorage is a very basic object, you can just subclass it and mock it out which is what I do in the ckanext-cloudstorage tests.

jrods: @TkTech I'm trying with a fake FieldStorage class, but somewhere in the test it's failing and the error message is a red herring and not telling me the actual fail point?

```
FAIL: ckan.tests.logic.action.test_create.TestResourceCreate.test_size_of_resource_by_upload
  File "/usr/lib/ckan/default/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/vagrant/ckan/lib/default/src/ckan/ckan/tests/logic/action/test_create.py", line 498, in test_size_of_resource_by_upload
    assert size
AssertionError:
-------------------- >> begin captured stdout << ---------------------
FieldStorage('upload', 'test.txt')
{'name': 'A nice resource',
 'package_id': u'fdb3bb89-bd74-46ae-bdbb-44b896f7c7e8',
 'upload': FieldStorage('upload', 'test.txt'),
 'url': 'http://data'}

--------------------- >> end captured stdout << ----------------------
```

```
def test_size_of_resource_by_upload(self):
        import cgi
        class FakeFileStorage(cgi.FieldStorage):
            def __init__(self, fp, filename):
                self.file = fp
                self.filename = filename
                self.name = 'upload'


        with open('/home/vagrant/test.txt', 'rb') as test_file:
            test_resource = FakeFileStorage(test_file, 'test.txt')

        print(test_resource)
        context = {}
        params = {
            'package_id': factories.Dataset()['id'],
            'url': 'http://data',
            'name': 'A nice resource',
            'upload': test_resource
        }
        result = helpers.call_action('resource_create', context, **params)

        size = result.pop('size')
        assert size
```

ckan/tests/logic/action/test_create.py
```diff
@@ -470,23 +477,36 @@ def test_mimetype_by_user(self):
         assert_equals(mimetype, 'application/csv')
 
     def test_mimetype_by_upload(self):
-        pass
+        test_file = file('/home/vagrant/test.txt', 'rb')
```
jrods: could use ckan source file? or just make a string and make a file obj based on that, not sure yet

amercader:
@jrods to mock uploads you can try @TkTech 's approach of mocking cgi.FileStorage or have a look at these ones that use another library `pyfakefs` to mock a filesystem:

https://github.com/ckan/ckan/blob/f0ab2b39b32d16f07b215a82a44d0ee3a3d93021/ckanext/example_iuploader/test/test_plugin.py

I haven't used either myself though

kfishwick:
@amercader  can you please re-open issue https://github.com/ckan/ckan/issues/1517 ?

wardi:
I'd like to see a configuration option to keep the old behaviour because this is deep in ckan code that can't be overridden easily, and changes the metadata returned for resources.

You've got two strategies taken here, one guesses the type using magic and the other is based on the file extension. Do both these strategies return the same set of mime type values?

jrods:
The user can still override the guessed mimetype and file size, by just including the values they want when creating/updating the resource. 

Guessing by the file extension is more accurate. Magic is a last resort if resource url doesn't provide the file extension and then the file contents are used to guess the mimetype.

These changes only effect newly created resources or updating the resource data. The old default values for size and mimetype were both None. The new defaults values for size and mimetype are now 0 and <a mimetype> respectively.

amercader:
@jrods Pin the python-magic requirement to 0.4.6 as we pin all requirements to a specific version, and also add it to requirements.txt so the tests pass (the test environment is built from this list)

jrods:
@amercader @wardi @TkTech how do change config key/values within a test? Is there an example I can look at?

amercader:
Sorry @jrods totally missed this. Check for usage of the `change_config` decorator in the tests, eg here: https://github.com/ckan/ckan/blob/f0ab2b39b32d16f07b215a82a44d0ee3a3d93021/ckan/tests/lib/test_datapreview.py#L80

jrods:
@amercader can you look at the tests that are failing in circleci? They seem to be unrelated to my changes, but the change i made to test-core.ini may be the cause, i'm not sure yet.

amercader:
@jrods they are related to the org / groups image field, which is also an upload. These use the generic `Upload` class in lib.uploader, so maybe you need to investigate if it needs some tweaking to support the new changes

jrods:
@amercader I just ran the tests in my master branch a5408647c18bcc3a6317a8c83cfb37c72068e90c (none of my modifications are in the code base) and all the same tests fail in my local env.

amercader:
OK @jrods, this was a pain to debug but I finally got to the bottom of it (it ended up being a matter of pride!)

The bottom line is that the failures where caused by [enabling the file uploads](https://github.com/ckan/ckan/pull/3205/files#diff-28d7d03245b4b1b19e2acab0c903dc29R56) on the tests. This made the file upload fields appear in forms and triggered [this WebTest bug](https://github.com/Pylons/webtest/issues/146), which was the one appearing on CircleCI.

Even if we go to the bottom of this we can't enable the file uploads, at least as you did providing a specific path (`/home/ubuntu/uploads`) as this is not portable, because other developers or environments won't be able to create this directory (and files won't be cleaned up etc).

There was some discussion around this on #3037 and the solution was to mock a lot of things and use pyfakefs, as in the [example](https://github.com/ckan/ckan/blob/f0ab2b39b32d16f07b215a82a44d0ee3a3d93021/ckanext/example_iuploader/test/test_plugin.py#L79) I linked previoulsy on this thread. It is ugly but it's best shot we have now at making your tests pass. I've simplified the changes down to this, which makes the test pass:

```diff
diff --git a/ckan/tests/logic/action/test_create.py b/ckan/tests/logic/action/test_create.py
index 9a1b0eb..3cf44e3 100644
--- a/ckan/tests/logic/action/test_create.py
+++ b/ckan/tests/logic/action/test_create.py
@@ -3,7 +3,7 @@
 '''Unit tests for ckan/logic/auth/create.py.
 
 '''
-
+import ckan
 from ckan.common import config
 import mock
 import nose.tools
@@ -17,6 +17,22 @@ import ckan.plugins as p
 assert_equals = nose.tools.assert_equals
 assert_raises = nose.tools.assert_raises
 
+# Mock file uploads
+import __builtin__ as builtins
+from pyfakefs import fake_filesystem
+real_open = open
+fs = fake_filesystem.FakeFilesystem()
+fake_os = fake_filesystem.FakeOsModule(fs)
+fake_open = fake_filesystem.FakeFileOpen(fs)
+
+
+def mock_open_if_open_fails(*args, **kwargs):
+    try:
+        return real_open(*args, **kwargs)
+    except (OSError, IOError):
+        return fake_open(*args, **kwargs)
+
+
 
 class TestUserInvite(object):
 
@@ -450,7 +466,12 @@ class TestResourceCreate(object):
 
         assert not stored_resource['url']
 
-    def test_mimetype_by_url(self):
+    @helpers.change_config('ckan.storage_path', '/doesnt_exist')
+    @mock.patch.object(ckan.lib.uploader, 'os', fake_os)
+    @mock.patch.object(builtins, 'open', side_effect=mock_open_if_open_fails)
+    @mock.patch.object(ckan.lib.uploader, '_storage_path', new='/doesnt_exist')
+    def test_mimetype_by_url(self, mock_open):
+
         """
         The mimetype is guessed from the url

```

Remove the changes on `test-core.ini`, update the rest of your new tests as above and we should be good to go :chart_with_upwards_trend:.

jrods:
@amercader your example worked for the test that require a file upload, thank you

However on one failing test: [test_model_dictize.py](https://github.com/ckan/ckan/blob/master/ckan/tests/lib/dictization/test_model_dictize.py#L395-L423) (it fails in the [`assert_equals_expected`](https://github.com/ckan/ckan/blob/master/ckan/tests/lib/dictization/test_model_dictize.py#L318)  function), 
I'm stuck on the default type of 'size'; It's defaulting to unicode instead of int/bigint. In the Resource model, in resource_table, I have the size column as defined `Column('size', types.BigInteger, default=0)`. But in the failing test, it's u'0' on the assert comparison.
I don't have enough knowledge on adding a default value to the model, if what I'm doing is correct (probably not), but will you enlighten me on the proper way of defaulting a column's value?

jrods:
@amercader i found the issue with the size var being unicode in this context, it seems in the [table_dictize init call](https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/__init__.py#L42-L53), it wasn't accounting for long data types.

amercader:
@jrods glad we are making progress!

Why are you defaulting the size to 0 at the database level? Is that a requirement of the new code? I'm not saying we can not add the change but it's a change in the current behaviour, and perhaps `None` meaning that the size hasn't been checked is more intuitive.

jrods:
@amercader the reason was for a failing test, along with changing the test expected dict for a resource. I believe this was due to where i was initializing `self.filesize` in the uploader module.
I moved `self.filesize` to be initialed when there's only an actual file upload. I also reverted the expect dict for a resource, where size defaults to `None`. All the tests are passing now =)

ckan/lib/dictization/__init__.py
```diff
@@ -45,6 +45,8 @@ def table_dictize(obj, context, **kw):
             result_dict[name] = value
         elif isinstance(value, int):
             result_dict[name] = value
+        elif isinstance(value, long):
```
amercader: Leave it, sounds like the right approach anyway. But add a quick note on the changelog file for the next version saying that the `resource['size']` property will change from a string to a number, in case someone relies on it.

amercader:
@jrods a couple minor comments and we are good to go! :rocket:

ckan/lib/dictization/__init__.py
jrods: I've added a minor note to the changelog in my latest commit

jrods:
@amercader I've added those changes to the PR for the changelog and pep8.

amercader:
Thanks for your patience and your great work @jrods

jrods:
@amercader no problem, thanks for all the help =) I'm glad it's finally merged in

PR #3206, opened 2016-08-18
[#3196] Flask debugging and testing

amercader:
This targets master so can be merged when reviewed.

Adds the Flask [DebugToolbar](https://flask-debugtoolbar.readthedocs.io/) middleware, which provides lots of useful information when debugging. It requires the `SECRET_KEY` config option to be set up, which we will need for the sessions anyway. As with the repoze.who auth token, we fall back to `beaker.session.secret` if it's not present.

Also sets a testing config option required by Flask when running the tests. Flask behaves differently if `TESTING` is True, eg not capturing exceptions when serving a page, which is the behaviour we have in the Pylons tests.


amercader:
Merged as part of #3212

PR #3208, opened 2016-08-18
[#3196] Common session for Flask and Pylons

amercader:
This is based off #3207 

Beaker backend for Flask, so we can share the session across Pylons and Flask. Includes tests.

All fantastic work from @brew in the POC branch


torfsen:
These SSL failures look like the ones from #3243 which was supposed to be fixed by #3247.

amercader:
Test failure unrelated

ckan/config/middleware/flask_app.py
```diff
@@ -71,6 +73,27 @@ def make_flask_stack(conf, **app_conf):
...
+        def open_session(self, app, request):
+            if 'beaker.session' in request.environ:
+                return request.environ['beaker.session']
+
+        def save_session(self, app, session, response):
+            session.save()
+
+    cache_dir = app_conf.get('cache_dir') or app_conf.get('cache.dir')
+    session_opts = {
+        'session.data_dir': '{data_dir}/sessions'.format(
```
TkTech: Do we *really* want to use the file-based session type? We could do `'session.type': 'ext:database'` and store it in postgres.

amercader:
@TkTech I pushed a fix for supporting all beaker config options.

Regarding using the db backend by default I also prefer it and that's what I used on my last deployments. The only thing worth noting is that adds another step when users are setting up a new instance (they need to edit `beaker.session.url`). Unless we do some magic and set the value to be the same as `sqlalchemy.url` if it's not set.

Personally as long as you can use any Beaker options to configure them however you like (as with my latest push) and this is documented I don't see a big problem with using the file backend by default, but happy to consider it.

TkTech:
Yup, either approach works now that it can be fully customized. 👍

amercader:
Happy to merge @TkTech ?

PR #3192, opened 2016-08-18
IPermissionLabels

wardi:
This is a suggested change to the dataset view permissions.

At the moment we return only public datasets from `package_search` #3176 and `package_list`. Private datasets are visible from the front end on the organization page if you're a member of the organization (unless the controller has been overridden). If you know the dataset names or ids with `package_show` (unless the package_show auth function has been overridden).

Users might never find out about datasets they have permission to view e.g. if they only use the API, if they don't know the correct org to search or if search controller is overridden. Users might see datasets listed in an organization search page that they are prevented from viewing if the package_show auth function doesn't match logic in search controller.

Let's deprecate customizing the package_show auth function and replace it with permission labels in the search index.

We implement our current permissions by adding `permission_labels = [pkg.owner_org]` to the search index for private datasets, and `permission_labels = ['public']` for public datasets. For non-sysadmins `package_search` will call a new plugin interface to get a list of permission labels for the search. By default that interface will return `'public'` and the UUID of organizations the current user is a member of.

CKAN plugins would be able to implement that new interface and override before_index to set their own permission labels. This scheme should cover many types of permission business logic, including full 1:1 user permissions for viewing datasets.


TkTech:
`(public OR member-abcd OR member-9876 OR creator-1234)`

Just wanted to mention, don't do this with `OR`. Instead, issue multiple fq filters.

```
&fq=permission_labels:public
&fq=permission_labels:member-abcd
&fq=permission_labels:member-9876
etc...
```

In general for solr caching you want your filter queries to be simple especially in the case of `permission_labels:public`. The above example will generate 3 caching paths that will benefit future queries of the same type before they're pushed out of the cache (if you're filtering on a member, it's likely that user will be looking at other datasets very soon). The compound query on the other hand will generate a massive cache key that will not benefit any subset.

wardi:
@TkTech `&fq=permission_labels:public &fq=permission_labels:member-abcd` is the same as `&fq=permissions_labels:(public AND member-abcd)` unless I'm mis-reading the solr docs

TkTech:
@wardi while functionally equivalent, they have different penalties.

`&fq=permissions_labels:(public AND member-abcd)` generates a single filter query cache entry, with a cache key of `(public AND member-abcd)`. This cache is useful for this, and only this query.

`&fq=permission_labels:public &fq=permission_labels:member-abcd`

Generates 3 cache entries with 3 different keys that can be re-used for similar queries.

wardi:
@TkTech ok, but I'm looking for a way for my code to do "OR", not "AND"

TkTech:
That just clicked, ignore me :)

wardi:
This is a functioning replacement for our current package_search and package_show auth based on permission labels. The new interface IPermissionLabels could use a test extension to show how it could be used to change the permission caculation.

Any ideas for a plausible permission scheme?  Maybe a plugin that lets users create "proposed" datsets that are only visible to the creator and admins of the organization.

wardi:
@TkTech @amercader this PR now has an `example_ipermissionlabels` extension and tests that show the `IPermissionLabels` interface in action.

ckanext/example_ipermissionlabels/tests/test_example_ipermissionlabels.py
```diff
@@ -0,0 +1,115 @@
...
+    def teardown_class(cls):
+        ckan.plugins.unload(u'example_ipermissionlabels')
+
+    def test_normal_dataset_permissions_are_normal(self):
+        user = factories.User()
+        user2 = factories.User()
+        user3 = factories.User()
+        org = factories.Organization(user=user)
+        org2 = factories.Organization(user=user2)
+        call_action(
```
amercader: You can do:

```python

org2 = factories.Organization(user=user2, users=[{'name': user3['id'], 'capacity': 'member'}])

```

ckan/lib/search/index.py
```diff
@@ -279,6 +279,13 @@ def index_package(self, pkg_dict, defer_commit=False):
 
         assert pkg_dict, 'Plugin must return non empty package dict on index'
 
+        # permission labels determine visibility in search, can't be set
+        # in original dataset or before_index plugins
+        labels = lib_plugins.get_permission_labels()
+        dataset = model.Package.get(pkg_dict['id'])
+        pkg_dict['permission_labels'] = labels.get_dataset_labels(
+            dataset) if dataset else [] # TestPackageSearchIndex-workaround
```
amercader: Why are we passing the model object instead of the dict?

ckan/lib/search/query.py
```diff
@@ -310,18 +315,23 @@ def run(self, query):
...
-            fq += " +state:active"
-        query['fq'] = [fq]
+        if not '+state:' in query.get('fq', ''):
+            fq.append('+state:active')
 
-        fq_list = query.get('fq_list', [])
-        query['fq'].extend(fq_list)
+        # only return things we should be able to see
+        if permission_labels is not None:
+            fq.append('+permission_labels:(%s)' % ' OR '.join(
```
amercader: Should we check if an extension has already messed with a `permission_labels` `fq` param before this point (eg adding it in the `before_search` plugin hook) and completely override it?

wardi:
@amercader private datasets in the search results was merged in #3191, that's not new behaviour in this PR

wardi:
@amercader the old search API calling lib.search is fine, that's where these permissions checks are implemented.

ckan/lib/search/index.py
wardi: I started building this interface expecting user and dataset dicts, but I ended up switching to the models instead.

These permissions are partially enforced below the logic layer (lib.search). Calling back up to the logic layer to get the dict version of the user object doesn't make sense, also it would make what used to be a super-cheap package_show auth function way more expensive to call. It's also inconsistent to have one part of the interface take a model and the other part take a dict.

Also with @TkTech's plugin extras changes we might have extras attached to users and datasets that aren't part of the dict presented to the user (e.g. permissions or labels applied by someone else) and having the model will make those extras easier to access.

ckan/lib/search/query.py
wardi: I decided to clearly separate permissions-related information from user metadata. The only place to set permission labels is the IPermissionLabels interface.

Issue #3151, opened 2016-07-05
Plural translation strings in JavaScript are not extracted


torfsen:
It seems that plural translation strings (using Jed's `isPlural` function) don't get extracted to `ckan.pot` when running `python setup.py extract_messages`.

Here's a small demo JS file:

``` javascript
this.ckan.module('message-extraction-test', function ($, _) {

  initialize: function() {
    console.log(_('Raw Singular Translation').fetch());
    console.log(_('Raw Plural Translation').isPlural(1, 'I can haz plural?').fetch());
    console.log(this.i18n('singular'));
    console.log(this.i18n('plural', {count: 2}));
  },

  options: {
    i18n: {
      singular: _('Nothing fancy'),
      plural: function (data) {
        return _('%(count)d item').isPlural(data.count, '%(count)d items');
      }
    }
  }
});
```

From that, `python setup.py extract_messages` extracts

```
#: ckan/message_extraction_test.js:4
msgid "Raw Singular Translation"
msgstr ""

#: ckan/message_extraction_test.js:5
msgid "Raw Plural Translation"
msgstr ""

#: ckan/message_extraction_test.js:12
msgid "Nothing fancy"
msgstr ""

#: ckan/message_extraction_test.js:14
#, python-format
msgid "%(count)d item"
msgstr ""
```

Note that `"I can haz plural?"` and `"%(count)d items"` are missing. The correct behavior would have been to recognize them as plural forms of `"Raw Plural Translation"` and `"%(count)d item"`, respectively, e.g.

```
#: ckan/message_extraction_test.js:5
msgid "Raw Plural Translation"
msgid_plural "I can haz plural?"
msgstr[0] ""
msgstr[1] ""
```

It seems that CKAN [configures Babel](https://github.com/ckan/ckan/blob/4bffd8c2611ae9eacec6b8593681fcc287d9a622/setup.cfg) to take `isPlural` into account:

```
[extract_messages]
keywords = translate isPlural
add_comments = TRANSLATORS:
output_file = ckan/i18n/ckan.pot
width = 80
```

However, that doesn't seem to be working.

torfsen:
This might be a limitation of babel, see python-babel/babel#72.

torfsen:
[...]

PR #3211, opened 2016-08-22
[#3151] Extractable plural translations in JavaScript

torfsen:
This PR makes plural translations in JS extractable (#3151).

As discussed in #3151, this PR adds two new JS functions `ckan.i18n._` and `ckan.i18n.ngettext` for singular and plural translations. In contrast to the previous `ckan.i18n.translate` function they do not expose the Jed instance which we use for the actual translation but return the fully translated strings (with inserted placeholder values where appropriate).

Correspondingly, JS CKAN modules now have `this._` and `this.ngettext` methods as shortcuts for the functions described above. The methods replace the old `this.i18n` / `this.options.i18n` functionality. Similarly, module declaration functions should not take `ckan.i18n` and `ckan.i18n.translate` arguments anymore.

The existing modules have been updated accordingly.

The old behavior has been left in place and should still work but is marked as deprecated. The changes should therefore be backwards-compatible.

The PR is a WIP, in particular the documentation has not been updated. However, the main changes are implemented and I think it's a good time to get some feedback to make sure that we agree on the details.

I haven't included the resulting changes in `ckan.pot` (i.e. the now extracted plural strings) since, as far as I understand, the actual translations are not managed via PRs.

### DIFF
--- a/ckan/public/base/javascript/i18n.js
+++ b/ckan/public/base/javascript/i18n.js
@@ -1,3 +1,47 @@
+'use strict';
+
+/*
+ * CKAN JavaScript i18n functionality.
+ *
+ * Adds an `i18n` attribute to the CKAN object (`this.ckan`).
+ *
+ * Singular strings can be translated using `i18n._`:
+ *
+ *     ckan.i18n._('Hello, world!')
+ *
+ * The function will return the translated string in the currently selected
+ * language. Often, a translateable string contains dynamic parts, for example
+ * a username. These can be included via named `%`-placeholders:
+ *
+ *     ckan.i18n._('Hello, %(name)s!', {name: 'Jessica'})
+ *
+ * After translation the placeholders are replaced using the values passed in
+ * the second argument.
+ *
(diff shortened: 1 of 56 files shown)


amercader:
@torfsen is this still needed after #3272? If so is it ready for reviewing?

torfsen:
@amercader This is different from #3272: while #3272 allows extensions to provide translations for JS, this PR fixes #3151, i.e. the problem that plural translations in JS code are currently not extracted.

The PR is ready in the sense that it contains the necessary fixes as discussed with @TkTech in #3151. The documentation still needs to be updated, but since the PR introduces a new public JS i18n API I wanted to get some feedback first.

So feel free to review this at any time.

amercader:
@TkTech you gave some thought to this on #3151 so you might want to have a look at this. It looked fine to meet without knowing the full context.
@torfsen feel free to send the rest of the docs missing.

torfsen:
I've updated the documentation.

TkTech:
👍

amercader:
great stuff @torfsen

PR #3212, opened 2016-08-23
[#3196] Common requests code for Flask and Pylons

amercader:
This PR is based off #3208, and also contains #3194

There is a bunch of code located on the `BaseController` which is executed before and after every single request. This PR abstracts it to the `ckan.views` module so it can be reused by Flask as well, with the `before_request` and `after_request` handlers.

Code relates to:
- Identifying the user via the repoze cookie or the API header
- Check session cookies
- Set CORS headers
- Timer for requests
- Update CKAN app_globals


ckan/config/middleware/flask_app.py
```diff
@@ -14,7 +37,18 @@ def make_flask_stack(conf, **app_conf):
     """ This has to pass the flask app through all the same middleware that
     Pylons used """
 
+    root = os.path.dirname(
+        os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
TkTech: We lose support for deploying as an archive or other importers doing this don't we? Although I bet it's already broken for similar things in other places. We should use the helpers in `pkg_resources` to get the true root.

ckan/config/middleware/flask_app.py
```diff
@@ -24,6 +58,81 @@ def make_flask_stack(conf, **app_conf):
...
+
+    @app.after_request
+    def ckan_after_request(response):
+        # Check session cookie
+        response = check_session_cookie(response)
+
+        # Set CORS headers if necessary
+        response = set_cors_headers_for_response(response)
+
+        # Log time between before and after view
```
TkTech: Not a fan of this being a built-in/default behaviour. If you're developing/debugging it's part of Flask-debugtoolbar already and if you're in production it's part of whatever layer you're using above CKAN or 99% of tools like Sentry for logging. Would just get rid of this.

ckan/config/middleware/flask_app.py
```diff
@@ -24,6 +58,81 @@ def make_flask_stack(conf, **app_conf):
...
+        jinja_extensions.UrlForExtension
+    ]
+    for extension in extensions:
+        app.jinja_env.add_extension(extension)
+    app.jinja_env.filters['empty_and_escape'] = \
+        jinja_extensions.empty_and_escape
+    app.jinja_env.filters['truncate'] = jinja_extensions.truncate
+
+    # Common handlers for all requests
+    @app.before_request
```
TkTech: Move this, `ckan_after_request`, `helper_functions` and `c_obj` out to module-level methods.

ckan/config/middleware/flask_app.py
```diff
@@ -32,6 +141,46 @@ def hello_world():
...
+    # Auto-register all blueprints defined in the `views` folder
+    _register_core_blueprints(app)
+
+    # Set up each IBlueprint extension as a Flask Blueprint
+    for plugin in PluginImplementations(IBlueprint):
+        if hasattr(plugin, 'get_blueprint'):
+            app.register_extension_blueprint(plugin.get_blueprint())
+
+    # Start other middleware
+
```
TkTech: Extra newline

ckan/config/middleware/flask_app.py
```diff
@@ -57,17 +231,61 @@ def can_handle_request(self, environ):
...
+        Sets the rule property `ckan_core` to False, to indicate that the rule
+        applies to an extension route.
+        '''
+        self.register_blueprint(blueprint, **kwargs)
+
+        # Get the new blueprint rules
+        bp_rules = [v for k, v in self.url_map._rules_by_endpoint.items()
+                    if k.startswith(blueprint.name)]
+        bp_rules = list(itertools.chain.from_iterable(bp_rules))
```
TkTech: How does this work if you have for example a blueprint called `user` and `users` which result in endpoints of `user.test` and `users.test`? You'll match the `users.test` endpoint for both.

Also, don't need to make three extra lists here...

```
bp_rules = itertools.chain.from_iterable(
    v for k, v in self.url_map._rules_by_endpoint.iteritems()
    if k.startswith(blueprint.name)
)
```

ckan/config/middleware/flask_app.py
```diff
@@ -57,17 +231,61 @@ def can_handle_request(self, environ):
...
+            r.ckan_core = False
+            r.match_compare_key = lambda: top_compare_key
+
+
+def _register_core_blueprints(app):
+    u'''Register all blueprints defined in the `views` folder
+    '''
+    views_path = os.path.join(os.path.dirname(__file__),
+                              u'..', u'..', u'views')
+    module_names = [f.rstrip(u'.py')
```
TkTech: Doesn't need to be a list, can be a generator.

ckan/config/middleware/flask_app.py
```diff
@@ -57,17 +231,61 @@ def can_handle_request(self, environ):
...
+                    for f in os.listdir(views_path)
+                    if f.endswith(u'.py') and not f.startswith(u'_')]
+    blueprints = []
+    for name in module_names:
+        module = importlib.import_module(u'ckan.views.{0}'.format(name))
+        blueprints.extend([m for m in inspect.getmembers(module)
+                           if isinstance(m[1], Blueprint)])
+    if blueprints:
```
TkTech: No need to check for a non-empty list here, `for blueprint in blueprints` on an empty list is a nop anyways.

amercader:
@TkTech I'm going to address your comments. Just to confirm, you are happy to base the review on this PR and not the previous individual ones that led to this one (see #3196)? ie should I push fixes just to these one? Cheers

ckan/config/middleware/flask_app.py
amercader: `app` is only available in the context of the `make_flask_stack` function, so I guess I can't use the decorator at the module level. Is there a way of moving the functions and assigning them directly to `app`? All I could find is `app.before_request_funcs`, but it looks like it not should be messed with directly (the docs say to use the decorator).

ckan/config/middleware/flask_app.py
amercader: To avoid this we can check if the key starts with the blueprint name plus a dot

ckan/config/middleware/flask_app.py
amercader: Can you point me to the method on `pkg_resources` to use or an example? All my googling returned approaches like the one above.

ckan/config/middleware/flask_app.py
TkTech: You can just call the decorator as a method and pass it.

amercader:
@TkTech I addressed all your comments except for https://github.com/ckan/ckan/pull/3212#discussion_r87624527

ckan/views/api.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def _finish_ok(response_data=None,
+               content_type=u'json',
+               resource_location=None):
+    u'''If a controller method has completed successfully then
+    calling this method will prepare the response.
+    @param resource_location - specify this if a new
+       resource has just been created.
+    @return response message - return this value from the controller
```
TkTech: What style of comment parsing is this? I'm not familiar with it, kinda looks like javadoc instead of sphinx?

ckan/views/api.py
```diff
@@ -0,0 +1,103 @@
...
+API_DEFAULT_VERSION = 3
+API_MAX_VERSION = 3
+
+
+# Blueprint definition
+
+api = Blueprint(u'api', __name__, url_prefix=u'/api')
+
+
+# Private methods
```
TkTech: Getting flashbacks to `#pragma` blocks in CodeWarrior :) Let's not do section comments if we can help it, the code itself should be self-describing. `_` prefix means "private", `__` "protected", a Blueprint is clearly a blueprint, etc...

amercader:
@TkTech how is this looking? any other issues to address?

amercader:
Bump @TkTech

TkTech:
👍

amercader:
@TkTech you mean the Flask debugger? or the old Pylons-based custom one? Let me know if I can help with this.

TkTech:
@amercader The existing one - go ahead and raise an error with debugging enabled, then try to use the interactive prompt. I will resolve.

TkTech:
@amercader the broken pylons debugger is fixed with the Pylons 1.0.2c upgrade PR.

amercader:
@TkTech great. Do you want to merge this one first (if happy) or wait for the other?

wardi:
huzzah!

amercader:
Woo!! Thanks @TkTech, huge step forward on our journey to Flask!

I've updated all the remaining PRs to make sure they are up to date with these changes:

* #3208 - Common session object
* #3213 - I18n
* #3228 - url_for
* #3239 - API Blueprint 

They should be a piece of cake after this one! :)

tino097:
@mattfullerton we have a branch using boto3 and fixed that issue, but yet to be merged. If you can, you can review the code and we could make PR 
https://github.com/keitaroinc/ckanext-s3filestore/blob/boto3-fix/ckanext/s3filestore/controller.py#L22

mattfullerton:
@tino097 Thanks, I've reviewed code (see one minor pull request), switched our Dockerfile to that branch and will hand over to the deployment guys. I'll let you know if its working well there.

mattfullerton:
@tino097 What is the purpose of the new config setting ckanext.s3filestore.host_name? I don't see any use of it in the code but would still like to set it to something plausible/something that will work in the future.

mattfullerton:
@tino097 And quick docs on the new settings (ckanext.s3filestore.region_name  - I assume something like "eu-central-1" is what's wanted here, and ckanext.s3filestore.signature_version - "s3" or "s3v4" is what's required here (after some digging)?) as well as host name would probably be good before making the pull request

tino097:
@mattfullerton  those should be part of the other branch as we are tried to use minio and host_name  should be used whenever you want to use S3 or Minio. As signature(it should be s3v4) or region name(set to eu-central-1 as default), those were added to get more configurable code instead of hardcoded values in the code. 
Sorry that there are no updated documents as some of the branches are still in test period

PR #3216, opened 2016-08-25
[#2842] config object fixes

amercader:
Small fixes for the common `config` object introduced in #3163


amercader:
@LinkSDM @davidread  I've added this to the CKAN 2.6 milestone so we can keep track of what needs to be cherry-picked to the release-v2.6.0 branch. If that does not work we can change to labels or something else but let's see how it goes for now.

PR #3219, opened 2016-08-26
Fix typo in 'paster datastore resubmit'

argv0:
"resources_ids" should be "resource_ids".  Causes resubmit command to fail.


amercader:
Nice catch @argv0, thanks

PR #3221, opened 2016-08-29
3220 switched abort from 404 to 401, user redirected to login page

gleb-rudenko:
Fixes #
Anonymous users are now redirected to login page instead of Page not found 404 when trying to access /dashboard page.
### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -80,7 +80,7 @@ class UserController(base.BaseController):
         try:
             user_dict = get_action('user_show')(context, data_dict)
         except NotFound:
-            abort(404, _('User not found'))
+            abort(401, _('Authorize to see this page'))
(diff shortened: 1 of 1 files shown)


ckan/controllers/user.py
```diff
@@ -80,7 +80,7 @@ def _setup_template_variables(self, context, data_dict):
         try:
             user_dict = get_action('user_show')(context, data_dict)
         except NotFound:
-            abort(404, _('User not found'))
+            abort(401, _('Authorize to see this page'))
```
wardi: 401 indicates an authentication error, not an authorization error, so it's not appropriate here.

If you want to send a redirect, send a redirect. Don't rely on repoze's magic redirection to a login when a 401 is sent.

PR #3224, opened 2016-08-30
Redirect to login page instead of Page not found 404 when anonymous user is trying to access to /dashboard page and all user related pages

gleb-rudenko:
Fixes #
Anonymous users are now redirected to login page instead of Page not found 404 when trying to access /dashboard page.
### Features:
- [ ] includes API changes


gleb-rudenko:
Unlike the previous PR (https://github.com/ckan/ckan/pull/3221), I used redirect as you proposed

ckan/controllers/user.py
```diff
@@ -80,7 +80,8 @@ def _setup_template_variables(self, context, data_dict):
         try:
             user_dict = get_action('user_show')(context, data_dict)
         except NotFound:
-            abort(404, _('User not found'))
+            h.flash_error('Authorize to see this page')
+            h.redirect_to(controller='user', action='login')
```
torfsen: This doesn't seem to distinguish between a missing authentication and a user that genuinely doesn't exist. If `get_action('user_show')` raises `NotFound`, then a 404 seems like the right response. If that's not the case in the situtation you're trying to fix then perhaps `user_show` shouldn't raise `NotFound` in the first place, but `NotAuthorized` (or, if we're just requiring that the user is logged in, a currently not existing `NotAuthenticated`).

gleb-rudenko: There is another way that we can try, by creating NotAuthenticated
`diff --git a/ckan/controllers/user.py b/ckan/controllers/user.py
index 7147b83..0df45c0 100644
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -29,6 +29,7 @@ render = base.render
 check_access = logic.check_access
 get_action = logic.get_action
 NotFound = logic.NotFound
+NotAuthenticated = logic.NotAuthenticated
 NotAuthorized = logic.NotAuthorized
 ValidationError = logic.ValidationError
 UsernamePasswordError = logic.UsernamePasswordError
@@ -80,6 +81,8 @@ class UserController(base.BaseController):
         try:
         except NotFound:
+            abort(404, _('User not found'))
+        except NotAuthenticated:
             h.flash_error('Authorize to see this page')
             h.redirect_to(controller='user', action='login')
         except NotAuthorized:
diff --git a/ckan/logic/__init__.py b/ckan/logic/__init__.py
index 7f86888..1218ffa 100644
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -56,6 +56,14 @@ class NotAuthorized(ActionError):
     '''
     pass

+class NotAuthenticated(ActionError):
+    '''Exception raised when the user is not authorized to call the action.
+
+    For example :py:func:`~ckan.logic.action.create.package_create` raises
+    :py:exc:`~ckan.plugins.toolkit.NotAuthorized` if the user is not authorized
+    to create packages.
+    '''
+    pass

 class ValidationError(ActionError):
     '''Exception raised by action functions when validating their given
diff --git a/ckan/logic/action/get.py b/ckan/logic/action/get.py
index 8d375e0..1cf2f60 100644
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -38,6 +38,7 @@ _validate = ckan.lib.navl.dictization_functions.validate
 _table_dictize = ckan.lib.dictization.table_dictize
 _check_access = logic.check_access
 NotFound = logic.NotFound
+NotAuthenticated = logic.NotAuthenticated
 ValidationError = logic.ValidationError
 _get_or_bust = logic.get_or_bust

@@ -1416,7 +1417,7 @@ def user_show(context, data_dict):
     elif provided_user:
         context['user_obj'] = user_obj = provided_user
     else:
-        raise NotFound
+        raise NotAuthenticated

     _check_access('user_show', context, data_dict)`

torfsen: I think `NotAuthenticated` is a good idea (just like HTTP which distinguishes between 401 and 403), but it would have to be implemented properly: The whole authentication work-flow around `ckan.logic.check_access` would have to be updated accordingly.

wardi:
I think we just need the tests updated (and maybe use the same "Unauthorized" text so we aren't adding a new translation string)

PR #3213, opened 2016-08-25
[#3196] i18n for Flask endpoints

amercader:
This is based off #3212, so you might want to review this first.

This PR includes changes to allow Flask endpoints to translate strings via [Flask-Babel](https://github.com/python-babel/flask-babel). It adds the Flask-babel requirement, hooks its middleware into the Flask stack and provides a wrapper in `ckan.common` for `_` and `ungettext`, the functions used for translations. There are also the usual cleanups and consolidations.

~~**Note:** This is marked as WIP because it relies on unpublished changes in Flask-babel regarding the support for different gettext domains: https://github.com/python-babel/flask-babel/pull/91 (cc @TkTech)~~

Issue #3083, opened 2016-06-02
url_for support for Flask / Pylons


amercader:
While the transition to Flask is ongoing we will have routes and controllers which are served by Pylons and routes and views which are served by Flask. We need to ensure that URLs can be generated in the same way regardless of the context of the request. Although similar, Pylons (or Routes) `url_for` and Flask don't accept the same parameters. There are different syntaxes supported on Pylons but basically it works like this:

``` python

    # Pylons
    url = url_for(controller='api', action='action', ver=3, qualified=True)

    # Flask
    url = url_for('api.action', ver=3, _external=True)

```

To complicate things a bit more, Flask insists on being under an [Application Context](http://flask.pocoo.org/docs/0.11/appcontext/) when generating URLs. The approach we've followed in our POC branches is the following:
1. **All** imports for `url_for` point to our wrapper on `ckan/lib/helpers.py`.
2. This `url_for` in helpers has been [modified](https://github.com/ckan/ckan/blob/5a13507426b794fbcc5c2a515c9b63e30d0d2891/ckan/lib/helpers.py#L179:L260) to support both Flask and Pylons parameters. Regardless of what parameters are passed, first we try the Flask url generation, and if it doesn't work we fall back to the Pylons one.
3. In order for this to work in the context of both applications, some [changes](https://github.com/ckan/ckan/blob/5a13507426b794fbcc5c2a515c9b63e30d0d2891/ckan/lib/helpers.py#L179:L260) are made at the top of the middleware stack just before dispatching the request to the relevant app. Essentially, if Flask is serving the request we create a `routes_request_config` object and attach the Routes mapper to it, and if it is a Pylons request, we wrap it with `flask_app.test_request_context(environ_overrides=environ)`. This allows both routers to work even in the context of a request being served by the other app.

These changes allow most of the existing code in core and extensions to work unchanged, except for the tests.

Test are going to need updating in order to use the combined `url_for` function. Essentially as we are outside the context of a web request, whenever we call `url_for` directly in the tests (or a function that triggers `url_for` later on) Flask will raise an exception:

```
RuntimeError: Attempted to generate a URL without the application context being pushed. This has to be executed when application context is available.
```

These calls need to be wrapped in the [`test_request_context`](http://flask.pocoo.org/docs/0.11/api/#flask.Flask.test_request_context) context manager:

``` python
    def test_atom_feed_page_negative_gives_error(self):
        group = factories.Group()

        app = self._get_test_app()
[...]

PR #3228, opened 2016-09-01
[#3196] Common URL generation for Flask and Pylons

amercader:
This is based off #3213 

The goal of this set of changes is to allow URLs for Flask and Pylons endpoints to be generated regardless of the app that is serving the request.

To do so we:
- provide the necessary environment on each app so the other app routing module can work (19f2be8)
- consolidate a single `url_for` function that first tries to generate a URL using the Flask router and falls back to the Pylons one if Flask doesn't know about that particular endpoint. The Flask router also supports (most of) the current Pylons syntax for backwards compatibility (b5ea2e7).
- fix a ton of tests :) As discussed in #3083 when calling `url_for` from the tests (directly or indirectly) we need to wrap the call in a `test_request_context()` context manager, as Flask requires an application context to be present. 

The bulk of this PR are changes in the imports and tests.


amercader:
Test failure unrelated

amercader:
@smotornyuk @tino097 @wardi @TkTech 

This one looks overwhelming but the actual significant changes are small. A better summary of the changes can be found in the description, but basically this includes:

* Changes in the environment so both Flask and Pylons routers can work regardless of who is serving the request (19f2be8e)
* A common `url_for` in helpers that works both in Flask and Pylons, which is mostly backwards compatible (https://github.com/ckan/ckan/commit/b5ea2e7ffa312534017275337c3d709d3a0a5e7b)
* Loads and loads of changes in tests to provide a `test_request_context()` when building a URL (directly o indirectly). See discussion in #3083.

Let me know if you have any comments

ckan/lib/helpers.py
```diff
@@ -200,43 +202,156 @@ def get_site_protocol_and_host():
...
-        kw['protocol'], kw['host'] = get_site_protocol_and_host()
-    my_url = _routes_default_url_for(*args, **kw)
+            raise Exception('API URLs must specify the version (eg ver=3)')
+
+    try:
+        # First try to build the URL with the Flask router, making a copy of
+        # the params in case they are modified
+        flask_args = tuple(args)
+        flask_kw = kw.copy()
+
```
smotornyuk: I'm not quite understand purpose of this. You are going to unpack this values on the next line, so this this two lines do nothing - original values are passed to function, right? You rather need `copy.deepcopy`. And is it ever possible, that arguments will be modified..?

amercader: @smotornyuk `args` and `kw` are the arguments and key words that `h.url_for()` gets. When passing them to `_url_for_flask()` they might get changed, eg trying to adapt them to what Flask actually understands from the Pylons-based syntax (eg [here](https://github.com/ckan/ckan/blob/0581d42a535e7c3bf0a0e9768b58d3be3db91ef6/ckan/lib/helpers.py#L298) and [here](https://github.com/ckan/ckan/blob/0581d42a535e7c3bf0a0e9768b58d3be3db91ef6/ckan/lib/helpers.py#L307), that's why I pass copies of them to `_url_for_flask`:

```python
     my_url = _url_for_flask(*flask_args, **flask_kw)
```

If for some reason Flask can't build the URL, `_url_for_routes` is called with the original args and kw. Does this make sense?

amercader: I understand now @smotornyuk! Fixed in 291a016

amercader:
@smotornyuk @wardi @TkTech can we get one final push and see if we can get this merged to progress on the Flask front? I believe that @smotornyuk has already reviewed it but I'd like to make sure that everyone is happy with the changes required in the tests.
I've written some documentation that describes them in https://github.com/ckan/ckan/pull/3723, which might help.

ckanext/example_iauthfunctions/tests/test_example_iauthfunctions.py
```diff
@@ -260,18 +272,33 @@ def test_group_create_with_curator(self):
...
+        # Make the TestApp that we'll use to simulate HTTP requests to
+        # CKAN.
+        cls.app = helpers._get_test_app()
 
     @classmethod
     def teardown_class(cls):
         ckan.plugins.unload('example_iauthfunctions_v3')
+        # Restore the config to its original values
+        config.clear()
+        config.update(cls._original_config)
```
wardi: can we use `FunctionalTestBase`'s `_load_plugins` class variable instead of this boilerplate?

amercader: I've rewritten all these tests, which where a mess. There is no more messing with the config (other than via the `change_config` decorator.

wardi:
LGTM

amercader:
woo!

Let's :point_down: press :point_down: the :point_down: button!

amercader:
Thanks @wardi! Can you merge #3723 if happy? I want to send an email to the list in case people find the `Runtime` exceptions working with master.

PR #3234, opened 2016-09-05
Fixes #3230

### LINKED ISSUE #3230: Reset link not including site_url

master

The reset email should include the complete link to reset a users password. e.g  http://mydomain.com/user/reset/<id>?key=<somekey>

The email only includes the subpath. e.g  /user/reset/<id>?key=<somekey>

Try to send a password recovery email.

I think the fix lies in https://github.com/ckan/ckan/blob/master/ckan/lib/mailer.py#L147 Shouldn't it use `ckan.site_url` instead of just `site_url`

mattfullerton:
Fixes #3230 by implementing suggestion from @torfsen
### Proposed fixes:

pass qualified=True to url_for 
### Features:
- [ ] includes API changes

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -141,7 +141,8 @@ def get_reset_link(user):
                    h.url_for(controller='user',
                              action='perform_reset',
                              id=user.id,
-                             key=user.reset_key))
+                             key=user.reset_key,
+                             qualified=True))
 
 
 def send_reset_link(user):


ckan/lib/mailer.py
```diff
@@ -141,7 +141,8 @@ def get_reset_link(user):
                    h.url_for(controller='user',
                              action='perform_reset',
                              id=user.id,
-                             key=user.reset_key))
+                             key=user.reset_key,
+                             qualified=True))
```
torfsen: Just passing `qualified=True` is not enough, since the code still joins the result with `site_url`. The latter is wrong and needs to be removed. All you should need is the result of `url_for` (with `qualified=True`).

mattfullerton: I'm afraid I don't understand. On my CKAN I went from a missing site_url in the link to having one, so it appears to work. I was only implementing the suggested solution and trying to feed back as a pull request.

amercader: @mattfullerton The `urljoin` function is joining `site_url` (which is the original bug reported, as it should be `ckan.site_url` ) with the result of `h.url_for`. If `url_for` is passed the `qualified=True` parameter, it will be constructed with the `site_url` already appended, so there is no need to join both. So we need to drop the urljoin and just return the output of `h.url_for`

amercader:
@mattfullerton do you want to revisit this one? Should be very easy to finish off

mattfullerton:
@amercader Yes, will do

mattfullerton:
@amercader This has already been fixed in https://github.com/ckan/ckan/pull/3486

PR #3237, opened 2016-09-08
Add architecture diagram to docs

wardi:
This diagram is based on one from https://delftswa.gitbooks.io/desosa2016/content/ckan/chapter.html and improvements by @jqnatividad 

Updates the surrounding text are coming.


wardi:
@amercader this is good for review now

Issue #3229, opened 2016-09-02
Migrate API controller to Flask blueprint


amercader:
First controller to be migrated on top of all the foundation work compiled in #3196

amercader:
Done in #3239

PR #3239, opened 2016-09-12
[#3229] API Blueprint

amercader:
Based off #3228

This is a genuine, organic, GM and Pylons free, 100% certified Flask-only API for CKAN. :cow: :tangerine: :white_check_mark: 

It adds a new blueprint with the relevant API view functions (`ckan/views/api.py`). There is nothing clever about its structure:
- Blueprint definition
- Private methods
- View functions
- URL rules

I'd love feedback about this as it will be replicated on all upcoming blueprints/controllers. I personally quite like having all routes definitions together at the end of the file rather than on decorators (specially if like in the REST ones you need to tweak them quite a lot).

There are a couple of minor endpoints that call the template rendering functions so the rendering logic on `ckan.lib.base` had to be modified so it worked on Flask. This needs to be addressed properly when
we migrate the first frontend controller, specially the logic around caching, but for now it works fine for the low-level template rendering done via the API.

While I was at it I also did some cleaning and removed a lot of endpoints. Here's what's kept and what's left (again please provide any feedback you might have):
#### Kept (migrated to Flask):
- Action API (`/api/action`) - Obvs
- Autocomplete endpoints - (`/api/util/{x}/autocomplete`) - Used in the frontend
- Snippet endpoint - (`/api/util/snippet`) - Used in the frontend tests
#### Deprecated (kept active in the Pylons controller, to be removed in the next CKAN version
- REST API (`/api/rest`) - REST wrapper around actions. Deprecated, but some third party apps still use it
- Search API (`/api/search/`) - Old, extensions can't hook into it
- "Munge" endpoints
#### Removed in this PR:
- `/tag_counts` - Unused and easy to obtain with package_search
- `/qos/throughput/` - Broken, returns exception
- `/util/is_slug_valid` - Does not respect validation
- `/util/resource/format_icon` - Unused, output broken
- `/util/markdown` - Unused
- `/util/status` - Unused, use `status_show` action


deinok:
What is the state of this?

TkTech:
This is a large PR that requires and deserves significant review. It will take awhile to do properly and as an additional set of eyes you are more than welcome to participate.

deinok:
I am not very experimentated with python, so my opinion should not be important, but im happy to drop pylons in favour of flask

amercader:
@TkTech @deinok This is final PR that aggregates all the previous individual ones related to the Flask migration so far. There is an overview here: https://github.com/ckan/ckan/issues/3196

TkTech:
@amercader This is going to be a pretty big release - can we take this time to finally kill off this iteration of the very-broken REST api? It's been deprecated for sufficiently long, I think. We have a lot of upcoming work on things like permission label support and implementing it twice is going to be a royal pain. Updating code to work with the supported API isn't a huge month-long task.

I would love a RESTful API, but it's probably best to just focus on the supported action api for now.

amercader:
@TkTech I'm +1 on removing it (it will also vastly simplify all the routing logic on the blueprint). Realistically since 2.0 came around version 3 of the API (action API) has been the recommended way and this has been clearly described in the [docs](http://docs.ckan.org/en/latest/api/index.html).

Given that it's a big change I want to make sure that everyone is happy with this @wardi @davidread (also @smotornyuk and @tino097)

If we go ahead we should announce it as soon as it gets merged.

Thanks, that's exactly the sort of discussion I wanted to get from these PRs.
What do you think of the general approach for the blueprints and handling of `url_for` (#3228)?

wardi:
I'd love to see the rest API removed. If someone really wants it they are welcome to write an extension.

I might be biased because I joined the CKAN team after the action API was introduced, but any code in core that isn't being actively kept up do date will continue to cause us problems.

davidread:
Yes, I'm all for removing code and this old rest api in principle.

My only slight concern is harvesting. The CKAN Harvester was 'only' switched to the new API in February 2016, but that's probably enough time. But we should flag this up in the changelog. https://github.com/ckan/ckanext-harvest/pull/158

I find it handy occasionally to use the REST API to look at the unvalidated package dict, but as you say, I should probably write an extension for that.

amercader:
Great, this will allow a great cleanup!

> My only slight concern is harvesting. The CKAN Harvester was 'only' switched to the new API in February 2016, but that's probably enough time. But we should flag this up in the changelog. ckan/ckanext-harvest#158

So that's one good reason to drop support for the REST API no? It means that harvest no longer relies on it. The only issue would be if harvest tried to harvest an instance running CKAN < 2.0 (no action API) but that's already an issue now.

> I find it handy occasionally to use the REST API to look at the unvalidated package dict, but as you say, I should probably write an extension for that.

Since 2.2 you can use `use_default_schema` on `package_show` to get exactly this, eg

https://data.gov.uk/api/action/package_show?id=road-accidents-safety-data&use_default_schema=True

davidread:
The thing about harvesting is that if someone upgrades their CKAN and loses the REST API, then someone harvesting them will probably need to upgrade their ckanharvester to be less than a year old. DGU will need some work, and possibly others too.

davidread:
@amercader cheers for the tip about use_default_schema!

amercader:
I see, this definitely needs to be announced as soon as it gets merged (hopefully soon!) but I think the transition on these cases would be smooth, at least when upgrading ckanext-harvest.

TkTech:
> My only slight concern is harvesting. The CKAN Harvester was 'only' switched to the new API in February 2016, but that's probably enough time. But we should flag this up in the changelog. ckan/ckanext-harvest#158

Out of curiosity, is this exposed to an end-user of harvest in any way, other than requiring a minimum CKAN version, or is the API being used transparent?

davidread:
@TkTech No, there's no exposure to admins or end users of what API version is being used, apart from if they read the code or logs (or if they show the harvest logs in the web interface).

amercader:
@TkTech OK, I've removed the REST endpoints from the new Flask-based Blueprint, but I haven't completely removed them from Pylons because we will need to remove dozens of dependant legacy tests and this PR is already too noisy.

I'll remove the REST API in a separate PR once this is merged.

amercader:
Speaking of which, what can we do to help you review this? Most of the changes and noise on this PR are changes on the tests caused by how the URL generation is handled in Flask (#3228) so perhaps it's worth reviewing that first. Whatever works best.

TkTech:
👍 I will leave this up until Saturday for any final comments, but otherwise we're good to go.

On Saturday I will be releasing a new flask-babel version so we'll want to version bump this as well.

tino097:
while scrolling trough user activity on the dashboard, following error occurs 
`2017-04-10 15:11:37,630 ERROR [ckan.config.middleware.flask_app] Exception on /api/action/user_activity_list_html [GET]
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/flask/app.py", line 1988, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/flask/app.py", line 1641, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/flask/app.py", line 1544, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/views/api.py", line 305, in action
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/logic/__init__.py", line 457, in wrapped
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/logic/action/get.py", line 2748, in user_activity_list_html
    context, activity_stream, extra_vars)
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/lib/activity_streams.py", line 245, in activity_list_to_html
    snippet = activity_snippet_functions[match](activity, detail)
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/lib/activity_streams.py", line 30, in get_snippet_dataset
    link = h.dataset_link(pkg_dict) if pkg_dict else ''
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/lib/helpers.py", line 1449, in dataset_link
    text = dataset_display_name(package_or_package_dict)
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/lib/helpers.py", line 1435, in dataset_display_name
    return get_translated(package_or_package_dict, 'title') or \
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/lib/helpers.py", line 2413, in get_translated
    language = i18n.get_lang()
  File "/home/ksivakov/ckan/lib/local/src/ckan/ckan/lib/i18n.py", line 276, in get_lang
    langs = i18n.get_lang()
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/pylons/i18n/translation.py", line 191, in get_lang
    return getattr(pylons.translator, 'pylons_lang', None)
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/paste/registry.py", line 137, in __getattr__
    return getattr(self._current_obj(), attr)
  File "/usr/lib/ckan/local/local/lib/python2.7/site-packages/paste/registry.py", line 197, in _current_obj
    'thread' % self.____name__)
TypeError: No object (name: translator) has been registered for this thread`

TkTech:
I guess we have tests for the main page but not the api action, we should fix that as well.

amercader:
@tino097 I believe 2d7c28b should fix the exception.

@TkTech is user_activity_list_html one of the API endpoints that will be gone eventually?

tino097:
@amercader yes, thats fixed now

TkTech:
@amercader it's completely removed, activities are just normal templates now. I hope to have it finished tomorrow.

tino097:
follow/unfollow is not working correctly. There is no error message but when i was debugging got unauthorized message

amercader:
@smotornyuk I added your repoze patch from #3762. That fixes the follow / unfollow issue that @tino097 raised. Basically API calls could not be identified via cookie. This is not covered by our tests and I was only focusing on authorization via Header (the most common when using the API). I'm not sure how to test it to be honest.

AFAIC this PR is ready for final reviewing and merging. We can polish it as we work on the home and admin controllers @smotornyuk @tino097 @wardi @TkTech :arrow_down:

amercader:
One thing to think about, is it worth settling on a pattern for defining the views in the blueprints? This PR uses just individual view functions. @tino097 and @smotornyuk Home and Admin blueprints (#3762 and #3775) use [pluggable views](http://flask.pocoo.org/docs/0.12/views/) (`View` and `MethodView` classes) or a mix of both.

Or should we use what works best on each case?

TBH I just used view functions here because it seemed simpler at the time but thinking about it some of them could benefit from using Views, eg all the autocomplete ones.

Again, we can improve it later.

amercader:
Give me an M!
Give me an E!
Give me an R!
Give me a G!
...

ckan/views/api.py
```diff
@@ -95,17 +105,413 @@ def _finish_ok(response_data=None,
...
+            for item in entity:
+                new_items.append(make_unicode(item))
+            return new_items
+        elif isinstance(entity, dict):
+            new_dict = {}
+            for key, val in entity.items():
+                new_dict[key] = make_unicode(val)
+            return new_dict
+        else:
+            return entity
```
wardi: This kind of recursive function is really slow over long lists and deeply nested data, why is it required?

ckanext/datastore/tests/test_search.py
```diff
@@ -928,10 +929,9 @@ def test_self_join(self):
             where a.author = b.author
             limit 2
             '''.format(self.data['resource_id'])
-        data = {'sql': query}
-        postparams = json.dumps(data)
+        data = urllib.urlencode({'sql': query})
```
wardi: why is this change required?

ckan/views/api.py
amercader: I ported this straight from the current API controller, as many did before me starting from this ancient commit 8c4d651addab960ffd50ec50deab7d5014abf6f2 in 2010 that seem to be related to the REST API.

I'm pretty sure that Flask [will take care](http://flask.pocoo.org/docs/0.12/unicode/) of all the unicode handle automatically so I'll just remove it.

ckanext/datastore/tests/test_search.py
amercader: I can't remember. Probably a change in how Flask encodes special characters in query args. I'll have a look.

amercader: Flask is a bit more picky than Pylons when parsing POST data (with good reason).

In these tests we are doing calls via the webtest app like this:

```python
        res = self.app.post('/api/action/datastore_search_sql', params=postparams,
                              extra_environ=auth
        )
```

Which by default uses `Content-type: 'multipart/form-data'`. But what we are actually sending is JSON (see removed lines).

So we either send Form data (with `urllib.urlencode()` like here) or we use the appropriate content type header:

```python
        res = self.app.post('/api/action/datastore_search_sql', params=postparams,
                              extra_environ=auth,
                              headers={'Content-type': 'application/json'})
```

amercader: And this particular test failed because of the equal sign in the params

wardi: thanks that makes sense! I was just concerned that we broke one of the ways of calling the api.

wardi:
LGTM

PR #3240, opened 2016-09-12
Issue 2882 activity time

### LINKED ISSUE #2882: Activity timestamps are stored in local time instead of UTC
https://github.com/ckan/ckan/blob/master/ckan/model/activity.py#L41

If server has local time set to some timezone that differs from UTC, activity streams show -1 days ago instead of just now, since datetime.now() return local time instead of UTC. There probably are others as well which use local time instead of UTC, at least based on a quick search https://github.com/ckan/ckan/search?utf8=%E2%9C%93&q=datetime.now%28%29&type=Code

jrods:
Fixes #2882 

### Proposed fixes:

Activity timestamps are now created with a UTC timestamp.

A migration script is also included to adjust all stored timestamps in the Activity table, based on using the server's set timezone.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/migration/versions/085_adjust_activity_timestamps.py
+++ b/ckan/migration/versions/085_adjust_activity_timestamps.py
@@ -5,7 +5,7 @@ import datetime
 
 def upgrade(migrate_engine):
     """
-    The script assumes that the current timestamp was 
+    The script assumes that the current timestamp was
     recorded with the server's current set timezone
     """
     utc_now = datetime.datetime.utcnow()
--- a/ckan/migration/versions/085_adjust_activity_timestamps.py
+++ b/ckan/migration/versions/085_adjust_activity_timestamps.py
@@ -2,9 +2,11 @@
 
 import datetime
 
+
 def upgrade(migrate_engine):
     """
-    The script assumes that the current timestamp was recorded with the server's current set timezone
+    The script assumes that the current timestamp was 
+    recorded with the server's current set timezone
     """
     utc_now = datetime.datetime.utcnow()
     local_now = datetime.datetime.now()
--- a/ckan/migration/versions/085_adjust_activity_timestamps.py
+++ b/ckan/migration/versions/085_adjust_activity_timestamps.py
@@ -1,33 +1,14 @@
 # encoding: utf-8
 
-import pytz
-import tzlocal
-import logging
-
-log = logging.getLogger(__name__)
+import datetime
 
 def upgrade(migrate_engine):
-"""
-The script assumes that the current timestamp was recorded with the server's current set timezone
-"""
-    local_tz = tzlocal.get_localzone()
+    """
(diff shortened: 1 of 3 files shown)


ckan/migration/versions/085_adjust_activity_timestamps.py
```diff
@@ -0,0 +1,33 @@
...
+
+log = logging.getLogger(__name__)
+
+def upgrade(migrate_engine):
+"""
+The script assumes that the current timestamp was recorded with the server's current set timezone
+"""
+    local_tz = tzlocal.get_localzone()
+
+    with migrate_engine.begin() as connection:
```
wardi: transactions in a migration. What a great idea. We need more of these :-)

ckan/migration/versions/085_adjust_activity_timestamps.py
```diff
@@ -0,0 +1,33 @@
...
+        sql = "select id, timestamp from activity"
+        results = connection.execute(sql)
+
+        log.info("Adjusting Activity timestamp, server's localtime zone: {tz}".format(tz=local_tz))
+
+        for row in results:
```
wardi: big old ckan instances might have monstrous numbers of activities. Can you issue the update as a single postgres command instead of looping in python?

jrods: I tried doing this all in sql, but one issue i ran into was getting the system's set timezone. Another problem is, it's been awhile since I've had to write some complex sql =p.

wardi: if the offset is e.g. one hour then this seems to do the trick
```sql
update activity set timestamp = timestamp + interval '1 hour';
```

wardi: even easier would be to let posgres do the math:
```sql
update activity set timestamp = timestamp + (? - ?);
```
and pass in a local time and a UTC version of the same time.

ckan/migration/versions/085_adjust_activity_timestamps.py
```diff
@@ -0,0 +1,16 @@
+# encoding: utf-8
+
+import datetime
+
+
+def upgrade(migrate_engine):
+    """
```
jrods: @wardi is it necessary to prefix this string literal with `u`? same with line 15? i ask because it fails in circleci

torfsen: Yes, you need to prefix all string literals with either `u` (if they contain Unicode) or `b` (if they contain raw bytes). In almost all cases, `u` is the correct choice. As you've already seen from the tests, this is also required for docstrings. See the discussion in #3049 for details.

wardi:
@jrods thanks! merged with a small change

PR #3241, opened 2016-09-13
RDF template returns server error if package has no extras

smotornyuk:
Fixes #

### Features:

- [ ] includes API changes

In very special situations, pkg_dict hasn't such key as 'extras'.
it results in server error, because rdf template uses `None` as
fallback value.

Replaced it with empty list - it's more logically


amercader:
Thanks @smotornyuk, backported to 2.3.5 and 2.4.4. Note that the core RDF output is incorrect and incomplete on these versions, but you can still use [ckanext-dcat](https://github.com/ckan/ckanext-dcat) as a drop in replacement for semantically correct RDF outputs.

PR #3244, opened 2016-09-20
Fix html rendering of apache and nginx config file paths.

rgommers:

Sphinx doesn't allow substitutions in verbatim (double backticks) blocks.
This caused the substitutions to not be expanded.  See
http://docs.ckan.org/en/latest/maintaining/installing/deployment.html
at the start of sections 6 and 8.

Alternatively, if it's preferred to keep the current markup, the substitution can be replaced by the full path (with a comment in ``conf.py``.


amercader:
Thanks for the patch @rgommers

PR #3247, opened 2016-09-20
[#3243] Fix problems with database usage in background workers

### LINKED ISSUE #3243: Multiple background jobs that access the database fail
(This problem was originally reported by @wardi in #3165)

Accessing the database from a background job succeeds once, but subsequent runs of the same job fail:
```
DatabaseError: (DatabaseError) SSL error: decryption failed or bad record mac
 ```

torfsen:
Fixes #3243.

RQ workers execute jobs in separate processes. However, SQLAlchemy engines cannot be shared between processes, so a new engine has to be initialized in each new process. This wasn't done before, resulting in various exceptions when workers accessed the database.

This PR fixes the problem by initializing a new SQLAlchemy engine in each new process directly after forking. This is slightly different from the usual suggestion of how to use SQLAlchemy in a multi-process setting, which is to call `engine.dispose` after the fork. The problem with `engine.dispose` is that while it works well in the forked child (the inherited connection pool is discarded, and a new connection is created when the DB is accessed for the first time afterwards) it doesn't work well in the parent, which is not notified when the child closes the inherited (and shared!) connection. This means that, from the parent's view, the connection to the DB simply was closed unexpectedly.

I therefore think that in our case -- where we want to have a working engine in both the parent and the child process -- the correct approach is to simply initialize a new engine in the child. However, I'm not expert w.r.t. SQLAlchemy so I would love to get some feedback on this

I've added test cases that check that database access works in the workers and the main process in general and that running a background job in the middle of an on-going transaction behaves as it should (worker views the DB like before the transaction, the transaction itself is not changed). I've also added a section about database access from background jobs to the documentation.

Also note that I've refactored the code that initializes the SQLAlchemy engine in `ckan.config.environment`. The old code explicitly checked for PostgreSQL, but since we only support PostgreSQL anyways I dropped that check.

(Note that this issue was not related to SSL as I had previously considered in #3243. The SSL error messages were simply caused by `engine.dispose` closing the connection without the parent's knowledge)

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -251,17 +251,9 @@ def update_config():
     # CONFIGURATION OPTIONS HERE (note: all config options will override
     # any Pylons config options)
 
-    # for postgresql we want to enforce utf-8
-    sqlalchemy_url = config.get('sqlalchemy.url', '')
-    if sqlalchemy_url.startswith('postgresql://'):
-        extras = {'client_encoding': 'utf8'}
-    else:
-        extras = {}
-
-    engine = sqlalchemy.engine_from_config(config, 'sqlalchemy.', **extras)
-
-    if not model.meta.engine:
-        model.init_model(engine)
+    # Initialize SQLAlchemy
+    engine = sqlalchemy.engine_from_config(config, client_encoding='utf8')
+    model.init_model(engine)
 
     for plugin in p.PluginImplementations(p.IConfigurable):
         plugin.configure(config)
--- a/ckan/lib/jobs.py
+++ b/ckan/lib/jobs.py
@@ -28,6 +28,7 @@ from rq.utils import ensure_list
 
 from ckan.lib.redis import connect_to_redis
 from ckan.common import config
+from ckan.config.environment import load_environment
 
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


torfsen:
After thinking about #3243 and #3249 some more I came to the conclusion that the best option is probably to just call `ckan.config.load_environment` after the fork. This will handle both the re-initialization of the SQLAlchemy engine and trigger a call to the `IConfigurable.configure` implementations. It is also the way we initialize the paster processes.

This required a small change to `load_environment`: Previously, `load_environment` did not re-initialize the SQLAlchemy engine if it had already been initialized. This worked well for the main process and for paster, since both of these start from a new process. Background worker processes, however, are forked from the main process and therefore inherit the environment, including the engine. Therefore `load_environment` now always re-initializes the engine.

The change in `load_environment` triggered a hidden bug in the test case `ckan.tests.config.test_environment.TestUpdateConfig.test_update_config_env_vars`, which changes several config options, including the PostgreSQL URL, to invalid values but had previously failed to correctly restore them. This has been fixed.

I've also updated the documentation of the situations in which `IConfigurable.configure` is called, which fixes #3249.

The remaining failing test cases seem to be unrelated to my changes, I think they belong to the new `IPermissionLabels` functionality. @wardi, could you please have a look at that?

wardi:
@torfsen please merge from master

PR #3252, opened 2016-09-20
Removing set pkg_dict in package/edit_base.html

jrods:
### Proposed fixes:

This was causing `pkg_dict` to be overwritten by `c.pkg_dict`, which isn't a set variable while rendering `package/new_resource.html`.

I did a quick search for any other instances of `package/edit_base.html` being extended; Those templates look like they're not using `pkg_dict`.
### Features:
- [ ] includes API changes

### DIFF
--- a/ckan/templates/package/edit_base.html
+++ b/ckan/templates/package/edit_base.html
@@ -1,7 +1,6 @@
 {% extends 'package/base.html' %}
 
 {% set pkg = c.pkg_dict %}
-{% set pkg_dict = c.pkg_dict %}
 
 {% block breadcrumb_content_selected %}{% endblock %}
 


wardi:
@jrods circleci doesn't seem to be building your PR. Try enabling circleci on your fork from circleci.com and then merge from master to get the latest test fix

jrods:
@wardi The test that is failing in circleci, you already fixed here: e0fdb8dcd279857890a301f90d82d18293983a32

jrods:
@wardi I did a quick refactor of the other pkg = c.pkg_dict, so far no problems in the web ui

wardi:
This is a fix for pkg_dict gettting overwritten in the new_resource case, right?  It's probably better to make sure that pkg_dict is being explicitly passed to the templates that require it. I really don't like *sometimes* having a value passed my mutable global state and other times as a normal variable.

This change has the possibility of breaking extensions that have extended this template and copied parts, or just extended this template or another child of it and relied on these variables being present. If we remove those set statements has you have done but just always pass and use the pkg_dict variable from our controllers to the templates we'll have simpler code that shouldn't break existing extensions.

jrods:
@wardi I'm running into this issue again with ckan v2.7 , but now `c.pkg_dict` doesn't exist for the `new_resource` view.

Removing the `c.pkg_dict` var assignments in `package/edit_base.html` to just `{% set pkg = pkg_dict %}` fixes the issue.

wardi:
@jrods so this no longer applies IIUC, would you submit a new PR? I'll close this one now.

boykoc:
In [2.8.2](https://github.com/ckan/ckan/blob/8ed0534f00b2b25f9e8df1962f821b9a9a979479/ckan/templates/package/edit_base.html#L3) this is still an issue and @wardi's comment does fix this.

I've noticed in current master branch that the setting of vars was completely removed in [PR4359](https://github.com/ckan/ckan/pull/4359/files#diff-04ab76c3a054d2522d26317f64addb7cL3). May cause issues but I haven't checked yet.

boykoc:
Correction.  Quick test and it seems fine. I'm guessing because the template isn't overriding `pkg_dict` it's being passed as part of extra_vars properly. 👍

PR #3258, opened 2016-09-27
[#2345] datastore: only create _table_metadata view if it does not exist yet

### LINKED ISSUE #2345: Datastore _table_metadata view recreated on every startup
We fixed the read/write test having conflicting table names by randomizing them (https://github.com/ckan/ckan/pull/1397, https://github.com/ckan/ckan/issues/2082) but there's still another database call that can cause problems on busy sites: creating the _table_metadata view.
https://github.com/ckan/ckan/blob/b1c12f74b09c4703407563861e089e37ec0e19e2/ckanext/datastore/plugin.py#L80

My understanding of PostgreSQL views was once it's created, they update automatically without any further interaction. And if it's being recreated, all read queries against that view are locked until it is done?

Does this view really need to be created and/or recreated every time the plugin is loaded? 
Could it be done once on installation (either with documentation for a paster command or checking that it does not exist/is not the right structure before proceeding)?

fanderegg:
Fixes #2345
### Features:
- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -120,7 +120,6 @@ class DatastorePlugin(p.SingletonPlugin):
                      'of _table_metadata are skipped.')
         else:
             self._check_urls_and_permissions()
-            self._create_alias_table()
 
     def notify(self, entity, operation=None):
         if not isinstance(entity, model.Package) or self.legacy_mode:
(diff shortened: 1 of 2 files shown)


wardi:
@fanderegg thanks, this shrinks the race condition we have when a number of workers are starting at the same time, but there's still a race. Would you mind moving the view-creation code to https://github.com/ckan/ckan/blob/master/ckanext/datastore/set_permissions.sql instead? That should completely eliminate the race.

fanderegg:
@wardi I updated the pull request as you suggested.

fanderegg:
Ok that wasn't as easy as i thought. I will check the errors and update the pull request.

fanderegg:
@wardi Now all tests pass.

fanderegg:
@wardi Any updates on this?

amercader:
This is of course the right approach but is there a chance that people with an existing DataStore database will end up with no `_table_metadata` view? I guess that then they need to run the set permissions command to create the view. This sh be mentioned in the Change log.

wardi:
I think the only way a user would end up with a missing `_table_metadata` is if they use an old version of the set permissions script, then upgrade ckan before ever enabling the datastore extension, then enable the datastore extension, right?

Issue #2346, opened 2015-03-13
dashboard_activity_list_html() doesn't pass the user id to activity_list_to_html()


clementmouchet:
The load more functionality in the dashboard doesn't work, because the `id` that is used to query more data is not passed to the template.

I've fixed it by passing the context user id:

``` diff
diff --git a/ckan/logic/action/get.py b/ckan/logic/action/get.py
index ad59266..2519744 100644
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py

@@ -3216,8 +3221,6 @@ def dashboard_activity_list_html(context, data_dict):
     The activity stream is rendered as a snippet of HTML meant to be included
     in an HTML page, i.e. it doesn't have any HTML header or footer.

-    :param id: the id or name of the user
-    :type id: string
     :param offset: where to start getting activity items from
         (optional, default: 0)
     :type offset: int
@@ -3233,6 +3236,7 @@ def dashboard_activity_list_html(context, data_dict):
     model = context['model']
     offset = data_dict.get('offset', 0)
     extra_vars = {
+        'id': model.User.get(context['user']).id,
         'controller': 'user',
         'action': 'dashboard',
         'offset': offset,
```

Will submit a PR asap

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #3269, opened 2016-10-12
[#2346] Dashboard_activity_list_html() dosen't pass user id

Engerrs:
Fixes #

Load more don't work on Dashboard page, because dashboard_activity_list_html() don't pass id as an argument, so in activity-stream.js the user_activity_list_html() gives us an error.

### Proposed fixes:

Fix it by passing id from context as we do with model= context['model']. As a result context['user'] gives us the name of current user and we can use it, because user_activity_list_html() can take both id and username as id.

### Features:

- [X] includes API changes


amercader:
Thanks for the patch @Engerrs! Merged and backported

Issue #3270, opened 2016-10-12
ne translations have been merged incorrectly


torfsen:
It seems that [`ckan/i18n/ne/LC_MESSAGES/ckan.po`](https://github.com/ckan/ckan/blob/dbf5173c1b8c2e64048435de35d190e47441ac29/ckan/i18n/ne/LC_MESSAGES/ckan.po) has been merged incorrectly, the file still contains Git conflict markers. The offending commit seems to be 8f38725138e629c22a6d2170b7cb00c9e778f66f by @amercader. As far as I can tell the other files in that commit are fine.

amercader:
PR is here: https://github.com/ckan/ckan/pull/3246 @TkTech maybe you can push the button on that one?

torfsen:
@amercader Must have missed that one, sorry for the noise!

PR #3246, opened 2016-09-20
Fix NE locale, which was broken in a bad merge

amercader:
Fix the po file and recompile the mo file. @LinkSDM this needs
to be cherry-picked to the dev-v2.6 branch and perhaps fixed in
Transifex too

PR #3272, opened 2016-10-12
Support for JS translations in extensions

torfsen:
This PR is a WIP to support JavaScript translations in extensions (ckan/ideas-and-roadmap#176).

Currently, CKAN's translation files for JavaScript are generated "manually" using `paster trans js`. The generated files take only the translations from core CKAN into account.

This PR introduces the following changes:
- Translations from extensions that implement `ITranslation` are taken into account. Hence, extensions may override CKAN translations used in JS or introduce their own translatable strings in JS.
- Translations are only (re-)generated if they're out-of-date, i.e. if there's a PO-file for the language that's newer than the generated JS file.
- Translations are re-generated during server startup (if necessary).

With these changes, translations in extensions work transparently for both Jinja and JS. No changes to the extensions are necessary. Running `paster trans js` isn't required anymore, but the command has been kept in case one wants to work on translations without constantly restarting the server.

I'd love to get some feedback on that approach. Once there's a consensus I'll add the missing tests and update the documentation.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -22,6 +22,7 @@ import ckan.lib.search as search
 import ckan.logic as logic
 import ckan.authz as authz
 import ckan.lib.jinja_extensions as jinja_extensions
+from ckan.lib.i18n import build_js_translations
 
 from ckan.common import _, ungettext, config
 from ckan.exceptions import CkanConfigurationException
@@ -102,6 +103,11 @@ def load_environment(global_conf, app_conf):
     p.load_all()
 
     app_globals.reset()
+
+    # Build JavaScript translations. Must be done after plugins have
+    # been loaded.
+    build_js_translations()
+
 
 # A mapping of config settings that can be overridden by env vars.
 # Note: Do not remove the following lines, they are used in the docs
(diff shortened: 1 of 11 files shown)


torfsen:
Note: The failing test is due to #3270 which will be fixed by #3246.

torfsen:
@TkTech Some feedback from you on this would be much appreciated.

amercader:
@torfsen this looks great. Happy to review it fully and merge it if you sort tests and docs. Cheers

torfsen:
@amercader I've added tests and updated the documentation so this is ready to go from my point of view.

I didn't find an established convention of how to bundle extra files with test cases, so I created the subdirectory `ckan/tests/lib/_i18n_build_js_translations` for them. Let me know if there's a better place.

amercader:
This all looks great @torfsen. There is already an example `ITranslation` plugin with tests on `ckanext/example_itranslation`. Any reason why we can add the new tests there?

torfsen:
@amercader Only some of the tests target `ITranslation`, so I'd rather keep most of them where they currently are. I could re-use the `example_itranslation` plugin for the `ITranslation`-related tests, but then I would have to include the necessary edge case translation strings in its `.pot` and `.po` files, where they might confuse people who just want to learn from the example.

Hence, in my opinion it's cleaner to keep the tests as they are, but at the end it's probably a matter of taste. Let me know if you'd rather have them as part of `example_itranslation`.

amercader:
Thanks @torfsen !

PR #3286, opened 2016-10-24
[#1517] Set default file size during resource upload instead of empty size

### LINKED ISSUE #1517: Empty mime type and size for uploaded files

gleb-rudenko:
Fixes #1517

### Proposed fixes:
Default file size is added to the data_dict (if no user file size provided from API ) from the file_upload object that is provided by ResourceUpload Class, during resource_create action.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -296,7 +296,7 @@ def resource_create(context, data_dict):
 
     upload = uploader.get_resource_uploader(data_dict)
 
-    if not 'size' in data_dict and hasattr(upload, 'upload_file'):
+    if 'size' not in data_dict and hasattr(upload, 'upload_file'):
         file_object = upload.upload_file
         file_object.seek(0, 2)
         data_dict['size'] = file_object.tell()
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -296,6 +296,12 @@ def resource_create(context, data_dict):
 
     upload = uploader.get_resource_uploader(data_dict)
 
+    if not 'size' in data_dict and hasattr(upload, 'upload_file'):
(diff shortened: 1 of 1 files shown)


ckan/logic/action/create.py
```diff
@@ -296,6 +296,12 @@ def resource_create(context, data_dict):
 
     upload = uploader.get_resource_uploader(data_dict)
 
+    if not 'size' in data_dict and hasattr(upload, 'upload_file'):
```
TkTech: `if 'size' not in data_dict...`

gleb-rudenko: @TkTech , I tried to follow the same code style as in the CKAN code.
From the same method above my code:
`if not 'resources' in pkg_dict:`

gleb-rudenko: @TkTech , if you still want me to change it, let me know please and I will.

TkTech: Yes, please. Your IDE/linter should be yelling at you to fix it :) At lot of the old CKAN code does not follow PEP8 and it's cleaned up as we go.

gleb-rudenko:
@TkTech , done, next time I'll know 👍

amercader:
Thanks @gleb-rudenko, there's another PR which is almost done and which among other things has the same functionality: https://github.com/ckan/ckan/pull/3205. The approach seems very similar and it has tests so it probably makes sense to merge that one. Maybe you could check it to review the approach.

Does that sound good?

gleb-rudenko:
@amercader , I've reviewed the approach, it looks good. I agree with merging it instead of mine.
Left a comment there about PEP8 standards.

PR #3288, opened 2016-10-27
[#2649] Fix resource extras being lost during resource edit

torfsen:
This is a partial fix for the issues regarding resource extras being lost during resource edits (#2649). In particular, this PR fixes the problem in the case of existing resources being edited via the web UI. That sub-problem was caused by not storing the extras in the edit form and then calling `resource_update` with the POSTed data. This PR changes that behavior to use `resource_patch` instead.

This PR *does not fix* the related issue also described in #2649 by @wardi where extras from custom schemas are dropped during the editing of resource drafts.

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -554,7 +554,7 @@ class PackageController(base.BaseController):
                 clean_dict(dict_fns.unflatten(tuplize_dict(parse_params(
                                                            request.POST))))
             # we don't want to include save as it is part of the form
-            del data['save']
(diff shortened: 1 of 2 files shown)


torfsen:
This was discussed in today's CKAN dev meeting: We agreed that the approach used in this PR is a bit fragile. In addition, @TkTech is currently working on an improved system for extras (#3072). The consensus was to try to use that to fix the extra-issues in general instead of fixing all small problems one-by-one.

PR #3295, opened 2016-11-02
Solr Dockerfile Update

deinok:
Upgrade Solr Version 6.0 -> 6.2

Schema.xml is getted from the master (SysLinks not work in docker)

### DIFF
--- a/contrib/docker/solr/Dockerfile
+++ b/contrib/docker/solr/Dockerfile
@@ -1,4 +1,4 @@
-FROM solr:6.0
+FROM solr:6.2
 MAINTAINER Open Knowledge
 
 # Enviroment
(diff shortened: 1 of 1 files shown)


contrib/docker/solr/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN mkdir -p /opt/solr/server/solr/$SOLR_CORE/data
 
 # Adding Files
 ADD ./solrconfig.xml \
-https://raw.githubusercontent.com/ckan/ckan/ckan-2.5.2/ckan/config/solr/schema.xml \
+https://raw.githubusercontent.com/ckan/ckan/master/ckan/config/solr/schema.xml \
```
wardi: so how do we fix this? Moving the solr schema under contrib/docker is pretty terrible.

How about using make to build our docker images? make doesn't have silly limitations like this.

wardi: or a simpler idea: change this to /dev-v2.6/ instead of /master/.  We don't change the solr schema very often and that will keep old versions of this code building the right solr image.

deinok: [https://cwiki.apache.org/confluence/display/solr/Schema+Factory+Definition+in+SolrConfig#SchemaFactoryDefinitioninSolrConfig-Switchingfromschema.xmltoManagedSchema](https://cwiki.apache.org/confluence/display/solr/Schema+Factory+Definition+in+SolrConfig#SchemaFactoryDefinitioninSolrConfig-Switchingfromschema.xmltoManagedSchema)

This don't solve the problem at all, but probably will help a lot.

Should i send a PR with this?

wardi: If we can get ckan using the ManagedSchema API we don't need the schema.xml in our solr image and users can start solr without additional configuration (a big chunk of our docs are devoted to this) but, it's a fair bit of work because some sites extend the schema in their own ways, so we might need a new interface to let plugins configure solr.

PR #3296, opened 2016-11-02
Docker Postgres Update

deinok:
Fix a bug with password of the Postgres Instance
Upgrades the postgres image from 9.5 to 9.6


deinok:
@wardi Any Update about this?

Issue #3271, opened 2016-10-12
make facets sortable by function parameter


hbunke:
Sorting of facet items is hardcoded (in `lib/helpers/get_facet_items_dict()`). However, there might be usecases where this is not the desired sorting and sorting key should be the item title only (in our case this would be a journal's volume or issue number).

Presumably it should be no problem to make the sorting key a parameter for `get_facet_items_dict()` (with the current behavior as default). 

Any thoughts? I'd try to submit a PR if no one objects.

amercader:
Sounds good, as long as you can keep the default behaviour so extensions don't break. Feel free to submit a PR for this

hbunke:
See #3298. For now it's a very conservative change that enables reverse sorting by name for int type values only. I'm not sure if sorting by name would make sense for other values.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #3298, opened 2016-11-03
[#3271] make facets sortable by name

hbunke:
### Features:

- [X] includes user-visible changes (optionally)
- [ ] includes API changes

Sorting of facet items is currently hardcoded. However, there might be
usecases where this is not the desired sorting and sorting key should be the
item title only.

This PR introduces a int_sort parameter for helpers.get_facet_items_dict(). It's
a very conservative change that enables sorting by name for facets with ints
(real or literal) only, assuming that this is the only usecase in which
descending sorting by name might be desirable.

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -708,7 +708,8 @@ def default_group_type():
 
 
 @core_helper
-def get_facet_items_dict(facet, limit=None, exclude_active=False, int_sort=False):
+def get_facet_items_dict(facet, limit=None, exclude_active=False,
+                         int_sort=False):
     '''Return the list of unselected facet items for the given facet, sorted
     by count (default) or name.
 
@@ -741,7 +742,8 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False, int_sort=False
             all_ints = all(map(isint, names))
             if all_ints:
                 return sorted(f, key=lambda it: -int(it['display_name']))
-        return sorted(f, key=lambda it: (-it['count'], it['display_name'].lower()))
+        return sorted(f, key=lambda it: (-it['count'],
+                      it['display_name'].lower()))
 
     try:
         facet_items = c.search_facets.get(facet)['items']
@@ -762,6 +764,7 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False, int_sort=False
     # zero treated as infinite for hysterical raisins
     if limit is not None and limit > 0:
         return sort(facets)[:limit]
+
     return sort(facets)
 
 
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -727,7 +727,6 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False,
     limit -- the max. number of facet items to return.
     exclude_active -- only return unselected facets.
     int_sort -- sort facets with int literals in 'display_name' numerically
-        and descending
     '''
 
     def sort(f):
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -708,14 +708,14 @@ def default_group_type():
 
 
 @core_helper
-def get_facet_items_dict(facet, limit=None, exclude_active=False):
+def get_facet_items_dict(facet, limit=None, exclude_active=False, int_sort=False):
     '''Return the list of unselected facet items for the given facet, sorted
-    by count.
+    by count (default) or name.
 
     Returns the list of unselected facet contraints or facet items (e.g. tag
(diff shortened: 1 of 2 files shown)


ckan/lib/helpers.py
```diff
@@ -725,28 +726,46 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
...
+    int_sort -- sort facets with int literals in 'name' by name and descending
     '''
-    if not c.search_facets or \
-            not c.search_facets.get(facet) or \
-            not c.search_facets.get(facet).get('items'):
+
+    def isint(v):
+        try:
+            i = int(v)
+        except:
```
torfsen: Please only catch the appropriate exception (in this case it's `ValueError`).

ckan/lib/helpers.py
```diff
@@ -725,28 +726,46 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
...
+            i = v
+        return isinstance(i, int)
```
torfsen: You don't need a separate call to `isinstance`: You can return `True` if `int(v)` doesn't raise an exception and `False` otherwise.

ckan/lib/helpers.py
```diff
@@ -725,28 +726,46 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
...
+
+    def sort(f):
+        if int_sort:
+            names = map(lambda it: it['display_name'], f)
```
torfsen: Nowadays list comprehensions and generator expressions are usually considered more Pythonic than `map`:

    names = [it['display_name'] for it in f]
    all_ints = all(isint(n) for n in names)

ckan/lib/helpers.py
```diff
@@ -725,28 +726,46 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
...
+            all_ints = all(map(isint, names))
+            if all_ints:
+                return sorted(f, key=lambda it: -int(it['display_name']))
```
torfsen: You're first converting everything to `int` to see if it's an integer and then discard these integers only to compute them again later on. Why not do the conversion once (if `int_sort` is true) and use it for sorting, falling back to the default sort order if there's an error during conversion?

torfsen:
I like the idea, but would prefer to have a single `order` parameter that can take multiple values, for example `count` to sort the facets by count (current default) or `title` to sort them alphabetically. While we don't need to implement all of the sort orders right now it would be easier to add another one at a later point (compared to having multiple `sort_count`, `sort_title`, etc. parameters).

Regarding sorting by the title's numerical value, I suggest support for the so-called [natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order) instead, since it achieves the same effect for integer titles but also works for mixed letter/digit titles. See, for example, the [`natsort` package](https://github.com/SethMMorton/natsort).

ckan/lib/helpers.py
hbunke: You're right. I tend to be a bit lazy about exceptions.

ckan/lib/helpers.py
hbunke: I know, but personally I prefer `map` (more functional, better composable). But in this case it really doesn't matter, so I'll change that.

hbunke:
Having a single parameter for `order` has also been my first idea. I just wasn't sure --as explained in the PR-- if it makes sense at all to sort facets by title (except for ints). Discussion point.

Will have a look at `natsort`.

ckan/lib/helpers.py
torfsen: Something like this:
```python
def sort(f):
    if int_sort:
        # Sort numerically
        try:
           return sorted(f, key=lambda it: -int(it['display_name']))
        except ValueError:
            # Some of the display names could not be converted to ints,
            # fall back to standard sort
            pass
    # Sort alphabetically
    return sorted(f, key=lambda it: (-it['count'], it['display_name'].lower()))
```

hbunke: Uh, much better and more concise, thx! Seems like I've been thinking too complicated here.

wardi:
Sorting numeric tag names is a pretty site-specific thing. How about just overriding this helper in your own exension, or as @torfsen suggested make this sort-overriding more generic.

hbunke:
Thanks, Ian. Actually I have already overwritten this in my extension, before submitting the PR, and in deed I tried to keep it more generic there ;-) (https://github.com/hbunke/ckanext-edawax/blob/master/ckanext/edawax/plugin.py#L33-L70 ). So, I'll be glad to modify the PR accordingly as soon as I find some time.

wardi:
@hbunke thanks, we can do that in a new PR.

PR #3301, opened 2016-11-04
[#3292] iUploader interface two attributes were added to documentation

### LINKED ISSUE #3292: Create a file filesize function as a part of iUploader interface
This PR was merge https://github.com/ckan/ckan/pull/3205 recently, and it works great!
The only problem is that other extensions that override the default upload functionality don't have those alts, and it would be cool to implement a function that would be a part of an interface so that further extensions development will require this function.

gleb-rudenko:
Fixes #3292
Added two new attributes description into the iUploader interface.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -1618,6 +1618,10 @@ class IUploader(Interface):
         :param resource: resource dict
         :type resource: dictionary
 
+        Optional attributes:
+            filesize (int):  Uploaded file filesize.
+            mimetype (str):  Uploaded file mimetype.
+
         ``upload(id, max_size)``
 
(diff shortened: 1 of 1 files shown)


ckan/plugins/interfaces.py
```diff
@@ -1618,6 +1618,10 @@ def get_resource_uploader(self):
         :param resource: resource dict
         :type resource: dictionary
 
+        Optional attributes:
```
amercader: Can you expand this a bit more to make it more clear to people less familiar with how it works? Something like:

"Optionally, this method can set the following two attributes on the class instance so they are set in the resource object:"

PR #3305, opened 2016-11-05
Requirements.in Update

deinok:
Updates Libraries that doesn't need any change


deinok:
@amercader Any problem whit this, this are the libraries that don't require any change to be upgraded

amercader:
Thanks for the patch @deinok

PR #3307, opened 2016-11-08
suppression of the id argument on group_list api requests

datalocale:
Fixes #

### Proposed fixes:

suppression of the id argument

### Features:

- [x ] includes API changes

Suppression of the id=data-explorer on the httpie and the python script examples


amercader:
Thanks @datalocale

PR #3311, opened 2016-11-13
Add setup.py classifiers

sirex:
It took me some time to find what Python versions are supported by CKAN, to fix that added classifiers to the setup.py.


amercader:
Thanks @sirex, great idea

PR #3318, opened 2016-11-18
[#3317] Don't track automatically generated JS translation files in git

### LINKED ISSUE #3317: Don't track generated JS translation files in git
The git repository contains stubs for the automatically generated JS translation files. Since #3272 has been merged, these are automatically overwritten with the full translations during server startup and therefore show up as modified files in the repository. That's rather annoying when doing git stuff.

Automatically generated content should usually not be tracked in a VCS, so I suggest we ignore these files.

torfsen:
Fixes #3317.


TkTech:
👍

PR #3320, opened 2016-11-18
fix KeyError: 'content-type'

nbtetreault:
Fixes #

### Proposed fixes:

Prevent keyerror by using get() instead of accesing dict index directly

### Features:

- [ ] includes API changes

For an unknow reason "KeyError: 'content-type'" is triggered when content-type is not set. It is triggered or not depending on error level in configuration

### DIFF
--- a/ckanext/resourceproxy/controller.py
+++ b/ckanext/resourceproxy/controller.py
@@ -58,7 +58,7 @@ def proxy_resource(context, data_dict):
         if not did_get:
             r = requests.get(url, stream=True)
 
-        base.response.content_type = r.headers['content-type']
+        base.response.content_type = r.headers.get('content-type', '')
         base.response.charset = r.encoding
 
         length = 0


ckanext/resourceproxy/controller.py
```diff
@@ -58,7 +58,7 @@ def proxy_resource(context, data_dict):
         if not did_get:
             r = requests.get(url, stream=True)
 
-        base.response.content_type = r.headers['content-type']
+        base.response.content_type = r.headers.get('content-type', '')
```
nbtetreault: I agree about **None**.

nbtetreault: I did some test and it looks like my version of [Requests](http://docs.python-requests.org/en/master/) is out of date. By updating it I fixed the problem.

Sorry about that. I guess there is no need to fix then.

PR #3322, opened 2016-11-21
Optimize Docker image build to minimize its size

PeterDaveHello:
Combine apt-get related commands in a single RUN statements,
to prevent some additional but useless layers created during the build.

Size difference:
```
$ docker images
REPOSITORY TAG    IMAGE ID CREATED         SIZE
after      <none> 84cac1   9 minutes ago   629 MB
before     <none> ff3e9b   32 minutes ago  651.5 MB
```

Reference:
  Best practices for writing Dockerfiles
  https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,8 @@ ENV CKAN_CONFIG /etc/ckan/default
 ENV CKAN_DATA /var/lib/ckan
 
 # Install required packages
-RUN apt-get -q -y update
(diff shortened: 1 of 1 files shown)


TkTech:
Interesting, I guess there's no way for docker to understand what's being run so it simply deltas after each command.

👍

PR #3323, opened 2016-11-21
Don't crash if PO files aren't exactly where we expected them when ge…

TkTech:
Simple fix for a bug introduced in #3272 that caused a crash on startup if a PO file was not found.

@torfsen

### DIFF
--- a/ckan/lib/i18n.py
+++ b/ckan/lib/i18n.py
@@ -381,10 +381,20 @@ def build_js_translations():
 
     # Build translations for each language
     for lang in sorted(langs):
-        po_files = [os.path.join(i18n_dir, lang, u'LC_MESSAGES',
-                                 domain + u'.po')
-                    for i18n_dir, domain in i18n_dirs.iteritems()]
-        po_files = [fn for fn in po_files if os.path.isfile(fn)]
+        po_files = [
+            fn for fn in (
+                os.path.join(
+                    i18n_dir,
+                    lang,
+                    u'LC_MESSAGES',
+                    domain + u'.po'
+                )
+                for i18n_dir, domain in i18n_dirs.iteritems()
+            ) if os.path.isfile(fn)
+        ]
+        if not po_files:
+            continue
+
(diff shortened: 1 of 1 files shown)


torfsen:
No `.po`-files? I admit that I didn't think about that. In what situations could that occur?

ckan/lib/i18n.py
```diff
@@ -381,10 +381,20 @@ def build_js_translations():
...
+            fn for fn in (
+                os.path.join(
+                    i18n_dir,
+                    lang,
+                    u'LC_MESSAGES',
+                    domain + u'.po'
+                )
+                for i18n_dir, domain in i18n_dirs.iteritems()
+            ) if os.path.isfile(fn)
+        ]
```
torfsen: In my opinion that is far less readable than my version, but I guess you wouldn't have changed it if you considered mine readable :laughing:

wardi:
We ran into this on our fork when we backported the change that requires a translation even for English (likely because we didn't do the backport correcly) but the error that was generated was pretty confusing.

@TkTech should we improve the error instead of silently continuing?

ckan/lib/i18n.py
wardi: @torfsen I'm with you, but @TkTech has some pretty strongly-held beliefs around coding style :grinning:

TkTech:
@wardi This is not the same error as what we ran into - this specific fix is for a plugin that doesn't follow the path assumptions made by Torfsen's work because it does some experimental stuff with babel. CKAN runs just fine.

ckan/lib/i18n.py
TkTech: @torfsen Sorry, bit of a whitespace/lisp-y fan. I can revert it if you like but the first one should still be a generator instead of a list.

torfsen: @TkTech No, keep the changes. And you're obviously 100% correct regarding the generator.

Issue #3245, opened 2016-09-20
datastore_active is being rewritten due to race condition


ZtF:
### CKAN Version if known (or site URL)

2.5.2
### Please describe the expected behaviour

When file resource saved to FileStore, datapusher is notified about it, downloads the resource, process the file, create new table in DataStore and sets datastore_active flag for the particular resource. 
### Please describe the actual behaviour

It's working as described, though when uploading multiple resources (eg. through API), there is a race condition probably in `ckan.lib.dictization.model_save.package_dict_save` (or somewhere in the stack trace).
Sometimes the problem can even produce "Resource not found" in logs though it's most certainly there at the time. That's a case where datapusher tries to patch the resource and set datastore_active to True.
### What steps can be taken to reproduce the issue?

In httpd conf file, the race condition can be "resolved" with forcing one process and one thread for CKAN.
Something like this in ckan httpd config file:
`
WSGIDaemonProcess ckan display-name=ckan processes=1 threads=1
`
If it's eg. 15 threads and 2 processes, the described issue appears.

torfsen:
This might be the same issue as ckan/datapusher#100.

wardi:
A nice solution would be to make package_update atomic, but that needs a new error class (and API version) for conflicts that should be retried.

A less-nice solutions is making sure datapusher only updates one resource at a time per dataset. Not a perfect solution, but will be much less likely to fail.

A quick hack would be to sneak the datastore_active change past package_patch/package_update so it doesn't update all the other dataset metadata fields, then make sure to re-index the dataset.

amercader:
> to sneak the datastore_active change past package_patch/package_update

@wardi what do you mean by this?

wardi:
@amercader package_patch calls package_show+package_update to change the values passed. When the datastore just needs to set datastore_active=True on a resouce that's a lot of extra work and a much larger chance of conflict with action in progress.

Skipping all the validation and work done by package_update by changing the model directly would be a layering violation, but I could see the argument for it in this case, and it's unlikely to harm any real ckan instances. We are already doing something like this in our bulk actions that set the state to public/private.

amercader:
I see, sounds very sensible. +1

wardi:
I might not get to this very quickly, so if someone wants to submit a PR, here is the code that bulk updates the dataset private flag without going through package_update: https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L1138-L1149

[...]

PR #3325, opened 2016-11-22
#3245 Atomic actions

smotornyuk:
Added new decorator `ckan.logic.atomic_action` in order to solve race condition in update actions

### DIFF
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -5,6 +5,9 @@ import logging
 import re
 import sys
 
+from threading import Condition
+from collections import Iterable
+
 import formencode.validators
 
 import ckan.model as model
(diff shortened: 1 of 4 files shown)


smotornyuk:
Solution for [#3245](https://github.com/ckan/ckan/issues/3245)

ckan/logic/__init__.py
```diff
@@ -695,3 +701,63 @@ def _import_module_functions(module_path):
...
+                    if flag:
+                        break
+            else:
+                flag = data_dict.get(param)
+
+            if not flag:
+                return action(context, data_dict)
+            lock.acquire()
+            while flag in wrapper._in_progress:
+                lock.wait(.5)
```
torfsen: I think using a [`threading.Event`](https://docs.python.org/2/library/threading.html#event-objects) would be a better choice instead of a fixed-delay wait loop here.

wardi:
This approach won't work. CKAN processes can access the database independently so a mutex in a web process won't help. Please look at my suggested fixes in #3245 instead.

PR #3337, opened 2016-11-28
[#3332] Properly close database connections in background workers

### LINKED ISSUE #3332: [CircleCI] - Test fails - SSL connection has been closed unexpectedly
Sometimes CircleCI test fails, the cause is that SSL connection has been closed unexpectedly.

torfsen:
Fixes #3332.

RQ background workers exit via `os._exit`, which (by design) omits all of Python's usual clean up routines. As a result, the SQLAlchemy session and its associated database connections have to be cleaned up manually. This hadn't been done before, leading to sporadic exceptions due to SSL connections being unexpectedly closed.

This commit fixes that situation by properly closing each worker's SQLAlchemy session and its associated database connections.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -284,6 +284,8 @@ def update_config():
     except sqlalchemy.exc.InternalError:
         # The database is not initialised.  Travis hits this
         pass
-    # if an extension or our code does not finish
-    # transaction properly db cli commands can fail
+
+    # Close current session and open database connections to ensure a clean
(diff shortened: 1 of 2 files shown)


amercader:
Great find @torfsen! It seemed odd that RQ used `os._exit`, so I investigated a bit and found this in the python [docs](https://docs.python.org/2/library/os.html#os._exit):

> Note The standard way to exit is sys.exit(n). _exit() should normally only be used in the child process after a fork().

PR #3342, opened 2016-12-02
javascript: use ckan locale instead of browser locale to format dates

fanderegg:
The current implementation of the automatic conversion of dates into the timezone of the browser uses the language of the browser to format/translate the date. The problem is when the browser language is different from the current ckan language, the date is translated into a different language than the rest of the ckan interface. With this pull request we always use the current ckan locale to translate the date.

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/main.js
+++ b/ckan/public/base/javascript/main.js
@@ -20,7 +20,6 @@ this.ckan = this.ckan || {};
   ckan.initialize = function () {
     var body = jQuery('body');
     var locale = jQuery('html').attr('lang');
-    var browserLocale = window.navigator.userLanguage || window.navigator.language;
     var location = window.location;
     var root = location.protocol + '//' + location.host;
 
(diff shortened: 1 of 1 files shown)


wardi:
You should be able to add
```ini
ckan.display_timezone = server
```
See http://docs.ckan.org/en/latest/maintaining/configuration.html?highlight=display_timezone#ckan-display-timezone

If not, we should fix that.

fanderegg:
The problem here is not the time zone being used but the selected translation from moment.js.
moment.locale switches the translation of moment.js:
http://momentjs.com/docs/#/i18n/changing-locale/

Lets say my browser is set to English and my time zone is UTC+2, but I use ckan in German, it would render the date/time like this:
Änderungsdatum: October 4, 2016, 6:14 PM (UTC+02:00)
Instead of this (= with my pull request applied):
Änderungsdatum: 4. Oktober 2016, 18:14 (UTC+02:00)

If your browser language matches your ckan language, nothing will change with this pull request.

CC: @metaodi

wardi:
Oh I get it now.

This change formats dates according to the language the user has selected, right?

fanderegg:
Yes exactly, the ckan language the user selected, taken from the lang attribute of the <html> tag.

PR #3344, opened 2016-12-02
Allow larger datastore dumps

wardi:
This is a small improvement to datastore that allows large CSV dumps and shrinks the maximum memory used by paging over the data and streaming the output CSV

### DIFF
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -5,45 +5,56 @@ import unicodecsv as csv
 
 import pylons
 
-import ckan.plugins as p
-import ckan.lib.base as base
-import ckan.model as model
+from ckan.plugins.toolkit import (
+    Invalid,
+    ObjectNotFound,
+    get_action,
+    get_validator,
+    _,
+    request,
+    response,
+    BaseController,
+    abort,
+)
 
-from ckan.common import request
-
+int_validator = get_validator('int_validator')
 
 PAGINATE_BY = 10000
 
 
-class DatastoreController(base.BaseController):
+class DatastoreController(BaseController):
     def dump(self, resource_id):
-        context = {
-            'model': model,
-            'session': model.Session,
-            'user': p.toolkit.c.user
-        }
+        try:
+            offset = int_validator(request.GET.get('offset', 0), {})
+        except Invalid as e:
+            abort(400, u'offset: ' + e.error)
+        try:
+            limit = int_validator(request.GET.get('limit'), {})
+        except Invalid as e:
+            abort(400, u'limit: ' + e.error)
 
-        offset = 0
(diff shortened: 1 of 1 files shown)


mattfullerton:
:+1: Thanks for this!

amercader:
In the original code it seems like people could define the limit on the request (eg I just want the first 10 rows). If I'm understanding your changes correctly people now will always get all the rows regardless of `limit` as it is used as the page size. So passing the `limit` parameter would just affect the internal paging size. Shouldn't we separate `limit` and `page`?

wardi:
You're right. Are those parameters documented? (Do we *want* to keep them?) The way they were written made it easy to consume all the memory on the web server if a user requests enough rows.

I can add them back but I would like to document them at least.

amercader:
There should probably be a hard limit either way, with params or without, even if it's a big one (1m ?)
Your offset is not actually used on the following loops if there is an `offset` passed as param. I fine dropping support for the params to simplify things.
What would probably be more useful for selective dumps is allow an option to download as CSV on `datastore_search`.

wardi:
I agree, a `datastore_search` CSV output option would be more flexible. It doesn't fit the action API protocol though. Ideally we could get any of our (paginated or not) API calls' data returned as a CSV streamed back to the end user.

Let me just add those options back here in case someone is using them.

On another topic: Should I be including a UTF-8 BOM here so that Excel can open these files? It would be a user-visible change. Add an option for that at the same time?

amercader:
I personally would not add it by default. I'm happy with an option to enable it optionally.

wardi:
There's no reason for a hard limit on this call. If the client keeps reading the data this code will continue looping to the next 10k records until the end. There's no benefit in making the client send multiple requests.

amercader:
Fair enough, we'll assume there are no memory leaks there :)

TkTech:
@wardi regarding incompatibility with the existing api, we could support `Accept: text/csv` with a utility wrapper around certain actions. A quick `LazyCSVWriter` that doesn't write the headers until the first dict/row is consumed to get the structure? Makes the assumption that every row will be identical.

ckanext/datastore/controller.py
```diff
@@ -20,27 +23,34 @@ def dump(self, resource_id):
...
+            }
+
+            action = p.toolkit.get_action('datastore_search')
+            try:
+                result = action(context, data_dict)
+            except p.toolkit.ObjectNotFound:
+                base.abort(404, p.toolkit._('DataStore resource not found'))
+
+            if not wr:
+                pylons.response.headers['Content-Type'] = 'text/csv'
```
amercader: `Content-Type: text/csv; charset=utf-8` ?

Issue #3319, opened 2016-11-18
docker-compose up will fail on initial database timeout


rwillmer:
### CKAN Version if known (or site URL)
2.6.0 (docker images tagged dev-v2.6)

### Please describe the expected behaviour
With contrib/docker/docker-compose.yml modified to add required environment variables (#3075)
'docker-compose up' should bring up a working system

### Please describe the actual behaviour
The 3 containers all appear to be able to start. The use of depends_on starts them in the correct order.

However, the db container has not fully started by the time the ckan container looks for it.

It fails with 
```
ckan    | sqlalchemy.exc.OperationalError: (OperationalError) could not connect to server: Connection refused
ckan    | 	Is the server running on host "db" (172.18.0.3) and accepting
```

### What steps can be taken to reproduce the issue? 
```
$ cd contrib/docker
$ docker-compose up
```

rwillmer:
I shall add a 'wait-until-ready' into my ckan entry point, as suggested by docker

https://docs.docker.com/compose/startup-order/

However, I wonder whether ckan needs, or already has, a reconnect option which is used if the database connection fails at a later stage.

rwillmer:
I'll also add a link to https://github.com/dmfenton/ckan-docker, which is a working implementation.

deinok:
@rwillmer Working on it, thanks

deinok:
@rwillmer Should be fixed, can you check it?

frafra:
Still need to reboot ckan because the db is not ready.

Issue #3075, opened 2016-05-31
`docker-compose up` does not work


dmfenton:
The way that you are specifying environmental variables in these lines is incorrect: https://github.com/ckan/ckan/blob/master/contrib/docker/my_init.d/50_configure#L36-L49

As a result the CKAN container cannot connect to the DB or SOLR

``` bash
ckan | *** Running /etc/my_init.d/50_configure...
ckan | /etc/my_init.d/50_configure: 37: /etc/my_init.d/50_configure: DB_ENV_POSTGRES_USER: parameter not set
ckan | no DATABASE_URL specified and linked container called 'db' was not found
ckan | *** /etc/my_init.d/50_configure failed with status 1
ckan |
ckan | *** Killing all processes...
ckan exited with code 1
```

I added the ENV's for `DATABASE_URL` and `SOLR_URL` directly into docker-compose.yml and everything works.

``` yml
version: '2'
services:
  ckan:
    container_name: ckan
    image: ckan/ckan:latest
    depends_on:
      - db
      - solr
    links:
       - db
       - solr
    ports:
       - "80:80"
    environment:
       DATABASE_URL: "postgresql://ckan:ckan@db:5432/ckan"
       SOLR_URL: "http://solr:8983/solr/ckan"
  db:
    container_name: db
    image: ckan/postgresql:latest
  solr:
    container_name: solr
    image: ckan/solr:latest
```

My Docker version info:

``` bash
$ docker version                                                                                                                      [16:37:48]
Client:
 Version:      1.10.2
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   c3959b1
 Built:        Mon Feb 22 22:37:33 2016
 OS/Arch:      darwin/amd64

Server:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Wed Apr 27 00:34:20 2016
 OS/Arch:      linux/amd64
```

deinok:
It's a know issue, im working improving the docker image of ckan, you can use the docker-cloud.yml in [http://cloud.docker.com](http://cloud.docker.com) for now, or follow this legacy instructions [http://docs.ckan.org/en/ckan-1.3.3/maintaining/installing/install-using-docker.html#installing-ckan](http://docs.ckan.org/en/ckan-1.3.3/maintaining/installing/install-using-docker.html#installing-ckan)

The actual image is not ready for production, CMD script and CKAN docker image needs to be full rewrite

[...]

PR #3346, opened 2016-12-04
Docker BugFix

deinok:


deinok:
Requires #3346

deinok:
Closes #3319

deinok:
Closes #3075

deinok:
Requires #3347

deinok:
@wardi any update on this

circle.yml
```diff
@@ -65,3 +82,12 @@ test:
...
         - mocha-phantomjs http://localhost:5000/base/test/index.html
+
+        # Docker
+        - docker run -d --name db postgresql
+        - docker run -d --name solr solr
+        - docker run -d --name redis redis:latest
+        - docker run -d --name ckan -p 5001:5000 --link db:db --link redis:redis --link solr:solr ckan
+        - sleep 10
+        - curl --retry 10 --retry-delay 5 -v http://localhost:5001
+
```
wardi: You've added about 2 minutes to the all CI containers. Would you add your docker build and test code to https://github.com/ckan/ckan/blob/master/.circleci-matrix.yml so that it only runs in one container instead of all 4?

amercader:
I think we definitely should test the Docker images, building and basic functional tests (as in "site is up and running"), but probably separating it from the normal tests. We could use one of the Circle CI containers for that or use Travis for Docker testing exclusively and have two checks.

circle.yml
wardi: The circlce-matrix docs are more than a little lacking, let me know if you'd like more direction here

amercader:
@deinok great work, it seems like we are almost there. This will be a really good base for people to work on Docker based setups

contrib/docker/solr/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN mkdir -p /opt/solr/server/solr/$SOLR_CORE/data
 
 # Adding Files
 ADD ./solrconfig.xml \
-https://raw.githubusercontent.com/ckan/ckan/dev-v2.6/ckan/config/solr/schema.xml \
+https://raw.githubusercontent.com/ckan/ckan/master/ckan/config/solr/schema.xml \
```
deinok: @amercader For now i don't know how to make the diference between the schemas.
The best aproach is to set a managed schema. I will work on it ;)

deinok:
@amercader For now i think we should use Travis for docker, or wait until CircleCI 2.0 is released.

For now, i will try to set the Docker test in a container.
(For now, docker only tests if it can be build and executed, and the frontpage returns a HTTP-200)
(In a future, we should be able to run the tests inside the docker)

contrib/docker/ckan-entrypoint.sh
```diff
@@ -0,0 +1,86 @@
...
+  ckan-paster make-config ckan "$CONFIG"
+
+  ckan-paster --plugin=ckan config-tool "$CONFIG" -e \
+      "sqlalchemy.url = ${DATABASE_URL}" \
+      "solr_url = ${SOLR_URL}" \
+      "ckan.redis.url = ${REDIS_URL}" \
+      "ckan.storage_path = /var/lib/ckan" \
+      "email_to = disabled@example.com" \
+      "error_email_from = ckan@$(hostname -f)" \
+      "ckan.site_url = http://localhost:5000"
```
wardi: Would it be easier to use the environment variables for these config options instead?

circle.yml
deinok: Yes, i will try to make it ;)
But im not sure if i can cache the images in the matrix :(

contrib/docker/ckan-entrypoint.sh
wardi: Please just add the ones you need, making arbitrary config options override-able from environment vars makes me nervous.

deinok:
@wardi or @amercader Can anybody help me with the travis.yml?

deinok:
#3375

wardi:
If you remove the changes to the circle tests I would merge this, then I can re-enable travis on this repo

contrib/docker/ckan-entrypoint.sh
```diff
@@ -0,0 +1,78 @@
...
+
+write_config () {
+  ckan-paster make-config ckan "$CONFIG"
+
+  ckan-paster --plugin=ckan config-tool "$CONFIG" -e \
+      "sqlalchemy.url = ${DATABASE_URL}" \
+      "solr_url = ${SOLR_URL}" \
+      "ckan.redis.url = ${REDIS_URL}" \
+      "ckan.storage_path = ${CKAN_DATA}" \
+      "ckan.site_url = ${SITE_URL}"
```
wardi: Do we need to call config-tool at all? We've already got environment variables for these settings.

* `CKAN_SQLALCHEMY_URL` will override `sqlalchemy.url`
* `CKAN_SOLR_URL` will override `solr_url`
* `CKAN_REDIS_URL` will override `ckan.redis.url`
* `CKAN_STORAGE_PATH` will override `ckan.storage_path`
* `CKAN_SITE_URL` will override `ckan.site_url`

deinok:
Done

deinok:
Any issue?

contrib/docker/ckan-entrypoint.sh
wardi: These are environment variables recognised by CKAN and used to override INI settings. See https://github.com/ckan/ckan/blob/master/ckan/config/environment.py#L117

If you use these environment variable names you don't need to generate an ini file with `paster config-tool` at all.

wardi:
@deinok just that small improvement to skip generating an ini file, then I'll merge

deinok:
@wardi Done, take a look ;)

wardi:
@deinok great work. Thanks!

deinok:
Also thanks for your help @wardi ;)

PR #3351, opened 2016-12-05
Check group name and id in data_dict during package creation

smotornyuk:
With old checks api won't work correctly. If api_version == 3 and in group name specified instead of group id in package dict, then group won't be added in `groups` variable and no checks, whether user have permissions for this group, are performed


smotornyuk:
I changed `_create_package` method so that package creator added into group as admin and can pass consequent auth checks. 
Also i changed generation of time for checking correctness of activity's timestamp - previously it generated timezone-awared dates and tests do not passed in timezones, different from UTC

@amercader, could you review it now?

amercader:
Looks good @smotornyuk

PR #3353, opened 2016-12-06
[#3352] Introduce all_fields option to user_list with test.

### LINKED ISSUE #3352: Add a user_list option that returns just names instead of dicts
user_list currently returns a dict for each user and the enhancement is to add an option to allow returning the user names.

DGU has 30k users and it takes 3 minutes, which is no good for a web request. Whereas returning just the names takes 10 seconds, skipping the dictization and the additional parts of the query that adds up revisions by the user.

package_list and group_list both return either names or dicts, depending on the `all_fields` parameter, so the proposal is to do the same, the only difference being that the default is True, for backward compatibility. (Other list actions just return dicts.)

davidread:
Fixes #3352 

### Features:

- [no] includes user-visible changes
- [yes] includes API changes
- [na] includes bugfix for possible backport

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -818,6 +818,9 @@ def user_list(context, data_dict):
     :param order_by: which field to sort the list by (optional, default:
       ``'name'``). Can be any user field or ``edits`` (i.e. number_of_edits).
     :type order_by: string
+    :param all_fields: return full user dictionaries instead of just names.
+      (optional, default: ``True``)
+    :type all_fields: boolean
 
     :rtype: list of user dictionaries. User properties include:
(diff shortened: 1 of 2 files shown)


amercader:
Looks good!

wardi:
:confused:

davidread:
But you didn't merge it @amercader...

amercader:
oh no, sorry. I could swear that I had merged it :cold_sweat: Do you have a local copy of your branch to resubmit?

amercader:
If not I'll use this diff to resubmit it and/or merge it

davidread:
Ok looks good to merge again :)

amercader:
Sorry about the mess up before!

PR #3355, opened 2016-12-06
[#3354] Migrate table is kept during tests

### LINKED ISSUE #3354: Initializing test db is not idempotent
Running nosetests deletes the `migrate_version` table, which means that subsequent runs of `paster db init -c test-core.ini` [don't work](https://gist.github.com/davidread/af650b9227e5ed54d891dbeeb0e2c254). This is an issue when you are writing provision scripts, because the easiest thing is for these commands to be idempotent so they can be run repeatedly without harmful effect or error. So it is helpful if our tests don't delete the `migrate_version` table.

How to reproduce:
```
dropdb ckan_test
createdb ckan_test -O ckan_default
paster --plugin=ckan db init -c test-core.ini
nosetests --ckan --with-pylons=test-core.ini ckan/tests/controllers/test_admin.py:TestTrashView.test_trash_purge_deleted_datasets  # or any test
# You can see the table is gone with this:
sudo -u postgres psql ckan_test -c 'select * from migrate_version;'
# Now run init again, as if you are running the provision script again
paster --plugin=ckan db init -c test-core.ini
# That didn't work - see the error here: https://gist.github.com/davidread/af650b9227e5ed54d891dbeeb0e2c254
```

Because of this issue, my ansible play is:
```
  shell: psql ckan_test -c 'select name from package limit 1' | grep row && echo TestDatabaseAlreadyInitialized || /usr/lib/ckan/bin/paster --plugin=ckan db init -c /usr/lib/ckan/src/ckan/test-core.ini
  sudo: true
  sudo_user: postgres
  register: cmd_result
  changed_when: cmd_result.stdout.find('TestDatabaseAlreadyInitialized') == -1
```
whereas with this fix it can simply be:
```
  command: /usr/lib/ckan/bin/paster --plugin=ckan db init -c /usr/lib/ckan/src/ckan/test-core.ini
```

davidread:
Fixes #3354

- [na] includes tests covering changes
- [na] includes updated documentation
- [no] includes user-visible changes
- [no] includes API changes
- [no] includes bugfix for possible backport

### DIFF
--- a/ckan/model/__init__.py
+++ b/ckan/model/__init__.py
@@ -255,6 +255,8 @@ class Repository(vdm.sqlalchemy.Repository):
         else:
             tables = reversed(self.metadata.sorted_tables)
         for table in tables:
+            if table.name == 'migrate_version':
+                continue
             connection.execute('delete from "%s"' % table.name)
         self.session.commit()
         log.info('Database table data deleted')


davidread:
Cheers!

wardi:
@amercader I cherry-picked this over to dev-v2.6 because I'm running in to this issue in my scheming tests now that I've added more ckan versions to test against

wardi:
Hmm. Unfortunately that only got me a bit further. I needed #3366 too.

PR #3358, opened 2016-12-08
Bugfix for #3357, fix package group listing

### LINKED ISSUE #3357: Package group listing should take group type into consideration
CKAN version 2.6.0

Only groups with type 'group' should be listed when viewing package/group_list.html. An identical issue occurs with the group select input when viewing the same template

All groups with the attribute is_organization == false are listed and the group's type-attribute is not taken into account. If e.g. a group with the type 'foo' exists, it will also be listed here. The group listing in route '/group' is working as expected. 

Issue can not be reproduced without new functionality that creates additional group types than 'group' and 'organization'. We have a work in progress [extension](https://github.com/6aika/ckanext-collection) which provides this functionality.

A PR to fix the issue will follow.

mikkonieminen:
Fixes #3357 

### Proposed fixes:

Fixes package group listing to check for group type in addition to the is_organization attribute.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -244,7 +244,8 @@ def package_dictize(pkg, context):
                from_obj=member.join(group, group.c.id == member.c.group_id)
                ).where(member.c.table_id == pkg.id)\
                 .where(member.c.state == 'active') \
-                .where(group.c.is_organization == False)
+                .where(group.c.is_organization == False) \
+                .where(group.c.type == 'group')
     result = execute(q, member, context)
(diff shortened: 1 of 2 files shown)


wardi:
I'd like group types to behave like dataset types in how they work with search. I'm ok with changing the default but we should have a config option to allow all group types to be shown like http://docs.ckan.org/en/latest/maintaining/configuration.html?highlight=all_types#ckan-search-show-all-types does for datasets

mikkonieminen:
Good point, we will look into it when we have time. Closing this one for now.

PR #3360, opened 2016-12-08
cli: TypeError: expected a character buffer object

wardi:
fix a couple places where exceptions are passed to `error()`, causing a failure while trying to fail.


ckan/lib/cli.py
```diff
@@ -1492,9 +1493,7 @@ def profile_url(url):
             except KeyboardInterrupt:
                 raise
             except:
```
TkTech: Might as well make this an `except Exception` while you're there? :)

wardi:
@TkTech done!

TkTech:
Tests pass locally, +1

PR #3364, opened 2016-12-12
`render_markdown` breaks links with ampersands

smotornyuk:
Added few additional allowed tags to `bleach.clean`
function and changed sanitization sequence so that
markdown applied first and only after that result cleaned


smotornyuk:
In addition, updated tests and noticed that previously it was possible to add links like `<a href="javascript: alert(1)">link</a>`. With this PR such cases will be sanitized as well. Is it ok?

amercader:
Great stuff @smotornyuk! Marking for backporting

Issue #2438, opened 2015-05-20
Rename deleted datasets when they conflict with one a user wants to create


wardi:
It's common for users to want to delete a dataset then create one with the same name (URL slug) only to have CKAN tell them they can't do that.

But they should be able to do that! And, we shouldn't have to purge the first package and its history first.

Why don't we munge the name of an existing conflicting deleted dataset from package_create (in a safe way) that would allow the creation of the new dataset?

Munging could be something like `popular-name` -> `popular-name-deleted-1`, `popular-name-deleted-2` and would only happen on package_create, not package_delete.

joetsoi:
I have spoken to two groups of people who want this recently, so I'm going to plus one this twice.
+1
+1

amercader:
Big +1 as well

It's still possible for sysadmins to edit a deleted dataset (they can find them via their dashboard) and edit it to be active again. Should we have a `popular-name-resurrected-1` munging as well? :)

rossjones:
Isn't the problem with doing this rename in package_create that the check for existing package is done in validation with the ```package_id_does_not_exist``` validator. This means either an explicit check for a specific validation error in package_create, or the validator doing more than it should.

Renaming to some version of the existing name then makes it difficult to undelete as the name will be in used.  If we are expecting the user to have to rename during undelete - we may as well just set the name to the id when it is deleted.  On undelete we can follow the usual munging of the title to reset it?

wardi:
yeah, I was thinking of doing the explicit check for a specific validation error in package_create.

The problem with renaming immediately on delete is it's hard to go back and undelete when you deleted by mistake. We could work around that by redirecting to the new package edit url after delete, but once you lose that url how do you find your dataset?

rossjones:
I wish we had numbers on whether anyone actually ever undeletes, or uses history, etc etc.

wardi:
history has been pretty broken for a while, so probably not many users there. Breaking things is a pretty good way of finding out who's using a feature :-/

rossjones:
I think this problem may only actually be one-way.  We've had lots of 'complaints' that package_create doesn't allow people to use the name of a 'deleted' dataset, perhaps we should optimise for that and worry less about people having to change the name field when it is undeleted?  

How about on package_delete we just change the name to ```'__deleted__%s' % package['name']``` and document it for users wanting to find the deleted dataset to undelete?

wardi:
[...]

PR #3370, opened 2016-12-16
Rename deleted datasets

smotornyuk:
When user creates new dataset with the same name as in previously removed, old dataset renamed - its id used instead of name, in order to prevent feature conflicts.

I just noticed #2438 and decided to start progress, because it's quite essential problem in many of our projects.

Currently old dataset renamed inside validator - it's not very nice and can be placed into create/update actions, but i just don't want trigger additional queries in order to prevent this validation. And, at the same time, it can produce some errors in feature if we just check whether `errors` dict contains particular validation message and do renaming after that.

As for undeleting - now dataset restored with its id inside name field. As was discussed earlier in issue, restoring old name is not a good idea. I don't think that new dataset should be somehow referred from our restored dataset - maybe just some flash-warning(like suggested Ross) with message, that name has been changed, but without link to new owner of this name


smotornyuk:
Dataset is renamed inside *lib.dictization.__init__*, after validation happens. Validation itself, ignores previously existing dataset with the same name if it has `deleted` state.
And i accidentally found duplicated line at [122 and 130](https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/__init__.py#L122-L130). I think that no one against removing first one?

ckan/lib/dictization/__init__.py
```diff
@@ -132,7 +131,11 @@ def table_dict_save(table_dict, ModelClass, context):
             params = dict((key, table_dict.get(key)) for key in constraint)
             obj = session.query(ModelClass).filter_by(**params).first()
             if obj:
-                break
+                if 'name' in params and getattr(obj, 'state', None) == State.DELETED:
+                    obj.name = obj.id
+                    obj = None
+                else:
+                    break
```
wardi: IIUC this will clear a datasets name when it's deleted, but if a dataset was deleted before this change is made the change will fail when postgres enforces the unique name constraint.

Instead of this I suggest catching the postgres failure and rename the deleted dataset before retrying the one we're trying to save.

ckan/logic/validators.py
```diff
@@ -348,16 +349,16 @@ def package_name_validator(key, data, errors, context):
...
         package_id = package.id
     else:
         package_id = data.get(key[:-1] + ('id',))
     if package_id and package_id is not missing:
         query = query.filter(model.Package.id <> package_id)
     result = query.first()
-    if result:
-        errors[key].append(_('That URL is already in use.'))
+    if result and result.state != State.DELETED:
+            errors[key].append(_('That URL is already in use.'))
```
wardi: this is good but the indentation is off :confused: wonder why our code style checker didn't catch that

ckan/lib/dictization/__init__.py
wardi: Oh, I'm wrong. this code is trying to prevent the postgres error, so it's actually fine. :+1:

At a later time we can update this code to rely on postgres enforcement instead of avoid it, but that's out of scope for this change.

wardi:
@smotornyuk just the little style fix and I'll merge. Thanks for your work on this!

smotornyuk:
Updated. Don't know, how my linter missed this line

Issue #3365, opened 2016-12-13
Intermittent test failures in test_jobs


amercader:
e.g. https://circleci.com/gh/ckan/ckan/2035

@torfsen just to keep this open as we closed #3359

torfsen:
I've spent some time on isolating the problem, see http://stackoverflow.com/q/41279157/857390.

torfsen:
Thanks for the suggestion, @TkTech! I've [re-posted my question to the SQLAlchemy mailing list](https://groups.google.com/forum/#!topic/sqlalchemy/_SN2BDipBw8).

torfsen:
With [the help of StackOverflow](http://stackoverflow.com/a/41722129/857390) I think I've found the underlying problem. Here's how I understand what happens:

1. The main CKAN process initializes its SQLAlchemy engine and stores references to the engine and session in global variables (`ckan.model.meta.engine`, etc.)
2. When a worker is started, rq forks a new child process. That process inherits its parent's global state, including the session and the global variables referencing it.
3. The worker process now initializes its own engine and session, which is necessary because engines cannot be shared between processes.
4. The new worker session is then stored in the global variable that previously pointed to the worker's copy of the parent's session. This causes the refcount of that copy to drop to 0.
5. When the garbage collector in the child now happens to run it will finalize the worker's copy of the parent's session since it's not referenced anymore (within the worker process). This also closes the associated connections, which are shared with the parent. 
6. When the parent now tries to use these connections it finds them unexpectedly closed and raises an exception.

There are two possible solutions to this problem:

1. Keep references to the copy of the parent's session in the worker process so that it doesn't get garbage collected. This is nice because nothing changes for the parent but it's a bit hacky.
2. In the parent, close all connections and dispose the engine before forking and then re-initalize things afterwards. This ensures that no connections are shared with the child, since no connections exist during the fork. This is a cleaner way of doing things but more complicated, since starting a worker now affects the current session.

I'll have a look at how complicated these two approaches are to implement in CKAN (and whether they actually solve the issue :stuck_out_tongue_winking_eye:).

wardi:
@torfsen great detective work! I agree approach #2 sounds like the cleaner option, but a working fix is better than a clean one. :-)

wardi:
fixed by #3377

PR #3377, opened 2016-12-21
[#3365] Fix intermittent test failures in test_jobs

torfsen:
This PR is WIP to fix the intermittent test failures of the background job tests (#3365).

I can't reproduce these failures locally, so I'm going to use this PR to find out what's going wrong when running the tests on CircleCI.

### DIFF
--- a/ckan/lib/jobs.py
+++ b/ckan/lib/jobs.py
@@ -207,6 +207,14 @@ def test_job(*args):
 class Worker(rq.Worker):
     u'''
     CKAN-specific worker.
+
+    Note that starting an instance of this class (via the ``work``
+    method) disposes the currently active database engine and the
+    associated session. This is necessary to prevent their corruption by
+    the forked worker process. Both the engine and the session
+    automatically re-initialize afterwards once they are used. However,
+    non-committed changes are rolled back and instance variables bound
+    to the old session have to be re-fetched from the database.
     '''
     def __init__(self, queues=None, *args, **kwargs):
         u'''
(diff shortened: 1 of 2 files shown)


amercader:
@torfsen any update on this?

torfsen:
@amercader Just came back from my winter holidays, so I'm looking into this again.

torfsen:
I think that I have found the problem and fixed it. As described [in my comment](https://github.com/ckan/ckan/issues/3365#issuecomment-273496859) over at #3365 the issue seems to be that CKAN stores the SQLAlchemy engine and the associated connections in global state which is inherited by the worker process and garbage collected there after the worker has initialized its own engine. That closes the connections initially copied from the main process which are still in use there.

This PR solves the issue by explicitly disposing the engine and the associated connections in the main process before the fork. That way, the worker process doesn't inherit any shared connections in the first place. This is clean and deterministic (instead of the random failures we had before which were likely caused by the "random" times when the GC ran). However, it also implies that anything associated with a session before the fork cannot be re-used after the fork (in the parent), since that session has been disposed. Doing so will (deterministically) raise an exception. In practice this isn't a real problem, though, since worker processes are only forked in calls to `paster jobs worker` and specially crafted test cases.

From my point of view this PR is now ready to merge.

wardi:
@torfsen care to remove the SkipTest I added as part of this PR?

torfsen:
@wardi I think [I did that](https://github.com/ckan/ckan/pull/3377/files#diff-11fce695903e1f45a8c41bc91ec9f279L247) :wink:

wardi:
@torfsen I'm clearly blind :smile:

Great work on this! I'm looking forward to all this things we can move to jobs.  Updating dataset index on org changes, a local datapusher, there's so much we can do with this code :+1:

PR #3380, opened 2016-12-22
Add an ckan.preview.image_formats config option for the image viewer

mattfullerton:
Addresses http://stackoverflow.com/questions/41213685/recognised-file-types-for-image-preview

### Features:
Add an ckan.preview.image_formats config option for the image viewer	
- [ ] includes API changes

### DIFF
--- a/ckanext/imageview/plugin.py
+++ b/ckanext/imageview/plugin.py
@@ -6,7 +6,7 @@ import ckan.plugins as p
 log = logging.getLogger(__name__)
 ignore_empty = p.toolkit.get_validator('ignore_empty')
 
-DEFAULT_IMAGE_FORMATS = ['png', 'jpeg', 'jpg', 'gif']
+DEFAULT_IMAGE_FORMATS = 'png jpeg jpg gif'
 
 
 class ImageView(p.SingletonPlugin):
@@ -17,8 +17,7 @@ class ImageView(p.SingletonPlugin):
 
     def update_config(self, config):
         p.toolkit.add_template_directory(config, 'theme/templates')
-        image_formats = config.get('ckan.preview.image_formats', '').split()
-        self.formats = image_formats or DEFAULT_IMAGE_FORMATS
+        self.formats = config.get('ckan.preview.image_formats', DEFAULT_IMAGE_FORMATS).split()
 
     def info(self):
         return {'name': 'image_view',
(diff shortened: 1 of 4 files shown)


mattfullerton:
Tests are failing on docs; will have to review that - don't review yet!

mattfullerton:
This is ready now; failing test seems to be due to a DB connection problem and nothing source-code related.

doc/maintaining/configuration.rst
```diff
@@ -1214,6 +1214,19 @@ Default value: ``text plain text/plain``
...
+ckan.preview.image_formats
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Example::
+
+ ckan.preview.text_formats = png jpeg jpg gif
+
+Default value: ``png jpeg jpg gif``
+
+Image based resource formats that will be rendered by the Image view plugin (``image_view``)
```
TkTech: I would just add a couple words to this to explicitly state it's a space-delimited list of formats.

TkTech:
Can you add it (with the default values) as a commented out item in the config template? https://github.com/ckan/ckan/blob/master/ckan/config/deployment.ini_tmpl

mattfullerton:
Done and did the same (on both points) for text_view

TkTech:
👍 I'm happy with this if someone wants to second and merge.

ckanext/imageview/plugin.py
```diff
@@ -17,6 +17,8 @@ class ImageView(p.SingletonPlugin):
 
     def update_config(self, config):
         p.toolkit.add_template_directory(config, 'theme/templates')
+        image_formats = config.get('ckan.preview.image_formats', '').split()
```
wardi: use the default implied in the comment in the config above?

ckanext/imageview/plugin.py
```diff
@@ -17,6 +17,8 @@ class ImageView(p.SingletonPlugin):
...
+        self.formats = image_formats or DEFAULT_IMAGE_FORMATS
```
wardi: Just saw this and deleted my last comment, but I think it's better to use the DEFAULT_IMAGE_FORMATS in the get() so that explicitly setting no image formats in the ini actually works (it's what the user requested, even if it's strange)

mattfullerton: Agreed and committed

amercader:
Thanks @mattfullerton!

PR #3382, opened 2016-12-29
Update pylons to 1.0.2

TkTech:
Updates several dependencies and includes minor backwards-compatibility aliases to leave most well-behaved extensions unchanged.

This upgrade is necessary for multiple timing and XSS attacks that are patched in downstream libraries that have been forever pinned to old versions. It also fixes the interactive debugger failing.

This should last us until Flask is ready :)


TkTech:
So this is actually fully functional now, but there are still issues with running tests I'm looking into.

TkTech:
Hm, we seem to be hitting circleci's container timeout. Might be time to do another test performance sprint.

TkTech:
To reiterate, for those that absolutely most upgrade pylons due to active exploits that affect them, *this PR can be used right now*. It's currently up to date with master and fully functional other than breaking the test framework. The outstanding Pylons exploits are very niche and do not affect stock CKAN.

davidread:
Note: to run tests after this change you need to:
* pip install -r requirements.txt   # as you might expect
* python setup.py develop    # because pylons 1.0.1 delivers its plugin to nose differently. If you don't do this you might get: "AttributeError: No configuration has been registered for this process or thread" or "nosetests: error: Error reading config file 'setup.cfg': no such option 'with-pylons'" errors

BTW I'm seeing tests run pause often, which may be the cause of the timeouts, specific to this PR.

kmbn:

PR #3384, opened 2017-01-02
datastore cli: add dump command

wardi:
see below https://github.com/ckan/ckan/pull/3384#issuecomment-290787636


wardi:
TBD: add load_config to plugins.toolkit

wardi:
I didn't add the create-test-data command, but there's plenty of good stuff here to consider for merging as-is.

- Factored out the CKAN init into a single `load_config` function call, instead of requiring that CLI commands inherit from CkanCommand class. CkanCommand ties us to paster and argparse.
- Re-implemented the simple datastore cli on top of click, the most commonly used cli library with Flask applications
- Added a `datastore dump` subcommand that allows dumping from the datastore the same way and with the same new options the dump controller has

I'm leaving performance improvements and new commands for separate PRs.

wardi:
@amercader forgot to mention this at the meeting. Would you mind having a look?

amercader:
This looks great @wardi. Click's syntax looks weird at first but I think it will be a great improvement over paster.

Do you want to pin a particular Click version on the requirements rather than relying on whatever Flask requires?

> TBD: add load_config to plugins.toolkit

^ This would be cool as well

wardi:
hmm. pip-compile hasn't been run recently, should I leave that for before release and just add the one line I need to avoid conflicts?

amercader:
On this PR add just Click for sure. We should run pip-compile on master on a separate PR, and we should do it well before the 2.7 release (ie now) to identify any potential issues.

wardi:
added load_config and pinned click

ckan/plugins/toolkit.py
```diff
@@ -68,9 +68,9 @@ class _Toolkit(object):
         'StopOnError',
         # validation invalid exception
         'Invalid',
-        # old class for providing cli interfaces
+        # old class for providing CLI interfaces
         'CkanCommand',
-        # new funtion for initializing cli interfaces
```
wardi: :smile: perfect!

PR #3385, opened 2017-01-03
group/org show include_dataset_count param

wardi:
Make it possible to suppress package_count calculation if we don't need it from some action calls


wardi:
changelog for organization_list_for_user

amercader:
Quick Changelog mention and we are good to go

PR #3390, opened 2017-01-06
datastore: dump more formats

wardi:
Extend the datastore controller to dump CSV with BOM (CSV for Excel), JSON, TSV, XML formats as well as plain CSV

### DIFF
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -18,12 +18,13 @@ from ckan.plugins.toolkit import (
 from ckanext.datastore.writer import (
     csv_writer,
     tsv_writer,
+    json_writer,
 )
 
 int_validator = get_validator('int_validator')
 boolean_validator = get_validator('boolean_validator')
 
-DUMP_FORMATS = 'csv', 'tsv'
+DUMP_FORMATS = 'csv', 'tsv', 'json'
 PAGINATE_BY = 10000
 
 
@@ -45,6 +46,8 @@ class DatastoreController(BaseController):
                 return csv_writer(response, columns, resource_id, bom)
             if fmt == 'tsv':
                 return tsv_writer(response, columns, resource_id, bom)
+            if fmt == 'json':
+                return json_writer(response, columns, resource_id, bom)
             abort(400, _(
                 u'format: must be one of %s') % u', '.join(DUMP_FORMATS))
 
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -19,12 +19,13 @@ from ckanext.datastore.writer import (
     csv_writer,
     tsv_writer,
     json_writer,
+    xml_writer,
 )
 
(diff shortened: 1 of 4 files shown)


ckanext/datastore/writer.py
```diff
@@ -0,0 +1,169 @@
...
+        response.headers['Content-Type'] = (
+            b'application/json; charset=utf-8')
+        if name:
+            response.headers['Content-disposition'] = (
+                b'attachment; filename="{name}.json"'.format(
+                    name=encode_rfc2231(name)))
+    if bom:
+        response.write(UTF8_BOM)
+    response.write(
+        b'{\n  "columns": %s,\n  "data": [' % json.dumps(
```
wardi: @amercader actually, I'm planning to replace the "columns" here with the "fields" definition in the same style that is accepted by `datastore_create`, so don't merge just yet

ckanext/datastore/controller.py
```diff
@@ -32,14 +39,24 @@ def dump(self, resource_id):
...
+            if fmt == 'csv':
+                return csv_writer(response, columns, resource_id, bom)
+            if fmt == 'tsv':
+                return tsv_writer(response, columns, resource_id, bom)
+            if fmt == 'json':
+                return json_writer(response, columns, resource_id, bom)
+            if fmt == 'xml':
+                return xml_writer(response, columns, resource_id, bom)
+            abort(400, _(
+                u'format: must be one of %s') % u', '.join(DUMP_FORMATS))
```
wardi: If we want to add a plugin interface for other formats (or overriding formats) this would be the place, but let's do that as after this PR to keep this change small.

ckanext/datastore/writer.py
amercader: Also `data` is called `records` in `datastore_create`

wardi: Sure I can change that but the data here is a list of lists not a list of objects. Should we change `datastore_create` so that it accepts a list of lists too?

amercader: We would need to keep track of the ordering of the fields right? I guess we can assume that the order of the `fields` dicts matches the order of the `records` lists. It will make the params parsing logic a bit more complex but perhaps it's worth to be consistent. Up to you.

wardi: I won't bother changing `datastore_create` here because the fields list there is optional. We can address that later if it's an issue

amercader:
Nice! Definitely a major feature for next release

mattfullerton:
@wardi @amercader Any plans for a dropdown UI for this and/or labelling the resources with multiple formats and/or putting all these formats into the Solr index if datastore loading has taken place? If not I'm happy to do some or all of these - would be a valuable addition for end users who are otherwise painstakingly creating multiple versions of files.

tino097:
@mattfullerton i've worked on this a little bit with #3479. I could make a PR and you could add your work on top of it

PR #3391, opened 2017-01-07
some tests running on all circleci containers

wardi:
I'm guessing it's just new ones that are written as functions at the module level, like ckan.tests.test_coding_standards. I would like to support tests in this style, let's see how hard that will be.


wardi:
Easy fix, and shaves ~50 seconds off our test run time

PR #3392, opened 2017-01-09
Changes to the way of uploading the CKAN logo image. In the form we w…

tino097:
The feature will change the way of setting up the CKAN logo. Instead of current setting up the path to the image, it could be uploaded with the image upload control. 

Current status of the code, give us the control on the form. @amercader Please check the following snippet

`{{ form.image_upload(data, errors, **is_upload_enabled=true**, is_url=is_url, is_upload=is_upload, upload_label = _('Site logo'), url_label=_('Site logo')) }}`

the bold parameter value should be retrieved with the helper, h.uploads_enabled() but i was unable to access to helper.


ckan/controllers/admin.py
```diff
@@ -50,7 +50,7 @@ def _get_config_form_items(self):
             {'name': 'ckan.site_title', 'control': 'input', 'label': _('Site Title'), 'placeholder': ''},
             {'name': 'ckan.main_css', 'control': 'select', 'options': styles, 'label': _('Style'), 'placeholder': ''},
             {'name': 'ckan.site_description', 'control': 'input', 'label': _('Site Tag Line'), 'placeholder': ''},
-            {'name': 'ckan.site_logo', 'control': 'input', 'label': _('Site Tag Logo'), 'placeholder': ''},
+            {'name': 'ckan.site_logo', 'control': 'image_upload', 'label': _('Site Tag Logo'), 'placeholder': ''},
```
amercader: @tino097 The form for the config options is created with `autoform.generate()` based on this dict, so you can pass the `is_upload_enabled` param here (with the value resulting from calling `h.upload_enabled()`, which you can call on this controller). Then modify `autoform.html` to use this value instead of the hardcoded value. Does this make sense?

amercader:
@tino097 for handling the actual upload have a look at `group_create` ([here](https://github.com/ckan/ckan/blob/61ba91f9ade7dcf703910734909b023c949ad714/ckan/logic/action/create.py#L626:L628) and [here](https://github.com/ckan/ckan/blob/61ba91f9ade7dcf703910734909b023c949ad714/ckan/logic/action/create.py#L702)) and replicate it on [`config_option_update`](https://github.com/ckan/ckan/blob/61ba91f9ade7dcf703910734909b023c949ad714/ckan/logic/action/update.py#L1187), replacing `group` by `admin` when calling `uploader.get_uploader()`

ckan/lib/app_globals.py
```diff
@@ -39,6 +39,9 @@
     'ckan.favicon': {}, # default gets set in config.environment.py
     'ckan.template_head_end': {},
     'ckan.template_footer_end': {},
+    'image_url':{},
+    'image_upload':{},
+    'clear_upload':{},
```
wardi: this looks out of place..

ckan/logic/schema.py
```diff
@@ -665,6 +665,9 @@ def default_update_configuration_schema():
         'ckan.site_custom_css': [unicode],
         'ckan.main_css': [unicode],
         'ckan.homepage_style': [is_positive_integer],
+        'image_url':[ignore_missing, unicode],
```
wardi: `unicode` here is certainly wrong

tino097:
@amercader @wardi hey guys, can you check the test that is failing ?

amercader:
@wardi @tino097 OK, this turned out to be much more complex than it appears.

1. Our widget for image uploads uses `image_url`, `image_upload` and `clear_upload` as actual name fields but we need to keep using `ckan.site_logo` so we actually edit the relevant configuration setting, so you need these changes to use custom name fields specific to the logo: https://gist.github.com/amercader/8b637e5e3a9358d11852a040aa84144c


2. Also this will just store the relative path to the logo, but in order to be used in the frontend we need a URL (similarly to what is done for the [groups](https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L457) when dictizing them:

```diff
diff --git a/ckan/logic/action/update.py b/ckan/logic/action/update.py
index 3ff509c..a5fba86 100644
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -11,6 +11,8 @@ import mimetypes
 from ckan.common import config
 import paste.deploy.converters as converters
 
+
+import ckan.lib.helpers as h
 import ckan.plugins as plugins
 import ckan.logic as logic
 import ckan.logic.schema as schema_

@@ -1259,6 +1260,10 @@ def config_option_update(context, data_dict):
 
     for key, value in data.iteritems():
 
+        # Set full Logo URL
+        if key == 'ckan.site_logo' and value and not value.startswith('http'):
+            value = h.url_for_static('uploads/admin/{0}'.format(value))
+
         # Save value in database
         model.set_system_info(key, value)

```
   You might want to test a bit this logic with all combinations, I didn't check too much,

3. Reverting the change in the failing test should fix it:

```diff
diff --git a/ckan/tests/controllers/test_admin.py b/ckan/tests/controllers/test_admin.py
index c9e031a..8915c48 100644
--- a/ckan/tests/controllers/test_admin.py
+++ b/ckan/tests/controllers/test_admin.py
@@ -140,7 +140,7 @@ class TestConfig(helpers.FunctionalTestBase):
         # remove logo
         env, config_response = _get_admin_config_page(app)
         config_form = config_response.forms['admin-config-form']
-        config_form['image_url'] = ''
+        config_form['ckan.site_logo'] = ''
         webtest_submit(config_form, 'save', status=302, extra_environ=env)
 
         # new tagline
```

With all these changes the feature works fine and tests pass

amercader:
Thanks @tino097!

PR #3395, opened 2017-01-11
[#3394] Add `i18n` directory to extension template

### LINKED ISSUE #3394: Extension template doesn't contain an i18n directory
The paster template for extensions currently doesn't include the `ckanext/<name>/i18n` directory. That means that the user needs to create the directory manually, contrary to [what the documentation says](http://docs.ckan.org/en/latest/extensions/translating-extensions.html#extract-strings).

I'll prepare a PR to include the directory in the template.

torfsen:
Fixes #3394.

Previously the paster extension template was missing the `ckanext/<name>/i18n` directory, so users had to create it manually. This PR adds that directory to the template.


amercader:
:+1:

PR #3400, opened 2017-01-16
upgrade font-awesome to v4.0.3

fanjinfei:
Fixes #

### Features:

- [ X] includes user-visible changes
- [X ] includes API changes
  upgrade Font-Awesome from version 3.2.1 to 4.0.3
  please refer to https://github.com/FortAwesome/Font-Awesome/wiki/Upgrading-from-3.2.1-to-4 to upgrade your code accordingly.


amercader:
@fanjinfei @wardi This is going to involve changing the templates. eg new FA class names, as [this](https://github.com/FortAwesome/Font-Awesome/wiki/Upgrading-from-3.2.1-to-4) seems to suggest.

For instance I get this because the class `icon-sitemap` no longer exists:

0fuj450

Changing it to `fa fa-sitemap` brings it back again. The code that builds this is [here](https://github.com/ckan/ckan/blob/07ce1ec4ed368a5258bef1a689ab0c2299da8be8/ckan/lib/helpers.py#L544).

I'm not opposed to this if we want to target a newer version of FA but the changes in the templates should be included in this PR (the one I mention and any other icon on the frontend). And the change should be mentioned on the changelog so people can update their themes.

TkTech:
I agree, it would be a great change to make. Backporting fa icons is a really miserable process. Being able to upgrade any time with the new convention would be great.

amercader:
@fanjinfei is this ready for review?

fanjinfei:
@amercader Not yet. There are a few other js and css files need to be updated.

fanjinfei:
@amercader Now it is ready for review.

fanjinfei:
need further patch to remove ie7 support

amercader:
@fanjinfei this is great, just before merging could you add a note on the CHANGELOG, section "API changes and deprecations" about the version change that includes a link to this page https://github.com/FortAwesome/Font-Awesome/wiki/Upgrading-from-3.2.1-to-4?

Thanks!

fanjinfei:
@amercader I just edited the commit message. Is that the same as CHANGELOG you mentioned?

amercader:
Many thanks @fanjinfei! :tada:

amercader:
@mattfullerton yes, this was discussed and we were aware of the impact it would have on extensions. There is always a trade-off between keeping things stable and using modern versions of libraries. We considered that having to update the class names when upgrading was fair to ask, and this was listed on the CHANGELOG as part of this PR.

You raise a good point with ckanext-pages, for extensions that target multiple CKAN versions I guess it's fine to use both syntaxes at the same time, as one will be ignored, eg:

```
<i class="icon-info fa fa-info"></i>
```

@fanjinfei @TkTech WDYT? Should this work? ^

mattfullerton:
@amercader Thanks for the response and sorry I missed it in the CHANGELOG. Knowing that it was a conscious decision is enough for me and great suggestion with the double-classing.

mattfullerton:
https://github.com/ckan/ckanext-pages/pull/51

PR #3404, opened 2017-01-17
EditorConfig File

### LINKED ISSUE #3402: [EditorConfig] - Add .editorconfig
This file will override IDE code style.

This will help maintaining a single code style.
[http://editorconfig.org/](http://editorconfig.org/)

deinok:
Closes #3402

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,12 +7,12 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
-indent_size = 4
-indent_style = tab
+indent_size = 2
+indent_style = space
 insert_final_newline = true
 tab_width = 4
 trim_trailing_whitespace = true
 
 # Python Code Style
 [*.py]
-indent_style = space
+indent_size = 4


.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+
+# Project Root
+root = true
+
+# Default Code Style
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = tab
```
wardi: Most of our non-python source (js and html) is 2-space indents. We shouldn't be using tabs anywhere.

deinok: This is the standard in a lot of languages so it should not be touched. 
The best way its to override JS an HTML, CSS and LESS

Ex:
 [*.{less,css}]
indent_size = 2
indent_style = space

[*.js]
indent_size = 2
indent_style = space

[*.hml]
indent_size = 2
indent_style = space

amercader: We don't want tabs *anywhere* on the CKAN code base :) docs, requirements files, Dockerfiles, etc So I'm +1 on setting `indent_style = space` globally

wardi: yes, I'd allow an exception for Makefiles, but I really don't want tabs as the default

deinok: So the default should be ?
indent_style = space
indent_size = 2

Any file format that allows tabs or size 4?

wardi: I like 2-space indents everywhere but Python and Makefiles. I don't think we have files that need other styles in our source directory.

deinok:
@wardi @amercader Take a look if this is the required code format

.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+root = true
+
+# Default Code Style
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+tab_width = 4
```
wardi: what does `tab_width = 4` mean here?

deinok: In case of using tabs, the indent size will be 4
indent_size its the equivalent to space_witdh

wardi: IIUC that gets inherited by the [*.py] block, which would be wrong because python sees tabs as being 8 spaces.

Why have this line at all?

deinok: Its better define all the setting that let the default, that can change.

But, that line dont afect python code. Python code will have [space : 4]

wardi: I'm saying `tab_width = 8` is a sensible default because it's the de facto standard and it aligns with what the python interpreter expects.  If you set it to 4 here we'll need to set it to 8 in the [*.py] block.

I don't see a reason to set this to 4 for our code base at all, so we should either remove this line or set it to 8 so it doesn't conflict with python's interpretation of tabs.

deinok: [https://www.python.org/dev/peps/pep-0008/#indentation](https://www.python.org/dev/peps/pep-0008/#indentation)

This is not how python understand (\t) is how a text editor understand a (\t).

Puting that a tab (in case of use) is going to be the equivalent of 8 spaces breaks rules of PEP8 and also will break the PEP8 tests in ckan

wardi: I'm talking about https://docs.python.org/2/reference/lexical_analysis.html#indentation

    First, tabs are replaced (from left to right) by one to eight spaces such that the total number of
    characters up to and including the replacement is a multiple of eight (this is intended to be the
    same rule as used by Unix)

So setting `tab_width = 4` on `[*.py]` is simply wrong (as inherited from `[*]`), and setting `tab_width = 4` on `[*]` is unnecessary because none of our files use tabs, so let's just remove it.

Issue #3398, opened 2017-01-13
Groups cannot be depended on in resource.config


torfsen:
In CKAN 2.6 I cannot depend on a previously defined group in the `[depends]` section of a Fanstatic `resource.config`:
```
[groups]

a =
    file1
    file2

b =
    file3
    file4

[depends]

b =
    a
```
raises
```
Traceback (most recent call last):
  File "/usr/lib/ckan/default/bin/paster", line 11, in <module>
    sys.exit(run())
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 102, in run
    invoke(command, command_name, options, args[1:])
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 141, in invoke
    exit_code = runner.run(args)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 236, in run
    result = self.command()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/serve.py", line 284, in command
    relative_to=base, global_conf=vars)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/serve.py", line 329, in loadapp
    **kw)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 247, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 272, in loadobj
    return context.create()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 710, in create
    return self.object_type.invoke(self)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 146, in invoke
    return fix_call(context.object, context.global_conf, **context.local_conf)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/deploy/util.py", line 55, in fix_call
    val = callable(*args, **kw)
  File "/usr/lib/ckan/default/src/ckan/ckan/config/middleware/__init__.py", line 45, in make_app
    load_environment(conf, app_conf)
  File "/usr/lib/ckan/default/src/ckan/ckan/config/environment.py", line 97, in load_environment
    p.load_all()
[...]

PR #3406, opened 2017-01-18
Groups inside resource.config can't be dependent on another groups

smotornyuk:
#3398 Now groups in resource.config['depends'] can define another groups from the same file as their dependencies.
It done by checking, whether dependency defined in the same *resource.config* as group. If so, it replaced with its resources.


amercader:
@torfsen does this help with #3398?

torfsen:
@amercader Need to check that once I have time

amercader:
Code looks good, thanks @smotornyuk

PR #3414, opened 2017-01-24
datastore Data Dictionary as postgresql comments

wardi:
Use postgres comments in the datastore DB to store a JSON object for each column. For now these objects may have a single "label" field containing a human-friendly label for the column in the site's default language.

These JSON objects will be returned from `datastore_search` along with the field ids and types and may be set or updated by including them in the field objects passed to `datastore_create`.


wardi:
This PR needs tests for the API changes, tests for the controller and a way to display the labels and description fields to users without edit access

jqnatividad:
Can confirm it works.  Comments show up in postgres tools like Navicat, psql, Toad, etc.

It'd be interesting to see how regular users can see the column info.  Should it be implemented as a Resource View?

wardi:
Maybe these fields should be returned from `datastore_info` instead of `datastore_search`

wardi:
Sadly I can't use `datastore_info` because it doesn't seem possible to extend, and it returns less information than the "fields" list returned with every `datastore_search`

ckanext/datastore/controller.py
```diff
@@ -84,3 +89,59 @@ def result_page(offset, limit):
...
+                action='dictionary',
+                id=id,
+                resource_id=resource_id)
+
+        return render(
+            'datastore/dictionary.html',
+            extra_vars={'fields': fields})
+
+
+def short_hash(field):
```
amercader: Is this used?

amercader:
@wardi regarding how this will play with the DataPusher you are right in that the DataPusher will remove the table before each run but it would also be easy to retreive the current info before and resubmit it (we are already [calling](https://github.com/ckan/datapusher/blob/master/datapusher/jobs.py#L392) [`datastore_search`](https://github.com/ckan/datapusher/blob/master/datapusher/jobs.py#L189) on the resource table before deleting). The question is if that would be intended behaviour as you might get whatever information you have there (field types, field labels...) out of sync with the new data.

ckanext/datastore/controller.py
wardi: no, I'll remove it

ckanext/datapusher/templates/package/resource_edit_base.html
```diff
@@ -0,0 +1,6 @@
+{% ckan_extends %}
+
+{% block inner_primary_nav %}
+  {{ super() }}
+  {{ h.build_nav_icon('resource_data', _('DataPusher'), id=pkg.name, resource_id=res.id) }}
```
wardi: I just find it strange that the "DataStore" tab only appears when the datapusher plugin is enabled.. I can switch it back though

wardi:
for datapusher the field types shouldn't be preserved, just the value of the "info" field (currently "label" and "notes" but some sites may extend that however they see fit)

amercader:
Great stuff @wardi

JerlynPeh:
Hi,

I was exploring the use of the Data Dictionary on the dataset level when I chanced upon this. 

Being a beginner at CKAN, may I know the prerequisites for using this plugin and if there are screenshots of the output, may I see it?  

Thank you in advance!

wardi:
@JerlynPeh this one is a feature of the datastore plugin which is part of the ckan repository, so you either wait for our next release (you can run master for development) or backport the feature to the version of ckan you're running.

JerlynPeh:
@wardi Thanks for your response! How do I go about backporting this feature? 

What I did is to copy the changes in the codes according to the commits for this feature into my version of CKAN. However, even after restarting the server and CKAN, there is no visible difference and there isn't any error with running the codes. 

What would you advise to do from here?

ggcatu:
What would be the correct way to expand this ?
I want to add an extra field to the form and keep some extra information along with description and label

wardi:
All you need to do is add your field to the template and it should get saved.

wardi:
@ggcatu good idea, PR welcome!

ggcatu:
Any idea on how should I proceed so that any changes made in the dictionary are reflected on the Activity stream of the org? Can't seem to find any documentation on this, would preffear asking before smashing my head against the code

wardi:
@TkTech something for extended activities?

PR #3418, opened 2017-01-27
[#3417] Missing `--ckan` option in extension template's Travis config

### LINKED ISSUE #3417: Extension template's Travis setup doesn't use --ckan for nosetests
The current [paster template for an extension's Travis call to `nosetests`](https://github.com/ckan/ckan/blob/0bb141ae43b51f2c095fba87750c5fdaa00ee816/ckan/pastertemplates/template/bin/travis-run.sh_tmpl) doesn't include the `--ckan` parameter.

In practice this leads to similar effects as if `--reset-db` had been specified.

torfsen:
Fixes #3417 by adding the missing `--ckan` parameter.


amercader:
Thanks @torfsen!

PR #3420, opened 2017-01-30
Jinja's truncate is better than ours, don't try to override it.

### LINKED ISSUE #3419: TypeError: do_truncate() takes at least 2 arguments (4 given)

2.6.0 and 2.7.1a (Master Branch)

User can view any object via a get request, examples are user, organization or dataset, but basically trying to view any overview page of a CKAN 'thing'.

Jinja template throws 500 error performing the do_truncate operation (see screenshot). For example, if I click on my username in the header toolbar or on the name of a dataset the error occurs.

screen shot 2017-01-30 at 12 03 30

So far I've built docker images with both 2.6.0 and the master branch built from source. The Jinja repository released a 2.8.1 on 29 December which could be the root cause of this issue. The CKAN requirements list a dependency on Jinja 2.8, so it's possible a breaking change was introduced in 2.8.1. I'm going to try to update to Jinja 2.9 to see if that helps.

TkTech:
Fixes #3419

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -256,7 +256,6 @@ def update_config():
     env.install_gettext_callables(_, ungettext, newstyle=True)
     # custom filters
     env.filters['empty_and_escape'] = jinja_extensions.empty_and_escape
-    env.filters['truncate'] = jinja_extensions.truncate
     config['pylons.app_globals'].jinja_env = env
 
     # CONFIGURATION OPTIONS HERE (note: all config options will override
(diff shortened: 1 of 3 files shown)


amercader:
@TkTech the original issue was caused by using a more recent version of Jinja2 than the one required (see https://github.com/ckan/ckan/issues/3419#issuecomment-276357220), so to be on the safe side unless really think it's worth it I'd say that we don't backport this in case we introduce a difference in behaviour.

TkTech:
@amercader Are you worried about the core jinja filter itself being incompatible? Jinja's version is handling compatibility with no breakage, it's our (very odd) overridden version (which has no functional effect) that is not handling backwards compatibility.

PR #3423, opened 2017-01-31
[#3315] Mark remaining strings in JS modules as translatable

### LINKED ISSUE #3315: Translate resource-view-filters.js
`Add filter` is not translated: https://github.com/ckan/ckan/blob/master/ckan/public/base/javascript/modules/resource-view-filters.js#L180

Default template should be moved to `initialize` method and i18n should be added. Solution: https://github.com/DanePubliczneGovPl/ckan/commit/c70cd20aec13057fde5a9ef2825e7335e0d0638d

torfsen:
Fixes #3315 by marking all remaining user-visible strings in JS code as translatable.


amercader:
Thanks @torfsen!

Issue #3422, opened 2017-01-31
If a resource is deleted, the corresponding data in the DataStore should be deleted as well


metaodi:
### CKAN Version if known (or site URL)
2.5.3

### Please describe the expected behaviour
If a resource is deleted, the corresponding data in the DataStore should be deleted as well. Otherwise the DataStore will quickly overflow if you harvest a big source and for each resource a new table is created and the old one still in place.

### Please describe the actual behaviour
Currently the deletion of a ressource does not affect the DataStore at all. But every table in the DataStore is linked to a resource in CKAN, so it's safe to assume they share the same lifecycle.

### What steps can be taken to reproduce the issue? 

1. Create a resource in CKAN (with DataStore enabled)
2. Upload a CSV & upload to DataStore
3. Delete the resource in CKAN

The resource is still available in the DataStore

amercader:
@metaodi You are of course absolutely correct. @smotornyuk will have a go at implementing this. It would be great if you could test this once there is a PR/

wardi:
We could also use a command for cleaning out orphaned datastore tables

metaodi:
@wardi maybe additionally, to cleanup "old" databases. But I think to have this implemented _properly_ makes sense, doesn't it?

wardi:
Absolutely! (that's why I used the word "also")

jqnatividad:
+1 to having a CLI command to clean orphaned datastore tables.

metaodi:
I created a "quick & dirty" paster command to empty the orphaned tables, which saves us a lot of space on our DB server: https://github.com/opendata-swiss/ckanext-switzerland/pull/137

It does not drop the tables, but simply calls the `datastore_delete` API. It's good enough for now for us.

cc @wardi @jqnatividad

amercader:
Fixed in #3425

jqnatividad:
Any chance the "quick & dirty" paster command to clean up orphaned datastore tables that @metaodi put together will be folded into core?

metaodi:
@jqnatividad if there is a need, I can create a PR with the command for this repository.

joeldouglass:
This appears to be fixed if you directly delete a resource. But if you delete and purge a data set (which also deletes its resources), the datastore tables are still there.

PR #3425, opened 2017-02-01
Datastore deleted inside `after_delete` of IResourceController

smotornyuk:
#3422 
After resource had been successfully removed:
1. Get all removed resources of parent dataset
2. Filter out those ones, without active datastore
3. Remove datastore for remaining(it will remove some old datastores first time and will only remove datastore of current resource in consequent runs, so no overhead here)
4. Catch `NotFound`. Resource was just deleted, so `datastore_delete`'s call to `resource_patch` will rise this error.
5. Manually update extras of removed dataset. Not sure whether it required, but this is expected behavior in some way, so just let's do it. Yeah, i directly update resource, but i definitely have access to this resource if `after_delete` have been called.

### DIFF
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -281,14 +281,13 @@ class DatastorePlugin(p.SingletonPlugin):
             if res.extras.get('datastore_active') is True]
 
         for res in deleted:
-            try:
-                logic.get_action('datastore_delete')(
-                    context, {'resource_id': res.id, 'force': True})
-            except logic.NotFound:
-                # resource was just deleted, so it's expected situation
-                res.extras['datastore_active'] = False
-                res_query.update(
-                    {'extras': res.extras}, synchronize_session=False)
+            db.delete(context, {
+                'resource_id': res.id,
+                'connection_url': self.write_url
+            })
+            res.extras['datastore_active'] = False
+            res_query.update(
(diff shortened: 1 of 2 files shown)


ckanext/datastore/plugin.py
```diff
@@ -265,6 +268,28 @@ def before_show(self, resource_dict):
...
+            model.Resource.state == State.DELETED
+        )
+        deleted = [
+            res for res in query.all()
+            if res.extras.get('datastore_active') is True]
+
+        for res in deleted:
+            try:
+                logic.get_action('datastore_delete')(
+                    context, {'resource_id': res.id, 'force': True})
```
wardi: We're inside the datastore code here, why not call the lower-level function to delete instead of the action?

smotornyuk:
Updated

camfindlay:
Which version of CKAN is this available from?

metaodi:
@camfindlay it's in CKAN 2.7.0 (see the merge commit https://github.com/ckan/ckan/commit/ef6cf50daec17dc56fa56eb4d3a3becbcc0afb2d)

camfindlay:
Will this fire the after_delete when a harvest "reharvest" is triggered? Testing this out at the moment on a fresh CKAN 2.7 and my datastore tables just continue to grow in size (ie. no clean out).

PR #3427, opened 2017-02-02
allow package_search to return id/name instead of full dataset dictionary

fanjinfei:
Fixes #

### Features:

- [ X] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1833,6 +1833,8 @@ def package_search(context, data_dict):
         The parameter that controls which fields are returned in the solr
         query cannot be changed.  CKAN always returns the matched datasets as
         dictionary objects.
+        if fl='id' or 'name', only this field in the dataset dictionary will
+        be returned.
     '''
     # sometimes context['schema'] is None
     schema = (context.get('schema') or
@@ -1875,8 +1877,10 @@ def package_search(context, data_dict):
         else:
             data_source = 'validated_data_dict'
         data_dict.pop('use_default_schema', None)
+
         # return a list of package ids
-        data_dict['fl'] = 'id {0}'.format(data_source)
+        if data_dict.get('fl') not in ('id', 'name'):
+            data_dict['fl'] = 'id {0}'.format(data_source)
 
         # Remove before these hit solr FIXME: whitelist instead
         include_private = asbool(data_dict.pop('include_private', False))
@@ -1903,21 +1907,25 @@ def package_search(context, data_dict):
         # Add them back so extensions can use them on after_search
         data_dict['extras'] = extras
 
-        for package in query.results:
-            # get the package object
-            package_dict = package.get(data_source)
-            ## use data in search index if there
-            if package_dict:
-                # the package_dict still needs translating when being viewed
-                package_dict = json.loads(package_dict)
-                if context.get('for_view'):
-                    for item in plugins.PluginImplementations(
-                            plugins.IPackageController):
-                        package_dict = item.before_view(package_dict)
-                results.append(package_dict)
-            else:
-                log.error('No package_dict is coming from solr for package '
-                          'id %s', package['id'])
+        if data_dict.get('fl') in ['id', 'name']:
+            for package in query.results:
+                results.append( {data_dict.get('fl'):package} )
+        else:
+            for package in query.results:
+                # get the package object
+                package_dict = package.get(data_source)
+                ## use data in search index if there
+                if package_dict:
+                    # the package_dict still needs translating when being viewed
+                    package_dict = json.loads(package_dict)
+                    if context.get('for_view'):
+                        for item in plugins.PluginImplementations(
+                                plugins.IPackageController):
+                            package_dict = item.before_view(package_dict)
+                    results.append(package_dict)
+                else:
+                    log.error('No package_dict is coming from solr for package '
+                              'id %s', package['id'])
 
         count = query.count
         facets = query.facets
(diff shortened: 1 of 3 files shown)


ckan/logic/action/get.py
```diff
@@ -1875,8 +1877,10 @@ def package_search(context, data_dict):
         else:
             data_source = 'validated_data_dict'
         data_dict.pop('use_default_schema', None)
+
         # return a list of package ids
-        data_dict['fl'] = 'id {0}'.format(data_source)
+        if data_dict.get('fl') not in ('id', 'name'):
+            data_dict['fl'] = 'id {0}'.format(data_source)
```
wardi: you're expecting the `fl` parameter to be a string value but a list seems like a better choice because solr allows multiple values for `fl`. If you add `fl` to the `default_package_search_schema` you can make sure the value gets converted to a list before it's used here.

ckan/logic/action/get.py
```diff
@@ -1903,21 +1907,25 @@ def package_search(context, data_dict):
...
-                    for item in plugins.PluginImplementations(
-                            plugins.IPackageController):
-                        package_dict = item.before_view(package_dict)
-                results.append(package_dict)
-            else:
-                log.error('No package_dict is coming from solr for package '
-                          'id %s', package['id'])
+        if data_dict.get('fl') in ['id', 'name']:
+            for package in query.results:
+                results.append( {data_dict.get('fl'):package} )
```
wardi: we should be able to support returning all fields listed in `fl` into the results list. That lets sites add their own values to solr in `before_index` then retrieve them quickly from here.

wardi:
@TkTech does this cause you any concern WRT your search refactoring work?

ckan/logic/action/get.py
fanjinfei: I noticed in ckan/ckan/lib/search/query.py class PackageSearchQuery:run() :   
```
            # if just fetching the id or name, return a list instead of a dict   
            if query.get('fl') in ['id', 'name']:   
                self.results = [r.get(query.get('fl')) for r in self.results]  
```
so we can be consistent with this?

wardi: That only lets us return one field in our results. Solr's `fl` parameter supports multiple fields, which is much more useful in general, e.g. to get a quick list of the ids, names and titles of matching datasets.

fanjinfei: I did a test. The solr 'fl' parameter does not support fields inside 'extras'. We may need to dictionize all the extras fields in plugin's after_search, then filter the result fields in ckan/package_search.

fanjinfei:
With this fix, in plugin's after_search(), the search_results['results']['extras'] may be dictionary or list, depending on the 'fl' value.

TkTech:
@wardi None at all, none of this still exists in the rework so there's no point thinking about it :) It is a very large rewrite.

ckan/logic/action/get.py
```diff
@@ -1875,8 +1876,12 @@ def package_search(context, data_dict):
...
             data_source = 'validated_data_dict'
         data_dict.pop('use_default_schema', None)
-        # return a list of package ids
-        data_dict['fl'] = 'id {0}'.format(data_source)
+
+        result_fl = data_dict.get('fl')
+        if not result_fl:
+            data_dict['fl'] = 'id {0}'.format(data_source)
+        else:
+            data_dict['fl'] = ' '.join(result_fl)
```
wardi: the solr API should accept a real list, doesn't it? converting it ourselves will cause strange behaviour if we generate something that doesn't parse as a real list.

wardi:
https://github.com/ckan/ckan/blob/master/ckan/logic/schema.py#L588 needs the `fl` parameter added, something like:
```
 'fl': [ignore_missing, list_of_strings],
```

ckan/logic/action/get.py
fanjinfei: No, class PackageSearchQuery:run() does not accept a real list 'fl'. It has to be a space separated fields name.

wardi: Does it work If we fix PackageSearchQuery at the same time?

fanjinfei: 'fl' argument is used by SolrConnection.raw_query() as a space separated fields string. If we pass a real list to PackageSearchQuery, it still need to be converted.

fanjinfei:
@wardi how to test default_package_search_schema() ?

wardi:
Ok, fl as a list might only be supported by solr 4+ https://wiki.apache.org/solr/CommonQueryParameters#fl so I guess we'll have to space or comma-separate the parameters. I notice that this lets users pass solr functions for subqueries. @TkTech any concerns here or should be continue to allow users to pass arbitrary fl values?

ckan/tests/logic/action/test_get.py
```diff
@@ -852,6 +852,17 @@ def test_search(self):
...
+    def test_search_fl(self):
+        factories.Dataset(title='Rivers', name='test_ri')
+        factories.Dataset(title='Lakes')
+
+        search_result = helpers.call_action('package_search', q='rivers', fl=['title', 'name'])
+
+        eq(search_result['results'][0]['title'], 'Rivers')
+        eq(search_result['results'][0]['name'], 'test_ri')
+        eq(len(search_result['results'][0]), 2)
+        eq(search_result['count'], 1)
```
wardi: you can replace these four lines with:
```python
eq(search_result['results'], [{'title': 'Rivers', 'name': 'test_ri'}])
```
for less code and a better error message when the test fails (the error will include the complete results list)

amercader:
Thanks for the patch @fanjinfei

PR #3428, opened 2017-02-07
datastore triggers

wardi:
1. Add an API call for syadmins (only) to define PL/pgSQL triggers in the datastore db
2. Extend datastore_create to allow applying those triggers when creating or modifying tables

Triggers are great for cleaning data, enforcing validation and/or keeping modification records as data is being loaded. Using triggers instead of validation logic at the python level lets us enforce those rules regardless of how the data is loaded (bulk loading data straight to postgres is very nice for large datasets)


wardi:
@amercader this is feature complete with test coverage now.

wardi:
For issues on #3414 please report them there and I'll fix on that branch and merge back into this one.

wardi:
need to fix quantified code error and exceptions aren't being caught by upsert.

wardi:
@amercader this is ready for review

ckanext/datastore/db.py
```diff
@@ -1173,6 +1195,37 @@ def create(context, data_dict):
...
+def _create_triggers(connection, resource_id, triggers):
+    u'''
+    Delete existing triggers on table then create triggers
+
+    Currently our schema requires "before insert or update"
+    triggers run on each row, so we're not reading "when"
+    or "for_each" parameters from triggers list.
+    '''
+    existing = connection.execute(
+        u'SELECT tgname FROM pg_trigger WHERE tgrelid = %s::regclass',
```
amercader: Should we use `datastore_helpers.identifier` here? Maybe not, I'm not familiar with `::regclass`

ckanext/datastore/logic/action.py
```diff
@@ -55,6 +55,10 @@ def datastore_create(context, data_dict):
     :type primary_key: list or comma separated string
     :param indexes: indexes on table (optional)
     :type indexes: list or comma separated string
+    :param triggers: triggers to apply to this table, eg: [
+        {"function": "trigger_clean_reference"},
+        {"function": "trigger_check_codes"}]
+    :type triggers: list of dictionaries
```
amercader: Perhaps add a note that says that these triggers can be created with ``:py:func:`datastore_function_create` `` so docs are linked. Are they limited to the ones created via this action? Or can you invoke triggers manually created on Postgres?

ckanext/datastore/db.py
wardi: the execute() call is quoting the resource_id parameter for us. I use `identifier` when I can't avoid hand-rolling sql.

ckanext/datastore/logic/action.py
wardi: Good suggestion. You can use any postgresql function that returns type `trigger`, whether created by datastore_function_create or manually.

wardi:
Thanks @amercader !

PR #3430, opened 2017-02-07
Query performance improvements.

TkTech:
Removes a forced order_by on revisions that causes poor query
performance on many other table queries.

Replaced `count()` queries on number_of_edits and number_of_packages
with vastly improved queries. SQLAlchemy uses subqueries for `count()`,
along with an odd default unindexed order_by on the table mapper
causing even trivial counts to perform unindexed table scans.

This change reduces the average query time for a user with 500k
revisions from 15 *seconds* to under 20ms.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/model/resource.py
+++ b/ckan/model/resource.py
@@ -189,7 +189,6 @@ meta.mapper(Resource, resource_table, properties={
                             ),
     )
 },
-order_by=[resource_table.c.package_id],
 extension=[vdm.sqlalchemy.Revisioner(resource_revision_table),
            extension.PluginMapperExtension(),
            ],
--- a/ckan/model/user.py
+++ b/ckan/model/user.py
@@ -9,7 +9,7 @@ import passlib.utils
 from passlib.hash import pbkdf2_sha512
 from sqlalchemy.sql.expression import or_
 from sqlalchemy.orm import synonym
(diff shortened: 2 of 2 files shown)


TkTech:
There is a second PR that is part of this series that adds/corrects migrations for poor indexes as well as a performance bug in the `{% snippet %}` jinja extension. We've managed to eliminate all queries over 1s and reduce page loads to under 400ms without caching.

TkTech:
Commit 2e3559a saves us about 200ms per 10 resources on our site, which ends up being 1/4 of the total page time.

Instead of checking if we can edit the package each time we display an individual resource, check it once and pass it through.

TkTech:
@amercader this PR is ready :)

amercader:
Great stuff @TkTech!

MrkGrgsn:
@amercader and @TkTech, will this PR or #3466 get backported to 2.5 or 2.6? We have clients' repositories with high numbers of revisions that would benefit greatly from these performance improvements. We could  probably get @smotornyuk to work on it.

TkTech:
@MrkGrgsn they're both small changes, you should be able to cleanly cherry pick them into your clients fork without running into conflicts.

I'm okay with having them backported since they change no behavior, @amercader?

PR #3435, opened 2017-02-13
Fix group, org member tables in mobile view

wildcatzita:

### DIFF
--- a/ckan/templates/group/members.html
+++ b/ckan/templates/group/members.html
@@ -8,15 +8,12 @@
 
 {% block primary_content_inner %}
   <h3 class="page-heading">{{ _('{0} members'.format(c.members|length)) }}</h3>
-  <table class="table table-header table-hover table-bordered" id="member-table">
-    <col width="70" />
-    <col width="40" />
-    <col width="20" />
+  <table class="table table-header table-hover table-bordered group-org-members" id="member-table">
     <thead>
       <tr>
-        <th scope="col">{{ _('User') }}</th>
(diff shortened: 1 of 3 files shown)


MrkGrgsn:
This new CSS will potentially alter the styling of more tables than you intend, since it will apply to the first 3 columns of any table that uses `col` elements. You need to use targeted CSS selectors, altering the template  to make if required (e.g., adding a class to the `col`'s or `table`)

wildcatzita:
@MrkGrgsn, I've added class in templates instead col tag.

smotornyuk:
@wildcatzita , i just noticed, that you'll get the same effect if you omit with attribute on col tags, even without adding anything to css. Could you check this and update you PR if i right?

wildcatzita:
@smotornyuk, fixed. Can yo check please?

ckan/templates/group/members.html
```diff
@@ -8,15 +8,12 @@
...
+  <table class="table table-header table-hover table-bordered group-org-members" id="member-table">
     <thead>
       <tr>
-        <th scope="col">{{ _('User') }}</th>
-        <th scope="col">{{ _('Role') }}</th>
-        <th scope="col"></th>
+        <th class="col-user">{{ _('User') }}</th>
+        <th class="col-role">{{ _('Role') }}</th>
+        <th class="col-actions"></th>
       </tr>
```
smotornyuk: Are you still need these classes?

smotornyuk:
I think it can be merged now

PR #3437, opened 2017-02-14
Alterations inside datastore for further integration of third party backends

smotornyuk:
Trying to add support of MongoDB for data persistence instead of PostgreSQL, i faced a lot of problem. The main one - i need to rewrite(or copy) everything from datastore, but in that case i won't receive any security updates of bug fixes from PRs to datastore. So i suggest to make changes inside datastore instead. Move any pgsql-specific code to separate file and enable it by adding as inner `backend` property into datastore extension. In future other extension will be allowed to replace backend by implementing `IDatastoreBackend`. Which one will be used - it's still a question. Probably it can be decided either by setting additional variable in config file, or, even better, by first part of write_url (e.g 'postges://...', 'mongodb://..')

Work still in progress. Currently just plugin.py changed. Next target is actions.py


ckanext/datastore/interfaces.py
```diff
@@ -145,3 +145,25 @@ def datastore_delete(self, context, data_dict, fields_types, query_dict):
...
+
+
+class IDatastoreBackend(interfaces.Interface):
+    """Allow custom implementations of datastore backend"""
+    def register_backends(self):
+        """
+        Register classes that inherits from DatastoreBackend.
+
+        Every registered class provides implementations of DatastoreBackend
+        and, depending on `datastore.write_url`, one of them will be used
```
amercader: Is there any check on `datastore.read_url`? eg could you end up using a different backend for read operations (search) and write ones?

ckanext/datastore/backend/example.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
```
amercader: Can you rename this file to `example-sqlite.py`?

amercader:
@smotornyuk I think this is great work that opens a lot of possibilities for the Datastore. I think is worth cleaning it up a bit and get the conflicts fixed.
@wardi WDYT?

wardi:
@amercader absolutely agree. There are plenty of potential ckan users that have a favourite data store that this would open us to.

ckan/tests/test_coding_standards.py
```diff
@@ -587,6 +587,7 @@ def find_unprefixed_string_literals(filename):
     u'ckanext/datastore/controller.py',
     u'ckanext/datastore/db.py',
     u'ckanext/datastore/backend/postgres.py',
+    u'ckanext/datastore/backend/example-sqlite.py',
```
torfsen: This list is for existing source files that have not yet been converted to the new standard. Please do not add new files to the list, instead make sure that they follow our [rules for Unicode handling](http://docs.ckan.org/en/latest/contributing/unicode.html) so that they pass the tests.

smotornyuk:
@amercader , i added [assertion](https://github.com/ckan/ckan/pull/3437/files#diff-e5334138b119ef88fd90b9c590c5ba0bR91) in order to prevent using different schemas for read and write urls.

ckan/tests/test_coding_standards.py
```diff
@@ -586,12 +586,15 @@ def find_unprefixed_string_literals(filename):
     u'ckanext/datapusher/tests/test_interfaces.py',
     u'ckanext/datastore/controller.py',
     u'ckanext/datastore/db.py',
+    u'ckanext/datastore/backend/postgres.py',
```
torfsen: Again, please do not add new files to the whitelist :wink:

Simply prefix all string literals with either `u` or `b` [as stated by the rules](http://docs.ckan.org/en/latest/contributing/unicode.html).

smotornyuk:
`ckanext/datastore/backend/postgres.py` isn't a new file - it's just renamed(and ok, slightly modified `db.py`). There are more than 300 so it would be overkill, if i change them all.

torfsen:
@smotornyuk great!

amercader:
@smotornyuk I think all of @wardi big PRs that touched the DataStore are now merged so now is a great time to get these conflicts fixed so it can get a final review and probably merged for 2.7.

smotornyuk:
ok, i'll try to resolve them this week

smotornyuk:
Hm, i think it works:)

amercader:
Thanks @smotornyuk, I've been playing with this and I think it looks great and has a lot of potential. From my side it looks good to merge and could be part of 2.7.

Two things I think that need to be done (but can be done on separate PRs):

* A small set of tests for the example sqlite backend, to ensure that the interface keeps working on other backends that the default postgres one over time.
* Some documentation [here](http://docs.ckan.org/en/latest/maintaining/datastore.html). A small section about custom backends, explaining what to implement and how (doesn't need to be really thorough, just the basics and point to the example in the source code).

@wardi would be good if you could have a look as well as you've been doing more datastore-y things recently.

wardi:
datastore_helpers.identifier and literal_string should be moved to the postgres backend, since they implement SQL quoting that works with postgres. This looks good so far though..

wardi:
@smotornyuk would you mind adding the work that @amercader mentioned: sqlite example tests (just the minimum to see that it works at all) and some description in the docs. I think we need that if we're releasing this feature in 2.7.

smotornyuk:
ok, i'll add them

smotornyuk:
Done. I decided to add new example extension and put custom backend there. There are not so much tests, they just check correctness of backend behavior.

wardi:
@smotornyuk great work!

PR #3440, opened 2017-02-16
[DOC] Added private parameter on package create method.

Germanaz0:


ckan/logic/action/create.py
```diff
@@ -60,6 +60,8 @@ def package_create(context, data_dict):
     :param title: the title of the dataset (optional, default: same as
         ``name``)
     :type title: string
+    :param private: If ``True`` creates a private dataset
+    :type private: string
```
wardi:
```
:type private: bool
```

otherwise great!

Germanaz0: Ok fixing now

Issue #3447, opened 2017-02-20
Resource creation date column don't use datetime.utcnow()


Engerrs:
### CKAN Version if known (or site URL)
Current version
### Expected behaviour
When creating resource for dataset and entering to it summary page, in additional information table, there are two dates of Creation. One is showing only month,day,year. Second one is showing when we click on 'Show more' button under the table, this field should show how much time ago the resource was created, for example, if it was created just now, it should display 'Just now'.

### Actual behaviour
When we create new resource for dataset, on resource summary page in additional information table, when we click 'Show more', field 'created' shows '-1 days ago'.

### What steps can be taken to reproduce the issue? 
1.Create dataset.
2.Create resource for this dataset.
3.Enter resource summary page.
4.Click 'Show more' under the Additional info table.
5.'created' field shows '-1 days ago'.

PR #3448, opened 2017-02-20
[#3447] Resource creation date use datetime.utcnow()

Engerrs:
### Fix:
Added datetime.utcnow() to the 'created' column, so it won't show now '-1 days ago' at additional information table.


wardi:
This is clearly correct. Thanks @Engerrs !

amercader:
Backport?

PR #3449, opened 2017-02-20
[#3443] Raise exception if Celery version is not supported

### LINKED ISSUE #3443: Missing requirement for Celery
`requirements.in` and `requirements.txt` currently don't contain an entry for `celery`. While use of Celery [is optional](https://github.com/ckan/ckan/blob/727fe76a4da67fb590e87ce75c5284f145a6e495/ckan/model/__init__.py#L191) it is the [documented background job system for CKAN](http://docs.ckan.org/en/ckan-2.6.0/maintaining/background-tasks.html) as of CKAN 2.6.

The problem is that CKAN's Celery code is incompatible with newer versions of Celery. For example, `pip install celery` pulls in 4.0.2, and using that within CKAN raises the following exception when initializing the database:
```$ /usr/lib/ckan/default/bin/paster --plugin=ckan db init -c /etc/ckan/default/production.ini
  File "/usr/lib/ckan/default/bin/paster", line 11, in <module>
    sys.exit(run())
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 102, in run
    invoke(command, command_name, options, args[1:])
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 141, in invoke
    exit_code = runner.run(args)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 236, in run
    result = self.command()
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/cli.py", line 223, in command
    model.repo.init_db()
  File "/usr/lib/ckan/default/src/ckan/ckan/model/__init__.py", line 192, in init_db
    import ckan.lib.celery_app as celery_app
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/celery_app.py", line 65, in <module>
    celery.loader.conf.update(default_config)
AttributeError: 'NoneType' object has no attribute 'update'
```
Reverting to the latest Celery 3.x version (3.1.25) solves the problem.

We should therefore add a requirement for Celery and pin it to version 3.1.25. Given that CKAN 2.7 introduces a new system based on RQ and deprecates the Celery system I don't think we need to update our Celery code to work with Celery 4.x.

torfsen:
Fixes #3443 by checking Celery's version at import time. Unsupported versions (anything other than `3.x`) raise an `ImportError` and therefore behave as if Celery wasn't installed at all in places where it is optional. In places where Celery is mandatory the `ImportError` informs the user about the real problem.

### DIFF
--- a/ckan/lib/celery_app.py
+++ b/ckan/lib/celery_app.py
@@ -10,9 +10,13 @@ import ConfigParser
 import logging
 import os
 
-from celery import Celery
 from ckan.common import config as ckan_config
 from pkg_resources import iter_entry_points, VersionConflict
+
+from celery import __version__ as celery_version, Celery
+if not celery_version.startswith(u'3.'):
+    raise ImportError(u'Only Celery version 3.x is supported.')
+
 
 log = logging.getLogger(__name__)
 
--- a/ckan/tests/lib/test_celery_app.py
+++ b/ckan/tests/lib/test_celery_app.py
@@ -1,0 +1,27 @@
+# encoding: utf-8
+
+u'''
+Tests for ``ckan.lib.celery_app``.
+'''
+
+from nose.tools import raises
+import mock
+
+
+class TestCeleryVersion(object):
+    u'''
+    Make sure that Celery's version is checked.
+    '''
+    @mock.patch.dict(u'sys.modules', {u'celery': mock.MagicMock()})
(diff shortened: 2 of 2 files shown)


amercader:
Thanks @torfsen

Issue #3281, opened 2016-10-18
user 'manage' dialog usability


rleir:
### CKAN Version if known (or site URL)

dev-v2.6 http://beta.ckan.org/user/edit/rleir#markdown
### Please describe the expected behaviour

A minor quibble. The help text should match the linking policy. Either we allow links (preferred), as shown in the help text, or we disallow them and take them out of the help text.
### Please describe the actual behaviour

Add a link as shown in the help text, and get an error 'The form contains invalid entries: About: Edit not allowed as it looks like spam. Please avoid links in your description.' 
### What steps can be taken to reproduce the issue?

Open your user 'manage' dialog. Click on 'Markdown formatting'. It suggests you can use the format 'http://auto.link.ed/' for a link. Edit such a link into your 'About' text, click 'Update Profile'. You get the error message.

gleb-rudenko:
@amercader , I'll work on it.

gleb-rudenko:
@amercader , ready for review.

gleb-rudenko:
@amercader, any progress with this one?

amercader:
@gleb-rudenko Is there a PR for this?

gleb-rudenko:
@amercader, I guess https://github.com/ckan/ckan/compare/master...gleb-rudenko:user-manage-dialog-usability

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #3452, opened 2017-02-21
#3281 / removing validation rule for about page

gleb-rudenko:
Fixes #
Removed the validation rule for the about field, so now you can post links.
### Features:

- [ ] includes API changes


TkTech:
This needs to be a configuration option. Most (all?) sites with public registration will not want this, such as the demo.ckan site.

gleb-rudenko:
@TkTech , the validator is back again, in additional, added config to allow posting links. Also noted in the tooltip link that you can use links on some markdown fields only if CKAN configured

TkTech:
👍

ckan/logic/validators.py
```diff
@@ -607,8 +607,9 @@ def user_password_not_empty(key, data, errors, context):
             errors[key].append(_('Missing value'))
 
 def user_about_validator(value,context):
-    if 'http://' in value or 'https://' in value:
-        raise Invalid(_('Edit not allowed as it looks like spam. Please avoid links in your description.'))
+    if not config.get('ckan.markdown_filter_links', False):
```
wardi: Let's keep the old default. It makes sense for sites that allow anonymous users to sign up

Issue #2651, opened 2015-09-18
delete resource on draft dataset should redirect to new_resource page


wardi:
currently it breaks out of the new-dataset-creation process and drops users at the dataset show page

wildcatzita:
@wardi, fixed redirect. Can you check please?

wardi:
fixed by #3454

PR #3454, opened 2017-02-21
[#2651] Fix for delete resource on draft dataset should redirect to new_resource page

wildcatzita:


wildcatzita:
@wardi, fixed.

PR #3464, opened 2017-03-02
Fix an incorrect error message.

TkTech:
This is a small legacy codefix.

If a sub-template cannot be found anywhere inside an IDatasetForm template assumptions are made in the displayed error message as to which template failed. This PR turns:

```
Viewing record datasets in <built-in function format> format is not supported (template file scheming/package/read.html not found).
```

into:

```
Viewing datasets of type "record" is not supported ('Template scheming/display_snippets/text cannot be found').
```

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -390,11 +390,14 @@ class PackageController(base.BaseController):
         try:
             return render(template,
                           extra_vars={'dataset_type': package_type})
-        except ckan.lib.render.TemplateNotFound:
-            msg = _("Viewing {package_type} datasets in {format} format is "
-                    "not supported (template file {file} not found).".format(
-                        package_type=package_type, format=format,
-                        file=template))
+        except ckan.lib.render.TemplateNotFound as e:
+            msg = _(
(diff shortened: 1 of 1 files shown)


TkTech:
Merging this tomorrow if there are no objections by then.

wardi:
This change still hides the real error you were seeing: A sub-sub-template was missing. Can we check for the topmost templates' existence instead of hijacking every template not found error here?

TkTech:
@wardi... sorry? I might be misunderstanding but this explicitly *doesn't* do that? It's kinda the entire point of the PR.

wardi:
You're right, this change does now show which template was missing.

For missing sub-templates though this still renders the response as a 404 with a reference to a file that the end user has no access to in the message. Ideally we would let the TemplateNotFound exception be raised in those cases so that the developers get notified that they did something wrong.

TkTech:
I'm merging this as-is for now, since it's less-broken then what's already there. Don't have time to improve the existing behaviour.

wardi:
"less-broken than what's already there" is not wrong

PR #3466, opened 2017-03-04
Performance improvements in group_activity_query

TkTech:
Currently, attempting to view the activity feed (a very user-visible feature) of an organization or group with even a few thousand packages will take between 10s to never. This is caused by poor logic inside `group_activity_query` which among other things loads every single dataset owned by that group/org into memory, gets the IDs, throws them out, and then does a gigantic `IN` query against every `object_id`.

With this change, rendering org activity with 18000 datasets takes <= 700ms (with no query being above 250ms) and most of that time can be reduced through two changes:

- The controller should be changed to sort and page based off the `Activity.timestamp` column, which should also be indexed. This solves the offset/limit performance once you're many pages in.
- Template improvements as a significant amount of time is spent rendering each activity item.

However, these two changes may break existing themes and will be left for another PR. This PR as it is causes no breaking changes and can be safely backported.

### DIFF
--- a/ckan/model/activity.py
+++ b/ckan/model/activity.py
@@ -10,6 +10,7 @@ from sqlalchemy import (
     ForeignKey,
     desc,
     or_,
+    and_,
     union_all
 )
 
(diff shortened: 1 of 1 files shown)


amercader:
Tests seem to suggest that some activities are being missed with these changes (or the tests may be fragile). Can you check?

TkTech:
Haven't been able to figure it out yet. Logically the behaviours are identical. The tests fail differently locally and need to be run as the entire set since it fails differently depending on adjacent tests which is making debugging this slow.

TkTech:
Ian clued me into the fact that the tests on circleci are using segments to divide the tests among workers, causing different behaviour than when they're run all together.

TkTech:
image

The tests should be passing now. With an example organization with 60k records our time spent in queries is <= 250ms with the majority of the 1s page time spent in rendering templates.

wardi:
@smotornyuk gives this a :+1: merging

PR #3467, opened 2017-03-06
faster datastore total calculation + make optional

wardi:
This PR separates the total rows calculation from the select statement in datastore_search, which for large tables can give a significant speed improvement. It also adds an `include_total` parameter to datastore_search (default True for backwards compatibility) that allows skipping the calculation completely.


amercader:
This looks great but can you merge master @wardi? There is a conflict involving the new stuff you've worked on that I'd prefer you checked first.

wardi:
@amercader done

Issue #3471, opened 2017-03-07
Small design issues on Create Dataset page and my organization, my groups tabs on Dashboard page.


Engerrs:
### CKAN Version
2.5 , 2.7

### Issue
On Create dataset page, license field and info-block not positioned on one line, witch also creates more space between license filed and organization field.
![screenshot 1](https://cloud.githubusercontent.com/assets/13872453/23661397/575bcdb6-0355-11e7-8f23-412a7bb00674.jpg)

On Dashboard page, my organizations and my groups tabs, blocks go beyond the boundaries of the parent block.
![screenshot](https://cloud.githubusercontent.com/assets/13872453/23661609/f6a76dda-0355-11e7-9c5e-8a51a61ba76f.jpg)

Engerrs:
Created a fix for those two issues.

PR #3472, opened 2017-03-07
[#3471] Fix for small issues and Create dataset page and Dashboard tabs

Engerrs:
Fixes #
Fixed the position of info-block and tabs my organization and my groups, on Dashboard page


smotornyuk:
Fix for media-greed looks ok, but as for moving info-block - i don't think that this is a good idea. At least info-block [displayed on new line by default](https://github.com/ckan/ckan/blob/d73326ad4beb3743cb4a347b15c04dace8c8654b/ckan/public/base/less/forms.less#L152)

PR #3476, opened 2017-03-09
Much better handling of metadata_modified.

### LINKED ISSUE #3475: metadata_modified is inconsistently set & updated

All

On any kind of package change, we should see a correctly updated UTC timestamp in metadata_modified.

metadata_modified becomes `null` (creation) or does not reflect the actual time of insertion (update).

Create a package.

TkTech:
Fixes #3475.

See existing issue for details.

### DIFF
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -290,8 +290,6 @@ def package_dict_save(pkg_dict, context):
 
     if 'metadata_created' in pkg_dict:
         del pkg_dict['metadata_created']
-    if 'metadata_modified' in pkg_dict:
-        del pkg_dict['metadata_modified']
 
     pkg = d.table_dict_save(pkg_dict, Package, context)
 
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -294,11 +294,6 @@ def package_update(context, data_dict):
     else:
         rev.message = _(u'REST API: Update object %s') % data.get("name")
 
-    #avoid revisioning by updating directly
-    model.Session.query(model.Package).filter_by(id=pkg.id).update(
-        {"metadata_modified": datetime.datetime.utcnow()})
-    model.Session.refresh(pkg)
-
     pkg = model_save.package_dict_save(data, context)
 
     context_org_update = context.copy()
--- a/ckan/migration/versions/086_improve_metadata_modified.py
+++ b/ckan/migration/versions/086_improve_metadata_modified.py
@@ -1,0 +1,55 @@
+# encoding: utf-8
+
+# -*- coding: utf-8 -*-
+
+
+def upgrade(migrate_engine):
+    update_default_q = (
+        'ALTER TABLE ONLY package ALTER COLUMN metadata_modified '
+        'SET DEFAULT (statement_timestamp() at time zone \'utc\');'
+    )
+
+    add_trigger_q = """
+CREATE OR REPLACE FUNCTION update_metadata_modified() RETURNS TRIGGER
+LANGUAGE plpgsql
+AS
+$$
+BEGIN
+    NEW.metadata_modified = (CURRENT_TIMESTAMP at time zone 'utc');
+    RETURN NEW;
+END;
+$$;
+
+DROP TRIGGER IF EXISTS update_metadata_modified_t on "package";
+CREATE TRIGGER update_metadata_modified_t
+  BEFORE UPDATE
+  ON package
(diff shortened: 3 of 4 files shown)


wardi:
This would prevent setting metadata_modified with package_update, which I think is something we want to have when migrating and harvesting datasets.

I'm also worried about taking on some extra complexity by updating just this one field with postgres default function and trigger, when all the other metadata fields are set with python code.

wardi:
If we *do* go with this approach we'd also need to update the metadata_modified when resources belonging to the package are modifies.

TkTech:
> I'm also worried about taking on some extra complexity by updating just this one field with postgres default function and trigger, when all the other metadata fields are set with python code.

The trigger can be rewritten to handle all the sub-tables with minimal repetition. `metadata_created` and `metadata_modified` are special and should never be user editable.

> This would prevent setting metadata_modified with package_update, which I think is something we want to have when migrating and harvesting datasets.

This is the current reason I have left `metadata_created` alone, since it's being used for such purposes. However, using both `metadata_created` and `metadata_modified` like this is quite wrong. We absolutely need to have proper solid values for both fields for internal usage.

Keep in mind that you cannot currently correctly set either field reliably from create and any value you add will be trampled.

If you want to track an arbitrary "origin" created/modified date from harvesting, this is what extras are perfect for.

torfsen:
> `metadata_created` and `metadata_modified` are special and should never be user editable.

I beg to differ. Anytime I want to synchronize metadata from another source (e.g. harvesting) I want to synch the metadata timestamps, too. In my opinion there's a clear difference between modifying what is stored in CKAN's DB and modifying the metadata itself (i.e. the "virtual" data about the data which is not tied to a particular storage). In my understanding, `metadata_created` and `metadata_modified` should talk about the latter, while you seem to favor the former.

Obviously there is a certain overlap: most of the time if I edit the stored metadata in CKAN that's equivalent to changing the virtual metadata and therefore `metadata_modified` should be updated. But in certain situations (harvesting, etc.) this is not the case, so it should be possible to prevent an automatic update of `metadata_modified` and to set both timestamps "manually".

In my opinion CKAN would benefit greatly from a unified and consistent metadata concept that applies to all its entities and which documents exactly how CKAN handles the issues outlined above. Being able to clearly document when and how data and metadata have changed is very important for us, for example to reduce the risk of liability suits w.r.t. outdated data.

TkTech:
I think I explained it poorly.

`metadata_created`, and `metadata_modified` are two core fields on the package record itself. I believe (and these values are mostly treated this way already) that these should be exclusively treated as the timestamps on *CKANs* metadata.

Like I suggested, in your exact scenario you would simply store something like `origin_modified` and `origin_created` in the extras.

wardi:
At one point metadata_modified was a property on the model that looked up the last time stamp in the revision table. We discussed this at length in a dev meeting and changed metadata_modified to a proper column so that we could do the sort of thing @torfsen is talking about. It's really useful to let users control this field, in particular because it's a default sort for datasets on the search page.

The metadata_modified you're looking for (date last modified on this particular ckan instance) can be found in activities and revisions whenever it's needed.

I'd like to just fix the bug that's causing this field to be None when it's not specified by the user.

TkTech:
> `The metadata_modified you're looking for (date last modified on this particular ckan instance) can be found in activities and revisions whenever it's needed.`

We never record activities for private datasets. Having to lookup revisions (which we're trying to remove) just to find recently modified packages isn't great since you'll have to JOIN and ORDER BY. It's the kind of field that really should be unrolled into the object.

I just don't agree with using these two fields instead of extras, which seem ideal for storing arbitrary values from the origin. Up to you two. We can close this PR.

amercader:
I agree both with

1. That `metadata_modified` should be editable.
2. That we need a straight-forward, read-only way of finding out when a dataset was last modified *in a particular CKAN instance* (so the `created` , `modified` fields that @TkTech mentions, which in most cases will have the same value as `metadata_created`  and `metadata_modified`)

So let's close this PR and tackle it in a new one.

We still need to find out a way to prevent the original issue of `metadata_modified` being empty of course.

wardi:
According to https://github.com/ckan/ckan/issues/3768 setting the metadata_modified on a dataset isn't actually allowed at the moment. @amercader is on the side of having separate user-modifiable fields and leaving metadata_modified and metadata_created as read-only. (we could call these new fields created and last_modified to match the read/write ones in resources)

Reopening..

Vanuan:
Can we also make it sortable, i.e. in datasets list?

PR #3479, opened 2017-03-13
[#3474]fixing XML dump

### LINKED ISSUE #3474: Datastore resource dump as XML

Latest version

### Please describe the expected behavior

Exporting resource as XML

### Please describe the actual behavior
if columns in datastore contain only numeric values or there are spaces in between , XML is not well formatted

adding additional checks in `ckan/ckanext/datasore/writer.py`

tino097:
Fixes #3474 

Exporting resource as XML dump now is well formatted XML.

### DIFF
--- a/ckanext/datastore/writer.py
+++ b/ckanext/datastore/writer.py
@@ -9,7 +9,8 @@ import unicodecsv
 
 UTF8_BOM = u'\uFEFF'.encode(u'utf-8')
 # Element names can contain letters, digits, hyphens, underscores, and periods
-SPECIAL_CHARS = u'#$%&!?\/@'
+SPECIAL_CHARS = u'#$%&!?\/@}{[]'
+
 
 @contextmanager
 def csv_writer(response, fields, name=None, bom=False):
--- a/ckanext/datastore/writer.py
+++ b/ckanext/datastore/writer.py
@@ -8,7 +8,8 @@ from xml.etree.cElementTree import Element, SubElement, ElementTree
 import unicodecsv
 
 UTF8_BOM = u'\uFEFF'.encode(u'utf-8')
-
+# Element names can contain letters, digits, hyphens, underscores, and periods
+SPECIAL_CHARS = u'#$%&!?\/@'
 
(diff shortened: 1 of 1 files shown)


ckanext/datastore/writer.py
```diff
@@ -161,9 +161,20 @@ def writerow(self, row):
...
+            if unicode(k).isnumeric():
+                k = u'_' + unicode(k)
             if v is None:
                 SubElement(root, k).attrib[u'xsi:nil'] = u'true'
                 continue
             SubElement(root, k).text = unicode(v)
         ElementTree(root).write(self.response, encoding=u'utf-8')
         self.response.write(b'\n')
+
+    def getXmlElement(self, element_name):
```
wardi: `snake_case` is the convention for methods and functions. I'd also move this outside of the class and make it a simple function because `self` is not used anywhere in the body

wardi: Also why not add the `isnumeric` check and `_`-prepending into this function. Maybe call it `clean_xml_element_name(element_name)` or similar.

wardi:
Is prefixing with `_` not part of getting a valid element name? If it is please move that into the function as well and I'll merge.

tino097:
i will do that

tino097:
i assumed that those column names would be accepted when the file is uploaded to datastore. E.g `2!` so im not checking all other cases. But if you think that could be a real scenario, i should add more checks in the method

i haven't notice that we had those test locally, for future commits i will use it as must.

wardi:
I still see a problem with get_xml_element(). e.g. looks like the string "1 2" would return "12"

Would you add some tests for this function so that it stays working in the future?

tino097:
@wardi i think it is expected as the element name could not contain spaces. I was looking to add some tests as next step

wardi:
I don't understand. I can make a column named "1 2 3" and then this function will produce an element name of "123". If you move the numeric check to the end it would be "_123" which is a valid element name.

kmbn:

PR #3481, opened 2017-03-14
Setting of datastore_active flag moved to separate function

janantk:
Changes from #3331 put into separate function.

Because the same approach is needed not only when creating new DataStore item (fixed in #3331), but also when deleting one (and setting datastore_active to False).

### DIFF
--- a/ckanext/datastore/logic/action.py
+++ b/ckanext/datastore/logic/action.py
@@ -149,47 +149,7 @@ def datastore_create(context, data_dict):
         log.debug(
             'Setting datastore_active=True on resource {0}'.format(resource.id)
         )
-        # issue #3245: race condition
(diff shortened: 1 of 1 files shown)


ckanext/datastore/logic/action.py
```diff
@@ -598,6 +556,53 @@ def datastore_make_public(context, data_dict):
     db.make_public(context, data_dict)
 
 
+def set_datastore_active_flag(model, data_dict, flag):
+    '''
+    Set appropriate datastore_active flag on CKAN resource.
+
+    Called after creation or deletion of DataStore table.
```
wardi: let's keep the link to `issue #3245: race condition` in the description to explain why we're doing this direct update of the ckan models ourselves for now.

PR #3485, opened 2017-03-15
Completely remove the revision controller.

### LINKED ISSUE #3484: Removal of deprecated revision controller.
The revision UI (/revision/list, /revision/{id}, /revision/edit/{id}) has long been unsupported. For several release versions the UI has not even been [themed correctly](http://demo.ckan.org/revision/list). I'm proposing the immediate removal of this controller and associated templates.

1) The UI is not "useful". When even small sites generate millions of revisions showing 12000+ pages is not great. The fields on individual revision details have not displayed properly in years. Most comments will always either be blank or "REST API" (which is also a bug).
2) The page performs very, very poorly and a bot crawling all pages is responsible for disabling large CKAN instances.
3) The end-user functionality of this page is provided by activity feeds.
4) ~No~ Limited test coverage.

TODO:

- [x] Remove controller, templates, routes, tests.
- [ ] Remove translation strings.
- [x] Update the `controllers/package.py:history` view to use activities instead of the revision controller diff. Or just remove it entirely.

TkTech:
Fixes #3484, see issue for details.


TkTech:
Commit cf9b198 will be the last commit in this series that will work without database migrations.

TkTech:
Friendly reminder that writing stateful tests will make Santa deliver coal for Christmas. 😭

ckan/controllers/package.py
```diff
@@ -400,87 +394,7 @@ def read(self, id):
...
-                )
-            response.headers['Content-Type'] = 'application/atom+xml'
-            return feed.writeString('utf-8')
-
-        package_type = c.pkg_dict['type'] or 'dataset'
-
-        return render(
-            self._history_template(c.pkg_dict.get('type', package_type)),
-            extra_vars={'dataset_type': package_type})
+        h.redirect_to(controller='pacakge', action='activities', id=id)
```
davidread: 'pacakge' typo

davidread:
This PR look promising :)

TkTech:
I forget why I paused on this. If I remember correctly I think I couldn't get the tests working although in hindsight it might actually have been circle.ci that was failing.

ckan/controllers/package.py
```diff
@@ -1277,18 +1191,24 @@ def activity(self, id):
...
             abort(404, _('Dataset not found'))
         except NotAuthorized:
             abort(403, _('Unauthorized to read dataset %s') % id)
 
-        return render('package/activity.html',
-                      {'dataset_type': dataset_type})
+        return render(
+            'package/activity.html',
+            extra_vars={
+                'dataset_type': dataset_type,
```
amercader: I see `dataset_type` being passed as extra var to templates in several places but I can't find where it is used on the actual templates. Is it necessary?

ckan/controllers/package.py
```diff
@@ -343,29 +341,21 @@ def read(self, id):
...
             c.pkg = context['package']
+
+            if activity_id:
+                c.pkg_dict = context['session'].query(model.Activity).get(
+                    activity_id
+                ).data['package']
+                # Don't crash on old activity records, which do not include
+                # resources or extras.
+                c.pkg_dict.setdefault('resources', [])
+                c.is_activity_archive = True
```
amercader: We are actively trying to avoiding setting new things into `c` (or Flask's `g`), can you pass this explicitly as an extra var?

ckan/templates/snippets/activity_stream.html
```diff
@@ -0,0 +1,51 @@
...
+{% endmacro %}
+
+{% macro dataset(activity) %}
+<span class="dataset">
+  {{ h.dataset_link(activity.data.package) }}
+</span>
+{% endmacro %}
+
+{% macro organization(activity) %}
+{{ h.organization_link(activity.data.group) }}
```
amercader: If I'm not mistaken this uses the Org / group name, not the title. Would it be possible to use the title? Datasets seem to use the title. And I understand that this is very likely how it was implemented before your changes, just wondering

ckan/controllers/package.py
wardi: ckanext-scheming and any site that uses multiple IDatasetForm plugins use this value, but it's not used by the core templates because they only support a single type

ckan/controllers/package.py
wardi: ckanext-scheming and any site that uses multiple IDatasetForm plugins use this value, but it's not used by the core templates because they only support a single type

ckan/model/package.py
```diff
@@ -527,17 +527,40 @@ def activity_stream_item(self, activity_type, revision, user_id):
...
-            d = {'package': dictization.table_dictize(self,
-                context={'model': ckan.model})}
-            return activity.Activity(user_id, self.id, revision.id,
-                    "%s package" % activity_type, d)
+            # We save the entire rendered package dict so we can support
+            # viewing the past packages from the activity feed.
+            dictized_package = ckan.logic.get_action('package_show')({
+                'model': ckan.model,
+                'session': ckan.model.Session,
+                'for_view': True,
```
wardi: for_view is a thing that makes me sad. IIRC it will do things like translate strings into the current user's language which doesn't seem appropriate for storing as a permanent record in the activity. Is there a reason it's required here?

ckan/templates/snippets/activity_stream.html
```diff
@@ -0,0 +1,51 @@
...
+<span class="actor">
+  {{ h.tag_link(activity.data.tag) }}
+</span>
+{% endmacro %}
+
+{% macro group(activity) %}
+<span class="group">
+  {{ h.group_link(activity.data.group) }}
+</span>
+{% endmacro %}
```
wardi: This looks like these macros make it easy and efficient to override the style of the links created, just by extending this template. Very nice

TkTech: Yup, record activities of any type you want and just add a template snippet to render them in your extension (or in core). The goal for activities is to be as simple as possible to use.

ckan/controllers/package.py
amercader: But do they use it on the actual `package/activity.html` template? Why do you need to pass it to the activity templates in particular? Or is it just a convention in case someone wants to use it on their own custom `package/activity.html`?

TkTech: I include it because that's what it was doing previously, and didn't want to remove it and potentially break templates in ways I did not realize. If it can be removed that's fine with me :)

Since the entire dict is now in an activity you can switch based off the type field in the package instead.

davidread: I agree with leaving it for customized activity pages.

ckan/controllers/package.py
davidread: I've addressed this in #3972 now

Issue #3478, opened 2017-03-12
password reset email omits the FQDN


MartinHowitt:
### CKAN Version if known (or site URL)
2.6.1 http://libraries.thedata.place
Source install as per instructions

### Please describe the expected behaviour
I expected to receive an email with a clickable link in it.

### Please describe the actual behaviour
when receiving a password reset email, the text of the email shows as:
"Dear testy,

You have requested your password on Libraries Open Data to be reset.

Please click the following link to confirm this request:

   /user/reset/87a1d440-2a1a-4790-bfc2-e03f8fa800da?key=b8fd884d49

Have a nice day."

pasting the above string to the end of the FQDN works and sends me to the password reset page.

### What steps can be taken to reproduce the issue? 
- register as a new user.
- log out
- click the masthead to log in
- click on the password reset button
- check your email (you may find it's gone to spam, I've not set up SPF as yet).

If I can find the code that constructs the email I might be able to fix this myself but I'm quite new to CKAN so that may take a while.

TkTech:
@MartinHowitt have you set your `site_url` in the .ini? The logic behind that link is very simple and will fail like that if it is not set. You can take a look here, https://github.com/ckan/ckan/blob/d73326ad4beb3743cb4a347b15c04dace8c8654b/ckan/lib/mailer.py#L114

amercader:
We should use `h.url_for` with `qualified=True` here:

https://github.com/ckan/ckan/blob/d73326ad4beb3743cb4a347b15c04dace8c8654b/ckan/lib/mailer.py#L146

MartinHowitt:
@TkTech thanks for getting back to me. ckan.site_url = http://libraries.thedata.place
so that looks ok. One thing I did in deviation from the source script was to update all the python libraries as I was getting an error starting CKAN. I got the idea to that from 
https://github.com/ckan/ckan/issues/3130
although that page was specifically about a Docker install it fixed my problem at the time.

davidread:
From these comments, I believe this is fixed now, so closing.

PR #3486, opened 2017-03-15
Use h.url_for and qualified=True for reset mails

### LINKED ISSUE #3230: Reset link not including site_url

Zharktas:
Fixes #3230 and #3478 

### Proposed fixes:

Suggested fixes in #3478 and #3234.

Replaced urljoin with h.url_for and use qualified=True.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -144,11 +144,11 @@ def get_invite_body(user, group_dict=None, role=None):
 
 
 def get_reset_link(user):
-    return urljoin(config.get('site_url'),
-                   h.url_for(controller='user',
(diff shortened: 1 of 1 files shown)


TkTech:
👍

Issue #3477, opened 2017-03-10
Datastore dump of nested json returns strange characters


arvidT:
### CKAN Version if known (or site URL)
2.5.3

### Please describe the expected behaviour
I have a datastore resource that I'm trying to dump to CSV format using the /datastore/dump/[resource-id] url. The expected behaviour is to get a CSV representation of my resource returned.

### Please describe the actual behaviour
The problem is that any nested json within my resource will be formatted beginning with a "u"-character. See example:
```
_id,measureId,lastUpdated,validFrom,validTo,publiclyAvailableInformation,regionalIndicator,subCategoryDefinition,parentMeasureDefinition,author,authenticator
105,02c0ac9c-b26c-449c-b064-2a2da84fa763,2016-05-26T10:04:23,,,False,{},"{u'parentMeasureId': u'e101817d-2226-4f48-a1d2-fcb61fde050e', u'subCategoryCode': [{u'code': u'545779e637874129a8ef8af5', u'system': u'5457794537874129a8ef8aaf'}]}",,,
```

see `{u'parentMeasureId'` in the second row and in all the other nested json types. 
The "u"-character is not in the original resource and if I access the reource using the usual datastore API it is returned ok:

```json
{
"validFrom": null,
"measureId": "053eb6bf-3da5-4264-a056-87dc4deb3075",
"author": null,
"subCategoryDefinition": {
"parentMeasureId": "a87440aa-30c5-43f0-9e51-d07f5fccc131",
"subCategoryCode": [
{
"code": "fb9376b6-12fa-47b4-8a2a-031657a72a0d",
"system": "19bb7abe-7de4-420f-8259-7aca0a461b6f"
}
]
}
```

### What steps can be taken to reproduce the issue? 
Create a datastore resource using nested json and try to dump it to CSV.

amercader:
Objects are being dumped as their Python representation instead of JSON strings.

This happens both on CKAN <= 2.6 and on master (with @wardi 's new datastore dump writers).

The following adds a test case and solves it for CKAN <= 2.6:

```diff
diff --git a/ckanext/datastore/controller.py b/ckanext/datastore/controller.py
index b872c1d..12f8a10 100644
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -1,5 +1,6 @@
 # encoding: utf-8
 
+import json
 import StringIO
 import unicodecsv as csv
 
[...]

PR #3487, opened 2017-03-16
Nested columns serialized with json.dumps instead of python str representation

smotornyuk:
Pull request for #3477

### DIFF
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -1,9 +1,6 @@
 # encoding: utf-8
 
-import StringIO
-import md5
-
-import pylons
+import json
 
 from ckan.plugins.toolkit import (
     Invalid,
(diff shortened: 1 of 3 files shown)


wardi:
This looks good but can we move it into the writer classes instead? It should be possible to dump these objects in a way that is most appropriate for the format being dumped.

smotornyuk:
I've did it and added base tests for all output formats. I have also changed output of xml format, could you take a look at it(you can see in tests, how it supposed to look like)?  If you don't agree, i can change it back, but it more `xml-way`, i think

wardi:
How about using something like https://pypi.python.org/pypi/xmljson for creating the XML from the arbitrary JSON that might be stored?

smotornyuk:
Initially i thought about it, but this solution will give us error for keys like '1' or 'composite key'. So i used tag 'value' with key attribute in order to preserve hint about original structure and treat all keys exactly the same

Issue #3489, opened 2017-03-17
Ckan release-v2.4.x got no fix for package_search method that exists in versions 2.5 and higher


Engerrs:
### CKAN Version
all ckan 2.4.x versions

### Expected behaviour
We've got user in this way "user = context.get('user')", witch wont lead to KeyError when such key is not there. Such implementation is already on version 2.5 and higher.

### Actual behaviour
We've got user in this way "user = context['user']", witch leads to KeyError when it not exists.

MrkGrgsn:
Was fixed in master for 2.5, the original commit was a523b44ccbef3b18157d2684137fb0aa92295d32

amercader:
Backported to 2.4.7

PR #3490, opened 2017-03-17
[#3489] Adding fix for package_search method that not being added.

Engerrs:
Adding small fix for package_search method that been not added in earlier releases.


amercader:
@smotornyuk backported this to the upcoming 2.4.7 release

PR #3494, opened 2017-03-20
support chained action in plugins

fanjinfei:
Fixes #
Add chained_action support so that user can modify previously defined actions.
### Proposed fixes:
Use chained_action decorator.

### Features:

- [ ] includes API changes


ckan/logic/__init__.py
```diff
@@ -312,6 +313,15 @@ def clear_actions_cache():
     _actions.clear()
 
 
+def chained_action(func):
+    func.__is_chained_action__ = True
```
wardi: `__double_underscore__` should be reserved for attributes that have a special meaning to the python interpreter. You can use `func.chained_action = True` here instead.

ckan/logic/__init__.py
```diff
@@ -394,20 +404,32 @@ def get_action(action):
...
+                resolved_action_plugins[name] = plugin.name
+                # Extensions are exempted from the auth audit for now
+                # This needs to be resolved later
+                auth_function.auth_audit_exempt = True
+                fetched_actions[name] = auth_function
+    for name, func_list in chained_actions.iteritems():
+        if name not in fetched_actions:
+            raise NotFound('The action %r is not found for chained action' % (
+                name))
+        for func in func_list:
```
wardi: I think this should be `for func in reversed(func_list)`

For all other interfaces we have a "first plugin wins" pattern, so the first plugin declaring a chained action in the plugin list should be the first called for that action.

wardi:
This also needs a mention in docs next to where IActions is described

ckan/plugins/interfaces.py
```diff
@@ -892,6 +892,11 @@ def get_actions(self):
         By decorating a function with the `ckan.logic.side_effect_free`
         decorator, the associated action will be made available by a GET
         request (as well as the usual POST request) through the action API.
+
+        By decrorating a function with the 'ckan.plugins.toolkit.chained_action,
+        the action will be chained to another function defined in plugins with a
+        "first plugin wins" pattern, which means the first plugin declaring a
+        chained action should be called first.
```
wardi:
```
Chained actions must be defined as `action_function(original_action, context, data_dict)` where the first parameter will be set to the action function in the next plugin or in core ckan. The chained action may call the `original_action` function, optionally passing different values, handling exceptions, returning different values and/or raising different exceptions to the caller.
```

ckanext/datastore/tests/test_chained_action.py
```diff
@@ -0,0 +1,81 @@
...
+
+@p.toolkit.chained_action
+def datastore_delete(up_func, context, data_dict):
+    global deleted_count
+    res = helpers.call_action(u"datastore_search",
+                              resource_id=data_dict[u'resource_id'],
+                              filters=data_dict[u'filters'],
+                              limit=10,)
+    result = up_func(context, data_dict)
+    deleted_count = res.get(u'total', 0)
```
wardi: instead of modifying a global why not return this as part of the response to demonstrate that this is possible?

ckanext/datastore/tests/test_chained_action.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return result
+
+
+class ChainedDataStorePlugin(p.SingletonPlugin):
```
wardi: How about `ExampleDatastoreDeleteWithCountPlugin`

`ChainedDatastorePlugin` sounds like something a user might want to use..

ckan/plugins/interfaces.py
wardi: ^ I meant to say I would add something like this paragraph to the one you've written.

wardi:
@fanjinfei please update the name to fix the test and I'll merge

fanjinfei:
@wardi test case fixed

wardi:
@fanjinfei great work!

PR #3495, opened 2017-03-21
resource_search API optionally returns private resources

grahame:
### Proposed fixes:

Add a new `include_private` flag to the `resource_search` endpoint

If set, return resources attached to private packages to which the
authenticated user has access.

CKAN 2.6 added the `include_private` flag to `package_search`. Adding equivalent functionality to the `resource_search` API is very useful to CKAN deployments which make use of private datasets.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2064,12 +2064,17 @@ def resource_search(context, data_dict):
     :type offset: int
     :param limit: Apply a limit to the query.
     :type limit: int
+    :param include_private: if ``True``, resources from private datasets will
+        be included in the results. Only resources from private datasets in
+        the user's organizations will be returned and sysadmins will be
+        returned all private resources. Optional, the default is ``False``.
 
     :returns:  A dictionary with a ``count`` field, and a ``results`` field.
     :rtype: dict
 
     '''
     model = context['model']
+    user = context.get('user')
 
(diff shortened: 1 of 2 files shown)


TkTech:
@wardi I imagine this should be done using the permission labels recently introduced instead, yeah? Otherwise the conditions for package visibility will be wrong.

wardi:
Yes, and those are only stored in solr at the moment.

wardi:
Instead we need to store the dataset permission labels in postgres as an indexed `_text` column on the `Package` model, then we can filter on it for `resource_search`.

I can't think of a way to implement `resource_search` on the existing solr index. Offset pagination would be very difficult to implement since we need to filter the results in two passes. We could duplicate resources as their own entity in solr, but I don't think that might complicate the pluggable search backends work.

grahame:
Ah, I see where I've gone wrong - this was a patch against CKAN 2.6, and I didn't spot the changes in `package_search` when I cherry-picked this onto master.

One advantage of putting the resource data into solr -  we could use solr for the resource queries, and drop all the special code that peeks into the extra fields column in the resources table. We could probably also lose the `ckan.extra_resource_fields` config option.

TkTech:
I'm more than happy to see resources (and users, groups, orgs) added to the search index, the indexer rewrite will follow along. The interface will just end up with new `index_*` variants.

kmbn:

PR #3514, opened 2017-03-28
[3513] minimum setuptools version

### LINKED ISSUE #3513: Minimum setuptools version
CKAN requires a recent version of setuptools ever since some of CKAN's dependencies were upgraded. However there's nothing in the install instructions to sort this out. The default setuptools for our recommended distro, Ubuntu Trusty 14.04, is 2.2, which is too old.

Specifically:
* repoze.who needs setuptools 11.3 or later https://github.com/repoze/repoze.who/issues/26
* html5lib needs setuptools 18.5 or later https://github.com/ckan/ckan/issues/3232

davidread:
Fixes #3513 

It's a problem to pin setuptools in requirements.txt, because it is generated from requirements.in by pip-compile, and that specifically excludes setuptools:
```
# The following packages are commented out because they are
# considered to be unsafe in a requirements file:
# setuptools
```
I can't see a way to hack pip-compile, so we'd have add it back to requirements.txt manually every time we run pip-compile, which is not ideal.

I tried adding it to setup.py:
```
    install_requires=[
        'setuptools>=18.5',  # required for html5lib
        # ALL other dependencies should be put in requirements.in/.txt
    ],
```
But I get this error when I do `python setup.py develop`:
```
  File "/home/vagrant/ckan/local/lib/python2.7/site-packages/packaging/requirements.py", line 59, in <module>
    MARKER_EXPR = originalTextFor(MARKER_EXPR())("marker")
TypeError: __call__() takes exactly 2 arguments (1 given)
```
Clearly it is difficult for setuptools to upgrade itself while running.

Final try was to add the requirement as code in setup.py, so that it checks the version but won't try and upgrade itself - it tells the user how to do this. See PR.

## Testing

```
(ckan)vagrant@precise64:/vagrant/src/ckan$ pip install setuptools==2.2
...

(ckan)vagrant@precise64:/vagrant/src/ckan$ python setup.py develop
  File "setup.py", line 31, in <module>
    suggested=SUGGESTED_SETUPTOOLS_VERSION))
AssertionError: setuptools version error
You need a newer version of setuptools.
You have 2.2, you need at least 18.5 and the suggested version is 18.5.
Do this:
    pip install setuptools==18.5
and then try again to install ckan into your python environment.

(ckan)vagrant@precise64:/vagrant/src/ckan$ pip install setuptools==18.5
...
  Found existing installation: setuptools 2.2
    Uninstalling setuptools-2.2:
      Successfully uninstalled setuptools-2.2
Successfully installed setuptools-18.5

(ckan)vagrant@precise64:/vagrant/src/ckan$ python setup.py develop
...
Installed /vagrant/src/ckan
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,16 +6,23 @@ if os.environ.get('USER', '') == 'vagrant':
     del os.link
 
 try:
-    from setuptools import setup, find_packages, __version__
+    from setuptools import (setup, find_packages,
+                            __version__ as setuptools_version)
 except ImportError:
     from ez_setup import use_setuptools
     use_setuptools()
-    from setuptools import setup, find_packages, __version__
+    from setuptools import (setup, find_packages,
+                            __version__ as setuptools_version)
 
-assert __version__ >= '18.5', 'setuptools version error'\
-    '\nYou need a newer version of setuptools. Do this:\n' \
-    '    pip install -U setuptools\n' \
-    'before installing ckan into your python environment again.'
+MIN_SETUPTOOLS_VERSION = 18.5
+assert setuptools_version >= str(MIN_SETUPTOOLS_VERSION) and \
+    int(setuptools_version.split('.')[0]) >= int(MIN_SETUPTOOLS_VERSION),\
+    ('setuptools version error'
+     '\nYou need a newer version of setuptools.\n'
+     'You have %s and you need at least %s.\nDo this:\n'
+     '    pip install -U setuptools\n' \
+     'and then try again to install ckan into your python environment.' %
+     (setuptools_version, MIN_SETUPTOOLS_VERSION))
     
 from ckan import (__version__, __description__, __long_description__,
                   __license__)
--- a/setup.py
+++ b/setup.py
@@ -17,18 +17,18 @@ except ImportError:
 from ckan import (__version__, __description__, __long_description__,
                   __license__)
 
-MIN_SETUPTOOLS_VERSION = 18.5
-SUGGESTED_SETUPTOOLS_VERSION = 18.5
+MIN_SETUPTOOLS_VERSION = 20.4
 assert setuptools_version >= str(MIN_SETUPTOOLS_VERSION) and \
     int(setuptools_version.split('.')[0]) >= int(MIN_SETUPTOOLS_VERSION),\
     ('setuptools version error'
      '\nYou need a newer version of setuptools.\n'
-     'You have {current}, you need at least {minimum} and the suggested '
-     'version is {suggested}.\nDo this:\n'
-     '    pip install setuptools=={suggested}\n'
+     'You have {current}, you need at least {minimum}'
+     '\nInstall the recommended version:\n'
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -6,12 +6,17 @@
     del os.link
 
 try:
-    from setuptools import setup, find_packages
+    from setuptools import setup, find_packages, __version__
 except ImportError:
     from ez_setup import use_setuptools
     use_setuptools()
-    from setuptools import setup, find_packages
+    from setuptools import setup, find_packages, __version__
```
amercader: Maybe import `as setuptools_versions` to avoid confusion / clashes

davidread: Ta, that's changed now

davidread:
The 'danger' of installing setuptools is mentioned here, to do with old versions of pip:
https://github.com/pypa/setuptools/issues/937#issuecomment-274629937
However I tried using pip all the way back to 1.0.1, so without wheel support, and I don't have any issue installing various setuptools versions with it, or running `python setup.py develop`.

davidread:
It now suggests using setuptools==1.8 rather than the latest, as suggested here: https://github.com/ckan/ckan/issues/3513#issuecomment-290022839

davidread:
Now recommending 20.4, as per comments on the issue. Using a requirements file too, so the recommended version is not duplicated in the docs.

davidread:
@thenets if you could test this, that'd be great

thenets:
@davidread I don't know how to use it. Can you teach me how to use on IRC? Or give me a tutorial?

davidread:
@thenets Try this code with:
```
git fetch origin
git checkout -b require-setuptools-version origin/require-setuptools-version
git merge master
```
And then this new setuptools check simply runs when you type `python setup.py develop` when in the ckan source directory. (See the example in this PR's description section called "Testing".)

Also of interest is whether the recommended version of setuptools 'works'. You said you had some problems with it. For me the problem occurred when CKAN starts up using setuptools < 18.5 - it caused an exception, but maybe you see it at another point - it would be good to know the details.

davidread:
Ah I just saw your response on the issue. Ah well, at least we can as a group go for setuptools 20.4 and any problems will surface quicker. That at least is better than everyone using different versions.

thenets:
@davidread I've tried configuring CKAN 2.7 but my Docker container does not work for the new schema. I'll try again later. I am creating a new installation script to automate the entire installation and testing process.
I'll let you know when I'm done!

davidread:
Thanks!

davidread:
A person tried needed this and benefited from the fix over on #3544, which I've now marked as depending on this PR. Some weird bug about 16.04 benefits from reinstalling setuptools, which this PR adds to the install instructions.

davidread:
@amercader any thoughts on this?

amercader:
I'm not too keen on adding extra install steps but this seems like can help difficult to track install problems. Thanks @davidread

Issue #3457, opened 2017-02-23
Changing resource format for an existing resource, won't change/create new view for new format.


Engerrs:
### CKAN Version
All

### Expected behaviour
When we created new resource for dataset, with/without format, after updating the resource for new format it creates/changes view for it.

### Actual behaviour
When we created new resource for dataset, with/without format, after updating the resource for new format it doesn't create/change the view for resource.

### What steps can be taken to reproduce the issue? 
1.Create Dataset
2. Create new resource for it with format CSV
3. Enter the resource and update it's format to PNG
4. We can see that the view for the resource hasn't changed.

Engerrs:
I can take a look at this one if need.

wardi:
@Engerrs please do

Engerrs:
Hi @wardi , created a fix for that one.

gleb-rudenko:
@wardi, what is the progress of this one?

wardi:
@Engerrs please create a PR so we can review it

Engerrs:
Sorry @wardi , thought that  I've created one. Add it.

wardi:
fixed by #3515

PR #3515, opened 2017-03-29
[#3457]Create new resource view if resource format changed

Engerrs:
Creating new resource view if resource format is changed.


ckan/logic/action/update.py
```diff
@@ -62,6 +62,8 @@ def resource_update(context, data_dict):
     model = context['model']
     user = context['user']
     id = _get_or_bust(data_dict, "id")
+    old_resource = _get_action('resource_show')(context, {'id': id})
+
```
smotornyuk: I think that you should allow CKAN to perform all its default checks first and obtain old information no sooner than required

smotornyuk: Or you can even use resource's model...

ckan/logic/action/update.py
```diff
@@ -120,6 +122,12 @@ def resource_update(context, data_dict):
 
     resource = _get_action('resource_show')(context, {'id': id})
 
+    if old_resource['format'] != resource['format']:
+        _get_action('package_create_default_resource_views')(
+            {'model': context['model'], 'user': context['user'],
+             'ignore_auth': True},
+            {'package': updated_pkg_dict})
```
smotornyuk: Wouldn't `resource_create_default_resource_view` fit here better?

smotornyuk:
Now it looks ok :+1: 
@amercader , what do you think? It replicates logic from `resource_create` if format changed and tests are provided, so i have no objection against this PR

amercader:
Looks great, thanks @Engerrs.

Feel free to merge @smotornyuk. Also marked for backporting.

Issue #3508, opened 2017-03-27
Document undocumented interface methods


torfsen:
There are several methods in the interfaces of [`ckan.plugins.interfaces`](https://github.com/ckan/ckan/blob/af2f061fdd08219d5070415be118d6063d38341f/ckan/plugins/interfaces.py) which are undocumented and therefore do not show up in the [interface reference](http://docs.ckan.org/en/latest/extensions/plugin-interfaces.html). This is problematic, since implementing an interface partially will raise an exception sooner or later (unless you inherit the default implementations by explicitly passing `inherit=True` to `plugins.implements`).

Steps that should be taken:

1. Document all interface methods
2. Tell Sphinx not to hide undocumented methods (will also help if an undocumented method is added in the future)

smotornyuk:
Resolved in #3519

PR #3519, opened 2017-03-30
Updated documentation of interfaces

smotornyuk:
#3508
* Removed authz_add/remove_role from IPackageController, IOrganizationController and IGroupController
* Added docstrings to undocumented methods
* Now sphinx shows all, even undocumented classes and interfaces from `ckan.plugins.interfaces`


amercader:
Thanks @smotornyuk, this is super useful

PR #3526, opened 2017-04-04
Avoid duplicate Datapusher jobs

### LINKED ISSUE #2334: DatapusherPlugin implementation of notify() can call 'datapusher_submit' multiple times
Hi,

Currently `DatapusherPlugin` implements `notify()` from `IDomainObjectModification` & `IResourceUrlChange`.

On our CKAN 2.3 portal this this means it's called twice; the first time, because operation == model.domain_object.DomainObjectOperation.new (IDomainObjectModification) and the second time because operation is not passed (IResourceUrlChange).

It's been like that for a [long time](https://github.com/ckan/ckan/blame/8f298cd9e658d788b031a4059271210aff6a3d94/ckanext/datapusher/plugin.py) and I wonder it this has not been picked up before because it has not been tested with extensions that implement other interfaces leading to notify being called multiple times ? 

In my case implementing IResourceUrlChange sounds like a quick fix, but I'm wondering if this doesn't highlight that the logic in DatapusherPlugin should be re-considered?

What do you think?

### LINKED ISSUE #2721: DataPusher does not lock table while uploading data
Create a dataset with a CSV file with 3 rows.
Go to the resource page > Manage > DataStore, Click on "Upload to DataStore" many successively

Result:

pgb4oo5

davidread:
Fixes #2721
Fixes #2334

Thanks @clementmouchet for the initial code for this.

screen shot 2017-04-04 at 11 43 00

Test this PR manually by double-clicking the "Upload to Datastore" button and seeing that the second one of the tasks is skipped, seeing this in the logs:
```
2017-04-04 10:32:17,633 INFO  [ckanext.datapusher.logic.action] A pending task was found 3fcabb2a-a965-4216-877e-0202c00ee8e7 for this resource, so skipping this duplicate task
```

Ideally in this code some signals would be added in, to do the check and update atomically, but as it stands, it avoids the problem for 99.9% of cases.

### DIFF
--- a/ckanext/datapusher/logic/action.py
+++ b/ckanext/datapusher/logic/action.py
@@ -89,12 +89,13 @@ def datapusher_submit(context, data_dict):
             'task_type': 'datapusher',
             'key': 'datapusher'
         })
-
-        if existing_task and existing_task.get('state') == 'pending':
+        assume_task_stale_after = datetime.timedelta(seconds=int(
+            config.get('ckan.datapusher.assume_task_stale_after', 3600)))
+        if existing_task.get('state') == 'pending':
             updated = datetime.datetime.strptime(
                 existing_task['last_updated'], '%Y-%m-%dT%H:%M:%S.%f')
             time_since_last_updated = datetime.datetime.now() - updated
-            if time_since_last_updated > datetime.timedelta(days=1):
+            if time_since_last_updated > assume_task_stale_after:
                 # it's been a while since the job was last updated - it's more
                 # likely something went wrong with it and the state wasn't
                 # updated than its still in progress. Let it be restarted.
--- a/ckanext/datapusher/logic/action.py
+++ b/ckanext/datapusher/logic/action.py
@@ -84,18 +84,33 @@ def datapusher_submit(context, data_dict):
         'error': '{}',
     }
     try:
-        task_id = p.toolkit.get_action('task_status_show')(context, {
+        existing_task = p.toolkit.get_action('task_status_show')(context, {
             'entity_id': res_id,
             'task_type': 'datapusher',
             'key': 'datapusher'
-        })['id']
-        task['id'] = task_id
+        })
+
+        if existing_task and existing_task.get('state') == 'pending':
+            updated = datetime.datetime.strptime(
+                existing_task['last_updated'], '%Y-%m-%dT%H:%M:%S.%f')
+            time_since_last_updated = datetime.datetime.now() - updated
+            if time_since_last_updated > datetime.timedelta(days=1):
+                # it's been a while since the job was last updated - it's more
+                # likely something went wrong with it and the state wasn't
+                # updated than its still in progress. Let it be restarted.
+                log.info('A pending task was found %r, but it is only %s hours'
+                         'old', existing_task['id'], time_since_last_updated)
+            else:
+                log.info('A pending task was found %s for this resource, so '
+                         'skipping this duplicate task', existing_task['id'])
+                return False
+
+        task['id'] = existing_task['id']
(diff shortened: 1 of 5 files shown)


ckanext/datapusher/logic/action.py
```diff
@@ -84,18 +84,33 @@ def datapusher_submit(context, data_dict):
...
+        })
+
+        if existing_task and existing_task.get('state') == 'pending':
+            updated = datetime.datetime.strptime(
+                existing_task['last_updated'], '%Y-%m-%dT%H:%M:%S.%f')
+            time_since_last_updated = datetime.datetime.now() - updated
+            if time_since_last_updated > datetime.timedelta(days=1):
+                # it's been a while since the job was last updated - it's more
+                # likely something went wrong with it and the state wasn't
+                # updated than its still in progress. Let it be restarted.
```
wardi: A day seems like a very long time. Maybe make this shorter and use a nice descriptive constant name like `ASSUME_TASK_STALE_AFTER = timedelta(..)`

ckanext/datapusher/logic/action.py
```diff
@@ -84,18 +84,33 @@ def datapusher_submit(context, data_dict):
...
-        task_id = p.toolkit.get_action('task_status_show')(context, {
+        existing_task = p.toolkit.get_action('task_status_show')(context, {
             'entity_id': res_id,
             'task_type': 'datapusher',
             'key': 'datapusher'
-        })['id']
-        task['id'] = task_id
+        })
+
+        if existing_task and existing_task.get('state') == 'pending':
```
smotornyuk: if task does not exist, [error is raised](https://github.com/ckan/ckan/blob/42406b15de78bff6e486a0e263f98114c644fffc/ckan/logic/action/get.py#L2348). In that case, could first check be omitted in order to avoid confusing expectations in future?

ckanext/datapusher/logic/action.py
davidread: Sounds good, I'll do an hour

smotornyuk:
@davidread i wrote simplest test for this fix, could you add it if you find it suitable
```
diff --git a/ckanext/datapusher/tests/test_action.py b/ckanext/datapusher/tests/test_action.py
index 5c74c98..52f1e75 100644
--- a/ckanext/datapusher/tests/test_action.py
+++ b/ckanext/datapusher/tests/test_action.py
@@ -3,6 +3,7 @@
 import datetime
 
 from nose.tools import eq_
+import mock
 
 import ckan.plugins as p
 from ckan.tests import helpers, factories
@@ -278,3 +279,21 @@ class TestDataPusherAction(object):
 
         # Not called
         eq_(len(mock_datapusher_submit.mock_calls), 1)
+
+    def test_duplicated_tasks(self):
+        def submit(res, user):
+            return helpers.call_action(
+                'datapusher_submit', context=dict(user=user['name']),
+                resource_id=res['id'])
+
+        res = factories.Resource(user=user)
+        with mock.patch('requests.post') as r_mock:
+            r_mock().json = mock.Mock(
+                side_effect=lambda: dict.fromkeys(
+                    ['job_id', 'job_key']))
+            r_mock.reset_mock()
+            submit(res, user)
+            submit(res, user)
+
+            eq_(1, r_mock.call_count)

```

ckanext/datapusher/logic/action.py
davidread: good spot! thanks

davidread:
@wardi That 'task stale' option has been added now.

@smotornyuk That test has been added now.

amercader:
Thanks @davidread (and @clementmouchet)

PR #3529, opened 2017-04-04
Add notes on running tests with Solr multicore

davidread:
I'm aware there is a desire to change the install instructions to use Solr multicore, a more recent version of Solr and switch to Tomcat. However, right now the tests fail because the test-core.ini specifies a solr_url that assumes you're multi-core. So this is just a quick fix for that really. But when the wider solr changes do occur, there will still be plenty of people still running the existing single core set-up and they may appreciate these hints to get the tests going too.

I tested this on Trusty Solr 3.6 on jetty and have reasonable confidence it will work on other set-ups. But I'd welcome any advice on any incompatibilities with these instructions.


amercader:
Cheers, we should revisit these instructions once we officially target a modern version of Solr.

PR #3531, opened 2017-04-05
do not allow to modify exsiting username

fanjinfei:
Fixes #

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -349,14 +349,6 @@ class UserController(base.BaseController):
 
             email_changed = data_dict['email'] != c.userobj.email
 
-            if id != data_dict['name']:
-                old_user = model.User.get(id)
-                if old_user.name != data_dict['name']:
-                    errors = {'username': [_('Can not modify username')]}
-                    error_summary = {_('Username'):
-                                     _('Can not modify username')}
-                    return self.edit(id, data_dict, errors, error_summary)
-
             if (data_dict['password1'] and data_dict['password2']) \
                     or email_changed:
                 identity = {'login': c.user,
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -530,8 +530,11 @@ class UserController(base.BaseController):
         if request.method == 'POST':
             try:
                 context['reset_password'] = True
+                user_state = user_dict['state']
                 new_password = self._get_form_password()
-                user_dict['password'] = new_password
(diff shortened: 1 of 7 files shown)


fanjinfei:
WIP

fanjinfei:
ready for review

ckan/controllers/user.py
```diff
@@ -349,6 +349,14 @@ def _save_edit(self, id, context):
 
             email_changed = data_dict['email'] != c.userobj.email
 
+            if id != data_dict['name']:
```
amercader: Shouldn't this check be done at the validation level? Otherwise the name will be able to be changed via the API

amercader:
On second thoughts I'm not too sure about this. For instance when you get an invite to an organization (via email) you get a user name assigned based on your email plus some random characters, eg `adria-mercader-1559`. Even the invite email says "You can change it later."

Another use case is if someone leaves the organization but you still want to keep using the CKAN user changing its name.

I think it's a useful feature and if there are auth implications somewhere related to changing the name we should fix those.

wardi:
invites shouldn't be creating users, they should just send a URL that can be used to create a user already added to an org

wardi:
If changing usernames is genuinely useful we could live with just being able to override the schema so we can disable it on our site cleanly. That's not something we can do right now.

camfindlay:
Happy to provide the related reasoning behind this security enhancement - is there a CKAN core team email I can interact with?

raylatchmanan-qol:
Hello Adrià and crew,

Could you please advise the security reason why the username editing is now been disabled? 

In our organisation (gov), we use a distributed publishing model where each dept (~40) manages its own publishers (~400). To keep the publisher names anonymous from the public, we use "<agency-acronym>_publisher_123" username pattern (e.g. transport_publisher_123). We do maintain an internal registry with the publisher's details etc. However, like every other gov, we do have dept name changes after elections/new gov, where we need to update the usernames to reflect the name changes.

With this feature now been disabled, it is a massive headache for us as we have no other option than creating new user account and delete old ones. This is very time consuming and also publishers will lose history, follows etc.

I do appreciate you may have valid reasons behind this change, however, is it possible for the sysadmin or on the production.ini to set the flag if the username is editable or not instead of bake it to the code? Happy to hear your suggestions too. 

Thank you and appreciate the team's contribution to CKAN.

Kind regards,
Ray

PR #3535, opened 2017-04-06
Dataset modified timestamp gets updated even if nothing was changed in dataset

gleb-rudenko:
Fixes #
#2373
### Proposed fixes:
metadata_modified only changes when the package is modified

### Features:
- [X] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -294,12 +294,11 @@ def package_update(context, data_dict):
     else:
         rev.message = _(u'REST API: Update object %s') % data.get("name")
 
-    #avoid revisioning by updating directly
-    model.Session.query(model.Package).filter_by(id=pkg.id).update(
-        {"metadata_modified": datetime.datetime.utcnow()})
-    model.Session.refresh(pkg)
-
     pkg = model_save.package_dict_save(data, context)
+
+    # Update metadata modified ONLY when pkg metadata was changed.
+    if model.Session.is_modified(pkg):
+        pkg.metadata_modified = datetime.datetime.utcnow()
 
     context_org_update = context.copy()
     context_org_update['ignore_auth'] = True
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -297,9 +297,6 @@ def package_update(context, data_dict):
     pkg = model_save.package_dict_save(data, context)
 
     # Update metadata modified ONLY when pkg metadata was changed.
-    if model.Session.is_modified(pkg):
-        pkg.metadata_modified = datetime.datetime.utcnow()
-
     context_org_update = context.copy()
     context_org_update['ignore_auth'] = True
     context_org_update['defer_commit'] = True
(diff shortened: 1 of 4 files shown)


ckan/logic/action/update.py
```diff
@@ -294,13 +294,12 @@ def package_update(context, data_dict):
...
 
-    #avoid revisioning by updating directly
-    model.Session.query(model.Package).filter_by(id=pkg.id).update(
-        {"metadata_modified": datetime.datetime.utcnow()})
-    model.Session.refresh(pkg)
-
     pkg = model_save.package_dict_save(data, context)
 
+    # Update metadata modified ONLY when pkg metadata was changed.
+    if model.Session.is_modified(pkg):
```
wardi: we need to update metadata_modified when any of the extras or resources in this package have been modified as well

gleb-rudenko: @wardi, ye, you are right, digging into it.

ckan/tests/legacy/functional/api/test_activity.py
```diff
@@ -361,12 +361,12 @@ def _create_package(self, user, name=None):
         details = self.activity_details(activity)
         # There should be five activity details: one for the package itself,
         # one for each of its two resources, and one for each of its two tags.
-
-        assert len(details) == 5, "There should be five activity details."
+        # import pdb; pdb.set_trace();
```
torfsen: I see I'm not the only one who mistakenly commits debugging statements :wink:

TkTech: 😆

gleb-rudenko: @TkTech oops...

ckan/tests/legacy/functional/api/test_activity.py
```diff
@@ -466,6 +465,10 @@ def _add_resource(self, package, user):
 
         # Test for the presence of a correct activity detail item.
         details = self.activity_details(activity)
+        details = list(filter(
+            lambda x: x['object_type'] == "Resource",
+            details
+        ))
```
TkTech: Filter is kinda deprecated, use list comprehensions. Tends to perform better as since you're removing multiple function calls.

`details = [x for x in details if x['object_type'] == 'Resources']`

gleb-rudenko:
finally, all tests are passed, can someone look at the code and the changes I've made?

gleb-rudenko:
during my testing, found another issue:
when we add tags to the dataset, modified date is updated as well (both postgresql and solr) but when we remove tags, it doesn't update the modified date. Will need to take a look at this one.

amercader:
I stumbled into this old PR and I really like that we are not blindly updating metadata_modified each time package_update is called but when there's an actual update. It looks like is no longer WIP so I removed the label. @wardi let me know if you don't have time to look at this and I'll try to find some.

wardi:
@amercader yes please take a look. I agree not changing the date when nothing changed is better and we can short circuit things like the solr index being updated at the same time.

wardi:
@amercader you offered to take a look at this one and I forgot to assign you

smotornyuk:
This PR is not going to be updated as author no longer works with CKAN

PR #3537, opened 2017-04-07
fix organization members translation

fanjinfei:
Fixes #
Do not put argument inside translation
### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/organization/members.html
+++ b/ckan/templates/organization/members.html
@@ -10,7 +10,9 @@
 {% endblock %}
 
 {% block primary_content_inner %}
-  <h3 class="page-heading">{{ _('{0} members'.format(c.members|length)) }}</h3>
+  {% set count = c.members|length %}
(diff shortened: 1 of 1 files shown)


ckan/templates/organization/members.html
```diff
@@ -10,7 +10,9 @@
 {% endblock %}
 
 {% block primary_content_inner %}
-  <h3 class="page-heading">{{ _('{0} members'.format(c.members|length)) }}</h3>
+  {% set count = c.members|length %}
+  {% set members_count = _(ungettext('{0} member', '{0} members', count)).format(count) %}
```
torfsen: You shouldn't need to pass the result of `ungettext` to `_`, that would translate it twice. In addition, please use named placeholders in your strings:

    set members_count = ungettext('{count} member', '{count} members', count).format(count=count)

fanjinfei: thanks

PR #3523, opened 2017-04-07
datastore dump + search performance

wardi:
Profiling dumping datastore resources and the `datastore_search` API it uses to identify the easiest improvements we can make.

Timings for `python -m cProfile paster datastore dump` on a datastore table with ~920k rows
- commit c1132574ef
- total time: 218.7 s - all percentages below relate to this
- 68.8 % within datastore_search (datastore/logic/action)
   - 41.1 % within format_results (datastore/db)
      - 21.8 % convert (datastore/db)
      - 13.5 % fetchone (sqlalchemy)
   - 27.4 % within execute (sqlalchemy)
- 27.7 % within writerow (datastore/writer)
   - 18.6 % writerow (unicodecsv)
   - 6.5 % _json_dump_nested (datastore/writer)

More than 40% of our time is in sqlalchemy fetchone + execute. Switching to the postgres `COPY` function instead of iterating over select results is promising. We can also skip intermediate total row calculations and see if index-based offsets will perform better for large tables.

Almost 30% of our time is spent formatting the results as CSV. We can lean on postgres to format the CSV for us we could skip almost all of this work.

More than 20% of our time is spent converting results returned from postgres with column type and python `isinstance` checks. We can investigate optimizations in the python code and use more efficient data libraries and conversions.

### DIFF
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -28,7 +28,7 @@ int_validator = get_validator('int_validator')
 boolean_validator = get_validator('boolean_validator')
 
 DUMP_FORMATS = 'csv', 'tsv', 'json', 'xml'
-PAGINATE_BY = 10000
+PAGINATE_BY = 32000
 
 
 class DatastoreController(BaseController):
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -100,16 +100,22 @@ class DatastoreController(BaseController):
 
 
 def dump_to(resource_id, output, fmt, offset, limit, options):
+    if fmt == 'csv':
+        writer_factory = csv_writer
+        records_format = 'csv'
+    elif fmt == 'tsv':
+        writer_factory = tsv_writer
+        records_format = 'tsv'
+    elif fmt == 'json':
+        writer_factory = json_writer
+        records_format = 'lists'
+    elif fmt == 'xml':
+        writer_factory = xml_writer
+        records_format = 'objects'
+
     def start_writer(fields):
         bom = options.get(u'bom', False)
-        if fmt == 'csv':
-            return csv_writer(output, fields, resource_id, bom)
-        if fmt == 'tsv':
-            return tsv_writer(output, fields, resource_id, bom)
-        if fmt == 'json':
-            return json_writer(output, fields, resource_id, bom)
-        if fmt == 'xml':
-            return xml_writer(output, fields, resource_id, bom)
+        return writer_factory(output, fields, resource_id, bom)
 
     def result_page(offs, lim):
         return get_action('datastore_search')(None, {
@@ -118,6 +124,7 @@ def dump_to(resource_id, output, fmt, offset, limit, options):
                 PAGINATE_BY if limit is None
                 else min(PAGINATE_BY, lim),
             'offset': offs,
+            'records_format': records_format,
             })
 
     result = result_page(offset, limit)
@@ -128,13 +135,20 @@ def dump_to(resource_id, output, fmt, offset, limit, options):
             if limit is not None and limit <= 0:
                 break
 
-            for record in result['records']:
-                wr.writerow([record[column] for column in columns])
+            records = result['records']
 
-            if len(result['records']) < PAGINATE_BY:
+            wr.write_records(records)
+
+            if records_format == 'objects' or records_format == 'lists':
+                if len(records) < PAGINATE_BY:
+                    break
+            elif not records:
                 break
+
             offset += PAGINATE_BY
             if limit is not None:
                 limit -= PAGINATE_BY
+                if limit <= 0:
+                    break
 
             result = result_page(offset, limit)
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -125,6 +125,7 @@ def dump_to(resource_id, output, fmt, offset, limit, options):
                 else min(PAGINATE_BY, lim),
             'offset': offs,
             'records_format': records_format,
+            'include_total': 'false',  # XXX: default() is broken
             })
 
     result = result_page(offset, limit)
--- a/ckanext/datastore/logic/action.py
+++ b/ckanext/datastore/logic/action.py
@@ -122,8 +122,6 @@ def datastore_create(context, data_dict):
             resource_id = data_dict['resource_id']
             _check_read_only(context, resource_id)
 
-    data_dict['connection_url'] = config['ckan.datastore.write_url']
-
     # validate aliases
     aliases = datastore_helpers.get_list(data_dict.get('aliases', []))
     for alias in aliases:
@@ -153,7 +151,6 @@ def datastore_create(context, data_dict):
 
     result.pop('id', None)
     result.pop('private', None)
-    result.pop('connection_url')
     return result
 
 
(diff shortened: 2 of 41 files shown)


wardi:
With 856ae96 the time spent in `datastore_search` was significantly reduced:

| datastore_search | seconds for 920k records |
| --- | --- |
| before change | 150.4 |
| after change | 84.1 |

Action API users will see further benefit from JSON not needing to be re-encoded when sent to the client.

We're taking advantage of the new `RawJSON` support in simplejson 3.10 and the postgres `row_to_json()` function to render row JSON.

Overall time for dumping to CSV was only slightly improved because we're still converting from JSON -> python dictionaries -> CSV. We're also still looping over every row in python instead of processing a whole batch of rows at a time. Fixing those issues is next.

wardi:
a553c1d hands off CSV, TSV and our JSON-lists-style dumps to postgres with a new `records_format` parameter. It also starts using COPY instead of looping over results. This code is running ~ 3.5x the speed of what we have now.

| datastore dump csv | seconds for 920k records |
| --- | --- |
| before this branch | 218.7 |
| after a553c1d | 62.8 |

This change is a little rough and I'd like to run separate profiles on each of the export formats because their code paths are quite different now. I also need to implement select filtering for the CSV and TSV output (not required for the dump command but should be supported by the API)

The new profile shows almost half our time being spent in queries not related to dumping the actual data, so there should be another opportunity for optimization there that will help all `datastore_search` calls.

wardi:
Total calculation is unnecessary for dumping all rows, so d7ab29e disables that. This branch is now running at more than 6x the speed of master for dumping CSV files.

| datastore dump csv | seconds for 920k records |
| --- | --- |
| before this branch | 218.7 |
| after d7ab29e | 34.4 |

This time is dominated (>90%) by the postgres copy operation. We can `explain` and start optimizing the postgres query to go further, but after a 6x improvement I'm tempted to circle back around and collect some of the other performance numbers and make sure everything is still working properly.

amercader:
@benscott @aliceh75 This has the potential to affect `IDataStore` plugins (apart from providing significant performance improvements). As you were using it (IIRC) perhaps you want to give it a try with your instance to flag any potential issues?

wardi:
More numbers, this time covering calls to `datastore_search` API with curl in a few different ways. I'm comparing efcfe97 (before include_total), bd361dc (before this PR) and d7ab29e. All numbers are the smallest "real" seconds from three runs of `time curl ...`. Blank cells represent unsupported calls.

| parameters | A: efcfe97 | B: bd361dc | C: d7ab29e | A/B | A/C |
| --- | --- | --- | --- | --- | --- |
| distinct, 1 field | 1.002 s | 1.028 s | 1.015 s | 0.97 | 0.99 |
| distinct, 1 field, no total | | 0.650 s | 0.646 s | 1.54 | 1.55 |
| distinct, 10k | 2.746 s | 1.288 s | 0.761 s | 2.13 | 3.61 |
| distinct, 10k, no total | | 1.000 s | 0.418 s | 2.75 | **6.57** |
| 10k | 2.763 s | 1.259 s | 0.751 s | 2.19 | 3.68 |
| 10k, no total | | 0.969 s | 0.417 s | 2.85 | **6.63** |
| 10k, lists, no total | | | 0.441 s | | **6.27** |
| 10k, csv, no total | | | 0.157 s | | **17.77** |
| 100k | 11.883 | 10.485 s | 4.061 s | 1.13 | 2.93 |
| 100k, no total | | 10.211 s | 3.705 s | 1.16 | 3.21 |
| 100k, lists, no total | | | 3.680 s | | 3.68 |
| 100k, csv, no total | | | 0.863 s | | **13.77** |

After this change for 10k+ row queries of datastore tables with ~ 1M rows expect `datastore_search` to be ~ 3x as fast as CKAN 2.6 with no change to your client code, ~ 6x as fast when using `include_total=false`, and more than 12x as fast when using `include_total=false` and `records_format=csv` or `tsv`.

the `records_format=csv` and `tsv` options are the only ones using `COPY` in d7ab29e. Based on this converting the other formats to use `COPY` internally is a small change that looks worth doing.

wardi:
Updated https://github.com/ckan/ckan/pull/3523#issuecomment-294190135 above with numbers for f51610a

wardi:
My last change bumps up the page size for the dump controller from 10k to 32k rows based on some timings that show it shaves 10-20s off our 920k row export: image

Note that xml is the last transformation being done in python (and cElementTree). If we generate our xml output in postgres it will have overall performance similar to the json format.

This work is complete. Here's a comparison of the before+after #3523 export times for my 920k row dataset in the each format: 
image

ckanext/datastore/controller.py
```diff
@@ -28,7 +28,7 @@
 boolean_validator = get_validator('boolean_validator')
 
 DUMP_FORMATS = 'csv', 'tsv', 'json', 'xml'
-PAGINATE_BY = 10000
+PAGINATE_BY = 32000
```
amercader: Is this some magic number? Just curious

ckanext/datastore/logic/action.py
```diff
@@ -393,6 +393,16 @@ def datastore_search(context, data_dict):
     :param sort: comma separated field names with ordering
                  e.g.: "fieldname1, fieldname2 desc"
     :type sort: string
+    :param include_total: True to return total matching record count
+                          (optional, default: true)
+    :type include_total: bool
+    :param records_format: the format for the records return value:
+        'objects' (default) list of {fieldname1: value1, ...} dicts,
+        'lists' list of [value1, value2, ...] lists,
+        'csv' string containing comma-separated values with no header,
```
amercader: I can see how this (not returning headers) is useful to stream rows, eg to populate a widget as you scroll, but also I would expect the headers to be returned if I was using the API to get a subset of the data.

What are your views on this? Is it worth adding another parameter that allows you to return headers or not?

TkTech: It's probably worth adding, it's a pretty typical option from APIs returning CSV.

ckanext/datastore/controller.py
wardi: it comes from these benchmarks https://github.com/ckan/ckan/pull/3523#issuecomment-296304558 : there's not much benefit to going higher but 32k is a pretty big improvement over 10k, especially for csv

ckanext/datastore/logic/action.py
wardi: we already have the dump controller that streams CSV with headers. If we want to add filtering to produce custom CSVs I'd rather add it to that controller.

wardi: including a header row in this call also 100% duplicates what's returned in the "fields" list

amercader: Sorry, I thought that when using `records_format=csv/tsv` the output was actually a CSV/TSV file, not that the `records` property contained a CSV string. It makes sense now.

ckanext/datastore/plugin.py
```diff
@@ -390,8 +390,24 @@ def datastore_search(self, context, data_dict, fields_types, query_dict):
         sort = self._sort(data_dict, fields_types)
         where = self._where(data_dict, fields_types)
 
-        select_cols = [
-            datastore_helpers.identifier(field_id) for field_id in field_ids]
+        select_cols = []
+        records_format = data_dict.get(u'records_format')
+        json_values = records_format in (u'objects', u'lists')
+        for field_id in field_ids:
+            fmt = u'to_json({0})' if records_format == u'lists' else u'{0}'
```
wardi: @amercader Looks like this change requires postgres 9.3.  Should I rewrite it with the functions available in 9.2 or should we bump the postgres version requiremnet?

amercader:
Is it going to be difficult to rewrite with 9.2 support? This is meant to be in CKAN 2.7 so perhaps is a bit late to bump the Postgres requirement without warning.

PR #3541, opened 2017-04-11
[#2334] Perform DataPusher timestamp check using UTC

amercader:
Task timestamps are stored in UTC but the check was done against `.now(),` causing a test failure locally.

cc @davidread @smotornyuk


davidread:
Thanks Adria

PR #3544, opened 2017-04-14
Tweak docs for Ubuntu 16.04

### LINKED ISSUE #3543: Ubuntu 16.04 Xenial compatibility
16.04 has been out a while, so it would be good to tweak the ckan install instructions to be compatible with it.

davidread:
Fixes #3543

Depends on https://github.com/ckan/ckan/pull/3514

### DIFF
--- a/doc/maintaining/installing/install-from-source.rst
+++ b/doc/maintaining/installing/install-from-source.rst
@@ -5,11 +5,11 @@ Installing CKAN from source
 ===========================
 
 This section describes how to install CKAN from source. Although
-:doc:`install-from-package` is simpler, it requires Ubuntu 16.04 64-bit, Ubuntu
-14.04 64-bit or Ubuntu 12.04 64-bit. Installing CKAN from source works with
-other versions of Ubuntu and with other operating systems (e.g. RedHat, Fedora,
-CentOS, OS X). If you install CKAN from source on your own operating system,
-please share your experiences on our
+:doc:`install-from-package` is simpler, it requires Ubuntu 16.04 64-bit or
+Ubuntu 14.04 64-bit. Installing CKAN from source works with other versions of
+Ubuntu and with other operating systems (e.g. RedHat, Fedora, CentOS, OS X). If
+you install CKAN from source on your own operating system, please share your
+experiences on our
 `How to Install CKAN <https://github.com/ckan/ckan/wiki/How-to-Install-CKAN>`_
 wiki page.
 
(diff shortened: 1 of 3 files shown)


taviroquai:
Hi,

Just tried on a fresh Ubuntu 16.04 and here are some points that I noticed.

On point 4, the following command only applies when installing in user home, so I did not use it.
```
sudo chown -R `whoami` ~/ckan/etc
```

On point 4, when running the following command:
```
paster make-config ckan /etc/ckan/default/development.ini
```

I got the following screen:
ckan_make-config

On point 6, when initializing the database, I also got some errors (maybe related with the previous ones):
ckan_db_init

Everything else, following the install from source doc, was OK.

davidread:
The error:

    UndefinedEnvironmentName: 'extra' does not exist in evaluation environment.
 
Is detailed here: https://github.com/pypa/setuptools/issues/523 and it doesn't appear to be fixed. The workaround is to reinstall setuptools. This will be sorted in CKAN when https://github.com/ckan/ckan/pull/3514 gets merged.

Could you try again with #3514 merged into your local repo as well? Or simply simply run:

    pip install setuptools==20.4 --allow-all-external

and then try your paster command again?

taviroquai:
@davidread Yes I can confirm that the error is gone when using setuptools version 20.4.

Also, came up a warning saying that the option --allow-all-external is deprecated and has no effect.

I'm not a python dev so I maybe "swimming in unknown waters" :)

davidread:
@taviroquai Great! I've added a note to this PR to say it relies on that over PR.

smotornyuk:
Works perfectly for me

doc/maintaining/installing/install-from-source.rst
```diff
@@ -5,10 +5,10 @@ Installing CKAN from source
 ===========================
 
 This section describes how to install CKAN from source. Although
-:doc:`install-from-package` is simpler, it requires Ubuntu 14.04 64-bit or
-Ubuntu 12.04 64-bit. Installing CKAN from source works with other versions of
-Ubuntu and with other operating systems (e.g. RedHat, Fedora, CentOS, OS X).
-If you install CKAN from source on your own operating system, please share your
+:doc:`install-from-package` is simpler, it requires Ubuntu 16.04 64-bit or
```
amercader: @davidread @smotornyuk This assumes that there will be packages available for the next CKAN version :)

AFAICT https://github.com/ckan/ckan-packaging hasn't been updated to support xenial packages, although after the great work on this PR it should be really easy to build them

davidread: Good spot, I'll change that text back to "it requires Ubuntu 14.04 64-bit or
Ubuntu 12.04 64-bit" so it is consistent for the time being. I'm not sure if I'll get a chance to look at the packages too...

PR #3547, opened 2017-04-19
Bootstrap3 migration

tino097:
Migration to bootstrap version 3.3.7


tino097:
@amercader @wardi  can you please add WIP label to this PR

many thanks to @gocemitevski  for the effort :)

wardi:
@TkTech you had some comments to add to this one?

TkTech:
@wardi he isn't basing it off my PR so nope

amercader:
This :clap: is :clap: great :clap: 

@ckan/core @tino097 @gocemitevski 

We discussed what would be the migration strategy for getting this merged and agreed that Bootstrap 2 and 3 templates would need to coexist for a while to give current extensions and instances time to upgrade. The rough plan is:

* Announce the change on 2.7 (but don't include anything)
* After 2.7 is branched, work on the compatibility mechanism. Suggestion is to have both sets of templates/assets for Bootstrap 2 and 3 and load one or the other depending on a config option (defaults to 3).
* These two sets can be in core, as any new changes in templates should be applied to both sets during the transition period (within reason, will be judged on a case by case basis)
* Once we have a compatibility mechanism working, merge this PR and make the Bootstrap 3 templates the default.
* We keep the Bootstrap 2 templates/assets on 2.8 and 2.9
* We remove them completely on 2.10

Does this sound reasonable? We can tweak it as we go, but as I said, if we are following the two template set support route we need to have a clear plan and communicate it well so we don't end up dragging the old templates for ages.

wardi:
I forgot to mention on Thursday: Let's just remove the grid displays for listing Organisations. It's not worth any effort to maintain because it's only makes Orgs harder to find alphabetically, and it only helps if all your orgs have more recognisable icons than names.

amercader:
This is looking better and better @gocemitevski!

I'm sure it's already on your radar, but can we fix this paddings aorund the containers:

audhib9 1

gocemitevski:
@amercader actually, that padding is intentional, as it adds more breathing space around the main content. You might have also noticed that the new Bootstrap 3 grid is a bit wider than the older one, so everything is a bit more distanced. I think this improves the visual appearance, but if you don't agree, I'm sure I can think of something. Anyway, I most certainly think that we should gradually (not radically) evolve the UI and try to incorporate contemporary Design/UI patterns into it.

wardi:
I agree with @amercader the extra space looks wrong because the headings still have borders on some sides but not others. maybe if those borders are removed it would look better.

Can we do all design changes after the conversion instead of some half-finished changes now and some later?

gocemitevski:
@wardi sure. Not a problem at all.

tino097:
@amercader @wardi @smotornyuk  managed to make configurable selection of bootstrap version. There should be two new .ini settings 
```
ckan.templates=templates-bs3
ckan.static_files=public-bs3
```

depending of this values, frontend will be build upon chosen version

tino097:
This PR is finished! 

With this PR we should be able to choose which bootstrap version will be used in the templates. For that reason, we will have `public` and `templates` for bootstrap2 and `public-bs3` and `templates-bs3` for bootstrap3. 

The configuration is set within `.ini`  using the properties 
`ckan.templates=templates-bs3`
`ckan.static_files=public-bs3`

https://github.com/ViderumGlobal/ckan/blob/bootstrap3-migration/ckan/config/environment.py#L72
will read this properties or it will use  bootstrap2 as default

wardi:
I'd rather we reverse the names, i.e. "templates-bs2, public-bs2" for the old templates and "templates, public" for the new ones. That way when we remove bs2 we'll be left with "templates, public"

gocemitevski:
@wardi agree. @tino097 even better if we can have only one configuration setting. For example:

> ckan.bootstrap = 2
> ckan.bootstrap = 3

smotornyuk:
I found a problem on pages with sidebar - it is not available at all(Chromium v58). I can't click on facets at search page or on org/share links at dataset detail page. It looks like something overlaps sidebar, because if i set z-index to 1 for *.secondary.col-sm-3* it fixes the problem

tino097:
@wardi i've changed folder names, @gocemitevski i dont want to have hardcoded values in the code as the folder names could be subject of change

smotornyuk:
There are some problems after full review:
- User's dashboard, news feed - JS error at [this line](https://github.com/ViderumGlobal/ckan/blob/bootstrap3-migration/ckan/public/base/javascript/modules/dashboard.js#L32) - cannot read property `tip` of undefined
- 
- 
- 
- 
- 
- 
-

klikstermkd:
@smotornyuk I fixed all the issues that you reported except:

- The "Members" tab not selected. This should not be selected because the current page is Add/Edit member and not "Members" page. This is the same if you go to a certain organization, the "Organizations" tab is not selected in the menu bar. Also in previous versions of CKAN it's the same behavior.
- The incorrect position for the search icon. I already fixed this in master. It's not related to Bootstrap 3, but by another PR merged in master few weeks ago.

smotornyuk:
Cool, in that case i think that this PR should be merged after resolving conflicts

tino097:
fixed and conflicts resolved, guess its good to go

tino097:
conflicts resolved

klikstermkd:
CSS and JS files shouldn't be minified in master. Building of the frontend should be done prior a release in a release branch.

I will remove the minified CSS and JS files.

amercader:
@klikstermkd was about to say this :) Keep anything in the vendor libraries folders though (anything in `*/vendor/*`).

And while you are at it remove the `.idea` folder as well.

I'll let you know anything else after finishing my review

ckan/lib/fanstatic_resources.py
```diff
@@ -227,9 +228,15 @@ def create_resource(path, lib_name, count, inline=False):
     registry = get_library_registry()
     registry.add(library)
 
-base_path = os.path.abspath(os.path.join(os.path.dirname(__file__),
-                                         '..', 'public', 'base'))
 
+from ckan.common import config
```
amercader: Can you move this import to the top?

tino097:
@amercader as for the folder names, that  can be changed, but i would keep both config options as we will have less hardcoded values in the code. Your suggestion is ok, and i agree with `ckan.base_templates_folder` and `ckan.base_public_folder` , if anyone have other idea please share it with us.

amercader:
Cool, let's go for these names and add the docs and I'll merge.

klikstermkd:
@amercader If an invalid value is provided for the config options, for example:

`ckan.base_public_folder = invalid_value`
`ckan.base_templates_folder = invalid_value`

Should an exception be raised that the provided value for the specific config option is invalid, or fallback to their default values, `public` and `templates` respectively?

amercader:
@klikstermkd raise an exception with a clear message please. It's best to do what the user expects to happen, or not do it at all.

klikstermkd:
@amercader I added helper functions for getting the base public and templates folder. I think it's better like that instead of using `config.get()` everywhere where it's needed. Additionally this way an exception can be raised for invalid values.

ckan/lib/helpers.py
```diff
@@ -2344,6 +2344,35 @@ def mail_to(email_address, name):
     return html
 
 
+@core_helper
+def get_base_public_folder(app_conf):
```
wardi: ckan/lib/helpers are functions for templates to call. It makes no sense for a template to need a directory, could we move these closer to where they're actually needed?

klikstermkd: The functions are used in multiple files, so I thought `helpers.py` is the right location to include them into.

What do you suggest as other location?

wardi: this is something site-configuration related and some of the extensions need it but it seems silly to add to the plugins toolkit. How about just loading the default into the config object when reading the config?

klikstermkd: Well that was how it was done before I added the helper functions. But in that way how are we going to raise an exception for invalid value? We can't just raise exceptions everywhere where it needs to be read from config.

wardi: we can have it fail in update_config like it does when site_url isn't specified. (an exception other than RuntimeError would be better of course)

klikstermkd: If we raise an exception in `update_config`, it won't work when doing `paster less`, because the `ckan.base_public_folder` config is used there.

klikstermkd: Or to be more clear, the `update_config` function is not called when doing `paster less`.

wardi: let's be boring and have just one way of loading our configuration. If LessCommand calls self._load_config like the other CkanCommand subclasses then it should work.

wardi: note: I say this as the person that added a second way to load the configuration :smile: but in my defence the new way is much simpler.

klikstermkd: I removed the helpers. Can you take a look at now if it looks good to you?

wardi: looks great, thanks! Just trying to keep lib/helpers from becoming a dumping ground for every utility function in ckan.

klikstermkd: Yes it makes sense when I think about it.

tino097:
@amercader @wardi i added some text to the theming guide, but please check it for grammar :)

klikstermkd:
@amercader @wardi I added additional docs for how to properly set up the new config options.

amercader:
Merged now, many many thanks to @gocemitevski, @tino097, @klikstermkd and anyone else involved in this, it's a big step forward for CKAN! :tada: :confetti_ball: :rocket: 

I've created a new issue with small things that can be fixed separately: https://github.com/ckan/ckan/issues/3678

PR #3549, opened 2017-04-20
download icon for no_view

hvwaldow:
### Proposed fixes:
If there are no resource views available (no view plugins loaded) but the resource is in the Filestore, then

1. the *Explore* dropdown-menu on the dataset-page will have an entry "*Go to resource*", and
2. the top right button on the resource-page will be labeled "*Go to resource*"

Both times this is misleading, as "*Download*" would be appropriate. 

### Features:

- [ ] includes API changes


smotornyuk:
It has sense and it works. Verified

PR #3553, opened 2017-05-02
Remove exposed ports

mattfullerton:
No need to expose ports externally when using the container names already provides the necessary access to the ports internally from the containers.

### DIFF
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -12,17 +12,11 @@ ckan:
 db:
   container_name: db
   image: ckan/postgresql:latest
-  ports:
-    - "5432:5432"
   
 solr:
   container_name: solr
   image: ckan/solr:latest
-  ports: 
-    - "8983:8983"
     
 redis:
   container_name: redis
   image: redis:latest
-  ports: 
-    - "6379:6379"


deinok:
@mattfullerton well done, also we should move to the latest docker-compose file. But we probably have problems with things like URLs

wardi:
@canoas would you provide some examples of the errors you're seeing? Maybe we need to add some more tests of the docker images.

mattfullerton:
@canoas and/or post what ckan.ini you ended up with, i.e.
`docker-compose exec ckan bash`
`cat /etc/ckan/default/ckan.ini`

canoas:
Not on my computer now, but I remember i've seen many errors regarding internal connections failing with urls like http://localhost:/....
As i started to re-enable de exposed ports, new ones were appearing which means there were still internal references/configurations using the previous ports.
Btw: exposing the ports as a workaround was not sufficient to make the ckan instance start/run. That's when i asked here if the commit was complete.

Sent from my Mobile

-------- Original Message --------
Subject: Re: [ckan/ckan] Remove exposed ports (#3553)
From: Matt Fullerton
To: ckan/ckan
CC: Jose Antonio Silva ,Mention


@canoas<https://github.com/canoas> and/or post what ckan.ini you ended up with, i.e.
docker-compose exec ckan bash
cat /etc/ckan/default/ckan.ini

mattfullerton:
Well it should just work:
https://github.com/ckan/ckan/blob/master/contrib/docker/my_init.d/50_configure#L3

But redis might be the problem; that mechanism of setting URLs for Solr and Postgres hasn't been applied for the redis URL, and it supplies its own default of locahost:
https://github.com/ckan/ckan/blob/a5408647c18bcc3a6317a8c83cfb37c72068e90c/ckan/lib/redis.py#L23

I.e. what's needed is this: https://github.com/derilinx/ckan-docker/blob/master/etc/my_init.d/50_configure#L40

I'll add that in but I need to get on to a machine I can test it on before resubmitting. Seeing the ckan.ini you ended up with in the container and any error logs would speed me along.

mattfullerton:
I finally got on to a machine where I can test this today and have a few insights. The main problem is this:
https://github.com/ckan/ckan/blob/master/contrib/docker/ckan-entrypoint.sh#L29
-- the container names (internal host names) aren't written to the config

Beyond that, there can be the problem that CKAN doesn't connect to the DB because it hasn't started yet. This problem might have now first reared its head because Docker when it maps a port to the outside world will not drop the connection if the internal service hasn't started yet (maybe, !???). In any case the fix is to do docker-compose up -d ckan afterwards to restart CKAN.

With uncommenting that stuff in ckan-entrypoint.sh I get CKAN connecting to the other services without exposed ports but I have another problem that paster's DB init crashes. What release of CKAN is the docker build based on? Master?

PR #3554, opened 2017-05-02
Support of stream responses

smotornyuk:
Very simple change that can be backported(at least, change inside **pylons_app.py**) to previous versions. 
I added one function to `ckan.common`, that can convert something iterable to stream reponse, which can be handled(i.e addition of new headers) as usually. For example:
```python

@pylons or flask route
def action(args):
    data = compute_iterable() # any list, string or anything that can be passed to built-in iter()
    # or
    def generator():
          yield chunk
          yield chunk
    data = generator()


    response = streaming_response(data)
    # and if you want to change headers:
    response.headers['content-type'] = 'text/plain'
    return data

```

Does anyone have suggestions? For example, currently 'text/html' is considered to be default content type, but it has(probably) more sense to change it to 'text/plain' or 'application/octet-stream' for streams? 
:information_source:  and you'll get error because of utf-8 characters if you try to send html stream, so in should be fixed

PS and does anyone have idea, how to test(programatically) stream responses?


wardi:
@TkTech this is a small flask improvement, do you have time to review?

TkTech:
I do, will review tomorrow.

TkTech:
@wardi there's a small conflict but otherwise I'm fine with it. Do we want it in 2.7?

wardi:
I don't think it hurts, nothing in core will be using it yet

amercader:
I think it's a big late in the day for 2.7 and if this is Flask-related and still not used perhaps it's a better idea to merge it to master and test it until the next release.

amercader:
@smotornyuk I know there are no conflicts but just because this is quite old, anything you want to add / check before merging this?

smotornyuk:
I'll take a look at it and try to check everything before tomorrows meeting

smotornyuk:
I merged master into this branch and everything is still working:) Review completed, nothing should be changed

amercader:
@smotornyuk Considering the extent of the Flask migration included in 2.7 do you think it's worth backporting this? I don't mind either way, I just want to make sure it would be useful at all in 2.7 so it's worth the risk.

smotornyuk:
This code doesn't change much, so it should be quite safe to merge it into 2.7, but actually, I do not want to:) 2.7 is last relatively "clean-from-flask" release, so it probably shouldn't contain any new features. if anyone wants to use such feature in their portal with an old version, it quite easy to cherry pick these commits. And for those, who want all fresh cool features it would be much better to switch to v2.8+. Writing new code with streaming responses on pre-flask versions has no sense because it will require reworking during the upgrade, so no one with v2.7 will use it anyway

amercader:
That was my impression too, just wanted to confirm thanks!

PR #3559, opened 2017-05-09
Add localization support for groups

### LINKED ISSUE #3558: Groups should support multilingual schema fields
Group templates are missing translation support for multilingual fields. E.g. templates should call get_translated for displayed group fields. I'll provide suggestions with a PR in a minute.

mikkonieminen:
# Fixes
#3558 

### Proposed fixes:

Fixes missing support for multilingual group fields in group templates.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/group/about.html
+++ b/ckan/templates/group/about.html
@@ -3,11 +3,19 @@
 {% block subtitle %}{{ _('About') }} - {{ super() }}{% endblock %}
 
 {% block primary_content_inner %}
-  <h1>{% block page_heading %}{{ c.group_dict.display_name }}{% endblock %}</h1>
+  <h1>
+    {% block page_heading %}
+      {% if c.group_dict.title %}
+        {{ h.get_translated(c.group_dict, 'title') }}
+      {% else %}
+        {{ c.group_dict.display_name }}
+      {% endif %}
+    {% endblock %}
+  </h1>
   {% block group_description %}
-    {% if c.group_dict.description %}
-      {{ h.render_markdown(c.group_dict.description) }}
-  	{% endif %}
+    {% if h.get_translated(c.group_dict, 'description') %}
+      {{ h.render_markdown(h.get_translated(c.group_dict, 'description')) }}
+    {% endif %}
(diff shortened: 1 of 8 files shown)


wardi:
@mikkonieminen thanks! would you take a look at the test failures?

mikkonieminen:
@wardi: fixed

ckan/templates/group/about.html
```diff
@@ -3,14 +3,22 @@
 {% block subtitle %}{{ _('About') }} - {{ super() }}{% endblock %}
 
 {% block primary_content_inner %}
-  <h1>{% block page_heading %}{{ c.group_dict.display_name }}{% endblock %}</h1>
+  <h1>
+    {% block page_heading %}
+      {% if c.group_dict.title %}
```
wardi: this doesn't look right. The translated version will be stored in `title_translated` so checking the title relies on the correct validator being set up for `title`. Why don't you follow the `h.get_translated(c.group_dict, 'title') or c.group_dict.display_name` pattern from below instead?

ckan/templates/group/snippets/info.html
```diff
@@ -11,16 +11,20 @@
...
         [{{ _('Deleted') }}]
       {% endif %}
     </h1>
     {% endblock %}
     {% block description %}
-    {% if group.description %}
+    {% if h.get_translated(group, 'description') %}
       <p>
-        {{ h.markdown_extract(group.description, 180) }}
+        {{ h.markdown_extract(h.get_translated(group, 'description'), 180) }}
```
wardi: you're not extracting the description from the current language here

ckan/templates/group/about.html
mikkonieminen: True, had the same idea before but forgot to actually do the fix. I'll commit the same fix to the other remaining templates in a moment.

mikkonieminen:
For the localization to actually work in group index the include_extras needs to be set to true for the group_list action call in GroupController. Is this a valid solution performance wise?

PR #3568, opened 2017-05-18
email validation for user input

tino097:
Fixes #
user input for email when creating datasets

### Features:

- [ x] includes user-visible changes
- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -831,3 +831,10 @@ def empty_if_not_sysadmin(key, data, errors, context):
         return
 
     empty(key, data, errors, context)
+
+def email_validator(value, context):
+    '''Validate email input '''
+    if value:
+        if not re.match("\A(?P<name>[\w\-_]+)@(?P<domain>[\w\-_]+).(?P<toplevel>[\w]+)\Z",value,re.IGNORECASE):
+            raise Invalid(_('Email "%s" is not valid format' % (value)))
+    return value
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -832,9 +832,16 @@ def empty_if_not_sysadmin(key, data, errors, context):
 
     empty(key, data, errors, context)
 
+#pattern from https://html.spec.whatwg.org/#e-mail-state-(type=email)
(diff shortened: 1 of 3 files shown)


ckan/logic/validators.py
```diff
@@ -831,3 +831,10 @@ def empty_if_not_sysadmin(key, data, errors, context):
         return
 
     empty(key, data, errors, context)
+
+def email_validator(value, context):
+    '''Validate email input '''
+    if value:
+        if not re.match("\A(?P<name>[\w\-_]+)@(?P<domain>[\w\-_]+).(?P<toplevel>[\w]+)\Z",value,re.IGNORECASE):
```
TkTech: Line is too long, PEP8.

ckan/logic/validators.py
```diff
@@ -831,3 +831,10 @@ def empty_if_not_sysadmin(key, data, errors, context):
...
+            raise Invalid(_('Email "%s" is not valid format' % (value)))
```
TkTech: This string can't be translated. Should be:

`_('Email "%s" is not valid format') % value`

But even better,

`_('Email {email} is not a valid format').format(email=email)`. It's good practice to use named blocks when doing translations.

ckan/logic/validators.py
wardi: This is the one we're using, based on the HTM5 spec for email validation https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/validators.py#L112-L125

ckan/logic/validators.py
TkTech: Actually, I just realized showing the failed validation string in a flash message is...not ideal...

tino097: what are you suggesting ?!

amercader:
Tests complain `Email test_author@test_author.com is not a valid format` which seems pretty valid to me.

adamamyl:
@amercader Underscores aren't valid in domain names (maybe use something from https://tools.ietf.org/html/rfc2606#section-3 ?).

amercader:
Thanks @adamamyl! You learn something everyday :)

Then it seems like it's just a matter of updating the test.

tino097:
@adamamyl thank you !!!

It's ok now

amercader:
Thanks @tino097

PR #3572, opened 2017-05-22
Apply env. variables to CKAN config

mattfullerton:
Complements https://github.com/ckan/ckan/pull/3553 by applying the environment variables. I'm not sure if this will really work on its own, but adding the environment variables (DB_PORT_5432_TCP_ADDR etc.) in the docker-compose.yml environment section will definitely work.

### DIFF
--- a/contrib/docker/ckan-entrypoint.sh
+++ b/contrib/docker/ckan-entrypoint.sh
@@ -26,13 +26,13 @@ write_config () {
 
   ckan-paster make-config ckan "$CONFIG"
 
-  # In case want to use the config from ckan.ini use this
-  #ckan-paster --plugin=ckan config-tool "$CONFIG" -e \
-  #    "sqlalchemy.url = ${CKAN_SQLALCHEMY_URL}" \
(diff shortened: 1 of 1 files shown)


wardi:
IIUC these settings shouldn't need to be in the ini file if they are in the environment when running ckan. Are they getting cleared from the environment in some cases?

mattfullerton:
That's the first time I ever heard of CKAN taking an .ini setting from an environment variable?

wardi:
It is a thing, yes https://github.com/ckan/ckan/blob/c871c719d148027fa54663a3decf428bf0b749d3/ckan/config/environment.py#L117-L136

amercader:
Not relevant here, but there is also ckanext-envvars that allows setting (almost) all settings via environment variables.

mattfullerton:
@wardi I see what you mean. I'm pretty sure we need to pass through the container names into those env. variables; the misleading comments in the script give the impression something will magically happen if the container is linked to one called "db", "solr", etc. but I don't think that's the case. Closing this and put the other change in #3582.

PR #3576, opened 2017-05-24
[#3146] Rename example theme plugin internally

### LINKED ISSUE #3146: ImportError: No module named plugin when creating a new theme extension

2.5.2

Starting CKAN with a new plugin based on my custom extension

I can't start Ckan after creating a new exension to define my own theme.
I followed the 2.5.2 guide , but i am getting an error message :

ImportError: No module named plugin
- Create new extension with the command :

``` sh
paster --plugin=ckan create -t ckanext ckanext-example_theme
```
- Edit setyp.py to include extension class in entry points
- python setup.py develop
- Add the new plugin in the my development.ini file

amercader:
Fixes #3146 

See issue for background. When users were following the theming tutorial they got an ImportError because the name suggested (`example_theme`) was already used by the example plugin itself.

This patch renames the internal plugin and updates the code imports shown on the docs, but not the docs themselves, so users should not see any change externally, and can use `example_theme` without conflicts. And if someone wants to chek the sources it's pretty obvious that they should look at `example_theme_docs`.

@davidread WDYT? I think this should work fine


davidread:
This is a good solution - thanks @amercader 

The only thing to note is that those who have already run the version before this change will have to get rid of the pyc files in the old ckan/ckanext/example_theme directory, to make the error go away. e.g.

    rm -rf ckan/ckanext/example_theme

PR #3577, opened 2017-05-25
get_translated: fall back to .po files

fanjinfei:
if the translated language for the field is not provided by a `*_translated` value, then we should try to find it from .po files.

### Features:

- [ ] includes API changes


wardi:
what do you think about this change @amercader

amercader:
Backported to 2.7

Issue #2938, opened 2016-04-01
Odd feature in sorting organizations by packages


junsk1:
In our Finnish Etsin CKAN portal we sort data producer organizations by default on the basis of the number of datasets they have published, which is achieved by setting sort == 'packages'. A confusing feature is, however, the way 0 dataset organizations are displayed, which are mixed in order at the end of the sort list. This is caused by a sql query which also sums up the datasets, which have been transferred to another organizations  and thus the membership is not active anymore. 

For example: if the organization has had 5 datasets but now has zero, the aforementioned occurs.

Code: in logic.get.py method group_list() when sort_info == 'package_count':  Does not check if the member record is active, so if a package has moved from organization to another, the package is still active and the old organization membership is not. 

CKAN Version: 2.4.1
Site URL: https://etsin.avointiede.fi/organization

amercader:
That makes sense @junsk1 . So if I understand correctly it's just a matter of adding `` here, right?

https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L368

Can you send a quick PR with the proposed fix? You can add a small test here to make sure the fix gets applied: https://github.com/ckan/ckan/blob/master/ckan/tests/logic/action/test_get.py#L399

wildcatzita:
@amercader, I can start working on it if it need.

amercader:
@wildcatzita That would be great! I updated the comment with the proposed fix

wildcatzita:
@amercader , ok. I'll fix it.

Engerrs:
Hi @amercader , I've made fix for that one. Also I've added sorting in Order by on organization page, for sort by datasets and made another fix in query because it counted private datasets, while normal users can't see private datasets. Sysadmin can see counted private datastes on organization page, but it will sort only for those that are public.
Added test for all of this logic.

PR #3578, opened 2017-05-27
[#2938] Org list sort package_count added to UI and sort logic modification

Engerrs:
Adding package_count sorting to UI for Organization list.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -320,6 +320,7 @@ def member_list(context, data_dict=None):
 
 def _group_or_org_list(context, data_dict, is_org=False):
     model = context['model']
+    user = context.get('user_id')
     api = context.get('api_version')
     groups = data_dict.get('groups')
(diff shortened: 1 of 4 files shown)


Engerrs:
Hi @amercader , I've added a full PR with all changes in one.

wardi:
actually, we can get counts of datasets for all orgs from the facets returned by package_search, and those numbers will always take permissions into account.

Engerrs:
@wardi, its a good idea, I'll rewrite then the logic for this one.

Engerrs:
@wardi , it appears that the method is done by one big query and in the result he got the counts of datasets from **organization_show** method, which displays the right count already.

wardi:
@Engerrs I'm saying this is duplicating code and only shows the correct counts if the site hasn't defined a custom policy for dataset visibility with an IPermissionLabels plugin.

If we make one call to package_search with limit=0 the facet counts returned will always have the correct count.

kmbn:

Issue #3579, opened 2017-05-28
My organizations on dashboard page show wrong number of datasets for organization.


Engerrs:
CKAN Version
All versions

Expected behaviour
When logged user switch to My organizations tab at dashboard, the number of datasets for the organization should match the number of datasets that found on **/organization** page.

Actual behaviour
The number of datasets are not match the number of datasets on **/organization** page.

To reproduce the issue there should be at least 1 organization with couple of public datasets and at least one private.

tino097:
#3580

PR #3580, opened 2017-05-28
[#3579]  Fix for number of datasets that displayes on My orgs tab

Engerrs:
It seems that the problem was with the query that showed only public datasets, while /organization page showed with private for users that at least members of those datasets.


tino097:
in future releases this will be subject of changes using correct api actions

kmbn:

Issue #2794, opened 2015-12-17
Add "Add Dataset" button to user's and group's page


vitorbaptista:
As far as I could see, every page which list datasets have the "Add Dataset" button, as:
## Search datasets

![search datasets](https://cloud.githubusercontent.com/assets/76945/11870414/646f8fe2-a4a6-11e5-821f-f9ab23baa8dd.png)
## Organization's page

![organization page](https://cloud.githubusercontent.com/assets/76945/11870415/67345be0-a4a6-11e5-8b4d-b97e85d55374.png)
## User's dashboard

![user dashboard](https://cloud.githubusercontent.com/assets/76945/11870416/68662e8a-a4a6-11e5-983e-4b8b3c8f3a2c.png)

The only pages that list datasets and don't add these links are:
## User's page

![users page](https://cloud.githubusercontent.com/assets/76945/11870446/c6147280-a4a6-11e5-8cf2-2f3d0666270c.png)
## Group's page

![group page](https://cloud.githubusercontent.com/assets/76945/11870451/c99ce504-a4a6-11e5-98da-7acb0bc89cbc.png)

This feels inconsistent, so I would suggest adding to those places as well.

vitorbaptista:
The code that adds the "Add Datasets" buttons is repeated in each place as well, so I would suggest extracting that to a snippet. This way, it can be easily overwritten by templates/extensions (I'm doing that on https://github.com/ckan/ckanext-datapackager to add a "Import Data Package" button)

rossjones:
@vitorbaptista Can I assign this to you for a PR?

vitorbaptista:
@rossjones sure! :+1:

Engerrs:
Hi @amercader, I've added a PR for this one.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

If this issue is still affecting the version of CKAN you're working with now, please feel free to comment or reopen the issue.

If you do reopen this issue, please update it with new details. One reason it might not have been resolved in the past is that it wasn't clear how a contributor could address the issue.

PR #3583, opened 2017-05-30
[#2794] Moved Add dataset button to snippet

Engerrs:
Moved 'Add Dataset' button to snippet as it used like 4 time in different templates.
About adding this button 'Groups' page, is not correct, because CKAN logic done is such way that Datasets can add Groups, not Groups create Datasets. Also adding this button to the page, means that when we create new Dataset on this page, it should be related to this group from the start, which means that new logic for this one should be added to 'new' method in 'package' controller or creating new method in group controller that will handle this, but its not related to this ticket.
We can add a simple button that will create new Dataset it wont make any effect creating it from Groups page.


smotornyuk:
@Engerrs , could you resolve conflict please?

Engerrs:
Hi @smotornyuk, I've solved this one.

Issue #3588, opened 2017-06-02
Upload to Datastore tab shows incorrect time at Upload Log


Engerrs:
CKAN Version
2.5,2.6,2.7

Expected behaviour
When Upload to Datastore, the time in table and at logs, shows display correctly.

Actual behaviour
When Upload to Datastore, the time at logs, shows display incorrectly. -1 day

Steps to reproduce the issue 
1.Configure datastore and datapusher.
2.Create dataset with CSV resource.
3.Enter the Upload to Datastore tab and click on the button, reset the page.

amercader:
Fixed on #3589

PR #3589, opened 2017-06-02
[#3588] Fix for Upload Log time at resource Datastore tab

Engerrs:
Added a fix for the time that display at Upload Log, it seems that **job_detail** at datapusher_status action gives already the time with +3 hours.
I've changed the variable that template uses for the display to the one that used in the table as the Upload Log will be match to the value in the table.

### DIFF
--- a/ckanext/datapusher/logic/action.py
+++ b/ckanext/datapusher/logic/action.py
@@ -290,10 +290,10 @@ def datapusher_status(context, data_dict):
             job_detail = r.json()
             for log in job_detail['logs']:
                 if 'timestamp' in log:
-                    date = datetime.datetime.strptime(
+                    date = time.strptime(
                         log['timestamp'], "%Y-%m-%dT%H:%M:%S.%f")
                     date = datetime.datetime.utcfromtimestamp(
-                        time.mktime(date.timetuple()))
+                        time.mktime(date))
                     log['timestamp'] = date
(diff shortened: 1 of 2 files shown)


smotornyuk:
@Engerrs , as i understand, you suggesting to use time of last status change, that is the same for each [log record](https://github.com/Engerrs/ckan/blob/83b7c2dbf02501400bc27ab35cbb65f48b5f13d4/ckanext/datapusher/templates/datapusher/resource_data.html#L64). In that case information for tasks that are executing for a while, becomes irrelevant. It would be much better to use item's timestamp as it was before, but use `datetime.utcnow` it the point of item creation

Engerrs:
Hi @smotornyuk , I've add a small fix for timestamp, rather then change the item used in template, but I couldn't change the timestamp, because the info that it gets for display is set in [**ckanserviceprovider**
package](https://github.com/ckan/ckan-service-provider/blob/master/ckanserviceprovider/util.py#L53) and save if in our **job_store.db** file. So I've found another way to deal with the timestamp for logs by converting it to UTC format from it's current.

smotornyuk:
As i understand, you can use `time.strptime` instead of `datetime.strptime`. It's not critical, but more obvious

Engerrs:
Hi @smotornyuk, changed that one.

smotornyuk:
@Engerrs , thanks)

PR #3595, opened 2017-06-07
fix broken language toggle url when there is params in current url

fanjinfei:
Fixes #
if current url is /en/dataset?q=xxxx, then change to another language will generate bad url.
### Features:

- [ ] includes API changes


tino097:
Tested, works fine

Issue #3594, opened 2017-06-06
requests dependency failure


mattfullerton:
### CKAN Version if known (or site URL)
master running inside Docker

### Please describe the expected behaviour
CKAN starts

### Please describe the actual behaviour
This is inside Docker, at CKAN startup (stack trace at end). The paster command `ckan-paster make-config ckan "$CONFIG"` is completing successfully and this is presumably at `ckan-paster --plugin=ckan db init -c "${CKAN_CONFIG}/ckan.ini"`. In the end the complaint is that urllib3 `must be version >= 1.21.1, < 1.22`. From debian:jessie we get "python-urllib3 (1.9.1-3)" and later in the build 1.7.1 is installed:
`Downloading/unpacking urllib3==1.7.1 (from httpretty==0.8.3->-r /usr/lib/ckan/default/src/ckan/dev-requirements.txt (line 8))`
(but that ought to be irrelevant as its a sub-dependency)

Not sure why I never saw this before; does it warrant adding that requirement explicitly to requirements.txt?
(the workaround right now is to do `ckan-pip install --upgrade urllib3` in the Dockerfile)

### What steps can be taken to reproduce the issue? 
My exact setup is docker and docker-compose on Mac - CKAN docker image created manually with docker build and tagged; this tag used in the docker-compose file. But the CKAN version used is master, pulled today.

---
```
ckan     | Traceback (most recent call last):
ckan     |   File "/usr/local/bin/ckan-paster", line 11, in <module>
ckan     |     sys.exit(run())
ckan     |   File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 102, in run
ckan     |     invoke(command, command_name, options, args[1:])
ckan     |   File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 141, in invoke
ckan     |     exit_code = runner.run(args)
ckan     |   File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 236, in run
ckan     |     result = self.command()
ckan     |   File "/usr/lib/ckan/default/src/ckan/ckan/lib/cli.py", line 334, in command
ckan     |     self._load_config(cmd!='upgrade')
ckan     |   File "/usr/lib/ckan/default/src/ckan/ckan/lib/cli.py", line 303, in _load_config
ckan     |     load_config(self.options.config, load_site_user)
[...]

PR #3596, opened 2017-06-08
docker-compose without exposing ports - working version

### LINKED ISSUE #3569: [DOCKER] docker-compose up not working
Running `docker-compose up -d` on the `docker-compose.yml` file in `contrib/docker` seems to result in the `ckan` container not running? When inspecting the `ckan` container logs it shows me this:
```
Distribution already installed:
  ckan 2.7.0a0 from /usr/lib/ckan/default/src/ckan
Creating /etc/ckan/default/ckan.ini
Now you should edit the config files
  /etc/ckan/default/ckan.ini
2017-05-18 17:46:43,354 ERROR [ckan.lib.redis] Redis is not available
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/redis.py", line 61, in is_redis_available
    return redis_conn.ping()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/redis/client.py", line 682, in ping
    return self.execute_command('PING')
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/redis/client.py", line 578, in execute_command
    connection.send_command(*args)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/redis/connection.py", line 563, in send_command
    self.send_packed_command(self.pack_command(*args))
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/redis/connection.py", line 538, in send_packed_command
    self.connect()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/redis/connection.py", line 442, in connect
    raise ConnectionError(self._error_message(e))
ConnectionError: Error 111 connecting to localhost:6379. Connection refused.
2017-05-18 17:46:43,355 CRITI [ckan.config.environment] Could not connect to Redis.
  File "/usr/local/bin/ckan-paster", line 11, in <module>
    sys.exit(run())
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 102, in run
    invoke(command, command_name, options, args[1:])
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 141, in invoke
    exit_code = runner.run(args)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/paste/script/command.py", line 236, in run
    result = self.command()
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/cli.py", line 334, in command
    self._load_config(cmd!='upgrade')
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/cli.py", line 303, in _load_config
    load_config(self.options.config, load_site_user)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/cli.py", line 221, in load_config
    load_environment(conf.global_conf, conf.local_conf)
  File "/usr/lib/ckan/default/src/ckan/ckan/config/environment.py", line 103, in load_environment
    p.load_all()
  File "/usr/lib/ckan/default/src/ckan/ckan/plugins/core.py", line 129, in load_all
    unload_all()
  File "/usr/lib/ckan/default/src/ckan/ckan/plugins/core.py", line 182, in unload_all
    unload(*reversed(_PLUGINS))
  File "/usr/lib/ckan/default/src/ckan/ckan/plugins/core.py", line 210, in unload
    plugins_update()
  File "/usr/lib/ckan/default/src/ckan/ckan/plugins/core.py", line 121, in plugins_update
    environment.update_config()
  File "/usr/lib/ckan/default/src/ckan/ckan/config/environment.py", line 266, in update_config
    model.init_model(engine)
  File "/usr/lib/ckan/default/src/ckan/ckan/model/__init__.py", line 157, in init_model
    version_table = Table('migrate_version', meta.metadata, autoload=True)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/sql/schema.py", line 352, in __new__
    table._init(name, metadata, *args, **kw)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/sql/schema.py", line 425, in _init
    self._autoload(metadata, autoload_with, include_columns)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/sql/schema.py", line 448, in _autoload
    self, include_columns, exclude_columns
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1670, in run_callable
    with self.contextual_connect() as conn:
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1737, in contextual_connect
    self.pool.connect(),
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 332, in connect
    return _ConnectionFairy._checkout(self)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 630, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 433, in checkout
    rec = pool._do_get()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 949, in _do_get
    return self._create_connection()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 278, in _create_connection
    return _ConnectionRecord(self)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 404, in __init__
    self.connection = self.__connect()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/pool.py", line 530, in __connect
    connection = self.__pool._creator()
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/engine/strategies.py", line 95, in connect
    connection_invalidated=invalidated
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/util/compat.py", line 189, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/engine/strategies.py", line 89, in connect
    return dialect.connect(*cargs, **cparams)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/sqlalchemy/engine/default.py", line 376, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/psycopg2/__init__.py", line 179, in connect
    connection_factory=connection_factory, async=async)
sqlalchemy.exc.OperationalError: (OperationalError) could not connect to server: Connection refused
	Is the server running on host "172.17.0.3" and accepting
	TCP/IP connections on port 5432?
 None None
```

Any ideas on what goes wrong?

mattfullerton:
Fixes #3569 

- Apply environment variables, always; no need to use ckan.ini (see https://github.com/ckan/ckan/pull/3572#issuecomment-303596016)
- Allow connection to postgres via docker instead of coming via external port mapping
- Includes a workaround for #3594
- ~~~One remaining problem: the redis URL does not make it into CKAN despite using correct environment variable(?). Hence WIP label; but this will commit will already improve those suffering from completely broken docker-compose functionality.~~~


mattfullerton:
@wardi This is now working well. The Redis issue was a general one that would have affected anyone depending on environment variables, and I've also figured out the relationship between the cryptically names environment variables in the entrypoint script and those [needed to be] passed through from docker-compose.

contrib/docker/docker-compose.yml
```diff
@@ -1,4 +1,8 @@
 # docker-compose build && docker-compose up
+# If you experience problems with the CKAN container not being
+# able to connect to the DB, then most likely the DB has not
+# started up quickly enough. Just do "docker-compose up ckan"
+# again to retry
```
wardi: would a hack like https://github.com/datacats/datacats/blob/master/datacats/scripts/install_postgis.sh#L10-L15 be of any use to us?

SergK: probably this can be done from separate container which will configure ckan using https://github.com/jwilder/dockerize since we also need to be sure that all redis, solr, pg are up and running. I'll try to provide PR

SergK: but probably this one can be used in ckan-entrypoint.sh, rather simple, but we need to install netcat package as well
https://github.com/GSA/catalog-app/blob/master/docker/webserver/entrypoint.sh#L12-L14

contrib/docker/ckan-entrypoint.sh
```diff
@@ -58,29 +61,30 @@ link_redis_url () {
 
 # If we don't already have a config file, bootstrap
 if [ ! -e "$CONFIG" ]; then
+  write_config
+fi
```
SergK: I think we need to move this at LL:87-88

contrib/docker/docker-compose.yml
Cdower: adding 
`depends_on:
    - redis`
to the ckan service will likely help with this

wardi:
@mattfullerton I'd like to merge this so we can get it in 2.7, would you like to make any more changes first?

SergK:
@wardi I have issue with this patch https://github.com/ckan/ckan/pull/3596/commits/466fe093e891f672516cc30714f6a363e3364526 it doesn't work for me
And IMO write_config should be called just after set_environment

mattfullerton:
Will get back to you all tomorrow (Wednesday)

contrib/docker/ckan-entrypoint.sh
mattfullerton: Why? All this does is create a fresh config if none exists. The stuff in set_environment is setting environment variables that are independent of the config file.

contrib/docker/docker-compose.yml
mattfullerton: @Cdower We don't have that in docker-compose file format version 1. There's actually no good reason not to move to version 2 or 3, but I was originally trying to make a minimal change and didn't want to take that step.

Beyond that though, I'm not sure depends_on will help anything. Docker explicitly state: "However, Compose will not wait until a container is “ready” (whatever that means for your particular application) - only until it’s running." - https://docs.docker.com/compose/startup-order/

mattfullerton:
@SergK Please note that testing is very tricky because the current published ckan image on docker (will not update till PR merged) is not what you need. To test you need to build the ckan image manually with docker and then specify the image in docker-compose (possibly specifying the docker file with a relative path will work too).

contrib/docker/ckan-entrypoint.sh
SergK: hmm, probably I misunderstand, but need to check. The case is, what if I need to change default values for connection string to redis and pg

mattfullerton: Then you just need to pass the variables $CKAN_SQLALCHEMY_URL, $CKAN_SOLR_URL or $CKAN_REDIS_URL to the container

mattfullerton:
re testing: https://github.com/mattfullerton/ckan/blob/8a91622ed52905c957fb5e1bee0b4f93fb6313c4/contrib/docker/docker-compose.yml#L9

contrib/docker/ckan-entrypoint.sh
SergK: Hm, I thought that the logic was, always fall back to variables defined in compose:

    environment:
        - DB_ENV_POSTGRES_USER=ckan
        - DB_ENV_POSTGRES_PASSWORD=ckan
        - DB_ENV_POSTGRES_DB=ckan
        - DB_PORT_5432_TCP_ADDR=db
        - DB_PORT_5432_TCP_PORT=5432
        - SOLR_PORT_8983_TCP_ADDR=solr
        - SOLR_PORT_8983_TCP_PORT=8983
        - REDIS_PORT_6379_TCP_ADDR=redis
        - REDIS_PORT_6379_TCP_PORT=6379

so first of all we need to grad those variables, and then proceed with config

SergK:
I'm just working on running ckan with UWSGI in docker, https://github.com/SergK/ckan/commit/5e0a54b5938756eed8b46f6df0894ad779cbd17b so that's why tested everything

mattfullerton:
@SergK Remember we are modifying the CKAN docker image here and not necessarily working with compose. There are three scenarios:

1. Someone has non-docker Postgres etc. but wants to dockerize CKAN
2. Someone doesn't like docker-compose and starts up containers manually and then starts CKAN with links to those containers
3. Someone uses docker-compose

in 2 & 3, we get those variables automatically BUT in (2) the hostname is "wrong" (its an IP). That's why in the current PR only hostnames are overwritten from docker-compose (what you pasted is not from the current PR). In 2 they are provided by Docker as IPs and should also work. In (1) you have to set the finished env. variables ($CKAN_SQLALCHEMY_URL etc.) yourself when starting the container.

All of that said, this started as a docker-compose issue and I'm only testing with compose.

mattfullerton:
@wardi This PR works but I haven't been able to resolve the postgres startup problem despite trying to add in your suggested hack. The file documents the course of action if this problem occurs and all-in-all this is a major security improvement compared to exposing those ports. I would suggest merging and opening a new issue for this remaining problem, potentially combined with moving to docker-compose format 3 (with depends_on I think we could at least trick docker-compose into starting postgres first).

wardi:
@mattfullerton thanks, merging as you suggest because this is already an improvement.

Issue #3598, opened 2017-06-08
IPackageController Interface don't have before_update and before_create methods


Engerrs:
### CKAN Version
All versions

### Expected behaviour
When user(admin) creates/updates dataset he can hook the dict that he got in method to modify the data that he gets before the actuall create/update.

### Actual behaviour
User can't hook dict before package creates/updates.

Those methods would be very useful for situations, when user got some hidden fields that need to get all data or part of the data from another field in the form. In a result user wont need a custom javascript to deal with such situations and can use CKAN hooks more often.
Issue that can appear after adding those methods related to naming issue, because **IResourceController** got same methods, so when user tries to use both of those methods he got naming conflict.

HilariousGil:
I was looking for this functionality. Thanks @Engerrs 
@wardi I was wondering, is there a specific reason for this not to be merged into any release? or was simply overlooked...

frafra:
Related pull request: https://github.com/ckan/ckan/pull/3599

wardi:
conflicting methods have been renamed also this can be accomplished by customizing the schema or overriding package_create/package_update actions

PR #3599, opened 2017-06-08
[#3598] Adding two new methods for IPackageController

Engerrs:
Added two methods to IPackageController, so that user can hook the dict that he sends before the actual create/update.
Modified their name in order to solve the issue with naming conflicts.

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -152,6 +152,9 @@ def package_create(context, data_dict):
 
     _check_access('package_create', context, data_dict)
 
+    for item in plugins.PluginImplementations(plugins.IPackageController):
+        item.before_package_create(context, data_dict)
+
     if 'api_version' not in context:
         # check_data_dict() is deprecated. If the package_plugin has a
         # check_data_dict() we'll call it, if it doesn't have the method we'll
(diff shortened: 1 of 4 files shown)


wardi:
@Engerrs this change overlaps with functionality already available by implementing validation in an IDatasetForm (even better it works per-dataset-type) or by chaining the package_create/package_update actions themselves. I'm reluctant to add a third way to modify data passed by the user unless there's a really compelling reason.

wardi:
@Engerrs here's the best place to override package_creation/updating the way you want: https://github.com/ckan/ckan/blob/master/ckan/plugins/interfaces.py#L1267

you get the data_dict to make any changes you want, then end your function with a
```python
return ckan.plugins.toolkit.navl_validate(...)
```

Looks like this method isn't in the docs where it should be. Also the docstring is a little scarier than probably necessary. I'll try to fix that.

rossjones:
@wardi I guess it's safe to close this now if it is unnecessary.

gt-novelt:
In case anyone is looking for the line @wardi was referring to, here is the permalink : https://github.com/ckan/ckan/blob/02e65d05815912e7afbe40094f8aa104a2ab64b3/ckan/plugins/interfaces.py#L1284

kmbn:

PR #3600, opened 2017-06-09
CLI for adding user - prompt for email address, now it is required, rather than show error.

davidread:
See user displaying the problem: https://stackoverflow.com/questions/44462382/ckan-create-new-user-issue/44466583#44466583

Instead of this:
```
$ paster sysadmin add seanh
...
Need name and email of the user.
```
i.e. an error. Now with this PR it prompts:
```
$ paster sysadmin add seanh
...
Email address: sean@h.com
...
Added seanj3 as sysadmin
```
or you can specify the email address as a param.

And the same for the 'paster user add' command.

### DIFF
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -91,7 +91,7 @@ def user_add(args):
     or paster user add
     '''
     if len(args) < 1:
-        error('You need to specify the user name.')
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @davidread

PR #3601, opened 2017-06-09
Increase minimum password requirements in line with industry standard

davidread:
Requires: >8 chars, upper & lower & number.

8 chars is mentioned here: https://www.gov.uk/service-manual/design/passwords

The constraint of requiring one of upper/lower/number is typical these days.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -3,10 +3,10 @@
 '''API functions for adding data to CKAN.'''
 
 import logging
-import mimetypes
 import random
 import re
 from socket import error as socket_error
+import string
 
 import paste.deploy.converters
 from sqlalchemy import func
@@ -1023,6 +1023,7 @@ def user_invite(context, data_dict):
     :returns: the newly created user
     :rtype: dictionary
     '''
+    import string
     _check_access('user_invite', context, data_dict)
 
     schema = context.get('schema',
@@ -1037,7 +1038,19 @@ def user_invite(context, data_dict):
         raise NotFound()
 
     name = _get_random_username_from_email(data['email'])
-    password = str(random.SystemRandom().random())
+    # Choose a password. However it will not be used - the invitee will not be
+    # told it - they will need to reset it
+    while True:
+        password = ''.join(random.SystemRandom().choice(
+            string.ascii_lowercase + string.ascii_uppercase + string.digits)
+            for _ in range(12))
+        # Occasionally it won't meet the constraints, so check
+        errors = {}
+        logic.validators.user_password_validator(
+            'password', {'password': password}, errors, None)
+        if not errors:
+            break
+
     data['name'] = name
     data['password'] = password
     data['state'] = ckan.model.State.PENDING
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -583,8 +583,9 @@ def user_password_validator(key, data, errors, context):
         errors[('password',)].append(_('Passwords must be strings'))
     elif value == '':
         pass
-    elif len(value) < 4:
-        errors[('password',)].append(_('Your password must be 4 characters or longer'))
+    elif len(value) < 8:
+        errors[('password',)].append(_('Your password must be 8 characters or '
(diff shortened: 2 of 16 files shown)


adamamyl:
I think the GOV.UK guidance is possibly a little out of sync with that of CESG/NCSC (as they now are), especially looking at the `Technical controls vs complex passwords` section of https://www.ncsc.gov.uk/guidance/password-guidance-simplifying-your-approach ?

davidread:
My thinking is that 8+ chars and requiring 1 each of upper/lower/number is a conservative & widely used minimum. I interpret the NCSC and GDS guidance means don't increase from this further, which is the temptation. If you're using 4 characters then that is too weak by any standards and a no brainer to "Steer users away from choosing predictable passwords". It is annoying that the guidance is so open to interpretation though, and will check with colleagues.

amercader:
@davidread Thanks, the default requirement definitely needs to be updated.

We create a password for invited users [here](https://github.com/ckan/ckan/blob/b4dc4c73d634ddeebcffa2405bdaa44983642a1d/ckan/logic/action/create.py#L1037) that should be updated with whatever new convention, and also the one used in the tests [factory](https://github.com/ckan/ckan/blob/b4dc4c73d634ddeebcffa2405bdaa44983642a1d/ckan/tests/factories.py#L110) (and any other manually created in tests).

I personally find the Upper case requirement annoying as long as we require a certain length and a number but that's a personal preference.

As an aside, the important thing is that extension can change this to suit their needs and right now this is a pain as there is no way to customize the User schema (no `IUserForm`), so you have to [monkey-patch](https://github.com/data-govt-nz/ckanext-security/blob/master/ckanext/security/plugin.py#L16:L19) like in your case or override the whole user_update to pass a custom schema. We need to provide an IUserForm that allows the same customization as eg IDatasetForm.

davidread:
I dropped the requirements for complexity - @adamamyl you're right. Can't do this if you follow the NCSC advice to choose a password made of three words. But the minimum length of 8 stands.

davidread:
Has anyone any idea how to fix the mocha tests? It just says:
```
mocha-phantomjs http://localhost:5000/base/test/index.html
Error loading resource http://localhost:5000/base/test/index.html (1). Details: Connection refused
Failed to load the page. Check the url: http://localhost:5000/base/test/index.html
Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL file:///home/ubuntu/nvm/v0.10.33/lib/node_modules/mocha-phantomjs/lib/mocha-phantomjs.coffee. Domains, protocols and ports must match.
```
I can't see anything to do with passwords in the ckan/public/base/ files either.

TkTech:
@davidread instead, why not use a regex and help text from the ini in the validator? ex:

```
ckan.auth.default_password_regex = "^(?=.*[A-Za-z])(?=.*\d)(?=.*[$@$!%*#?&])[A-Za-z\d$@$!%*#?&]{8,}$"
ckan.auth.default_password_helptext = "Your password must contain at least one letter, one number, and one special character of at least 8 characters."
```

This makes the validator very flexible and can easily be adjusted for three words or other simple policies.

davidread:
Sure, if you think there's a need? @TkTech is this something you'd use, or others?

TkTech:
@davidread yes, it's a change we need to implement ourselves at statscan. Saw you'd already started this while I was looking. We'd backport and use this ASAP.

davidread:
Brill, feel free to add it into this branch then. As long as the default requires 8 chars then I'm happy.

davidread:
Hooray, a rebase fixed the last of the tests somehow. @TkTech feel free to add to this PR before it's merged or simply do a follow-on one.

TkTech:
@davidread not a fire here so I won't get to it immediately. This is a good change and we'll keep the regex-based verification default the same as this so might as well merge it.

davidread:
@amercader are you ok with merging this now?@TkTech is good to follow up with something which is configurable, avoiding those monkey patches you mentioned.

PR #3557, opened 2017-06-16
type override for next datapusher run

wardi:
This adds a "Type Override" field to the data dictionary form that will be read by datapusher when re-uploading.

If datapusher guesses the wrong column types the upload will fail part-way. With this change the user can go to the data dictionary to override the column types so that the next upload may work better.

Currently only text, numeric and timestamp types are supported (the ones used by the default datapusher install)

### DIFF
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -21,6 +21,14 @@ from ckanext.datastore.writer import (
     json_writer,
     xml_writer,
 )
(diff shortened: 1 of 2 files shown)


tino097:
tested it, looks good

amercader:
Thanks for reviewing @tino097. Cherry-picked to 2.7 (and datapusher updated)

PR #3611, opened 2017-06-16
Fix render_datetime for dates before year 1900

### LINKED ISSUE #2228: Render_datetime can't handle dates before year 1900
Due to the python 2.7 feature datetime.strftime won't handle dates before year 1900 (http://bugs.python.org/issue1777412)

https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L947

klikstermkd:
Fixes #2228 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1176,6 +1176,20 @@ def render_datetime(datetime_, date_format=None, with_hours=False):
 
     # if date_format was supplied we use it
     if date_format:
+
+        # See http://bugs.python.org/issue1777412
+        if datetime_.year < 1900:
+            date_format = date_format.replace('%y', str(datetime_.year)[-2:])
+            date_format = date_format.replace('%Y', str(datetime_.year))
+
+            # "2017" can be any year, since it will be replaced with the one
+            # provided in "date_format"
+            datetime_ = datetime.datetime(2017, datetime_.month, datetime_.day,
+                                          datetime_.hour, datetime_.minute,
+                                          datetime_.second)
+
+            return datetime_.strftime(date_format)
+
         return datetime_.strftime(date_format)
     # the localised date
     return formatters.localised_nice_date(datetime_, show_date=True,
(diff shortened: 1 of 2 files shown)


ckan/lib/helpers.py
```diff
@@ -1176,6 +1176,20 @@ def render_datetime(datetime_, date_format=None, with_hours=False):
 
     # if date_format was supplied we use it
     if date_format:
+
+        # See http://bugs.python.org/issue1777412
+        if datetime_.year < 1900:
+            date_format = date_format.replace('%y', str(datetime_.year)[-2:])
+            date_format = date_format.replace('%Y', str(datetime_.year))
```
wardi: `'%%y'` won't be handled properly here

klikstermkd: After a research, I found out that there isn't an easy way to fix this. But, we can use a third-party library that patches `strftime` to use year before 1900. Here's one which I tested it and it works https://pypi.python.org/pypi/virtualtime

klikstermkd: @wardi What do you propose, how should we proceed with this?

wardi: how about a simple regexp that only replaces '%y' or '%Y' when not preceded by a '%'?

klikstermkd: @wardi Here's a regexp `r'(?<!%)%y'` that uses negative lookbehind, but this will work correctly only for `%y` and `%%y`, and not for `%%%y` and so on.

wardi: how about
```python
re.sub('(?<!%)(%%)*%y', r'\g<1>1890', '%%y')
```

klikstermkd: It works!

ckan/lib/helpers.py
```diff
@@ -1176,6 +1176,20 @@ def render_datetime(datetime_, date_format=None, with_hours=False):
...
+
+            # "2017" can be any year, since it will be replaced with the one
+            # provided in "date_format"
+            datetime_ = datetime.datetime(2017, datetime_.month, datetime_.day,
```
wardi: Use 2016 instead for Feb 29 support.

klikstermkd:
@wardi Ready for review.

PR #3613, opened 2017-06-17
[#2944] Update API docs for import datasets example

### LINKED ISSUE #2944: CKAN API documentation

release-v2.4-final
http://ec2-52-87-202-213.compute-1.amazonaws.com:5000
## Story

I've tried today to create a simple API example for an institution, that only contained creating a dataset and creating a resource for that dataset. However, I've gone through two problems and it took me a lot more time than expected.

First, the [example](http://docs.ckan.org/en/latest/api/index.html#example-importing-datasets-with-the-ckan-api) for creating a new dataset does not work because it's missing the `owner_org` [key](https://github.com/govro/ckan-api-upload-example/blob/master/script.py#L18). I think we should, at least, patch the documentation till https://github.com/ckan/ckan/issues/2789 is ready.

Second, there is the problem with the `URL` field for a resource https://github.com/ckan/ckan/pull/2844. I think we should also patch the [documentation](http://docs.ckan.org/en/latest/maintaining/filestore.html#filestore-api).

Is it okay if I go and change the docs and add to-do items within the corresponding issues?

klikstermkd:
Fixes #2944 

Update API docs for [this example](http://docs.ckan.org/en/latest/api/index.html#example-importing-datasets-with-the-ckan-api) to include the missing `owner_org` field which is required.

### Features:

- [ ] includes API changes

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -184,6 +184,7 @@ to import datasets into CKAN.
     dataset_dict = {
         'name': 'my_dataset_name',
         'notes': 'A long description of my dataset',
+        'owner_org': 'org_id_or_name'
     }
 
     # Use the json module to dump the dictionary to a string for posting.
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @klikstermkd!

Issue #3615, opened 2017-06-18
datastore_search action gives 0 results with query search


Engerrs:
### CKAN Version
2.6 and lower

###  Expected behaviour
When uploading CSV file and trying to do query search with datastore_search for example: http://demo.ckan.org/api/action/datastore_search?resource_id=335bda41-93ea-4f57-b274-cb4347425e73&q=mary, it should give results, where 'mary' is set in text or is a single word. Searching by small text "Native of NSW" should give results, where this text appears.

### Actual behaviour
0 results return when trying to search by single words that ends with 's', 'y' latters, for example: "Mary" and "Francis".

###  Reproduce the issue 
1.Set Postgres version 9.4.
2.Make custom CSV with some names and texts. Some name should consists 's' or 'y' endings.
3.Setup pure CKAN with datastore and datapusher.
4. Create dataset and resource with CSV you've created.
5. Enter to the resource summary page, view should be created.
6. Try to search by name that ends with 's' or 'y'. No result returns.
7. Try to search by some small text. No result returns.

Engerrs:
This issue can be closed, as I found out that there is an config option "ckan.datastore.default_fts_lang = english" that can solve this one if we put "simple".
But still the actual issue with 'english' still appears on Postgres 9.4.7, 9.5.4, while 9.5.7 working with 'english' fine.

PR #3616, opened 2017-06-18
[#3615] Fix for datastore_search action

Engerrs:
I was trying to reproduce this bug with pure SQL query right in database on 9.5 Postgres version, but it seems to work as it should be.
On 9.4 version it's broken.
After some investigation I found the problem in **_build_query_and_rank_statements method**,
as far a I understand from [this](https://www.postgresql.org/docs/9.4/static/textsearch-controls.html) page at postgres doc, 'english' param is used to split the words by small parts for text search, but it seems that we don't actually need this for our plain text search, as without it, the query returns all results. I also assume that this also be problem with Postgres version, as at 9.5 it works fine with the 'english' param.

This fix is only for 2.6 and lower version, as 2.7 plugin file was slightly changed.

### DIFF
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -506,10 +506,10 @@ class DatastorePlugin(p.SingletonPlugin):
         lang_literal = literal_string(lang)
         query_literal = literal_string(query)
         if plain:
-            statement = u"plainto_tsquery({lang_literal}, {query_literal}) {query_alias}"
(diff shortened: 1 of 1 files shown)


amercader:
@Engerrs Can you provide more context as to what was the original issue?
Also we definitely need a new test that demonstrates the problem and that gets fixed by the patch. Finally we don't merge PRs into already released versions, all fixes need to be cherry-picked from master.

Feel free to resubmit a pull request with more context that targets master. Thanks!

PR #3627, opened 2017-06-21
[#3606] Allow plugins to react to forks of the CKAN process

### LINKED ISSUE #3606: Problems in background workers with non-core database relations
@amercader has [reported](https://gist.github.com/amercader/7d7d3ffa89af834205974cad0d0a57d0) problems with the new background job system where a job succeeds on the first run but then fails on the following runs. This is similar to issue #3243 (fixed by PR #3247). Also of interest may be the issue #3332 (with PR #3337) as well as issue #3365 (with PR #3377).

@amercader has suggested that the problem might be due to DataStore keeping its own set of database connections (one for reading, one for writing). Those are currently not taken into account when the database connection is disposed before a worker is forked and re-established afterwards.

If that is indeed the case then the DataStore DB connections (and all other similar connections established by extensions) need to be taken care of. One way of doing that would be via an interface that provides access to worker events (e.g. `before_worker_fork` and `after_worker_fork`).

The first step, however, should be to create a reproducibly failing test case from the example above.

torfsen:
This is a WIP to fix #3606.

As discussed in #3606, the PR introduces a new interface (`IForkObserver`) to allow extensions to react to forks of the CKAN process. This is then used to prevent a sharing of DataStore's SQLAlchemy engines between parent and child processes, similar to what we're already doing for the core's engines.

In its current state, this PR fixes the problem of #3606 for me: if I manually enable the `datastore` plugin in `test-core.ini` the tests still succeed (without the PR some tests in `ckan/tests/lib/test_jobs.py` fail).

However, some parts of the PR are less than elegant, so I'd love to get some feedback and suggestions for improvements:

- In `ckan/lib/jobs.py` I monkey-patch `os.fork` to make sure that the implementations of `IForkObserver` are correctly notified. The monkey-patch is only temporarily while calling the `rq` method that performs the fork. An alternative would be to re-implement the `rq` method in question with the `IForkObserver` notifications. That would avoid the monkey-patch but would make updating `rq` more difficult since we would be messing even more with its internals than we already are.

- AFAIK only plugins can implement interfaces. This means that the `ckanext.datastore.plugin.Datastore` implements `IForkObserver` and then passes these events on to the active backend. Hence `ckanext.datastore.backend.DatastoreBackend` gets two new methods `before_fork` and `after_fork`, and `ckanext.datastore.backend.postgres.DatastorePostgresqlBackend` gets a new `before_fork` method. This is a lot of redirection and certainly not ideal. An slightly better alternative might be to keep these methods out of `DatastoreBackend` and then do some type checking in `Datastore.before_fork` to see whether the active backend is PostgreSQL (and if so, call its `before_fork` method).

In addition to these issues the PR is missing proper documentation. In addition, we haven't so far succeeded in creating a reproducibly failing test case for the original issue that doesn't rely on adding `datastore` to the list of plugins in `test-core.ini` (somehow, temporarily loading the plugin during the test case does not trigger the problem).

### DIFF
--- a/ckan/lib/jobs.py
+++ b/ckan/lib/jobs.py
@@ -19,7 +19,6 @@ prefixed names. Use the functions ``add_queue_name_prefix`` and
 '''
 
 import logging
-import os
 
 import rq
 from rq.connections import push_connection
@@ -206,23 +205,6 @@ def test_job(*args):
     print(args)
 
 
-_os_fork = os.fork
-
-
-def _fork():
-    u'''Variant of os.fork that notifies IForkObserver implementations.
-
-    Works the same as ``os.fork`` does, but notifies implementations of
-    :py:class:`ckan.plugins.IForkObserver` according to its protocol.
-    '''
-    for plugin in plugins.PluginImplementations(plugins.IForkObserver):
-        plugin.before_fork()
-    result = _os_fork()
-    for plugin in plugins.PluginImplementations(plugins.IForkObserver):
-        plugin.after_fork(result)
-    return result
-
-
 class Worker(rq.Worker):
     u'''
     CKAN-specific worker.
@@ -278,13 +260,9 @@ class Worker(rq.Worker):
         queue = remove_queue_name_prefix(job.origin)
         log.info(u'Worker {} starts job {} from queue "{}"'.format(
                  self.key, job.id, queue))
-        try:
-            # Monkey patch os.fork to make sure IForkObserver
-            # notifications are sent.
-            os.fork = _fork
-            result = super(Worker, self).execute_job(job, *args, **kwargs)
-        finally:
-            os.fork = _os_fork
+        for plugin in plugins.PluginImplementations(plugins.IForkObserver):
+            plugin.before_fork()
+        result = super(Worker, self).execute_job(job, *args, **kwargs)
         log.info(u'Worker {} has finished job {} from queue "{}"'.format(
                  self.key, job.id, queue))
 
--- a/ckan/lib/jobs.py
+++ b/ckan/lib/jobs.py
@@ -19,6 +19,7 @@ prefixed names. Use the functions ``add_queue_name_prefix`` and
 '''
 
 import logging
+import os
 
 import rq
 from rq.connections import push_connection
@@ -30,6 +31,7 @@ from ckan.lib.redis import connect_to_redis
 from ckan.common import config
 from ckan.config.environment import load_environment
 from ckan.model import meta
+import ckan.plugins as plugins
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


amercader:
@torfsen thanks for your quick work on this.

Regarding the two main points:

* Why do we need to go down to the level of `os.fork()`? To solve the main engine disposal issues on #3247 you added the environment refresh code in the [`main_work_horse`](https://github.com/torfsen/ckan/blob/df570b9d74bbe9ffd44bbb4846a19929b0aad6a7/ckan/lib/jobs.py#L254) method, which as you mention on the comment gets called shortly after forking. Can we not add the plugins check at this point so it has the same effect on extensions? eg:

    ```python

    def main_work_horse(self, job, queue):
        # This method is called in a worker's work horse process right
        # after forking.
        load_environment(config[u'global_conf'], config)
        for plugin in plugins.PluginImplementations(plugins.IForkObserver):
           plugin.after_fork(self.pid)
        return super(Worker, self).main_work_horse(job, queue)
    ```
   It could well be that I'm missing something but this seems cleaner, even if we don't have a `before_fork` method (maybe we can implement it similarly from another of the `Worker` methods).

* I agree is not ideal. Perhaps something like your last suggestion will be simpler:

    ```python

     class DatastorePlugin(p.SingletonPlugin):

          p.implements(IForkObserver)

         def after_fork(self, pid):
             if hasattr(self.backend, 'after_fork'):
                 self.backend.after_fork(pid)

     # ...

     class DatastorePostgresqlBackend(DatastoreBackend):

          def after_fork(self, pid):
                # Whatever
    ```

   On the other hand, having it explicitly added to `DatastoreBackend` kind of documents the methods (it should say these are optional). I don't have a strong opinion either way.

torfsen:
Consensus in the dev meeting today:
- The `after_fork` is currently not used but adds a lot of complexity. Remove it. We can add it later on if it turns out that it would be useful.
- Instead of polluting the `DatastoreBackend` interface with the events from `IForkObserver` use the duck-typing approach suggested by @amercader. A possible future improvement would be to make Datastore backends true plugins which can implement interfaces themselves.

torfsen:
I've implemented the changes discussed above (remove `after_fork` and propagate `before_fork` via duck-typing).

In addition I've added a test case which reproduces the problem: the test [fails](https://circleci.com/gh/ckan/ckan/2777) without the fix (dragging a lot of other test cases down with it by messing up the database connection) and [succeeds](https://circleci.com/gh/ckan/ckan/2779) when the fix is in place.

Note that GitHub displays the commits in their chronological order, so it appears that the last commit of the PR fails. However, while f46c80f is technically the youngest commit in this branch it is logically the first one. This is because I re-ordered the commits (`git rebase -i`) to put the test case to the front so that I could verify that it fails on CircleCI before pushing the fix. I wasn't aware that this would turn out to be this confusing in the GitHub UI:

```
$ git log --pretty=oneline -n 4
4d39e1e052e8047aa530f3e52596b9cbc2c748ca [#3606] Use duck-typing to call before_fork in the PostgreSQL backend
05a28a598b925bc0c6bf60736e9aae71387bb943 [#3606] Remove IForkObserver.after_fork
2e40ceb436a5164653777d1801c9896636c3129b [#3606] Allow plugins to react to forks of the CKAN process
f46c80fa34e29f3873c7110f9d594c37e0e31f6c [#3606] Add test case for Datastore backend jobs problem
```

This PR is now ready for review.

wardi:
@torfsen that's pretty funny.  too bad github doesn't handle rewritten history well :smile:   Having that test to demonstrate the issue is great.

amercader:
Thanks @torfsen, this is now on 2.7

PR #3631, opened 2017-06-23
Changed required permission from 'update' to 'manage_group'

### LINKED ISSUE #3308: Problem adding datasets to a group on package_create, the user has 'Editor' capacity
Found in 2.2.2 and later

I manage a customized CKAN for a client. The create dataset page is changed in a way it is possible to add all metadata to a dataset on 'package_create'. Also it should be possible to add the dataset direktly to  groups. The user has the capacity 'Editor' on the group.

The auth function 'package_create' always does the `check2 = _check_group_auth(context,data_dict)`, which is a different approach than in 'package_update' auth function.
That leads to using the call to `authz.has_user_permission_for_group_or_org(group.id, user, 'update')`.
Later this leads to a comparison of permission '**update**' with the permissions of 'Editor' role  ('editor', ['read', 'delete_dataset', 'create_dataset', 'update_dataset', 'manage_group']). 
`if 'admin' in perms or permission in perms:
            return True`
In my opinion this can never be true and thus is bug.

Could you please check this?

Regards,
Daniel

iewil:
Fixes #3308 

### Proposed fixes:
Changed the required permission for has_user_permission_for_group_or_org from "update" to "manage_group". There is no such permission as "update" in ROLE_PERMISSIONS.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/auth/create.py
+++ b/ckan/logic/auth/create.py
@@ -195,7 +195,7 @@ def _check_group_auth(context, data_dict):
         groups = groups - set(pkg_groups)
 
     for group in groups:
-        if not authz.has_user_permission_for_group_or_org(group.id, user, 'update'):
+        if not authz.has_user_permission_for_group_or_org(group.id, user, 'manage_group'):
             return False
 
     return True


amercader:
Thanks @iewil !

PR #3642, opened 2017-06-26
3641 setuptools instructions

### LINKED ISSUE #3641: No such file or directory: '/usr/lib/ckan/default/src/ckan/requirement-setuptools.txt' during installation from source
Latest (setting up development environment)

When following the CKAN [installation from source](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) instructions, I should be able to complete the development setup on Ubuntu 16.04. This implies that all requisite steps should be fully documented.

I get the following error when creating the virtualenv:

```
$ virtualenv --no-site-packages /usr/lib/ckan/default
Using base prefix '/home/brylie/anaconda3'
New python executable in /usr/lib/ckan/default/bin/python
/usr/lib/ckan/default/bin/python: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
ERROR: The executable /usr/lib/ckan/default/bin/python is not functioning
ERROR: It thinks sys.prefix is '/home/brylie' (should be '/usr/lib/ckan/default')
ERROR: virtualenv is not compatible with this system or executable
```

I am getting an error on the step **2b: Install the recommended version of ‘setuptools’:**

```
 pip install -r /usr/lib/ckan/default/src/ckan/requirement-setuptools.txt

Could not open requirements file: [Errno 2] No such file or directory: '/usr/lib/ckan/default/src/ckan/requirement-setuptools.txt'

```

Follow the steps on [Installing CKAN from source](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html)

torfsen:
Fixes #3641.

Instead of pointing to a file that is only installed later, the docs now simply tell the user to update `setuptools` to the latest version.

I've kept `requirement-setuptools.txt` because it is used in the CI scripts where reproducibility is important. However, I've taken the chance and removed a duplicate hard-coded `setuptools` version from `setup.py` which now reads it from `requirement-setuptools.txt` instead.


torfsen:
I've updated the PR as discussed in the dev meeting today: the setuptools version for the docs is now read dynamically from `requirement-setuptools.txt`.

This PR is now ready for review.

amercader:
Thanks @torfsen!

davidread:
Thanks for fixing this up @torfsen and there are some neat improvements in here.

PR #3643, opened 2017-06-26
Use name of dataset type in search page

mattfullerton:
The use of the word "dataset" is basically hardcoded in a number of places even though the rest of the CKAN logic is fine with handling non "dataset" packages. I'm not sure if this PR is desired because maybe such customizations belong in the extensions where such types are created, but I've had a stab at handling any other dataset type generically. There are a few rough edges regarding translatability, which was maybe the reason never to do this in the first place? Or was it an oversight?


ckan/templates/snippets/search_result_text.html
```diff
@@ -28,18 +28,25 @@
   {% set text_query_none = _('No organizations found for "{query}"') %}
   {% set text_no_query = ungettext('{number} organization found', '{number} organizations found', count) %}
   {% set text_no_query_none = _('No organizations found') %}
+
+{% else %}
+  {% set text_query = ungettext('{number} {type} found for "{query}"', '{number} {type}s found for "{query}"', count) %}
```
wardi: let's substitute the type before passing the string to ungettext. That lets extensions define their own plurals and translations of the types. (the list of types will be static for a given site so static translations are fine)

ckan/templates/package/search.html
```diff
@@ -4,7 +4,8 @@
 {% block subtitle %}{{ _("Datasets") }}{% endblock %}
 
 {% block breadcrumb_content %}
-  <li class="active">{{ h.nav_link(_('Datasets'), controller='package', action='search', highlight_actions = 'new index') }}</li>
+  {% set type_text = dataset_type.title() %}
+  <li class="active"><a href="{{ dataset_type }}">{{ _(type_text) }}</a></li>
```
wardi: should be possible to generate these urls with nav_link or url_for. They are registered  as `'%s_search' % package_type` https://github.com/ckan/ckan/blob/master/ckan/lib/plugins.py#L104

mattfullerton: Ah! _that's_ how its supposed to work ... OK, will rework

wardi:
@mattfullerton would you be able to resolve conflicts and make the change I suggested?

mattfullerton:
@wardi see my comment: It doesn't work for me as I'm not using IDatasetForm but ckanext-scheming

wardi:
@mattfullerton but, ckanext-scheming uses IDatasetForm to create new dataset types

mattfullerton:
@wardi I clarified your suggestion with @amercader and have it working now in all places where URL creation happens. I've also merged in latest CKAN master.

mattfullerton:
@wardi Also happy to squash the commits if you give a green light to the final version

mattfullerton:
@wardi Are the failing circleci tests coming from the use of the ~ concatenation operator? That converts to unicode:
https://github.com/pallets/jinja/blob/master/jinja2/nodes.py#L757

But there are plenty of places that do it in ckan:

```
$ pwd
... ckan/ckan/templates
ack "~"
```

mattfullerton:
@wardi Removing the tildes allowed the tests to pass. Should I squash or can you take it as is?

ckan/templates/snippets/add_dataset.html
```diff
@@ -1,7 +1,9 @@
 {# Adds 'Add Dataset' button #}
 
+{% set dataset_type = dataset_type if dataset_type else 'dataset' %}
+
 {% if group %}
     {% link_for _('Add Dataset'), controller='package', action='new', group=group, class_='btn btn-primary', icon='plus-square' %}
 {% else %}
-    {% link_for _('Add Dataset'), controller='package', action='new', class_="btn btn-primary", icon="plus-square" %}
+    {% link_for _('Add ' + dataset_type.title()), controller='package', action='new', named_route=dataset_type + '_new', class_='btn btn-primary', icon='plus-square' %}
```
wardi: @amercader does this cause any problem with translations? Now collecting the strings won't pick up "Add Dataset" anymore, will it get removed from the template and po files? (That would be bad)

PR #3646, opened 2017-06-27
Upgrade SQLAlchemy to 1.1

amercader:
After diving into this for quite a while I think this is ready for review. I don't think it adds more uncertainty to VDM's behaviour, and any issues we can pick up before the next release.

As usual when involving vdm, these are really small patches with huge commit messages for background but all commits should be straightforward to follow.

Most important one is 26dd7ad. The Tl;DR is that if in a single transaction the Session has recorded a change for the creation of an object and for the update of the same object, we will just record an activity for the creation of the object. @TkTech as you have worked with activities I'd like to make sure that you are happy with this (26dd7ad).

@wardi a85b684 is a a change on the trigger functions creation that you worked on recently.

And of course, VDM must go :hocho:

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -610,7 +610,7 @@ def member_create(context, data_dict=None):
                               table_id=obj.id,
                               group_id=group.id,
                               state='active')
-
+        member.group = group
     member.capacity = capacity
 
     model.Session.add(member)
--- a/ckan/migration/versions/016_uuids_everywhere.py
+++ b/ckan/migration/versions/016_uuids_everywhere.py
@@ -170,18 +170,25 @@ def create_uuids(migrate_engine, primary_table_name, revision_table_name):
         q = revision_table.update().values(id=revision_table.c.continuity_id)
         migrate_engine.execute(q)
 
+
 def drop_sequencies(migrate_engine):
 
-    sequencies = ['package_extra', 'package',
-                  'package_resource',
-                  'package_tag',
-                  'revision', 'tag']
+    sequences = ['package_extra', 'package_extra_revision', 'package',
+                 'package_resource', 'package_resource_revision',
+                 'package_revision', ' package_tag', 'package_tag_revision',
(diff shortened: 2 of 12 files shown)


TkTech:
@amercader more than happy with it, I think someone has actually implemented similar behaviour in our fork already since our clients complained.

ckan/logic/action/create.py
```diff
@@ -610,7 +610,7 @@ def member_create(context, data_dict=None):
                               table_id=obj.id,
                               group_id=group.id,
                               state='active')
-
+        member.group = group
```
wardi: is this an unrelated change?

ckan/migration/versions/016_uuids_everywhere.py
```diff
@@ -166,19 +172,18 @@ def create_uuids(migrate_engine, primary_table_name, revision_table_name):
 
 def drop_sequencies(migrate_engine):
 
-    sequencies = ['package_extra', 'package_extra_revision', 'package',
-                  'package_resource', 'package_resource_revision',
-                  'package_revision',' package_tag', 'package_tag_revision',
+    sequencies = ['package_extra', 'package',
+                  'package_resource',
+                  'package_tag',
```
wardi: looks like a change in behaviour. we'd still want to drop these sequences if they were created in an earlier migration, right?

ckan/logic/action/create.py
amercader: No, with the 1.1 upgrade not explicitly assigning the group to the new member object led to the relation between the two not being set at the database level, which caused loads of [test failures](https://circleci.com/gh/ckan/ckan/2679?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) due to Authorization errors

ckan/migration/versions/016_uuids_everywhere.py
amercader: You are right, fixed in dafa45b

PR #3650, opened 2017-06-29
Fixes #2806

### LINKED ISSUE #2806: Allow webpageview on external URLs
Right now, webpageview only works for files with format HTM/HTML.

https://github.com/ckan/ckan/blob/master/ckanext/webpageview/plugin.py#L27-L31

Why not just allow webpageview for an external URL?

This could be a great generic way to handle external resources without having to create a special purpose preview handler (see #1536)

For example, [Google Sheets](http://data.beta.nyc/dataset/hack4heat-2015-related-datasets/resource/bb4924bf-26cd-4b60-8006-e15038205b95), Docs, [Slides](http://data.beta.nyc/dataset/hack4heat-2015-related-datasets/resource/5cd5fd78-ade7-4223-b402-82408e814805), etc. and anything that can live in an iframe.

e.g. http://data.beta.nyc/dataset/hack4heat-2015-related-datasets

We actually allowed this on this CKAN instance.

mattfullerton:
Fixes #2806


amercader:
I think the Fullscreen view button makes sense. I found confusing that it opened in a new tab but maybe that's just me.

The main issue I see is that most views are not prepared to be displayed on their own in a page, specially there's padding lacking. I guess that this can be solved on a per-view basis afterwards.


untitled

PR #3651, opened 2017-06-29
[Dockerfile] Ckan dockerfile refactoring

SergK:
* Define DEBIAN_FRONTEND noninteractive as ENV var, since dist-upgrade
  can bring some terminal warnings
* Add postgresql-client, since we are using it during postgres check
  readiness
* Add and run ckan container under ckan user
* Decrease number of docker layers, by grouping all RUN stuff in single
  RUN command
* Update ckan env variables:
  + ENV CKAN_HOME /usr/lib/ckan
  + ENV CKAN_VENV $CKAN_HOME/default

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,49 +3,49 @@
 FROM debian:jessie
 MAINTAINER Open Knowledge
 
-ENV CKAN_HOME /usr/lib/ckan/default
+ENV CKAN_HOME /usr/lib/ckan
+ENV CKAN_VENV $CKAN_HOME/default
 ENV CKAN_CONFIG /etc/ckan/default
 ENV CKAN_STORAGE_PATH /var/lib/ckan
 ENV CKAN_SITE_URL http://localhost:5000
 
+ENV DEBIAN_FRONTEND noninteractive
+
 # Install required packages
-RUN apt-get -q -y update && apt-get -q -y upgrade && DEBIAN_FRONTEND=noninteractive apt-get -q -y install \
-		python-dev \
-        python-pip \
-        python-virtualenv \
-        libpq-dev \
-        git-core \
-	&& apt-get -q clean
+RUN apt-get -q -y update && apt-get -q -y upgrade && apt-get -q -y install \
+    python-dev \
+    python-pip \
+    python-virtualenv \
+    libpq-dev \
+    git-core \
+    postgresql-client \
+    && apt-get -q clean && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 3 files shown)


SergK:
Hi @mattfullerton @wardi Can I kindly ask you for review. A little bit update on the latest changes. Thank you

Dockerfile
```diff
@@ -3,49 +3,49 @@
...
+    python-dev \
+    python-pip \
+    python-virtualenv \
+    libpq-dev \
+    git-core \
+    postgresql-client \
+    && apt-get -q clean && rm -rf /var/lib/apt/lists/*
+
+# add ckan user which runs all the ckan related stuff
+RUN useradd -r -u 900 -m -c "ckan account" -d $CKAN_HOME -s /bin/false ckan
```
florianm: This will only work if run after L28, where $CKAN_HOME is created.

SergK: Why? **CKAN_HOME=/usr/lib/ckan** 
.. and **_/var/lib_** - exists, so there is no need in explicit creation, system will do it itself.

florianm: Yes, the variable $CKAN_HOME exists, but the useradd command has trouble creating the dir $CKAN_HOME by itself (contradicting man useradd).

```
Step 9/18 : RUN useradd -r -u 900 -m -c "ckan account" -d $CKAN_HOME -s /bin/false ckan
 ---> Running in 4ad1ca329589
useradd: cannot create directory /usr/lib/ckan/default
ERROR: Service 'ckan' failed to build: The command '/bin/sh -c useradd -r -u 900 -m -c "ckan account" -d $CKAN_HOME -s /bin/false ckan' returned a non-zero code: 12
```

When I move the useradd command below the mkdir step, all is well.

https://askubuntu.com/a/374878 says that adduser is preferred over useradd.

SergK: I've checked your patch and in your case the issue exists because you have another paths:

```
ENV CKAN_HOME /usr/lib/ckan/default
ENV CKAN_VENV $CKAN_HOME/default
```

that's why it can not add user since /usr/lib/ckan/ does not exist

florianm: my bad, thanks for the heads-up! 
Works on my end: https://github.com/parksandwildlife/ckan/blob/3649-docker-upgrade/Dockerfile
Additionally, SITE_URL comes from docker-compose.yml reading .env

SergK:
Closed, since the most of the stuff was merged within #3692

PR #3652, opened 2017-06-29
Fix 'owner_org' shown as optional in API docs for package_create

### LINKED ISSUE #3647: In API docs "package_create" lists "owner_org" as optional
In API docs, the action `package_create` lists the parameter `owner_org` as optional which is not correct.

The parameter `owner_org` can be optional only if `ckan.auth.create_unowned_dataset` is set to `true` in the config.

I suggest to remove the "(optional)" text and add to the description of the parameter that it can be optional only if `ckan.auth.create_unowned_dataset` is set to `true` in the config.

klikstermkd:
Fixes #3647 

### Features:

- [ ] includes API changes


klikstermkd:
Ready for review.

PR #3657, opened 2017-06-30
Fix broken default logo image

### LINKED ISSUE #3656: Default logo image not properly saved
2.7 and 2.8.0a

By default, CKAN uses a default logo image. This can be changed through the UI in the sysadmin config page or in configuration settings.

If you go to the sysadmin config page, and if you don't change the logo image and just click Update, the logo image will break. The reason is because its been given wrong path to the image.

Go to sysadmin config page. If you previously changed the default logo image, click the Reset button to bring the default settings. After that click on Update and the logo image will break.

I isolated the issue and will provide a fix.

klikstermkd:
Fixes #3656 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -1272,8 +1272,10 @@ def config_option_update(context, data_dict):
         raise ValidationError(errors)
 
     image_in_ckan = False
+    public_path = config.get('ckan.static_files', 'public')
     ckan_images_path = os.path.abspath(os.path.join(os.path.dirname(__file__),
-                                       '..', '..', 'public', 'base', 'images'))
+                                       '..', '..', public_path, 'base',
+                                                    'images'))
 
     for key, value in data.iteritems():
 
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -1271,23 +1271,12 @@ def config_option_update(context, data_dict):
         model.Session.rollback()
         raise ValidationError(errors)
 
-    image_in_ckan = False
-    public_path = config.get('ckan.static_files', 'public')
-    ckan_images_path = os.path.abspath(os.path.join(os.path.dirname(__file__),
-                                       '..', '..', public_path, 'base',
-                                                    'images'))
-
     for key, value in data.iteritems():
 
         # Set full Logo url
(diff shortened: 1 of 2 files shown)


klikstermkd:
Ready for review.

ckan/logic/action/update.py
```diff
@@ -1270,11 +1271,27 @@ def config_option_update(context, data_dict):
...
     for key, value in data.iteritems():
 
         # Set full Logo url
-        if key =='ckan.site_logo' and value and not value.startswith('http'):
-            value = h.url_for_static('uploads/admin/{0}'.format(value))
+        if key == 'ckan.site_logo' and value and not value.startswith('http'):
+            for f in os.listdir(ckan_images_path):
+                if f == value:
+                    image_in_ckan = True
+                    break
```
wardi: os.isfile can be used instead of a for loop here

klikstermkd:
@wardi Now it should be good.

klikstermkd:
@wardi Can you check this again? :)

amercader:
I think the underlying problem is that the admin form is receiving the correct value (`/base/images/ckan-logo.png`), even if the widget just displays the file name:

```html
<input id="field-image-url" type="url" name="ckan.site_logo" value="/base/images/ckan-logo.png" placeholder="http://example.com/my-image.jpg" class="form-control" readonly="">
```

But sending the wrong value to the controllers, it will only send the file name (`ckan-logo.png`). Rather than adapt the action to this, it would be best to ensure that the form is sending data unchanged.

klikstermkd:
That still won't resolve the issue because the controller doesn't know whether the default image is uploaded from a user or available already from CKAN's local images directory.

amercader:
True, but then it's just a matter of covering the local image case in the `config_option_update()` check:

```python
if key =='ckan.site_logo' and value and not (value.startswith('http') or value.startswith('/')):
```

klikstermkd:
Uploaded images from user and CKAN's images from local folder will pass the condition.

amercader:
Sorry, I meant: 

```python
if key =='ckan.site_logo' and value and not value.startswith('http') and not value.startswith('/'):
```

klikstermkd:
@amercader Can you check now?

PR #3666, opened 2017-07-04
Fix solr multicore not persisting restarts

davidread:
When making SOLR run multicore, I found that using the HTTP request to SOLR to tell it to create a SOLR core worked only until you restart Jetty/SOLR. I couldn't get the 'persist' parameter to work, so had to fall back to editing the solr.xml file.

### DIFF
--- a/doc/contributing/test.rst
+++ b/doc/contributing/test.rst
@@ -71,12 +71,13 @@ Configure Solr Multi-core
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The tests assume that Solr is configured 'multi-core', whereas the default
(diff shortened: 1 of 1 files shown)


tino097:
thumbs up!

amercader:
Just out of curiosity, what Solr version where you using when testing this @davidread?

davidread:
@amercader it's Postgres 9.3.16 (as comes with Ubuntu 14.04)

amercader:
@davidread thanks, was asking about Solr (as this issue relates to Solr) :)

davidread:
:-/

SOLR 3.6.2 (i.e. solr-jetty package default for Ubuntu 14.04)

PR #3668, opened 2017-07-04
Validate image URL in resource form

### LINKED ISSUE #3660: Resource URL is not validated on create/update
All versions.

On resource create/update, if a URL is provided it should be validated.

On resource create/update, if a URL is provided it's not validated.

Either through the API or the web form, try to create or update a resource and choose a URL as a resource and type some text which is not a valid URL. It will pass for any typed text.

klikstermkd:
Fixes #3660 

The image URL field is validated only through the form and not through the API, since as discussed on the meeting call if it's validated on the backend it can cause many issues.

### Features:

- [ ] includes API changes


klikstermkd:
Ready for review.

PR #3674, opened 2017-07-06
Radio helper

### LINKED ISSUE #3667: radio helper missing
CKAN 2.6.2

The radio button helper is removed from helpers.py and it's still used in revisions_table.html

As I am informed history revision will be removed, but we should be taken into consideration that some extensions may still use it. The radio button helper should be displayed and not produce an error.

JGulic:
Fixes #3667 

### Proposed fixes:
Radio helper was removed but it's still used in revisions_tables template. By adding the radio button helper we are avoiding error in revisions_table template

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2342,6 +2342,15 @@ def mail_to(email_address, name):
     return html
 
 
+@core_helper
+def radio(selected, id, checked):
+    if checked:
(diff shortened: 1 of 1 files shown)


tino097:
@amercader resolved

amercader:
Thanks @JGulic and @tino097

PR #3686, opened 2017-07-12
OpenID column removed from userModel and code

smotornyuk:
Just removed this field(migration included)


ckan/config/routing.py
```diff
@@ -358,8 +358,6 @@ def make_map():
     map.redirect('/user/', '/user')
     with SubMapper(map, controller='user') as m:
         m.connect('/user/edit', action='edit')
-        # Note: openid users have slashes in their ids, so need the wildcard
-        # in the route.
```
amercader: Can we remove the wildcards in the routes that this comment was referring to? eg `'/user/followers/{id:.*}` -> `'/user/followers/{id}`. Other routes don't use them anyway, and a year from now we will we wondering why they are there.

smotornyuk:
Good idea. I removed them

tino097:
There were no problems, its good to go in master :)

PR #3695, opened 2017-07-14
[#3694] Fix datastore setup that was being scuppered by log statement…

### LINKED ISSUE #3694: datastore set-up error - logging getting in the way
master

Setting up the datastore permissions hits an error (which can be ignored, but still)
```
$ paster datastore set-permissions -c test-core.ini | sudo -u postgres psql
You are now connected to database "datastore_test" as user "postgres".
ERROR:  syntax error at or near "2017"
LINE 1: 2017-07-14 14:35:01,978 INFO  [ckan.lib.fanstatic_resources]...
        ^
```
This is because of some stray logging appearing:
```
$ paster datastore set-permissions -c test-core.ini
2017-07-14 14:37:10,623 INFO  [ckan.lib.fanstatic_resources] Base path /home/dread/ckan-master/ckan/ckan/public/base
/*
This script configures the permissions for the datastore.
...
```

davidread:
Also, an unrelated log spacing improvement.

Fixes #3694 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/fanstatic_resources.py
+++ b/ckan/lib/fanstatic_resources.py
@@ -235,7 +235,7 @@ public = config.get('ckan.base_public_folder')
 base_path = os.path.abspath(os.path.join(
     os.path.dirname(__file__), '..', public, 'base'))
 
-log.info('Base path {0}'.format(base_path))
+log.debug('Base path {0}'.format(base_path))
 create_library('vendor', os.path.join(base_path, 'vendor'), depend_base=False)
 
 create_library('base', os.path.join(base_path, 'javascript'),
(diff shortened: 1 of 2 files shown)


amercader:
Thanks @davidread !

torfsen:
I know this has already been merged, but doesn't this PR just fix a symptom instead of removing the real cause? AFAIK #3694 is the reason why logging messages should go to STDERR by default, since STDERR won't be propagated when piping the output to another command.

I'm not sure how the log output ends up in STDOUT in the first place, but I suggest to fix that instead.

davidread:
@florianm good point. It appears I was using test-core.ini which outputs logs to stdout (which I think is arguable). Whereas the docs say I should have used development.ini which logs to stderr. So my bad here.

PR #3556, opened 2017-07-16
datastore upsert performance

wardi:
The `datastore_upsert` and `datastore_create` actions currently issue separate insert statements to the datastore database for every row added or updated instead of using a single query, which is simple but slow. The insert/update also abort completely if any individual row fails to be loaded.

This ticket is to track work improving performance and adding an option to continue when some records fail. When continuing on errors, each error will be collected and returned as a list of validation errors to the caller instead of as a single error.

Like #3523 we'll consider using the postgres `COPY` command to feed data to postgres. We can use [notices](http://initd.org/psycopg/docs/connection.html#connection.notices) to communicate errors from postgres to CKAN while loading data. This approach can work alongside data validation implemented in triggers #3428  if we choose a standard format for error text.

### DIFF
--- a/ckanext/datastore/set_permissions.sql
+++ b/ckanext/datastore/set_permissions.sql
@@ -49,6 +49,7 @@ GRANT SELECT ON ALL TABLES IN SCHEMA public TO {readuser};
 ALTER DEFAULT PRIVILEGES FOR USER {writeuser} IN SCHEMA public
    GRANT SELECT ON TABLES TO {readuser};
 
+-- a view for listing valid table (resource id) and view names
 CREATE OR REPLACE VIEW "_table_metadata" AS
     SELECT DISTINCT
         substr(md5(dependee.relname || COALESCE(dependent.relname, '')), 0, 17) AS "_id",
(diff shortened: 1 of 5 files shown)


wardi:
`json_populate_recordset` isn't working yet (so doesn't appear here) but the required work to introduce a trigger to populate _full_text and removing the python-side equivalent code is done. CSV upsert is going to be much faster than loading json regardless, so it's probably better to focus on that now that the trigger is in place.

I need to set this work aside for a bit but I hope to finish it off soon.

ckanext/datastore/set_permissions.sql
```diff
@@ -49,23 +49,60 @@ GRANT SELECT ON ALL TABLES IN SCHEMA public TO {readuser};
...
+$body$ LANGUAGE plpgsql;
+ALTER FUNCTION populate_full_text_trigger() OWNER TO {writeuser};
+
+-- migrate existing tables that don't have full text trigger applied
+DO $body$
+    BEGIN
+        EXECUTE coalesce(
+            (SELECT string_agg(
+                'CREATE TRIGGER zfulltext BEFORE INSERT OR UPDATE ON ' ||
+                quote_ident(relname) || 'FOR EACH ROW EXECUTE PROCEDURE ' ||
```
davidread: I needed a space before the FOR: `' FOR`

davidread:
We need the populate_full_text_trigger function added into the test database too - this just needs documenting I guess?

wardi:
Ideally we set up the test database with the same sql (and make sure we aren't deleting this function when we clear it)

davidread:
The code in this PR isn't directly about upsert performance, so I've moved that to a new PR https://github.com/ckan/ckan/pull/3786 (and rebased the commits in a new branch).

I suggest future work on upsert performance is done in a new PR.

PR #3701, opened 2017-07-16
Fix layout on mobile/desktop for bulk process in organization

klikstermkd:
_Issue only in master._

In the organization's bulk process, the layout got messed up with the inclusion of Bootstrap 3. 

There was extra space and the layout was using fixed widths instead of using Bootstrap's responsive classes.

With this PR, the layout is properly shown on mobile and desktop.

Here are 2 screenshots showing the extra space on smaller/bigger screen.

screenshot from 2017-07-16 16-58-35

screenshot from 2017-07-16 16-58-59


klikstermkd:
Ready for review.

PR #3702, opened 2017-07-16
Fix size for sidebar image

klikstermkd:
The image that shows up in the left sidebar in organizations/groups/user profiles is not properly scaled to the available space.

Here are 2 screenshots the demonstrate the issue. The first one is taken on a smaller size screen and the second on a bigger size screen.

screenshot from 2017-07-16 18-03-08

screenshot from 2017-07-16 18-03-20

This PR will change it to use responsive size of the image so that it scales nicely to whatever browser size.


klikstermkd:
Ready for review.

PR #3704, opened 2017-07-16
Remove minified files from test directory

klikstermkd:
_Issue appears only in master._

This PR removes minified files from the `test` directory, both from `public` and `public-bs2`.

These files are minified on `front-end-build` on a `release` branch.


klikstermkd:
Ready for review.

PR #3699, opened 2017-07-16
Upgrade jQuery to v3

klikstermkd:
Fixes #3665 

This PR upgrades jQuery to v3 which is the latest supported version of jQuery.

Beside the automated frontend tests which all pass, I did a manual test as well and everything seems to work.

PR #3705, opened 2017-07-16
Remove support for older versions of IE

klikstermkd:
With soon to be merged PR #3699 for jQuery 3, which drops support for IE 6-8, it's good time to remove some JS shims and polyfills for older versions of IE. 

I removed them only when in Bootstrap 3 mode, which is the default from CKAN 2.8, since maybe some folks are still gonna be on Bootstrap 2 and they probably have users with older versions of IE.


klikstermkd:
Ready for review.

PR #3720, opened 2017-07-23
Fix Explore button not working

klikstermkd:
_Issue appears only in master._

When jQuery was upgraded to v3, the Explore button stopped working in a dataset page.

This PR fixes that.

### DIFF
--- a/ckan/templates-bs2/package/snippets/resource_item.html
+++ b/ckan/templates-bs2/package/snippets/resource_item.html
@@ -18,11 +18,11 @@
   {% block resource_item_explore %}
   {% if not url_is_edit %}
   <div class="dropdown btn-group">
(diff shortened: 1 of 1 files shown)


klikstermkd:
Ready for review.

wardi:
check if the hover effect still works

klikstermkd:
@smotornyuk I checked this one, and the hover effect works, meaning the cursor changes on button hover.

wardi:
There's also some a11y issue with <a> tags missing an href. Something about screen readers not seeing them as proper links.

klikstermkd:
@wardi Now when I think better about this one, it doesn't make sense to me to use an `<a>` HTML tag for this one, since it's a dropdown toggle, and not an actual link that you can navigate to. 

I suggest to change it to use `<button>` instead. I tested it and it works.

smotornyuk:
Verified

PR #3722, opened 2017-07-24
Add a link to ckanext-rq to the background jobs documentation

torfsen:
Adds a link to [ckanext-rq](https://github.com/davidread/ckanext-rq) to the background jobs documentation.

It would be awesome if this could be part of the 2.7.0 documentation :wink:


davidread:
Good idea

smotornyuk:
Backported

PR #3726, opened 2017-07-24
add email validation to default_user_schema

tino097:
As for PR #3600 and making email required, its good to have validation of the user prompt for the email. In the default_user_schema is added `email_validator` method to check if the entered email is valid and in case of wrong input, adequate message will be displayed


smotornyuk:
Verified

PR #3733, opened 2017-07-26
Fix embed modal not properly shown

### LINKED ISSUE #3731: Embed modal window not working
master (2.8.0a)

@klikstermkd another small issue with the Bootstrap3 templates. The modal that comes up when you click on "Embed" in a resource view does not work. It looks like the ordering of the divs is wrong

localhost-5000-dataset-test-tartan-1-resource-ebcd7122-2207-45ad-9f11-0094823ff28d

klikstermkd:
Fixes #3731

### DIFF
--- a/ckan/public-bs2/base/javascript/modules/resource-view-embed.js
+++ b/ckan/public-bs2/base/javascript/modules/resource-view-embed.js
@@ -5,6 +5,7 @@ this.ckan.module('resource-view-embed', function ($) {
   function initialize() {
     self = this;
     modal = $('#embed-'+this.options.id)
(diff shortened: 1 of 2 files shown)


klikstermkd:
Ready for review.

smotornyuk:
@klikstermkd , it looks like you added fix to bs2 templates, where it was working initially. And bs3 templates still behave incorrectly

klikstermkd:
@smotornyuk Yes you are right. Instead of adding the fix for bs3, I put the code in the bs2 folder. Now it should be good.

Issue #3607, opened 2017-06-15
Compute and store hashes of resource files uploaded


wardi:
Optionally compute and store hashes of resource files uploaded to CKAN based on a new configuration option, eg:

```ini
ckan.resource_hash = md5 sha384
```

Would compute the MD5 and SHA384 hashes of resource files as they are uploaded (not as a separate pass) and store those values in the `resource.hash` field. CKAN's resource hash field has been unused since datastorer was replaced with datapusher. If `ckan.resource_hash` is not specified in the ini no existing values in `resource.hash` will be modified.

We'll use the SRI format for resource hash fields, eg a resource hash value would look like this if the site is configured to compute MD5 and SHA384:
```
md5-q0Zw+hGQWEK17VcrqD2uQg== sha384-dOTZf16X8p34q2/kYyEFm0jh89uTjikhnzjeLeF0FHsEaYKb1A1cv+Lyv4Hk8vHd
```

sorki:
+1, we managed to implement this in our plugin with after_create and after_update and were wondering why there's this field already.

https://github.com/otevrenamesta/ckanext-odczdataset/blob/74abed3e5b9684fb4f0ea0637c2246d0f9087893/ckanext/odczdataset/plugin.py#L158

jqnatividad:
Is this still being considered?  There are some installations interested in exposing the hash to ensure dataset integrity.

wardi:
My original issue description was incorrect: The resource hash **is** set by datapusher unconditionally to the md5 of the file pushed, when it runs its jobs.

I'd like this field to be changed as described above but it will require either a new field or changes to datapusher. In my mind it's much nicer to compute the hash while a file is uploaded instead of depending datapusher pushing the same data to a datastore table first.

jqnatividad:
+1 to @wardi's suggestion to compute the hash on the CKAN side, rather than datapusher.
* so we can have a hash for non-tabular data
* to catch when the same file has been uploaded, and not trigger datapusher

IMHO, its best to keep the existing field and show it in the resource's Additional Information:

However, MD5 should not be supported as it [has well-documented vulnerabilities](https://en.wikipedia.org/wiki/MD5). MD5 was used by datapusher to detect duplicate files.  For that usecase, MD5 is OK.  However, MD5 is not good enough for cryptographic purposes.

[...]

PR #3748, opened 2017-08-08
use "select for update" when retrieving package information for resource_create() / resource_update()

alexandru-m-g:
The discussion thread about this problem can be found in the ckan dev list here: https://lists.okfn.org/pipermail/ckan-dev/2017-July/011047.html

The main idea:
The resource_create() ( and similarly resource_update() ) does a package_show() followed later by a package_update() can potentially lead to a lost resource in some special cases.

Since postgres uses by default "read committed" transaction isolation the following could happen:
2 almost simultaneous requests (R1, R2) come to the API and are dealt with by different processes/threads. Both are modifying dataset D which already has one resource (resource1)

TIMELINE
 (R1) starts resource_create(resource2) on dataset D
 (R2) starts resource_update(resource1) on dataset D
 (R2) does package_show(D) => gets D with just resource1
 (R2) changes resource1 in D
 (R1) does package_show(D) => gets D with resource1
 (R1) adds resource2 to D => D.resources = [resource1, resource2]
 (R1) does package_update(D)
 (R1) resource_create() finishes, everything is committed successfully to the db => D has 2 resources in the db
 (R2) does package_update(D) - please note that here D only has one resource as read in step 3
 (R2) resource_update() finishes, everything is committed successfully to the db => D has just resource1 (resource2 disappears)

I've changed a bit package_show() and the get() function from Package model to use "with_for_update()" from sqlalchemy.

CAVEATS/TODO
- This does NOT tackle in any way the problem mentioned by @wardi about "a user opens a resource or dataset edit page in their browser in the morning, then makes some changes and submits them in the afternoon"
- this does NOT tackle potential similar problems with package_patch()

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -286,11 +286,9 @@ def resource_create(context, data_dict):
     if not data_dict.get('url'):
         data_dict['url'] = ''
 
-    # lock the package record in the db when selected via package_show()
-    context['for_update'] = True
-
+    # using "for_update" to lock the package record in the db when selected via package_show()
     pkg_dict = _get_action('package_show')(
-        dict(context, return_type='dict'),
+        dict(context, return_type='dict', for_update=True),
         {'id': package_id})
 
     _check_access('resource_create', context, data_dict)
@@ -330,7 +328,6 @@ def resource_create(context, data_dict):
                   uploader.get_max_resource_size())
 
     model.repo.commit()
-    context.pop('for_update')
 
     #  Run package show again to get out actual last_resource
     updated_pkg_dict = _get_action('package_show')(context, {'id': package_id})
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -286,6 +286,9 @@ def resource_create(context, data_dict):
     if not data_dict.get('url'):
         data_dict['url'] = ''
 
+    # lock the package record in the db when selected via package_show()
+    context['for_update'] = True
+
     pkg_dict = _get_action('package_show')(
         dict(context, return_type='dict'),
         {'id': package_id})
@@ -327,6 +330,7 @@ def resource_create(context, data_dict):
                   uploader.get_max_resource_size())
 
     model.repo.commit()
+    context.pop('for_update')
 
     #  Run package show again to get out actual last_resource
     updated_pkg_dict = _get_action('package_show')(context, {'id': package_id})
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -286,7 +286,7 @@ def resource_create(context, data_dict):
     if not data_dict.get('url'):
         data_dict['url'] = ''
 
-    # using "for_update" to lock the package record in the db when selected via package_show()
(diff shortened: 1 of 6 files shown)


ckan/logic/action/update.py
```diff
@@ -123,6 +125,8 @@ def resource_update(context, data_dict):
     upload.upload(id, uploader.get_max_resource_size())
     model.repo.commit()
```
wardi: if select_for_update we can expect some new errors from postgres here, right? We should catch those and turn them into validation errors we don't cause a server error and so the user knows to retry.

ckan/logic/action/create.py
```diff
@@ -286,6 +286,9 @@ def resource_create(context, data_dict):
     if not data_dict.get('url'):
         data_dict['url'] = ''
 
+    # lock the package record in the db when selected via package_show()
+    context['for_update'] = True
+
     pkg_dict = _get_action('package_show')(
         dict(context, return_type='dict'),
```
wardi: please add for_update=True here instead of modifying the passed context.

ckan/logic/action/update.py
alexandru-m-g: The main thing that should happen is for other transactions to wait for the lock to be released. Quoting the postgres documentation: "_FOR UPDATE causes the rows retrieved by the SELECT statement to be locked as though for update. This prevents them from being modified or deleted by other transactions until the current transaction ends. That is, other transactions that attempt UPDATE, DELETE, or SELECT FOR UPDATE of these rows will be blocked until the current transaction ends_."

There is also this in the documentation "_Within a SERIALIZABLE transaction, however, an error will be thrown if a row to be locked has changed since the transaction started_" BUT I think this is only the case when postgres is set to use the "SERIALIZABLE" transaction isolation level.

ckan/logic/action/create.py
alexandru-m-g: Committed another version where I modified the context just for the specific package_show() action ( same for both cases: update and create )

wardi:
I think I see a new problem: If a user uploads a file the dataset is locked for the duration. That could take minutes or hours with locks on the database.

We'll need to rearrange these functions to upload before locking the record, which also requires changing the IUploader interface. That's some work I've started but haven't made great progress on yet. #3607 

Ideas welcome.

alexandru-m-g:
Indeed, you are right, if a user uploads a big file there is a chance for the lock to remain on the package row for a long time. 

BUT, I was testing and I think that happened until now anyway (even without my changes in this PR). More precisely, resource_create() (or resource_update()) call **package_update()** and when package row update happens (https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L310 and https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L314) there's a lock (RowExclusiveLock) set. *That lock remains set till the end of the transaction which means it remains locked until the upload finishes*.
So I think that this PR, just locks the record a bit sooner.

@wardi I understand that this is a sensitive matter right at the core of CKAN, so would it be better if I add a config option that would allow the user to select whether to user "for update" ?

kmbn:

PR #3749, opened 2017-08-10
3499 skip url parsing in redirect

### LINKED ISSUE #3499: Redirecting to same page in non-root hosted ckan adds extra root_path to url
2.6.0, 2.6.1, master

In ckanext-harvest reharvest and clear buttons and in ckanext-report button redirect to the same page, if ckan is hosted in for example /data, I expect the redirection to /data/harvest/admin/someharvestname.

Resulting url is /data/data/harvest/admin/someharvestname

Clear or reharvest non-root hosted ckan.

### Some additional info

https://github.com/ckan/ckan/pull/2998 is probably the culprit since redirection works in ckan 2.3.5.
Will supply PR which should be validated for not breaking anything else.

Zharktas:
Fixes #3499 

### Proposed fixes:
If URL given to the h.redirect_to is a full URL or it begins with root_path if root_path is set, skip url_for parsing as it gives malformed URLs when the root_path is set. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -172,11 +172,16 @@ def redirect_to(*args, **kw):
                    '/user/logged_in']
     matching = [s for s in uargs if any(xs in s for xs in exempt_urls)]
 
+    _url = ''
+    skip_url_parsing = False
     if uargs and len(uargs) is 1 and isinstance(uargs[0], basestring) \
         and uargs[0].startswith('/') and len(matching) is 0:
-        return _routes_redirect_to(uargs[0])
+        skip_url_parsing = True
+        _url = uargs[0]
 
-    _url = url_for(*uargs, **kw)
+    if skip_url_parsing is False:
+        _url = url_for(*uargs, **kw)
+
     if _url.startswith('/'):
         _url = str(config['ckan.site_url'].rstrip('/') + _url)
 
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -153,6 +153,15 @@ def redirect_to(*args, **kw):
         toolkit.redirect_to('dataset_read', id='changed')
 
     '''
+    # Remove LANG from root_path so that we do not need to parse locales
+    root_path = config.get('ckan.root_path', None)
+    if root_path:
+        root_path = re.sub('/{{LANG}}', '', root_path)
+
+    # If args contain full url eg. http://example.com or url starting with root_path skip url parsing
+    if args and (is_url(args[0]) or ( root_path and args[0].startswith(root_path))) :
+        _routes_redirect_to(args[0])
+
     if are_there_flash_messages():
         kw['__no_cache__'] = True
 
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -166,17 +166,12 @@ def redirect_to(*args, **kw):
     uargs = map(lambda arg: str(arg) if isinstance(arg, unicode) else arg,
                 args)
 
-    # Repoze.who redirects with single strings
-    # which need to be parsed by url_for
-    exempt_urls = ['/user/logout', '/user/logged_out_redirect',
-                   '/user/logged_in']
-    matching = [s for s in uargs if any(xs in s for xs in exempt_urls)]
-
     _url = ''
     skip_url_parsing = False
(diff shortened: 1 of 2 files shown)


wardi:
@fanjinfei would you take a look at this if you have time?

fanjinfei:
looks good

ckan/lib/helpers.py
```diff
@@ -159,6 +159,18 @@ def redirect_to(*args, **kw):
     # Routes router doesn't like unicode args
     uargs = map(lambda arg: str(arg) if isinstance(arg, unicode) else arg,
                 args)
+
+    # Remove LANG from root_path so that we do not need to parse locales
+    root_path = config.get('ckan.root_path', None)
+    if root_path:
+        root_path = re.sub('/{{LANG}}', '', root_path)
```
wardi: is there a reason for re.sub here instead of simply using .replace?

ckan/lib/helpers.py
```diff
@@ -159,6 +159,18 @@ def redirect_to(*args, **kw):
...
+
+    # If args contain full url eg. http://example.com
+    # or url starting with root_path skip url parsing
+    if uargs and (is_url(uargs[0])
+                  or (root_path and uargs[0].startswith(root_path))):
+        return _routes_redirect_to(uargs[0])
```
wardi: if the root path starts with ``/{{LANG}}`` and that's included in `uargs[0]` then `uargs[0].startswith(root_path)` will be false. That's not what you want, right?

wardi:
Out of curiosity, what's generating urls that include the root path but not the host name and sending them to redirect_to? If the root path is the same as part of other normal urls isn't it hard to determine whether the prefix is the root path or not (e.g. if our root_path is /dataset how do you know if a '/dataset/...' url includes the root path?)

If we can keep these rules simple we should have fewer surprises. Ideally all our redirect_to calls to the same site will be passing the controller, action and args not a pre-populated url, right?

Zharktas:
This is just a thought but shouldn't this just redirect to whatever the function gets as a parameter if it is a single string ? That way we could skip the whole root_path parsing and we could redirect to: 
 * different hosts eg. http://example.com
 * to the same ckan site if the url is given as a string instead of controller, action, etc. for whatever reason. 
 * to the same domain but a different service, for example CMS which wouldn't have controller actions.

wardi:
@Zharktas that sounds like a much simpler idea!

Zharktas:
@wardi I've updated the PR to redirect if the arguments contain a single string.

Repoze.who redirects with single strings so there was a few urls that actually need to go through url_parsing, those are in exempt_urls list. Don't know if it is smart way to do it, but at least it works.

Zharktas:
Named routes apparently fail, will try to fix that soonish when I have time.

tino097:
This PR is fixing #3750

ckan/lib/helpers.py
```diff
@@ -152,14 +152,37 @@ def redirect_to(*args, **kw):
...
 
     # Routes router doesn't like unicode args
     uargs = map(lambda arg: str(arg) if isinstance(arg, unicode) else arg,
                 args)
-    _url = url_for(*uargs, **kw)
+
+    # Repoze.who redirects with single strings
+    # which need to be parsed by url_for
+    exempt_urls = ['/user/logout', '/user/logged_out_redirect',
+                   '/user/logged_in']
```
wardi: sadly, these can be cusomized in the who.ini so this won't work. Is there any other way we can detect this caller?

wardi: I guess we can update the docs for root_path to say that who.ini must also be updated to include the root.

Zharktas: It might be detectable by using __ckan_no_root parameter, since the only place it is used is in user controller during logins like [here](https://github.com/ckan/ckan/blob/master/ckan/controllers/user.py#L400). But I'll test it tomorrow if it helps on update the PR if it does.

Zharktas:
@wardi There was one repoze url which needs to go though url_for some reason, so I've added extra parameter for that. This also requires that the root_path is added to repoze.who paths post_login_url and post_logout_url so those should be documented. I've tested this locally with and without root_path and with locales in url and without it.

wardi:
@Zharktas looks good, and @tino097 reports this is working well. Thanks!

PR #3751, opened 2017-08-10
added blocks

ggcatu:
Fixes #

### Proposed fixes:

Add blocks in datastore plugin templates, dictionary.html and resource_read.html so we can easily customize and expand


ggcatu:
Why can't this be done?

```
      {% for f in ddict %}
            <tr>
 +            {% block resource_data_dictionary_items %}
              <td>{{ f.id }}</td>
              <td>{{ f.type }}</td>
              <td>{{ h.get_translated(f.get('info', {}), 'label') }}</td>
              <td>{{ h.render_markdown(
                h.get_translated(f.get('info', {}), 'notes')) }}</td>
 +            {% endblock %}
            </tr>
          {% endfor %}
```

ggcatu:
So, you cannot declare a variable inside a block, and expect to use it inside an inner one.

Don't know what's the workaround, but as for now should be working.

Any tip appreciated.

wardi:
A common pattern in the ckan templates is to use a snippet for repeated blocks like this, that makes it easy for an extension to override, and all variables are passed in.

We have a separate problem that snippets aren't being cached and slow down rendering, but that can be fixed in another PR.

ggcatu:
@wardi like this?

ckanext/datastore/templates/package/snippets/dictionary_table.html
```diff
@@ -0,0 +1,7 @@
+<tr>
+  <td>sdfasdf {{ f.id }}</td>
```
wardi: sdfasdf?

ckanext/datastore/templates/package/resource_read.html
```diff
@@ -9,29 +9,29 @@
...
-            <td>{{ f.type }}</td>
-            <td>{{ h.get_translated(f.get('info', {}), 'label') }}</td>
-            <td>{{ h.render_markdown(
-              h.get_translated(f.get('info', {}), 'notes')) }}</td>
-          </tr>
-        {% endfor %}
+        {% block resource_data_dictionary_data %}
+          {% set ddict=h.datastore_dictionary(res.id) %}
+          {% for f in ddict %}
+            {% snippet "package/snippets/dictionary_table.html", f=f, h=h %}
```
wardi: no need to pass h. please use a name like 'field' instead of 'f' since inside the snippet it's no longer clear from context what it contains.

ckanext/datastore/templates/datastore/dictionary.html
```diff
@@ -13,20 +13,7 @@
...
-          {'name': 'numeric', 'value': 'numeric'},
-          {'name': 'timestamp', 'value': 'timestamp'},
-          ], selected=f.get('info', {}).get('type_override', '')) }}
-        {{ form.input('info__' ~ loop.index ~ '__label',
-          label=_('Label'), id='field-f' ~ loop.index ~ 'label',
-          value=f.get('info', {}).get('label', ''), classes=['control-full']) }}
-        {{ form.markdown('info__' ~ loop.index ~ '__notes',
-          label=_('Description'), id='field-d' ~ loop.index ~ 'notes',
-          value=f.get('info', {}).get('notes', '')) }}
+          {% snippet "datastore/snippets/dictionary_form.html", f=f, position=loop.index, form=form %}
```
wardi: please pass `f` as `field`, no need to pass form (the snippet can import it itself)

ggcatu:
Should "No newline at end of file" be fixed?

ckanext/datastore/templates/package/resource_read.html
```diff
@@ -9,29 +9,29 @@
...
-            <td>{{ f.type }}</td>
-            <td>{{ h.get_translated(f.get('info', {}), 'label') }}</td>
-            <td>{{ h.render_markdown(
-              h.get_translated(f.get('info', {}), 'notes')) }}</td>
-          </tr>
-        {% endfor %}
+        {% block resource_data_dictionary_data %}
+          {% set dict=h.datastore_dictionary(res.id) %}
+          {% for field in dict %}
+            {% snippet "package/snippets/dictionary_table.html", field=field, h=h %}
```
wardi: no need for h=h here

tino097:
@amercader @wardi i think this can be backported too

amercader:
Unless it's fixing an actual bug, I wouldn't backport changes in the template structure, as we might break extensions.

Issue #3752, opened 2017-08-11
Uploading large files never frees tmp-file disk space (disk space leak)


curtisyallop:
### CKAN Version if known (or site URL)
2.6.0 and 2.7.0 (both versions tested)

### Please describe the expected behaviour
After uploading 1GB file repeatedly (and removing the old one each time), ckan should keep running just fine - not die due to running out of disk space. Restarting the web server should not free up all the missing space and make it work again.

### Please describe the actual behaviour
Uploading a 1GB file multiple times as described above makes ckan die after about 5-6 uploads. As I watch free disk space ("watch -n .2 df -h", Avail column) as the upload executes, I can see the free space dropping. Ignore the first upload. It often drops 2GB each 1GB upload (it drops 4-5GB in total but recovers a couple GB). But sometimes it stays the same for 1-2 uploads. But after about 4-6 uploads, ckan dies (I have 6GB free space in my virtual machine and both /tmp and datastore are on the root filesystem). If I restart ckan (either paster-serve or apache), the disk space immediately is all freed up.

I made a possible fix which prevents this leak. In /usr/lib/ckan/default/src/ckan/ckan/lib/uploader.py, `ResourceUpload.upload`, after "os.rename" and before "return", I do "self.upload_file.close()". In /usr/lib/ckan/default/lib/python2.7/site-packages/paste/cascade.py (pylons), `Cascade.__call__`, I altered the method to perform "f.close()" if file "f" was created (it is created on line "f = tempfile.TemporaryFile()"), just before returning from the method. I am trying to figure out if there is any better solution to preventing this disk-space-leak problem.

Note: "strace -f" on paster serve revealed 3 files were created "/tmp/tmpT_rCJn", "/tmp/tmpET4MoY", "/mnt/datastore/resources/e92/813/3a-8f09-4f2d-8900-e89d0de1ce7c~".

lsof -c python | grep /tmp
python  12815 vagrant    9u   REG     253,0 1073743123 26707027 /tmp/tmp0V5iAe (deleted)
python  12815 vagrant   10u   REG     253,0 1073743123 28579426 /tmp/tmp9DcPv0 (deleted)
Why are these files being held open?

### What steps can be taken to reproduce the issue? 
[...]

PR #3756, opened 2017-08-16
[#3752] CloseWSGIInput middleware + update in ResourceUploader

smotornyuk:
After some investigation of #3752  i found that webob  [creates copy](https://github.com/Pylons/webob/blob/1.0.8/webob/request.py#L546) of request body in some cases(details are not so relevant here). This results in temporary files that still opened until server restart. I suggest additional middleware,that will close such files. 
And, thanks to @curtisyallop for solution, this PR contains fix for ResourceUpload class - it previously relied on automatic file closing as well. Now it performed via context manager


ckan/lib/uploader.py
```diff
@@ -277,25 +277,25 @@ def upload(self, id, max_size=10):
...
-                    break
-                output_file.write(data)
-                if current_size > max_size:
-                    os.remove(tmp_filepath)
-                    raise logic.ValidationError(
-                        {'upload': ['File upload too large']}
-                    )
+            with open(tmp_filepath, 'wb+') as output_file:
+                with self.upload_file:
+                    self.upload_file.seek(0)
```
TkTech: How is `seek` working on a stream?

smotornyuk: It's not a stream, it is file descriptor

TkTech: Right, of an unbuffered stream?

smotornyuk: But what if `self.upload_file` is a StringIO object(this is case for test and, probably, that could be used for programmatical resource creation)? In that case you need to reset pointer and then copy buffer content into file

neilmayhew:
@TkTech Any chance of getting this finished up and merged soon?

@smotornyuk I think it would be helpful to squash all the commits into one, and rebase the branch onto the current state of `master`.

TkTech:
There are some circle.ci tests failing but I can't see how they were related to the changes in this PR. Will merge tomorrow if there are no further comments.

smotornyuk:
This errors came from current migration to flask(i added fresh changes into this PR).

ckan/lib/uploader.py
```diff
@@ -277,25 +277,25 @@ def upload(self, id, max_size=10):
...
+
+                    if not data:
+                        break
+                    output_file.write(data)
+                    if current_size > max_size:
+                        os.remove(tmp_filepath)
+                        raise logic.ValidationError(
+                            {'upload': ['File upload too large']}
+                        )
+                self.upload_file.close()
```
neilmayhew: Isn't there a possibility of a file leak here, eg if the 'upload too large' exception is thrown? Shouldn't the code be wrapped in a `try ... finally` block?

curtisyallop: Sometimes "upload" is type "file" but for small files, ckan uses type StringO (cStringIO.OutputType) which does not implement __enter__/__exit__ so we cannot use "with:". smotornyuk's previous version used "with:" but we had to switch back to ".close()".

neilmayhew: Yes, I was talking about leaking `self.upload_file`, not the file returned by `open(tmp_filepath)`.

tino097: i see what you meant, i thought we have `file` object for upload too

smotornyuk: Yeah, it has sense. @neilmayhew , thanks, i'll check memory leak here(i believe, this piece will cause it, but checks are always valuable:) ) and then i'll update this piece as you suggested

smotornyuk:
@neilmayhew , you are right, temporary file not removed in case of ValidationError. 

And you may be interested in little change in middleware - now actual file descriptor from request checked rather than `cgi.FileField` itself, because I noticed that it treated like falsy value. Not sure, whether it's because of some update, or there was such behavior initially(in that case, fix shouldn't be working previously :thinking: )

curtisyallop:
@smotornyuk , I can confirm that the falsey behavior was there initially. And the previous fix did not fix the leak. Regular "upload" values of type cgi.FieldStorage evaluate to false for us on ckan 2.3.5 and 2.6 (we run both versions right now). I reported this falsey problem earlier. It used to be here but seems to have disappeared from the githib history. I made a fix and you incorporated this suggestion - to change the code from "if upload:" to "if upload is not None:" (this change also disappeared from github history). Unfortunately my fix, while fixing the falsey problem, had a bug in it - creating url links now failed. At the time, I suggested that the falsey problem might be specific to 2.3.5 but actually it was not.

So to summarize possible values for "upload": "upload" can be None (non-upload regular ckan web site requests), empty string (update resource, create link) or type cgi.FieldStorage (regular uploads). And "if upload:" evaluates to false for type cgi.FieldStorage.

Your latest change looks great. Thanks for your help with this.

curtisyallop:
To test the upload, I propose 5 cases to be tested: (1) Small file (eg 5 bytes), (2) Large file (eg 60MB), (3) Very large file (eg 1GB) (4) Create link (5) Regular site accesses to any page. To test that the leak is fixed, I find the following line works well: `watch -n .2 'df -h; echo; sudo lsof -c python -c httpd -c nginx | grep /tmp'`. lsof shows the non-closed '(deleted)' file handles eating up tmp space. df shows available free disk space dropping. 1GB uploads work well to see space getting eaten up. Note that small files (eg 5 bytes) and large files (eg 60MB) are handled differently in lib/uploader.py.

ckan/config/middleware/common_middleware.py
```diff
@@ -28,6 +29,30 @@ def __call__(self, environ, start_response):
...
+
+    This middleware is supposed to close such files after each request.
+    '''
+    def __init__(self, app, config):
+        self.app = app
+
+    def __call__(self, environ, start_response):
+        wsgi_input = environ['wsgi.input']
+        if isinstance(wsgi_input, FakeCGIBody):
+            upload = wsgi_input.vars.get('upload')
```
wardi: These file objects could be passed as any parameter, not just 'upload'. Can you make this fix more general?

smotornyuk:
Updated. 
PS. i checked flask_app - it doesn't has such problem, so middleware should be applied only to pylon

ckan/config/middleware/common_middleware.py
curtisyallop: Good point. For example the manage-organization image/icon upload is called "image_upload".

ckan/lib/uploader.py
```diff
@@ -22,6 +22,21 @@
 _max_image_size = None
 
 
+def _copy_file(input_file, output_file, max_size):
+    input_file.seek(0)
+    current_size = 0
+    while True:
+        current_size = current_size + 1
+        # MB chunks
+        data = input_file.read(2**20)
```
wardi: since you've moved this code, do you mind giving `2**20` a name? Just that and a rebase or merge to resolve the conflict and I'll merge, Thanks!

smotornyuk:
[Updated](https://github.com/smotornyuk/ckan/blob/e14ede256d93323db2f6cff70f18d7cf183b0a5f/ckan/lib/uploader.py#L18-L33)

wardi:
@smotornyuk thank you!

PR #3762, opened 2017-08-17
Blueprints home controller

tino097:
Work in progress migrating home controller to flask


ckan/lib/helpers.py
```diff
@@ -813,8 +817,12 @@ def _make_menu_item(menu_item, title, **kw):
 
     This function is called by wrapper functions.
     '''
+    # display the menu item if it comes from flask
+    if '.' in menu_item:
+        link = _link_to(title, menu_item, suppress_active_class=True)
+        return literal('<li>') + link + literal('</li>')
```
smotornyuk: here we are missing active class from current route. i think that this piece should be removed and copying named routes can be slightly improved

ckan/config/middleware/flask_app.py
```diff
@@ -341,3 +341,10 @@ def is_blueprint(mm):
         for blueprint in inspect.getmembers(module, is_blueprint):
             app.register_blueprint(blueprint[1])
             log.debug(u'Registered core blueprint: {0!r}'.format(blueprint[0]))
+            # print app.url_map
+            # building the routes and saving to config
+            # need to be improved
+            for rule in app.url_map.iter_rules():
+                config['routes.named_routes'].update({
+                    rule.endpoint: rule.rule
+                })
```
smotornyuk: How about:
```python
            for rule in app.url_map.iter_rules():
                if '.' not in rule.endpoint:
                    continue
                controller, action = rule.endpoint.split('.')
                route = {
                    rule.endpoint: {
                        'action': action,
                        'controller': controller,
                        'highlight_actions': action,
                        'needed': list(rule.arguments)
                    },

                }
                config['routes.named_routes'].update(route)
```
this will convert our blueprints to items quite similar to pylons routes

smotornyuk:
I currently working on admin controller(based on your PR :) ) and i found out, that there are some problems with identifying current user. This is my current state of `flask_app.py`. In short words, currently Flask do not uses repoze.who so user is not authenticated at all. This if fix for problem. Could you review and, maybe apply it?
```diff
diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
index 50e9fab..c66aae5 100644
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -18,6 +18,8 @@ from flask_babel import Babel
 from beaker.middleware import SessionMiddleware
 from paste.deploy.converters import asbool
 from fanstatic import Fanstatic
+from repoze.who.config import WhoConfig
+from repoze.who.middleware import PluggableAuthenticationMiddleware
 
 import ckan.model as model
 from ckan.lib import helpers
@@ -194,6 +196,23 @@ def make_flask_stack(conf, **app_conf):
     for key in flask_config_keys:
         config[key] = flask_app.config[key]
 
+    # Initialize repoze.who
+    who_parser = WhoConfig(conf['here'])
+    who_parser.parse(open(app_conf['who.config_file']))
+
+    app = PluggableAuthenticationMiddleware(
+        app,
+        who_parser.identifiers,
+        who_parser.authenticators,
+        who_parser.challengers,
+        who_parser.mdproviders,
+        who_parser.request_classifier,
+        who_parser.challenge_decider,
+        logging.getLogger('repoze.who'),
+        logging.WARN,  # ignored
+        who_parser.remote_user_key
+    )
+
     # Add a reference to the actual Flask app so it's easier to access
     app._wsgi_app = flask_app
 
@@ -217,6 +236,7 @@ def ckan_before_request():
     # Update app_globals
     app_globals.app_globals._check_uptodate()
 
+    g.controller, g.action = request.url_rule.endpoint.split('.')[:2]
     # Identify the user from the repoze cookie or the API header
     # Sets g.user and g.userobj
     identify_user()
@@ -344,7 +364,18 @@ def _register_core_blueprints(app):
             # print app.url_map
             # building the routes and saving to config
             # need to be improved
+
             for rule in app.url_map.iter_rules():
-                config['routes.named_routes'].update({
-                    rule.endpoint: rule.rule
-                })
+                if '.' not in rule.endpoint:
+                    continue
+                controller, action = rule.endpoint.split('.')
+                route = {
+                    rule.endpoint: {
+                        'action': action,
+                        'controller': controller,
+                        'highlight_actions': action,
+                        'needed': list(rule.arguments)
+                    },
+
+                }
+                config['routes.named_routes'].update(route)
```

ckan/lib/helpers.py
tino097: yes i agree about this, i've added just for testing to keep the flow going with less modifications possible

tino097:
@smotornyuk thank you very much, now i will start working on the index

ckan/lib/helpers.py
```diff
@@ -678,8 +679,8 @@ def nav_link(text, *args, **kwargs):
 
     '''
     if is_flask_request():
-        # checking if its flask request and calling directly 'url_for' for test
-        link = url_for(str(args))
+        # checking if its flask request and calling '_link_to'
```
tino097: @smotornyuk @amercader i would appreciate if you can check if this could be a good workaround

tino097:
https://pastebin.com/GgjjXpGr i have this error, any suggestions

tino097:
Thanks @amercader, my initial idea with this PR was to have some starting point where we could discuss about next steps and after to come with cleaner PR

ckan/config/middleware/flask_app.py
```diff
@@ -341,3 +364,19 @@ def is_blueprint(mm):
         for blueprint in inspect.getmembers(module, is_blueprint):
             app.register_blueprint(blueprint[1])
             log.debug(u'Registered core blueprint: {0!r}'.format(blueprint[0]))
+            # print app.url_map
+            # building the routes and saving to config
+            # need to be improved
+            for rule in app.url_map.iter_rules():
```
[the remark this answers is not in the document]
tino097: i think we have done this to get the routes in same place, because i had issues with the routes 
https://github.com/ckan/ckan/blob/3229-api-blueprint/ckan/lib/helpers.py#L869 so this looks like good solution. 

Maybe we could do this after core and extension blueprints are registered

ckan/config/environment.py
```diff
@@ -9,6 +9,7 @@
 
 import sqlalchemy
 from pylons import config as pylons_config
+from flask import config as flask_config
```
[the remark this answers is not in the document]
tino097: i will remove it

ckan/lib/app_globals.py
```diff
@@ -70,6 +70,7 @@
     'ckan.user_list_limit': {'default': '20', 'type': 'int'},
     'search.facets.default': {'default': '10', 'type': 'int',
                              'name': 'facets_default_number'},
+    'search_facets_limits': {'default': 3},
```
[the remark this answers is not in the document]
tino097: i will remove this but im not sure why im getting that error

ckan/config/middleware/flask_app.py
```diff
@@ -217,6 +238,8 @@ def ckan_before_request():
     # Update app_globals
     app_globals.app_globals._check_uptodate()
 
+    g.controller, g.action = request.url_rule.endpoint.split('.')[:2]
```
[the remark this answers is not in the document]
smotornyuk: This variables are used inside helpers that build navigation links. Without them [this](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L689) line ends up with error. 
As a quick solution, i use [this code](https://github.com/ckan/ckan/pull/3775/commits/e3cb97851f8d2d6cd64f6d35affcb6a3f57effea), but i'm not sure, whether it will work for index actions of new blueprints..

ckan/views/home.py
```diff
@@ -0,0 +1,85 @@
...
+        return render_template(u'home/index.html')
+
+
+class AboutView(View):
+    def dispatch_request(self):
+        return render_template(u'home/about.html')
+
+
+util_rules = [
+    (u'/', HomeView.as_view("/")),
```
smotornyuk: I think, that it would be better to preserve historical name, like `index` for such routes

ckan/lib/app_globals.py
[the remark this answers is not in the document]
tino097: @amercader im not able to start the app without this line, any suggestions ?

amercader:
> Thanks @amercader, my initial idea with this PR was to have some starting point where we could discuss about next steps and after to come with cleaner PR

You are right @tino097, sorry if I was being too thorough. I'll ignore comments and formatting stuff from now on

ckan/lib/app_globals.py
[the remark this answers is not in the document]
smotornyuk: As i understand, this variable is required by [search snippet](https://github.com/ckan/ckan/blob/master/ckan/templates/home/snippets/search.html#L1). So, if we are going to avoid adding global variables - we need to modify `get_facet_items_dict` helper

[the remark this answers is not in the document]
amercader: Exactly, this [helper](https://github.com/tino097/ckan/blob/d5917bc8fc56498a503d07b5145e2916355bfbc8/ckan/lib/helpers.py#L931) is using `c.search_facets_limit`, which in the old Pylons based globals failed silently to an empty string so the helper didn't fail.

It looks like this property is actually set by the package and group controllers to track some state so it can't be moved to `config`:

```
ag search_facets_limits
ckan/lib/helpers.py
931:    if c.search_facets_limits and limit is None:
932:        limit = c.search_facets_limits.get(facet)
963:    if c.search_facets_limits and limit is None:
964:        limit = c.search_facets_limits.get(facet)

ckan/controllers/group.py
358:            c.search_facets_limits = {}
362:                c.search_facets_limits[facet] = limit

ckan/controllers/package.py
298:        c.search_facets_limits = {}
307:            c.search_facets_limits[facet] = limit
```

In this case lets do `hasattr(c, 'search_facets_limit')` on the helper to avoid the exception and not add it to the globals object by default. Does this make sense?

Also worth reading https://github.com/ckan/ckan/pull/3203 for context on `c` / `g`.

ckan/views/home.py
tino097: @smotornyuk i will set it to `index`

ckan/lib/app_globals.py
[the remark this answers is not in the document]
wardi: can we not pass the filter information as a normal parameter instead?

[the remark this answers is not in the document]
amercader: It is trickier than it seems: https://gist.github.com/amercader/ce4cae4a789a7db6fb78d304164e552a, and I didn't manage to make it work on the group controller because of how it is structured. Also there are a ton of other search and facet related variables in `c`, it might be best to refactor the whole thing consistently and not individual vars at this point.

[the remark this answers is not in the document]
wardi: yes, that makes sense to me

tino097:
I guess we should identify the working changes and i could work on separate PR with those

ckan/config/middleware/flask_app.py
```diff
@@ -360,3 +379,16 @@ def is_blueprint(mm):
         for blueprint in inspect.getmembers(module, is_blueprint):
             app.register_blueprint(blueprint[1])
             log.debug(u'Registered core blueprint: {0!r}'.format(blueprint[0]))
+            for rule in app.url_map.iter_rules():
```
amercader: After looking at the code it's now clear to me as well that we can't avoid `config['routes.named_routes']` until we just have Flask endpoints. So let's keep this, but I think it should be moved to another place where it only gets run once and it also cover extension. A good spot seems to be [here](https://github.com/tino097/ckan/blob/8d8c222c71a9d4c6be2d1f9e8613ebfc2286363c/ckan/config/middleware/flask_app.py#L157) @smotornyuk WDYT?

ckan/lib/helpers.py
```diff
@@ -684,6 +685,10 @@ def are_there_flash_messages():
 
 def _link_active(kwargs):
     ''' creates classes for the link_to calls '''
+    if is_flask_request():
+        endpoint = request.url_rule.endpoint
+        action = request.url_rule.rule
+        return endpoint.split('.')[1] == action.translate('/')
```
amercader: Maybe I'm missing something but this will never match a menu item (passed via `kwargs`) against the current endpoint loaded. I think it should work like this:

```diff
diff --git a/ckan/lib/helpers.py b/ckan/lib/helpers.py
index e451ba0..bc54a58 100644
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -687,8 +687,11 @@ def _link_active(kwargs):
     ''' creates classes for the link_to calls '''
     if is_flask_request():
         endpoint = request.url_rule.endpoint
-        action = request.url_rule.rule
-        return endpoint.split('.')[1] == action.translate('/')
+        return (
+            kwargs.get('controller') == endpoint.split('.')[0] and
+            kwargs.get('action') == endpoint.split('.')[1]
+        )
+
     highlight_actions = kwargs.get('highlight_actions',
                                    kwargs.get('action', '')).split()
     return (c.controller == kwargs.get('controller')
```

Also to make easier to get rid of Pylons code in the future let's use separate functions for Flask / Pylons rather than doing checks inside. So something like:

```python

def _link_active(kwargs):
     ''' creates classes for the link_to calls '''
     if is_flask_request():
         return _link_active_flask(kwargs)
     else:
         return _link_active_pylons(kwargs)
```

ckan/lib/helpers.py
```diff
@@ -730,6 +735,10 @@ def nav_link(text, *args, **kwargs):
     :param condition: if ``False`` then no link is returned
 
     '''
+    if is_flask_request():
+        # checking if its flask request and calling '_link_to'
+        link = _link_to(text, *args, **kwargs)
+        return link
```
amercader: See above for separate functions for Flask / Pylons

ckan/lib/helpers.py
```diff
@@ -922,11 +930,12 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
...
-    if c.search_facets_limits and limit is None:
-        limit = c.search_facets_limits.get(facet)
-    # zero treated as infinite for hysterical raisins
-    if limit is not None and limit > 0:
-        return facets[:limit]
+    if hasattr(c, 'search_facets_limits'):
+        if c.search_facets_limits and limit is None:
+            limit = c.search_facets_limits.get(facet)
+        # zero treated as infinite for hysterical raisins
+        if limit is not None and limit > 0:
```
amercader: These two lines should not be inside the `if hasattr(c, 'search_facets_limits'):` check

ckan/views/home.py
```diff
@@ -0,0 +1,85 @@
...
+        if g.userobj and not g.userobj.email:
+            url = h.url_for(controller=u'user', action=u'edit')
+            msg = _(u'Please <a href="%s">update your profile</a>'
+                    u' and add your email address. ') % url + \
+                _(u'%s uses your email address'
+                    u' if you need to reset your password.') \
+                % config.get(u'ckan.site_title')
+            h.flash_notice(msg, allow_html=True)
+
+        return render_template(u'home/index.html')
```
amercader: We need to use the common render function in CKAN (after making sure it works in Flask) but for now let's use this.

ckan/views/home.py
```diff
@@ -0,0 +1,85 @@
...
+                u'rows': 4,
+                u'start': 0,
+                u'sort': u'views_recent desc',
+                u'fq': u'capacity:"public"'
+            }
+            query = logic.get_action(u'package_search')(
+                context, data_dict)
+            g.search_facets = query['search_facets']
+            g.package_count = query['count']
+            g.datasets = query['results']
```
amercader: As mentioned in https://github.com/ckan/ckan/pull/3203 we should take the opportunity of the rewrite for deprecating as much of these as possible (ie we still keep them in `g` for a version, but we pass explicitly the vars to the templates). This might not be possible with all of them (like in `g.search_facets_limits`) but if possible, let's deprecate them.

ckan/views/home.py
```diff
@@ -0,0 +1,85 @@
...
+                    u' if you need to reset your password.') \
+                % config.get(u'ckan.site_title')
+            h.flash_notice(msg, allow_html=True)
+
+        return render_template(u'home/index.html')
+
+
+class AboutView(View):
+    def dispatch_request(self):
+        return render_template(u'home/about.html')
```
amercader: This could be a simple view function

ckan/views/home.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+util_rules = [
+    (u'/', HomeView.as_view("index")),
+    (u'/home', HomeView.as_view("home")),
```
amercader: Why not use the same name for both? eg `index`

ckan/views/home.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return render_template(u'home/index.html')
+
+
+class AboutView(View):
+    def dispatch_request(self):
+        return render_template(u'home/about.html')
+
+
+util_rules = [
```
amercader: `util_rules` was just a name related to the Utils API on the API blueprint, I think `rules` will be less confusing.

amercader:
@tino097 good idea. I added some comments, and here's a summary of them:

* `config['routes.named_routes']` in Flask: let's keep it, but in a better place if possible
* `_link_active()` helper: fix in Flask and separate in 2 functions
* `_nav_link()` helper: separate in 2 functions
* `get_facet_items_dict()` helper: Fix two lines
* Home blueprint: See how many `g` vars can we deprecate, minor comments

amercader:
Closed in favour of #3891

Issue #3776, opened 2017-08-22
[DOCKER] - Change .ini name


deinok:
ref: #3764
The .ini its called ckan.ini
Rename it to production.ini

PR #3777, opened 2017-08-22
Fix Issue #3776

deinok:


deinok:
Travis known issue. Not dependent to this.

amercader:
We could change the name to production.ini for consistency with other installation methods, but I don't think this fixes an actual problem. If the file was created with this name if all commands used the same name (ckan.ini) it should work fine (ie it is not required that the file is named production.ini)

In any case this will conflict with changes in https://github.com/ckan/ckan/pull/3692, which should be ready to be merged soon, so if necessary, the change should be made there.

PR #3779, opened 2017-08-22
Prevent multiple consecutive form submissions

### LINKED ISSUE #3773: Multiple resources created for a dataset
All versions.

When a new resource is created through the web UI, if the "Add" button in the form is clicked multiple times consecutively, only one resource should be created.

When a new resource is created through the web UI, if the "Add" button in the form is clicked multiple times consecutively more than one resource is created.

klikstermkd:
Fixes #3773 

### Features:

This change is going to prevent sending multiple form submissions on every form where the `basic-module` is attached, such as the Organization/Group/Dataset forms.


klikstermkd:
Ready for review.

tino097:
@amercader another candidate for backporting

amercader:
This one adds a bit of complexity on the JS side and also broke the tests, so perhaps better not to backport it.

Issue #3781, opened 2017-08-23
_mail_recipient header override


ggcatu:
### CKAN Version if known (or site URL)
https://github.com/ckan/ckan/blob/3fa8d9b7476a9f37761749804e4613c196f4fa26/ckan/lib/mailer.py#L30-L36
Mailer is actually adding more keys to the MIMEText, isntead of replacing them.

### Please describe the expected behaviour
To change the Content-Type to the new one

### Please describe the actual behaviour
The old Content-Type is kept, but a new Content-Type key is created

amercader:
@ggcatu This sounds really sensible and looks like you know where the code should be changed. Do you want to submit a PR for this?

ggcatu:
Check https://github.com/ckan/ckan/pull/3784

amercader:
Thanks @ggcatu

PR #3784, opened 2017-08-25
Fix replace, add

ggcatu:
#3781 

### Proposed fixes:

Use MIMEText methods instead of trying to modify headers by hand.


tino097:
@amercader could be backported too

PR #3792, opened 2017-08-31
Fix JS error on tests that was breaking the CircleCI build

amercader:
A change introduced in #3779 had a syntax that is not supported in node.
This broke the JS tests, and the CircleCI build.


wardi:
let's just get this in so we can have passing tests

PR #3795, opened 2017-08-31
Update copyright name, years and include contributors, as per README.rst

davidread:
I notice that the README copyright message now acknowledges contributors too:
https://github.com/ckan/ckan/blob/master/README.rst#copying-and-license
so this brings the other copyright messages in line with that one - in the license file and docs footer.


amercader:
+1

PR #2562, opened 2017-09-01
allow datastore_search_sql on private datasets

wardi:
We're using EXPLAIN to list the tables that a user is accessing in datastore_search_sql, but we're blocking access to all private tables by removing access at the postgres permission level.

Instead of denying all sql on private datasets, let's call the resource_show auth method for each table the user is accessing and refuse the query only when any of those checks fail.

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -46,7 +46,7 @@ def owner_org_validator(key, data, errors, context):
     if not group:
         raise Invalid(_('Organization does not exist'))
     group_id = group.id
-    if not(user.sysadmin or
+    if not context.get(u'ignore_auth', False) and not(user.sysadmin or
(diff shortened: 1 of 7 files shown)


wardi:
This still needs a migration in set_permissions.sql, maybe better to merge #3786 first?

wardi:
Re-running `paster datastore set-permissions | psql ...` is necessary for this change to work with existing private datastore tables, which is the same thing #3786 suggests in its documentation change

wardi:
Marking this as WIP again because this change combined with some postgres extensions (which would have to be manually installed) can expose private tables and I'm working on a general fix.

wardi:
@amercader, @smotornyuk  this is good to go, crosstabs are the only extension included with postgres that could leak data and this small change makes it safe to use them.

PR #3802, opened 2017-09-05
Try to fix travis

ggcatu:
### Proposed fixes:

> apt-get install -y build-essential libssl-dev libffi-dev

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,9 @@ RUN apt-get -q -y update && apt-get -q -y upgrade && DEBIAN_FRONTEND=noninteract
         python-virtualenv \
         libpq-dev \
         git-core \
+        build-essential \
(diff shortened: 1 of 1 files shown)


ggcatu:
Probably not all 3 libraries are needed. 

Maybe trial and error would give the right one.

amercader:
Thanks @ggcatu!

deinok:
@ggcatu The travis CI only setup a docker instance and tes if it is serving CKAN.

So it seems more a problem with undefined or unmantained libs or a commit that breaks the compatibility with the docker images.

So, this fix only shows a more heavy issue.
Also, nice catch ;)

ggcatu:
@deinok Yep, of course, problem was while trying to build the image, that cryptography was complaining during compilation, because missing dependencies

PR #3803, opened 2017-09-05
[#3799] Add some missing major changes to change log.

### LINKED ISSUE #3799: Add missing major changes to change log
[As discussed on ckan-dev](https://lists.okfn.org/pipermail/ckan-dev/2017-September/011103.html): Our [update-guide](http://docs.ckan.org/en/ckan-2.7.0/maintaining/upgrading/index.html) doesn't mention version-specific changes (AFAIK by design), and the [change log](https://github.com/ckan/ckan/blob/177143fca601cad334d0e6be09a4224ac9c85645/CHANGELOG.rst) only lists certain categories of changes (e.g. database scheme changes, Solr scheme changes). This means that some important changes that require attention during an update (e.g. for 2.7 the new hard dependency on Redis) are not properly documented (they are documented in the *installation* guide, but not in the update guide).

We should therefore extend the change log so that it also covers:

* New/changed/removed configuration options (for 2.7: `ckan.redis.url`)
* New/changed/removed software dependencies (for 2.7: Redis)

torfsen:
Fixes #3799.

The new hard requirement of Redis and the corresponding configuration option `ckan.redis.url` previously wasn't explicitly documented.

This commit also clarifies the update procedure for package and source installs w.r.t. new/changed 3rd-party package-requirements and configuration options.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,20 +25,20 @@ Note: This version requires re-running the 'datastore set-permissions' command
 v2.7.0 2017-08-02
 =================
 
-Note: Starting from this version, CKAN requires at least Postgres 9.3
+General notes:
+ * Starting from this version, CKAN requires at least Postgres 9.3
+ * Starting from this version, CKAN requires a Redis database. Please
+   refer to the new `ckan.redis.url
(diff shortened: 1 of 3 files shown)


amercader:
Thanks @torfsen!

amercader:
@smotornyuk we forgot to pull the changes from the last releases to the master changelog

PR #3807, opened 2017-09-08
config option for type used to display all dataset types

mattfullerton:
This commit allows display of all package types via url argument (_show_all_types=True) similar to the ckan.search.show_all_types config option.

The User story is such:
I start building a CKAN site and realise I want to archive a bunch of things that aren't really "datasets" - let's say documents. I create a new dataset type and schema and am broadly happy that datasets and documents are handled separately in CKAN. However, there are few, _theme-specific_ places where I would like to provide a common listing, for example by hyperlinking metadata fields on a dataset/document description field (e.g. now show me _everything_ from year 2005) or under a list of common categories I use on my front page or in a menu to different listings of data.

The only way I've been able to think of to resolve such a use case is to create a third (second custom) dataset type for my actual datasets and then set the ckan.search.show_all_types option to True and rearrange links so that most of the time I go to one dataset type or the other and in certain cases use the default /dataset to show everything. This seems like overkill and also creates terrible problems with nomenclature: what alternative name to I give a dataset that is actually... a dataset? I can't think of a good English word.

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -228,10 +228,12 @@ class PackageController(base.BaseController):
                 # Only show datasets of this particular type
                 fq += ' +dataset_type:{type}'.format(type=package_type)
             else:
-                # Unless changed via config options, don't show non standard
-                # dataset types on the default search page
+                # Unless changed via config options or url parameter,
+                # don't show non standard dataset types on the default
+                # search page
                 if not asbool(
-                        config.get('ckan.search.show_all_types', 'False')):
+                        config.get('ckan.search.show_all_types', 'False')) \
+                        and not request.params.get('_show_all_types', False):
                     fq += ' +dataset_type:dataset'
 
             facets = OrderedDict()
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -224,17 +224,26 @@ class PackageController(base.BaseController):
                        'user': c.user, 'for_view': True,
                        'auth_user_obj': c.userobj}
 
-            if package_type and package_type != 'dataset':
+            # Unless changed via config options, don't show other dataset
+            # types any search page. Potential alternatives are do show them 
+            # on the default search page (dataset) or on one other search page
+            search_all_type = config.get('ckan.search.show_all_types', 'false')
+            search_all = False            
+
+            try:
+                #If the "type" is set to True or False, convert to bool
+                search_all = asbool(search_all_type)
+            #Otherwise we treat as a string representing a type
+            except ValueError:
+                if package_type and package_type == search_all_type:
+                    search_all = True
+                
+            if not package_type:
+                package_type = 'dataset'
+ 
+            if not search_all:
                 # Only show datasets of this particular type
                 fq += ' +dataset_type:{type}'.format(type=package_type)
-            else:
-                # Unless changed via config options or url parameter,
-                # don't show non standard dataset types on the default
-                # search page
-                if not asbool(
-                        config.get('ckan.search.show_all_types', 'False')) \
-                        and not request.params.get('_show_all_types', False):
-                    fq += ' +dataset_type:dataset'
 
             facets = OrderedDict()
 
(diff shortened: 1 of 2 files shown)


wardi:
@mattfullerton how about we make the "show all" dataset type configurable, instead of hard-coded to /dataset? e.g. you could have your /dataset for datasets /doc for documents and /all for all the types.

mattfullerton:
@wardi That certainly seems better than my original plan where I was then "losing" the dataset noun. In my case I'm now fairly persuaded that having to create an extra dataset type just to achieve the full listing is not the way to go. But of course, there's no reason to force the user/developer to use the dataset listing to get the "list all" functionality. I see some options here, and want to understand which you prefer, before writing more code:

1. Stick with just the current option `ckan.search.show_all_types`, allowing options of False (never do it), True (do it for datasets) or any dataset type (so we'd be redefining the option a bit)
2. Leave the current option as it is, and add a new one `ckan.search.show_all_on_type` which would potentially override the first, and show all types of dataset on its search page

3.  And then, almost Independent of those first two - allow the change I submitted to work - override the default functionality and show all types, on _any_ search page

OR

4. Allow the override functionality but only on a given type of dataset specified by `ckan.search.show_all_on_type` - i.e. when `ckan.search.show_all_types` is False, `ckan.search.show_all_on_type` has no effect until that URL option is given

Valid options are 1&3, or 2&3 or 2&4

By the way, with the "all" option - I still see a problem/can of worms with noun choice - so many parts of the template code expect a single, singular noun. We would start getting things like "No alls found" or "No all datasets found".

wardi:
My preference is your option 1 or 2. I'm strongly against taking a HTTP GET parameter to override the site owner's choice w.r.t. mixing types. Types dictate the fields and facets available, and assuming that one types' search templates will accommodate all dataset types is sure to fail in unpredictable ways.

Maybe we should support white-listing types to be included in a given search page, but this is better done as a new function on IDatasetForm than in configuration. Maybe just use IDatasetForm and deprecate the show_all_types option since it's too rigid.

Instead of "all" someone might choose "record" or something similarly generic. Theme plugins can override translations to provide text that makes sense for their sites too.

As a separate PR it'd be nice to make the default search type used in places that now direct to `/dataset` configurable too.

mattfullerton:
I see where you're coming from with the GET parameter being user defined and the option being site owner defined. As far as I can tell though, the facets cope just fine with no datasets containing the fields in question. The point about a given search template failing is valid, but with option (4) the site owner would have control over which type is allowed to exhibit this kind of functionality. In the unlikely event that I've persuaded you, let me know, otherwise read on 🙂 

I think the quickest and easiest is to take option 1. Once I have that, I can use a noun like "record" for these type-agnostic searches. With being limited to having the search-all control only on dataset datasets, I was having to search for a new noun for datasets as I was going to have to move them to another type where they could be searched in isolation.

mattfullerton:
@wardi I've implemented option 1 in the commit; happy to squash if need be

TkTech:
> As a separate PR it'd be nice to make the default search type used in places that now direct to /dataset configurable too.

This would be very desirable, I've had to hack this in on more than one site where "Item" is the only type we want to display, and want to hide the meta-types.

mattfullerton:
@wardi Tests are passing now

ckan/controllers/package.py
```diff
@@ -224,15 +224,25 @@ def pager_url(q=None, page=None):
...
-            if package_type and package_type != 'dataset':
+            # Unless changed via config options, don't show other dataset
+            # types any search page. Potential alternatives are do show them
+            # on the default search page (dataset) or on one other search page
+            search_all_type = config.get('ckan.search.show_all_types', 'false')
+            search_all = False
+
+            try:
+                # If the "type" is set to True or False, convert to bool
+                search_all = asbool(search_all_type)
```
wardi: won't this show all types on all types if show_all_types = true? We should maintain the current behaviour of only showing all on the 'dataset' type

mattfullerton:
@wardi Sorry this has taken so long - please re-review; I've tested it thoroughly

wardi:
@mattfullerton looks great, thanks!

PR #3810, opened 2017-09-11
Shrink datastore_create response size

wardi:
when passing records to datastore_create, the same records are all passed back in the response. There's no reason I can see for this and there are no records returned from datastore_upsert. Remove the records from the response for speed and consistency.


smotornyuk:
Good point. We had run into this few times as well:)

PR #3816, opened 2017-09-15
Resource Queries (CREATE TABLE AS in datastore)

wardi:
from https://github.com/ckan/ideas-and-roadmap/issues/200

Add the option to build a resource from an SQL query instead of an uploaded or linked file, internally implemented as PostgreSQL materialized views

q1

Queries are the same SQL that users can send to `datastore_search_sql`

q2

This will solve several problems:

- make existing visualization extensions more useful. The problem with the low usability of the existing viz extensions is that vizzes in general are meant to work against aggregated data, not raw observation-level data
- allow us to do joins across related packages/resources
- allow us to publish public views of private datasets with private columns removed
- and make it all performant as the queries are "materialized"

This feature can be thought of as a performance/usability improvement over what users can do today with `datastore_search_sql` and `datastore_create`. A query in a resource may only be modified if the *user requesting the change has access to make the same query with datastore_search_sql*. The same is true for refreshing the underlying data: the user requesting the refresh must have permission to run the same query.

This way we're not introducing any new ways for less privileged users to expose data that they shouldn't have access to.

Setting up triggers to automatically refresh the underlying views and/or adding jobs for this purpose are sensible next steps but not included in this initial PR.


wardi:
This PR includes two other PRs not yet merged into master, so they should be reviewed first: #3865 and #2562

wardi:
I'm not 100% sure about my javascript changes here. There was no obvious way to extend the url/upload javascript for queries from the datastore extension so that change was made directly in core.

This PR is working and should have all access and validation checks in place, it just needs some tests for the front-end and datastore API changes.

wardi:
In this version I've replaced materialized views with the simpler "CREATE TABLE AS" datastore feature. Snapshots of data are created unconnected from their parent resources. This lets us freely delete or modify the parents unlike when we use materialized views. It's simpler and easier for users to understand.

I've also added a new `ckan.datastore.resource_query_enabled` parameter to disable the feature if a site administrator so chooses. `ckan.datastore.sqlsearch` must be enabled or this feature is automatically disabled.

Documentation of resource queries and create_table_as_sql parameter is included. Tests are coming soon, but feature is complete and ready to be reviewed.

amercader:
Not in 2.8

ckanext/datastore/set_permissions.sql
```diff
@@ -55,17 +55,20 @@ CREATE OR REPLACE VIEW "_table_metadata" AS
         substr(md5(dependee.relname || COALESCE(dependent.relname, '')), 0, 17) AS "_id",
```
amercader: Can you summarize what the changes in the permissions mean here? It's a bit hard to follow.

Will all existing DataStore users need to re-run this against their existing database?

doc/maintaining/configuration.rst
```diff
@@ -270,6 +270,23 @@ Default value:  ``True``
...
+.. _ckan.datastore.resource_query.enabled:
+
+ckan.datastore.resource_query.enabled
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Example::
+
+ ckan.datastore.resource_query.enabled = False
+
+Default value:  ``True``
```
amercader: I'm not sure how I feel about having this feature enabled by default. It is definitely an advance feature and I think most users will be confused by it. My vote is for defaulting it to False.

ckan/templates/macros/form.html
```diff
@@ -439,6 +438,19 @@
...
-    {% if is_upload_enabled %}</div>{% endif %}
+  {% if query %}
+    {% call input_block('query', label=_('Query'), error='', classes=['control-full', 'editor']) %}
+      <textarea id="field-query" name="query" placeholder='SELECT * FROM "23af..."' rows="5" cols="20" class="form-control">{{ data.get('query', '') | empty_and_escape }}</textarea>
+      <span class="editor-info-block">{% trans %}Create a snapshot of data
+        from other DataStore tables using an SQL Query. You may re-run this
+        query at any time from the resource preview screen.<br/>
+        <b>Note:</b> This query will run
+        with your permissions but the data will be published and visible
+        along with other data in this dataset.{% endtrans %}
```
amercader: Does this mean that considering

* Dataset A, public
* Dataset B, private
* User X, with admin permissions on Dataset B

If user X is creating a resource query in dataset A that includes data from a resource in dataset B, the data from that new resource query will be publicly accessible under dataset A?

If so there's a potential for users exposing private data without realizing. The message should reflect this better, and there should be a warning in the documentation (`datastore.rst`)

doc/maintaining/configuration.rst
wardi: I'm fine with that.

ckan/templates/macros/form.html
wardi: Agreed, can you suggest some better text?

A query could be used to publish from a private table while removing all PII information, so this isn't always a drawback. Also there's nothing new being allowed here that couldn't already be done by the same user with `datastore_search_sql` and `datastore_upsert`.

amercader: > Also there's nothing new being allowed here that couldn't already be done by the same user with datastore_search_sql and datastore_upsert

True, but this feature makes it much easier to do, and so more prone to accidents.

What about something like:

> The query will be run using your permissions but the data will be published using this dataset visibility settings, regardless of the permissions on other resources used. Keep this in mind when querying private datasets.

wardi: Sounds much better!

jqnatividad:
Hi @wardi , just wondering if the core team is still considering merging this feature.

If not, up to what version does it work? 

Thanks!

wardi:
@jqnatividad I believe so, it just needs a few small changes @amercader requested, conflicts resolved and python 3 compatibility.

jqnatividad:
Thanks @wardi, that's good to know!  In the meantime, what older version of CKAN do you recommend to use this with if I want to experiment with it?

rufuspollock:
Just as a question - is there a need for this to be in core vs an extension? Being in an extension would reduce the burden on the core team for maintenance and make it a choice for people to add?

wardi:
@rufuspollock The code would be much larger as an extension than as a core change and would also require some new interfaces in core. Regardless there's no active development on this feature right now, so it will likely continue to sit here until someone has time to pick it up.

rufuspollock:
@wardi thanks for the context - just for my understanding (as i understand this stuff better) could you give a bit of info on:

* Why it would be larger
* What interfaces would be needed

Many thanks in advance - and don't worry if you don't have time 🙏

tino097:
@wardi this PR was promising, what is left to be done, is there anything that me or @TomeCirun could help to get finalized and maybe merged

wardi:
@tino097 the code is pretty old now, so lots of conflicts to clean up and updates for python3. It could use the new way of extending the data field on resources, so that's an improvement and would actually be easier to do in an extension now. I think the underlying API should still be in the datastore extension since datastore has a way to swap out back-ends now, but if you build the UI in an extension it would be easier for someone to replace with a fancy query-builder or something if they like.

tino097:
Thanks @wardi 
Closing as old PR

PR #3821, opened 2017-09-21
add download option from dump functionality

tino097:
Enables users to choose to what format resource will be downloaded

### Features:

- [ ] includes API changes


ckan/templates/package/resource_read.html
```diff
@@ -43,6 +44,24 @@
                     <i class="fa fa-arrow-circle-o-down"></i> {{ _('Download') }}
                   {% endif %}
                 </a>
+                {%if res.url_type == 'upload' and res.datastore_active %}
```
amercader: Don't check for `res.url_type == 'upload'` you can have stuff in the DataStore that hasn't been uploaded

wardi:
No need for the upload check and make sure this is inside a block so it can be overridden

TkTech:
Better yet , can we get away with no hardcoding at all? There should be an action method for DataStore to advertise supported formats.

ckan/templates/package/resource_read.html
```diff
@@ -43,6 +44,25 @@
                     <i class="fa fa-arrow-circle-o-down"></i> {{ _('Download') }}
                   {% endif %}
                 </a>
+                 {% block donwload_resource_button %}
```
wardi: download_resource_button

ckan/templates/package/resource_read.html
```diff
@@ -43,6 +44,25 @@
...
+                  {%if res.datastore_active %}
+                <button class="btn btn-primary dropdown-toggle" data-toggle="dropdown">
+                    <span class="caret"></span>
+                  </button>
+                <ul class="dropdown-menu">
+                  <li>
+                    <a href="{{ h.url_for(controller='ckanext.datastore.controller:DatastoreController', action='dump', resource_id=res.id) }}"
+                      target="_blank"><span>CSV</span></a>
```
wardi: Maybe default to bom=True for CSV and TSV so that Excel can load them properly

wardi:
@TkTech Having a block that can be overridden is our best option for now. That lets sites define whichever options make sense for them. Also the list of formats datastore supports exporting isn't extensible with a plugin interface yet.

PR #3822, opened 2017-09-21
Added Ubuntu 16.04 instructions and host ports

GabrielNicolasAvellaneda:


doc/maintaining/installing/install-from-package.rst
```diff
@@ -20,9 +20,21 @@ importing data to CKAN's :doc:`/maintaining/datastore`.
 1. Install the CKAN package
 ---------------------------
 
-On your Ubuntu 14.04 or 12.04 system, open a terminal and run these commands to install
+On your Ubuntu system, open a terminal and run these commands to install
 CKAN:
 
+#. Host ports requirements
```
amercader: Is this really needed? I We are also using Postgres and Redis, which would use a port as well, and DataPusher if its installed. While it might be useful to some, I'm not sure about having this so prominently in the install instructions.

GabrielNicolasAvellaneda: I added those ports because during installation I've already apache setup at port 80 (I've also nginx running on another port), and this will not work and I think that probably this could happen with other people. So, I had to take a look at the nginx configuration of ckan to known that actually it was expecting Apache on 8080. About postgresql, I didn't add it because by default it is using a pipes instead of a tcp port. About Redis, I think that I should add it too.

amercader: @GabrielNicolasAvellaneda fair enough, can you add Postgres and Redis but move the table to the introduction section above, to not mix it with the actual instructions.

amercader:
Thanks @GabrielNicolasAvellaneda, I made the changes myself and merged this

PR #3835, opened 2017-09-22
[#3834] Keep root path in new user url

### LINKED ISSUE #3834: Registering a new account redirects to an unprefixed url
2.7.0 and master

Upon registering a new account, the user should be redirected to a url starting with the root path

The url to which a new user is redirected should include the root path

On an installation that uses `ckan.root_path`, register a new account.

alexbrault:
Fixes #3834

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -258,7 +258,7 @@ class UserController(base.BaseController):
         if not c.user:
             # log the user in programatically
             set_repoze_user(data_dict['name'])
-            h.redirect_to(controller='user', action='me', __ckan_no_root=True)
+            h.redirect_to(controller='user', action='me')
         else:
             # #1799 User has managed to register whilst logged in - warn user
             # they are not re-logged in as new user.


amercader:
@wardi this reverts a change made in a recent PR that has been backported https://github.com/ckan/ckan/pull/3749. Should we backport this as well?

wardi:
@amercader does it? the `__ckan_no_root=True` part that is being removed was added in 4936c7c2ceb0987f80446ae69128281757f8e69e

IIUC this fix is required because we removed some url_for magic that we used to have when passed a plain string

amercader:
@wardi sorry, you are right this does not revert any change. Too many redirects in the user controller.

But I stand corrected as removing `__ckan_no_root=True` in other redirects was part of #3749 so IIUC this should be backported as well for consistency in all user redirects. Does this make sense?

amercader:
I'm pretty sure this needs to go into 2.6 and 2.7 as #3749 was backported there so I'll backport it too. @wardi @Zharktas let me know if you think otherwise

wardi:
@amercader ok, I understand now. Sounds good.

Zharktas:
Looks like that commit removed the last one of the "ckan_no_root" magics, at least according to the search https://github.com/ckan/ckan/search?utf8=%E2%9C%93&q=__ckan_no_root&type=, the logic should be removed from the helper too since it just complicates unnecessarily?

amercader:
@Zharktas absolutely

PR #3839, opened 2017-09-27
Fix setup.py classifiers syntax

amercader:
Was preventing the package from being uploaded to PyPI


wardi:
looks uncontroversial

PR #3841, opened 2017-09-28
fixed negative date vaue when creating a data request or updating a p…

precious-ijege:
### What does this PR do?
Fix the negative date value being displayed when a user creates a new datarequest and also when a user updates his/her profile.

The image below shows when a user successfully creates a data request.
image

The image below shows when a user successfully updates his/her profile and then checks via the activity stream
image

### DIFF
--- a/ckan/lib/formatters.py
+++ b/ckan/lib/formatters.py
@@ -99,7 +99,7 @@ def localised_nice_date(datetime_, show_date=False, with_hours=False):
         return months
 
     if not show_date:
-        now = datetime.datetime.utcnow()
+        now = datetime.datetime.now()
         if datetime_.tzinfo is not None:
             now = now.replace(tzinfo=datetime_.tzinfo)
         else:
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -120,8 +120,7 @@ def _datestamp_to_datetime(datetime_):
 
     # all dates are considered UTC internally,
     # change output if `ckan.display_timezone` is available
-    datetime_ = datetime_.replace(tzinfo=pytz.utc)
-    datetime_ = datetime_.astimezone(get_display_timezone())
+    datetime_ = datetime_.replace(tzinfo=get_display_timezone())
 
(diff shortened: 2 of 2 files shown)


smotornyuk:
@andela-pijege , thanks for your PR, but i suggest slightly different solution
```diff
diff --git a/ckan/lib/formatters.py b/ckan/lib/formatters.py
index f731144c6..607459738 100644
--- a/ckan/lib/formatters.py
+++ b/ckan/lib/formatters.py
@@ -99,9 +99,9 @@ def localised_nice_date(datetime_, show_date=False, with_hours=False):
         return months
 
     if not show_date:
-        now = datetime.datetime.utcnow()
+        now = datetime.datetime.now(pytz.utc)
         if datetime_.tzinfo is not None:
-            now = now.replace(tzinfo=datetime_.tzinfo)
+            now = now.replace(tzinfo=pytz.utc)
         else:
             now = now.replace(tzinfo=pytz.utc)
             datetime_ = datetime_.replace(tzinfo=pytz.utc)
diff --git a/ckan/model/activity.py b/ckan/model/activity.py
index cfd6d02e2..fdfa52660 100644
--- a/ckan/model/activity.py
+++ b/ckan/model/activity.py
@@ -40,7 +40,7 @@ class Activity(domain_object.DomainObject):
     def __init__(self, user_id, object_id, revision_id, activity_type,
             data=None):
         self.id = _types.make_uuid()
-        self.timestamp = datetime.datetime.now()
+        self.timestamp = datetime.datetime.utcnow()
         self.user_id = user_id
         self.object_id = object_id
         self.revision_id = revision_id
```

Previously we had a lot of troubles because of using tz aware datetime objects. Failed tests, problems in case of server migration, etc. So generally we decided to use UTC time in wherever it possible. Actually, this problem was fixed in latest 2.7 release and my suggestion is based on it. Maybe it even be better for your application to switch to 2.7?

andela-pijege:
@smotornyuk I would check out the latest version like you said. Thanks a lot for your response.

PR #3843, opened 2017-09-29
[#3798] Keep root_path out of site_logo until needed

### LINKED ISSUE #3798: Site logo is broken when site got root_path and WSGIScriptAlias
### CKAN Version
Master(2.8), 2.7

### Expected behaviour
When sysadmin adds external link to image or upload an image for site.logo, logo is displayed.

### Actual behaviour
When sysadmin adds external link to image or upload an image for site.logo, the url that is generated is broken. For upload/link it adds extra /data and the end of the url.

### Steps to reproduce the issue:
1. Setup CKAN with master or 2.7 version branch through apache.wsgi, add WSGIScriptAlias /data at conf and root_path /data/{{LANG}} at config.
2. As sysadmin go to configuration page.
3. Try to upload or add an external link to an image.
4. Save and you can see, that the url is being broken.

Shouldn't there be also an extra check "if url_path.startswith('http')" before the actual url is generated, so it wont add the root_path?

It seems that doc http://docs.ckan.org/en/ckan-2.7.0/sysadmin-guide.html for site.logo is not updated. There is no mention, that now it can be uploaded.
Also the logo can't be added as it done in 2.5, when we add "/base/images/ckan-logo.ong" path in the input. If we try to do this in link, it think that we uploaded it. Is it a bug?

alexbrault:
Fixes #3798 

### Proposed fixes:

For external urls, `url_for_static_or_external` would prefix it with the root_path. Instead, it now skips the last bit of processing if the url is already a fully-qualified url.

For site logo uploads, since the header template passes the value through `url_for_static_or_external` when rendering, I removed the call to `url_for_static` which would have added prefix twice

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -438,7 +438,9 @@ def url_for_static_or_external(*args, **kw):
     if kw.get('qualified', False):
         kw['protocol'], kw['host'] = get_site_protocol_and_host()
     my_url = _routes_default_url_for(*args, **kw)
-    return _local_url(my_url, locale='default', **kw)
+    if not is_url(my_url):
+        my_url = _local_url(my_url, locale='default', **kw)
+    return my_url
 
 
 @core_helper
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -1278,7 +1278,7 @@ def config_option_update(context, data_dict):
                 and not value.startswith('/'):
             image_path = 'uploads/admin/'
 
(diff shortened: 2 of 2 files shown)


smotornyuk:
@alexbrault, it seems, that your fix do not cover all cases - [thist test](https://github.com/ckan/ckan/blob/master/ckan/tests/lib/test_helpers.py#L77-L83) fails

kmbn:

PR #3854, opened 2017-10-06
Fixes #3796 and an issue with some deployments of CKAN losing Action API

### LINKED ISSUE #3796: Incorrect API link on /dataset
master

Text at bottom of dataset listing (search.html template) should include a working API link

A link of the form https://beta.ckan.org/packages?ver=%2F3 is generated
(https://beta.ckan.org/dataset)

This line is no longer working, I suspect because of the recent work to move the API over to Flask(?):
https://github.com/ckan/ckan/blob/master/ckan/templates/package/search.html#L52

mattfullerton:
This PR fixes an issue with some deployments of CKAN losing the Action API. The patch from @amercader resolves an issue reported on ckan-dev and IRC where sometimes flask routes aren't getting loaded and the action API is lost as a result. Incidentally, it fixes #3796 :-D

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -355,7 +355,9 @@ def _register_core_blueprints(app):
     def is_blueprint(mm):
         return isinstance(mm, Blueprint)
 
-    for loader, name, _ in pkgutil.iter_modules(['ckan/views'], 'ckan.views.'):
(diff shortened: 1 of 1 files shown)


amercader:
To give more context, the old code made assumptions about the environment in which the ckan package was installed. This didn't always work, for example in Docker deployments. The new code uses absolute paths to look for blueprint modules.

PR #3855, opened 2017-10-06
Change CORS header keys and values to string instead of unicode

ljupchokotev:
The CORS header keys and values being unicode raised problems when using uWSGI with CKAN. uWSGI expects the header keys to be string instead of unicode which results in a TypeError. 

### Features:

- [ ] includes API changes


amercader:
This makes our test that checks for string literals fail. Rather than remove the `u''` bit, can you replace it by a `b''`? Check the docs on this: http://docs.ckan.org/en/ckan-2.7.2/contributing/unicode.html#string-literals

cc @torfsen

amercader:
Thanks @ljupchokotev!

PR #3858, opened 2017-10-09
Fix of `Add Dataset` button on search page.

smotornyuk:
[#3643] assumes, that `add dataset` route is named as `dataset_new`.
I just updated route's name, as it not used anywhere else and it won't
break anything.


ggcatu:
What happend here? Why added , rever, add ?

Issue #3833, opened 2017-09-22
Templates from extensions are not registered in the Flask app


amercader:
This only affects maser. The best way to reproduce it is to click on the Data API button on a resource with DataStore. This uses the API to look for a template and return it rendered, and right now triggers a 500 error, as the template comes from the datastore extension:

```
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/views/api.py", line 448, in snippet
    return render(snippet_path, extra_vars=dict(request.args))
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/lib/base.py", line 187, in render
    return flask_render_template(template_name, **extra_vars)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/flask/templating.py", line 133, in render_template
    return _render(ctx.app.jinja_env.get_or_select_template(template_name_or_list),
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/jinja2/environment.py", line 869, in get_or_select_template
    return self.get_template(template_name_or_list, parent, globals)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/jinja2/environment.py", line 830, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/jinja2/environment.py", line 804, in _load_template
    template = self.loader.load(self, name, globals)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/jinja2/loaders.py", line 113, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/flask/templating.py", line 57, in get_source
    return self._get_source_fast(environment, template)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/flask/templating.py", line 85, in _get_source_fast
    raise TemplateNotFound(template)
TemplateNotFound: ajax_snippets/api_info.html
```

Perhaps we need to add something to the [`add_template_directory()`](https://github.com/ckan/ckan/blob/master/ckan/plugins/toolkit.py#L326) function that plugins use

cc @tino097 @smotornyuk

tino097:
@amercader im not able to reproduce the error

wardi:
[...]

PR #3860, opened 2017-10-09
[#3833] Extend Flask template dirs

smotornyuk:
Add additional template directories to template_loader during
creation of Flask app stack. Previously we put list of additional
dirs into `config['pylons.app_globals'].template_paths`, but
as we are migrating to Flask I moved this value to
`config['computed_template_paths']`. After this, inside
`make_flask_app`, default `jinja_loader` replaced with ChoiceLoader,
that will try to load templates from either default dirs or custom
ones.

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -57,13 +57,10 @@ def make_flask_stack(conf, **app_conf):
     app.template_folder = os.path.join(root, 'templates')
     app.app_ctx_globals_class = CKAN_AppCtxGlobals
     app.url_rule_class = CKAN_Rule
-    extra_paths = config.get('computed_template_paths')
-
-    if extra_paths:
-        app.jinja_loader = ChoiceLoader([
-            app.jinja_loader,
-            CkanextTemplateLoader()
-        ])
+    app.jinja_loader = ChoiceLoader([
+        app.jinja_loader,
+        CkanextTemplateLoader()
+    ])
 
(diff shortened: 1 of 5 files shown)


ckan/config/middleware/flask_app.py
```diff
@@ -54,6 +57,13 @@ def make_flask_stack(conf, **app_conf):
     app.template_folder = os.path.join(root, 'templates')
     app.app_ctx_globals_class = CKAN_AppCtxGlobals
     app.url_rule_class = CKAN_Rule
+    extra_paths = config.get('computed_template_paths')
+
+    if extra_paths:
```
amercader: Shouldn't we use the `ChoiceLoader` always regardless of whether we have `computed_template_paths` in the config at this point?
Otherwise if you start CKAN without any plugin loaded at all you won't detect templates loaded at runtime, like in tests. I know this is impossible right now as we have "core plugins", but wondering if it makes more sense to future proof this.

amercader:
This looks fantastic @smotornyuk, just one tiny question

ckan/config/middleware/flask_app.py
smotornyuk: Yeah, it has sense. I prefer to keep things as simple as possible, but here we can allow permanent ChoiceLoader even if there is only one tpl source

smotornyuk:
Updated.

amercader:
Thanks @smotornyuk! Good to have green tests again

PR #3862, opened 2017-10-10
Flask helpers update

tino097:
This PR comes out from #3762 and it contains changes to helpers and adding the flask routes to config['routes.named_routes']

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -731,6 +731,30 @@ def nav_link(text, *args, **kwargs):
     :param condition: if ``False`` then no link is returned
 
     '''
+    if is_flask_request():
+        return nav_link_flask(text, *args, **kwargs)
+    else:
+        return nav_link_pylons(text, *args, **kwargs)
+
+
+def nav_link_flask(text, *args, **kwargs):
+    if len(args) > 1:
+        raise Exception('Too many unnamed parameters supplied')
+    controller, action = request.url_rule.endpoint.split('.')
+    if args:
+        kwargs['controller'] = controller or None
+    named_route = kwargs.pop('named_route', '')
+    if kwargs.pop('condition', True):
+        if named_route:
+            link = _link_to(text, named_route, **kwargs)
+        else:
+            link = _link_to(text, **kwargs)
+    else:
+        link = ''
+    return link
+
+
+def nav_link_pylons(text, *args, **kwargs):
     if len(args) > 1:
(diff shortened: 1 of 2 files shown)


ckan/lib/helpers.py
```diff
@@ -923,11 +960,12 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
...
-    if c.search_facets_limits and limit is None:
-        limit = c.search_facets_limits.get(facet)
-    # zero treated as infinite for hysterical raisins
-    if limit is not None and limit > 0:
-        return facets[:limit]
+    if hasattr(c, 'search_facets_limits'):
+        if c.search_facets_limits and limit is None:
+            limit = c.search_facets_limits.get(facet)
+        # zero treated as infinite for hysterical raisins
+        if limit is not None and limit > 0:
```
amercader: These two lines should not be inside the `if hasattr(c, 'search_facets_limits'):` check, they need one level of indentation less

ckan/lib/helpers.py
```diff
@@ -685,12 +685,25 @@ def are_there_flash_messages():
...
+
+
+def _link_active_pylons(kwargs):
     highlight_actions = kwargs.get('highlight_actions',
                                    kwargs.get('action', '')).split()
     return (c.controller == kwargs.get('controller')
             and c.action in highlight_actions)
 
 
+def _link_active_flask(kwargs):
```
amercader: We should avoid using Pylons terms on Flask functions. `controller` and `action` are Pylons terms that not apply to Flask, so in the context of a Flask function let's use:

```python
blueprint, endpoint = request.url_rule.endpoint.split('.')
```

---

Of course we need to check against kwargs.get('controller') and kwargs.get('action') to keep backwards compatibility, but we also need to introduce the new Flask-based equivalent to be ready to switch calls to `nav_link()` to the Flask syntax, eg something like:

```python
if kwargs.get('endpoint') == request.url_rule.endpoint:
```
So we can go from this

```
{{ h.nav_link(_('Login'), controller='user', action='login') }}
```

to this


```
{{ h.nav_link(_('Login'), endpoint='user.login') }}
```

I suggest we do this in one go when we update all `url_for`, `nav_link` etc calls after all controllers have been migrated, to avoid having function calls in the two syntaxes.

tino097:
@amercader looks good now

amercader:
Thanks @tino097!

PR #3864, opened 2017-10-10
[#3863] Change permission_labels type to string in schema.xml

### LINKED ISSUE #3863: Permission labels are indexed by type text in SOLR
2.7, master

Organization Admin should see organizations private datasets even when SOLR indexing has been changed to accommodate different languages. The permission label should not be index by language rules.

If indexing rules are changed for example to use [EdgeNGram](https://wiki.apache.org/solr/AnalyzersTokenizersTokenFilters#solr.EdgeNGramFilterFactory), the permission label id's are indexed by these rules and organization admin, editor or member can not see organizations private datasets.

Apply EdgeNGram to solr indexing.

Create organization.
Create user.
Add user to organization.
Create private Dataset to organization.

### Possible fix
Use type textgen instead of text in the default schema.xml.

Zharktas:
Fixes #3863 

### Proposed fixes:
Changed permission_labels type to string as proposed in https://github.com/ckan/ckan/issues/3863#issuecomment-335463338

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/solr/schema.xml
+++ b/ckan/config/solr/schema.xml
@@ -112,7 +112,7 @@ schema. In this case the version should be set to the next CKAN version number.
     <field name="organization" type="string" indexed="true" stored="true" multiValued="false"/>
 
     <field name="capacity" type="string" indexed="true" stored="true" multiValued="false"/>
-    <field name="permission_labels" type="text" indexed="true" stored="false" multiValued="true"/>
+    <field name="permission_labels" type="string" indexed="true" stored="false" multiValued="true"/>
 
     <field name="res_name" type="textgen" indexed="true" stored="true" multiValued="true" />
     <field name="res_description" type="textgen" indexed="true" stored="true" multiValued="true"/>


wardi:
@Zharktas I forgot, this is going to need a version bump on the solr schema, mind taking care of that too?

Zharktas:
@wardi created PR for that too :)

wardi:
@amercader does this rise to the level of needing a backport? It's a solr schema change which isn't usually something we do in a patch release so I'm leaning towards "no"

amercader:
We want 2.7.x to be supported for a long time so maybe it's worth thinking about it. If I understand correctly backporting this would be safe in that if people do a normal patch upgrade (ie just upgrading the sources but not restarting solr (jetty/tomcat) the search won't be affected. But if they have this problem the just need to restart jetty/tomcat and reindex right? Or is there something else that might break existing searches?

wardi:
As discussed at meeting: backport plan is to include the new (2.8) schema in 2.7 but modify code to allow both versions 2.7 and 2.8 so that sites aren't forced to upgrade and reindex if they're not affected by the issue

PR #3865, opened 2017-10-11
Allow IValidator to override existing validators

wardi:
There are cases where overriding a built-in or plugin-defined validator is much simpler than redefining a one or more whole schemas.

This change allows IValidator plugins to override existing validators, following the established "first plugin wins" pattern. Validators are usually simple functions and they can be created with a number of different function signatures, so we don't attempt to "chain" validators like we do when overriding action methods. one validator wins and is the only one that will be called.

Current behaviour is to raise a NameConflict error, this will no longer happen.

The core schema.py imports validator functions directly, should we change schema.py to call get_validator() as part of the same PR for consistency, or make this only affect schemas defined in plugins or with ckanext-scheming?

### DIFF
--- a/ckan/lib/navl/validators.py
+++ b/ckan/lib/navl/validators.py
@@ -2,7 +2,7 @@
 
 import ckan.lib.navl.dictization_functions as df
 
-from ckan.common import _
+from ckan.common import _, json
 
 missing = df.missing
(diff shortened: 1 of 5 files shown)


wardi:
@amercader thanks!

PR #3871, opened 2017-10-13
cli: fix passing of concurrency arg for celery

sorki:
Allows running only one worker, required as a work-around for race conditions that can happen if e.g. two jobs update extras.


amercader:
Thanks @sorki, this was obviously wrong. Just FYI, the celery based tasks are deprecated in favour of a new background jobs framework http://docs.ckan.org/en/ckan-2.7.2/maintaining/background-tasks.html

sorki:
Right, I'm planning to migrate when time permits. Thanks.

PR #3881, opened 2017-10-23
Add support for postgresql 10

groundrace:
Updated the psycopg2 drivers to support latest Postgresql version

Fixes #

### Features:

- [ ] includes API changes


amercader:
@groundrace  thanks for this. We need to ensure that the actual code also works against postgres 10. Looking at the failures it looks like there are some JSON related ones on the DataStore, so perhaps this needs to be investigated a bit more. Can you check what needs updating?

PR #3882, opened 2017-10-25
Fixed hardcoded 'Dataset'

ggcatu:
Fixes #

Fixing some hardcoded 'Dataset', 

Breadcrumb, Search form, New 'package'

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -181,7 +181,7 @@ def make_map():
     map.redirect('/package/{url:.*}', '/dataset/{url}')
 
     with SubMapper(map, controller='package') as m:
-        m.connect('search', '/dataset', action='search',
+        m.connect('search', action='search',
                   highlight_actions='index search')
         m.connect('dataset_new', '/dataset/new', action='new')
         m.connect('/dataset/{action}',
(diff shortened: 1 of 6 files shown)


wardi:
We just discussed this at our dev meeting and our feeling is we need a new IDatasetForm method that can return the user-visible strings for the type defined. 
e.g.
```python
def get_strings(self):
    return {
        'Datasets': _('Records'),
        'View datasets': _('View records'),
        'Create dataset': _('Create record'),
        'Search datasets': _('Search records'),
        'Add dataset': _('Add record'),
    }
```
This way the strings can be picked up by tools that generate .po files (from core and extensions) instead of requiring them to be manually entered.

This affects the changes merged in #3643 as well. I know this expands the scope of the PR, @ggcatu @mattfullerton do either of you have time to implement this change?

ggcatu:
@wardi and how should #3885 be correctly adressed in the future, as I think that is working only in kind of a hacky way, on the other hand maybe implement a way that a single IDataSetForm can manage multiple schemas.

Correct me if I'm wrong but ckanext-scheming forces you to use the same templates for all the schemas you define?

wardi:
You're right,
```python
get_strings(self, dataset_type):
```
would be better, then scheming and any other IDatasetForms can do their thing properly

wardi:
@ggcatu you can override the template that scheming uses for its pages to split off to per-type ones at the template level. That's a separate issue though, happy to discuss on the ckanext-scheming repo

ggcatu:
Facets filtering was still sending to /dataset.... instead of /datatype

kmbn:

PR #3885, opened 2017-10-26
checking by named_route

ggcatu:
Fixes #

### Proposed fixes:
Fix the active menu, when different datatypes.
Would only work for Pylons, 
Don't know if this would be the best way, and not sure how to port it to work with flask, since I think there are not named routes?

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -912,9 +912,7 @@ def _make_menu_item(menu_item, title, **kw):
         raise Exception('menu item `%s` cannot be found' % menu_item)
     item = copy.copy(_menu_items[menu_item])
     item.update(kw)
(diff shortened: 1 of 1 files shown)


ckan/lib/helpers.py
```diff
@@ -909,7 +912,9 @@ def _make_menu_item(menu_item, title, **kw):
         raise Exception('menu item `%s` cannot be found' % menu_item)
     item = copy.copy(_menu_items[menu_item])
     item.update(kw)
+    item.update({'named_route': menu_item})
     active = _link_active(item)
```
wardi: Instead of modifying and reusing a dict this is safer:

```python
active = _link_active(dict(item, named_route=menu_item))
```

kmbn:

PR #3908, opened 2017-11-06
change solr schema.xml ref in dockerfile to local

### LINKED ISSUE #3894: Move docker-compose.yml to project root
By moving the docker-compose file to the root of the project we can make the whole repo available in the build context. 

The first benefit is that it solves the issue when using a version of CKAN different from master, at the moment it's loading the `schema.xml` via a github url from the master branch, which will raise a schema conflict error. (https://github.com/ckan/ckan/issues/2255#issuecomment-336146034). By moving the docker-compose file up to the root we can reference the schema file present in `ckan/config/solr/`, ensuring it's referencing the correct version. 

Secondly it would also open up the opportunity to mount the local files into the container, which would be very helpful for local development.

Only `docker-compose.yml` and `.env` would need to be moved to the root, the remaining files can be referenced to in `contrib/docker`. 

Please let me know your thoughts or if I've missed anything. Happy to put together a PR if this seems acceptable.

gerbyzation:
Fixes #3894
Changes the Dockerfile to use the local schema.xml located in `ckan/config/solr/` instead of the one in github master branch, avoiding version conflicts.

### Features:

- [ ] includes API changes

### DIFF
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -60,6 +60,11 @@ services:
   solr:
     container_name: solr
     image: ckan/solr:latest
+    # To build locally, remove the image reference above and uncommend this:
+    # build:
+    #   context: ../../
+    #   dockerfile: contrib/docker/solr/Dockerfile
+
 
   redis:
     container_name: redis
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -59,11 +59,9 @@ services:
 
   solr:
     container_name: solr
-    image: ckan/solr:latest
-    # To build locally, remove the image reference above and uncommend this:
-    # build:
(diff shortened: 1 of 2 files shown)


contrib/docker/docker-compose.yml
```diff
@@ -60,6 +60,11 @@ services:
   solr:
     container_name: solr
     image: ckan/solr:latest
+    # To build locally, remove the image reference above and uncommend this:
+    # build:
+    #   context: ../../
+    #   dockerfile: contrib/docker/solr/Dockerfile
```
wardi: why not make this the default?

gerbyzation: I don't see a reason why not

wardi:
This should go on dev-2.6 and dev-2.7 so we can start building those images again

PR #3909, opened 2017-11-07
Allow overriding celery list params

philpep:
The condition 'if key in LIST_PARAMS' was never satisfied because config
keys are lower case and LIST_PARAMS is upper case.

This allow overriding the CELERY_IMPORTS, ADMINS and ROUTES in ckan
config.

Also overriding list params should extend existings values (i.e. for
CELERY_IMPORTS)


smotornyuk:
@philpep  thanks. I'll merge it.

PR #3915, opened 2017-11-10
restrict anonymous access to user information, with ckan.public_user_…

fanjinfei:
…details (default true)

Fixes #
By default CKAN exposes all account names, full names, org membership of its users.

### Proposed fixes:
A new configuration option ckan.public_user_details with a default true value to maintain the current behavior.

When ckan.public_user_details=false, anonymous user access to user_show, user_list, and group member will be denied.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/logic/auth/get.py
+++ b/ckan/logic/auth/get.py
@@ -3,9 +3,11 @@
 import ckan.logic as logic
 import ckan.authz as authz
 from ckan.common import _
+from ckan.common import config
 from ckan.logic.auth import (get_package_object, get_group_object,
-                            get_resource_object)
+                            get_resource_object, restrict_anon)
 from ckan.lib.plugins import get_permission_labels
+from paste.deploy.converters import asbool
 
 
 def sysadmin(context, data_dict):
@@ -88,7 +90,10 @@ def tag_list(context, data_dict):
 
 def user_list(context, data_dict):
     # Users list is visible by default
-    return {'success': True}
+    if not asbool(config.get('ckan.public_user_details', True)):
+        return restrict_anon(context)
+    else:
+        return {'success': True}
 
 def package_relationships_list(context, data_dict):
     user = context.get('user')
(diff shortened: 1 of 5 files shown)


ckan/lib/helpers.py
```diff
@@ -2595,3 +2595,16 @@ def sanitize_id(id_):
...
     return str(uuid.UUID(id_))
+
+
+@core_helper
+def public_user_details():
+    return converters.asbool(config.get('ckan.public_user_details', True))
+
+
+@core_helper
+def restrict_anon(context):
```
wardi: this doesn't make sense as a template helper, please make this a simple function in logic/auth/get.py instead

ckan/lib/helpers.py
```diff
@@ -2595,3 +2595,16 @@ def sanitize_id(id_):
     ValueError.
     '''
     return str(uuid.UUID(id_))
+
+
+@core_helper
+def public_user_details():
```
wardi: is this used in the templates anywhere?

ckan/logic/auth/get.py
```diff
@@ -161,7 +166,9 @@ def group_show(context, data_dict):
     user = context.get('user')
     group = get_group_object(context, data_dict)
     if group.state == 'active':
-        return {'success': True}
+        if h.public_user_details() or (not asbool(data_dict.get('include_users', False)) and (
+		data_dict.get('object_type', None) != 'user')):
```
wardi: check your editor settings, we shouldn't be using real tabs for indentation

wardi:
@fanjinfei this looks pretty good to me so far. Would you add a reference to the new config option in the configuration section of the docs and the ckan/config/deployment.ini_tmpl file?

fanjinfei:
@wardi Which file is for `the configuration section of the docs`?

wardi:
https://github.com/ckan/ckan/blob/master/doc/maintaining/configuration.rst

wardi:
It's probably worth mentioning that the new setting should only be used if user registration is also disabled (or someone could just register an account to get access to all the users)

PR #3936, opened 2017-11-23
Get rid of debconf issues during ckan image build

SergK:
* Get rid of:
          TERM is not set, so the dialog frontend is not usable
  while building ckan docker image
* Clean up apt mirrors


SergK:
Please check related #3937 fix for CI

PR #3937, opened 2017-11-23
Fix CI by changing build docker context for solr

SergK:
Fix is related to #3908


SergK:
This fix is also related to #3938 since we try to use cached images, patch from #3908 passed CI

PR #3948, opened 2017-12-05
Use wheel & update pip in Dockerfile

ad-m:
### Proposed fixes:

Added ``wheel`` to system in Dockerfile and disable cache in pip.

### Features:

Adding ``wheel`` allows faster python package installation. We use Python libraries, which are partially written in C ++. They require a compilation that takes a while, unless we use the compiled libraries provided by the wheel. There is many compatible, compiled libraries than to [manylinux](https://github.com/pypa/manylinux).

We disable cache in pip to avoid preparing wheel packages. Besides, the cache does not make sense, because the rebuilt container will not have access to it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get -q -y update \
         python-dev \
         python-pip \
         python-virtualenv \
+        python-wheel \
         libpq-dev \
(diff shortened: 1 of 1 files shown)


smotornyuk:
Thanks, @ad-m

PR #3951, opened 2017-12-07
[docker] remove config and home volumes

gerbyzation:
I would like to open up discussion on the current docker-compose volume setup. At the moment the ckan service is using three volumes:

https://github.com/ckan/ckan/blob/6f311192c5a89319772cd1e4e970c89e28aaf405/contrib/docker/docker-compose.yml#L36-L39

The main point is on `ckan-home`, which includes the virtualenv and source code. This is somewhat deceiving when updating dependencies or code. 

Multiple people have reported issues where code or dependencies have been updated, used `docker-compose down` and rebuild, but changes don't propagate because the volume still contains previous iteration. To resolve this `docker volume prune` needs to be run when rebuilding. 

Sharing the config & source code through a volume is an anti-pattern when used with containers, as it prevents an immutable approach for containers. I'm also unaware of any uses or benefits this might have. If this is to keep several instances 'in sync' this probably would be better addressed through orchestration. For that reason I propose removing the home and config volumes. 

Please let me know if anyone has any thoughts on this.

### DIFF
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -3,8 +3,6 @@
 version: "3"
 
 volumes:
-  ckan_config:
-  ckan_home:
   ckan_storage:
   pg_data:
 
@@ -34,8 +32,6 @@ services:
       - DS_RO_PASS=${DATASTORE_READONLY_PASSWORD}
 
     volumes:
-      - ckan_config:/etc/ckan
-      - ckan_home:/usr/lib/ckan
       - ckan_storage:/var/lib/ckan
 
   datapusher:


Vanuan:
I agree. The problem stems from the fact that CKAN plugins are not microservices, i.e. plugin's code is called directly without any RPC.
The problem worsens by the fact that plugin's code needs to reside inside ckan's home.

This means that to install any plugin you have to rebuild an image. So each CKAN deployment is unique. Thus it becomes meaningless to share Docker configuration for CKAN.

Here's mine: https://github.com/Vanuan/ckan-base/blob/master/README.md

gerbyzation:
@Vanuan You're right that we can't create a docker configuration that works for everyone everywhere, but we can create a base configuration that will get a base instance of CKAN running. When needed this can then be extended for someone's specific purpose.

To solve the configuration issue it could be worth considering including chanext-envvars or a script that will modify the config based on environment variables. That way settings could be customised in the .env file. This would have to be a seperate issue from this though.

Vanuan:
Well, I don't see how it's different. When you configure CKAN, e.g. change logo, you want the changes to be persisted. Since CKAN stores changes to the same place it stores sources, the whole source directory needs to be put in a volume. Otherwise you need to reconfigure CKAN on each restart.

gerbyzation:
@Vanuan it's the difference between treating the container as an advanced virtual environment or a self-contained and immutable executable/process. Volumes are there for data that needs to persist past the lifetime of a single container instance. Using it for source code is staying with old ops patterns but using a new technology. 

Can you explain why the entire source would have to be in a volume? What stops you from putting the config and user provided files in a volume? Or copying these files into the container when building?

I'm not sure yet what the most appropriate solution is for installing plugins.
One option (and at the moment I think the most obvious and straight forward) would be to have a base ckan image that can be customised and configured by using it as a parent image. 

Kubernetes also has an interesting pattern with init containers, but I don't know of any way something similar can be achieved in compose.

Vanuan:
@gerbyzation I agree that this is the problem. But I don't see how removing volumes from the example compose file solves the issue.

Have you tried running it with your changes, changing the logo and then re-creating the container?

gerbyzation:
@Vanuan Okay, I'll just put together an example in the next few days and show what I mean.

gerbyzation:
@Vanuan Okay, have a look at this: https://github.com/thedataplace/ckan/tree/improved-docker. 

I have renamed the dockerfile to `Dockerfile.base`. The changes:

1. it installs `ckanext-envvars`
2. It includes a base `config.ini` file (found in repo root) that loads the  `ckanext-envvars` plugin.
3. I've commented out the `ckan.site_url` build argument as we want a generic image. This is still required as env variable though to run the image.

If you build this and tag it with `ckan:2.7.2`, you can then use this as a base image in another Dockerfile:

https://github.com/thedataplace/ckan/blob/6433aa5e8970db17bf7dc0fdb9e78adab6df5b0e/Dockerfile#L1

If you add the following lines to the docker file

```
RUN mkdir -p /usr/lib/ckan/public
ADD rabbit.png /usr/lib/ckan/public
```
 and these as env vars in docker compose (can also be done in `.env` if you like)

```
- CKAN___EXTRA_PUBLIC_PATHS=/usr/lib/ckan/public
- CKAN__SITE_LOGO=/rabbit.png
```
and every time you rebuild you'll have a rabbit logo.

Installing additional plugins can be done in this Dockerfile as well. Another benefit is that the configuration is all declared, so it can be kept in version control. 

I realise I should probably put this forward as a proposal, as just removing the volumes is only half a solution.  As it's expanding the scope of this PR a bit it would seem better to do this in a new issue/PR.

But please let me know what you think

kmbn:

PR #3952, opened 2017-12-08
Upgrade Bleach to latest

davidread:
Bleach hasn't changed a great deal, judging by the [changelog](https://github.com/mozilla/bleach/blob/master/CHANGES). CKAN only uses it in render_markdown(). There's a comprehensive test for it: test_format_text which passes, so I'm happy.


smotornyuk:
I tested it and didn't found any incompabilities.
Thanks, @davidread

PR #3956, opened 2017-12-15
Remove hardcoded list with langs, and take it form ini

Krokop:
Fixes #
If you set to the ckan.locale_default something that is not in [LANGS](https://github.com/ckan/ckan/blob/master/ckanext/multilingual/plugin.py#L10) list, it'll raise error in [here](https://github.com/ckan/ckan/blob/master/ckanext/multilingual/plugin.py#L235)

I remove hardcode and get list of langs from ckan.ini#locale_order


e0ne:
+1. LGTM

smotornyuk:
Yeah, i think it must be there. Thanks, @Krokop

PR #3958, opened 2017-12-15
Check that resource is not None in "datastore_delete" action

dlax:
"resource" comes from a model.Resource.get() above which may return
None.

I have observed this while running "datastore_delete" action on no longer existing resources (cleanup up the datastore).


smotornyuk:
Thanks, @dlax

PR #3962, opened 2017-12-18
Updating `setup.py` to include package data

### LINKED ISSUE #3961: Package resources not included when installing from source in non-editable mode.
Known affects:
 - CKAN 2.7.2; Python 2.7.14; Ubuntu 16.04 & 17.10; 
   ```
   $ virtualenv --version
   15.1.0
   $ virtualenv --no-site-packages test-venv
   <snip>
   $ ./test-venv/bin/pip freeze --all 
   pip==9.0.1
   setuptools==38.2.4
   wheel==0.30.0
    ```

### Expected behaviour
Checking out the repository and installing it(without the editable flag) should install the required package resources like `ckan/migration/migrate.cfg`.

## Actual behaviour
Installing the package without the editable flag does not include the package resources meaning all JavaScript, CSS, templates and the migration config noted above are not included.

This makes the package non-functional. 

For me the problem arose because `ckan/migration/migrate.cfg` did not exist in my install directory, and hence the database could not be created. There are numerous other files listed in `MANIFEST.in` which are also required for CKAN to run which are not included.

Following the [installing from source](http://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html) instructions, omitting the `-e` flag from step **c**

## Resolution.

The solution to this is to add `include_package_data=True` to the `setup` call in `setup.py`

leesdolphin:
When installing from sources in non-editable mode, include the package resources like JavaScript, CSS, templates, and migration configuration.

Fixes #3961

### Proposed fixes:

Include packange data in non-editable builds.

This does not update the installation documentation; however it would be wise to advise against editable installations for non-development environments.

### Features:

- [ ] includes API changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -216,6 +216,7 @@ setup(
     keywords='data packaging component tool server',
     long_description=__long_description__,
     zip_safe=False,
+    include_package_data=True,
     packages=find_packages(exclude=['ez_setup']),
     namespace_packages=['ckanext', 'ckanext.stats'],
     message_extractors={


e0ne:
Looks like a reasonable fix

Issue #3929, opened 2017-11-18
CKAN's state table clashes with PostGIS generated TIGER state table


philtweir:
### CKAN Version if known (or site URL)

2.8.0a HEAD [e219c2] (although observed earlier tags)

### Please describe the expected behaviour

On running `docker-compose up` in a freshly cloned repo, all services start and database migrations run successfully.

### Please describe the actual behaviour

`ckan` container exits during start-up with errors (from `db` and `ckan` containers):

```
ckan          | Traceback (most recent call last):
ckan          |   File "/usr/local/bin/ckan-paster", line 11, in <module>
ckan          |     sys.exit(run())
ckan          |   File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/paste/script/command.py", line 102, in run
ckan          |     invoke(command, command_name, options, args[1:])
ckan          |   File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/paste/script/command.py", line 141, in invoke
ckan          |     exit_code = runner.run(args)
ckan          |   File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/paste/script/command.py", line 236, in run
ckan          |     result = self.command()
ckan          |   File "/usr/lib/ckan/venv/src/ckan/ckan/lib/cli.py", line 356, in command
ckan          |     model.repo.init_db()
ckan          |   File "/usr/lib/ckan/venv/src/ckan/ckan/model/__init__.py", line 187, in init_db
db            | ERROR:  column "id" referenced in foreign key constraint does not exist
db            | STATEMENT:  
db            |         CREATE TABLE package (
db            |                 id SERIAL NOT NULL, 
db            |                 name VARCHAR(100) NOT NULL, 
db            |                 title TEXT, 
db            |                 version VARCHAR(100), 
db            |                 url TEXT, 
db            |                 download_url TEXT, 
db            |                 notes TEXT, 
db            |                 license_id INTEGER, 
ckan          |     self.upgrade_db()
ckan          |   File "/usr/lib/ckan/venv/src/ckan/ckan/model/__init__.py", line 284, in upgrade_db
ckan          |     mig.upgrade(self.metadata.bind, self.migrate_repository, version=version)
ckan          |   File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/migrate/versioning/api.py", line 186, in upgrade
db            |                 state_id INTEGER, 
db            |                 revision_id INTEGER, 
[...]

PR #3965, opened 2017-12-20
[#3929] Allow custom metadata for database connection, to define schema for migrations

philtweir:
When PostGIS starts up using the mdillon/postgis Docker image, the TIGER
target schema ends up with a table named "state", conflicting with CKAN's
state table.

This change makes the schema used by CKAN during migration explicit, and
configurable via an environment variable, CKAN_MIGRATIONS_TARGET_SCHEMA.

Fixes #

### Proposed fixes:

- Add a `CkanMetaData` class to provide a singly-settable default target schema
- Use this in migrations all migrations
- Use a `CKAN_MIGRATIONS_TARGET_SCHEMA` to make this settable via environment variables
- Update tests to ensure environment variable works
- Add environment variable into documentation

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -47,9 +47,6 @@ who.log_file = %(cache_dir)s/who_log.ini
 ## Database Settings
 sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_default
 
-## Migration Settings
-ckan.migrations.target_schema = public
-
 #ckan.datastore.write_url = postgresql://ckan_default:pass@localhost/datastore_default
 #ckan.datastore.read_url = postgresql://datastore_default:pass@localhost/datastore_default
 
@@ -96,6 +93,9 @@ ckan.site_id = default
 # ckan.cors.origin_allow_all = true
 # cors.origin_whitelist is a space separated list of allowed domains.
 # ckan.cors.origin_whitelist = http://example1.com http://example2.com
+
+## Migration Settings
+ckan.migrations.target_schema = public
 
 
 ## Plugins Settings
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -47,6 +47,9 @@ who.log_file = %(cache_dir)s/who_log.ini
 ## Database Settings
 sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_default
 
+## Migration Settings
+ckan.migrations.target_schema = public
+
 #ckan.datastore.write_url = postgresql://ckan_default:pass@localhost/datastore_default
 #ckan.datastore.read_url = postgresql://datastore_default:pass@localhost/datastore_default
 
--- a/ckan/model/metadata.py
+++ b/ckan/model/metadata.py
@@ -9,7 +9,7 @@ from sqlalchemy import MetaData
 from ckan.common import config as ckan_config
 
 
-class CkanMetaData(MetaData):
+class CkanMigrationMetaData(MetaData):
     """CKAN custom metadata
 
     Allow for setting metadata from CKAN's own configuration.
@@ -21,4 +21,4 @@ class CkanMetaData(MetaData):
         if u'schema' not in kwargs and schema:
             kwargs[u'schema'] = schema
 
-        super(CkanMetaData, self).__init__(*args, **kwargs)
+        super(CkanMigrationMetaData, self).__init__(*args, **kwargs)
--- a/ckan/model/metadata.py
+++ b/ckan/model/metadata.py
@@ -1,7 +1,13 @@
 # encoding: utf-8
 
+"""Wrapper for SqlAchemy MetaData
+
+Keep a consistent set of metadata for CKAN.
+"""
+
 from sqlalchemy import MetaData
 from ckan.common import config as ckan_config
+
 
 class CkanMetaData(MetaData):
     """CKAN custom metadata
--- a/ckan/model/metadata.py
+++ b/ckan/model/metadata.py
@@ -1,0 +1,18 @@
+# encoding: utf-8
+
+from sqlalchemy import MetaData
+from ckan.common import config as ckan_config
+
+class CkanMetaData(MetaData):
+    """CKAN custom metadata
+
+    Allow for setting metadata from CKAN's own configuration.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        schema = ckan_config.get('ckan.migrations.target_schema')
+        if 'schema' not in kwargs and schema:
+            kwargs['schema'] = schema
+
+        super(CkanMetaData, self).__init__(*args, **kwargs)
--- a/ckan/model/metadata.py
+++ b/ckan/model/metadata.py
@@ -17,8 +17,8 @@ class CkanMetaData(MetaData):
     """
 
     def __init__(self, *args, **kwargs):
-        schema = ckan_config.get('ckan.migrations.target_schema')
-        if 'schema' not in kwargs and schema:
-            kwargs['schema'] = schema
+        schema = ckan_config.get(u'ckan.migrations.target_schema')
+        if u'schema' not in kwargs and schema:
+            kwargs[u'schema'] = schema
 
         super(CkanMetaData, self).__init__(*args, **kwargs)
--- a/ckan/tests/config/test_environment.py
+++ b/ckan/tests/config/test_environment.py
@@ -22,7 +22,6 @@ class TestUpdateConfig(h.FunctionalTestBase):
 
     ENV_VAR_LIST = [
         ('CKAN_SQLALCHEMY_URL', 'postgresql://mynewsqlurl/'),
-        ('CKAN_MIGRATIONS_TARGET_SCHEMA', 'public2'),
         ('CKAN_DATASTORE_WRITE_URL', 'http://mynewdbwriteurl/'),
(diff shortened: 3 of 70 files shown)


ckan/tests/config/test_environment.py
```diff
@@ -81,6 +84,8 @@ def test_update_config_db_url_precedence(self):
 
         nosetools.assert_equal(config['sqlalchemy.url'],
                                'postgresql://mynewsqlurl/')
+        nosetools.assert_equal(config['ckan.migrations.target_schema'],
```
ad-m:
```
  File "/usr/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/home/ubuntu/virtualenvs/venv-system/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/ubuntu/ckan/ckan/tests/config/test_environment.py", line 87, in test_update_config_db_url_precedence
    nosetools.assert_equal(config['ckan.migrations.target_schema'],
  File "/home/ubuntu/ckan/ckan/common.py", line 82, in __getitem__
    return self.store[key]
KeyError: 'ckan.migrations.target_schema'
```
See https://circleci.com/gh/ckan/ckan/3473#tests/containers/2 .

ckan/model/metadata.py
```diff
@@ -0,0 +1,18 @@
+# encoding: utf-8
+
+from sqlalchemy import MetaData
+from ckan.common import config as ckan_config
+
```
ad-m: Add newline here.

philtweir:
As this is an unusual component to change, I am not 100% sure if there anything particular that should be added in terms of tests or docs other than the last few commits above - is there anything I should look at? Thanks!

amercader:
@philtweir Thanks for the thorough PR, specially the tests.

I'm not sure why you didn't include the`MetaData()` instantiation [here](https://github.com/flaxandteal/ckan/blob/0dea2b3ac914cfaedbf462cc5e2d610162b7e66d/ckan/model/meta.py#L153), which is pretty important as it is the one used by most of the application. 

If I understood correctly, if you want to migrate to a custom schema you'll need to use CKAN on that schema from then on, so you'll need all metadata instances to point to the custom schema right? If this feature is about allowing CKAN to work on a different schema than `public` I'll be happy for it to be merged (although I doubt many users would need for it). But if it's just a workaround for a particular deployment setup maybe is pushing it. I doubt you'll need the TIGER functions from PostGIS to run a standard CKAN so that's what should be moved to a separate schema or not installed at all.

Even if you get PostGIS and CKAN installed on separate schemas you'll need extra steps to enable PostGIS in the CKAN DB right?

philtweir:
Thanks @amercader - yes, that's a good point.

Originally, (which is why the pull request was delayed), I had a look at making the schema used by CKAN globally settable in a single PR. However, there are a number of locations where raw SQL is used, which (I think) would require setting the schema in or before the call, therefore sending an escaped environment variable (CKAN_MIGRATIONS_TARGET_SCHEMA) to the DB embedded in an SQL string - I thought that may have broader implications than the team would be up for to solve the current issue. Given that, until the raw SQL can be replaced, it seemed better to leave the general running of the application to be internally consistent, rather than have some parts overridden. In general though, that'd be the aim.

However, the issue in the migrations is a little different to what you describe - TIGER does install to a separate schema, but when SqlAlchemy is migrating it checks table existence before creating a table, which passes if the `state` relation name exists in _any_ schema (afaict, unless a schema is explicitly passed). `CkanMetaData` makes the schema explicit to ensure it only searches that one relevant schema, usually "public". However, rather than hardcoding "public" as your default schema, this env lets you provide an alternative. If I understand correctly, this also means that if you are using any other tables in other schema which happen to clash with the CKAN names, this should prevent the migrations skipping creating them.

That means that CKAN's default Docker setup, which is used in `contrib/docker`, no longer works (unless it was run before that docker-compose.yml refactor, so not on new installs). As far as I am aware, there isn't anything else you need to do to get PostGIS on the CKAN database - PostGIS is mostly PostgreSQL extension(s), so afaiu (at least in that mdillon image) is automatically available for all schemas.

philtweir:
@amercader  If this approach is too broad, I can focus down just to either (a) hard-code the "public" schema into the CkanMetaData or (b) remove CkanMetaData entirely, and just hard-code the "public" schema per migration (and maybe come back to the PG schema selection at a later date) -- just as this is a bug which leads to a wrongly laid out DB on first docker-compose start-up, and people have had difficulty undoing that (as in #3929), cutting down to a minimal, more merge-friendly solution would at least solve the primary problem quickly?

ckan/config/deployment.ini_tmpl
```diff
@@ -47,6 +47,9 @@ who.log_file = %(cache_dir)s/who_log.ini
 ## Database Settings
 sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_default
 
+## Migration Settings
```
davidread: This is an obscure setting. How about sticking this more round L99?

davidread:
LGTM. Let's get this merged :) (And backported for the most recent release, I suggest)

ckan/config/deployment.ini_tmpl
philtweir: Yep, that sounds logical - just pushed a commit

davidread:
@amercader do you have further comment on this, or may I merge?

philtweir:
One thing I don't understand is why there is the security/snyk new dependency issues - I can't view the report, but can't see anything in the commits that would have changed/updated them - any ideas?

amercader:
@philtweir @davidread @wardi Sorry it took me a while to get back to this. As tempting as the one-liner in #4016 looks my initial reservations were along the lines of your latest comments in #3929. This is certainly noisy but perhaps safer than #4016 and even if I still don't think that you can use a different schema than public after the migration, this opens he door to implementing it in the future.

So I'm happy to merge this but I'd ask to rename `CkanMetaData` to `CkanMigrationMetaData` to make clear that it only applies to that.

@wardi I know we didn't decide this in the meeting but does this sound good to you?

philtweir:
Sure, will update that now. Thanks @amercader

wardi:
As I said at the meeting, this feels like half a feature. We're adding new a configuration option (and environment variable!) but IIUC it's not actually safe to change these from the default value of 'public'. I still vote for the one line change because this one invites users to try something that we know doesn't work.

philtweir:
The rationale is that there are two reasons to change the setting:
1. installing CKAN to a non-default schema -- think this is what @wardi is referring to, which can confirm is not implemented here
2. installing CKAN when the default schema is other than `public` -- afaict, a one-liner would prevent CKAN installing in this case, whereas this approach still allows that use case - however, I'd agree that would cause very few people an issue. Happy to test non-public default schema against a fresh install if requested, but would be keen to have a provisional go-ahead before spending too much longer on modifications to this PR.

Equally, there's a couple of levels that still accommodate (2) that are smaller (as [above](https://github.com/ckan/ckan/pull/3965#issuecomment-363419963)):

1. One-liner, as in #4016, as `public` is widely used as the default name, and `state` is the only table currently breaking
2. One-liner in each migration, as `public` is standard (but touches all migration files)
3. Config setting with _no_ env for `state` table - touches only a couple of files
4. Config setting with no env used in all migration files
5. Env read into config setting and used in all migration files (this PR)

davidread:
Some good ideas here. I've implemented option 3:
https://github.com/ckan/ckan/pull/4026
In the unlikely event people have changed their default schema, then they have this get-out option. And it's a minimal change for everyone else.
Is this a reasonable compromise?

philtweir:
If you want to use PostGIS, (which seems equivalent to supporting queriable geospatial data storage by plugins?), that seems to me to be the standard base image, but the flip-side is that that'll be evolving over time too.

Strictly, in terms of use-cases, I would have one, as we are working with a parallel app that is to be deployed (in Docker) with CKAN alongside, but we'll just be running two DBs instead - in any case, I wouldn't assume that behaviour should be officially supported. My point above is more about hard-coding a postgres setting (regardless of how the user is using the DB), which CKAN is (afaict) currently independent from - I think that was @amercader's original concern in the issue, which makes sense to me also. Ironically, I think it's relevant primiarly when _not_ using default Docker images :)

@davidread cool, cheers - my only suggestion would be that it be consistent between the migrations, rather than special behaviour (effectively) for the `state` table, i.e. option 4, as this problem was a right pain to diagnose when it unexpectedly came up - happy to provide a commit to #4026 (or just remove the env-reading-line from this one) However, from @wardi's comment, sounds like that may be worse?

davidread:
I'm easy on all these options tbh.

amercader:
Thanks for all the comments. I feel like we need to move on so unless someone strongly objects I'm going to go ahead and merge the compromise option #4026, which is almost @wardi's second favourite option.

I still think that the CKAN database should only contain CKAN tables and that the PostGIS image should not install the Tiger extension, but if that's the current situation in our docker setup so be it.

Aside: the underlying issue here is that PostGIS should not be a requirement. It is only required for ckanext-spatial for historical reasons and because I have never found the time to make it optional (ckan/ckanext-spatial#195). Hopefully when that is implemented we can rely on the standard Postgres image and not the PostGIS one.

@philtweir apologies for all the hassle, it took me a while to wrap my head around this

philtweir:
Agree that the priority is getting set-up working again and think #4026 covers the main issues succinctly - certainly, I not realised PostGIS was being removed, so in that context, it certainly makes more sense. +1 from me to switch to #4026 - thanks for providing the compromise @davidread !

PR #3972, opened 2018-01-02
Revision UI removed & activity stream improved

### LINKED ISSUE #3484: Removal of deprecated revision controller.

davidread:
Fixes #3484

* Revision UI removed (as warned in CKAN 2.7.0 changelog):
  - `/revision/list` - List of which datasets changed. Not easy to page through 200 at a time. Not used or useful as far as we can tell.
  - `/revision/{id}` - Detail of a revision - displays timestamp and the dataset - no use. Problematic when there are billions of revisions you have billions of these pages that get indexed.
  - `/revision/edit/{id}` - to undelete a revision. No-one uses this method. Better ways to achieve this.
  - `/revision/diff/{id}` - diffs different versions of a dataset. Replaced by `/dataset/changes/`
* History page removed - `/dataset/{id}/history` now redirects to Activity Stream. The History page was not supported or linked to from anywhere (on default templates) but it was useful for admins. The functionality of seeing the past versions is fully replaced by the Activity Stream now. (The Atom feed at the same URL is also removed: `/dataset/{id}/history?format=atom`, again deprecated. But there is a feed of datasets ordered by metadata_modified at `/feed/dataset.atom`, and the changes are in the package_activity_list API which could be made an Atom feed if there is interest.)
* Activity Stream saves the full dictized dataset, rather than just the package table. Migration script: `migrate_package_activity.py` (discussed below)
* Activity Stream for a dataset now has 2 new links against each activity item: "View this version" to view the dataset as it was and "Changes" - the diff since the previous version:
  screen shot 2018-05-12 at 10 32 47
    The changes (diff) page looks like this:
        screenshot 2019-02-15 at 17 32 05
        ...
        screenshot 2019-02-15 at 17 32 23
    This isn't the clearest - changes are not highlighted - you have to look in the margin for a letter. But it's a pretty good start and can be improved.

    * NB The "View this version" and "Changes" pages can viewed by admins and optionally others too - depending on new option `ckan.auth.public_activity_stream_detail`. The default is False, so that sites upgrading don't suddenly splurge this detail without taking the conscious act of changing it, although we encourage it in the changelog. New sites will have it set to True by default, because it is added that way in `deployment.ini_templ`.
* Viewing old versions of datasets is done with a different parameter name. Uses `?activity_id=<id>` rather than `@<revision_id>` or `@<date>`. (This breaking change is ok because the only links to these were from the history page which was not linked to itself.)
* Activity Stream model is simplified - we use just Activity and remove ActivityDetail. Previously when you update a resource, it'd create an Activity('changed package') and also a connected ActivityDetail('resource changed'). This just seems too complex in the model. Now if you change just an extra field or a resource field, it'll simple say 'X updated the dataset Y', rather than 'X added the resource Z to the dataset Y'.
   Before:
   screenshot 2019-02-08 at 11 58 43
   After:
   screenshot 2019-02-08 at 11 58 53
   Since we store the pkg_dict on every commit, we could in future write logic that could analyze the diff to describe the change in more detail, like before. But it's worth it for the simpler model.
   * ActivityDetail is no longer used and is deprecated. It remains in the model for now, in case someone has some custom details they want to keep. But the user is asked to agree to deletion and then the table is emptied by when they run `migrate_package_activity.py`.
   * `activity_detail_list` logic function is removed
* New action functions for viewing an activity: `activity_show` and very similar: `activity_data_show` making command-line diffs possible e.g.:
```
diff <(ckanapi action activity_data_show id=484f01d7-d9c8-4a4d-bb42-00da6b8525f7 object_type=package -c $CKAN_INI) <(ckanapi action activity_data_show id=293a87ed-7ff9-4191-9d92-cb523763b8dd object_type=package -c $CKAN_INI)
15c15
<   "metadata_modified": "2018-01-02T16:29:44.571621",
---
>   "metadata_modified": "2018-01-02T16:29:44.834676",
80c80
<       "revision_id": "1e6a3a25-2099-4318-891c-a2dd78b173de",
---
>       "revision_id": "9e779436-fe43-400d-92ad-9f1583a60a60",
91,92c91,92
<   "revision_id": "9eddc9f4-a29b-4feb-aff6-bfcea87f2139",
<   "state": "draft",
---
>   "revision_id": "9e779436-fe43-400d-92ad-9f1583a60a60",
>   "state": "active",
```
* Diffing between package versions is also possible with a new action function `activity_diff`.
* `migrate_package_activity.py` - migrates activity stream to store the full package dict, using the package revision tables.
    * Package revision tables (package_revision/package_extra_revision/package_tag_revision/resource_revision) are not removed and are still being populated, but are not used for anything. (We keep them just in case: There is a chance that there are packages have been create/changed/deleted but there are not corresponding activity stream objects (due to disabling ckan.activity_streams_enabled, or for changes while running ckan before 1.6, or due to a bug certain changes are not recorded as activity). Regenerating activity stream objects for these would probably be possible from the revisions, so why not leave them.)
    * This does not migrate old versions of groups/organizations/member/package_relationship/system_info so those revision tables remain, to be tackled in future. Eventually we'll get rid of vdm.
    * This migration takes a while to run (could be hours for the biggest sites), and the effect is only noticeable to admins by default, so rather than include it in the normal `paster db upgrade` migrations, it is a standalone script. Ideally you download it and run it against any version of CKAN (I've tested it 2.3-2.8) before you upgrade to this version of CKAN. But if you don't, that is fine - you can upgrade to this or a future CKAN version, and run this migration if you want - the only downside of not migrating is that the "View this version" and "Changes" links will 404 until you do run the migration (and by default only admins can see Changes page anyway). (An earlier version of this PR prevented you upgrading until the migratino was done, but is no longer the case.)

I think we are agreed that it is ok to remove Revision UI and History page without prior deprecation because they are long unlinked and unsupported.

This PR removes a chunk of code to do with 'revisions'. However migrate_package_activity.py requires the revisions still in the model. A future PR (looking like https://github.com/ckan/ckan/pull/4664/files#diff-67c809480fe3487c995fb33d7d134d84) can remove the rest:
* tables: group_revision/group_extra_revision/member_revision/package_relationship_revision/system_info_revision - maybe we need a similar way to view changes to groups at least? maybe the rest can go.
* model code
* vdm

This PR follows on from https://github.com/ckan/ckan/pull/4627: Activity Stream is displayed with standard templates, rather than via action functions that returned HTML.

This PR took over from the original work on this issue that was #3485

Todo:
- [x] Go through review comments https://github.com/ckan/ckan/pull/3485
- [x] diffing - see if it is poss using the api and diff.
- [x] Check: _filter_activity_by_user commenting out
- [x] Remove ActivityDetail in a migration script
- [x] write test for migrate_package_activity.py (ensure it doesn't break in future)
- [x] move the revision stuff out of model_dictize code into migrate_revisions.py (including calling package_dictize with the (context['revision_id'] option)
- [x] config option for everyone to see diffs
- [x] test against old/real activity streams
- [x] Really old versions of CKAN included activity event types that no longer exist, and trying to render those activity events fails. We need to add a fallback template to render these without exception. This will be more important once this PR goes in since it will be much easier to add arbitrary event types. (TkTech: One important test we should have is to ensure activity doesn't fail to render when the database includes an activity type or objects no longer exist. There have been 4-5 issues about this on the mailing list from what I've seen.)
- [x] Rendering org/group events when the org/group has been deleted cause crashes on some templates.
- [x] check templates-bs2 are in sync

### Features:

- [x] includes API changes


amercader:
@davidread this is great!, are you still working on it?

davidread:
@amercader Yes :)

davidread:
I've gone down the path that by default *only admins see the old versions of datasets*. The concern I have is where existing CKAN sites would make the old versions much more visible, when the admin had previously assumed old versions were not released. They wouldn't wish to show old versions which may have incorrect metadata, or multiple draft versions, which may be embarrassing.

However the auth functions can be simply overridden, to allow anyone to see the old versions, if that is wanted.

@wardi and anyone else - it would be good to hear if you think this is right.

TkTech:
Leaving it here as a note, but one important test we should have is to ensure activity doesn't fail to render when the database includes an activity type or objects no longer exist. There have been 4-5 issues about this on the mailing list from what I've seen.

- Really old versions of CKAN included activity event types that no longer exist, and trying to render those activity events fails (do we just not render them? Do we show a generic event?). This will be more important once this PR goes in since it will be much easier to add arbitrary event types.
- Rendering org/group events when the org/group has been deleted cause crashes on some templates.

davidread:
Thanks @TkTech ! I'll add it to the list of things to look at.

davidread:
I'm wondering whether to go ahead with the 'simplification of the activity model' thing, which means getting rid of ActivityDetail and thus messages like "User X added a resource/extra/tag to Package Y" in favour of simply "User X updated Package Y". Actually the detail might be nice for users to see, but it's a bit of code to maintain. @TkTech took it out and I've taken that through, but I'm not totally convinced. Either I'd restore this code or I'll do a migration script to get rid of ActivityDetail objects.

davidread:
I'm tempted to write new tests. The legacy ones are a huge bit of work, checking lots and lots at once, covering so many edge cases I'd not considered.

wardi:
@davidread I agree only sysadmins seeing old versions is a good call. It's not trivial to determine who should be able to see what when talking about historical metadata. Should the permissions on the metadata at that time apply or the current ones? Do we need to check who was a member of each org during the time that metadata was active too? The answers will vary from site to site.

Getting rid of activity detail is fine with me. It can be reconstructed from the json versions we're storing if we want it back, and currently it assumes no customization of the dataset schemas.  Keep this change as simple as possible and we can re-build more advanced reporting later.

davidread:
Thanks for helping me resolve those @wardi. A couple more questions to come.

TkTech:
@davidread I originally removed ActivityDetail because it was a fairly large duplication of information, and my overriding goal with the original PR was to simplify things. Those kinds of messages can be easily done just by storing some extra fields on the activity dict and adding some template logic.

```if fields_modified == 1: data.update({'field_before': ..., 'field_after': ...})```

**Comment by davidread** [comment]
I've done some simplification to my changes here - reverted the activity queries without the extra joins, but it relies on this PR now: https://github.com/ckan/ckan/pull/4317

**Comment by bribroder** [comment]
Hey all, just wanted to comment on the visibility of past revisions--it seems reasonable that revisions can be made private and hidden, but public revision history could perhaps increase rather than decrease trust in the dataset, as a way of being transparent about how the dataset has changed over time--perhaps some bit of research used a previous revision and future researchers doing some kind of update or continuance would need to see how the dataset has changed.

Just wanted to point out the perspective that older revisions are not embarrassing, they're part of the history of the dataset. Dataverse, a similar project to CKAN, did a feature comparison with other datahub projects and actually highlights, as a positive, access to older revisions: https://docs.google.com/spreadsheets/d/1KptHzDHIdB3s1v5m1mMwphcwXhOVWdkRYdjEWW1dqrE/edit?ts=5b742653#gid=355072175

Thanks for all your work on CKAN!

**Comment by davidread** [comment]
@bribroder I agree. I think I have a solution:

* New installs of CKAN will have the changes public
* Those upgrading exsiting sites to this version of CKAN will only show the changes to admins, but they will be asked on upgrade to change this setting.

(I can achieve this by putting into the CKAN config template 'ckan.activity_streams_detail_is_public=True', but the default value (if you don't specify it, like those upgrading) is False)

**Comment by amercader** [comment]
I'm trying to add custom activities for a project and after all the nastiness I had to do I can wait for this to be ready! @davidread where are you at with this? is it in a stage when it can be reviewed?

File: ckan/views/dataset.py [hunk 1]
```diff
@@ -1026,116 +1061,55 @@ def activity(package_type, id):
     # TODO: remove
     g.pkg_dict = pkg_dict
     g.pkg = pkg
-    g.package_activity_stream = package_activity_stream
```
davidread: Normally we don't fiddle with context variables and globals because even this style is deprecated, we don't want it to break someone's customized template. However the `package/activity.html` template is fundamentally changed in this PR (`package_activity_stream` changes type from pre-rendered HTML to a list), necessitating changes to a customized template anyway.

ckan/controllers/package.py
```diff
@@ -1311,18 +1215,23 @@ def activity(self, id):
         data_dict = {'id': id}
         try:
             c.pkg_dict = get_action('package_show')(context, data_dict)
-            c.pkg = context['package']
-            c.package_activity_stream = get_action(
-                'package_activity_list_html')(
-                context, {'id': c.pkg_dict['id']})
+            package_activity_stream = get_action(
+                'package_activity_list')(
+                context, {'id': id})
```
davidread: Normally we don't fiddle with context variables and globals because even this style is deprecated, we don't want it to break someone's customized template. However the `package/activity.html` template is fundamentally changed in this PR (`package_activity_stream` changes type from pre-rendered HTML to a list), necessitating changes to a customized template anyway.

ckan/controllers/user.py
```diff
@@ -673,17 +667,20 @@ def dashboard(self, id=None, offset=0):
 
         c.followee_list = get_action('followee_list')(
             context, {'id': c.userobj.id, 'q': q})
+
         c.dashboard_activity_stream_context = self._get_dashboard_context(
             filter_type, filter_id, q)
-        c.dashboard_activity_stream = h.dashboard_activity_stream(
+        dashboard_activity_stream = h.dashboard_activity_stream(
```
davidread: Normally we don't fiddle with context variables and globals because even this style is deprecated, we don't want it to break someone's customized template. However the `user/dashboard.html` template is fundamentally changed in this PR (`dashboard_activity_stream` changes type from pre-rendered HTML to a list), necessitating changes to a customized template anyway.

davidread:
I'm in the closing stages of this - a month more tops. Note that it will remove the need for revisions/vdm in Datasets - we will be storing the complete package_dict on every change. Then we could decide whether to extend these activity stream changes to groups/organizations/member/package_relationship/system_info (which is not a huge task) or simply lose their history (maybe people need the audit trail or maybe they don't care too much - they would have had to poke around in the db to see it anyway). Then we can remove revisions/vdm.

amercader:
That is fantastic news @davidread. Can't wait to get this in the review process.

> we will be storing the complete package_dict on every change

Is this equivalent to the current revisions in terms of storge in the DB? Sounds a bit scary if you have a large metadata schema or lots of extras but maybe we are doing it now anyway. I guess that cleanup jobs (remove revisions older than x) can be also added in the future.

davidread:
@amercader So yes it is storing more. Hopefully I can test this on a large CKAN instance and get an idea of the scale of the change.

davidread:
> @TkTech wrote: One important test we should have is to ensure activity doesn't fail to render when the database includes an activity type or objects no longer exist. There have been 4-5 issues about this on the mailing list from what I've seen.

@TkTech Yes, so as you say, a custom activity will cause a TemplateNotFound exception in snippets/activity_stream.html trying to render`snippets/activities/{}.html`. I *could* render a fallback 'generic activity' saying "Bob did something called 'datastore changed' yesterday" but without any more detail, just to get it to render without exception. However it adds code complexity. Shouldn't a site owner upgrading their CKAN be forced to add a template for this in their extension? I'm trying to get my head round this.

wardi:
I like the fallback idea. Maybe update the snippet tag to take a list of templates and it will render the first it finds?

davidread:
@wardi Cool, fallback is now added

ckan/lib/base.py
```diff
@@ -76,17 +76,31 @@ def abort(status_code=None, detail='', headers=None, comment=None):
...
-                  % (template_name, output, template_name))
-    return literal(output)
+    for template_name in template_names:
+        try:
+            output = render(template_name, extra_vars=kw)
+            if config.get('debug'):
+                output = ('\n<!-- Snippet %s start -->\n%s\n<!-- Snippet %s end -->\n'
+                        % (template_name, output, template_name))
+            return literal(output)
+        except TemplateNotFound:
```
wardi: is there a way to make sure that the TemplateNotFound came from the first-referenced snippet and not some other template that snippet renders? If we catch both we could hide real template errors that should be raised

davidread: Sure, I've added that in now.

TkTech:
I believe where I was going with that (long time ago!) was plugins adding and removing activity types. If I temporarily disable a plug-in that added custom types of activity events, it shouldn't cause pages to fail. This would have been the case for open.canada which would have implemented access history events as activities.

davidread:
@TkTech Ah, that makes perfect sense now. So with these latest commits it'll render "bob changed datastore", rather than raising TemplateNotFound. 👍

davidread:
I did a migration on a test database to get a sense of the time taken and the change in database size (@amercader asked about the latter)

Timing: Migration took 49s for 236 datasets. If you scale that linearly to 900k datasets (europeandataportal.eu) then you're talking 1 hour. There's likely to be more activities per dataset though, so it would be more like several hours. Seems reasonable to want to do the migration while the site is up.

Size: size of the activity table went up about 20%, because every revision we now store resources, extras and tags on each package. However the overall size came down when we deleted the content of the activity_detail table (which is no longer needed), because that had a quirk that when you add a tag, it stored the dictized package as well in that table - this seems like a mistake.

Full details here: https://gist.github.com/davidread/0cb2442dde65222f0dd41e9b5700035f

If anyone else has a CKAN they could give it a try on, I'd most welcome getting more feedback.

davidread:
This PR's migration is intended to be downloaded and run **before** you upgrade CKAN - see https://github.com/ckan/ckan/wiki/Migrate-package-activity . However a small site that doesn't care about a bit of site down-time can upgrade to this code, run the migration and then bring the site up. That's because the model still has the revision stuff embedded in it. (In this PR we add [a check that this PR's migration is done before "paster db upgrade" will run](https://github.com/ckan/ckan/pull/3972/files#diff-c27306712b18b475d7662f0d5d5eab14R277)). 

Question: Should we remove the revision stuff from the model in this same release of CKAN? If we did, then anyone upgrading their code would find that the migration is not possible - they'd have to downgrade their CKAN again and run the migration first. I quite like this idea, to get it out all out of the way in the next CKAN release, with our eye on getting rid of revisions/pylons/python2 asap. What do you think @wardi ?

davidread:
This is ready to review now @wardi

amercader:
:tada: Amazing news @davidread! Thanks for your work on this.

Regarding the dropping of the revisions let's think about it, and see how it fits in the wider python 3 support effort. Personally, I can't wait to get rid of the revisioning but if it prevents people migrating maybe we should pause and see what the options are. Could we provide a separate extension that included vdm and the revisioning bits necessary to perform a migration so users that care about old revisions can import them? My sense is that this would be difficult but you know the code better than anyone now.

TkTech:
@amercader I am currently in the process of removing vdm including migrations as part of a contract with TBS.

ckan/config/deployment.ini_tmpl
```diff
@@ -72,6 +72,8 @@ ckan.auth.user_delete_organizations = true
 ckan.auth.create_user_via_api = false
 ckan.auth.create_user_via_web = true
 ckan.auth.roles_that_cascade_to_sub_groups = admin
+ckan.auth.public_user_details = true
+ckan.auth.public_activity_stream_detail = true
```
wardi: are these the defaults we want in the ini template?

CHANGELOG.rst
```diff
@@ -10,6 +10,26 @@ Changelog
 v.2.9.0 TBA
 ==================
 
+ * This version requires script 'migrate_package_activity.py' to be run
```
wardi: would it be possible to detect un-migrated activities and display them with a <needs migration> message instead of requiring the site to be offline while the migration is happening?

ckan/lib/dictization/model_dictize.py
```diff
@@ -696,19 +634,29 @@ def vocabulary_list_dictize(vocabulary_list, context):
...
+        # delete all the data apart from the title field on each data object,
+        # because that is needed to display it in the activity stream
+        for obj_key in activity_dict['data'].keys():
+            obj_data = activity_dict['data'][obj_key]
+            if isinstance(obj_data, dict):
+                for key in obj_data.keys():
+                    if key != 'title':
+                        del obj_data[key]
+            else:
+                del activity_dict['data'][obj_key]
```
wardi: You could replace the deepcopy and loops with
```python
activity_dict = {
    key: {'title': val['title']}
    for (key, val) in activity_dict.items() if 'title' in val}
```

ckan/logic/action/get.py
```diff
@@ -2546,7 +2547,9 @@ def package_activity_list(context, data_dict):
     '''
     # FIXME: Filter out activities whose subject or object the user is not
     # authorized to read.
-    _check_access('package_show', context, data_dict)
+    data_dict['include_data'] = False
+    include_hidden_activity = asbool(context.get('include_hidden_activity'))
```
wardi: Why not make `include_hidden_activity` part of the data_dict instead?

ckan/config/deployment.ini_tmpl
davidread: `ckan.auth.public_user_details` is unrelated to this PR but I noticed it was missing so added it for completeness, same value as the default.

`ckan.auth.public_activity_stream_detail` is true to encourage new ckan sites to have an open history. It's different to the default, so that old sites that upgrade have no change unless they consciously set this to true.

CHANGELOG.rst
davidread: The migration is designed to be run while the site is online. I've just re-worded this to make that clear.

Hopefully the site admin would practise a CKAN upgrade before doing it on live. They would get the [error message](https://github.com/ckan/ckan/blob/d253130526ea538e1fa49229e91f1d19d8711c6c/ckan/model/__init__.py#L284) when they did 'paster db upgrade' telling them to the this migration first, as explained [here](https://github.com/ckan/ckan/wiki/Migrate-package-activity#if-you-tried-to-upgrade-from-ckan-28-or-earlier-to-ckan-29-and-it-stopped-at-paster-db-upgrade). Hopefully that's clear.

ckan/logic/action/get.py
```diff
@@ -2588,7 +2595,9 @@ def group_activity_list(context, data_dict):
     '''
     # FIXME: Filter out activities whose subject or object the user is not
     # authorized to read.
-    _check_access('group_show', context, data_dict)
+    data_dict['include_data'] = False
```
wardi: this modifies the caller's data_dict, use `dict(data_dict, include_data=False)` in the call below instead

ckan/lib/dictization/model_dictize.py
davidread: that is just the ticket! :) I've put that in now.

ckan/logic/action/get.py
davidread: 'hidden' activity is that by site_user, such as harvests, scripts, which are not shown in the activity stream because they can be too numerous and aren't attributed to a person. This is not something this PR changes. (I guess you could argue they should all be exposed, but I didn't want to change anything unnecessarily.)

However we do need to get at these hidden activities in the migration, because these hidden activities do have Activity objects, and if a hidden Activity is followed be a non-hidden one and you look at the changes of that non-hidden Activity, then it does a diff with the hidden one (rather than the most recent non-hidden one), so it is important to store the package_dict in hidden Activity objects.

So because we don't want to start exposing these activities in the Activity Stream web page or API, it's added as an option on the context, not the data_dict.

ckan/logic/action/get.py
davidread: good point. done.

ckan/migration/migrate_package_activity.py
```diff
@@ -0,0 +1,261 @@
...
+        model.Session.commit()
+        print(u'  saved')
+    print(u'  This package\'s {} activities are migrated'.format(
+        len(package_activity_stream)))
+
+
+def wipe_activity_detail(delete_activity_detail):
+    from ckan import model
+    num_activity_detail_rows = \
+        model.Session.execute(u'SELECT count(*) FROM "activity_detail";') \
```
wardi: `SELECT count(*) from (SELECT 1 FROM "activity_detail" LIMIT 1) as t` to make this query faster

ckan/model/__init__.py
```diff
@@ -266,12 +267,29 @@ def upgrade_db(self, version=None):
...
+            num_unmigrated_dataset_activities = num_unmigrated(meta.engine)
+            if num_unmigrated_dataset_activities:
+                print('''
+    !!! ERROR !!!
+    You have {num_unmigrated} unmigrated package activities.
+
+    You cannot do this db upgrade until you completed the package activity
+    migration first. Full instructions for this situation are here:
+
+    https://github.com/ckan/ckan/wiki/Migrate-package-activity#if-you-tried-to-upgrade-from-ckan-28-or-earlier-to-ckan-29-and-it-stopped-at-paster-db-upgrade
```
wardi: I thought we could upgrade the db and then migrate the activities later, why is the upgrade being prevented here? Also the `num_unmigrated` implementation walks the whole activity table doing string matching on every record every, that's going to be very slow for large sites.

ckan/migration/migrate_package_activity.py
```diff
@@ -0,0 +1,261 @@
...
+    from ckan import model
+
+    context = get_context()
+    # 'hidden' activity is that by site_user, such as harvests, which are
+    # not shown in the activity stream because they can be too numerous.
+    # However these do have Activity objects, and if a hidden Activity is
+    # followed be a non-hidden one and you look at the changes of that
+    # non-hidden Activity, then it does a diff with the hidden one (rather than
+    # the most recent non-hidden one), so it is important to store the
+    # package_dict in hidden Activity objects.
```
wardi: ok I was confused by hidden but that makes sense. How about calling these "system" activity or "automated" activity instead of "hidden"?

ckan/templates-bs2/package/base.html
```diff
@@ -16,7 +16,7 @@
 	{% else %}
 	    <li>{% link_for _('Datasets'), named_route='dataset.search' %}</li>
 	{% endif %}
-	<li{{ self.breadcrumb_content_selected() }}>{% link_for dataset|truncate(30), named_route='dataset.read', id=pkg.name %}</li>
+	<li{{ self.breadcrumb_content_selected() }}>{% link_for dataset|truncate(30), named_route='dataset.read', id=pkg.name if is_activity_archive else pkg.name %}</li>
```
wardi: the if/else looks like a no-op. Did you mean to use the id instead of the name when is_activity_archive is true?

ckan/migration/migrate_package_activity.py
davidread: cool, changed

ckan/model/__init__.py
davidread: We want people to migrate before they upgrade for the following case: this migration is CKAN 2.9, and then 2.10 comes out which removes revisioning and now migrate_package_activity.py isn't compatible. If a site is upgrading from 2.8 to 2.10 it seems easier if we force people to migrate before upgrade, rather than saying they have to upgrade to 2.9 first, start CKAN then run the migration script and then upgrade to 2.10.

Yeah good point about num_unmigrated. I'll have a ponder about this.

ckan/templates-bs2/package/base.html
davidread: good spot! changed

ckan/tests/migration/test_revision_legacy_code.py
```diff
@@ -0,0 +1,287 @@
...
+
+import ckan.lib.search as search
+from ckan.lib.dictization.model_save import package_dict_save
+from ckan.lib.create_test_data import CreateTestData
+
+from ckan.migration.revision_legacy_code import package_dictize_with_revisions as package_dictize
+
+
+# tests here have been moved from ckan/tests/legacy/lib/test_dictization.py
+class TestPackageDictizeWithRevisions:
```
wardi: `...(object):` to make sure we don't get a "classic class"

wardi:
added some more comments to the changes, @smotornyuk @amercader are assigned, let's wait to hear what they think too

ckan/migration/migrate_package_activity.py
```diff
@@ -0,0 +1,263 @@
...
+
+        # get the dataset as it was at this revision
+        context[u'revision_id'] = activity[u'revision_id']
+        # call package_show just as we do in package.py:activity_stream_item(),
+        # only with a revision_id
+        try:
+            dataset = logic.get_action(u'package_show')(
+                context,
+                {u'id': activity[u'object_id'], u'include_tracking': False})
+        except logic.NotFound as exc:
```
wardi: other errors are possible here, an old revision might have data that doesn't conform to the latest dataset schema or is incompatible with currently installed plugins. Those errors shouldn't prevent the migration from going ahead.

ckan/tests/migration/test_revision_legacy_code.py
davidread: 👍 changed

ckan/migration/migrate_package_activity.py
davidread: I've put a fix in https://github.com/ckan/ckan/commit/8f30186f8

ckan/migration/migrate_package_activity.py
davidread: I should have added that it's not just changes by the site_user, but you can configure other users' changes to hide using `ckan.hide_activity_from_users` config option: https://github.com/ckan/ckan/pull/1330/files It's a bit woolly, and "hidden" is not the best name, but I didn't want to stray from the existing language of this functionality, rooted in the config option. I think another option would be to call it "activity_by_filtered_out_users" which comes from the existing function `_activity_stream_get_filtered_users`, but it's a bit of a mouthful. Alternatively I rename all these things, bar the config option?

ckan/model/__init__.py
davidread: To avoid num_unmigrated being run on every upgrade in the future, I've added in migration 88. So when you upgrade to this version of CKAN, it does the check each in the two places, and never again. I seems better than trying to optimize the query or only checking the first package activity. https://github.com/ckan/ckan/pull/3972/commits/b6e72784dc17b42e0bbcffb27bdc901945bca1f5

davidread:
@wardi I think I've responded to all your points. Do have another look through.
 
@smotornyuk @amercader please comment too, and we can move towards getting this thing in, and get CKAN much closer to getting rid of revisions/vdm/pylons/python2!

amercader:
@davidread IIRC the only thing this was missing before merging was an easier migration path for users that didn't involve running separate scripts or changing CKAN versions (I might be summarizing the issue wrong). Have you made any progress on that? Do you need more discussion around options?

davidread:
@amercader yeah, to do that I've been looking at how to move the versioned model out of CKAN and into a separate .py file that the migrate script can import. If I can get it working then it will allow this migration to be run in future CKAN versions, and also make it easy to get rid of vdm.

davidread:
@wardi I've now removed the restrictions about having to do the migration before you upgrade beyond this version. I think this is good to merge, so please do take a look!

davidread:
I was hoping this would reduce CKAN's code size, but the extra is due to the new diff functionality, and ckan/migration/revision_legacy_code.py which is starting to include hacked bits of vdm, so it's not really in the CKAN codebase proper to concern devs in the future.

hardingalexh:
Anything else to contribute for this MR? My team has been evaluating implementing a similar feature and it'd be great to get this merged so we can use it :)

ckan/logic/action/get.py
wardi: what if `include_hidden_activity` was an option in the data_dict but only available to sysadmin users, i.e with the `ignore_not_sysadmin` validator? Adding options to the context means those calls are limited to a user running locally, but I think a sysadmin should be able to access the same things from an API whether local or remote.

wardi:
looking great, just had one minor comment about an API parameter above

ckan/logic/action/get.py
davidread: Nice idea. I've done this now in 75592d95b

davidread:
@wardi said:
> #3972 is fine with me, just need the ok from @amercader and/or @smotornyuk  since I'm not assigned

@amercader @smotornyuk can either of you merge this PR now, please? It's been a long-time fork and I'm really keen to move on and get the rest of vdm removed.

amercader:
Thanks for your work on this @davidread (and @TkTech for starting it)! Really thorough and well-thought.

PR #3973, opened 2018-01-03
setup.py: add include_package_data=True

sorki:
Respects MANIFEST.in and makes template loader working
when installed as package in site-packages.

I'm trying to package CKAN for NixOS and this seems required as I don't (can't) run from ckan directory directly.


tino097:
This is same as #3962

PR #3975, opened 2018-01-04
DOCS Note for sql searches with UPPERCASE column names

camfindlay:
Something we found when using the datastore_search_sql API call was that in situations where the source data headings were all uppercase, the sql would return no results (even though we knew results should show). We eventually traced this to some postgres related config around uppercase identifiers. Solution is to double quote the column name. Adding here to document for future reference incase others get stumped by this. 

Tested in CKAN 2.6.0, 2.6.4 you may want to backport this note into some of the older docs.

Original ckan-dev list reference about this at: https://lists.okfn.org/pipermail/ckan-dev/2017-November/011201.html

CC/ @amercader 

### Features:

- [ ] includes API changes


amercader:
Thanks @camfindlay !

PR #3978, opened 2018-01-07
Changes to support Postgres 10+

maxious:
Extends/fixes #3881 

### Proposed fixes:
- Upgrade Psycopg2 to a version supporting and compiled against libpq 10+
- Disable support in psycopg2 2.5+ for automatic conversion of native json types to python objects http://initd.org/psycopg/docs/extras.html#adapt-json
- Increase version number in postgres version check unit test of version that shouldn't exist from 10 to 20

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -23,7 +23,7 @@ from ckan.lib.lazyjson import LazyJSONObject
 import ckanext.datastore.helpers as datastore_helpers
 import ckanext.datastore.interfaces as interfaces
 
-import psycopg2.extras
+from psycopg2.extras import register_default_json, register_composite
 import distutils.version
 from sqlalchemy.exc import (ProgrammingError, IntegrityError,
                             DBAPIError, DataError)
@@ -117,6 +117,14 @@ def _get_engine_from_url(connection_url):
                                                'ckan.datastore.sqlalchemy.',
                                                **extras)
         _engines[connection_url] = engine
+
+    # don't automatically convert to python objects
+    # when using native json types in 9.2+
+    # http://initd.org/psycopg/docs/extras.html#adapt-json
+    register_default_json(conn_or_curs=engine.raw_connection().connection,
+                          globally=False,
+                          loads=lambda x: x)
+
     return engine
 
 
@@ -274,9 +282,7 @@ def _cache_types(context):
             # redo cache types with json now available.
             return _cache_types(context)
 
-        psycopg2.extras.register_composite('nested',
-                                           connection.connection,
-                                           True)
+        register_composite('nested', connection.connection, True)
 
 
 def _pg_version_is_at_least(connection, version):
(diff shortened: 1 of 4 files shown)


ckanext/datastore/tests/test_unit.py
```diff
@@ -43,7 +43,7 @@ def test_pg_version_check(self):
         engine = db._get_engine_from_url(config['sqlalchemy.url'])
         connection = engine.connect()
         assert db._pg_version_is_at_least(connection, '8.0')
-        assert not db._pg_version_is_at_least(connection, '10.0')
+        assert not db._pg_version_is_at_least(connection, '20.0')
```
amercader: @maxious wouldn't this mean that we support PG 11 to 19? at least in this unit test of course

ckanext/datastore/backend/postgres.py
```diff
@@ -277,6 +277,10 @@ def _cache_types(context):
         psycopg2.extras.register_composite('nested',
                                            connection.connection,
                                            True)
+        # don't automatically convert to python objects
+        # when using native json types in 9.2+
+        # http://initd.org/psycopg/docs/extras.html#adapt-json
+        psycopg2.extras.register_default_json(loads=lambda x: x)
```
wardi: interesting! This looks like a big performance win in many cases, but does it break anything?

wardi: We could consider using https://github.com/ckan/ckan/blob/master/ckan/lib/lazyjson.py#L7 to wrap the string returned from postgres so that if it's pased passed to simplejson.dump it will be rendered intact.

ckanext/datastore/tests/test_unit.py
maxious: Good for another 10 years or so ;) https://www.postgresql.org/developer/roadmap/
20 seems like a reasonable upper bound for a version that could exist in the future.

ckanext/datastore/backend/postgres.py
maxious: The change in psycopg2 (which needed to be upgraded to have pg10 support) automatically converting json to a object without a json string representation *does break* the assumptions in ckan/tests ie. about the "nested" custom type. Hence this PR to disable that behaviour.

LazyJSONObject could be considered but would need to refactor further eg. using simplejson in unit test assertions, adding a __eq__ implementation etc.

wardi:
@amercader LGTM

ckanext/datastore/backend/postgres.py
TkTech: Don't do this please, you're going to break other extensions. Either pass the cursor into `register_default_json` to scope it (which would be fine) or better yet if possible cast the JSON columns to `::text` (best performance and explicit behavior).

maxious: Refactored to scope more closely to just datastore.

TkTech: 👍

amercader:
Thanks @maxious!

Issue #3923, opened 2017-11-16
Clean up template rendering code


amercader:
[`ckan.lib.base.render()`](https://github.com/ckan/ckan/blob/a923e00e636ad8b83bc0a689c9c7e8833aa63219/ckan/lib/base.py#L97) is a mess of old code that is probably no longer used. Now that we are starting to port Pylons controllers to Flask in earnest is the perfect time to do a long overdue cleanup.

The Canadian open data portal has been running for a while in a [extremely simplified](https://github.com/open-data/ckan/blob/canada-v2.6/ckan/lib/base.py#L88) version of it. @wardi and @TkTech provided context in this [IRC conversation](https://botbot.me/freenode/ckan/2017-11-10/?msg=93378033&page=1).

Essentially:

* Remove all caching logic as it is not applied anyway
* Increase Jinja 2 cache size (https://github.com/open-data/ckan/commit/aa071b063)
* Check if the [`PageCacheMiddleware`](https://github.com/ckan/ckan/blob/a923e00e636ad8b83bc0a689c9c7e8833aa63219/ckan/config/middleware/common_middleware.py#L31) is being used at all, and if not nuke it.

I'm happy to keep all current stuff in a `_render_template_pylons()` function and have the cleaned up version just apply to Flask

smotornyuk:
For simplicity, all existing caching logic from `render` will be moved to separate function called only for pylons' routes, thus it can be safely dropped after finishing flask migrations. 

In addition: 
- enable Jinja2's bytecode cache
- leave cache size with default value(it already increased inside Jinja2 itself)
- completely remove PageCacheMiddleware. First reason for this - it wasn't used for rendering snippets and extended templates, so profit of using it is quite small. In addition, it seems that most of CKAN users configure different variations of caching using separate tools(ex. serving over nginx)

amercader:
Fixed in #3979

PR #3979, opened 2018-01-08
[#3923] Simplify render for Flask

smotornyuk:
@wardi, @amercader, [PageCacheMiddleware](https://github.com/ckan/ckan/blob/a923e00e636ad8b83bc0a689c9c7e8833aa63219/ckan/config/middleware/common_middleware.py#L31) is working for anonymous requests served by pylons, when `ckan.page_cache_enabled` is set to true.  Wouldn't it be better to enable caching middleware for Flask(and migrate `allow_cache` variable as well) instead? 
 
PS in that case `set_pylons_response_headers` should be applied to flask response as well. Otherwise, this function can be just ignored as it only controls cache headers


ckan/config/environment.py
```diff
@@ -275,7 +276,8 @@ def update_config():
                     jinja_extensions.LinkForExtension,
                     jinja_extensions.ResourceExtension,
                     jinja_extensions.UrlForStaticExtension,
-                    jinja_extensions.UrlForExtension]
+                    jinja_extensions.UrlForExtension],
+        bytecode_cache=jinja2.FileSystemBytecodeCache()
```
amercader: I can see how cloud or Docker based deployments may not want to use a file system based cache, and use the memcache one shipped by jinja2 or a [custom one](https://gist.github.com/rtt/5029885). Perhaps we could have a `ckan.jinja2.bytecode_cache_class` config option that points to the class to use (which defaults to `jinja2.FileSystemBytecodeCache`), but if it's too much hassle that can be a separate feature.

TkTech: When I cleaned up the Jinja stuff I originally had this same change in there (but as an INI configurable). We ended up removing it because it was not behaving correctly with translations, resulting in the wrong language text showing up for snippets.

Not sure if this has been fixed, but this is why we removed it the 1st time.

amercader: > Seems like the cache key wasn't including language, which it should.

Is this something that should be done at the CKAN level (ie something we can fix) or at the Jinja2 one?

in any case @smotornyuk this seems easy to test, just change langugages a few times with the cache enabled and see if it gets updated.

smotornyuk: As for different languages, there are two points:
1) cache key does not include language key, as said Tyler.
2) Jinja compiles and caches code to produce page, not page itself. So translation points stays there and there are no visible problems with translations.

So we shouldn't worry about custom locale handling cache.

As for docker - doesn't it allow to create files inside container? I thought, that those files won't be preserved during container restart, but inside one session they should be safely stored inside container's temporary folder. Isn't it so?

amercader: Docker does allow to create files in the container but the aim is to have state-less containers (ie rebuilding the container should preserve any state including caches etc.

Different backends can be a separate feature later on (@TkTech maybe your ini file based changed is the history somewhere?) I was more concerned about the handling of locales. You seem to suggest that this shouldn't be an issue and I made a quick test switching languages and everything seem to work fine. Let's merge it and keep an eye on it for a few weeks.

PR #3993, opened 2018-01-19
Search by autocomplete dataset with 'like' (% searchTerm %)

galcober-dtt:
Modified:

like_q = u"%s%%" % q

by

like_q = u'%' + q + u'%'

In order to search as the format_autocomplete method. This way we can search terms as '%term%'

Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1576,7 +1576,7 @@ def package_autocomplete(context, data_dict):
     limit = data_dict.get('limit', 10)
     q = data_dict['q']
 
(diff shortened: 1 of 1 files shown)


wardi:
changing the query as suggested will mean we can't use the existing index so the whole list of packages would be walked in the DB after every keypress. That might be ok for small sites but won't work for larger sites. @smotornyuk mentioned possibly using a solr autocomplete feature instead, that sounds like a better approach to me

smotornyuk:
Reimplemented using solr in #4191

Issue #3679, opened 2017-07-10
Overriding datastore authorization system


ggcatu:
Hello guys,

I've been looking through and found I can override ckan authorization system with a plugin that implements IAuthFunctions and just return the new fuction that I want.

But, as Datastore is already a plugin, how should I proceed in order to modify it s behaviour ?
Of course, I don't want to directly modify the Datastore functions but instead write a plugin.

In the examples, you show how to implement IDatastoreBackend but I want to override

![image](https://user-images.githubusercontent.com/9058818/28022519-b6756508-658b-11e7-95b4-57f288e6bf14.png)

Edit:

Actually I just found this:

https://github.com/ckan/ckan/blob/master/ckan/authz.py#L74

![image](https://user-images.githubusercontent.com/9058818/28022794-71aac124-658c-11e7-8f82-5e55e8e330b0.png)

Seems like having 2 plugins override the same auth function will raise an error.

amercader:
This sounds totally reasonable.

@wardi @fanjinfei  you recently pushed code to allow chaining of action functions from extensions. Could the same principle be applied to auth functions?

wardi:
@amercader absolutely. That's something we simply overlooked.

CRImier:
I'm affected by this problem. Is there something I could do to help fix this? I want to have additional checks in 'datastore_search' auth function, but I can't for now, since checks are application-specific and customer desires to have them in a separate plugin.

PR #3995, opened 2018-01-23
[#3679] Allow chaining of authentication functions

jrdh:
In a prior piece of work the ability to chain action functions was added, this commit extends that concept to authentication functions and adds tests for it.

Action chaining work: https://github.com/ckan/ckan/pull/3494

Fixes https://github.com/ckan/ckan/issues/3679

### Features:
Implements chaining auth functions, allowing auth function override from extensions. To chain an auth function, simply add the new `chained_auth` decorator to your function as well as registering it through the usual `IAuthFunctions` interface. Just like the chained action functions (PR linked above) the list of chaining functions is reversed so that the _last_ plugin to provide an overriding function will be the _first_ called when the auth is checked. Each chained authentication function will receive the next authentication function as the first parameter, followed by the usual `context` and `data_dict`.

- [ ] includes API changes


ckan/logic/__init__.py
```diff
@@ -644,6 +644,14 @@ def wrapper(context, data_dict):
     return wrapper
 
 
+def chained_auth(func):
```
wardi: I think we should rename this to `chained_auth_function` to distinguish from user authentication and because `IActions` -> `chained_action`; `IAuthFunctions` -> `chained_auth_function`

wardi:
@jrdh great work! Just that one small change I suggested above and I'll merge

ckan/logic/__init__.py
jrdh: Yep, good idea, thanks!

jrdh:
There we go @wardi, apologies for the slightly messy commit history fixing that sphinx test.

Issue #3905, opened 2017-11-06
Broken support for ITranslations on Flask requests


amercader:
Work on #3891 has made surface a bit of i18n support that wasn't included in #3213, the ability for extensions to add their own translations via `ITranslation`. This is currently handled on Pylons on a per-request basis as part of the [`i18n.handle_request()`](https://github.com/ckan/ckan/blob/8e67312b9de7e0386707b9a503d350f2da244049/ckan/lib/i18n.py#L233) call on the base controller.

The rest of code in that function is not relevant to Flask, but we need to manage somehow the `ITranslation` hook and support for the three `ckan.i18n.extra_*` config options, as the current code is Pylons specific.

If possible I would like to not have to handle this on every single request and do it just once on startup, unless someone can think of a reason not to? Just to be clear, this just tells Babel that there are extra directories and domains where it should look for translations, so it seems wasteful to ask on every single request. We might need to rewrite the `ITranslation` texts but that's all right.

I'd suggest to incorporate this logic in [here](https://github.com/ckan/ckan/blob/8e67312b9de7e0386707b9a503d350f2da244049/ckan/config/middleware/flask_app.py#L123), while the Flask app is being created. There are two Flask-babel [configuration options](https://github.com/python-babel/flask-babel/blob/master/docs/index.rst#configuration) that we should set:

* `BABEL_TRANSLATION_DIRECTORIES`: this seems to support multiple values so we should be fine here (just append the extra directories coming from `ITranslation` or the config options)
* `BABEL_DOMAIN`: I think we'll have a problem with this one, beacuse I'm pretty sure that currently Flask-babel only supports one domain. CKAN core uses the `ckan` domain, and by default extensions use `ckanext-<name>`. So Flask-babel would need to be updated. @TkTech do you think that is going to be tricky, or can you think of a work around for this?

amercader:
Following a [conversation](https://botbot.me/freenode/ckan/2017-11-10/?msg=93378335&page=1) with @TkTech on IRC, we have two options:

1. Initialize a `Babel` object for each of the translation domains (the default `ckan` one plus one for each of the extensions,  `ckanext-*`). Use `Babel.init_app(app)` with each object rather than a single `Babel(app)`.

2. Revive one of the old PRs in flask-babel that adds multiple domains support, update and get ready for @TkTech review. Suggestion was to look at https://github.com/python-babel/flask-babel/pull/40/

My gut feeling is to go for 1 but whoever takes a look will decide.

[...]

PR #3891, opened 2017-10-30
rewrite home blueprint

tino097:
Clean home blueprint from #3762

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #4003, opened 2018-02-05
[#3905] Added CKAN wrapper around flask Babel

smotornyuk:
As mentioned in #3905, there are two ways of fixing this bug. I tried first one, with separate initialization of babel object for each translation domain, but it seems that there is no possibility to do this during the initialization process.

If you'll take a look at [babel init](https://github.com/python-babel/flask-babel/blob/master/flask_babel/__init__.py#L80), you may see, that latest call will replace single babel instance inside application so that any future calls will have only last initialized app. 

As making PR into flask-babel is a little bit longer way(and I am not sure, that I have enough time for this right now), I introduce CKANBabel class, that inherited from FlaskBabel and adds most simple and transparent support of multi-domains(this, I hope, allow to replace it with enhanced FlaskBabel in future):
- There is additional `BABEL_MULTIPLE_DOMAINS` config, that contains domain for each translation path, separated with semi-columns(just like translation path)
- Previous `BABEL_DOMAIN` is used as fallback and set to `ckan`
- Iteration over translation paths(it implemented as property), shifts pointer of `domain` property. 
- if there is a domain for the current offset of domain pointer - it will be returned instead of the default domain
- otherwise, default domain returned

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -30,7 +30,7 @@ from ckan.lib.render import CkanextTemplateLoader
 from ckan.common import config, g, request, ungettext
 import ckan.lib.app_globals as app_globals
 from ckan.plugins import PluginImplementations
-from ckan.plugins.interfaces import IBlueprint, IMiddleware
+from ckan.plugins.interfaces import IBlueprint, IMiddleware, ITranslation
 from ckan.views import (identify_user,
                         set_cors_headers_for_response,
                         check_session_cookie,
@@ -39,6 +39,31 @@ from ckan.views import (identify_user,
 
 import logging
 log = logging.getLogger(__name__)
+
+
+class CKANBabel(Babel):
+    def __init__(self, *pargs, **kwargs):
+        super(CKANBabel, self).__init__(*pargs, **kwargs)
+        self.__i18n_path_idx = 0
+
+    @property
+    def domain(self):
+        default = super(CKANBabel, self).domain
+        multiple = self.app.config.get('BABEL_MULTIPLE_DOMAINS')
+        if not multiple:
+            return default
+        domains = multiple.split(';')
+        try:
+            return domains[self.__i18n_path_idx]
+        except IndexError:
+            return default
+
+    @property
+    def translation_directories(self):
+        self.__i18n_path_idx = 0
+        for path in super(CKANBabel, self).translation_directories:
+            yield path
+            self.__i18n_path_idx += 1
 
 
 def make_flask_stack(conf, **app_conf):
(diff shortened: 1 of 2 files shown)


ckan/config/middleware/flask_app.py
```diff
@@ -41,6 +41,31 @@
 log = logging.getLogger(__name__)
 
 
+class CKANBabel(Babel):
+    def __init__(self, *pargs, **kwargs):
+        super(CKANBabel, self).__init__(*pargs, **kwargs)
+        self.__i18n_path_idx = 0
```
wardi: double underscore names cause name mangling, is this necessary?

smotornyuk: Actually - no. This class does not suppose to be an ancestor for any further extending. I just wanted to notify, that this property not just internal value, but rather very-very internal pointer that does not assume not only external access but even internal(except for domain property)

amercader:
@smotornyuk There are two issues I've found. You should test against #3891, which has an actual route served with Flask that has translated strings (`home.index`). In that PR, `ckanext/example_itranslation/test/test_plugin.py` is failing because translations on extensions don't work at all.

* Running the `CKANBabel` logic on `flask_app.py` makes sense as we are initializing Flask plugins, but I now realized that this means that this only works if the plugins are already loaded when creating the app (ie they are defined in the ini file). If you load them dynamically like on [the extension tests](https://github.com/tino097/ckan/blob/a1383c3e88b85189a2adda992cdb060d72025a1e/ckanext/example_itranslation/tests/test_plugin.py#L11:L13), the new domains won't get registered (ie all 3 tests fail). This *might* be ok for now, but would prevent things like implementing a UI for dynamic loading of plugins in the future. The old Pylons use to get around this by checking [*on each web request*](https://github.com/ckan/ckan/blob/master/ckan/lib/base.py#L231) if there were new `ITranslation` plugins. I don't know what's preferrable tbh. Any thoughts @wardi?

  Anyway, the following changes in the test will make the plugin load at the right time (we can not use `_load_plugins` as this loads the plugins after the app is initialized):

   ```diff
    diff --git a/ckanext/example_itranslation/tests/test_plugin.py b/ckanext/example_itranslation/tests/test_plugin.py
    index 7cb0a34..7b2a884 100644
    --- a/ckanext/example_itranslation/tests/test_plugin.py
    +++ b/ckanext/example_itranslation/tests/test_plugin.py
    @@ -7,14 +7,19 @@ from nose.tools import assert_true, assert_false
  
  
     class TestExampleITranslationPlugin(helpers.FunctionalTestBase):
    +
    +    @classmethod
    +    def _apply_config_changes(cls, config):
    +        config['ckan.plugins'] = 'example_itranslation'
    +
         @classmethod
         def setup_class(cls):
             super(TestExampleITranslationPlugin, cls).setup_class()
    -        plugins.load('example_itranslation')
  
         @classmethod
         def teardown_class(cls):
    -        plugins.unload('example_itranslation')
    +        if plugins.plugin_loaded('example_itranslation'):
    +            plugins.unload('example_itranslation')
             super(TestExampleITranslationPlugin, cls).teardown_class()
  
         def test_translated_string_in_extensions_templates(self):

    ```

* Once we got pass this there is a more important issue. Overriding existing strings work (eg `Log in` in the example extension) but not translations marked with `{% trans %}` in the templates (`test_translated_string_in_extensions_templates` fails, while the other tests pass.)

PR #4007, opened 2018-02-08
3918 group blueprint

### LINKED ISSUE #3918: Porting Organizations/Group controllers to Flask Blueprints
As i will start porting organizations/groups controllers to Flask Blueprints, this issue should be used for tracking changes, suggestions or improvements that could be performed during the migration

tino097:
Fixes #3918 

Second attempt to port group controller to flask blueprint. Still work in progress, as some of the tests are failing,  but the issues that i had in previous attempt are solved.


tino097:
@amercader i need help regarding `TestGroupInnerSearch.test_group_search_within_org_no_results`. Within this test, we are invoking helpers method `_create_url_with_params` where we need to set controller/action from `c` but in our case it is not initialized. Do you have any suggestion what should we do here ?

ckan/views/group.py
```diff
@@ -0,0 +1,1094 @@
...
+        context = dict(model=model, user=g.user, auth_user_obj=g.userobj)
+        check_access(u'site_read', context)
+    except NotAuthorized:
+        _, action = request.url_rule.endpoint.split(u'.')
+        if action not in (u'group_list', ):
+            base.abort(403, _(u'Not authorized to see this page'))
+
+
+def index():
+    group_type = request.blueprint
```
amercader: This won't work for custom group types (ie if via `IGroupForm` we change `/group` to `/theme`. I think the approach we are currently following on master is safer, using `self._guess_group_type()`.

Of course this is to get the group `type` (ie `group`, `theme` , `collection` etc). If we need to known if the underlying class is a group or an org (eg `is_organization`) we could so something like:
```python
def index(is_organization):
     pass

group.add_url_rule(u'/', view_func=index, strict_slashes=False, defaults={'is_organization': False})

organization.add_url_rule(u'/', view_func=index, strict_slashes=False, defaults={'is_organization': True})


```

And the same when we register the custom routes on `IGroupForm`

amercader: But I don't think that would be necessary because the current controller already deals with this with the `_replace_group_org()` methods. So let's try with just `self._guess_group_type()`

ckan/lib/plugins.py
```diff
@@ -70,7 +73,14 @@ def lookup_group_controller(group_type=None):
     Returns the group controller associated with the given group type. The
     controller is expressed as a string that you'd pass to url_to(controller=x)
     """
-    return _group_controllers.get(group_type)
+    return _group_controllers.get(group_type)   
+
+
+def lookup_group_blueprints(group_type=None):
```
amercader: I don't think we'll need different blueprints. The mechanism should be the same as we have now. Different `IGroupForm` plugins get registered, and in `register_group_plugins()` we register the custom URLs (eg `/theme`) pointing the correct blueprint (`group` or `organization`). Basically we need to replace calls to the Pylons routes mapper with calls to the `.add_url_rule()` of the relevant blueprint.

The challenge might be to import the blueprint instances avoiding circular redirects. We might need to do it just before registering the routes.

```python
            map.connect('%s_index' % group_type, '/%s' % group_type, controller=group_controller, action='index')
```

amercader: Actually ignore me. We need different blueprints because we have to change the `url_prefix` of each.

So perhaps the best is to abstract the registering of the routes on the blueprint to a function and call that when registering the plugins. Something like

```python

# views/group.py

def add_group_url_rules(blueprint):
    blueprint.add_url_rule('u'/', view_func=index, strict_slashes=False)
    # ...

def add_org_url_rules(blueprint):
   #  Do we need two functions? Have groups and org different routes?
   blueprint.add_url_rule('u'/', view_func=index, strict_slashes=False)

group = Blueprint('group', __name__, url_prefix=u'/group')
organization = Blueprint('organization', __name__, url_prefix=u'/organization')

add_group_url_rules(group)
add_org_url_rules(organization)


# lib/plugins.py

def register_group_plugins():
     # ...
     for group_types in plugin.group_types():
        custom_type_blueprint = Blueprint(group_type, __name__, url_prefix=u'/' + group_type)
        add_group_url_rules(custom_type_blueprint)

```

WDYT @smotornyuk? For datasets we will need something similar

tino097:
group and organization are finished.

i think i will need #4037 merged before fixing failing tests

ckan/config/middleware/flask_app.py
```diff
@@ -173,6 +173,7 @@ def hello_world_post():
 
     # Auto-register all blueprints defined in the `views` folder
     _register_core_blueprints(app)
+    _register_plugin_blueprints(app)
```
tino097: @smotornyuk i tried as this, not sure if this is as your idea. Please check it and give some thoughts

smotornyuk: Yep, you've got the idea, but there are small problems bellow

ckan/lib/plugins.py
```diff
@@ -198,54 +200,57 @@ def register_group_plugins(map):
...
-            map.connect('%s_about' % group_type, '/%s/about/{id}' % group_type,
-                        controller=group_controller,
-                        action='about', ckan_icon='info-circle')
-            map.connect('%s_bulk_process' % group_type,
-                        '/%s/bulk_process/{id}' % group_type,
-                        controller=group_controller,
-                        action='bulk_process', ckan_icon='sitemap')
+            blueprint = Blueprint(group_type, group.group.import_name,
+                                  url_prefix='/{}'.format(group_type),
+                                  url_defaults={'package_type': group_type})
```
smotornyuk: let's use `group_type` as a name of this variable

ckan/lib/plugins.py
```diff
@@ -198,54 +200,57 @@ def register_group_plugins(map):
...
+            actions = ['group.member_delete', 'group.history', 
+                       'group.followers', 'group.follow', 'group.unfollow',
+                       'group.admins', 'group.activity']
+
+            blueprint.add_url_rule(u'/', view_func=group.index,
```
smotornyuk: adding `url_defaults` means that this keywords will be passed to each function registered under blueprint. i.e, `group.index` will be called as group_index(package_type), `group.read` - read(package_type, id), not just read(id). You can check my `dataset` file and you'll see, that all views are expecting this argument, even though it doesn't appears in the rules.

tino097:
@amercader, @smotornyuk please review it, and lets discuss how to get fixed creating duplicate blueprints  with same name

ckan/templates/group/member_new.html
```diff
@@ -55,7 +55,11 @@ <h1 class="page-heading">
     {{ form.select('role', label=_('Role'), options=c.roles, selected=c.user_role, error='', attrs=format_attrs) }}
     <div class="form-actions">
       {% if user %}
+<<<<<<< HEAD
```
smotornyuk: Oooops.

ckan/views/group.py
```diff
@@ -0,0 +1,1160 @@
...
+            link=_url_for_this_controller(
+                action='read', id=c.group_dict['name']),
+            description=_(u'Recent changes to CKAN Group: ') +
+            c.group_dict['display_name'],
+            language=unicode(get_lang()), )
+        for revision_dict in c.group_revisions:
+            revision_date = h.date_str_to_datetime(revision_dict['timestamp'])
+            try:
+                dayHorizon = int(request.params.get('days'))
+            except:
```
smotornyuk: I think we can expect only `TypeError` here

ckan/views/group.py
```diff
@@ -0,0 +1,1160 @@
...
+
+
+def add_group_type(group_type):
+    group_types.append(group_type)
+
+
+def index():
+    group_type = _guess_group_type()
+    page = h.get_page_number(request.params) or 1
+    items_per_page = 21
```
smotornyuk: Maybe it's time to use `int(config.get('ckan.datasets_per_page', 20))`
PS don't you know, why we are using 21 here?

ckan/views/group.py
```diff
@@ -0,0 +1,1160 @@
...
+
+    page = h.get_page_number(request.params)
+
+    # most search operations should reset the page counter:
+    params_nopage = [(k, v) for k, v in request.params.items() if k != 'page']
+    sort_by = request.params.get('sort', None)
+
+    def search_url(params):
+        controller = lookup_group_controller(group_type)
+        action = u'bulk_process' if getattr(c, u'action') == u'bulk_process' else 'read'
```
smotornyuk: Shouldn't it be `request.blueprint`? without group_controller, there will be no c.action

ckan/views/group.py
```diff
@@ -0,0 +1,1160 @@
...
+        url = h.url_for(controller=controller, action=action, id=id)
```
smotornyuk: And here, probably, we need `url_for('.'.join([action, controller]))

ckan/views/group.py
```diff
@@ -0,0 +1,1160 @@
...
+                   if isinstance(v, basestring) else str(v))
+                  for k, v in params]
+        return url + u'?' + urlencode(params)
+
+    def drill_down_url(**by):
+        return h.add_url_param(
+            alternative_url=None,
+            controller='group',
+            action='read',
+            extras=dict(id=c.group_dict.get('name')),
```
smotornyuk: 'group.read'

ckan/views/group.py
```diff
@@ -0,0 +1,1160 @@
...
+                'oldid': request.params.getone('selected2'),
+            }
+        except KeyError:
+            if 'group_name' in dict(request.params):
+                id = request.params.getone('group_name')
+            c.error = \
+                _('Select two revisions before doing the comparison.')
+        else:
+            params['diff_entity'] = 'group'
+            h.redirect_to(controller='revision', action='diff', **params)
```
smotornyuk: I believe that flask generates responses instead of exceptions(like pylons) for redirect, so you need to put return at this line. Same for https://github.com/ckan/ckan/pull/4007/files#diff-244c9c0a85f9bb077ce1799c1af6210cR1032

tino097: i was rebasing my branch i i think it got corrupted, some of the changes you mention, are already done

ckan/views/group.py
tino097: dont know why is 21, but i will change it as your suggestion to 20

smotornyuk: I suppose, that it related to masonry grid, which we used previously. There were 3 blocks in a row, so 21 groups are required visually. I think we've got rid of masonry for bs3

tino097: yes, i notice that tests are written in that manner. @gocemitevski create PR for updating masonry, after the merge, i will fix the tests

gocemitevski: @tino097 done. https://github.com/ckan/ckan/pull/4114

PR #4014, opened 2018-02-12
Update facet_list.html

smth:
Remove brackets and add some `span`s to the `facet_list` items, to offer more styling options.

Note: The `hidden` and `badge` classes are not necessary to achieve the above, but are in keeping with the use of Bootstrap styling throughout CKAN.


smotornyuk:
Thanks, @smth , i've merged it

PR #4027, opened 2018-02-15
[#4022] use javascript content-type for jsonp responses

### LINKED ISSUE #4022: making a JSONP call to the CKAN API returns the wrong mime type
2.7

When making an api call to the CKAN api, and using the callback parameter to generate jsonp instead of json, the content-type should be set to "application/javascript"

the content-type is set to application/json regardless of whether the callback parameter is used or not

Step 1. Make a request to https://demo.ckan.org/api/3/action/package_search?q=spending&callback=myFunc
Step 2. Inspect the response headers, and notice that the Content-Type header is set to application/json

wardi:
Fixes #4022 

### Proposed fixes:

Applying fix to both new API view and old API controller, just for backporting.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -38,6 +38,7 @@ CONTENT_TYPES = {
     'text': 'text/plain;charset=utf-8',
     'html': 'text/html;charset=utf-8',
     'json': 'application/json;charset=utf-8',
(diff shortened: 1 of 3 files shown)


smotornyuk:
yep, verified

Issue #3309, opened 2016-11-13
Python 3 support (OLD)


sirex:
Add Python 3 support.

[Python 2.7 support will end in 2 years](https://pythonclock.org/).

- [x] add Python 3 support to [ofs](https://pypi.python.org/pypi/ofs) and [pairtree](https://pypi.python.org/pypi/Pairtree): ~https://github.com/okfn/ofs/pull/14~, ~https://github.com/benosteen/pairtree/pull/2~
- [ ] remove [ofs](https://pypi.python.org/pypi/ofs) and [pairtree](https://pypi.python.org/pypi/Pairtree), they were needed by the old Filestore.
- [ ] replace Pylons with Flask: ckan/ckan#3119
- [x] remove [ordereddict](https://pypi.python.org/pypi/ordereddict)
- [ ] remove [ofs](https://pypi.python.org/pypi/ofs) and [pairtree](https://pypi.python.org/pypi/Pairtree) dependencies as they are no longer used
- [ ] migrate away from vdm or add Python 3 support to vdm since getting rid of vdm seems like a not an easy task, okfn/vdm#13
- [ ] replace [webhelpers](https://pypi.python.org/pypi/WebHelpers) with a modern alternative
- [ ] replace [repoze.who-friendlyform](https://pypi.python.org/pypi/repoze.who-friendlyform) with a modern alternative
- [ ] add Python 3 support to the ckan own code base
- [ ] Replace Paste with Click #4639

sirex:
I did some review of ckan dependencies and found following dependencies that do not support Python 3:

- [ofs](https://pypi.python.org/pypi/ofs) - last update: 2013-12-26
- [ordereddict](https://pypi.python.org/pypi/ordereddict) - last update: 2010-10-28
- [Pairtree](https://pypi.python.org/pypi/Pairtree) - last update: 2010-10-13
- [Pylons](https://pypi.python.org/pypi/Pylons) - last update: 2015-07-22
- [repoze.who-friendlyform](https://pypi.python.org/pypi/repoze.who-friendlyform) - last update: 2010-11-14
- [vdm](https://pypi.python.org/pypi/vdm) - last update: 2014-08-12
- [WebHelpers](https://pypi.python.org/pypi/WebHelpers) - last update: 2011-03-24

It looks like most of these packages are no longer maintained.

sirex:
[...]

Issue #3921, opened 2017-11-14
TypeError: expected byte string object, value of type unicode found


torfsen:
Since upgrading to CKAN 2.7.2 my Apache error logs are full of the following tracebacks:

```
[Tue Nov 14 15:08:11.096712 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973] mod_wsgi (pid=19417): Exception occurred processing WSGI script '/etc/ckan/default/apache.wsgi'.
[Tue Nov 14 15:08:11.096793 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973] Traceback (most recent call last):
[Tue Nov 14 15:08:11.096865 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]   File "/usr/lib/ckan/default/src/ckan/ckan/config/middleware/__init__.py", line 136, in __call__
[Tue Nov 14 15:08:11.096923 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]     return self.apps[app_name](environ, start_response)
[Tue Nov 14 15:08:11.096947 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]   File "/usr/lib/ckan/default/lib/python2.7/site-packages/paste/cascade.py", line 130, in __call__
[Tue Nov 14 15:08:11.096983 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]     return self.apps[-1](environ, start_response)
[Tue Nov 14 15:08:11.097003 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]   File "/usr/lib/ckan/default/src/ckan/ckan/config/middleware/common_middleware.py", line 61, in __call__
[Tue Nov 14 15:08:11.097035 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]     return self.app(environ, start_response)
[Tue Nov 14 15:08:11.097055 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]   File "/usr/lib/ckan/default/lib/python2.7/site-packages/paste/registry.py", line 379, in __call__
[Tue Nov 14 15:08:11.097086 2017] [wsgi:error] [pid 19417] [remote 127.0.0.1:7973]     app_iter = self.application(environ, start_response)
[...]

PR #4035, opened 2018-02-22
Change unicode() --> six.text_type() for Python 3

cclauss:
Related to #3309  (partial fix)

__unicode()__ was removed in Python 3 because all str are Unicode so this PR:
* Changes all calls to __unicode()__ into calls to [__six.text_type()__](https://pythonhosted.org/six/#six.text_type)
* Change all instances of __(str, unicode)__ to [__six.string_types__](https://pythonhosted.org/six/#six.string_types)
* Change remaining instances of __unicode__ to [__six.text_type__](https://pythonhosted.org/six/#six.text_type)

### DIFF
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -9,6 +9,7 @@ import urllib
 
 from webob.multidict import UnicodeMultiDict
 from paste.util.multidict import MultiDict
+from six import text_type
 
 import ckan.model as model
 import ckan.logic as logic
@@ -217,14 +218,14 @@ class ApiController(base.BaseController):
                                     'message': _('Access denied')}
             return_dict['success'] = False
 
-            if unicode(e):
+            if text_type(e):
                 return_dict['error']['message'] += u': %s' % e
 
             return self._finish(403, return_dict, content_type='json')
         except NotFound as e:
             return_dict['error'] = {'__type': 'Not Found Error',
                                     'message': _('Not found')}
-            if unicode(e):
+            if text_type(e):
                 return_dict['error']['message'] += u': %s' % e
             return_dict['success'] = False
             return self._finish(404, return_dict, content_type='json')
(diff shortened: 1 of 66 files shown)


amercader:
@torfsen As the champion of proper unicode handling what do you think of this?

@cclauss I think this is worth adding to the docs for future contributors (probably [here](http://docs.ckan.org/en/latest/contributing/unicode.html)).

More generally I wonder if we should add a section on the contributing docs about using the py2/3 compatible alternatives that @cclauss has been submitting, otherwise us and new contributors won't be aware of them

cclauss:
Added a few more files that we hidden by the __print__ Syntax Errors and also made changes to __unicode.rst__.  It would be helpful if you could also advise on the [__ur'string'__ issues](https://github.com/ckan/ckan/issues/created_by/cclauss) because I do not know how to resolve these and Python 3 Syntax Errors block progress.

torfsen:
In general, I think @cclauss did exactly the right thing. There will probably be some places where we'll run into problems with 3rd-party libraries, because they might have decided to use `bytes` instead of `str` in Python 3 -- I'm especially thinking about the networking stuff (things like #3921). However I'm not sure whether these can be discovered a priori, since we currently don't have a consistent string handling w.r.t. to bytes vs. Unicode in CKAN. Hence, in my opinion, this PR is the right approach, problems can then be figured out when they arise.

W.r.t. to `ur` strings there doesn't seem to be a totally Python 2 + 3 compatible solution. However, the differences between `ur` in Python 2 and `r` in Python 3 seem only to concern `\u` escape sequences -- which we currently don't use in CKAN (at least according to a [quick GitHub search](https://github.com/ckan/ckan/search?utf8=%E2%9C%93&q=%5Cu&type=)). However, we still need to figure out if we can simply use `r` instead of `ur` (which would allow us to use the same code for Python 2 and Python 3). The main use of raw strings is in regular expressions, but I don't know if the `re` module of Python 2 behaves differently for `unicode` vs. `str` patterns (like `%` formatting does, for example).

Also, obviously + 1 for updating the docs on these issues. I also suggest adding a separate page in the documentation regarding Python 2 / 3 compatibility in general.

amercader:
@cclauss can you add the convention of using `six.text_type()` instead of `unicode()` to the Unicode docs page before we merge this? 
http://docs.ckan.org/en/latest/contributing/unicode.html

Cheers

cclauss:
Done.

PR #4032, opened 2018-02-22
[#4031] Improve IGroupForm support

amercader:
Fixes #4031

The individual commits are self-contained and pretty self-explanatory, but basically this fixes:

* Ability for org based plugins to be detected as such by the controller
* Ability to register a fallback org plugin
* Allow generation of URLs using custom types (eg `/publisher` or `/theme`). Right now `/group` and `/organization` were hardcoded everywhere.

Only thing missing is updating the bootstrap 2 templates.

I really think a version of this should be backported to 2.7. I think everything is pretty safe, even the changes in plugin logic, but let's discuss.

Issue #4031, opened 2018-02-22
Improve IGroupForm support


amercader:
Support for custom group schemas is well behind dataset ones.

These are some issues that need addressing:
* Issues around differentiating group-based and org-based plugins
* No `group_type` passed to templates, routes hardcoded to the core `group` and `organization` controllers
* Hardcoded logic in the group controller that prevents using custom names (eg `/publisher` or `/theme`)
* Can not define a fallback `IGroupForm` for organizations

PR #4037, opened 2018-02-23
[#4031] Improve IGroupForm support (2.7 version)

amercader:
These are the same changes as in #4032 but targeting 2.7, so it's easier to decide what is safe to backport.

Although there are a lot of changes in templates IMO they are all safe, as they don't touch any template structure or markup, just the way that URLs are generated.

Adding names to routes should not affect any existing generation or resolution.

https://github.com/ckan/ckan/commit/8ec960bcc0b5940bd497bbabb65cdcff1546db0f is the one that touches more critical code, but it's basically fixing a bug and I can't see how it could break any existing plugin (more than they already are)

@wardi let me know what you think after you've reviewed #4032

### DIFF
--- a/ckan/config/routing.py
+++ b/ckan/config/routing.py
@@ -292,19 +292,23 @@ def make_map():
                   highlight_actions='index search')
         m.connect('group_list', '/group/list', action='list')
         m.connect('group_new', '/group/new', action='new')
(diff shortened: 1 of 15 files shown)


amercader:
Decided not to backport, will be available in 2.8

PR #4038, opened 2018-02-23
print() is a function in Python 3

cclauss:
__print()__ is a function in Python 3 so this PR makes appropriate changes to be compatible with both Python 2 and Python 3.

Also fixes other Python 3 Syntax Errors:
* __python3 -c "01"__
* __python3 -c "0 <> 1"__
* old style exceptions


amercader:
Thanks @cclauss! keep them coming :)

PR #4041, opened 2018-02-25
Remove datastore legacy mode

wardi:
In #2562 we start using the read-only role for all non-writing operations so it doesn't make sense for sites to disable  `datastore_sql_search` by removing the read-only role from their configuration. All CKAN sites using datastore should have both read-only and read-write roles for their datastore database, set with the usual `ckan.datastore.write_url` and `ckan.datastore.read_url` settings.

For sites that don't want to provide `datastore_search_sql` we point to the existing explicit configuration option in the documentation http://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-datastore-sqlsearch-enabled


wardi:
Please review #2562 first, which this is based on.

(I love PRs that remove 300 lines while adding significant new functionality)

smotornyuk:
Makes perfect sense and should be merged as soon as original PR approved :+1:

amercader:
:tada: The DataStore feels much lighter all of a sudden! Thanks @wardi

PR #4042, opened 2018-02-26
Major bug in datasotre plugin

gsueur:
Hi
I've just discovered this bug in the datastore plugin which set ALL the resources to datastore_active=False when any off the resources is deleted, just because the wrong query is used in the plugin to make the update. 
Thanks for taking this in consideration. I don't even understand how it hasn't been discovered before. 
Best regards

Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -196,7 +196,7 @@ class DatastorePlugin(p.SingletonPlugin):
                 'resource_id': res.id,
             })
             res.extras['datastore_active'] = False
-            res_query.update(
+            res.update(
                 {'extras': res.extras}, synchronize_session=False)
 
     # IDatastore


camfindlay:
Which versions of CKAN contain this bug? only new or any past versions?

gsueur:
I didn't dig into the archives. But the 2.7 version and the current code at least.

maxious:
https://github.com/ckan/ckan/commit/69094fff3b7f2f582cfad7db8e541093c2426ae6 in 2.7.0+ looks like the first commit with res_query.update

wardi:
We should also come up with a repair script that copies the extras out of the revisions or activities in the database.

wardi:
We should also consider an announcement on the mailing list because this affects everyone running 2.7 with the datastore if those users ever run the search-index rebuild command after the corruption of resource extras has happened.

gsueur:
Actually I don't need to run the search index rebuild to be annoyed by the bug. I guess CKAN is only using the database values to detect if the resource is datasource or file store located, which makes ReclineView behave differently (loading the file store data from the online web service for example. That's how I discovered the problem because my firewall was blocking the web service and thus the data didn't display where it was displaying nice few days before. Immediately after a resource deletion this behavior was noticeable.)
It has consequences on resource updates as well.

smotornyuk:
@gsueur , thanks for PR. Shame on me, i created this problem:( 

As for your fix - you are using `res.update`, where **res** is `ckan.model.Resource` object, thus it has no **update** method. Could you please update you PR with something like patch bellow? (:information_source: and attach test for verifying fix, please. If you don't mind, patch already contains one, that can be used)
```diff
diff --git a/ckanext/datastore/plugin.py b/ckanext/datastore/plugin.py
index 8d0c6b7b5..f194704ab 100644
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -196,7 +196,7 @@ class DatastorePlugin(p.SingletonPlugin):
                 'resource_id': res.id,
             })
             res.extras['datastore_active'] = False
-            res_query.update(
+            res_query.filter_by(id=res.id).update(
                 {'extras': res.extras}, synchronize_session=False)
 
     # IDatastore
diff --git a/ckanext/datastore/tests/test_delete.py b/ckanext/datastore/tests/test_delete.py
index 0bba49743..af6a4c130 100644
--- a/ckanext/datastore/tests/test_delete.py
+++ b/ckanext/datastore/tests/test_delete.py
@@ -124,6 +124,45 @@ class TestDatastoreDelete():
             NotFound, helpers.call_action, 'datastore_search',
             resource_id=resource_id)
 
+    def test_datastore_deleted_during_resource_only_for_deleted_resource(self):
+        package = factories.Dataset()
+        data = {
+            'boo%k': 'crime',
+            'author': ['tolstoy', 'dostoevsky'],
+            'package_id': package['id']
+        }
+
+        result_1 = helpers.call_action(
+            'datastore_create', resource=data.copy())
+        resource_id_1 = result_1['resource_id']
+
+        result_2 = helpers.call_action(
+            'datastore_create', resource=data.copy())
+        resource_id_2 = result_2['resource_id']
+
+        res_1 = model.Resource.get(resource_id_1)
+        res_2 = model.Resource.get(resource_id_2)
+
+        # `synchronize_session=False` and session cache requires
+        # refreshing objects
+        model.Session.refresh(res_1)
+        model.Session.refresh(res_2)
+        assert res_1.extras['datastore_active']
+        assert res_2.extras['datastore_active']
+
+        helpers.call_action('resource_delete', id=resource_id_1)
+
+        assert_raises(
+            NotFound, helpers.call_action, 'datastore_search',
+            resource_id=resource_id_1)
+        assert_raises(
+            NotFound, helpers.call_action, 'resource_show',
+            id=resource_id_1)
+        model.Session.refresh(res_1)
+        model.Session.refresh(res_2)
+        assert not res_1.extras['datastore_active']
+        assert res_2.extras['datastore_active']
+
     def test_delete_invalid_resource_id(self):
         postparams = '%s=1' % json.dumps({'resource_id': 'bad'})
         auth = {'Authorization': str(self.sysadmin_user.apikey)}
```

gsueur:
I'm not sure I know how to do that, sorry.

smotornyuk:
Not a problem - i've created [PR into your branch](https://github.com/gsueur/ckan/pull/1)

amercader:
Closed in favour of #4063

PR #4046, opened 2018-02-27
[#4031] Fix wrong action in BS2 template

amercader:
Spotted by @tino097 in https://github.com/ckan/ckan/pull/4032#discussion_r170934088


tino097:
@amercader its ok

PR #4048, opened 2018-02-28
Make parameters for wrapper() functions in ckan.logic generic args/kwargs

alycejenni:
### Proposed fixes:
Using both `@toolkit.chained_action` and one of the ckan.logic decorators (e.g. `@logic.side_effect_free`) caused an error as `chained_action` introduces a new parameter that the `wrapper` function in the decorator cannot handle. As the `wrapper` function doesn't actually use or modify the parameters, I've changed them to `*args, **kwargs` to allow for any number of parameters.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -577,12 +577,8 @@ def side_effect_free(action):
     your action function with CKAN.)
 
     '''
-    @functools.wraps(action)
-    def wrapper(*args, **kwargs):
-        return action(*args, **kwargs)
-    wrapper.side_effect_free = True
-
-    return wrapper
+    action.side_effect_free = True
+    return action
 
 
 def auth_sysadmins_check(action):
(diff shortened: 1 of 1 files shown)


wardi:
@smotornyuk https://github.com/ckan/ckan/pull/3494 is the chained action change mentioned with the new functions looking like https://github.com/ckan/ckan/pull/3494/files#diff-2373bbaa29e51613aa02630e46c0b7adR12

ckan/logic/__init__.py
```diff
@@ -578,8 +578,8 @@ def my_custom_action_function(context, data_dict):
 
     '''
     @functools.wraps(action)
-    def wrapper(context, data_dict):
-        return action(context, data_dict)
+    def wrapper(*args, **kwargs):
+        return action(*args, **kwargs)
     wrapper.side_effect_free = True
```
wardi: there's no need for the wrapper here at all, right? How about:
```python
def side_effect_free(action):
    '...'
    action.side_effect_free = True
    return action
```

alycejenni: that seems better - makes it much more like the chained_action definition as well.

PR #4056, opened 2018-03-01
[#4055] Remove Celery-based tasks

amercader:
Fixes #4055 

More spring cleaning

PR #4058, opened 2018-03-01
Fix reload() for Python 3

cclauss:
Related to #3309 (partial fix)

### Proposed fixes:

The builtin __reload()__ was moved into [importlib](https://docs.python.org/3/library/importlib.html#importlib.reload) in Python 3 because it was being overused and led to odd behavior that was often difficult to diagnose.  

In `__init__.py__` we are trying to __reload(sys)__ as a workaround to #2636  I suspect (but desire confirmation!) that this is about the lack of 'utf-8' and Unicode strs in Python 2 while both are the default in Python 3.  This PR therefor suggests that a __pass__ should be satisfactory in Python 3.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/pastertemplates/__init__.py
+++ b/ckan/pastertemplates/__init__.py
@@ -55,8 +55,11 @@ class CkanextTemplate(Template):
 
         # workaround for a paster issue https://github.com/ckan/ckan/issues/2636
         # this is only used from a short-lived paster command
-        reload(sys)
-        sys.setdefaultencoding('utf-8')
+        try:
+            reload(sys)  # Python 2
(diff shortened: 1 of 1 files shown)


ckan/tests/lib/test_celery_app.py
```diff
@@ -17,7 +18,7 @@ def check_celery_version(self, version):
         import celery
         celery.__version__ = version
         import ckan.lib.celery_app
-        reload(ckan.lib.celery_app)
+        reload_module(ckan.lib.celery_app)
```
wardi: this module is being removed by #4056

PR #4062, opened 2018-03-02
Flask migration: package controller

smotornyuk:
Stage #1: convert controller


ckan/lib/plugins.py
```diff
@@ -103,20 +107,18 @@ def register_package_plugins(map):
...
-                        controller='package', action='new')
-            map.connect('%s_read' % package_type, '/%s/{id}' % package_type,
-                        controller='package', action='read')
-
-            for action in ['edit', 'authz', 'history']:
-                map.connect('%s_%s' % (package_type, action),
-                            '/%s/%s/{id}' % (package_type, action),
-                            controller='package',
-                            action=action)
+            blueprint = Blueprint(
```
tino097: i have followed same approach but when executing the tests, it looks like this blueprint is not regiestered by flask_app, any idea how we could do it ?

smotornyuk: We need to actually register it - i/e, with `app` object. Current code block isn't working yet. I'm planning  to move call of this function(`register_package_plugins`) to initialisation of flask app, right after registering all core bluprints. Now it called from `routing` module and that has less sense as we are rather adding new blueprints, than new routes

ckan/lib/plugins.py
```diff
@@ -91,6 +96,7 @@ def register_package_plugins(map):
     # running unit tests.
     if _default_package_plugin is not None:
         return
+    dataset = import_module(app.blueprints['dataset'].import_name)
```
tino097: is this necessary because `from views import dataset as dataset` could do the job ?

ckan/config/middleware/flask_app.py
```diff
@@ -400,3 +405,7 @@ def is_blueprint(mm):
         for blueprint in inspect.getmembers(module, is_blueprint):
             app.register_blueprint(blueprint[1])
             log.debug(u'Registered core blueprint: {0!r}'.format(blueprint[0]))
+
+
+def _register_package_blueprints(app):
```
tino097: can we name this method `_register_plugin_blueprints` ?

ckan/lib/plugins.py
smotornyuk: Nope. On that point we'll get import error for direct import

ckan/config/middleware/flask_app.py
smotornyuk: Yep, i forgot that we'll need to extend this function with group blueprints

ckan/lib/plugins.py
tino097: yes, i had same issue with the import, it looks like it doesnt need the `from ckan.views...`, but  just 
`from views ...`  :)

smotornyuk: Yeah, but as i know, py2 relative imports are prohibited in ckan, because they work completely different in python3

tino097:
final thought, great job @smotornyuk

ckan/config/middleware/flask_app.py
tino097: thanks :)

ckan/lib/plugins.py
amercader: The `importlib` call and the reference used in the Blueprint creation are super cryptic.  Can't we not use a py2/py3 safe way of importing the module? I have not tested but does this help?

```
from ckan.views import dataset
```

Or maybe `from __future__ import absolute_import`?

ckan/lib/plugins.py
```diff
@@ -99,24 +105,22 @@ def register_package_plugins(map):
...
+                dataset.dataset.import_name,
+                url_prefix='/{}'.format(package_type),
+                url_defaults={'package_type': package_type})
+            blueprint.add_url_rule(
+                '/', view_func=dataset.search, strict_slashes=False)
+            blueprint.add_url_rule(
+                '/new', view_func=dataset.CreateView.as_view(str(u'new')))
+            blueprint.add_url_rule('/<id>', view_func=dataset.read)
+            blueprint.add_url_rule(
+                '/edit/<id>', view_func=dataset.EditView.as_view(str(u'edit')))
```
amercader: We are missing some routes here, follow and activity. Plugins should have the same routes as the main blueprints

ckan/templates-bs2/package/resource_read.html
```diff
@@ -1,15 +1,15 @@
 {% extends "package/base.html" %}
 
-{% set res = c.resource %}
+{% set res = resource %}
```
amercader: It's great that we are getting rid of this variables in the `c` object in the templates but AFAICT you are not keeping them in the dataset/resource blueprints. We need to keep them in `c` for at lest a version, otherwise all custom templates that use them will crash. See the approach I suggested here https://github.com/ckan/ckan/pull/3203 (question 1). So basically keep them in `c`, clearly marked for deletion but don't use them in templates.

ckanext/webpageview/tests/test_view.py
```diff
@@ -9,18 +9,19 @@
 
 
 class TestWebPageView(helpers.FunctionalTestBase):
+    @classmethod
+    def _apply_config_changes(cls, cfg):
```
amercader: Has the logic for loading plugins changed? I can see how `example_idatasetform` would be affected by my first comment in the PR but what about this and other view plugins?

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+        sort_by_fields = [field.split()[0] for field in sort_by.split(u',')]
+    extra_vars[u'sort_by_fields'] = sort_by_fields
+
+    pager_url = partial(_pager_url, params_nopage, package_type)
+
+    search_url_params = urlencode(_encode_params(params_nopage))
+    extra_vars[u'search_url_params'] = search_url_params
+
+    try:
+        # c.fields_grouped will contain a dict of params containing
```
amercader: fields_grouped not c.fields_grouped

ckan/lib/helpers.py
```diff
@@ -148,12 +148,12 @@ def redirect_to(*args, **kw):
         import ckan.plugins.toolkit as toolkit
 
         # Redirect to /dataset/my_dataset.
-        toolkit.redirect_to(controller='package', action='read',
+        toolkit.redirect_to('dataset.read',
```
amercader: This is the example on how to build a Pylons-based URL :) Change it to a controller still not migrated or just use placeholders like "my_controller" and "my_action"

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+    data_dict = {u'id': id, u'include_tracking': True}
+
+    # check if package exists
+    try:
+        pkg_dict = get_action(u'package_show')(context, data_dict)
+        pkg = context[u'package']
+    except (NotFound, NotAuthorized):
+        return base.abort(404, _(u'Dataset not found'))
+
+    # FEATURE: [start] https://github.com/ckan/ckan/pull/3972
```
amercader: What's this doing here? This is not yet merged to master AFAICT

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+            return base.abort(404, _(u'Dataset not found'))
+        # Earlier versions of CKAN only stored the package table in the
+        # activity, so add a placeholder for resources, or the template
+        # will crash.
+        pkg_dict.setdefault(u'resources', [])
+        is_activity_archive = True
+    # FEATURE: [end]
+
+    # can the resources be previewed?
+    for resource in pkg_dict[u'resources']:
```
amercader: great to remove all the old `preview` interface stuff :+1:

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+
+        context = {
+            u'model': model,
+            u'session': model.Session,
+            u'user': g.user,
+            u'auth_user_obj': g.userobj,
+            u'save': self._is_save()
+        }
+
+        # Package needs to have a organization group in the call to
```
amercader: I'm not following this comment and how it relates to call to `check_access` below

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+                u'pkg_dict': pkg_dict,
+                u'pkg': pkg,
+                u'resources_json': resources_json,
+                u'form_snippet': form_snippet,
+                u'errors_json': errors_json
+            }
+        )
+
+
+# FIXME: do we really need this route?
```
amercader: Nope, remove it!

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+        h.flash_success(
+            _(u"You are no longer following {0}").format(
+                package_dict[u'title']
+            )
+        )
+
+    return h.redirect_to(u'dataset.read', id=id)
+
+
+# FIXME: Do we need this?
```
amercader: There is a url registered and a page being rendered currently (although not linked from the UI) so let's keep it. We can check if it works as a separate issue.

ckan/views/dataset.py
```diff
@@ -0,0 +1,1175 @@
...
+)
+dataset.add_url_rule(u'/followers/<id>', view_func=followers)
+dataset.add_url_rule(
+    u'/groups/<id>', view_func=GroupView.as_view(str(u'groups'))
+)
+dataset.add_url_rule(u'/activity/<id>', view_func=activity)
+
+
+# XXX: [kill] history
+def history(package_type, id):
```
amercader: Let's do this as part of #3972

amercader: By that I mean keep it in this PR

ckan/views/resource.py
```diff
@@ -0,0 +1,866 @@
...
+        except (NotFound, NotAuthorized):
+            return base.abort(404, _(u'Resource view not found'))
+
+    if not view or not isinstance(view, dict):
+        return base.abort(404, _(u'Resource view not supplied'))
+
+    return h.rendered_resource_view(view, resource, package, embed=True)
+
+
+# FIXME: could anyone think about better name?
```
amercader: `EditResourceViewView` is not great either but slightly more clear :)

ckan/lib/plugins.py
amercader: The resource URLS are not being customized with the dataset type either. To be clear this is the current behaviour in master, but I wonder is now is the time to properly support all dataset related URLs? cc @wardi

amercader:
@smotornyuk my browser crashed mid-review and I thought I had lost some comments, so apologies for some of the comments being duplicated. If anything is not clear just let me know.

ckan/lib/plugins.py
smotornyuk: No, it won't work - the best I can choose is `dataset = import_module('ckan.views.dataset')`

smotornyuk: Currently i'm moving plugin registering/cleaning from initialization of flask stack back to routes and direct import `from .. import ..` works again.

ckan/config/middleware/flask_app.py
```diff
@@ -161,6 +162,9 @@ def ungettext_alias():
 
     babel.localeselector(get_locale)
 
+    # XXX: do we need such type of cleaning?
```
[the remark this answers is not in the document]
smotornyuk: ok, i put it back to `make_map`

ckan/lib/helpers.py
```diff
@@ -269,12 +269,12 @@ def url_for(*args, **kw):
 
     URLs built by Pylons use the Routes syntax::
 
-        url_for(controller='package', action='read', id='my_dataset')
+        url_for('dataset.read', id='my_dataset')
```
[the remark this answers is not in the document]
smotornyuk: I'll update it

ckan/templates-bs2/home/snippets/search.html
```diff
@@ -1,8 +1,8 @@
-{% set tags = h.get_facet_items_dict('tags', limit=3) %}
+{% set tags = h.get_facet_items_dict('tags', c.search_facets, limit=3) %}
```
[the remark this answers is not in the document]
smotornyuk: Good point. I updated changelog and set default value for this argument(which triggers using of c.search_facets). 
PS i mentioned, that in future new argument will be required

ckan/templates-bs2/package/search.html
```diff
@@ -13,17 +13,17 @@
...
         {% if h.check_access('package_create') %}
           <div class="page_primary_action">
-            {% link_for _('Add Dataset'), controller='package', action='new', class_='btn btn-primary', icon='plus-square' %}
+            {% link_for _('Add Dataset'), named_route='dataset.new', class_='btn btn-primary', icon='plus-square' %}
           </div>
         {% endif %}
       {% endblock %}
       {% block form %}
         {% set facets = {
-          'fields': c.fields_grouped,
```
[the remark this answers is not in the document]
smotornyuk: Yep, working on this

ckan/tests/controllers/test_package.py
```diff
@@ -1025,24 +996,25 @@ def test_anonymous_users_cannot_edit_resource(self):
...
 class TestResourceView(helpers.FunctionalTestBase):
+    @classmethod
+    def _apply_config_changes(cls, cfg):
+        cfg['ckan.plugins'] = 'image_view'
+
     @classmethod
     def setup_class(cls):
         super(cls, cls).setup_class()
 
-        if not p.plugin_loaded('image_view'):
```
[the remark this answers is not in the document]
smotornyuk: Hm, it looks like problem appears after i put plugin registration into flask initialization. Not, when i restored original workflow, this change no more required

amercader:
@smotornyuk I've been looking at this and playing a bit in the [package-blueprint-tests](https://github.com/ckan/ckan/tree/package-blueprint-tests) branch. I don't think the approach that we've followed here is going to work. By this I mean registering the `IDatasetForm` blueprints in `environment.py / routing.py`. It's simple a game of chicken and egg and we can't guarantee that the application will be ready when we add the blueprints.

In its current state when trying to load up an `IDatasetForm` plugin defined in the ini file the app will fail on startup with a `RuntimeError: Working outside of application context.` error.

The way ckan works is:

1. Load environment
2. Create app

If registering the `IDatasetForm` blueprints requires the app, then doing it in 1 it's always going to be fragile. On my branch I went back to your original approach of loading them after 2. It required a bit of changes in the plugins and there are still some errors but conceptually it sounds a bit safer, even if we need to rewrite some tests to load the plugins at the right time.

Here's the [current diff](https://gist.github.com/amercader/269cc25414c06d5f47aa0fbf743a1721)

Can you have a look at let me know what you think? Maybe I'm missing an obvious drawback.

smotornyuk:
:fireworks:

amercader:
@smotornyuk We are sooooooooooooo close!

I got this when trying to use scheming (or use an `IDatasetForm` that implemented the default `dataset` type):


    * AssertionError: A blueprint's name collision occurred between <flask.blueprints.Blueprint object at 0x7f06529dc990> and <flask.blueprints.Blueprint object at 0x7f0652a378d0>.  Both share the same name "dataset".  Blueprints that are created on the fly need unique names.


We need to tweak a bit how plugins are registered, to not re-register a blueprint that implements the core routes:

```diff
diff --git a/ckan/lib/plugins.py b/ckan/lib/plugins.py
index a849527..c9966b6 100644
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -103,8 +103,18 @@ def register_package_plugins(app):
                                  "IDatasetForm has been registered")
             _default_package_plugin = plugin
         for package_type in plugin.package_types():
-            # Create a connection between the newly named type and the
-            # package controller
+
+            if package_type == u'dataset':
+                # The default routes are registered with the core 
+                # 'dataset' blueprint
+                continue
+
+            elif package_type in _package_plugins:
+                raise ValueError("An existing IDatasetForm is "
+                                 "already associated with the package type "
+                                 "'%s'" % package_type)
+
+            _package_plugins[package_type] = plugin
+
             dataset_blueprint = Blueprint(
                 package_type,
                 dataset.import_name,
@@ -121,11 +131,6 @@ def register_package_plugins(app):
             dataset_resource_rules(resource_blueprint)
             app.register_blueprint(resource_blueprint)
 
-            if package_type in _package_plugins:
-                raise ValueError("An existing IDatasetForm is "
-                                 "already associated with the package type "
-                                 "'%s'" % package_type)
-            _package_plugins[package_type] = plugin
 
     # Setup the fallback behaviour if one hasn't been defined.
     set_default_package_plugin()

```

This I think we need to include in this PR as it pretty critical.

The other thing I've noticed is that we have lost the current new resource URL, it changed from `/dataset/new_resource/<dataset_id>` to `/dataset/<dataset_id>/resource/new`. The second is nicer and more consistent with the rest of resource URLs, but dropping breaks things, like eg [tests in scheming](https://github.com/ckan/ckanext-scheming/blob/master/ckanext/scheming/tests/test_form.py#L34) so it would be good to support it as well. I had a quick go but it was not trivial to add as a route on the dataset view (because of circular imports with the resource view) or a route on the resource view (because of the url_prefix). Perhaps a redirect in one of both views is easier but I'll let you find out. This can be done after we merge this though

smotornyuk:
Completely agree with your patch(applied it)

As for old-style resource create url - I added it as a lazy evaluated view. Such thing exists in Django for an ages now and it seems, that flask guys like that idea too - http://flask.pocoo.org/docs/1.0/patterns/lazyloading/#loading-late . So i just copied their suggestion(with slight modification for giving classBased views a name) and it is pretty reusable if we need such thing in feature

amercader:
:boom: Thanks @smotornyuk !

amercader:
FYI I added a second legacy resource route using the lazy view approach in 4b55d5c

davidread:
Nice work @smotornyuk !!

ckan/views/resource.py
```diff
@@ -0,0 +1,908 @@
...
+                )
+                data = extra_vars[u'data'] or old_data
+                view_type = old_data.get(u'view_type')
+                # might as well preview when loading good existing view
+                if not extra_vars[u'errors']:
+                    to_preview = True
+            except (NotFound, NotAuthorized):
+                return base.abort(404, _(u'View not found'))
+
+        view_type = view_type or request.args.get(u'view_type')
```
tino097: @smotornyuk here im getting error `UnboundLocalError: local variable 'view_type' referenced before assignment
` when clicking on 'New view' button when editing views of the resource

ckan/views/resource.py
```diff
@@ -0,0 +1,908 @@
...
+        data[u'view_type'] = view_type
+        view_plugin = lib_datapreview.get_view_plugin(view_type)
+        if not view_plugin:
+            return base.abort(404, _(u'View Type Not found'))
+
+        self._setup_template_variables(
```
tino097: `self` should be removed

ckan/views/resource.py
```diff
@@ -0,0 +1,908 @@
...
+                data = extra_vars[u'data'] or old_data
+                view_type = old_data.get(u'view_type')
+                # might as well preview when loading good existing view
+                if not extra_vars[u'errors']:
+                    to_preview = True
+            except (NotFound, NotAuthorized):
+                return base.abort(404, _(u'View not found'))
+
+        view_type = view_type or request.args.get(u'view_type')
+        data[u'view_type'] = view_type
```
wardi: I'm seeing
`UnboundLocalError: local variable 'data' referenced before assignment`
when creating new views here

ckan/views/resource.py
```diff
@@ -0,0 +1,908 @@
...
+            pkg_dict=pkg_dict,
+            pkg=pkg,
+            resource=resource
+        )
+        return context, extra_vars
+
+    def post(self, package_type, id, resource_id, view_id=None):
+        context, extra_vars = self._prepare(id, resource_id)
+
+        to_preview = request.POST.pop(u'preview', False)
```
wardi: `request.POST` isn't a thing in flask. This would need to be `request.form` but that's immutable, so we can't call `pop`

amercader: @smotornyuk can you solve these two issues? we are obviously missing some basic tests here or these would have been picked up

PR #4063, opened 2018-03-03
[#4042] update extras only for deleted resource

smotornyuk:
Improved PR for #4042, as there are no any activities there.

As for the script, that can revert the state of resource extras back. How about comparing latest resource revision with actual extras state? As we discussed, SOLR won't help much after index rebuild. And instead of blind rewriting all extras, we can show current and revisioned version of data and ask for confirmation. Here's code:
```python
from ConfigParser import ConfigParser
from argparse import ArgumentParser
from sqlalchemy import create_engine, exc
from sqlalchemy.sql import text

config = ConfigParser()
parser = ArgumentParser()
parser.add_argument('-c', '--config', help='Configuration file', required=True)

SIMPLE_Q = (
    "SELECT id, r.extras, rr.extras revision "
    "FROM resource r JOIN resource_revision rr "
    "USING(id, revision_id) WHERE r.extras != rr.extras"
)
IMPROVED_Q = (
    SIMPLE_Q + " AND  NOT (r.extras::JSON->>'datastore_active')::BOOLEAN"
)
UPDATE_Q = text("UPDATE resource SET extras = :extras WHERE id = :id")


def main():
    args = parser.parse_args()
    config.read(args.config)
    engine = create_engine(config.get('app:main', 'sqlalchemy.url'))
    try:
        records = engine.execute(IMPROVED_Q)
    except exc.ProgrammingError:
        records = engine.execute(SIMPLE_Q)

    total = records.rowcount
    print('Found {} datasets with inconsistent extras.'.format(total))
    print()

    skip_confirmation = False
    i = 0

    while True:
        row = records.fetchone()
        if row is None:
            break
        i += 1
        id, current, rev = row
        print('[{:{}}/{}] Resource <{}>'.format(i, len(str(total)), total, id))
        print('\tCurrent extras state in DB: {}'.format(current))
        print('\tAccording to the revision:  {}'.format(rev))
        if not skip_confirmation:
            choice = raw_input(
                '\tRequired action: '
                'y - rewrite; n - skip; ! - rewrite all; q - skip all: '
            ).strip()
            if choice == 'q':
                break
            elif choice == 'n':
                continue
            elif choice == '!':
                skip_confirmation = True
        engine.execute(UPDATE_Q, id=id, extras=rev)
        print('\tUpdated')


if __name__ == '__main__':
    main()

```

Folks, what do you think? 
PS should we actually commit this script or it would be better to link it inside mail notification?

### DIFF
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -161,7 +161,7 @@ class DatastorePlugin(p.SingletonPlugin):
                 'resource_id': res.id,
             })
             res.extras['datastore_active'] = False
-            res_query.update(
+            res_query.filter_by(id=res.id).update(
                 {'extras': res.extras}, synchronize_session=False)
 
     # IDatastore
--- a/ckanext/datastore/tests/test_delete.py
+++ b/ckanext/datastore/tests/test_delete.py
@@ -124,6 +124,45 @@ class TestDatastoreDelete():
             NotFound, helpers.call_action, 'datastore_search',
             resource_id=resource_id)
 
+    def test_datastore_deleted_during_resource_only_for_deleted_resource(self):
+        package = factories.Dataset()
+        data = {
+            'boo%k': 'crime',
+            'author': ['tolstoy', 'dostoevsky'],
+            'package_id': package['id']
+        }
+
+        result_1 = helpers.call_action(
+            'datastore_create', resource=data.copy())
+        resource_id_1 = result_1['resource_id']
+
+        result_2 = helpers.call_action(
+            'datastore_create', resource=data.copy())
+        resource_id_2 = result_2['resource_id']
+
+        res_1 = model.Resource.get(resource_id_1)
+        res_2 = model.Resource.get(resource_id_2)
+
+        # `synchronize_session=False` and session cache requires
+        # refreshing objects
+        model.Session.refresh(res_1)
+        model.Session.refresh(res_2)
+        assert res_1.extras['datastore_active']
+        assert res_2.extras['datastore_active']
+
+        helpers.call_action('resource_delete', id=resource_id_1)
+
+        assert_raises(
+            NotFound, helpers.call_action, 'datastore_search',
+            resource_id=resource_id_1)
+        assert_raises(
+            NotFound, helpers.call_action, 'resource_show',
+            id=resource_id_1)
+        model.Session.refresh(res_1)
+        model.Session.refresh(res_2)
+        assert not res_1.extras['datastore_active']
+        assert res_2.extras['datastore_active']
+
     def test_delete_invalid_resource_id(self):
         postparams = '%s=1' % json.dumps({'resource_id': 'bad'})
         auth = {'Authorization': str(self.sysadmin_user.apikey)}


amercader:
I think the proposed migration script looks good in general. A couple of questions:

* What version of postgres is required for the `(r.extras::JSON->>'datastore_active')` bit?

* I'm not following the IMPROVED_Q, can't we not have corrupted records with this extra?

As for where to keep this script, I don't have a strong preference. Perhaps I'd prefer it to keep it separate in a gist or something (properly linked from the changelog), but if not we can include it in the bin folder, but with a big fat warning to not use it if not necessary.

@wardi what do you think?

wardi:
IMPROVED_Q makes sense, corrupted extras should always have datastore_active=false

Shouldn't we also be ignoring the datastore_active differences between the revision and the primary table for all resources to avoid false positives?

We also need a command that will reset the datastore_active flags for all resources based on tables in the datastore db. This would be a useful command for cases where someone is making changes directly on their datastore db through backup/restore or other tools too.

I'd like to see these commands in the ckan repo so we can more easily direct users with corrupted resources. We should create some corrupted dbs and see how well they are repaired by our commands too.

amercader:
> IMPROVED_Q makes sense, corrupted extras should always have datastore_active=false

Right, my bad. I thought that tested for the presence of the extra

> Shouldn't we also be ignoring the datastore_active differences between the revision and the primary table for all resources to avoid false positives?

Do you mean `WHERE r.extras != rr.extras EXCEPT FOR THE datastore_active EXTRA`? Any idea on how to do that?

> We also need a command that will reset the datastore_active flags for all resources based on tables in the datastore db. This would be a useful command for cases where someone is making changes 
directly on their datastore db through backup/restore or other tools too.

This sounds useful but it is a separate issue right? (obvs related to this one but can be done separately)

> I'd like to see these commands in the ckan repo so we can more easily direct users with corrupted resources. 

Sounds good

> We should create some corrupted dbs and see how well they are repaired by our commands too.

@gsueur could you give the above script a try to see if it repairs the resource extras to the pre-bug state?

gsueur:
Actually yes, my test passes, the chosen resource is deleted while the others are still marked as datastore_active=True. 
Thanks !

amercader:
> the chosen resource is deleted 

What do you mean by this? The actual resource is deleted? or the wrong extras?
Just want to make sure we are deleting the right thing :)

gsueur:
Yes, the resource I want to delete is correctly deleted from the datastore, marked as 'state'='deleted' and 'extras' = datastore_active: false in the resource table.

wardi:
@smotornyuk I'm sure there's a clever way to do the comparison in postgres but for this kind of script we could just do a last check to ignore the datastore_active flag differences in python., something like:

```python
...
        id, current, rev = row
        # differences in datastore_active flag are expected
        if dict(current, datastore_active=None) == dict(rev, datastore_active=None):
            continue
```

smotornyuk:
Yes, fair enough. I updated initial comment, adding your suggestion instead if `improved_q`

amercader:
@smotornyuk is the first script needed (the one in the first comment) ready to paste in a gist?

I assume the second one (reset datastore_active based on the actual tables in the datastore db) won't be ready for tomorrow.  That's fine BTW, I'm just thinking ahead on how to put this in the changelog

PR #4068, opened 2018-03-05
Remove redundant label elements and keep search form inputs accessible

gocemitevski:
Fixes redundant labels in search form templates, previously added as duplicate elements (PR #3581) and modifies the implementation with the help of `aria-label` attribute. The hard-coded `Search Datasets` value is now dynamic and adapts to context. Several LESS blocks in `search.less` have been commented out or modified to improve the visual rendering of the forms.

### Proposed fixes:

- Don't include redundant `<label>` elements which are visually hidden with CSS -- Approach 2 @ https://www.w3.org/WAI/tutorials/forms/labels/
- Don't use a hard-coded value for `aria-label` and use `placeholder` value instead
- Use `aria-label` to keep search input accessible and provide the expected UX for screen readers
- Refactor markup in `search_form.html` snippet respecting Bootstrap 3 guidelines

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/home/snippets/search.html
+++ b/ckan/templates/home/snippets/search.html
@@ -5,7 +5,7 @@
   <form class="module-content search-form" method="get" action="{% url_for controller='package', action='search' %}">
     <h3 class="heading">{{ _("Search data") }}</h3>
     <div class="search-input form-group search-giant">
-      <input aria-label="{% block search_input_accessible_label %}{{ _('Search datasets') }}{% endblock %}" id="field-main-search" type="text" class="form-control" name="q" value="" autocomplete="off" placeholder="{% block search_placeholder %}{{ placeholder }}{% endblock %}" />
+      <input aria-label="{% block header_site_search_label %}{{ _('Search datasets') }}{% endblock %}" id="field-main-search" type="text" class="form-control" name="q" value="" autocomplete="off" placeholder="{% block search_placeholder %}{{ placeholder }}{% endblock %}" />
       <button type="submit">
         <i class="fa fa-search"></i>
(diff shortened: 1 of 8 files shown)


ckan/templates/home/snippets/search.html
```diff
@@ -5,8 +5,7 @@
   <form class="module-content search-form" method="get" action="{% url_for controller='package', action='search' %}">
     <h3 class="heading">{{ _("Search data") }}</h3>
     <div class="search-input form-group search-giant">
-      <label for="field-main-search">{% block header_site_search_label %}{{ _('Search Datasets') }}{% endblock %}</label>
-      <input id="field-main-search" type="text" class="form-control" name="q" value="" autocomplete="off" placeholder="{% block search_placeholder %}{{ placeholder }}{% endblock %}" />
+      <input aria-label="{% block header_site_search_label %}{{ _('Search datasets') }}{% endblock %}" id="field-main-search" type="text" class="form-control" name="q" value="" autocomplete="off" placeholder="{% block search_placeholder %}{{ placeholder }}{% endblock %}" />
```
klikstermkd: @gocemitevski I think "Search datasets" needs to be dynamic, meaning if a user is on the Datasets page it should be "Search Datasets", on the Organizations page it should be "Search organizations", etc.

gocemitevski: @klikstermkd this is not related to the contextual search form for datasets/organization/groups... in `templates/snippets/search_form.html`, but the home page widget, so it is no problem to have it hard-coded. Datasets are the only thing you can search for, from within the widget.

PR #4069, opened 2018-03-05
Remove REST API related code

### LINKED ISSUE #4065: Remove REST API
The REST API should be removed, since the Action API was introduced in CKAN 2.0.

I will create a PR that addresses this issue.

Part of #3536.

klikstermkd:
Fixes #4065 

This PR removes the REST API related code. Things to remove:
- [x] Routes mapping `/api/rest/*` from `routing.py`
- [x] REST API actions from `api.py`
- [x] Actions related to REST defined in `ckan/logic/action/*`
- [x] Legacy tests for REST API code
- [x] Refactor tests that use parts of the REST API

@amercader If needed, add more items to the list of things to remove.

Also, can you give me any pointers on which tests exactly need to be removed? Because there are some legacy tests not related to REST, but use old REST APIs to test other CKAN code. 

e.g. https://github.com/ckan/ckan/blob/master/ckan/tests/legacy/functional/api/model/test_package.py#L225


amercader:
@klikstermkd thanks, this looks great

Regarding the tests if they test REST endpoints directly, we just remove them (like the example you pasted).

If the REST API is used as part of a test testing another part of CKAN, like in the `TestActivity` ones failing now on CircleCI, then ideally we would rewrite to use the action API. For instance in these particular ones it looks like we can refactor the [functions](https://github.com/ckan/ckan/blob/master/ckan/tests/legacy/functional/api/test_activity.py#L221) that call this old endpoint

https://demo.ckan.org/api/2/rest/user/amercader/activity

to call this one, as the responses are the same:

https://demo.ckan.org/api/action/user_activity_list?id=amercader

It think that should be easy and fix a lot of these tests. I might be wrong of course so let me know if it' much more work that this.

klikstermkd:
@amercader Ready for review.

TkTech:
+45 −577 :+1:

smotornyuk:
Yep, i tested it and didn't find any problem

wardi:
:tada: Great work @klikstermkd !

PR #4084, opened 2018-03-08
Allows maximum resource upload size to be specified as an env variable

### LINKED ISSUE #4083: Specify max resource size in an env variable
2.7.2 (applies to head of master as well)

When deploying CKAN through docker it is not ideal to have to edit the configuration file in order to set maximum resource upload size. This would be better done through environment variables.

Currently the ckan.max_upload_size config can only changed via editing the config .ini files.

Running CKAN through Docker via Docker Compose.

Nimphal:
Fixes #4083 

### Proposed fixes:
This PR adds support for specifying the maximum resource upload size as an ENV variable.

### Features:

- [ x] includes tests covering changes
- [ ] includes API changes


amercader:
@Nimphal Are you familiar with [ckanext-envvars](https://github.com/okfn/ckanext-envvars)? It allows you to set (almost) any config option as an env var instead of an ini file setting. It is designed for Docker environments, and you can see it used [here](https://github.com/okfn/docker-ckan).

Nimphal:
Hi @amercader I was not familiar with that extension, thank you for that. It doesn't quite fit the use case for Docker though, because it requires extra steps to enable a plugin when running ckan in Docker.

amercader:
Thanks @Nimphal.
Yes, you will need to add the `ckan.plugins` config option to the ini file (or the `envvars` plugin won't be loaded), but other than that it allows full flexibility to configure CKAN via env vars (it is used successfully in many Docker-based deployments).

In any case the resource size is definitely something that makes sense to be supported directly in core so happy to merge it. Thanks again.

PR #4100, opened 2018-03-09
[#4094] Update Orgs/Groups logic for custom fields delete and update

### LINKED ISSUE #4094: Deleting first Group and Organization custom field is not possible

2.6+

- https://demo.ckan.org/organization/edit/20171207_mk
- https://demo.ckan.org/group/edit/16848646846
- https://beta.ckan.org/organization/edit/avg-data
- https://beta.ckan.org/group/edit/anotheraillysgroup

Each **Group or Organization custom field** marked for deletion should be deleted once the Update button is clicked. The deletion button should be displayed as a **red button with an icon**.

Whey you try to delete the first Group or Organization custom field, it only gets deleted when there are *at least two custom fields set*. When you try to delete the first custom field, when it's the only one in the group with paired values added, it does not get deleted once the Update button is clicked.

Also the **delete icon is not rendered** in < 2.8 and it's displayed as a simple checkbox.

Navigate to any of the links above and mark the first custom field for deletion, then click Update. The field will not be deleted. Then add values to at least two custom fields, Update and mark the first one for deletion. When you click on Update again, it will be deleted.

Also, note that the delete icon is not rendered and only a simple checkbox is shown for marking.

The custom fields functionality is fully working for datasets, so please check the Manage dataset page for reference.

Engerrs:
fixes #4094

### DIFF
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -133,19 +133,6 @@ def package_extras_save(extra_dicts, obj, context):
         state = 'deleted'
         extra.state = state
 
-def group_extras_save(extras_dicts, context):
-
-    model = context["model"]
-    session = context["session"]
-
-    result_dict = {}
-    for extra_dict in extras_dicts:
-        if extra_dict.get("deleted"):
-            continue
-        result_dict[extra_dict["key"]] = extra_dict["value"]
-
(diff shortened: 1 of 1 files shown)


gocemitevski:
Seems to work as expected.

PR #4102, opened 2018-03-12
[#4061] Recapture version no longer has a default value

### LINKED ISSUE #4061: recaptcha v1 will stop working 2018-3-31
All since 2.0ish

### The problem
Users will not be able to register, due to the re-captcha being switched off by Google on 2018-3-31.

Google's deprecation info: https://developers.google.com/recaptcha/docs/versions#v1

This affects those sites that:
* have setup recaptcha (i.e. registered with Google and [added their keys to the CKAN config](http://docs.ckan.org/en/latest/maintaining/configuration.html?highlight=captcha#ckan-recaptcha-publickey)). This is not part of the default CKAN install. Most installs only use the 'user' functionality for admins, so the impact should be limited.
* AND they use v1 of recaptcha. This is the default in the CKAN config, but Google deprecated v1 May 2016 and prevented new sites using it (I imagine it was the same time), so the issue only affects sites set-up before then.

### How to check if you have this problem
Show the relevant bits of your CKAN config:
```
grep recaptcha /etc/ckan/default/production.ini
```
IF `ckan.recaptcha.publickey` has a value (i.e. not blank or unset)
AND (`ckan.recaptcha.version = 1` OR `ckan.recaptcha.version` is blank or unset)
THEN you need to upgrade to recaptcha 2 before 2018-03-31.

### Action
I think we should change the default to v2 and warn the community.

davidread:
Because recapture v1 is long unsupported and will be switched off 2018-03-31.

Fixes #4061 

### Proposed fixes:

I've removed the default for `ckan.recaptcha.version=1`, so now if you've set-up recaptcha (by setting a value for `ckan.recaptcha.privatekey`) you have to specify the version. Yes this change will break an install that still uses v1 but has not specified that, should this release come out before the end of the month, but it is a softer hit than Google's switch-off in a couple of weeks time.

### Features:

- [ ] includes API changes


amercader:
Isn't it better to remove the `ckan.recaptcha.version` option all together as it always will be `2`?

What we should remove for sure is the [v1 handling code](https://github.com/ckan/ckan/pull/4102/files#diff-1abde95134b22e27a56858ba52461daaR20).

davidread:
I wasn't sure if this was going to go out before the 31st, but since it sounds like it isn't, it does sound like a good plan to remove the option and v1 code.

davidread:
@amercader this is done now

amercader:
Thanks @davidread !

Issue #4130, opened 2018-03-15
500 error when using HTTP queries that contain created or last_modified fields.


martinnaughton:
### CKAN Version if known (or site URL)
2.8

### Please describe the expected behaviour
To return a JSON representation

### Please describe the actual behaviour
when doing queries using created or last_modified on HTTP a 500 error is produced by CKAN. Using other fields the query is returned normally as json.
https://demo.ckan.org/api/3/action/resource_search?query=created:2017-10-25T10:21:45.319063

From looking at the logs this is the error output.
`  File "/usr/lib/ckan/venv/src/ckan/ckan/logic/action/get.py", line 2193, in resource_search
    count = q.count()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 3089, in count
    return self.from_self(col).scalar()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2843, in scalar
    ret = self.one()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2814, in one
    ret = self.one_or_none()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2784, in one_or_none
    ret = list(self)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2855, in __iter__
    return self._execute_and_instances(context)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2878, in _execute_and_instances
    result = conn.execute(querycontext.statement, self._params)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 945, in execute
    return meth(self, multiparams, params)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/sql/elements.py", line 263, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1053, in _execute_clauseelement
    compiled_sql, distilled_params
[...]

PR #4135, opened 2018-03-19
[#4130] allow to query resources by date fields

smotornyuk:
Fix for #4130 
I'm not sure, what value of searching by date with that microsecond precision, required by PostgreSQL, but adding any additional flags for more granular customization of precision is far beyond the scope of this problem, so I just make a change that uses equality instead of LIKE operator for DateTime fields


wardi:
@smotornyuk I think we agreed to change this to only use LIKE on text fields instead of only not on date fields, right?

smotornyuk:
Yep. I finally back, so i'll update this PR soon

smotornyuk:
Sorry, i completely forgot about it:((
Updated

ashmeet-kandhari:
Hi,

Any update on this issue?

smotornyuk:
ping @wardi

wardi:
This just needs a small test to make sure it keeps working then it's GTG

smotornyuk:
And test is finally there! @wardi , if you have time - this pr is ready for deploy

PR #4143, opened 2018-03-22
Group blueprint

tino097:
Cleaned version of #4007 with less comments in the log


tino097:
executing all tests that are throwing `AttributeError: 'NoneType' object has no attribute 'form_to_db_schema'` individually are passing. 

This could be issue to the plugin reset but i would like to get some help to locate the problem

amercader:
@tino097 to solve the group_type/is_organization handling issue in the group blueprint use the same `url_defaults` param approach that @smotornyuk uses in #4062:

Main Group and Organization blueprints have the following defaults:

```
url_defaults: {'group_type': 'group', 'is_organization': False}

url_defaults: {'group_type': 'organization', 'is_organization': True}
```
eg https://github.com/ckan/ckan/pull/4062/files#diff-2c4f7feb385fad48e72f4de013976946R43

All view functions accept a `group_type` and an `is_organization` parameter:
eg https://github.com/ckan/ckan/pull/4062/files#diff-2c4f7feb385fad48e72f4de013976946R160

When registering a plugin, we change the defaults on that particular blueprint:

```
url_defaults: {'group_type': group_type, 'is_organization': is_org}

```
eg https://github.com/ckan/ckan/pull/4062/files#diff-cf66d4c076e7091ab114f9e5bdb31c4cR112

ckan/lib/plugins.py
```diff
@@ -166,15 +174,17 @@ def register_group_plugins(map):
                 is_organization = group_controller == 'organization'
```
amercader: This whole check (the setting of the `is_organization` var) should be up one level, outside the `plugin.is_fallback()` if check, as we always need the var.

amercader:
:boom: :rocket: :fireworks: 

Thanks @tino097! Another great milestone passed!

PR #4144, opened 2018-03-22
commented statement in legacy test for coding standards

tino097:
Small fix for commented line in the 
`/ckan/ckan/tests/legacy/test_coding_standards.py` which was merged with #4069


tino097:
@amercader i think we should remove `/ckan/ckan/tests/legacy/functional/api/model/test_licenses.py` 
as it passes pep8 but it is blacklisted and it is pointing to rest endpoint. Any objection on the removal ?

amercader:
sounds good, let's remove it

wardi:
@tino097 the pep8 failure needs to be fixed on this one

tino097:
@wardi its fixed now

PR #4145, opened 2018-03-22
increase the strength of the password reset token

anotheredward:
This is the first PR to merge features from the plugin:
https://github.com/data-govt-nz/ckanext-security/issues/2
into ckan core

### Fixes 
Having weak/predictable password reset tokens

### Proposed fixes:
Use a longer password reset token from a less predictable source. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -1,5 +1,6 @@
 # encoding: utf-8
 
+import codecs
 import smtplib
 import socket
 import logging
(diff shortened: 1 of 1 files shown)


wardi:
@anotheredward looks like you're missing an `import os`

anotheredward:
Thanks @wardi , I've added the missing import.

anotheredward:
While this is the code we currently have in the ckan-security module, would
`random.SystemRandom`
be more appropriate?

I see that we use that elsewhere in the code base for cryptographically secure random number generation.

TkTech:
@anotheredward `random.SystemRandom()` just provides the top-level `random` interface, but with `urandom` as the source. Just using `urandom` directly in this case is fine.

PR #4149, opened 2018-03-26
 fix an if condition to be more pythonic

tino097:
small fix which prevents error that is occurring on master.ckan.org complaining on missing attribute 

`AttributeError: '_Globals' object has no attribute 'search_facets'`

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -948,8 +948,9 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
     exclude_active -- only return unselected facets.
 
     '''
-    if not hasattr(c, 'search_facets'):
-        return c.search_facets.get(facet, {}).get('items', [])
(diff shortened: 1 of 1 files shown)


ckan/lib/helpers.py
```diff
@@ -948,10 +948,8 @@ def get_facet_items_dict(facet, limit=None, exclude_active=False):
     exclude_active -- only return unselected facets.
 
     '''
-    if not c.search_facets or \
-            not c.search_facets.get(facet) or \
-            not c.search_facets.get(facet).get('items'):
-        return []
+    if not hasattr(c, 'search_facets'):
```
amercader: This should be the other way around right? `if hasattr(c, 'search_facets'):`

tino097: i thought it should return empty list if search_facets is missing ?

PR #4157, opened 2018-03-28
Legacy tests cleanup

tino097:
Cleaned legacy tests that have pointed to REST api 

Related to #4069 and #4144


tino097:
i've commented tests in `ckan/tests/legacy/functional/api/model/test_relationships.py` as im not sure if they should be rewritten or just deltede

smotornyuk:
I think that relationships worth rewriting - we still have this feature, but it wasn't updated for a long time(half a year ago i fixed one bug and everyone was surprised by the fact, that there is a person, who try this functionality:))

PR #4158, opened 2018-03-29
Fix datastore dump > 25k records

### LINKED ISSUE #4150: Datastore dump results are not the same as data in database
2.7.2

Downloaded dataset is not consistent with the actual data! 

We've upserted data using datastore api. The data is shown correctly in data explorer, Table view, API call or even querying postgres, But the data in downloaded csv file is different! Total number of records are the same but in csv we have random number of duplicates and missing data.

Upsert data using API. Download the CSV file using URL. Filter and compare downloaded data and data shown in data explorer or table view.

wardi:
Fixes #4150 

### Proposed fixes:

dump is using an offset without a sort specified which isn't reliable. Sort on `_id` field.

A better fix would add this sort and use the id of the last record retrieved to fetch the next page, but that's a larger change for a future version.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/controller.py
+++ b/ckanext/datastore/controller.py
@@ -137,6 +137,7 @@ def dump_to(resource_id, output, fmt, offset, limit, options):
                 PAGINATE_BY if limit is None
                 else min(PAGINATE_BY, lim),
             'offset': offs,
+            'sort': '_id',
             'records_format': records_format,
             'include_total': 'false',  # XXX: default() is broken
         })


TkTech:
For clarification if anyone is looking, skip+offset with large #s of rows starts to perform very poorly at high offsets, especially with unindexed filters. Changing `SELECT * FROM _ LIMIT 250000 OFFSET 250000` to `SELECT * FROM _ WHERE _id > last_returned_id ORDER BY "_id" ` where `_id` has a pre-sorted index (`CREATE INDEX id_desc_idx` ON resource_table (_id DESC NULLS LAST);`) is many times faster.

wardi:
@amercader no, there was no attempt to dump all records pre-2.7 so this doesn't apply before then

PR #4190, opened 2018-04-17
[#4066] map pylons routes to flask

### LINKED ISSUE #4066: Removing pylons routes breaks extensions
 2.8

The old named routes (e.g. 'home' or 'about') should be kept to improve the backwards compatibility of CKAN core. For extensions that want to support a range of CKAN versions this is important as they otherwise have to define these routes themselves, implement version-specific code or break one version. 

The routes were removed frim master and therefore e.g. code that build the main navigation is doomed to fail in extensions. E.g  https://github.com/ckan/ckanext-showcase/blob/92c9b2b7d4cc07b68eeb77587a832cb3c7607daa/ckanext/showcase/templates/header.html

Try to build a navigation to the homepage or the about page on both CKAN 2.7 and the latest master branch.
I would propose to either keep the old named routes or provide a way the old name to the new route.

tino097:
Fixes #4066 

### Proposed fixes:

added dictionary with mapping of route names and checking in `url_for`, `build_nav` and `build_nav_menu`  

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -1,7 +1,6 @@
 # encoding: utf-8
 
 '''CKAN environment configuration'''
-import json
 import os
 import logging
 import warnings
@@ -230,7 +229,6 @@ def update_config():
     # routes.named_routes is a CKAN thing
     config['routes.named_routes'] = routing.named_routes
     config['pylons.app_globals'] = app_globals.app_globals
-
     # initialise the globals
     app_globals.app_globals._init()
 
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -1,6 +1,7 @@
 # encoding: utf-8
 
 '''CKAN environment configuration'''
+import json
 import os
 import logging
 import warnings
@@ -229,6 +230,12 @@ def update_config():
     # routes.named_routes is a CKAN thing
     config['routes.named_routes'] = routing.named_routes
     config['pylons.app_globals'] = app_globals.app_globals
+    # Pylons to Flask legacy route names mapping0
+    import pdb; pdb.set_trace()
+    config['legacy_route_mappings'] = json.loads(
+                                      config.get('ckan.legacy_route_mappings',
+                                                 '{"home":"home.index",\
+                                                   "about": "home.about"}'))
     # initialise the globals
     app_globals.app_globals._init()
 
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -152,13 +152,6 @@ CONFIG_FROM_ENV_VARS = {
 # End CONFIG_FROM_ENV_VARS
 
 
-# Legacy route names
-LEGACY_ROUTE_NAMES = {
-    'home': 'home.index',
-    'about': 'home.about',
-}
-
-
 def update_config():
     ''' This code needs to be run when the config is changed to take those
     changes into account. It is called whenever a plugin is loaded as the
@@ -237,13 +230,6 @@ def update_config():
     # routes.named_routes is a CKAN thing
     config['routes.named_routes'] = routing.named_routes
     config['pylons.app_globals'] = app_globals.app_globals
-
-    # Pylons to Flask legacy route names mapping0
-    if config.get('ckan.legacy_route_mappings'):
-        config['legacy_route_mappings'] = LEGACY_ROUTE_NAMES.update(json.loads(
-               config.get('ckan.legacy_route_mappings')))
-    else:
-        config['legacy_route_mappings'] = LEGACY_ROUTE_NAMES
 
     # initialise the globals
     app_globals.app_globals._init()
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -152,6 +152,13 @@ CONFIG_FROM_ENV_VARS = {
 # End CONFIG_FROM_ENV_VARS
 
 
+# Legacy route names
+LEGACY_ROUTE_NAMES = {
+    'home': 'home.index',
+    'about': 'home.about',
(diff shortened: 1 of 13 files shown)


tino097:
@amercader @smotornyuk please check this and lets discuss about the option to add the dict to app_globals if we need to add option for updating from the extensions

smotornyuk:
I suggest adding config directive with additional mapping. Such approach has few advantages:
* avoid app_globals
* existing portals can use new CKAN with old extensions(by updating config file), without waiting for an update from extension maintainer

In addition, i suggest raising warning/deprecation message any time old name used.

tino097:
@smotornyuk 
i've created `ckan.route_mappings=home:home.index, about: home.about` config mapping, but if you have better idea how to look this, please do tell :) 

p.s same config should be added to test-core.ini as the test will fail

amercader:
Let's call the config option `ckan.legacy_route_mappings` so it's clear that this needs updating in the sources. What about defining it as a JSON object? It's easy to convert to a dict (and validate) and we can update the default one on helpers with the user defined ones.

`ckan.legacy_route_mappings={"home": "home.index", "about": "home.about"}`

We should show a deprecation notice (`INFO` probably) when one of the mappings is used as @smotornyuk suggests.

amercader:
@tino097 for the test you can use `@helpers.change_config` decorator to set the config option just on that test.

tino097:
@amercader its needed by all tests as im using the method in `url_for`. Any other suggestion ?

ckan/lib/helpers.py
```diff
@@ -873,6 +876,12 @@ def build_nav(menu_item, title, **kw):
     return _make_menu_item(menu_item, title, icon=None, **kw)
 
 
+def map_pylons_to_flask_route_name(menu_item):
+    '''returns flask routes for old fashioned route names'''
+    old_routes = config['legacy_route_mappings']
```
amercader:
```python
if menu_item in config['legacy_route_mappings']:
    log.info('Route name "{}" is deprecated and will be removed. Please update calls to use "{}" instead'.format(menu_item, config['legacy_route_mappings'][menu_item]))

```

ckan/tests/controllers/test_home.py
```diff
@@ -56,6 +56,13 @@ def test_email_address_no_nag(self):
 
         assert 'add your email address' not in response
 
+    @helpers.change_config('ckan.legacy_pylons_routes',
+                           '{"home":"home.index", "about": "home.about"}')
```
amercader: That test doesn't tell us much, because the values you are passing are the same ones used by default. A better test would be to set `@helpers.change_config('ckan.legacy_pylons_routes', '{"my_home_route":"home.index"}')` and see if both `url_for('home')` and `url_for('my_home_route')` still work.

ckan/config/environment.py
```diff
@@ -229,6 +230,11 @@ def update_config():
     # routes.named_routes is a CKAN thing
     config['routes.named_routes'] = routing.named_routes
     config['pylons.app_globals'] = app_globals.app_globals
+    # Pylons to Flask legacy route names mapping0
+    config['legacy_route_mappings'] = json.loads(
```
amercader: I'm not sure if defining this config option here will allow you to change the config dinamically like in the `change_config` decorator (if the test passes then great)

In any case, user defined routes should add to the existing default ones, so better do:

```python
LEGACY_ROUTE_NAMES = {
    'home': 'home.index',
    'about': 'home.about',
}

def update_config():
    # ...
    if config.get('ckan.legacy_route_mappings'):
        config['legacy_route_mappings'] = LEGACY_ROUTE_NAMES.update(
            json.loads(config['ckan.legacy_route_mappings']))



```

ckan/config/environment.py
```diff
@@ -229,6 +230,11 @@ def update_config():
...
+                                      config.get('ckan.legacy_route_mappings',
```
amercader: Please document this in `doc/maintaining/configuration.rst`

ckan/config/environment.py
tino097: i will add this change, it looks much cleaner

ckan/config/environment.py
```diff
@@ -1,6 +1,7 @@
 # encoding: utf-8
 
 '''CKAN environment configuration'''
+import json
```
amercader: not needed

ckan/config/environment.py
```diff
@@ -229,6 +230,7 @@ def update_config():
     # routes.named_routes is a CKAN thing
     config['routes.named_routes'] = routing.named_routes
     config['pylons.app_globals'] = app_globals.app_globals
+
```
amercader: noise :)

test-core.ini
```diff
@@ -14,7 +14,6 @@ cache_dir = %(here)s/data
 debug = false
 testing = true
 
-
```
amercader: noise

ckan/tests/controllers/test_home.py
```diff
@@ -56,6 +56,13 @@ def test_email_address_no_nag(self):
 
         assert 'add your email address' not in response
 
+    @helpers.change_config('ckan.legacy_route_mappings',
+                           '{"my_home_route": "home.index"}')
+    def test_map_pylons_to_flask_route(self):
+        app = self._get_test_app()
+        response = app.get(url_for('my_home_route'))
+        assert 'Welcome to CKAN' in response.body
```
amercader: Can you test if `url_for('home')` still works?

tino097:
merged to dev-v2.8

PR #4191, opened 2018-04-18
[#3993] package_autocomplete use solr

smotornyuk:
Alternative implementation of package_autocomplete. Contains schema chages -> requires rebuild of search index. 
I don't remember our strategy as for increasing solr version, so I updated it to 2.81. Everything else is pretty straightforward

### DIFF
--- a/ckan/config/solr/schema.xml
+++ b/ckan/config/solr/schema.xml
@@ -24,7 +24,7 @@
 <!-- We update the version when there is a backward-incompatible change to this
 schema. In this case the version should be set to the next CKAN version number.
 (x.y but not x.y.z since it needs to be a float) -->
-<schema name="ckan" version="2.81">
+<schema name="ckan" version="2.9">
 
 <types>
     <fieldType name="string" class="solr.StrField" sortMissingLast="true" omitNorms="true"/>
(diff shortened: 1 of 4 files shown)


ckan/config/solr/schema.xml
```diff
@@ -24,7 +24,7 @@
 <!-- We update the version when there is a backward-incompatible change to this
 schema. In this case the version should be set to the next CKAN version number.
 (x.y but not x.y.z since it needs to be a float) -->
-<schema name="ckan" version="2.8">
+<schema name="ckan" version="2.81">
```
amercader: This is targeting 2.9 so just use this as the new version

wardi:
The TestPackageAutocomplete class just needs another couple tests:

- can I autocomplete private datasets from my org?
- can I autocomplete with strings from the middle of the package name/title?

smotornyuk:
Updated and even found one problem - because i copied **fq='+capacity:public'**, permission labels had no effect. :)

smotornyuk:
ping @wardi

starlyrols:
OK
Envoyé depuis Yahoo Mail pour Android 
 
  Le ven., nov. 30, 2018 à 15:55, Sergey<notifications@github.com> a écrit :   

ping @wardi

PR #4192, opened 2018-04-18
[Security] Don't confirm whether a user exists or not

anotheredward:
Being able to enumerate users allows attackers to find valid username
which they can then try to guess/brute force the passwords for. This PR
makes the reset token request always return a "A reset token has been sent"
so attackers are unable to find out whether a user exists or not.

Attackers could also enumerate users through the users list, but upon
investigation it seems that the UK/US data portals either have this
disabled or behind a login. The NZ portal will soon be disabling this
listing page as well.

https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Authentication_and_Error_Messages

Fixes #
* Don't confirm whether a user exists or not on the reset form

### Proposed fixes:
* Remove any other places in the system where users can be enumerated from 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -485,12 +485,7 @@ class UserController(base.BaseController):
                         data_dict['id'] = user_list[0]['id']
                         user_dict = get_action('user_show')(context, data_dict)
                         user_obj = context['user_obj']
-                    elif len(user_list) > 1:
-                        h.flash_error(_('"%s" matched several users') % (id))
-                    else:
-                        h.flash_error(_('No such user: %s') % id)
-                else:
-                    h.flash_error(_('No such user: %s') % id)
+            helpers.flash_success(_('A reset token has been sent.'))
 
             if user_obj:
                 try:
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -485,7 +485,7 @@ class UserController(base.BaseController):
                         data_dict['id'] = user_list[0]['id']
                         user_dict = get_action('user_show')(context, data_dict)
                         user_obj = context['user_obj']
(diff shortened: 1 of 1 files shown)


ckan/controllers/user.py
```diff
@@ -485,12 +485,7 @@ def request_reset(self):
                         data_dict['id'] = user_list[0]['id']
                         user_dict = get_action('user_show')(context, data_dict)
                         user_obj = context['user_obj']
-                    elif len(user_list) > 1:
-                        h.flash_error(_('"%s" matched several users') % (id))
```
amercader: Can you log this (`log.info()`) and the other messages so maintainers can still track it?

amercader:
@anotheredward looks good, see minor comment

davidread:
I guess we should remove the "Please check your inbox" message too, as the absence of that indicates it is not a user?

A user might simply use a different email address, so we should be clear that it might not send in this case. So how about a message along the lines "A reset email has been sent (unless this is not a registered email address)"?

kmbn:
@anotheredward Are you still working on this PR? It looks like it's mergeable as soon as the logging is added.

kmbn:
@anotheredward I don't think I'll be able to get to it in the next week, so it would be great if you could update it. Ping me here or on Gitter if you run out of time.

davidread:
I've done an alternative PR that addresses the aims of this PR and covers a related change, to allow reset by email address: https://github.com/ckan/ckan/pull/4636

@anotheredward do let us know if you're happy to close this one in favour of the other.

anotheredward:
Collaborating on @davidread 's excellent work over at #4636 instead.
Closing this PR as it's now redundant :).

PR #4195, opened 2018-04-19
[#4194] Fix the "From" field with UTF-8 name and ASCII address in mail

### LINKED ISSUE #4194: lib/mailer: The header "From" with UTF-8 name and ASCII address

2.6.5

The "From" field (with UTF-8 name) in the header of emails sent by CKAN should be decoded and displayed by recipient's email client like:

From: "跨領域區域研究資料集" \<no-reply@data.depositar.io\>

The "From" field shows:

From: "跨領域區域研究資料集 \<no-reply@d"ata.depositar.io\>

Note that the latter double quote is in the wrong place, which ruins the field and brakes some email authentication methods, such as DKIM.

1. Set a site title with UTF-8 characters in SITE_URL/ckan-admin/config.
2. Let CKAN send a mail (invite someone via email, for example).

P.S. According the discussion in [this thread](https://stackoverflow.com/questions/41908883/how-to-add-an-utf-8-from-address-to-a-mail-mesage), we should assign the encoding to the name part and the address part respectively. I will submit a workable PR later.

u10313335:
Fixes #4194

### Proposed fixes:

According the discussion in [this thread](https://stackoverflow.com/questions/41908883/how-to-add-an-utf-8-from-address-to-a-mail-mesage), we should assign the encoding to the name part and the address part respectively.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -45,7 +45,8 @@ def _mail_recipient(recipient_name, recipient_email,
             msg.add_header(k, v)
     subject = Header(subject.encode('utf-8'), 'utf-8')
     msg['Subject'] = subject
-    msg['From'] = _("%s <%s>") % (sender_name, mail_from)
+    msg['From'] = Header(_("%s") % sender_name, 'utf-8')
+    msg['From'].append("<%s>" % mail_from, 'ascii')
     recipient = u"%s <%s>" % (recipient_name, recipient_email)
     msg['To'] = Header(recipient, 'utf-8')
     msg['Date'] = Utils.formatdate(time())
--- a/ckan/tests/lib/test_mailer.py
+++ b/ckan/tests/lib/test_mailer.py
@@ -3,6 +3,7 @@
 from nose.tools import assert_equal, assert_raises, assert_in, assert_raises
 from email.mime.text import MIMEText
 from email.parser import Parser
+from email.header import Header
 from email.header import decode_header
 import hashlib
 import base64
@@ -143,9 +144,12 @@ class TestMailer(MailerBase):
         msgs = self.get_smtp_messages()
         msg = msgs[0]
 
+        sender_name = config.get('ckan.site_title')
+        mail_from = config.get('smtp.mail_from')
+
         expected_from_header = '{0} <{1}>'.format(
-            config.get('ckan.site_title'),
-            config.get('smtp.mail_from')
+            Header(sender_name, 'utf-8'),
+            Header(mail_from, 'ascii')
         )
 
         assert_in(expected_from_header, msg[3])


u10313335:
Oops, it seems that some tests were failed. I will try to fix them.

wardi:
I think email headers are ascii-only. Unicode addresses are encoded like https://tools.ietf.org/html/rfc5234 (from https://tools.ietf.org/html/rfc6532#section-3.1 ) There should be a module that handles this for us we can use?

u10313335:
Thanks @wardi. But I think the [email.header.Header](https://docs.python.org/2/library/email.header.html#email.header.Header) class I am using encodes the from address well?

wardi:
@u10313335 you're right, I didn't read the change properly.

I'm confused now, though.. why were we using gettext on `_("%s <%s>")`? Let's get rid of that.

Why not change the From line to match the recipient one below instead?
```python
    from_ = u"%s <%s>" % (sender_name, mail_from)
    msg['From'] = Header(from_, 'utf-8')
```

According to the email.header.Header docs passing `'utf-8'` as the second argument does nothing (Header tries to encode unicode strings as us-ascii first regardless) but it's consistent with the rest of the code in the same function.

u10313335:
@wardi It seems that the problem is that the misplaced quotes put by recipient's email client. If we encode the all From line `%s <%s>`, email.header.Header will try to encode the whole line as utf-8 using [RFC2047 rules](https://tools.ietf.org/html/rfc2047.html#section-2) as long as possible. For example,

* The sender with shorter name `我是中文寄件者 <no-reply@localhost>` will be encoded as a single line `=?utf-8?b?5oiR5piv5Lit5paH5a+E5Lu26ICFIDxuby1yZXBseUBsb2NhbGhvc3Q+?=`. Then the recipient's email client will decode it as `"我是中文寄件者 <no-reply@localhost>"`

* The sender with longer name `我是比較長的中文寄件者 <no-reply@localhost>` will be encoded into two lines:
`=?utf-8?b?5oiR5piv5q+U6LyD6ZW355qE5Lit5paH5a+E5Lu26ICFIDxuby1yZXBseUBs?=`
`=?utf-8?q?ocalhost=3E?=`
Then the recipient's email client will decode it as `"我是比較長的中文寄件者 <no-reply@l"ocalhost>`. The quoted part is just the first line.

Note that **both situations produce the wrong From line** in recipient's email client. The correct From line should be `"我是中文寄件者" <no-reply@localhost>` and `"我是比較長的中文寄件者" <no-reply@localhost>`. Only the sender's name needs to be quoted.

The proposed PR (force the sender's address not to be encoded as utf-8 according to the [docs](https://docs.python.org/2/library/email.header.html#email.header.Header.append)) is the only workaround to bypass this problem as far as I know. The encoded From line of the first example will be `=?utf-8?b?5oiR5piv5Lit5paH5a+E5Lu26ICF?= <no-reply@localhost>`

P.S. 

1. I agree that we should remove the gettext function though. Like this:

  ```python
  msg['From'] = Header("%s" % sender_name, 'utf-8')
  msg['From'].append("<%s>" % mail_from, 'ascii')
  ```

2. It seems that the `msg['To']` line has the same issue.

tino097:
as the #4194 is closed as completed, this PR will be closed too

PR #4207, opened 2018-04-26
[#4031] More support for custom group types in templates

amercader:
Some bits and bobs that were missed out in #4032 to properly display custom group types in templates. Fixed the breadcrumbs in the dataset pages and the slug widget to use the custom types.


amercader:
There's a companion PR in ckanext-scheming: https://github.com/ckan/ckanext-scheming/pull/179

Issue #3567, opened 2017-05-18
Port feeds controller to Flask blueprint


amercader:
Once #3239 is merged (hopefully really soon after we branch out 2.7) we can start porting controllers to Flask blueprints and identify issues as we go.

One easy controller to get started is the feeds controllers, as it does not require template rendering, work can start on a branch from `3229-api-blueprint`.

The goal is too have a new blueprint `ckan.views.feeds.py`, modelled after `ckan.views.api.py`, with the same endpoints (URLs) and outputs.

After this first step, we can take the opportunity to migrate from `webhelpers.FeedGenerator` to something more maintained.

amercader:
@tino097 In general, the approach I would follow for migrating controllers is 

1. Familiarize yourself with how Blueprints are defined and registered (they are registered automatically if the module is on `ckan/views/`.
2. Make sure routes are registered and the Flask one is used instead of the Pylons one (comment the Pylons one on `ckan/config/routing.py`)
3. Start moving the first controller endpoint, plus all necessary imports and internal functions. Make sure the tests pass for it (if there are no tests, write one on `ckan/tests/controllers/`)
4. Move the rest of endpoints.

amercader:
Done in #3870

PR #4208, opened 2018-04-26
[#3567] Fix feed URL in group and org pages

amercader:
@tino097 this was missed in the flask feeds blueprint


tino097:
doesnt need to be in 2.7.4 as the feed blueprint was not part of it

PR #4215, opened 2018-04-30
restrict anonymous access to user information update

tino097:
Related to #3915 with small updates

PR #4225, opened 2018-05-03
403 exceptions with flask

tino097:
This PR comes from @smotornyuk comment to #4215 about handling the 403 exceptions from flask requests. 
Its a small workaround, by registering an errorhandler to the user blueprint and rendering 403 error page

### Features:

- [ ] includes API changes


ckan/tests/controllers/test_user.py
```diff
@@ -242,8 +242,9 @@ def test_user_edit_unknown_user(self):
         app = self._get_test_app()
         response = app.get(
             url_for('user.edit', id='unknown_person'),
-            status=403
-        )
+            status=200)
```
amercader: The status code should not change, it should still be 403. See below

ckan/views/user.py
```diff
@@ -28,6 +28,12 @@
 user = Blueprint(u'user', __name__, url_prefix=u'/user')
 
 
+@user.errorhandler(403)
+def error_handler(e):
+    extra_vars = {u'code': e.code, u'content': e.description}
+    return base.render(u'403.html', extra_vars)
```
amercader: This will return a 200 status, which is not what we want as we need to keep the original exception status code. To return the status code you can do the following ([docs](http://flask.pocoo.org/docs/1.0/quickstart/#about-responses)):

```python
return base.render(u'403.html', extra_vars), e.code
```

ckan/tests/controllers/test_user.py
amercader: So basically the tests should remain unchanged

ckan/views/user.py
```diff
@@ -28,6 +28,12 @@
 user = Blueprint(u'user', __name__, url_prefix=u'/user')
 
 
+@user.errorhandler(403)
```
amercader: We need to handle other exceptions as well, as currently any error returns a blank server page (400, 404, 500, etc). I think it's safer to explicitly add a decorator for each of the status codes returned by the blueprints.

ckan/templates-bs2/403.html
```diff
@@ -0,0 +1,21 @@
+{% extends "page.html" %}
```
amercader: Can we render the current `error_document_template.html`? See comment about capturing all errors

ckan/views/user.py
```diff
@@ -28,6 +28,12 @@
...
+def error_handler(e):
+    extra_vars = {u'code': e.code, u'content': e.description}
```
amercader: There's some extra logic we need to copy from the error controller, specially the one that bypasses templates rendering for the api and fanstatic, otherwise we'll get http errors on the api:

https://github.com/ckan/ckan/blob/master/ckan/controllers/error.py#L35

The rest I don't think we need for now

amercader:
Good stuff @tino097

wardi:
this doesn't override the error handling in the api controller, right?

tino097:
@wardi 
as @amercader mentioned to me, thats handled within api blueprint

amercader:
@wardi The API blueprint handles its own responses: https://github.com/ckan/ckan/blob/master/ckan/views/api.py#L75

PR #4229, opened 2018-05-04
Upgrade httpretty to 0.9.4 for Python 3 support

cclauss:
Helps with #3309

### Proposed fixes:
Upgrade development requirement [__httpretty__](https://github.com/gabrielfalcao/HTTPretty) from 0.8.3 to 0.9.4 for Python 3 support.

### Features:

- [ ] includes API changes

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -23,6 +23,7 @@ dependencies:
            && chmod +x ~/.local/bin/circleci-matrix"
 
     override:
+        - python --version ; pip --version
         - pip install -r requirement-setuptools.txt
         - pip install -r requirements.txt
         - pip install -r dev-requirements.txt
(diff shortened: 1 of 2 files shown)


amercader:
We've had lots of problems in the past with httpretty versions and tests that use Solr or the DataPusher, and sure enough there are already 2 failures related. It might just be that the tests need updating because of some change in httpretty, but in any case we need to get these fixed or rewritten before merging.

cclauss:
The root cause is that Circle CI is testing on Python 2.7.6 (current is 2.7.15) and the current httpretty is relying on [ssl.SSLContext which was added in Python 2.7.9](https://docs.python.org/2/library/ssl.html#ssl.SSLContext).

cclauss:
In CircleCI on the __pip install -r dev-requirements.txt__ step we are getting these warnings:
* httpretty 0.8.3 has requirement urllib3==1.7.1, but you'll have urllib3 1.23 which is incompatible.
* mox3 0.25.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.

PR #4236, opened 2018-05-07
datastore_search fix: total incorrect when distinct=true

wardi:
PR with test in progress


wardi:
not a regression, looks like it was never working correctly 6071824b9e

wardi:
@amercader anything holding this up?

PR #4243, opened 2018-05-08
Proposal: Extension support for highlighting

jbothma:
This is an initial proposal for adding support for extensions to invoke and use [solr highlighting/snippets](https://lucene.apache.org/solr/guide/6_6/highlighting.html)

I'm keen to hear your thoughts - whether this kind of thing is something you'd ever bring into CKAN, and how that might be done. I've pulled out what I could from my implementation overriding package_search and PackageQuery.run for plugins to carry their specific logic. I tried not to make it super specific to highlighting but perhaps it might be better to just focus on exposing highlighting?

The strategy is roughly:
- highlighting users might have diverse needs - I expect instances using highlighting to implement their own plugin to decide precisely what they want to expose and how.
- allow plugins to modify allowed solr query parameters (`update_valid_solr_parameters`)
- allow plugins to modify `fl` without invoking the pretty package output code which would then have to be re-implemented in the plugin (`fl_compatible`)
- allow plugins to see raw solr results (`include_raw_solr_results`)
  - this allows plugins to expose use what they need and include n their results as they wish
  - (I realised later it's not quite the raw json object you get back from solr directly - let's talk about the right way to expose the `highlighting` dict)
- expect plugins to then clean up the result from the raw solr response data (`search_results['raw_solr_results']`)

Fixes #

### Proposed fixes:
Enable use of solr highlighting/snippets for https://github.com/ckan/ideas-and-roadmap/issues/212

- 

### Features:

- [X] includes API changes

----

This would be used by a plugin looking something like

```python
class SATreasurySearchPlugin(plugins.SingletonPlugin):

    plugins.implements(plugins.IPackageController, inherit=True)

    def before_search(self, search_params):
        # log.info("before_search %r", search_params)
        if asbool(search_params.get('extras').get('ext_highlight')):
            search_params['hl'] = 'on'
            search_params['hl.fl'] = '*'
            search_params['hl.simple.pre'] = ''
            search_params['hl.simple.post'] = ''
            search_params['hl.snippets'] = 5
            search_params['hl.fragsize'] = 200
            # Make sure that matches where the query words are in close
            # proximity get higher ranking
            search_params['pf'] = ['name^4 title^4 tags^2 groups^2 text']
            # Request what package_search requests by default + index_id
            field_list = ['id', 'validated_data_dict', 'index_id']
            search_params['fl'] = search_params.get('fl', field_list)
            # Tell CKAN that the custom fl is compatible with its original
            # and it should behave like it's not modified
            search_params['extras']['fl_compatible'] = True

        return search_params

    def update_valid_solr_parameters(self, valid_solr_parameters):
        """
        Takes a set and returns a set
        """
        for param in HIGHLIGHTING_PARAMETERS:
            valid_solr_parameters.add(param)
        return valid_solr_parameters

    def include_raw_solr_results(self, search_params):
        if asbool(search_params.get('extras').get('ext_highlight')):
            return True
        else:
            return False

    def after_search(self, search_results, search_params):
        # log.info("after_search %r %r", search_results, search_params)
        if asbool(search_params.get('extras').get('ext_highlight')):
            ckan_results = search_results['results']
            solr_results = search_results['raw_solr_results']['results']
            highlighting = search_results['raw_solr_results']['highlighting']
            assign_highlighting(ckan_results, solr_results, highlighting)

            # Clean up the stuff we needed for assigning highlighting
            del search_results['raw_solr_results']

        return search_results


RESOURCE_RE = re.compile('^ckanext-extractor_([a-z0-9_-]+)_fulltext$')


def assign_highlighting(result_packages, solr_results, highlighting):
    for idx, solr_result in enumerate(solr_results):

        index_id = solr_result['index_id']
        package_highlighting = highlighting[index_id]
        package_result = result_packages[idx]
        package_result['highlighting'] = package_highlighting

        # Initialise resource highlighting key
        for resource in package_result['resources']:
            resource['highlighting'] = {}

        # Move resource highlighting to the resource
        for key in package_highlighting.keys():
            match = RESOURCE_RE.match(key)
            if match:
                resource_id = match.group(1)
                highlights = package_highlighting.pop(key)
                for resource in package_result['resources']:
                    if resource['id'] == resource_id:
                        resource['highlighting']['fulltext'] = highlights
```

Here's how I'm prototyping this https://github.com/OpenUpSA/ckanext-satreasury/pull/18/files

### DIFF
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -135,6 +135,7 @@ class SearchQuery(object):
     def __init__(self):
         self.results = []
         self.count = 0
+        self.highlighting = {}
 
     @property
     def open_licenses(self):
@@ -285,7 +286,6 @@ class PackageSearchQuery(SearchQuery):
         else:
             return solr_response.docs[0]
 
-
     def run(self, query, permission_labels=None, **kwargs):
         '''
         Performs a dataset search using the given query.
@@ -302,8 +302,13 @@ class PackageSearchQuery(SearchQuery):
         '''
         assert isinstance(query, (dict, MultiDict))
         # check that query keys are valid
-        if not set(query.keys()) <= VALID_SOLR_PARAMETERS:
-            invalid_params = [s for s in set(query.keys()) - VALID_SOLR_PARAMETERS]
+        valid_solr_parameters = VALID_SOLR_PARAMETERS
+        for item in plugins.PluginImplementations(plugins.IPackageController):
+            if 'update_valid_solr_parameters' in dir(item):
+                valid_solr_parameters = item.update_valid_solr_parameters(valid_solr_parameters)
+
+        if not set(query.keys()) <= valid_solr_parameters:
+            invalid_params = [s for s in set(query.keys()) - valid_solr_parameters]
             raise SearchQueryError("Invalid search parameters: %s" % invalid_params)
 
         # default query is to return all documents
@@ -384,7 +389,7 @@ class PackageSearchQuery(SearchQuery):
                               (query, e))
         self.count = solr_response.hits
         self.results = solr_response.docs
-
+        self.highlighting = solr_response.highlighting
 
         # #1683 Filter out the last row that is sometimes out of order
         self.results = self.results[:rows_to_return]
@@ -408,7 +413,12 @@ class PackageSearchQuery(SearchQuery):
         for field, values in six.iteritems(self.facets):
             self.facets[field] = dict(zip(values[0::2], values[1::2]))
 
-        return {'results': self.results, 'count': self.count}
+        query_response = {
+            'results': self.results,
+            'count': self.count,
+        }
+
+        return query_response
 
 
 def solr_literal(t):
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1864,7 +1864,7 @@ def package_search(context, data_dict):
         # Add them back so extensions can use them on after_search
         data_dict['extras'] = extras
 
-        if result_fl:
+        if result_fl and not extras['fl_compatible']:
             for package in query.results:
                 if package.get('extras'):
                     package.update(package['extras'] )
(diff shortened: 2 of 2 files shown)


ckan/lib/search/query.py
```diff
@@ -302,8 +302,13 @@ def run(self, query, permission_labels=None, **kwargs):
         '''
         assert isinstance(query, (dict, MultiDict))
         # check that query keys are valid
-        if not set(query.keys()) <= VALID_SOLR_PARAMETERS:
-            invalid_params = [s for s in set(query.keys()) - VALID_SOLR_PARAMETERS]
+        valid_solr_parameters = VALID_SOLR_PARAMETERS
+        for item in plugins.PluginImplementations(plugins.IPackageController):
+            if 'update_valid_solr_parameters' in dir(item):
```
jbothma: This is only needed because I haven't added fallback implementation to IPackageController yet

ckan/logic/action/get.py
```diff
@@ -1889,18 +1889,34 @@ def package_search(context, data_dict):
...
         'count': count,
         'facets': facets,
         'results': results,
-        'sort': data_dict['sort']
+        'sort': data_dict['sort'],
     }
 
+    include_raw_solr_results = False
+    for item in plugins.PluginImplementations(plugins.IPackageController):
+        if 'include_raw_solr_results' in dir(item):
```
jbothma: This is only needed because I haven't added fallback implementation to IPackageController yet

wardi:
Sounds like a great feature to me. @TkTech do you have any thoughts about the implementation?

TkTech:
I'm in favor of enabling more control over the search interface, and we should be able to allow everything needed for highlighting generically. It seems like the only functionality we really need is the `include_raw_solr_results`. There should be no need for `valid_solr_parameters`. Queries edited by plugins should be done *after* API/UI validation.

Lets discuss this on Thursday.

jbothma:
`update_valid_solr_parameters` is needed because `VALID_SOLR_PARAMETERS` is checked after `before_search`. Is there another way for plugins to extend the possible solr parameters at the moment?

TkTech:
I'm suggesting we change that behavior to simplify the interface.

jbothma:
Hi there!

What was the outcome of your planning meeting? What's the next step here?

kmbn:
@amercader @wardi Was there ever a decision about how @jbothma can proceed with this PR?

jbothma:
I'm about to upgrade so was getting tests running for my fork anyway -
happy to supply tests for an implementation.

Best
JD

tino097:
This PR is older than 5 years and some of the changes could not apply or have been fixed within other PR's

PR #4250, opened 2018-05-11
autocomplete tagsep - fix

hvwaldow:
### Proposed fixes:

Input of tags automatically separates tags (starts a new tag) when a comma is typed. However this doesn't work if a string containing commas is pasted. This severely slows down dataset submission for use-cases where tags or keywords are already available from another source.

This PR fixes that by making use of the [tokenSeparators option](http://select2.github.io/select2/#doc-tokenSeparators) of select2. It should work back to CKAN v2.3 and not break earlier versions. Additonally, other separators can be specified by passing `data-module-tagsep="[',', ';', '|']"`, for example.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/autocomplete.js
+++ b/ckan/public/base/javascript/modules/autocomplete.js
@@ -2,14 +2,16 @@
  * a list of terms from an API endpoint (provided using data-module-source).
  *
  * source   - A url pointing to an API autocomplete endpoint.
- * interval - The interval between requests in milliseconds (default: 1000).
+ * interval - The interval between requests in milliseconds (default: 300).
  * items    - The max number of items to display (default: 10)
  * tags     - Boolean attribute if true will create a tag input.
  * key      - A string of the key you want to be the form value to end up on
  *            from the ajax returned results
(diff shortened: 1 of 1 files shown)


ckan/public/base/javascript/modules/autocomplete.js
```diff
@@ -50,7 +53,8 @@ this.ckan.module('autocomplete', function (jQuery) {
         formatNoMatches: this.formatNoMatches,
         formatInputTooShort: this.formatInputTooShort,
         dropdownCssClass: this.options.dropdownClass,
-        containerCssClass: this.options.containerClass
+        containerCssClass: this.options.containerClass,
+        tokenSeparators: eval(this.options.tokensep)
```
smotornyuk: Could you please parse this options manually? Inside constructor, for example? I understand, that eval doesn't do any  harm here, but anyway no one like `eval` :)

hvwaldow: I see. I took the opportunity to change the syntax of the parametrization (value of `data-module-tagsep`) to something simpler. Now only single characters are possible as separation markers, but that is OK, I believe.

PR #4257, opened 2018-05-16
Error handler update

tino097:
We have issues with non HTTPExceptions and exceptions from extensions. Also small change for reading the `debug` configuration value

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -416,5 +416,5 @@ def _register_error_handler(app):
 
     for code in default_exceptions:
         app.register_error_handler(code, error_handler)
-    if not app.debug:
+    if not app.debug and not app.testing:
         app.register_error_handler(Exception, error_handler)
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -410,9 +410,10 @@ def _register_error_handler(app):
         if isinstance(e, HTTPException):
             extra_vars = {u'code': [e.code], u'content': e.description}
             return base.render(u'error_document_template.html', extra_vars), e.code
-        extra_vars = {u'code': 500, u'content': e.message}
+        extra_vars = {u'code': [500], u'content': u'Internal server error'}
         return base.render(u'error_document_template.html', extra_vars), 500
 
     for code in default_exceptions:
         app.register_error_handler(code, error_handler)
-    app.register_error_handler(Exception, error_handler)
+    if not app.debug:
+        app.register_error_handler(Exception, error_handler)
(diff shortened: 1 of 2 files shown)


tino097:
we need to add `try catch` blocks in some of the `helpers.py` functions 

For example for `check_access` as we have situation where `c.user` is throwing an `AttributeError` coming from extension

tino097:
Currently im testing few extensions against 2.8 and there are few scenarios where error handler is not working as it should.

First example is by enabling of `ckanext-pages` extension and when the instance is starting, the following [error](https://gist.github.com/tino097/d208a299fbbf2abb18ac21fb6dba6f10#file-pages-L103) is occurring
This error is result of the home blueprint endpoint and plugin is looking for pylons `c.action` attribute instead

Second example is by enabling `ckanext-ldap` extension and the error stack is [here](https://gist.github.com/tino097/e4ff24f67d5b88a41f2e66508615c941#file-ldap-L87)
In this case, while trying to display error page, `check_access` in `helpers.py` is trying to get `c.user` but because the extension had failed, user is not set

amercader:
OK, I had a closer look at this, here are some points to help clarify the situation:

* When running with `debug=True`, general uncaught exceptions should not be handled by Flask/Pylons. They should be allowed to raise so they can be picked up by the interactive debugger. By interactive debugger I mean this:
  * Pylons
screenshot-2018-6-12 server error

  * Flask 
      screenshot-2018-6-11 valueerror norr yo werkzeug debugger

  
* For Flask requests, I wrongly assumed that FlaskDebugToolbar was the responsible for showing the interactive debbuger, but it isn't. This is generated by Werkzeug, but only if running on Werkzeug of course! When running under paster and serving a Flask request the exception won't get handled and we will see a plain server error page like the cases @tino097 mentions in his comment:
   
    screenshot-2018-6-12 http 127 0 0 1

I think this is fine until we get finish #4921 and switch to run the local development server with Werkzeug, thus enabling the debugger for Flask requests. 

* To cover the `Return 500 status code + Return JSON error object showing "Internal Server Error" (no error details)` on API calls you will need something like this:

 ```diff
diff --git a/ckan/views/api.py b/ckan/views/api.py
index 8873f74..e3f776c 100644
--- a/ckan/views/api.py
+++ b/ckan/views/api.py
@@ -336,6 +336,13 @@ def action(logic_function, ver=API_DEFAULT_VERSION):
                        str(e)}
         return_dict[u'success'] = False
         return _finish(500, return_dict, content_type=u'json')
+    except Exception as e:
+        return_dict[u'error'] = {
+            u'__type': u'Internal Server Error',
+            u'message': u'Internal Server Error'}
+        return_dict[u'success'] = False
+        log.exception(e)
+        return _finish(500, return_dict, content_type=u'json')
 
     return _finish_ok(return_dict)
```

PR #4280, opened 2018-06-06
Add mention of the security@ckan.org email address in issue template

amercader:


TkTech:
We should probably do it the new way (https://help.github.com/articles/about-issue-and-pull-request-templates/) which let's us give multiple issue type buttons that a user mist explicitly click on. This one could be "Security Issue", "Bugfix", "New Feature", yada yada. I've found the one-big-template-for-everything gets messy after awhile.

davidread:
Could idea @TkTech. I'd be tempted to merge this PR in the meantime though, as it is an improvement, and the categories may need discussion.

kmbn:
Like @davidread wrote, this PR would be an improvement and we haven't gotten around to discussing a new category-based template in the last six months. Let's merge this.

PR #4282, opened 2018-06-07
Remove storage controller

tino097:
As the storage controller is not used after 2.2 its safe to be removed


tino097:
Should i remove this [db task](https://github.com/ckan/ckan/blob/master/ckan/lib/cli.py#L475) with this PR or create a new one ?

PR #4286, opened 2018-06-07
[#2293] Replace httpie doc examples with CURL examples

### LINKED ISSUE #2293: API docs should recommend ckanapi
The API docs currently have examples with:
- httpie (general use)
- urllib2 (python)
  http://docs.ckan.org/en/latest/api/index.html#making-an-api-request

In the meeting today we decided these would be better:
- curl (general use)
- ckanapi (python)
- urllib2 (python)

rossjones:
As per #2293 we replace the httpie examples with the equivalent in curl.
It appears that the examples were using http and therefore were getting
a 302 to the https url, and as we all know to our pain, POST data
doesn't follow redirects and so the examples weren't always working.
Have changes some occurrences of http://demo.ckan.org to
https://demo.ckan.org

Fixes #2293

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -79,7 +79,7 @@ of all the datasets in the ``data-explorer`` group on demo.ckan.org, install
 curl and then call the ``group_list`` API function by running this command
 in a terminal::
 
-    curl http://demo.ckan.org/api/3/action/group_list
+    curl https://demo.ckan.org/api/3/action/group_list
 
 The response from CKAN will look like this::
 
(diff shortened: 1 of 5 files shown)


doc/api/index.rst
```diff
@@ -73,13 +73,13 @@ To call the CKAN API, post a JSON dictionary in an HTTP POST request to one of
...
-client `HTTPie <http://httpie.org/>`_.  For example, to get a list of the names
+One way to post a JSON dictionary to a URL is using the command-line 
+client `Curl <https://curl.haxx.se/>`_.  For example, to get a list of the names
 of all the datasets in the ``data-explorer`` group on demo.ckan.org, install
-HTTPie and then call the ``group_list`` API function by running this command
+curl and then call the ``group_list`` API function by running this command
 in a terminal::
 
-    http http://demo.ckan.org/api/3/action/group_list
+    curl http://demo.ckan.org/api/3/action/group_list
```
davidread: missed one ;-) - needs to be 'https'

rossjones: Should be fixed now.

Issue #4161, opened 2018-03-30
tests failing without --segments


wardi:
### CKAN Version if known (or site URL)
master

### Please describe the expected behaviour
tests pass when run without --segments

### Please describe the actual behaviour
```
FAIL: ckan.tests.legacy.functional.api.test_activity.TestActivity.test_add_tag
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/project/ckan/ckan/tests/legacy/functional/api/test_activity.py", line 1583, in test_add_tag
    assert len(details) == 1
AssertionError
```

### What steps can be taken to reproduce the issue? 
run nosetests on master

Bisecting lead to commit a727c6373576702ff9f5226cf2a9ba5096e3c843 as the first bad commit

Issue #4088, opened 2018-03-08
Test failures in master in a fresh CKAN install


amercader:
Brand new virtualenv, fresh CKAN source install:

```
nosetests --nologcapture --with-pylons=test-core.ini -v ckan ckanext
```

gives:

```
----------------------------------------------------------------------
Ran 2511 tests in 1069.050s

FAILED (SKIP=10, errors=11, failures=14)


CKAN_SQLALCHEMY_URL in the env takes precedence over CKAN_DB ... ERROR
Setting an env var from the whitelist will set the appropriate option ... ERROR
ckan.tests.test_common.TestCommonRequest.test_params_also_works_on_flask_request ... ERROR
ckanext.reclineview.tests.test_view.TestReclineGraphView.test_title_description_iframe_shown ... ERROR
ckanext.reclineview.tests.test_view.TestReclineGridView.test_title_description_iframe_shown ... ERROR
ckanext.reclineview.tests.test_view.TestReclineMapView.test_title_description_iframe_shown ... ERROR
ckanext.reclineview.tests.test_view.TestReclineView.test_title_description_iframe_shown ... ERROR
ckanext.reclineview.tests.test_view.TestReclineViewDatastoreOnly.test_create_datastore_only_view ... ERROR
ckanext.textview.tests.test_view.TestTextView.test_js_included ... ERROR
ckanext.textview.tests.test_view.TestTextView.test_title_description_iframe_shown ... ERROR
ckanext.webpageview.tests.test_view.TestWebPageView.test_view_shown_on_resource_page ... ERROR
Test add tag activities. ... FAIL
Test deleted package extra activity stream. ... FAIL
ckan.tests.legacy.functional.api.test_package_search.TestPackageSearchApi1.test_05_uri_json_tags ... FAIL
ckan.tests.legacy.functional.api.test_package_search.TestPackageSearchApi1.test_06_uri_q_tags ... FAIL
ckan.tests.legacy.functional.api.test_package_search.TestPackageSearchApi2.test_05_uri_json_tags ... FAIL
ckan.tests.legacy.functional.api.test_package_search.TestPackageSearchApi2.test_06_uri_q_tags ... FAIL
ckan.tests.legacy.functional.api.test_package_search.TestPackageSearchApi3.test_05_uri_json_tags ... FAIL
ckan.tests.legacy.functional.api.test_package_search.TestPackageSearchApi3.test_06_uri_q_tags ... FAIL
Test user activity streams HTML rendering. ... FAIL
[...]

Issue #4060, opened 2018-03-01
[Python3] - CI


deinok:
ref:  #3309

The transition to Python3 and support of Python2 will require changes to the CI system.

I would recommend using CircleCI V2: https://circleci.com/docs/2.0/
And follow the following steps:

# Roadmap:
- [ ] Step 1: Port CircleCI V1 file to CircleCI V2
- [ ] Step 2: Enable CircleCI V2 Workflows https://circleci.com/docs/2.0/workflows/
- [ ] Step 3: Add Python 2 as a Workflow -> `python2`
- [ ] Step 4: Add Docker-Python2 test as a Workflow -> `docker-python2`
- [ ] Step 5: Remove TravisCI as it won't be required at this moment.
- [ ] Step 6: Add Python 3 as a Workflow -> `python3`. This test will fail until completely ported to Python 3. Set an easy mechanism to enable and disable this workflow. Disabled by default.
- [ ] Step 7: Add Docker-Python3 test as a Workflow. This test will fail until completely ported to Python 3. Set an easy mechanism to enable and disable this workflow. Disabled by default.
- [ ] Step 8: Python 3 is supported.
- [ ] Step 9: Set `python3` Workflow enabled by default.
- [ ] Step 10: Set `docker-python3` Workflow enabled by default.

# Related Issues:
Closes #4227 #4213

deinok:
@wardi I shouldn't require anything fixed until Step 3 is finished. So I will wait for your confirmation that everything is okey with this steps, and start working on it ;)

cclauss:
I would be interested to see [this test}(https://github.com/ckan/ckan/pull/4054/files#diff-354f30a63fb0907d4ad57269548329e3R19) added to the current __.travis.yml__ or __circle.yml__ so that we do not see backsliding on syntax errors and undefined names.  It would be especially valuabe if we could also get it [running on both Python 2 and Python 3](https://travis-ci.org/ckan/ckan/builds/347762819?utm_source=github_status&utm_medium=notification) quickly.

deinok:
@cclauss LGTM, im not expert in Python Linting but i really care about the style.
Adding it to the CI should not be a problem, this can be added inside the `python2` and `python3` workflows.
@wardi Are you agree with adding a lint system?
Also, is flake8 the recommended tool?

deinok:
As a reference of other linting tools for python:
* Pylint
* Pep8
* Flake8
* Pyflakes
* Pychecker
* Mypy
* PyLama

[...]

PR #4290, opened 2018-06-09
CI - IMPORTANT BUG

deinok:
Segmentation of Test have lead to don't trigger some tests.
This PR "fixes" the problem.

So, this PR would not pass the PR checks cause it is triggering tests that are now executed and not before this.

In order to fix this, we must do the following:
- Disable CircleCI parallelization.
- Skiping or fixing the 13 new tests that fail.

Even with this, this should be merged

### DIFF
--- a/.circleci-matrix.yml
+++ b/.circleci-matrix.yml
@@ -1,9 +1,0 @@
-env:
-- SEGMENTS="01234"
-- SEGMENTS="567"
-- SEGMENTS="89ab"
-- SEGMENTS="cdef"
-
-command:
-- mkdir -p $CIRCLE_TEST_REPORTS/nose
-- nosetests --ckan --reset-db --with-pylons=test-core.ini --nologcapture --with-coverage --cover-package=ckan --cover-package=ckanext --with-xunit --xunit-file=$CIRCLE_TEST_REPORTS/nose/junit.xml --segments=$SEGMENTS ckan ckanext
--- a/circle.yml
+++ b/circle.yml
@@ -58,8 +58,8 @@ database:
 test:
 
     override:
-        - circleci-matrix:
-            parallel: true
+        - mkdir -p $CIRCLE_TEST_REPORTS/nose
+        - nosetests --ckan --reset-db --with-pylons=test-core.ini --nologcapture --with-coverage --cover-package=ckan --cover-package=ckanext --with-xunit --xunit-file=$CIRCLE_TEST_REPORTS/nose/junit.xml ckan ckanext
 
     post:
         - paster serve test-core.ini:


amercader:
@deinok you are right, when run all tests sequentially there are failure, see https://github.com/ckan/ckan/issues/4088 Someone needs to go back in time and use `git bisect` to identify at what commit did this issues appear and try to fix them.

I agree this needs to be fixed but it should not work you to go ahead with migration to CircleCI 2.0 right? You can set up the workers or whatever is called using the same circle-matrix setup currently in used right?
Or if that's not possible work on the migration even with these [16 failures](https://circleci.com/gh/deinok/ckan/426) in place until we find a way to fix them.

deinok:
No, I would like to avoid using circle-matrix, it leads to problems.

Also, the final version its going to use the 4 instances at the same time, Python2, Python3, and the two Dockers

deinok:
If you could skip this tests for now (temporarily), i could send a PR with the new config for V2.
And reach Step 3 of #4060

wardi:
If we split the containers the way you suggest our python 2 tests will go from ~6 minutes to ~15 minutes to complete, and the python3 and docker containers will be done and idle almost right away.

How about keeping the tests split up and add the python3 and docker tests at the end?

deinok:
The point is that we can set parallelization in the tests.
With the aproach you are asking the complexity grows, and the time to run all the workflows grows also. (Time to setup environment)

amercader:
Can we keep the docker builds in Travis and for now run the python2 tests in 3 workers and the python3 checks in the 4th one?

I'm not completely opposed to running all tests in the same worker but I agree we should explore all other alternatives first, as the time spent really does make a difference.

deinok:
That will add complexity to the build, and also the bigest fear I have with the matrix system is a regresion of  this issue.

About paralelization, I think we can add something like this: http://nose.readthedocs.io/en/latest/doc_tests/test_multiprocess/multiprocess.html

For readonly tests it will improve the overall speed. I don't know how complex will be to add this to the actual codebase

wardi:
multiprocess tests against the same db is *more* complicated than splitting the tests across containers and could introduce other timing-related false-positives.

IIUC the failures when we run tests serially now are related to state in our flask app leaking from one test to another. That's just a bug in the way we set up/run our tests that we need to fix. Once we nail down the state that's leaking we can add a small regression test that can also be run independently, instead running all our tests serially in CI.

Which issue is this PR focused on: Fixing the failing tests or moving to circle V2?

deinok:
The PR is focused in notice that some tests are skiping due a missconfiguration of CI.
Is related to V2 cause I discovered this while I was porting the CI to V2.

> multiprocess tests against the same db is more complicated than splitting the tests across containers and could introduce other timing-related false-positives.

I agree with this. Even with spliting containers or not, multiprocess tests will improve speed of tests. I normaly use ReadWriteMutex for this kind of cases.

The case is that in a CI environment I think we should prefere reliability over speed. (Also, notice that in V2, matrix are implemented as workflows)

wardi:
Oh, are we completely skipping some tests right now? I thought we had some flask state leaking.

deinok:
@wardi Well, I'm not completely sure but I expect that the problem is the `--segments=$SEGMENTS`.
Like there is a range of tests that dont get inside any segment.

wardi:
here's a few of the failed tests above run on their own against master:

```
$ nosetests --ckan --reset-db --with-pylons=test-core.ini --nologcapture ckan.tests.test_common:TestCommonRequest.test_params_also_works_on_flask_request
.
Ran 1 test in 0.254s

OK
```

```
$ nosetests --ckan --reset-db --with-pylons=test-core.ini --nologcapture ckanext.example_iuploader.test.test_plugin:TestExampleIUploaderPlugin.test_resource_download_iuploader_called
.
Ran 1 test in 10.271s

OK
```

```
$ nosetests --ckan --reset-db --with-pylons=test-core.ini --nologcapture ckanext.reclineview.tests.test_view:TestReclineGridView.test_title_description_iframe_shown
.
Ran 1 test in 9.365s

OK
```

tino097:
hello @deinok, here are the failing tests and the log, when executed locally
https://gist.github.com/tino097/997229b880a173109189a96a4a9ef4b8

#4088

deinok:
@tino097 Yep, I want to see if its environment related or as @wardi said a test leak.

wardi:
Yes, I see the same failures as @tino097 when running all the tests

deinok:
Then clearly you have the reason @wardi its some kind of leak in the tests.
So, how we fix this?

wardi:
Start with git bisect to see what commit caused the issue. Recently I had to do this for a similar test failure when not using --segments: #4161 and found that the cause was a test written in a way that didn't unloading a plugin properly (fix was 312893cd4a8)

tino097:
As we talk today on the meeting, git bisect is really pain in the env :) because of DB and SOLR reconfiguring on every bisect but i will give it a another try these days

wardi:
shouldn't be too bad if you start from the e3c5fb7 (merging my earlier fix) that *should* have passing tests and isn't from a really old version

amercader:
Setting the tests failures aside (which I'm sure we'll get fixed soon) I'd really like for the Circle CI migration to go ahead in the meantime.
Given that we have two tools at our disposal why don't we combine them to make the best use o them:

* In Circle CI we use the 4 workflows available (I assume that's what we get in the free plan) to split the tests in the same way that we are doing now with circle-matrix
* In Travis we use their matrix feature to run the py2 docker build, p3 tests and py3 docker build. 

Is that possible? WDYT @deinok @wardi ?

wardi:
@amercader I like that approach, we keep getting the results we need back ASAP that way.

Circle V2 does report back separately on each workflow IIUC, so it might also be possible to move the other jobs there without delaying the main tests, but your way sounds like the simplest change.

deinok:
@amercader 

> In Circle CI we use the 4 workflows available (I assume that's what we get in the free plan)

Yes, we have 4 instances at the same time

> to split the tests in the same way that we are doing now with circle-matrix.

I dislike the idea of paralelize the tests. (But we can discuss this in another issue and dont block this).
Even accepting the fact of paralelize the tests, I should avoid having the matrix file. (It can be done without it).

> In Travis we use their matrix feature to run the py2 docker build, p3 tests and py3 docker build.

I dislike the idea of multiple CIs if we can do things with one. (Like the paralelization, we can discuss this in another issue.)

> Setting the tests failures aside (which I'm sure we'll get fixed soon) I'd really like for the Circle CI migration to go ahead in the meantime.

@amercader @wardi Could sombody skip that tests so I can commit the port of V2 with paralelization?

deinok:
Dont exec that tests.
Nevermaid, i will try paralelize an see if still failing, and report back

amercader:
@deinok did you manage to make progress with the Circle v2.0 migration?

deinok:
No, this test are blocking pass the CI build.
Nevermind, executin tests syncronus must work and pass. Paralelization should be optional.

tino097:
@deinok There is a merged fix #4390, for the tests, can you please check if everything is ok?

deinok:
https://circleci.com/gh/deinok/ckan/439

deinok:
FAILED (SKIP=3, errors=1, failures=3)

deinok:
Moving discussion to https://github.com/ckan/ckan/pull/4394

PR #4296, opened 2018-06-20
Fixing API Examples

sminez:
From working with a CKAN instance for the first time (2.5.2) it looks like the examples given here regarding file upload do not work. The request is rejected if the "url" key is not set and it needs to be set to an empty string for the file upload to work. If not, the following error is returned:

```
{"help": "http://.../api/3/action/help_show?name=resource_create", "success": false, "error": {"url": ["Missing value"], "__type": "Validation Error"}}
```

Setting the "url" key to anything other than an empty string seems to be accepted but the file contents are ignored and the resources is treated as an external link.

This commit updates the curl and requests examples to set the "url" and "name" keys appropriately.

### Fixes
Documentation fix for file upload API example.

### Proposed fixes:
N/A

### Features:

- [ ] includes API changes

### DIFF
--- a/doc/maintaining/filestore.rst
+++ b/doc/maintaining/filestore.rst
@@ -81,7 +81,7 @@ For example, to create a new CKAN resource and upload a file to it using
 
 .. parsed-literal::
 
- curl -H'Authorization: your-api-key' 'http://yourhost/api/action/resource_create' --form upload=@filetoupload --form package_id=my_dataset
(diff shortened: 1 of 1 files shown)


sminez:
Ah sorry, I've made this against the 2.8 branch and not the 2.5. I don't have access to a CKAN 2.8.X instance to check if this is a required change there as well.

amercader:
@sminez I'm happy to update the examples for the 2.5 version docs, but we should not add the param if it's not needed to newer versions. The `url` param is not required starting from CKAN 2.6.

amercader:
Fixed on the 2.5 docs: http://docs.ckan.org/en/2.5/maintaining/filestore.html
Thanks @sminez !

PR #4315, opened 2018-06-29
[WIP] Stop package_create polluting the context it is passed.

rossjones:
### Proposed fixes:

Although there are bigger discussions around mutable context (see
https://github.com/ckan/ideas-and-roadmap/issues/53 for more details)

`package_create`, at the end of the function, pollutes the context with
the package model and the id of the newly created package, just to call
another function.

As is done elsewhere in this function, this PR copies the context and
uses that in order to avoid issues with context dict re-use, until the
wider issue can be resolved.

### Features:

- [ ] includes API changes


ckan/logic/action/create.py
```diff
@@ -225,15 +225,18 @@ def package_create(context, data_dict):
         model.repo.commit()
 
     # need to let rest api create
-    context["package"] = pkg
     # this is added so that the rest controller can make a new location
-    context["id"] = pkg.id
+    api_context = context.copy()
+    api_context["package"] = pkg
+    api_context["id"] = pkg.id
```
wardi: let's just remove this instead, as you discussed

rossjones: 👍

rossjones:
@wardi I think this is done now.  I realise it is a tiny step and doesn't really solve the context problem, but it does remove at least one pain-point.

PR #4317, opened 2018-06-29
[#4316] Redirect /dataset/[id] -> /dataset/[name] etc

### LINKED ISSUE #4316: Redirect /dataset/[id] -> /dataset/[name] etc
Redirects:
* /dataset/[id] -> /dataset/[name]
* /group/[id] -> /group/[name]
* /organization/[id] -> /organization/[name]

Specifying these objects by *id* is hidden functionality really. There are no links to it. I've added it for two reasons:

1. As a developer it's occasionally I know a dataset's id and want to look at it, and it is useful to have it instead display on the normal url with the name in it. Why would you want the current functionality and keep it as the weird id url anyway? It's functionality I've wanted on many occasions over the years - maybe it's just me!

2. In Activity Stream the links to old versions of datasets/groups/organizations are currently given using the 'name' of those objects *as they were at the time of the change*. Obviously there are ways that these objects can change name, and in this case the link gets broken. So with my changes in PR https://github.com/ckan/ckan/pull/3972 I'm keen to change the link to be the id of the objects, so we avoid the broken link problem, and we don't add a whole bunch of joins on these queries that are potentially already quite big.

davidread:
Fixes #4316

(I didn't redirect user too as I don't need this for #3972  - the template for displaying a user does a lookup of the user in the db anyway. And converting a user id to a user name reveals personal info, which is probably the wrong trend in ckan, so I'd rather not do it purely for completeness sake.)

### Features:

- [ ] includes API changes


ckan/tests/controllers/test_organization.py
```diff
@@ -82,21 +82,35 @@ def test_error_message_shown_when_no_organization_list_permission(self, mock_che
...
+        response = app.get(url_for(controller='organization', action='read',
+                                   id=org['id']),
+                           status=302)
+        # redirect replaces the ID with the name in the URL
+        redirected_response = response.follow()
+        expected_url = url_for(controller='organization', action='read',
+                               id=org['name'])
+        assert_equal(redirected_response.request.path, expected_url)
+
+    # def test_organization_read(self):
```
rossjones: Worth removing is no longer necessary?  Or was an accidental commentout?

davidread: good spot, I'll remove

ckan/controllers/package.py
```diff
@@ -388,6 +388,11 @@ def read(self, id):
         except (NotFound, NotAuthorized):
             abort(404, _('Dataset not found'))
 
+        # if the user specified a package id, redirect to the package name
+        if data_dict['id'] == c.pkg_dict['id']:
+            h.redirect_to(controller='package', action='read',
+                          id=c.pkg_dict['name'])
```
wardi: Our site uses the same values for id and name, so this change would cause a redirect loop for us.

davidread: Good catch! I'll fix this.

davidread:
@wardi Comments are now all addressed now and conflicts resolved.

wardi:
Now that the package views are merged does this need to be implemented on the flask side too?

davidread:
It is already!

PR #4319, opened 2018-07-03
[#4310] Add global properties `controller` and `action`

### LINKED ISSUE #4310: Missing c.action attribute in 2.8.0 templates
It seems that in CKAN 2.8.0, the `c.action` attribute is at least sometimes missing when templates are being rendered. I can reproduce this for the `home/index.html` template, which currently breaks ckanext-pages for CKAN 2.8.0 (see ckan/ckanext-pages#78).

It seems that this is a consequence of [a change to `routing.py`](https://github.com/ckan/ckan/pull/3891/files#diff-eb995cd7c3a1045a463f8172c7adb128) as part of #3891, although I'm not sure about that.

smotornyuk:
fixes #4310.

 There are a bunch of extensions that rely on this props. This PR
is attempting to add them inside `before_each_request` hook. If
url.endpoint contains a dot, then part before dot will be considered
as controller and part after the dot - as action. otherwise(routes
without the dot, like test /hello route), view name is used for both,
controller and action name

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,15 @@
 Changelog
 ---------
 
+v.2.9.0 TBA
+==================
+
+Deprecations:
+ * `c.action` and `c.controller` variables should be avoided. Refer to either `request.enpoint`
+   (generally it will be string in format `BLUEPRINT_NAME.VIEW_NAME`, where blueprint is an
+   alternative of controller and view is an alternative of action) or `request.url_rule`(which is
+   object with additional information about current rule)
+
 v.2.8.0 2018-05-09
 ==================
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ v.2.9.0 TBA
 
 Deprecations:
  * ``c.action`` and ``c.controller`` variables should be avoided.
-   Refer to either ``request.enpoint`` (generally it will be string in format
+   Refer to either ``request.endpoint`` (generally it will be string in format
    ``BLUEPRINT_NAME.VIEW_NAME``, where blueprint is an alternative of
    controller and view is an alternative of action) or ``request.url_rule``
    (which is object with additional information about current rule).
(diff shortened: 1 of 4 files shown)


torfsen:
Should we also deprecate these attributes, for example by adding a note in the change log?

amercader:
Looks good. This is not going to cover all cases though. eg any dataset endpoint will set `c.controller == dataset` instead of `c.controller == package`, and the resource related endpoints will have `c.controller == resource` (eg this won't help with https://github.com/stadt-karlsruhe/ckanext-discovery/issues/1).

@torfsen I'm keen on your opinion as a extension maintainer here. Is it fair to document this in the changelog and ask extensions to do a bit of handling like

```python
try:
    is_dataset_request = toolkit.c.controller == 'package'    # CKAN < 2.8
except AttributeError:
    is_dataset_request = toolkit.request.endpoint.startswith('dataset.')   # CKAN >= 2.8
```

I guess that for the `package` case it would be all right to add an exception, but personally I'd like to avoid it. It's hard to keep the balance between moving forward and support all cases for extensions.

torfsen:
Personally, I'm definitely +1 for moving towards a consistent behavior without hard-coded exceptions, even if that requires modifications for our extensions. As long as the changes are well-documented, this should be no problem.

You could make things easier for extension maintainers by providing something like the following as a toolkit function (untested code):

```python
def get_endpoint():
    try:
        # CKAN >= 2.8
        return tuple(toolkit.request.endpoint.split('.'))
    except AttributeError:
        try:
            # CKAN < 2.8
            return toolkit.c.controller, toolkit.c.action
        except AttributeError:
            return None
````

That way one also has to update the extensions, but

1. the version-dependend code doesn't need to be rewritten every time
2. usage of the new `x.y` format is encouraged

One might still need to perform multiple checks due to difference in the versions (e.g. `dataset` vs. `package`), but one would need to do that anyway. And IMO it's better to do that only when necessary instead of manually adding selected exceptions to CKAN core:

```
if get_endpoint()[0] in ['dataset', 'package']:
    do_something()
````

amercader:
@smotornyuk will add the toolkit function suggested by @torfsen

PR #4348, opened 2018-07-18
[#4346] Update missing legacy routes

### LINKED ISSUE #4346: `search` legacy route missing
2.8+ (master)

There is a shortcut for the `search` legacy route in CKAN 2.8+, I guess it's missing in [LEGACY_ROUTE_NAMES](https://github.com/ckan/ckan/blob/df35eec2f5229bc60176cf96dc83ca1fcc259d1e/ckan/lib/helpers.py#L68-L71).

On master `search` is not available as route, basically the same as described in #4066

```python
h.build_nav_main(
          ('home', _('Homepage')),
          ('search', _('Datasets')),
          ('group_index', _('Groups'))
        )
```

tino097:
Fixes #4346

### Proposed fixes:
Update `LEGACY_ROUTE_NAMES` in the `helpers.py` with all default routes defined with navigation menu

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -68,6 +68,9 @@ MARKDOWN_ATTRIBUTES.setdefault('img', []).extend(['src', 'alt', 'title'])
 LEGACY_ROUTE_NAMES = {
     'home': 'home.index',
     'about': 'home.about',
+    'search': 'dataset.search',
+    'organizations_index': 'organization.index',
+    'group_index': 'group.index'
 }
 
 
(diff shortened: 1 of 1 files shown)


ckan/lib/helpers.py
```diff
@@ -68,6 +68,9 @@
 LEGACY_ROUTE_NAMES = {
     'home': 'home.index',
     'about': 'home.about',
+    'search': 'dataset.search',
+    'organizations_index': 'organization.index',
+    'group_index': 'group.index'
```
amercader: These two should be supported out of the box (https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L354). It's best to keep this list with only the really necessary legacy routes.

tino097: @amercader i guess `organizations` will be problem as the blueprint name is `organization`

amercader: You are absolutely right. This one should go in, I read too fast.

PR #4356, opened 2018-07-20
Update nodejs/less versions for frontend. 

davidread:
### Proposed fixes:

* Specify newer NodeJS (v4 -> v10. v4 is not being maintained)
* Specify newest `less` version
    * shorthand colors remain shorthand, e.g. `#fff`
* Rebuilt the frontend (i.e. ran 'less')

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/css/main.css
+++ b/ckan/public/base/css/main.css
@@ -1075,7 +1075,7 @@ body {
   font-size: 14px;
   line-height: 1.42857143;
   color: #333333;
-  background-color: #ffffff;
+  background-color: #fff;
 }
 input,
 button,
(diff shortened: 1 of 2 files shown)


davidread:
@gocemitevski you were last in this area - are you happy with this change?

gocemitevski:
@davidread yes, even though I would probably pick the LTS Node.js release - v. 8.x, as it is more stable. It is supported by April 2019. https://github.com/nodejs/Release#release-schedule

Nothing to add about the LESS version. Looks good to me.

davidread:
@johnglover thanks for this. Yeah, both NodeJS 8 and 10 work fine for this, but tbh we don't update this version very often, so next year will come around rather soon. 10.x is 'current' and is good til 2021 which I think is a better bet.

johnglover:
@davidread: I _could_ take credit, but it seems unfair :)

smotornyuk:
Yes, I agree with David - every even node version is LTS and as far as I know, by the end of the year v10 will replace v8 on current active LTS position. So it quite sensible to switch directly to  v10.

davidread:
Sorry @johnglover, but hello again! :)

johnglover:
👋

PR #4359, opened 2018-07-22
Fix usage of undefined variables

smotornyuk:
Fixes two problems, mentioned #4062. @tino097 , thank you for reporting this.


tino097:
@smotornyuk please check this 2 [lines](https://github.com/ckan/ckan/blob/master/ckan/templates-bs2/package/edit_base.html#L3-L4 ) too

smotornyuk:
Updated(and found few additional cases)

Issue #4338, opened 2018-07-16
[multilingual] AttributeError: '_Globals' object has no attribute 'fields'


saper:
### CKAN Version if known (or site URL)

2.8.0

### Please describe the expected behaviour

The site should load.

### Please describe the actual behaviour

After uploading a first data set, the site became unavailable.

During restart, the following exception is generated:

````
2018-07-16 10:14:55,940 DEBUG [pysolr] Found '1' search results.
[2018-07-16 10:14:55,944] ERROR in app: Exception on / [GET]
Traceback (most recent call last):
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/home/ckan/ckan/ckan/views/home.py", line 64, in index
    return base.render(u'home/index.html', extra_vars={})
  File "/usr/home/ckan/ckan/ckan/lib/base.py", line 127, in render
    return flask_render_template(template_name, **extra_vars)
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/templating.py", line 134, in render_template
    context, ctx.app)
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/flask/templating.py", line 116, in _render
    rv = template.render(context)
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/home/ckan/ukraine/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/home/ckan/ckan/ckan/templates/home/index.html", line 2, in top-level template code
[...]

PR #4360, opened 2018-07-22
Translate search facets only if they are present

smotornyuk:
Previously, attempting to get an unexisting key from `c` inside Pylons resulted in an empty string. As for Flask `g`, we are  raising AttributeError now. This PR updates multilingual extension with regard to the previous statement
Came from #4338 .


tino097:
@smotornyuk this looks ok, i think we should merge it and make it part of the 2.8.1 release 

@wardi whats your opinion regarding this ?

PR #4375, opened 2018-07-31
[#2713] Align IGroupForm customized facets with IDatasetForm

### LINKED ISSUE #2713: Group facets cannot be customized in the same way as dataset facets
When implementing IFacets interface, we want to override the default facet titles. For example, we may only want a part of the default facets for our dataset/group/organization page.
There are two ways to achieve this. Take `dataset_facets(self, facets_dict, package_type)` for example.

Method 1 is to change the facets_dict variable passed into the function, i.e

```
def dataset_facets(self, facets_dict, package_type):
        facets_dict.pop('organization')
        facets_dict.pop('license_id')
        return facets_dict
```

Method 2 is to return a new facet dictionary that contains all the facets titles that I want:

```
def dataset_facets(self, facets_dict, package_type):
        my_facet_titles = {
                    'groups': tk._('Categories'),
                    'tags': tk._('Tags'),
                    'res_format': tk._('Formats'),
                    }
        return my_facet_titles
```

Both methods work for dataset_facets, but the 2nd method will fail on group_facets. 
The issue is [here](https://github.com/ckan/ckan/blob/91e41b2e68faa3df5296a632f4862f5a55e69e62/ckan/controllers/group.py#L381). You can only make changes directly on the variable facets. Creating and returning any new dictionary in the `group_facets` method will not work.

Why is it necessary to define this [function](https://github.com/ckan/ckan/blob/91e41b2e68faa3df5296a632f4862f5a55e69e62/ckan/controllers/group.py#L379)? Why not make it consistent with the dataset counterpart?
Well, my current quick fix for my project is to always use method 1, or move the code outside `_update_facet_titles()`

Zharktas:
Fixes #2713

### Proposed fixes:
Aligns group_facets function calls with IDatasetForm.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/group.py
+++ b/ckan/controllers/group.py
@@ -370,7 +370,6 @@ class GroupController(base.BaseController):
         self._setup_template_variables(context, {'id': id},
                                        group_type=group_type)
 
-
     def bulk_process(self, id):
         ''' Allow bulk processing of datasets for an organization.  Make
         private/public or delete. For organization admins.'''
--- a/ckan/controllers/group.py
+++ b/ckan/controllers/group.py
@@ -322,7 +322,8 @@ class GroupController(base.BaseController):
                     facets[facet] = facet
 
             # Facet titles
-            self._update_facet_titles(facets, group_type)
+            for plugin in plugins.PluginImplementations(plugins.IFacets):
+                facets = plugin.group_facets(facets, group_type, None)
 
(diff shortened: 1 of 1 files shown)


smotornyuk:
[Fixed by some other PR](https://github.com/ckan/ckan/blob/master/ckan/views/group.py#L320)

Issue #4235, opened 2018-05-04
default SOLR query not searching text


jeff-at-h3:
### CKAN Version if known (or site URL)
2.7.2
Solr 7

### Please describe the expected behaviour
Using the search feature for datasets, providing text, eg. parks, returns datasets the have the work park (plus plurals) in the results.

### Please describe the actual behaviour
Our test env is using Solr 7 and our production env is using Solr 4.10, querying with 'park', there is no results with the search. 

### What steps can be taken to reproduce the issue? 
For the BCgov catalog, we haven't changed anything relating to the core serach functionality.

Querying the catalog with something like `/dataset?q=park` in our Production env, this works as expected, but in our upgraded envs, the same query returns no results.

However, specifying a field, eg. `text:park`, returns results as we expected.

In issue #3510, it seems like this is a similar problem?

### Questions
1. Has the default behaviour for `q` changed between Solr versions? @

dkelsey:
We have an extension that was injecting ```{!lucene q.op=AND}``` into the query.
The search query that resulted from a query for `park` was ```{!lucene q.op=AND}park```.
This was failing to return results. 
I removed the _injection_ of ```{!lucene q.op=AND}``` and the issue appears to be resolved.
We'll be testing to assert that this **does** resolve the issue.

smotornyuk:
Yes, i believe, that starting from v6.6 you need explicit definition of default query field. I'm using ```      <str name="df">text</str>
      <str name="q.op">AND</str>``` under `//requestHandler[@name=/select]/lst[@name="defaults]`(and some other changes) inside solrconfig.xml.

PR #4376, opened 2018-07-31
Explicitly set default search field in search query

### LINKED ISSUE #3510: Quirks with Solr6
Some quirks with CKAN and up-to-date versions of Solr:

- default field in the schema is deprecated: `[ckan] default search field in schema is text. WARNING: Deprecated,&#8203; please use 'df' on request instead.`
- When multiple terms are provided (like an org filter) the field name to search must be explicit. This is currently breaking searching for terms within an organization (`q=test +owner_org:...` fails, `q=text:test +owner_org:...` works, `q=test` works).
- While we're here, owner org should be a filter not part of the query.

Zharktas:
Fixes #3510, #4235 

### Proposed fixes:
Sets default search field in queries as the schema based default search field is deprecated in solr4 and non functioning in solr 6.6.

This should not affect instances running solr3.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -20,7 +20,7 @@ _open_licenses = None
 VALID_SOLR_PARAMETERS = set([
     'q', 'fl', 'fq', 'rows', 'sort', 'start', 'wt', 'qf', 'bf', 'boost',
     'facet', 'facet.mincount', 'facet.limit', 'facet.field',
-    'extras', 'fq_list', 'tie', 'defType', 'mm'
+    'extras', 'fq_list', 'tie', 'defType', 'mm', 'df'
 ])
 
 # for (solr) package searches, this specifies the fields that are searched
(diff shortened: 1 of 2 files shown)


TkTech:
Needs to be a setting, many sites use custom schemas. For example forcing the df to text in the query takes priority over other settings and would break our multilingual support. That said `text` is the proper default.

Zharktas:
I've moved it now to package_search, it can be overridden in before_search methods and be given as parameter to package_search itself.

wardi:
I don't know enough about solr to be comfortable will allowing this parameter to be user-defined.

Zharktas:
It needs be allowed in query.py, otherwise before_search can't manipulate it. But I'm fine with not exposing it in package_search by setting in it explicitly instead of using setdefault. I'll do a commit for it tomorrow once I'm back at the office.

PR #4384, opened 2018-08-03
[#4382] strip full URL before saving to DB

### LINKED ISSUE #4382: Full URL of the resource stored in the DB
2.6+
Reported in #4070 

When editing resource metadata, we should store the filename in the database, at `resource` table `URL` column. 

When editing resource metadata, full URL is saved in the database

1.Go to edit resource page
2.Save the changes

I've tested of this occurs on edit dataset and the issue is not reproducible.It occurs only on edit resource.

This occurs because we are generating the full url on [reading](https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L110) the resource and it is stored as it in the DB on save

tino097:
Fixes #4382 

### Proposed fixes:
Strip full url before saving 

### Features:

- [ ] includes API changes


ckan/lib/dictization/model_save.py
```diff
@@ -30,6 +32,11 @@ def resource_dict_save(res_dict, context):
     table = class_mapper(model.Resource).mapped_table
     fields = [field.name for field in table.c]
 
+    # Strip the full url for resources of type 'upload'
+    if res_dict.get('url_type') == u'upload':
+        url = res_dict.get('url')
+        res_dict[u'url'] = url[url.rfind(u"/")+1:]
```
wardi: This will fail if res_dict has no url or if it's set to `None`, also `rfind` is a pretty rarely used string function. How about:
```python
if res_dict.get('url') and res_dict.get('url_type') == u'upload':
    res_dict['url'] = res_dict['url'].rsplit('/')[-1]
```

wardi:
A quick test would be great too. How are you thinking of handling existing uploaded urls already in the db?

amercader:
Regarding on whether we need a migration for this, I think this change will correct resources if they are updated so it will be an improvement from the current situation and perhaps it's not worth the risk of having to update lots of fields in big instances.

ckan/lib/dictization/model_save.py
```diff
@@ -30,6 +32,10 @@ def resource_dict_save(res_dict, context):
     table = class_mapper(model.Resource).mapped_table
     fields = [field.name for field in table.c]
 
+    # Strip the full url for resources of type 'upload'
+    if res_dict.get('url') and res_dict.get('url_type') == u'upload':
+        res_dict['url'] = res_dict['url'].rsplit('/')[-1]
+
```
wardi: can we do the same when reading resources? That way sites changing their site_url will work even if they uploaded files before the fix.

wardi: Doesn't look like it. I mean actually `res_dict['url'] = res_dict['url'].rsplit('/')[-1]` like above or maybe just removing the `for_edit` check in https://github.com/ckan/ckan/blob/469d068a01a33960f263142ea671aefed031547b/ckan/lib/dictization/model_dictize.py#L110

A small change like this should mean existing sites can change their SITE_URL, then run a solr reindex to update all their download urls without having to touch the db.

ckan/lib/dictization/model_dictize.py
```diff
@@ -114,8 +114,8 @@ def resource_dictize(res, context):
                                     resource_id=res.id,
                                     filename=cleaned_name,
                                     qualified=True)
-    elif resource['url'] and not urlparse.urlsplit(url).scheme and not context.get('for_edit'):
-        resource['url'] = u'http://' + url.lstrip('/')
+    elif resource['url'] and not urlparse.urlsplit(url).scheme:
+        res_dict['url'] = res_dict['url'].rsplit('/')[-1]
```
wardi: this returns just the file name part of any resource URL that is missing an scheme when either `for_edit` is true or the `url_type` is *not* upload (see if statement above). That doesn't make sense to me.

tino097: now ive got lost a bit, the change that was requested a while ago, isnt been targeting this condition ?!

wardi:
This is the situation and the change I think should be part of this PR:

1. existing ckan code base has resources saved with `resource.url_type = 'upload'` and `resource.url = 'http://existingsitename/.../filename.csv'`

2. site upgrades to version with the fix in this PR and changes their site name to `newsitename`

3. package_dictize is used to return the resource urls from the db, i.e. `'http://existingsitename/.../filename.csv' which is now incorrect because it still has the old host name.

We should update package_dictize so that it ignores everything up to the `'/filename.csv' in url fields in the database when `resource.url_type = 'upload'` and instead generates the url based on the current sitename. This way users don't need to update all their datasets when their site name changes, they just need to rebuild the search index.

PR #4385, opened 2018-08-06
Separate IDatasetForm/IGroupForm blueprints and plugins registration

amercader:
@smotornyuk @tino097 Some extra tweaks on the registration of dataset/group plugins after the migration to Flask.

Consider the following test:

*test.ini*
```
[...]
ckan.plugins = my_idatasetform_plugin ...

```

*test_my_dataset.py*
```

class MyDatasetType(FunctionalTestBase):
    @classmethod
    def setup_class(cls):
        super(MyDatasetType, cls).setup_class()
        p.load('image_view')   # Completely unrelated plugin

    def test_my_dataset_type(self):
          dataset = call_action('package_create', {}, type='my_custom_type', 'custom_field'=123)

          eq(dataset['custom_field'], 123)  # Raises KeyError
```

The test above raises a KeyError because the dataset created was a standard default one. This was first noticed in [ckanext-dcat](https://travis-ci.org/ckan/ckanext-dcat/jobs/411278807#L1400)

When loading a new plugin we are resetting the environment by calling `ckan.config.environment.update_config()`. This will clear all dataset / group plugins previously loaded (calling the `lib.plugins.reset_package_plugins()`), but we are not re-registering the previously loaded.

If you recall from discussion in #4062, we moved the registration of the plugins to the `flask_app.,py` module because we required a valid application object to register the blueprints. This is still true but we also need to register the actual IDatasetForm plugins every time the environment is reloaded so the only sane way I can think of is to split the registration of both. For each `IDatasetForm` plugin:

* The blueprints that set the custom routes (if needed) are set in `make_flask_app()`
* The custom plugins are added to the internal `lib.plugins` plugin registry every time `environment.update_config()` is called.

Does this make sense?

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -220,7 +220,8 @@ def register_group_plugins():
             _group_plugins[group_type] = plugin
             _group_controllers[group_type] = group_controller
 
-        set_default_group_plugin()
+    # Setup the fallback behaviour if one hasn't been defined.
+    set_default_group_plugin()
 
 
 def register_group_blueprints(app):


smotornyuk:
Yes, it souds reasonable for me

tino097:
+1

amercader:
@smotornyuk can you merge if happy?

PR #4387, opened 2018-08-06
add SOLRv7 schema.xml

smotornyuk:
I added setup notes into `solr.rst` - everyone who want to check it may follow them in order to configure solr. 

PS i tested it on 7.3 and 7.4 and i will appreciate if anyone will try on older version.
PPS I'm using archLinux so i would also appreciate if somebody has a chance to test it on ubuntu :v:

### DIFF
--- a/ckan/config/middleware/pylons_app.py
+++ b/ckan/config/middleware/pylons_app.py
@@ -263,7 +263,7 @@ def execute_on_completion(application, config, callback):
     def inner(environ, start_response):
         try:
             result = application(environ, start_response)
-        except:
+        except Exception:
             callback(environ)
             raise
         # paste.fileapp converts non-file responses into list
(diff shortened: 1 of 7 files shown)


amercader:
From our discussion, if I understood correctly, a new schema file is needed because the classes used in some of the field definitions are not compatible with Solr 4

@wardi, @TkTech I can see how having two separate schema files, one of which is not generally used in tests, etc could lead to problems. Can you think of an alternative approach? The obvious seems to drop support of Solr<=4 but I'm pretty sure that's still used in lots of places (eg Ubuntu 16.04 LTS installs Solr 3 via packages!)

TkTech:
There are a number of breaking changes, not just the new "default" types. The one I imagine people keep running into is the removal of default field and default operator, since that also affects 6.

I believe there was a discussion on this in one of the other issues, but Solr pushes releases with breaking changes far faster then package maintainers update them. Solr is just a jar and very easy to install, upgrade and run, so there's no reason to limit ourselves to outdated packages. My suggestion would be to support the current solr release and one version prior **at the time** we do a release.

If we only support the current version and the prior version, I don't see a problem having the ci run the tests against both versions.

PR #4389, opened 2018-08-08
Wrong parameters in pg_dump

aborruso:
I have seen documentation https://www.postgresql.org/docs/current/static/app-pgdump.html and "-u" and "-d" do not exist

Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/doc/maintaining/database-management.rst
+++ b/doc/maintaining/database-management.rst
@@ -64,7 +64,7 @@ and restoring a database and its content to/from a file.
 
 For example, first dump your CKAN database::
 
-    sudo -u postgres pg_dump --format=custom -d ckan_default > ckan.dump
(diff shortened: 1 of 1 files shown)


TkTech:
These are not equivalent. `sudo -u postgres` is changing the UID to postgres, not passing a `-u` flag to pg_dump.

smotornyuk:
(wisper) And you'll probably get error due to peer authentification failure in case of `pg_dump -U`. Unless you have chnaged configuration

wardi:
Everyone's environment is going to be a little bit different (that's one of the reasons we removed the pg_dump wrapper that was in the cli) we could add something like "exact commands to back up and restore may vary from system to system, consult your PostgreSQL documentation"

PR #4394, opened 2018-08-13
[WIP] - CIRCLE CI V2

deinok:
FAILED (SKIP=3, errors=1, failures=3)

Any ideas on why this is happening?


tino097:
Not sure, something in ckanext tests could be still an issue

@amercader any thoughts ?

test-core-circle-ci.ini
```diff
@@ -0,0 +1,140 @@
+#
```
amercader: Rather than duplicating the whole file, if for some reason you need to override a config setting from `test-core.ini` you can just create this new file:

```
[server:main]
use = egg:Paste#http
host = 0.0.0.0
port = 5000

[app:main]
use = config:test-core.ini

sqlalchemy.url = postgresql://ckan_default:pass@ckan-postgres/ckan_test

ckan.datastore.write_url = postgresql://datastore_write:pass@ckan-postgres/datastore_test
ckan.datastore.read_url = postgresql://datastore_read:pass@ckan-postgres/datastore_test
```

amercader:
@deinok thanks a lot for your work on this!

I don't know what could be causing this failure. The test is one that @smotornyuk wrote for the new streaming features in Flask but looking at the stacktrace it looks like the Pylons stack is trying to serve it. This might suggest that the plugin is not loaded but then the other tests on the same file work so that's really confusing.

@deinok can you add the -v flag to the tests command so at least we see the order in which they are run?

I know we have discussed this in the past but I think it would be great if we parked these failures for now and focused on getting the split tests (circleci-matrix) working on Circle v2.0. Looking at the work you have done it seems that this should not be difficult.

In parallel we'll keep working on fixing these tests (@tino097 has done a lot of work on that already). Just to be clear, these failures also seem to happen when running all tests locally, so they are not related to Circle CI.

deinok:
@amercader The latest build -> https://circleci.com/gh/deinok/ckan/446

deinok:
Also, notice that the CI is using the our SOLR image. At some point we should move to schemaless SOLR and change that to use the official SOLR image

amercader:
@deinok, can you hook the circleci-matrix to run the tests across the 4 containers as part of this PR? We'll sort out the failures later.

deinok:
@amercader I'm sorry, but I can't see the point of spliting the tests. I doesn't give any benefit, just add more complexity. If testing speed is a requirement, there are better solutions.

If you want the matrix like style, you can merge this an then setup the matrix, or add commits to this PR.

PR #4399, opened 2018-08-14
[4398] Trash batch-purging

### LINKED ISSUE #4398: Trash purge timeouts

2.7

The sysadmin can purge datasets from trash. 

For sites with a large number of datasets in the trash, purge times out as it is an all or nothing operation.

Delete a large number of datasets in CKAN (e.g. > 100) and then try to purge them from Trash as sysadmin.

jqnatividad:
Related to #4398 

### Proposed fixes:
Purge datasets 10 at a time to avoid timeouts when there are a lot of trash items.

Make the trash list a numbered list. Since the list does not do pagination, and the purge button is all the way in the bottom, the sysadmin needs to go to the bottom of the list to press "Purge."  Making it a numbered list, gives the sysadmin useful info when there a lot of datasets available for purging.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/admin.py
+++ b/ckan/controllers/admin.py
@@ -183,7 +183,7 @@ class AdminController(base.BaseController):
                     except Exception as inst:
                         msg = _('Problem purging revision %s: %s') % (id, inst)
                         msgs.append(msg)
-                h.flash_success(_('Purge %s complete') % (limit))
+                h.flash_success(_('Purge %s complete') % (index))
             else:
                 msgs.append(_('Action not implemented.'))
 
--- a/ckan/controllers/admin.py
+++ b/ckan/controllers/admin.py
@@ -144,7 +144,8 @@ class AdminController(base.BaseController):
                                                         request.params):
                 if 'purge-packages' in request.params:
                     revs_to_purge = []
-                    for pkg in c.deleted_packages:
+                    limit = 10
+                    for index, pkg in zip(range(limit), c.deleted_packages):
                         revisions = [x[0] for x in pkg.all_related_revisions]
                         # ensure no accidental purging of other(non-deleted)
                         # packages initially just avoided purging revisions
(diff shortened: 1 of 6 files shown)


TkTech:
I'd suggest doing this a bit differently, especially because we allow hooking package delete/purge which can cause some pretty lengthy tasks to run (such as unmarking it for publication on the source site).

1. When datasets are selected for purge, mark their status as something new (say `deleted_pending_purge`). If we want to be fully backwards compatible we'd have to add a new flag to the model instead (ex: `purge=True`)
2. Put the actual purge task into a scheduled background task (or trigger it immediately when purge is pressed)

wardi:
I'm fine with this quick hack, but @TkTech 's approach is the best long term fix if someone wants to work on it.

jqnatividad:
We had a large 2.5.x installation with hundreds of datasets in the Trash, and we were getting web server timeouts when their sysadmin was trying to purge the trash.

Had to pull this together as they were running out of DB space as the datastore was filling it up.  Even had to use @metaodi's "quick-and-dirty" paster command to delete orphaned tables (#3422 ).

They're now on 2.7.3, and find we still need this because of their heavy usage of the site results in trash filling up.

Perhaps, until someone works on the long-term fix?

TkTech:
Sorry, my comment really should have been on the issue rather than the PR. I'm 100% 👍 on this change going in, but the issue is still an issue and should remain open until we can move it to the background workers.

wardi:
This is handled by the flask views now in master. Can you make the same change to https://github.com/ckan/ckan/blob/master/ckan/views/admin.py ?

kmbn:
The Travis build shows up as pending here, but if you [click through](https://travis-ci.org/ckan/ckan/builds/441697386?utm_source=github_status&utm_medium=notification), all the tests pass. @jqnatividad made the requested changes for Flask. Is there anything else that needs to be done before merging the PR, @wardi?

jqnatividad:
Closing this stale PR as this has been "partly" fixed with the removal of revisions.

"Partly" because the datasets are only marked as `deleted` and not actually purged from the datastore, resulting in orphaned tables. 

https://github.com/ckan/ckan/issues/3422#issuecomment-549511460

PR #4408, opened 2018-08-17
Switch config file reference to production

LukeTully:
Updated docs for consistent configuration file references

Fixes #
Switched the reference from development.ini to production.ini

### Proposed fixes:
Fixing there wherever else this might occur.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -885,23 +885,9 @@ class ApiController(base.BaseController):
             return request.GET.mixed()
 
         else:
-            try:
-                if request.method in ['POST', 'PUT']:
-                    request_data = request.body
-                else:
-                    request_data = None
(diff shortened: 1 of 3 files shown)


kmbn:
@amercader Should we go through _all_ the docs and settle on either `development.ini` or `production.ini` for all of them? PR #4411 attempts to address the same issue in a different doc, but in that case it's swapping production for development...

amercader:
We should just rename the file to `ckan.ini` and stop making the distinction. People will know if they are deploying to dev or prod :)

As for this particular one or #4411 I don't mind but obviously we should be consistent. Perhaps development.ini makes more sense for now but I don't know how often one or the other are used.

amercader:
Ini file is now called `ckan.ini`

PR #4410, opened 2018-08-17
Remove pyfakefs as a dependency

### LINKED ISSUE #4392: pbr version conflicts with dev-requirements.txt
```
ckan's requirements.txt:
    pbr==1.10.0               # via sqlalchemy-migrate

ckan's dev-requirements.txt:
    pyfakefs==2.9

pyfakefs's [setup.py](https://github.com/jmcgeheeiv/pyfakefs/blob/version-2.9/setup.py#L26)
    setup_requires=mox3

mox's setup.py:
    setup_requires=['pbr>=2.0.0']
```

master

* `pip install requirements.txt`
* `pip install dev-requirements.txt` gives error:
```
mox3 0.26.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
```

TkTech:
Fixes #4392

### Proposed fixes:

Removes pyfakefs as a dependency and from the tests that used it. The caveat is that these tests now require the storage path setting in the INI, which really should be done anyways. Previously it would try the real filesystem then pyfakefs if the real fs calls failed, but this was just potentially hiding other issues.


TkTech:
The circleci failures seem to be from the storage path not being configured, but we should probably give a clear, early error message instead of random failures.

TkTech:
Anyone willing to finish this? Don't have time to work on CKAN at the moment. This dependency causes problems and the tests are hiding potential issues with real file uploads. Good first issue.

PR #4412, opened 2018-08-20
 [#4291] werkzeug dev server

### LINKED ISSUE #4291: Run the development server using Werkzeug
As we transition to Flask, besides migrating the controllers we need to start picking up other tools from its ecosystem. As it surfaced in #4257 in order to get the debugger pages working on requests served by Flask we will need to switch from the Paster development server (`paster serve`) to a Werkzeug-based one. This will enable the debugger on Flask requests (and is probably more performant that the Paster server).

To keep things simple and maintain a syntax close to the current one I propose adding a new paster command. At its simplest form (run from the `ckan` folder) it should look like:

```
paster devserver
```
Additional options could mimic as close as possible the current `paster serve` ones. The Paster `serve` command requires the path to config file as a positional parameter, we could probably implement that as well on the new command, but my preference is to align it to use the `-c` option that all other paster commands use.

Just to be clear the `paster serve` command is not going anywhere so any existing scripts will work, but it should be deprecated and documentation, CI, etc updated to reflect the new command.

The following was enough to get it working, enabling the debugger on Flask requests (the Pylons debugger still works when running under Werkzeug):

```diff
diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
index 1f9fa22..b6631cd 100644
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -103,6 +103,10 @@ def make_flask_stack(conf, **app_conf):
         app.config['DEBUG_TB_INTERCEPT_REDIRECTS'] = False
         DebugToolbarExtension(app)
 
+        from werkzeug.debug import DebuggedApplication
+        app = DebuggedApplication(app, True)
+        app = app.app
+
     # Use Beaker as the Flask session interface
     class BeakerSessionInterface(SessionInterface):
         def open_session(self, app, request):
diff --git a/ckan/lib/cli.py b/ckan/lib/cli.py
index c487b49..84155f1 100644
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -330,6 +330,36 @@ class CkanCommand(paste.script.command.Command):
         self.site_user = load_config(self.options.config, load_site_user)
 
 
+class DevServerCommand(CkanCommand):
+    '''
+    Starts the development server
+
+    TODO: xx
+    '''
+
+    summary = __doc__.split('\n')[0]
+    usage = __doc__
+
+    # TODO: Options for debug , reload, port et al
+
+    def command(self):
+        self.serve()
+
+    def serve(self):
+        from werkzeug.serving import run_simple
+
+        conf = _get_config()
+
+        from ckan.config.environment import load_environment
+        load_environment(conf.global_conf, conf.local_conf)
+        app = make_app(conf.global_conf, **conf.local_conf)
+
+        # TODO: options
+        run_simple(
+            'localhost', 5000, app,
+            use_reloader=True, use_debugger=True, use_evalex=True)
+
+
 class ManageDb(CkanCommand):
     '''Perform various tasks on the database.
 
diff --git a/setup.py b/setup.py
index 4cc330e..2c54e5f 100644
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ entry_points = {
         'views = ckan.lib.cli:ViewsCommand',
         'config-tool = ckan.lib.cli:ConfigToolCommand',
         'jobs = ckan.lib.cli:JobsCommand',
+        'devserver = ckan.lib.cli:DevServerCommand',
     ],
     'console_scripts': [
         'ckan-admin = bin.ckan_admin:Command',

```

tino097:
Fixes #4291 

As it is discussed in this issue, we should start migrating to more Flask oriented development environment and using its libraries.  Within this PR  we will start using `werkzeug.serving.run_simple` for starting the CKAN instance. 

I believe that we should start cleaning up the middelware of the Pylons stuff and using the alternatives provided by Flask. For some time we could use both `paste serve` and `ckan run` commands but lets focus on the later.

Please add your suggestions in the comments

Current status is that we have `ckan run` command, but it needs to pass the configuration file to set the app variables


ckan/lib/flask_cli.py
```diff
@@ -0,0 +1,35 @@
...
+def _helper():
+    print('nothing')
+
+
+@click.help_option(u'-h', u'--help')
+@click_config_option
+def run():
+    # app = _load_config(config)
+    click.echo('Starting CKAN')
+    run_simple('localhost', 5000, make_app, use_reloader=True, use_evalex=True)
```
wardi: the host name, port and use_reloader should be added as click options

tino097: idea is to get them from config file or maybe as arguments while running the command. i've created the PR for the feedback

ckan/cli/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+click_config_option = click.option(
+    u'-c',
+    u'--config',
+    default=None,
+    metavar=u'CONFIG',
+    help=u'Config file to use (default: development.ini)')
+
+
+def _load_config(config=None):
```
wardi: This looks reusable, how about dropping the leading `_`?

There are parts of this function shared with the `_get_config` code in lib/cli.py, can they be factored out instead of repeated?

tino097: im keeping the same code as it is in `_get_config` in `ckan/lib/cli.py`, but ive renamed it to `load_config`,  because we are still using pylons and we should call `make_app` to create both stacks.

ckan/lib/flask_cli.py
```diff
@@ -0,0 +1,84 @@
+# encoding: utf-8
```
amercader: IIUC this file is no longer needed right?

tino097: yes, it can be removed

ckan/cli/db.py
```diff
@@ -0,0 +1,68 @@
...
+from flask import Flask, current_app
+from werkzeug.serving import run_simple
+
+from ckan.cli import click_config_option
+
+
+@click.group(name=u'db', short_help=u'Database commands')
+@click.help_option(u'-h', u'--help')
+def db():
+    pass
```
wardi: if you run db without a subcommand it should print the help, not do nothing right?

ckan/cli/server.py
```diff
@@ -0,0 +1,21 @@
...
+import os
+
+import click
+from flask import Flask, current_app
+from werkzeug.serving import run_simple
+
+from ckan.cli import load_config, click_config_option
+
+
+@click.command(u'run', short_help=u'Start development server')
```
wardi: is `run` a common way to start a flask server or did you just prefer `ckan run` over `ckan serve`?

ckan/cli/db.py
tino097: it lists Usage, Options and Commands sections without doing anything else

ckan/cli/server.py
tino097: Flask is using the command `run` for starting the app, i believe its a good practice to follow up the flask way thus i choose `run` over `serve`.

amercader:
TODO: logging

ckan/cli/db.py
```diff
@@ -0,0 +1,76 @@
...
+
+from ckan.cli import click_config_option
+import logging
+
+
+log = logging.getLogger(__name__)
+
+
+@click.group(name=u'db', short_help=u'Database commands')
+@click.help_option(u'-h', u'--help')
```
amercader: I assume this is to support `-h` in additon to `--help`?  It seems overkill to have to define this for every single command and subcommand. IIUC, [this](http://click.palletsprojects.com/en/6.x/documentation/#help-parameter-customization) should set it globally at the `ckan` command level, so we can remove all the individual decorators

ckan/cli/cli.py
```diff
@@ -0,0 +1,32 @@
+# encoding: utf-8
+
+import os
```
amercader: There's quite a lot of unused imports in the different files, can you remove them? Your IDE/editor should be able to help with that.

ckan/cli/search_index.py
```diff
@@ -0,0 +1,120 @@
...
+def clear(dataset_name):
+    from ckan.lib.search import clear, clear_all
+
+    if dataset_name:
+        clear(dataset_name)
+    else:
+        clear_all()
+
+
+@search_index.command(name=u'rebuild-fast',
```
amercader: The current name in `paster` is `rebuild_fast`. Perhaps is better to keep the same name for compatibility.

ckan/common.py
```diff
@@ -1,3 +1,4 @@
+
```
amercader: Remove

ckan/lib/cli.py
```diff
@@ -556,9 +556,6 @@ def rebuild_fast(self):
             package_ids.append(row[0])
 
         def start(ids):
-            ## load actual enviroment for each subprocess, so each have thier own
-            ## sa session
-            self._load_config()
```
amercader: Is this really needed? I'd expect no changes in the old CLI

ckan/cli/server.py
```diff
@@ -0,0 +1,24 @@
...
+@click.help_option(u'-h', u'--help')
+@click.option(u'-H', u'--host', default=u'localhost', help=u'Set host')
+@click.option(u'-p', u'--port', default=5000, help=u'Set port')
+@click.option(u'-r', u'--reloader', default=True, help=u'Use reloader')
+@click.pass_context
+def run(ctx, host, port, reloader):
+    u'''Runs development server'''
+    log.info(u"Running server {0} on port {1}".format(host, port))
```
amercader: The current paster server supports defining the port and the host in the ini file:

```
[server:main]
use = egg:Paste#http
host = 0.0.0.0
port = 5001
```

There's no need to add support for this now but it's worth keeping it in mind in case we want to implement it in the future.

setup.py
```diff
@@ -77,6 +77,7 @@ def parse_version(s):
         'jobs = ckan.lib.cli:JobsCommand',
     ],
     'console_scripts': [
+        'ckan = ckan.cli.cli:ckan',
         'ckan-admin = bin.ckan_admin:Command',
```
amercader: This line (`ckan-admin`) can be removed as it's no longer needed.

ckan/cli/cli.py
```diff
@@ -0,0 +1,32 @@
...
+
+@click.group()
+@click.help_option(u'-h', u'--help')
+@click_config_option
+@click.pass_context
+def ckan(ctx, config, *args, **kwargs):
+    ctx.obj = CkanCommand(config)
+
+
+ckan.add_command(server.run)
```
amercader: We probably want to automate this somehow and add support for extensions here, but we can do both these things at a later stage

amercader:
For reference, we are dropping the following subcommands in the new CLI:

```
    db create                      - alias of db upgrade
    db create-from-model           - create database from the model (indexes not made)
    db migrate-filestore           - migrate all uploaded data from the 2.1 filesore.
```

These are all fine by me but I want to double check with others @ckan/core

ckan/cli/search_index.py
wardi: I like the hyphen

ckan/cli/server.py
tino097: i will suggest to keep it for now

ckan/lib/cli.py
tino097: i guess i've removed this by mistake

amercader:
:tada: First step to get the CLI that CKAN deserves :) Thanks @tino097

PR #4422, opened 2018-08-24
support resource overrides in resources.config using fanstatic supersedes feature

OriHoch:
Fanstatic resources support a [supersedes](http://www.fanstatic.org/en/0.12/api.html#fanstatic.Resource) argument which allows to override resources.

This PR exposes this functionality via the `resource.config` file. It also has to be enabled via the Fanstatic app config - which is done by setting `fanstatic_enable_rollup = yes` in the ckan config.

### Manual Testing Procedure
* Under an existing plugin directory, create a directory to hold the resources: `override_select2`
* Under the `override_select2` directory create the following files:
  * `upgraded-select.js`
    * Copy the contents of this file from `public/base/vendor/select2/select2.js`
    * Add a `foo: "bar",` attribute to the `Select2` object towards the end of the file
  * `resources.config`
```
[supersedes]

upgraded-select.js = vendor/select2/select2.js
```
* Edit the plugin's `plugin.py` and add the following line to the `update_config` method:
  * `p.toolkit.add_resource('override_select2', 'override_select2')`
* Edit the ckan app config, add the following line:
  * `fanstatic_enable_rollup = yes`
* Restart the app
* Open the browser console, type: `Select2.foo`
* Expected: `"bar"`

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -206,6 +206,8 @@ def make_flask_stack(conf, **app_conf):
         app = plugin.make_middleware(app, config)
 
     # Fanstatic
+    fanstatic_enable_rollup = asbool(app_conf.get('fanstatic_enable_rollup',
+                                                  False))
     if debug:
         fanstatic_config = {
             'versioning': True,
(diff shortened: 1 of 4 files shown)


smotornyuk:
Fantastic, I really missed this feature and hadn't thought it already included in fanstatic library. Thank you, @OriHoch , i've checked it and now going to merge into upstream

PR #4430, opened 2018-08-26
Circle v2 w/parallel containers

wardi:
Continues work on #4394 by splitting tests across the 4 containers Circle CI provides for free to open source projects


tino097:
Looks like same tests are failing on both cases

.circleci/config.yml
```diff
@@ -0,0 +1,93 @@
...
+          NODE_TESTS_CONTAINER: 2
+      - image: postgres:10
+        environment:
+          POSTGRES_USER: ckan
+          POSTGRES_PASSWORD: ckan
+        name: ckan-postgres
+      - image: redis:3
+        name: ckan-redis
+      - image: ckan/solr:latest
+        name: ckan-solr
```
wardi: This doesn't use the solr configuration from the code being tested. We'll need to copy the schema over if we use to the docker image like this.

tino097:
I've debugged the issue with the test 
```
Iterator garbage collected without being closedException AssertionError: AssertionError('Iterator garbage collected without being closed',) in <bound method IteratorWrapper.__del__ of <webtest.lint.IteratorWrapper object at 0x7f714b484ed0>> ignored
Test extension sets up a unique route. ... Error - <type 'exceptions.RuntimeError'>: maximum recursion depth exceeded while calling a Python object
URL: http://localhost/stream/template
```
https://github.com/Pylons/webtest/issues/168

@wardi @smotornyuk @amercader

wardi:
Tests are green! Based on work from @deinok #4394 and @tino097 #4439

tino097:
Great work @wardi !!!

wardi:
@amercader @smotornyuk @davidread @TkTech If any of you have a minute to review this one, it would be great to have our circle CI tests working again.

smotornyuk:
From my point of view - it works and thus it should be merged as soon as possible :)

By the way, what's the problem with restarting jetty service? this line `service jetty9 restart || true  # erroring out but does seem to work` is pretty suspicious..

deinok:
From my point of view it should be merged.
But I dislike the idea of solr-jetty, I think we should move to shemaless config and move to newer versions of SOLR and use the oficial docker image.

wardi:
@smotornyuk just saw this error when ssh-ing into the CI container so I added that comment. When I run `service restart ...` it showed "failed" even though there were no errors in the log and the service was restarted successfully.

@deinok as a separate PR schemaless solr sounds great! We'd need updates to the docs and possibly new plugin endpoints too.

deinok:
@wardi Yes, I think that the benefits of having a schemaless is downgrade the complexity of a lot of use cases. Like simplificating the testing environment for CI.

PR #4438, opened 2018-08-31
right to left support

OriHoch:
Modify the site layout for right to left languages (Hebrew and Arabic) - to match the expected language directionality.

For right to left languages only - adds a [bootstrap 3 rtl css](https://github.com/parsmizban/Bootstrap-RTL) (MIT License) and a custom rtl css to fix ckan specific styles.

Added configuration options - 
* `ckan.i18n.rtl_languages` - list of languages rtl support will be activated for
* `ckan.i18n.rtl_css` - allows to override the rtl css

### DIFF
--- a/ckan/lib/app_globals.py
+++ b/ckan/lib/app_globals.py
@@ -56,6 +56,7 @@ app_globals_from_config_details = {
                       'name': 'facets'},
     'package_hide_extras': {'type': 'split'},
     'ckan.plugins': {'type': 'split'},
+    'ckan.i18n.rtl_languages': {'type': 'split'},
 
     # bool
     'debug': {'default': 'false', 'type' : 'bool'},
@@ -82,6 +83,12 @@ def set_main_css(css_file):
     # FIXME we should check the css file exists
     app_globals.main_css = str(new_css)
 
+def set_rtl_globals(rtl_css_file, rtl_languages):
+    assert rtl_css_file.endswith('.css')
+    new_css = rtl_css_file
+    # FIXME we should check the css file exists
+    app_globals.rtl_css = str(new_css)
+    app_globals.rtl_languages = rtl_languages
 
 def set_app_global(key, value):
     '''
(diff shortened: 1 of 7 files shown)


ckan/templates/base.html
```diff
@@ -67,6 +67,9 @@
     #}
     {%- block styles %}
       {% resource g.main_css[6:] %}
+      {% if h.lang() in g.rtl_languages %}
```
amercader: Can you use a `h.is_rtl_language()` helper function that checks the current lang and the config options needed? We are trying to limit the stuff that goes into the globals.

ckan/templates/base.html
```diff
@@ -67,6 +67,9 @@
...
+        {% resource g.rtl_css[6:] %}
```
amercader: Actually now I see that you'll need two helpers, as you'll also need to get the config value for this (`h.get_rtl_css()`)

OriHoch:
@amercader fixed

OriHoch:
Made some additional changes - 
* Added Persian to list of rtl languages
* Improved order of resource loading - replacing main.css with main-rtl.cc
* Additional CSS fixes

doc/maintaining/configuration.rst
```diff
@@ -1759,6 +1759,33 @@ If you have set an extra i18n directory using ``ckan.i18n.extra_directory``, you
...
+.. _ckan.i18n.rtl_languages:
+
+ckan.i18n.rtl_languages
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Example::
+
+  ckan.i18n.rtl_languages = he ar
+
+Default value: ``he ar``
```
amercader: IIUC we need Persian here as well right?

ckan/lib/app_globals.py
```diff
@@ -56,6 +56,7 @@
                       'name': 'facets'},
     'package_hide_extras': {'type': 'split'},
     'ckan.plugins': {'type': 'split'},
+    'ckan.i18n.rtl_languages': {'type': 'split'},
```
amercader: This can be removed now

OriHoch:
thanks @amercader fixed

klikstermkd:
@OriHoch @amercader Was the portal well tested before this PR was merged? I noticed that for example if `ar` locale is used the layout on the homepage is not properly shown.

PR #4439, opened 2018-08-31
Circle v2 fix

tino097:
This PR should fix #4394 tests execution.


tino097:
Locally all tests are passing, before this PR, only one test was falling because multiple execution of `load_environment` which was causing monkey patch of `find_controller` method to go into recursion with following stack trace
https://gist.github.com/tino097/ba8f678a8852da4fde35bc10dc484aa4

ckanext/imageview/tests/test_view.py
```diff
@@ -11,21 +11,20 @@
 class TestImageView(helpers.FunctionalTestBase):
 
     @classmethod
-    def setup_class(cls):
+    def _apply_config_changes(cls, cfg):
+        cfg['ckan.plugins'] = 'image_view'
```
wardi: better yet, use `_load_plugins = ['image_view']` at the class level and FunctionalTestBase will take care of loading and unloading the plugins

tino097: i've done this, because it was complaining with `template not found`  error

wardi:
I'm pretty sure the solr tests are failing because these are set up to use the ckan solr docker image without overriding the solr schema. I'm going to try removing that container because circle doesn't allow us to mount or copy the schema from the local system AFAICT

ckanext/imageview/tests/test_view.py
tino097: i see what have you meant, whole day debugging got me :)

PR #4441, opened 2018-09-03
Update tests so that one could understand, what feature is tested

smotornyuk:
@tino097 , i was unable to reproduce locally error with the missed application context, even when I executed all the tests. But, anyway, I slightly updated tests and now they are a little bit more transparent. Could you review them and, check, whether it somehow changes results of your local run? I changed check in *test_template_streaming*, which was causing problems and there is a chance, that this change may help you run tests(previously this test was attempting to get unexisting page).

### DIFF
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -1098,7 +1098,7 @@ def history(package_type, id):
             revision_date = h.date_str_to_datetime(revision_dict[u'timestamp'])
             try:
                 dayHorizon = int(request.args.get(u'days'))
-            except:
+            except Exception:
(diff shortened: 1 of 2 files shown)


tino097:
@smotornyuk its great that you've improved the docstrings. 

I've introduced a fix for the errors that i had on my local env regarding the `test_template_streaming` with https://github.com/ckan/ckan/pull/4439/commits/e1a3d0553781ec8e40884df42dc0b692f26646e8. 

Here is the error stack that ive been receiving upon execution of the tests https://gist.github.com/tino097/ba8f678a8852da4fde35bc10dc484aa4

PR #4447, opened 2018-09-07
`default` validator replaces all falsey values

wardi:
Change the current behaviour of the core `default` validator:

| default(42) | current return value | new return value |
| - | - | - |
| (missing) | 42 | 42 |
| None | 42 | 42 |
| 100 | 100 | 100 |
| 0 | 42 | **0** |
| False | 42 | **False** |
| [] | 42 | **[]** |
| '' | 42 | **''** |

### DIFF
--- a/ckan/lib/navl/validators.py
+++ b/ckan/lib/navl/validators.py
@@ -74,11 +74,12 @@ def ignore(key, data, errors, context):
     raise StopOnError
 
 def default(default_value):
-
+    '''
+    Return default_value if key is missing or None.
+    '''
     def callable(key, data, errors, context):
-
         value = data.get(key)
-        if not value or value is missing:
+        if value is None or value is missing:
             data[key] = default_value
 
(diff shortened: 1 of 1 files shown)


davidread:
Looks like we both fixed this! Closing in favour of https://github.com/ckan/ckan/pull/4448 which has tests.

wardi:
@davidread already has a similar fix in https://github.com/ckan/ckan/compare/reduce-datastore-counts?expand=1 . His keeps the current behaviour for empty strings

| default(42) | current return value | dread's return value |
| - | - | - |
| (missing) | 42 | 42 |
| None | 42 | 42 |
| 100 | 100 | 100 |
| 0 | 42 | **0** |
| False | 42 | **False** |
| [] | 42 | **[]** |
| '' | 42 | 42 |

Thoughts?

wardi:
agreed, #4448 does this better.

PR #4448, opened 2018-09-07
Fix and test for include_total not working

### LINKED ISSUE #4446: In datastore_search, parameter include_total=False doesn't work
master

When calling action function `datastore_search` you can set parameter include_total=False when you don't need the count. This should prevent an SQL query e.g. `SELECT  count(*) FROM "47e77d3c-6c77-44f2-bbf4-740b1b28bf5f"  ;` Counting rows is notoriously expensive, so we should avoid this when possible.

The parameter didn't change behaviour. The `SELECT count(*)...` occurred anyway. This causes the resource page to display slowly.

This is not normally a problem, but when the datastore db is being taxed (e.g. by ckan-xloader loading in data) the resource page for large datasets [can become unresponsive](https://github.com/ckan/ckanext-xloader/issues/41). So it is important to save on these expensive requests where we can.

You can see the `SELECT count(*) ...` occurring by turning on sqlalchemy logging, viewing a resource page (e.g. `/dataset/abc/resource/xyz` for a CSV that has been xloaded/datapushed.

If you put a pdb breakpoint in `ckan/ckanext/datastore/helpers.py:datastore_dictionary()` you see it calls datastore_search with include_total=False, however in datastore_search you see that gets changed to True when stepping past this line:
```
    data_dict, errors = _validate(data_dict, schema, context)
```

davidread:
include_total wasn't working because of the default() validator was broken, so I fixed that.

Fixes #4446

### Features:

- [ ] includes API changes

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ v.2.9.0 TBA
 Minor changes:
 
  * For navl schemas, the 'default' validator no longer applies the default when
-   the value is False, (), [] or {} (#4448)
+   the value is False, 0, [] or {} (#4448)
 
 Bugfixes:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,16 @@ Changelog
 v.2.9.0 TBA
 ==================
 
+Minor changes:
+
+ * For navl schemas, the 'default' validator no longer applies the default when
+   the value is False, (), [] or {} (#4448)
+
+Bugfixes:
+
+ * Action function "datastore_search" would calculate the total, even if you
+   set include_total=False (#4448)
+
 Deprecations:
  * ``c.action`` and ``c.controller`` variables should be avoided.
    ``ckan.plugins.toolkit.get_endpoint`` can be used instead. This function
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2280,7 +2280,8 @@ def resource_view_get_fields(resource):
 
     data = {
         'resource_id': resource['id'],
-        'limit': 0
+        'limit': 0,
+        'include_total': False,
     }
     result = logic.get_action('datastore_search')({}, data)
 
--- a/ckan/lib/navl/validators.py
+++ b/ckan/lib/navl/validators.py
@@ -78,7 +78,7 @@ def default(default_value):
     def callable(key, data, errors, context):
 
         value = data.get(key)
-        if not value or value is missing:
+        if value is None or value == '' or value is missing:
             data[key] = default_value
 
     return callable
--- a/ckan/lib/navl/validators.py
+++ b/ckan/lib/navl/validators.py
@@ -74,6 +74,8 @@ def ignore(key, data, errors, context):
     raise StopOnError
 
 def default(default_value):
+    '''When key is missing or value is 'empty', set it with a default value.
+    See the code for how 'empty' is defined.'''
 
     def callable(key, data, errors, context):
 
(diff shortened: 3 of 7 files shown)


ckanext/datastore/tests/test_search.py
```diff
@@ -634,6 +634,18 @@ def test_search_is_unsuccessful_when_called_with_invalid_fields(self):
...
+                'include_total': False}
+        postparams = '%s=1' % json.dumps(data)
+        auth = {'Authorization': str(self.normal_user.apikey)}
+        res = self.app.post('/api/action/datastore_search', params=postparams,
+                            extra_environ=auth)
+        res_dict = json.loads(res.body)
+        assert res_dict['success'] is True
+        result = res_dict['result']
+        assert 'total' not in result
+
```
wardi: Would you add this to the class at the top instead of with the legacy tests and like those tests use `helpers.call_action` to skip the request processing bits.

ckanext/datastore/controller.py
```diff
@@ -143,7 +143,7 @@ def result_page(offs, lim):
             'offset': offs,
             'sort': '_id',
             'records_format': records_format,
-            'include_total': 'false',  # XXX: default() is broken
+            'include_total': 'false',
```
wardi: we can change this to `False` now :raised_hands:

davidread:
@wardi thanks for the review. I've made those two improvements now :)

ckan/lib/navl/validators.py
```diff
@@ -78,7 +78,7 @@ def default(default_value):
     def callable(key, data, errors, context):
```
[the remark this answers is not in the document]
davidread: Good idea. Added!

davidread:
I've added these things to the changelog.

I've also removed an unnecessary total/`SELECT COUNT(*)` call by `resource_view_get_fields()`. I think it hangs well with this same PR since it is the same topic and is safe one-liner.

This PR now gets rid of 2 of the 4 `SELECT COUNT(*)` calls done when the user views the resource page. Which should make a difference to server load for large resources (e.g. 1 million rows / 100MB +) - see https://github.com/ckan/ckanext-xloader/issues/41

CHANGELOG.rst
```diff
@@ -10,6 +10,16 @@ Changelog
 v.2.9.0 TBA
 ==================
 
+Minor changes:
+
+ * For navl schemas, the 'default' validator no longer applies the default when
+   the value is False, (), [] or {} (#4448)
```
wardi: please remove `()` (shouldn't appear as part of a json payload) and add `0` to this list.

ckan/lib/navl/validators.py
```diff
@@ -74,6 +74,8 @@ def ignore(key, data, errors, context):
     raise StopOnError
 
 def default(default_value):
+    '''When key is missing or value is 'empty', set it with a default value.
+    See the code for how 'empty' is defined.'''
```
wardi: I don't think it hurts to repeat ourselves here.

`When key is missing or value is an empty string or None, replace it with a default value`

One day we could include the validator docstrings in the proper docs.

ckan/lib/helpers.py
```diff
@@ -2280,7 +2280,8 @@ def resource_view_get_fields(resource):
 
     data = {
         'resource_id': resource['id'],
-        'limit': 0
+        'limit': 0,
+        'include_total': False,
```
wardi: looks good!

davidread:
Comments addressed!

PR #4450, opened 2018-09-10
sqlalchemy-migrate -> Alembic

smotornyuk:
Use alembic instead of sqlalchemy-migrate. Main reason - I just like it. Second reason - it's more flexible and supports branching in an easy way. Third reason - [sqlalchemy-migrate has some problems with python>=3.6](https://github.com/openstack/sqlalchemy-migrate/pull/12)

I rewrote all migrations in alembic so that we can safely drop sqlalchemy-migrate before switching to Py3. Theoretically, there will be no problem for new installations, where alembic is used to create the database from scratch. For existing portals, I'm going to use Alembic for all migrations that were added in CKAN>=2.5(starting from 82). That's mean, that portals that use CKAN<2.5 must use a three-step upgrade workflow:

1.  upgrade codebase to some version >=2.5 and <2.9 
2. Upgrade database
3. Then upgrade both codebase and database to the desired version as per docs

As for CLI, I'm going to leave all commands as is and update commands internally, so there is no need in updating a lot of documentation or rewriting any of the existing installation scripts.

Are there any objections or suggestions?

PS Just for fun I'm also writing  downgrade scripts, even thou I'm not sure whether they have any value


ckan/migration/alembic.ini
```diff
@@ -0,0 +1,74 @@
...
+# version location specification; this defaults
+# to ckan/migration/alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat ckan/migration/alembic/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url = postgresql://ckan_default:root@localhost/ckan_alembic
```
amercader: Is this DB actually used? is it a placeholder?

ckan/migration/__init__.py
```diff
@@ -0,0 +1,8 @@
+# encoding: utf-8
+
+
+def skip_based_on_legacy_engine_version(op, filename):
```
amercader: Can you explain what this function does or what is needed for?

wardi:
This looks good to me. The handling and deletion of the sqlalchemy migrate version is nice.

ckan/migration/__init__.py
wardi: just a docsting would be nice e.g. "graceful upgrade from old migration numbers to alembic ids"

ckan/migration/versions/088_3ad397f70903_remove_migrate_version_table.py
```diff
@@ -0,0 +1,23 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute('DROP TABLE IF EXISTS migrate_version')
+
+
+def downgrade():
+    pass
```
wardi: maybe add a comment about creating migrate_version not being necessary once we've switched to alembic ids

ckan/migration/alembic.ini
smotornyuk: Nope, this is default configuration file benerated durinig alembic initialization, that can be used for development purposes and debuging(for example, you can run all alembic commands directly from folder, containing alembic.ini), but CKAN commands(paster, etc.) are initialized with database url from main configuration file. Probably, i should set `user:password` and `dbname` to some base values, like ones from initial configuration file. Or this line may be removed at all...

ckan/migration/versions/088_3ad397f70903_remove_migrate_version_table.py
smotornyuk: Yep, good idea

smotornyuk:
@amercader this solution should be compatible with existing instances starting from ckan-2.5. I'm checking in every migration existing of sqlalcemy-migrate's version table via `skip_based_on_legacy_engine_version` - i'll add docstring to it, but it basically compares prefix of migration file(001, 032, etc) with current version pulled from `migrate_version` table(if latter exists). If there is no such table, all alembic migrations will be applied, otherwise - as long as alembic migration is before the current version of sqlalchemy-migration, it applied without actual execution of migration script(this done via check in first two lines of `upgrade` functions). Theoretically it may work for any version of existing portals, but i tested it only for range between 2.5 and master.

davidread:
Another reason for moving to Alembic is because sqlalchemy-migrate is 'not maintained': https://groups.google.com/forum/#!topic/sqlalchemy/eF1DE20FMCQ

wardi:
I think this just needs conflicts resolved

smotornyuk:
@wardi , i've updated PR. There are two new migrations that i've converted to alembic migrations and small changes in CLI commands for database management

ckan/cli/db.py
```diff
@@ -56,17 +57,49 @@ def updatedb(version=None):
...
 
 
-@db.command(u'version', short_help=u'Returns current version of data schema')
-def version():
-    u'''Return current version'''
-    log.info(u"Returning current DB version")
+@db.command(u'downgrade', short_help=u'Downgrade the database')
+@click.option(u'-v', u'--version', help=u'Migration version', default=u'base')
+def downgradedb(version):
+    u'''Upgrading the database'''
```
wardi: Upgrading?

davidread: 😄 
```suggestion
    u'''Downgrading the database'''
```

smotornyuk: Thank you:)

PR #4249, opened 2018-05-11
Don't crash when plugin `identify()` doesn't set `g.user`.

ashleysommer:
Fixes #4247

### Proposed fixes:
Some authenticator plugins do not set `g.user` in `identify()`.
CKAN was then trying to read `g.user` and throwing an Attribute Error.
This fixes that.
No idea why this is needed now, it worked fine in 2.7.3.

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

PR #4467, opened 2018-09-20
Revert "Don't crash when plugin `identify()` doesn't set `g.user`."

amercader:
Reverts ckan/ckan#4249


amercader:
Sorry, I created this just to revert a merge done in a branch other than master

PR #4462, opened 2018-09-20
datastore_search: q/fields/records_format edge case tests and fixes

wardi:
The latest `datastore_search` code doesn't handle some combinations of parameters `q`, `fields` and `records_format`. Expand test coverage and fix issues found, including issues with `fields` list returned.

PR in progress


smotornyuk:
Looks cool(and simplified) - if no one minds, I merging it

PR #4473, opened 2018-09-21
Datastore search can estimate total, rather than expensive COUNT(*) [WIP]

davidread:
- [ ] datastore_search has new param to request estimation of total
- [ ] recline uses the estimation param
- [ ] recline tells user it is an estimate e.g. "about X results"

### Proposed fixes:

Unresponsive resource preview page, for large resources in datastore
https://github.com/ckan/ckanext-xloader/issues/41

### Features:

- [x] includes API changes


davidread:
@wardi Initial feedback welcome!

ckanext/reclineview/theme/public/vendor/recline/recline.js
```diff
@@ -2619,7 +2622,10 @@ my.MultiView = Backbone.View.extend({
         </div> \
       </div> \
       <div class="recline-results-info"> \
-        <span class="doc-count">{{recordCount}}</span> records\
+      {{#recordCountWasEstimated}} \
+        <span class="doc-count-approx">about</span> \
+      {{/recordCountWasEstimated}} \
+      <span class="doc-count">{{recordCount}}</span> records \
```
wardi: "about" and "records" need to be translated here.

Would be nice to truncate the estimate too, e.g. `about 315k records`

wardi:
this adds two extra round-trips to the DB when calculating the total. After measuring performance maybe they can be combined into the same sql statement that gets the actual total to see if it's noticeable for small datasets.

ckanext/datastore/backend/postgres.py
```diff
@@ -1287,17 +1288,60 @@ def search_data(context, data_dict):
...
-            resource=resource_id,
-            ts_query=ts_query,
-            where=where_clause)
-        count_result = _execute_single_statement(
-            context, count_sql_string, where_values)
-        data_dict['total'] = count_result.fetchall()[0][0]
+        total_estimation_threshold = \
+            data_dict.get('total_estimation_threshold') or 0
+        estimated_total = None
+        if total_estimation_threshold != 0 and not (where_clause or distinct):
```
wardi: I don't think 0 should be "off" in this case. How about we use the option to not pass this value when you want no limit. That lets sites apply their own default and validators to clamp this to a reasonable value like in #4484

ckanext/reclineview/theme/public/vendor/recline/recline.js
jqnatividad: What about just using [tilde (~)](https://en.wikipedia.org/wiki/Tilde)?

davidread: Recline is not translated anyway - I guess there are JS reasons for it. So I'd rather leave that to a separate effort [
screen shot 2018-10-12 at 13 56 23
](https://demo.ckan.org/it/dataset/newcastle-city-council-payments-over-500/resource/0c2bd47a-6ac5-412e-a337-1b45a952e07e)

davidread:
> this adds two extra round-trips to the DB when calculating the total. After measuring performance maybe they can be combined into the same sql statement that gets the actual total to see if it's noticeable for small datasets.

I tested a small table of 10 rows and overall the datastore_search request on my machine is taking about 75ms select count(*) and about 90ms with the count estimation.

postgres is reporting these database queries as taking 0.3-0.7ms, apart from the check of when the last analyze which is about 12ms, which is the culprit. So I'll have a dig into that.

wardi:
Maybe a function https://github.com/ckan/ckan/blob/master/ckanext/datastore/set_permissions.sql#L76-L89 to calculate the exact/approx row count that returns the number and whether an estimate was used?

davidread:
@wardi I've taken out the check to find out when the estimate/stat was last done (this is a slow lookup and we can infer if it is happened anyway by the value of the estimate), so there's only one potential extra round-trip now.

The logic is this: when doing datastore_search, and the user wants to know the total number of results, and the user indicated an estimate would be ok above a certain threshold, CKAN will query the db for the estimate (takes 0.5ms). If it's a smaller dataset (estimate is below the threshold) then it will do a second query, the SELECT COUNT(*) query (takes 0.3ms for 10 records, up to 1000ms for 1e6 records). 

So I don't think it is worth creating a function to shave off the 0.5ms.

If you're happy with this, then please review @wardi. I will just add in a configurable limit in a separate PR as that is a nice to have.

wardi:
@davidread yes that sounds good

ckanext/datastore/logic/action.py
```diff
@@ -405,6 +433,16 @@ def datastore_search(context, data_dict):
...
     :type include_total: bool
+    :param total_estimation_threshold: By default, the "total" returned is a
+            precise count of the query results, which can be computationally
+            expensive for large results (e.g. >100,000k rows). By setting this
+            value to non-zero, and the estimated total is above this value,
+            then datastore_search will skip the precise count - it just returns
+            the estimated total. The estimated total is generated by EXPLAIN,
+            which is triggered by Express Loader or DataPusher as part of
+            loading the data. NB Total estimation is not compatible when
+            specifying 'filters' or 'distinct' options. (optional, default: 0)
```
wardi: as I mentioned above, I find "0 means infinite" really confusing in an API. Can we use None(null) or some other value as the default "always calculate exact" value?

ckanext/datastore/backend/postgres.py
davidread: Agreed. I've changed it so that total_estimation_threshold=None means "off", which is also the default. And good idea - a site can clamp it by adding `limit_to_configured_maximum` to the schema (once #4484 is merged).

ckanext/datastore/logic/action.py
davidread: Done!

davidread:
@wardi I've addressed the threshold=0 problem. Would you mind looking at this again?

ckanext/datastore/logic/action.py
```diff
@@ -433,16 +433,17 @@ def datastore_search(context, data_dict):
...
+    :param total_estimation_threshold: If "include_total" is True and
+        "total_estimation_threshold" is not None and the estimated total
+        (matching record count) is above the "total_estimation_threshold" then
+        this datastore_search will return an *estimate* of the total, rather
+        than a precise one. This is often good enough, and saves
+        computationally expensive row counting for larger results (e.g. >100000
+        rows). The estimated total comes from the PostgreSQL table statistics,
+        generated when Express Loader or DataPusher finishes a load, or by
+        autovacuum. NB Total estimation is currently not compatible when
+        specifying 'filters' or 'distinct' options. (optional, default: None)
```
wardi: the "not compatible" wording makes it sound like the number will be wrong if you pass total_estimation_threshold with filters or distinct. Won't filters or distinct just prevent the estimation from happening (like if total_estimation_threshold was set to None?)

Issue #3511, opened 2017-03-28
organization_list / group_list [performance] in CKAN 2.6


TkTech:
### CKAN Version if known (or site URL)

2.6

### Please describe the expected behaviour

`org/group_list` with `all_fields=true` should have reasonable sub-1s performance. As reported on the mailing list:

> Hi,
> 
> in the CKAN project I work for we are currently upgrading from CKAN 2.3 to 2.6. 
> 
> One thing I noticed after the upgrade is that organization_list() (and group_list() for that matter) seems to be alot slower than it used to be on 2.3. Checked on demo.ckan.org ( http://demo.ckan.org/api/action/organization_list?package_count=true&all_fields=true ) and it takes over 40 seconds. 
> 
> For now I'm caching the output but is there a better way of dealing with this (especially for just getting name, title and package_count) ? 
> 
> Thanks for any advise,
> Alex

### Please describe the actual behaviour

40s+ on the demo site. This seems to be bound to the # of groups (every future time I say groups I mean both orgs and groups, same mechanic) and not the # of revisions or packages in them.

Using `all_fields=true` triggers a `group_show()` call on each group. This in turn triggers thousands of other queries (such as auth checks) and further queries incurred by `group_dictize`. I have not had a chance to determine the actual cause of the performance drop vs. v2.3.

`all_fields` probably shouldn't even be an allowed option, nor should the action allow getting every group/org with no realistic limit to enforce pagination. IMO We should back these entirely from the search index.

### What steps can be taken to reproduce the issue? 

http://demo.ckan.org/api/action/organization_list?package_count=true&all_fields=true

amercader:
I completely agree. There simply no way we can get decent performances with `all_fields=True` unless the objects are on the search index. Settling on a modern Solr version that allows schema-less indexing (or using ElasticSearch) will hopefully help with this in the future.

TkTech:
@amercader https://github.com/TkTech/sachi/blob/master/tests/test_basics.py#L15

Working on it. Solr & Postgres supported at the moment, facets will be next then it's suitable to drop into CKAN.

davidread:
In the meantime, I've added some tight limits on the number of results. See: https://github.com/ckan/ckan/pull/4484

tino097:
Old CKAN version, if the issue is still relevant we can reopen it

PR #4484, opened 2018-09-28
Add limits for action functions v2 [WIP]

### LINKED ISSUE #4480: Add limits for action functions
Several API calls are at risk of users requesting large expensive database queries. should add some sensible limits to the number of rows.

We currently limit the number of rows you can return in package_search.

davidread:
Fixes #4480

### Proposed fixes:

* [x] package_search - already has limit 1000
* [ ] group/organization_show
* [ ] group/organization_list (and see all_fields issue #3511)
* [x] recently_changed_packages_activity_list (_html)
* [x] package/user/group/organization_activity_list (_html)
* [ ] dashboard_activity_list (_html)

### Features:

- [ ] includes API changes


davidread:
@wardi it would be great to get this merged, if you have a chance to review?

davidread:
@ThrawnCA I have sympathy for this point of view - it does break functionality. However we were keen to add it as a patch for earlier versions, because it was seen as a DOS vector. The default limits are so low, as to try and make it obvious they are limited.

ThrawnCA:
> The default limits are so low, as to try and make it obvious they are limited.

That's a nice thought, but since it still truncates silently, and still allows enough to eg populate a dropdown list enough to need significant scrolling down, it really isn't obvious unless someone was already specifically looking for it.

davidread:
Agreed, it's problematic for the use case you describe. We had some awareness of this sort of thing when we designed it, balanced against the other concerns mentioned. The questions are whether that balance is wrong and what might be done now.

PR #4488, opened 2018-10-02
Remove instruction to create datastore db

### LINKED ISSUE #4378: Docker-compose: Datastore creation not working
Hi there, running into a problem with the docker-compose setup:

2.8.1

I'm following: http://docs.ckan.org/en/2.8/maintaining/installing/install-from-docker-compose.html
where:
docker exec -it db psql -U ckan -f 00_create_datastore.sh
should setup the database

docker exec -it db psql -U ckan -f 00_create_datastore.sh
returns:
./00_create_datastore.sh: No such file or directory

The Dockerfile seems to add the postgresql/docker-entrypoint-initdb.d directory
So I tried:
sudo docker exec -it db psql -U ckan -f docker-entrypoint-initdb.d/00_create_datastore.sh

However, this also cannot work because the command above already is a call to psql but the
file is a shellscript including a call to psql itself, resulting in:

psql:docker-entrypoint-initdb.d/00_create_datastore.sh:5: ERROR:  syntax error at or near "#!/"
LINE 1: #!/bin/bash

Is this a bug? I'm confused

Just follow: http://docs.ckan.org/en/2.8/maintaining/installing/install-from-docker-compose.html
until:

3. a) docker exec -it db psql -U ckan -f 00_create_datastore.sh

Thanks for your help,
Christoph

spikeheap:
Fixes #4378

### Proposed fixes:

This PR removes the instruction to create the datastore db, as  the datastore creation script is run automatically when the Postgres `db` container first runs. See the [docker image docs](https://docs.docker.com/samples/library/postgres/#how-to-extend-this-image) for more information on automatic script execution in`/docker-entrypoint-initdb.d`.

### DIFF
--- a/doc/maintaining/installing/install-from-docker-compose.rst
+++ b/doc/maintaining/installing/install-from-docker-compose.rst
@@ -168,20 +168,17 @@ against the database and import / export files from ``$VOL_CKAN_HOME``.
 ---------------------------
 3. Datastore and datapusher
 ---------------------------
-To enable the datastore, the datastore database and database users have to be created before
-enabling the datastore and datapusher settings in the ``production.ini``.
+The datastore database and user are created when the `db` container is first started, however we need to do some additional configuration before enabling the datastore and datapusher settings in the ``production.ini``.
 
-a. Create and configure datastore database
+a. Configure datastore database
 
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @spikeheap that makes sense

PR #4489, opened 2018-10-02
Correctly handles site_root in Flask requests

### LINKED ISSUE #4452: CKAN root path repeated in URL
2.8.0

We have a CKAN site that should be accessed via http://192.168.183.128/data
The expected CKAN URL should be http://192.168.183.128/data/<ckan_page>

When viewing some of the CKAN pages in a browser, the url root is repeated.
Therefore the site cannot be reached.

The browser tries to access http://192.168.183.128/data/data/<ckan_page>

Examples are the login, logout and admin pages
http://192.168.183.128/data/data/user/login
http://192.168.183.128/data/data/user/logged_out_redirect
http://192.168.183.128/data/data/user/ckan_admin

Once logged in, the URLs are correct for dataset, organization, group etc pages.

production.ini:
ckan.site_url http://192.168.183.128
ckan.root_path = /data/{{LANG}}

ckan_default.conf:
WSGIScriptAlias /data /etc/ckan/default/apache.wsgi

datapusher.conf:
WSGIScriptAlias /data /etc/ckan/datapusher.wsgi

smotornyuk:
Fixes #4452
And [here is the problem](https://github.com/pallets/werkzeug/blob/master/werkzeug/routing.py#L2085)

Running CKAN as wsgi script under nested path( `WSGIScriptAlias /data /path/to/wsgi`) ([small modification of suggested solution for deploing CKAN with apache](https://docs.ckan.org/en/2.8/maintaining/installing/deployment.html#create-the-apache-config-file))creates small problem if one is using `ckan.root_path`. Apache sets environment variable `SCRIPT_NAME` with 
 value equals `/data` in example above. Werkzeug prepends this value automatically to every generated url, and thus we have conflict with our root_path logic. Werkzeug solution not usable, as it does not account current locale, so i suggest just to strip this prefix


ckan/lib/helpers.py
```diff
@@ -331,14 +333,13 @@ def url_for(*args, **kw):
...
-        if _auto_flask_context:
+        # Pylons does not add current site_root
+        is_werkzeug = is_werkzg    if _auto_flask_context:
             _auto_flask_context.pop()
 
     # Add back internal params
     kw['__ckan_no_root'] = no_root
-
-    # Rewrite the URL to take the locale and root_path into account
+    kw['__is_werkzeug'] = is_werkzeuis_we tis_we the locale and root_path into account
```
wardi: ?

wardi:
https://github.com/ckan/ckan/blob/master/ckan/config/middleware/common_middleware.py#L14-L30 is a hack we've had for a long while that removed the SCRIPT_PATH environment variable for pylons so it wouldn't add the root_path twice. What do you think about a similar solution?

One problem with relying on SCRIPT_PATH is you can't support things like `root_path = /{{LANG}}/data` We should at least make the solution consistent.

smotornyuk:
Nice, hadn't noticed this middleware, but it definitely has sense to change my solution with existing tools, so i'll rewrite PR on Monday

smotornyuk:
I put logic into CKANFlask extension(middleware). SCRIPT_NAME must be removed on early stage, before flask internally creates url mapper(this happens [here](https://github.com/ckan/ckan/blob/master/ckan/config/middleware/flask_app.py#L371) ). 

I don't want to reuse existing middleware, because middlewares differ from flaskExtensions and thus I need to update existing class with obscure lines of code. Neither it a good idea, to create new class, because in that case relationship between wiping out SCRIPT_NAME and creation of url mapper will be too implicit

Zharktas:
Apparently this patch breaks pylons debugging URLs.

vagrant@vagrant-web:~$ curl https://10.10.10.10/data/dataset --insecure -I
HTTP/1.1 500 Internal Server Error
Server: nginx

Date: Mon, 14 Jan 2019 07:53:56 GMT
Content-Type: text/html
Connection: keep-alive
X-Debug-URL: https://10.10.10.10/_debug/view/1547452312
X-Content-Type-Options: nosniff

The Debug URL should contain site-root, but it doesn't.

smotornyuk:
@Zharktas , I slightly changed PR, restoring the state of environment variables after initializing Flask app, so now all Pylons URLs should be generated correctly

smotornyuk:
Updated once again. It appears that `root_path` still duplicated if I restore it after router initialization. Instead, i'm cleaning `SCRIPT_NAME` just before flask agrees to handle request

PR #4491, opened 2018-10-04
Use @chained_auth_function with core auth functions

alycejenni:
This is essentially an updated PR based on #4248 opened by @ashleysommer several months ago, with the formatting fixes and a new test. I take zero credit for the original fix!

### Proposed fixes:

From the original PR:
> Some authentication plugins implement chained auth functions on actions such as `user_create` and `user_update`. If no other plugins provide these functions, then CKAN raises the exception "The auth 'user_create' is not found for chained auth". The CKAN built-in auth system however does provide these functions. So it makes sense to fallback to having the chained auth function chain to the corresponding built-in function.
> 
> This is an example of one such plugin which attempts to chain off 'user_create' and 'user_update' and fails:
> https://github.com/NaturalHistoryMuseum/ckanext-ldap/blob/6f9aec4edadaf792e347b4d20adad2c2322c167f/ckanext/ldap/logic/auth/create.py#L12

I also fixed one other small bug in `authz.py`: the error message displays the `auth_function` name, but chained functions are partial functions and don't have names, so it was failing. It now uses the name of the `.func` if it's a partial.

### Features:

- [ ] includes API changes


amercader:
@wardi I took the liberty of reviewin, fixing and merging this as I was working on it.

PR #4497, opened 2018-10-12
Filtered Datatables Downloads

wardi:
Add a Download button to the datatables view to download filtered data based on view filters, user filters and datatables query string and sort order.


wardi:
**FLASK fix included**: I needed to change `request.params` to `request.args` in the snippet now that the dataset views are served with Flask. This likely affects other views, can we improve backwards compatibility for external plugins by aliasing `request.params` to `request.args`?

wardi:
should be complete now. Only the visible columns are downloaded

wardi:
@smotornyuk like 6bc233e ?

smotornyuk:
Yes, exactly

PR #4501, opened 2018-10-15
Remove old markup example.

boykoc:
Removed in favour of newer primer.html.

Fixes #

### Proposed fixes:

The older markup.html example currently does not load properly as can be seen on the [demo site](https://demo.ckan.org/development/markup.html). When looking into this I noticed that the [docs](https://docs.ckan.org/en/latest/theming/templates.html#html-tags-and-css-classes) do not reference this page. Instead they point to the newer primer.html and the bootstrap documentation. Both of these resources cover all of the examples given in the markup.html.

### Features:

- [ ] includes API changes


amercader:
Thanks @boykoc it sounds sensible to remove this.

PR #4505, opened 2018-10-17
Fix i18n API encoding

### LINKED ISSUE #4504: i18n API returns JSON-encoded JSON
- Correct: https://demo.ckan.org/api/i18n/fi
- Incorrect: https://beta.ckan.org/api/i18n/fi

Problem probably stems from changes [here](https://github.com/ckan/ckan/blob/master/ckan/views/api.py#L467).

The API should return JSON.

The API returns JSON-encoded JSON. This breaks JavaScript translations, as the returned object deserializes into a string rather than an object.

Call the API

bzar:
Fixes #4504 

### Proposed fixes:
- Make `i18n_js_translations` call `_finish_ok` with an object response data like all the others
- Fix `_finish` and `_finish_ok` documentation to reflect the expected `response_data` type for "json" `content_type`

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/views/api.py
+++ b/ckan/views/api.py
@@ -44,7 +44,7 @@ def _finish(status_int, response_data=None,
     :param status_int: The HTTP status code to return
     :type status_int: int
     :param response_data: The body of the response
-    :type response_data: object if content_type is `text`, a string otherwise
+    :type response_data: object if content_type is `text` or `json`, a string otherwise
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @bzar!

tino097:
@amercader i've checked, this dont need to be backported to 2.7.5

PR #4511, opened 2018-10-18
Ubuntu 18.04 docs

nushkovg:
Fixes documentation on installing CKAN on Ubuntu 18.04 64-bit.

- [ ] includes API changes

### DIFF
--- a/doc/maintaining/installing/install-from-source.rst
+++ b/doc/maintaining/installing/install-from-source.rst
@@ -88,7 +88,7 @@ a. Create a Python `virtual environment <http://www.virtualenv.org>`_
 
     If you are using ``python3`` and if you are having problems installing the
     requirements using ``pip`` inside the virtual environment, explicitly
-    install ``python2.7`` inside the virtual environment. 
+    instruct it to use ``python2.7``. 
 
     .. parsed-literal::
 
--- a/doc/maintaining/installing/install-from-source.rst
+++ b/doc/maintaining/installing/install-from-source.rst
@@ -90,10 +90,12 @@ a. Create a Python `virtual environment <http://www.virtualenv.org>`_
     requirements using ``pip`` inside the virtual environment, explicitly
     install ``python2.7`` inside the virtual environment. 
 
-    sudo mkdir -p |virtualenv|
-    sudo chown \`whoami\` |virtualenv|
-    virtualenv --python=/usr/bin/python2.7 --no-site-packages |virtualenv|
-    |activate|
+    .. parsed-literal::
+
+        sudo mkdir -p |virtualenv|
+        sudo chown \`whoami\` |virtualenv|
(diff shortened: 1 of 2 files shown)


doc/maintaining/installing/install-from-source.rst
```diff
@@ -21,7 +20,11 @@ work on CKAN.
 --------------------------------
 
 If you're using a Debian-based operating system (such as Ubuntu) install the
-required packages with this command for Ubuntu 16.04::
+required packages with this command for Ubuntu 18.04::
```
amercader: AFAICT the list of packages is the same in both right? so can we simplify it a bit and just say

> install required packages with this command for Ubuntu 16.04 and 18.04::

doc/maintaining/installing/install-from-source.rst
```diff
@@ -80,6 +83,19 @@ a. Create a Python `virtual environment <http://www.virtualenv.org>`_
        sudo chown \`whoami\` |virtualenv|
        virtualenv --no-site-packages |virtualenv|
        |activate|
+    
+.. note::
+
+    If you are using ``python3`` and if you are having problems installing the
```
amercader: Can we reword this section to something like:

```
.. note::
    If your system uses Python 3 by default (eg Ubuntu 18.04) make sure to create the virtualenv using the Python 2.7 executable with the ``--python`` option:

      .. parsed-literal::
         virtualenv --python=/usr/bin/python2.7 --no-site-packages |virtualenv|

```

doc/maintaining/installing/solr.rst
```diff
@@ -12,7 +12,22 @@ installed, we need to install and configure Solr.
    server, but CKAN doesn't require Jetty - you can deploy Solr to another web
    server, such as Tomcat, if that's convenient on your operating system.
 
-#. Edit the Jetty configuration file (``/etc/default/jetty8`` or
+.. tip::
+
+   Do this step only if your system is using ``jetty9``.
```
amercader: It's probably best to be explicit and say "Do this step only if you are using Ubuntu 18.04."

doc/maintaining/installing/install-from-source.rst
nushkovg: I agree. It looks much clearer this way.

doc/maintaining/installing/install-from-source.rst
nushkovg: I have to admit I had problems paraphrasing this section. This is much better, you just wrote everything I was thinking exactly like it should be. Thanks!

doc/maintaining/installing/solr.rst
nushkovg: Since Ubuntu 18.04 for some reason supports only jetty9, yes, this might be the better solution.

tino097:
Congrats @nushkovg on your first contribution

PR #4517, opened 2018-10-23
Unicode filenames

### LINKED ISSUE #4507: filenames with unicode characters should keep the original file name when saving to disk
### reproduction
* Upload a resource which has a file name with unicode characters
* View the resource page and inspect the download link

### expected
* Download link should contain the original filename with the unicode characters

### actual
* Download link truncates the unicode characters

## example
* https://demo.ckan.org/dataset/test-hebrew/resource/240cabfe-0057-4aea-9ce0-dbfb708bc484
* original file name: `נוכחות בישיבות ועדה 1-5-2017 עד 30-7-2017.xlsx`
* download url: `https://demo.ckan.org/dataset/0d9acb2e-638a-4d73-8511-2ff52a7f654a/resource/240cabfe-0057-4aea-9ce0-dbfb708bc484/download/---1-5-2017--30-7-2017.xlsx`

frafra:
Fixes #4507 

### Proposed fixes:
Do not remove/replace Unicode characters with 7-bit ASCII and allow Unicode letters and numbers.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/munge.py
+++ b/ckan/lib/munge.py
@@ -7,8 +7,9 @@
 
 import os.path
 import re
+import unicodedata
 
-from six import text_type
+from six import text_type, u
 
 from ckan import model
 from ckan.lib.io import decode_path
@@ -22,16 +23,16 @@ MAX_FILENAME_TOTAL_LENGTH = 100
 # Minimum total length of a filename (including extension)
 MIN_FILENAME_TOTAL_LENGTH = 3
 
+# Accepted Unicode categories
+UNICODE_CATEGORIES = ['Ll', 'Lm', 'Lo', 'Lt', 'Lu', 'Nd']
+
 
 def munge_name(name):
     '''Munges the package name field in case it is not to spec.'''
-    # substitute non-ascii characters
-    if isinstance(name, text_type):
-        name = substitute_ascii_equivalents(name)
     # separators become dashes
     name = re.sub('[ .:/]', '-', name)
     # take out not-allowed characters
-    name = re.sub('[^a-zA-Z0-9-_]', '', name).lower()
+    name = _unicode_cleanup(name, keep='-_').lower()
     # keep it within the length spec
     name = _munge_to_length(name, model.PACKAGE_NAME_MIN_LENGTH,
                             model.PACKAGE_NAME_MAX_LENGTH)
@@ -40,13 +41,10 @@ def munge_name(name):
 
 def munge_title_to_name(name):
     '''Munge a package title into a package name.'''
-    # substitute non-ascii characters
-    if isinstance(name, text_type):
-        name = substitute_ascii_equivalents(name)
     # convert spaces and separators
     name = re.sub('[ .:/]', '-', name)
     # take out not-allowed characters
-    name = re.sub('[^a-zA-Z0-9-_]', '', name).lower()
+    name = _unicode_cleanup(name, keep='-_').lower()
     # remove doubles
     name = re.sub('-+', '-', name)
     # remove leading or trailing hyphens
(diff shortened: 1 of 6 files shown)


wardi:
We should replace all non-"letter" category characters, i.e.
```python
not unicodedata.category(char).startswith('L')
```

frafra:
Unicode letters are ok, Unicode non-letters are not. Got it :-)

amercader:
@frafra Many thanks for your patch. As @wardi mentioned we still need to strip non-letter characters so IIRC your changes in e8f0631a1 are not needed (ie these tests should still pass unchanged).

Perhaps is useful to update the tests in test_munge.py with some examples like the ones @OriHoch mentions in the issue and work from there.

wardi:
Hmm. numbers should be allowed too. Better to list the specific categories we'll allow
http://www.fileformat.info/info/unicode/category/index.htm

@amercader what do you think of allowing:

Code | Description
-- | --
[Ll] | Letter, Lowercase
[Lm] | Letter, Modifier
[Lo] | Letter, Other
[Lt] | Letter, Titlecase
[Lu] | Letter, Uppercase
[Nd] | Number, Decimal Digit

wardi:
https://pypi.org/project/regex/ looks useful, it can match unicode categories without looping over every character

amercader:
> @amercader what do you think of allowing:

Just for completeness should we maybe also include the following?

Code | Description
-- | --
[Nl] | Number, Letter
[No] | Number, Other

frafra:
If we include every number and letter we could just use this regex (plus few additional characters like `-_.` that are specified inside the functions):

```python
regex.sub(r'[^\p{NUMBER}&&\p{LETTER}]', '', text)
```

I would also have to add a new requirement for `regex`. Let me know, so I can do a new commit.

frafra:
Ok, it seems pretty clean to me. Commits can be squashed when merged and/or more tests could be added.

ckan/lib/munge.py
```diff
@@ -134,9 +77,8 @@ def munge_filename_legacy(filename):
     Upload.update_data_dict() because if this routine changes then group images
     uploaded previous to the change may not be viewable.
     '''
-    filename = substitute_ascii_equivalents(filename)
     filename = filename.strip()
-    filename = re.sub(r'[^a-zA-Z0-9.\- ]', '', filename).replace(' ', '-')
```
wardi: the name and docstring of this function suggest we shouldn't change anything here. If we can remove this it would be great as a separate PR

frafra: @wardi Do you suggest to keep using `substitute_ascii_equivalents` in this function for this PR?

frafra: Right.  Removed.

frafra:
All test are go.
@OriHoch would you like to try this branch and see if it fixes the bug you reported, before merging to master?

OriHoch:
thanks, looks like there is something wrong with the encoding, this is what I'm getting - 


image

This is the file (remove .txt extension):

[נתוני נוכחות במליאה לפי נתוני הצבעה - כנסת 20 - מעודכנים ל18 ליולי 2018.ods.txt](https://github.com/ckan/ckan/files/2535112/-.20.-.18.2018.ods.txt)

(looks like github are munging it too..)

frafra:
Thank you @OriHoch. I think this is due to CKAN default `schema.xml`, which uses [ASCIIFoldingFilter](https://lucene.apache.org/solr/guide/6_6/filter-descriptions.html#FilterDescriptions-ASCIIFoldingFilter) for text strings. Could you please test again?

There is just one legacy test failing, so if it works for you, I am going to find a solution. Thanks!

OriHoch:
thanks, will test later today

OriHoch:
The bug persists, 

this is how the URL is stored in the DB - 
image

and this is in SOLR - 
image

frafra:
Does anyone has a clue about it? It seems something related more to Solr configuration that to CKAN.

kmbn:
@frafra Are you still working on this PR? In addition to the issue @OriHoch mentioned, the tests fail on CircleCI.

frafra:
@kmbn The patch is halfway trough, but as I wrote in the last comment I would need help to fix it, because it looks like Solr is involved.

pwalsh:
@frafra @amercader @wardi 

Instead of custom munging, would it be possible (and easier, and well tested) to use the [Mozilla slugify library](https://github.com/mozilla/unicode-slugify)? I have used this extensively on projects that require unicode URLs.

tino097:
This PR is old and some of the changes could not apply or have been fixed within other PR's

PR #4528, opened 2018-10-26
[#3592] Implement python-feedgen and deprecate webhelpers for Atom feeds.

### LINKED ISSUE #3592: Deprecate webhelpers.feedgenerator and use Jinja for feeds
Related to issue #3309 

latest

CKAN should be migrated to Python 3

The webhelpers package does not support Python 3.

# Goal
Replace webhelpers.feedgenerator based feed(s) with Jinja.

nushkovg:
Fixes #3592 

### Proposed fixes:

1. Implement python-feedgen as feed generator for Atom feeds. This module is maintained (the last update was in May 2018), unlike the current webhelpers module.
2. Remove the custom _FixedAtom1Feed class, because python-feedgen contains its features out-of-the-box.
3. Use the iso8601 module for simple ISO datetime format conversion, which makes the updated and published dates of datasets easier to maintain.

I had trouble with the link, categories and additional_fields features so I might need some help there. However, the weird thing with the link is that it gets generated even if that line is deleted, so I might be missing something. I have added comments in the code for these three too, so lets see if this fix fits the needs before anything else is done.

### DIFF
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -1121,7 +1121,7 @@ def history(package_type, id):
                 title=item_title,
                 link=item_link,
                 description=item_description,
-                author={'name': item_author_name},
+                author={u'name': item_author_name},
                 pubDate=item_pubdate,
             )
         response = make_response(feed.atom_str(encoding=u'utf-8'))
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -1084,13 +1084,13 @@ def history(package_type, id):
     format = request.args.get(u'format', u'')
     if format == u'atom':
         # Generate and return Atom 1.0 document.
-        from webhelpers.feedgenerator import Atom1Feed
-        feed = Atom1Feed(
+        from feedgen.feed import FeedGenerator
+        feed = FeedGenerator(
             title=_(u'CKAN Dataset Revision History'),
             link=h.url_for(
                 controller=u'revision', action=u'read', id=g.pkg_dict[u'name']
             ),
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -61,6 +61,8 @@ vdm==0.14
 webencodings==0.5.1       # via html5lib
 weberror==0.13.1          # via pylons
 webhelpers==1.3
+feedgen==0.7.0
```
torfsen: I haven't looked at the other parts, but the new dependencies should be added to `requirements.in` instead. Then use `pip-compile` to re-generate `requirements.txt`. Both `requirements.in` and `requirements.txt` are then committed.

kmbn:
@nushkovg some of the tests failed on CircleCI—do you want to check them out and update your PR?

ckan/views/dataset.py
```diff
@@ -1084,13 +1084,13 @@ def history(package_type, id):
     format = request.args.get(u'format', u'')
     if format == u'atom':
         # Generate and return Atom 1.0 document.
-        from webhelpers.feedgenerator import Atom1Feed
-        feed = Atom1Feed(
+        from feedgen.feed import FeedGenerator
+        feed = FeedGenerator(
```
frafra: `FeedGenerator` accepts no arguments, that is why the tests are failing. This piece of code and the tests an functions depending on the how Atom1Feed get initialized should be changed.

frafra:
This PR does not take into account `GeoAtom1Feed` at the moment.

tino097:
Fixed in #5054

Issue #4535, opened 2018-10-30
schema.xml has to be changed for non-extra fields


frafra:
### CKAN Version if known (or site URL)
2.8.1

### Please describe the expected behaviour
CKAN docs should refer to https://lucene.apache.org/solr/guide/6_6/defining-fields.html for custom fields.

### Please describe the actual behaviour
No reference about how to add fields to Solr schema. I was not able to add a custom field, so I asked on gitter and @Zharktas replied to me about that, explaining that schema.xml has to be modified.

frafra:
I tried to use Solr API, but it is not possible if you use ckan/solr Docker image because CKAN uses `ClassicIndexSchemaFactory` instead of `ManagedIndexSchemaFactory` that is the default in recent 6.6 releases of Solr.
See: https://github.com/ckan/ckan/blob/master/contrib/docker/solr/solrconfig.xml#L291

Should we switch to `ManagedIndexSchemaFactory` for the Docker image and then update the documentation?

frafra:
PR has been merged, but we still lack of documentation on how to extend the schema.
It could be related to: https://github.com/ckan/ckanext-scheming/issues/195

PR #4536, opened 2018-10-30
Switching from schema.xml to ManagedSchema

frafra:
Needed for #4535.

### Proposed fixes:

This patch allows to use Solr API to add fields, instead having to modify the schema.xml file directly.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/search/__init__.py
+++ b/ckan/lib/search/__init__.py
@@ -55,7 +55,8 @@ _QUERIES = {
     'package': PackageSearchQuery
 }
 
-SOLR_SCHEMA_FILE_OFFSET = '/admin/file/?file=schema.xml'
+SOLR_SCHEMA_FILE_OFFSET_MANAGED = '/schema?wt=schema.xml'
+SOLR_SCHEMA_FILE_OFFSET_CLASSIC = '/admin/file/?file=schema.xml'
 
 
 def _normalize_type(_type):
@@ -246,6 +247,21 @@ def clear_all():
     log.debug("Clearing search index...")
     package_index.clear()
 
+def _get_schema_from_solr(file_offset):
+    solr_url, solr_user, solr_password = SolrSettings.get()
+
+    http_auth = None
+    if solr_user is not None and solr_password is not None:
+        http_auth = solr_user + ':' + solr_password
+        http_auth = 'Basic ' + http_auth.encode('base64').strip()
+
+    url = solr_url.strip('/') + file_offset
+
+    req = urllib2.Request(url=url)
+    if http_auth:
+        req.add_header('Authorization', http_auth)
+
+    return urllib2.urlopen(req)
 
 def check_solr_schema_version(schema_file=None):
     '''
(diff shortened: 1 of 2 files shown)


frafra:
I am pretty sure that circleci is failing because the test are run against the old Solr configuration.
On my PC everything seems to work as usual, but I am not able to run the tests.

wardi:
This isn't a small change or something that we can do just in the docker images.

We'll need to update the source install instructions and add a command that will create and populate a solr core for use by ckan. Then we need a way for extensions to hook in to that command to add their own fields, languages and anything else they currently set up by customizing the solr schema.

@frafra the tests are very useful for development! Maybe set up a source install so you can run the tests locally?

frafra:
@wardi I will try to setup properly now that I am trying to do more complex commits.

Are you proposing to drop entirely the CKAN schema and just use the new API? I was thinking just to add the new capabilities related to managed schema without breaking compatibility at the moment, so no documentation change would be needed at the beginning.

The default schema.xml URL seems the only change that would break backwards compatibility; I do not know if the best way to handle it would be to check  both URLs by default or if there is a way to do a request to detect if the new schema factory is in use.

I am not sure how/when the `solrconfig.xml` located in the CKAN repository was generated and if it needs to be updated. I suppose that it is something shipped with the Sorl by the Linux distribution we should change it in order to change the parameter (I do not think that it is generated on-the-fly, because Solr documentation says that recent 6.6 releases use the new managed schema).

wardi:
I thought solr was either-or. You can't start with a schema.xml but also have solr managing the schema can you?

TkTech:
> 
> 
> I thought solr was either-or. You can't start with a schema.xml but also have solr managing the schema can you?

You can, solr will copy the schema.xml on startup to managed-schema by default and start using it instead.

frafra:
Ok, managed schema is used by default, but it uses the old one if it is not found. This should avoid breaking backwards compatibility. It is up to the system administrator or solr packager to choose between one of them, CKAN supports both.

We can extend and update the documentation in a successive pull request, if we want to go in this direction. It could be a feature to showcase for the next minor release (CKAN 2.9.0) I suppose.

torfsen:
+1 for an easy way for extensions to dynamically update the Solr schema. That would be very useful for ckanext-extractor.

frafra:
It seems to work as expected. Here is what I am using inside my plugin (that is an example):

```
    def update_config(self, config_):
        toolkit.add_template_directory(config_, 'templates')
        toolkit.add_public_directory(config_, 'public')
        toolkit.add_resource('fanstatic', 'coat')
        self._custom_schema()

    def _custom_schema(self):
        response = requests.get(CKAN_SCHEMA+'/copyfields')
        copyfields = response.json()['copyFields']
        if {"dest": "version_i", "source": "version"} in copyfields:
           return
        requests.post(CKAN_SCHEMA, json={
            "add-field":{
                "name": "version_i",
                "type": "int",
                "stored": True,
            }
        })
        requests.post(CKAN_SCHEMA, json={
            "add-copy-field":{
                "source": "version",
                "dest": ["version_i"],
            }
        })
```

It would be nice to have some helpers for such kind of requests and checks in the future.

frafra:
Proposal: accept the patch, which provide additional flexibility without breaking compatibility; after that a more inclusive discussion about managed schema could be started.

wardi:
@TkTech what do you think about this change as-is?

TkTech:
Certainly in favor, as long as we ensure dynamic fields are disabled. I can't recall what the default behaviour of managedindexfactory is.

We tend to send a lot more to solr than what is actually needed by the schema. If we allow dynamic fields it'll just be a mess.

Issue #1090, opened 2013-07-09
Can't create empty dataset


mwainwright:
When creating a dataset, allow user on screen 2 to continue without adding any resources. But give a pop-up warning:

"No data added. Create dataset with no data?
- Create
- Add data"

PR #4544, opened 2018-11-02
Allow dataset creation without adding any resources

howff:
# Fixes

Allows the user to create a dataset without adding a resource

### Proposed fixes:

See the following:
 https://github.com/ckan/ckan/issues/1090
 https://stackoverflow.com/questions/39496757/can-i-add-a-dataset-in-ckan-without-specifying-the-actual-data-link-or-file
 https://github.com/ckan/ckan/compare/master...smartlane:add-dataset-without-adding-data
 https://github.com/ckan/ckan/pull/2327
 https://github.com/ckan/ckan/pull/2316

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -667,9 +667,10 @@ class PackageController(base.BaseController):
                 except NotFound:
                     abort(404, _('The dataset {id} could not be found.'
                                  ).format(id=id))
-                if asbool(config.get('ckan.dataset.create_on_ui_requires_resources', 'True')) \
+                if asbool(config.get(
+                    'ckan.dataset.create_on_ui_requires_resources', 'True')) \
                     and not len(data_dict['resources']):
-                    # no data and configured to require resource so keep on page
+                    # no data and configured to require resource: stay on page
                     msg = _('You must add at least one data resource')
                     # On new templates do not use flash message
 
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -667,8 +667,8 @@ class PackageController(base.BaseController):
                 except NotFound:
                     abort(404, _('The dataset {id} could not be found.'
                                  ).format(id=id))
-                if not len(data_dict['resources']):
-                    # no data so keep on page
+                if asbool(config.get('ckan.dataset.create.require.resource', 'True')) and not len(data_dict['resources']):
+                    # no data and configured to require resource so keep on page
                     msg = _('You must add at least one data resource')
                     # On new templates do not use flash message
 
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -667,7 +667,8 @@ class PackageController(base.BaseController):
                 except NotFound:
                     abort(404, _('The dataset {id} could not be found.'
                                  ).format(id=id))
-                if asbool(config.get('ckan.dataset.create.require.resource', 'True')) and not len(data_dict['resources']):
+                if asbool(config.get('ckan.dataset.create_on_ui_requires_resources', 'True')) \
+                    and not len(data_dict['resources']):
                     # no data and configured to require resource so keep on page
                     msg = _('You must add at least one data resource')
                     # On new templates do not use flash message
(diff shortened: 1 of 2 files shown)


ckan/controllers/package.py
```diff
@@ -667,8 +667,8 @@ def new_resource(self, id, data=None, errors=None, error_summary=None):
                 except NotFound:
                     abort(404, _('The dataset {id} could not be found.'
                                  ).format(id=id))
-                if not len(data_dict['resources']):
-                    # no data so keep on page
+                if asbool(config.get('ckan.dataset.create.require.resource', 'True')) and not len(data_dict['resources']):
```
amercader: Can you rename the option to something like `ckan.dataset.create_on_ui_requires_resources`?

doc/maintaining/configuration.rst
```diff
@@ -1859,6 +1859,20 @@ example.
...
 -------------
 
+.. ckan.dataset.create.require.resource
+
+ckan.dataset.create.require.resource
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Example::
+
+ckan.dataset.create.require.resource = False
```
amercader: You need an indentation here to avoid the Sphinx warning error

ckan/controllers/package.py
amercader: This line should be shorter or the PEP8 test will complain.

howff: I have made those two changes

doc/maintaining/configuration.rst
howff: I have made that change

ckan/controllers/package.py
boykoc: @howff these are awesome changes, I'm looking forward to them in the next patch 👍 

I noticed the pep8 tests are still failing after your changes to `package.py` at lines 670, 671 and 672. 

```
> ---------------------------
> ckan/controllers/package.py
> ---------------------------
> E501 ln:670 line too long (95 > 79 characters)
> E125 ln:671 continuation line with same indent as next logical line
> E501 ln:672 line too long (80 > 79 characters)
> 
> Failing Files:
> ==============
> ckan/controllers/package.py
```

This should (I hope) fix it (I did a quick test of the whole file before and after at http://pep8online.com/, there's still some warnings but no more failures):

```
                if asbool(config.get(
                        'ckan.dataset.create_on_ui_requires_resources',
                        'True')) \
                        and not len(data_dict['resources']):
                    # no data and configured to require resource so keep on 
                    # page
```

howff: Can't you turn that error E501 into a warning? I am one of those people who edits code in a terminal window so this error message should be to my benefit. However it's actually creating code which is less readable. And at some point it won't even be possible to fit into 79 chars (a long config option with several levels of indenting) so what happens then? I am also surprised that a one-line addition is creating so much round-tripping and extra work for so many people; I naively imagined that making a contribution would be much easier than this ;-)

amercader:
Thanks @howff. I made a change in e09df0b to fix the PEP8 issue. I know this is annoying and looking at an individual change like this one it might seem frustrating to enforce these policies, but the end result is arguably more readable than the original code, and when you add all contributions overall it does make a difference to be consistent.

I hope that hasn't put you off of contributing more changes :) In the future I recommend running the pep8 tests locally to quickly check there are no errors:

     nosetests --nologcapture --with-pylons=test-core.ini -v ckan/tests/legacy/test_coding_standards.py

Or even better use a formatter like yapf (https://github.com/ckan/ckan/pull/4567) that will do the changes for you.

boykoc:
@howff I hear where you're coming from. I'm no expert but I'd say that when code is nesting this deep it's a prime candidate for refactoring into smaller pieces. But somethings are higher on the todo list than others.

I'm sure the core team would be happy for more PRs :).

Some things I try to consider (especially in an ideal state) are:

- methods / functions with more than 2 levels of nesting may be too long and could be refactored to be shorter separate methods/functions
- methods / functions with more than 2 levels of abstraction may be too long / trying to do too much

If these types of rules are applied it will help remove the issue of running out of space in a line.

boykoc:
@amercader did this get dropped in 2.9 or is it handled differently?  I don't see in the code where this is addressed now but the config is still referenced in the docs it seems.

PR #4556, opened 2018-11-18
View filter controls

wardi:
This PR includes 3 different changes:

1. Data dictionary "Show Controls" toggle buttons allow showing/hiding fields in the data dictionary form so that the form doesn't become too large and hard to use with more and more fields being added per column.

screenshot from 2019-01-04 15-01-22

2. New "search" view Filter Type that allows free-text searches on the column instead of an autocomplete field when users filter data on the front end. This should be the default for any columns with data not limited to a few hundred values or less. I haven't thought of an automatic way to detect that, so the default for all fields is still autocomplete.

3. datastore_search now accepts a blank key to its "q" parameter for searching across all text. q currently accepts either a string (search all text) or a dictionary (search text in given fields) but no way to combine the two types of search in the same query.

    e.g. combine a search across all text with one against only the street column:
`/api/action/datastore_search?q={"":"search this is across all text","street":"anystreet"}&resource_id=...`


wardi:
@amercader thoughts on this one? I'd like to build another feature on top of it.

amercader:
@wardi Will review it next wee for sure, I don't remember the details

wardi:
@amercader added one small change that makes the data dictionary for extensible from (multiple) plugins

wardi:
I could dust this off for inclusion in 2.9 WDYT?

amercader:
@wardi sounds good but I can't promise time for review so I'd consider it "only if there's time available"
Looks like it will need a Flask blueprint migration of the controller

pdelboca:
Closing as this is no longer relevant.

PR #4562, opened 2018-11-23
Add a hard upper limit to datastore_search(_sql) rows returned

### LINKED ISSUE #4561: Add a hard upper limit to datastore_search and datasearch_sql rows returned
For large datasets, we probably want to prevent users calling `datastore_search` or `datastore_seach_sql` and `limit` to 1000000 or more, because it chokes up postgres and ties up the server resources for the rest of the users. Users should page through instead. Or for these use cases they should download the data file itself and spend their own compute on processing it.

This follows on from the limits on core ckan action functions https://github.com/ckan/ckan/pull/4484

davidread:
Fixes #4561

### Proposed fixes:
`ckan.datastore.search.rows_max` is the new config option which limits `datastore_search` and `datastore_search_sql`.

When `datastore_search_sql` returns results that have hit this configured limit, then it includes `records_truncated: True` in the response.

It was too complicated to do the same for `datastore_search`, however you can tell when the limit you requested is above the `ckan.datastore.search.rows_max` limit, because the response includes a `limit` value which is changed from what you specified in the request to `ckan.datastore.search.rows_max`.

This PR includes some work to ensure 'datastore dump' still works (and remains not limited by `ckan.datastore.search.rows_max`). I've updated modernized the related tests, and that work is also found in a separate PR: https://github.com/ckan/ckan/pull/4581 if you want to merge that separately (before this one).

### Features:

- [x] includes API changes


ckanext/datastore/backend/postgres.py
```diff
@@ -1529,7 +1536,12 @@ def search_sql(context, data_dict):
 
         results = context['connection'].execute(sql)
 
-        return format_results(context, results, data_dict)
+        if results.rowcount == rows_max + 1:
+            data_dict['records_truncated'] = True
+        else:
+            data_dict['records_truncated'] = False
```
wardi: `data_dict['records_truncated'] = results.rowcount == rows_max + 1`

wardi:
ckan.datastore.search.rows_max is likely going to cause a problem with https://github.com/ckan/ckan/blob/master/ckanext/datastore/controller.py#L46

ckanext/datastore/backend/postgres.py
wardi: or better yet, not add anything to the dict in the common case
```python
        if results.rowcount == rows_max + 1:
            data_dict['records_truncated'] = True
```

davidread: why not

davidread:
I guess the datastore dump should be exempt from the rows_max. It's not flexible with lots of options like datastore_search, so is very cacheable. And it's just a bit having an API to get a straight dump of a file and it turning up truncated.

If so, dump_to could pass a context variable that tells datastore_search not to impose the rows_max.

wardi:
-1 on special behaviours based on context variables, just need to update dump to check for the results_truncated return value. If we try to be too clever we won'd be able to support custom validation rules because we end up repeating them.

Also for choosing a default datastore_search limit set it at least as large as the dump PAGINATE_BY value.

davidread:
Ok makes sense.

> choosing a default datastore_search limit set it at least as large as the dump PAGINATE_BY value.

Rather than disallow rows_max to be less than 32000, I've done some code that allows it to be lower, and in this case it simply reduces the PAGINATE_BY value, so no rows are missed by the dumper. `test_dump_with_low_rows_max` is the test (in test_dump.py).

Ready for rereview @wardi

ckanext/datastore/controller.py
```diff
@@ -43,7 +44,8 @@
 
 
 DUMP_FORMATS = 'csv', 'tsv', 'json', 'xml'
-PAGINATE_BY = 32000
+PAGINATE_BY = min(32000,
+                  int(config.get('ckan.datastore.search.rows_max', 32000)))
```
wardi: no, leave this as a constant and instead of checking the number returned check the records_truncated value returned from datastore search in the loop below

wardi: ^ should have said "in addition to checking the number of rows returned also check the records_truncated value"

wardi: also, trying to read a configuration value at module load time won't work when we need to reload the configuration file.

davidread: So in the loop we check the records_truncated value and break?

I'll have to add records_truncated to datastore_search too then, I've only got it in datastore_search_sql so far.

wardi: right I keep getting those confused

davidread:
@wardi Adding a 'records_truncated' to `datastore_search` seems too complicated. As discussed for this PR, for `datastore_search_sql` we return `records_truncated` to indicate if you get back less records than you would have done without rows_max. This is done by querying for rows_max+1 records and discarding the last one. However in `datastore_search` you can't remove the last record, because the result is returned from postgres as a CSV/TSV/JSON/XML string. So you *have to* query for exactly rows_max records, and if you want to know if you are truncating you need to do a second query for rows_max+1 and count the results. Yuk.

So I think the best we could do for datastore_search is say if the limit specified by the user has been lowered to rows_max. This is slightly useful, because the user doesn't know what rows_max is configured to be. I've not coded this yet, but it's simple.

For consistency the 'Datastore dump' call should also warn the user if the records are curtailed. Because it uses datastore_search it too cannot say if you are over the rows_max, without that annoying extra query. But we can tell the user if we have returned rows up to the rows_max, even with the pagination going on - it's complicated but I've managed to implement it with the help of a bunch of tests. I'm not sure it is worth it - let me know what you think.

wardi:
If we're going to limit the number of records that the dump controller returns that should be a separate option that defaults to unlimited.

The whole point of the dump controller is to dump all the data requested so that users don't need to paginate with the API themselves. The controller uses a constant amount of memory, it just takes more time if there are more records.

davidread:
Ok, I wasn't quite sure which way you were nudging, but that's clear now and helps!

davidread:
Backport for 2.7 is on branch 4561-limit-datastore_search-2.7-backport

davidread:
This is now ready for further comments @tino097 @smotornyuk

davidread:
@tino097 @smotornyuk Just a friendly ping to remind you about this PR :)

ckanext/datastore/backend/postgres.py
```diff
@@ -1392,6 +1392,8 @@ def format_results(context, results, data_dict):
                                                  field['type'])
         records.append(converted_row)
     data_dict['records'] = records
+    if data_dict.get('records_truncated', False):
+        data_dict['records'] = data_dict['records'][:rows_max]
```
smotornyuk: I'm wondering, won't there be any performance penalty for big lists with ~32000 records? I tested it locally with mock data and it doesn't seem to be a problem, but if you have a real dataset, it would be great to confirm it.  Otherwise, just popping the last record will do the same job, as I understand? and, due to lists nature, it won't consume additional memory/processor time

wardi: 32000 comes from https://github.com/ckan/ckan/pull/3523#issuecomment-296304558 measured against a real dataset

davidread: Nice thinking @smotornyuk! I tried it and `.pop()` is twice the speed of `[:rows_max]`
```
>>> import timeit
>>> t1 = timeit.Timer('a=500000*[\'a\'];a.pop()')
>>> t2 = timeit.Timer('b=500000*[\'b\'];b[:-1]')
>>> print(t1.timeit(1000)/1000, t2.timeit(1000)/1000)
(0.0023797738552093508, 0.004582674026489258)
```
I'll make the change

davidread:
Any more comments or is this good to merge @wardi @smotornyuk ?

ckanext/datastore/logic/schema.py
```diff
@@ -167,7 +169,9 @@ def datastore_search_schema():
         'plain': [ignore_missing, boolean_validator],
         'filters': [ignore_missing, json_validator],
         'language': [ignore_missing, text_type],
-        'limit': [ignore_missing, int_validator],
+        'limit': [default(100), natural_number_validator,
```
wardi: This is an API change that might break assumptions in external scripts, we should call it out in the docs and make it configurable too.

doc/maintaining/datastore.rst
```diff
@@ -249,6 +249,9 @@ JSON (``?format=json``) and XML (``?format=xml``). E.g. to download an Excel-com
 tab-separated file use
 ``{CKAN-URL}/datastore/dump/{RESOURCE-ID}?format=tsv&bom=true``.
 
+A number of parameters from :meth:`~ckanext.datastore.logic.action.datastore_search` can be used:
+    ``offset``, ``limit``, ``filters``, ``q``, ``distinct``, ``plain``, ``language``, ``fields``, ``sort``
+
```
wardi: nice, thanks

ckanext/datastore/controller.py
davidread: This discussion continued in the main PR comments below ⬇️ - Dec 7

The upshot of that was that changing the interface to the dump controller was reverted. Instead, we just get dump controller's calls to datastore_search to be sensitive to the situation when rows_max < PAGINATE_BY and reduce the page size to match. So this concern is resolved.

ckanext/datastore/logic/schema.py
davidread: Resolved in https://github.com/ckan/ckan/pull/4562/commits/51337ff02aaf1ac2f216563a4c0c5b1a27c5700a

ckanext/datastore/logic/action.py
```diff
@@ -420,7 +422,10 @@ def datastore_search(context, data_dict):
     :param language: language of the full text query
                      (optional, default: english)
     :type language: string
-    :param limit: maximum number of rows to return (optional, default: 100)
+    :param limit: maximum number of rows to return
+        (optional, default: ``100``, unless set in the site's configuration
+        ``ckan.datastore.search.rows_default``, upper limit: ``32000`` unless
+        set in site's configuration ``ckan.datastore.search.rows_max``)
```
wardi: Should have raised this in my last comment, but If we're adding the hard upper limit why do we need a default limit? AFAICT this just breaks external users scripts for no real benefit.

We've got a pretty big overhead serving `datastore_search` calls and if a user sees that they're only getting 100 records they might start paginating themselves 100 records at a time, creating *way* more load on the server. Is there a reason you don't want to just rely on the new rows_max setting?

davidread: The default of 100 has been there since 2012 (https://github.com/ckan/ckan/commit/3dec0e8#diff-5816a6ed066765bf38ec73a848d96e3aR150), so I don't think we're breaking anything, are we? I think if someone is going to seek out the 'offset' parameter to achieve paging, then they'll likely see the 'limit' too and increase that value to make their life easier. So arguably on balance, removing the default might increase load overall. However I can also see that maybe it's an unnecessary complication having both default and max.

wardi: you're right! nevermind then

wardi:
this looks good to me

davidread:
Thanks @wardi.

@tino097 @smotornyuk did either of you want to look at this any more before I merge?

PR #4563, opened 2018-11-23
Add doc language around patching resources.

boykoc:
Fixes #

### Proposed fixes:

Some of the functionality with `package_patch` was unclear to me. I thought others may experience the same at some point so I've updated the documentation.

- **No default views for new resources:** As per PR #3061 update methods are not supposed to create the default views for resources, even if the update method creates a resource, but this was not referenced in the docs.
- **Handling resources:** I've added language around expected behaviour of `package_patch` when updating, creating or deleting resources. Although it now makes sense, I originally expected a resource name to be used to verify if the resource exists and update it before removing and creating a new one and to leave existing resources not referenced in the list of resource dicts.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/patch.py
+++ b/ckan/logic/action/patch.py
@@ -19,14 +19,14 @@ def package_patch(context, data_dict):
     The difference between the update and patch methods is that the patch will
     perform an update of the provided parameters, while leaving all other
     parameters unchanged, whereas the update methods deletes all parameters
-    not explicitly provided in the data_dict. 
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @boykoc. The resource views issue was discussed in https://github.com/ckan/ckan/issues/3011 and I think there haven't been changes on that front so there's no easy solution for now. It's good to have it clarified in the docs though.

PR #4574, opened 2018-12-06
c.user not defined fix

### LINKED ISSUE #4247: '_Globals' has no attribute 'user' : exception when using an IAuthenticator on CKAN 2.8.0
I'm putting together a new deployment based on the new CKAN v2.8.0 release. I'm using ckanext-ldap as an authenticator, though it looks like this bug would apply to any authenticator plugin.
This exact setup worked fine on CKAN v2.7.3.

CKAN v 2.8.0
ckanext-ldap @ `ckan-upgrade-2.8.0a`

If the IAuthenticator plugin cannot authenticate the user, it does not set `g.user`, and CKAN should run the default authenticator.

If the IAuthenticator plugin cannot authenticate the user, it does not set `g.user`, and CKAN tries to lookup `g.user` and crashes with traceback:
```
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/flask/app.py", line 1610, in full_dispatch_request
    rv = self.preprocess_request()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/flask/app.py", line 1831, in preprocess_request
    rv = func()
  File "/usr/lib/ckan/venv/src/ckan/ckan/config/middleware/flask_app.py", line 281, in ckan_before_request
    identify_user()
  File "/usr/lib/ckan/venv/src/ckan/ckan/views/__init__.py", line 101, in identify_user
    if g.user:
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/usr/lib/ckan/venv/src/ckan/ckan/config/middleware/flask_app.py", line 334, in __getattr__
    return getattr(app_globals.app_globals, name)
AttributeError: '_Globals' object has no attribute 'user'
```
* Install CKAN v2.8.0 as per documented instructions
* Install a plugin that implements IAuthenticator (In this case I am using the ckanext-ldap plugin in the 2.8.0 branch), that may not be able to authenticate the user, so may not set `g.user`.
* Run CKAN normally
* Attempt to load any page. 

What is odd is that this section of code at `identify_user` in `ckan/views/__init__.py` has not changed between v2.7.3 and v2.8.0. And the way the authenticator plugin handles/sets `g.user` has not changed either. I'm guessing this is caused by a change in the way the _Globals object behaves when it cannot find an attribute.

frafra:
Fixes #4247 
This issue should have been ready fixed, but it seems that the patch did not covered all the cases.
This is not suppose to be the perfect patch for this issue, it seems a workaround for me, but this seems enough for my usage of CKAN.

More about it on: https://github.com/ckan/ckan/issues/4247#issuecomment-443727710

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -48,7 +48,7 @@ import ckan.authz as authz
 import ckan.plugins as p
 import ckan
 
-from ckan.common import _, ungettext, c, request, session, json
+from ckan.common import _, ungettext, c, g, request, session, json
 from markupsafe import Markup, escape
 
 
@@ -1154,10 +1154,10 @@ def sorted_extras(package_extras, auto_clean=False, subs=None, exclude=None):
 
 @core_helper
 def check_access(action, data_dict=None):
-    if not getattr(c, u'user', None):
-        c.user = None
+    if not getattr(g, u'user', None):
+        g.user = ''
     context = {'model': model,
-               'user': c.user}
+               'user': g.user}
     if not data_dict:
         data_dict = {}
     try:
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1154,6 +1154,8 @@ def sorted_extras(package_extras, auto_clean=False, subs=None, exclude=None):
 
 @core_helper
 def check_access(action, data_dict=None):
+    if not getattr(c, u'user', None):
+        c.user = None
     context = {'model': model,
                'user': c.user}
     if not data_dict:


smotornyuk:
Just two small alterations:
* Previously, pylons used to return empty string(`''`) for any undefined value, so let's use the same, instead of `None`
* We are replacing `c` global variable with `g` - they are pretty same, but latter is more Flask style, so could you switch to it as well?

frafra:
Hi @smotornyuk, thank you for your suggestions.
I can use an empty string instead of None, but switching from c to g would not fix the issue. I tried to use `g.user` instead of `c.user` without any additional code, but it fails. I tried to setup `g.user` and then use `c.user`, but it still fails. The only way I found to avoid this issue is just to set `c.user`.

smotornyuk:
No, I'm perfectly ok with your solution and switching to `g` didn't suppose to fix the problem, because `g` and `c` are the same, as I said before. You just should use `g` instead of `c` in the feature, because `c` - it's how Pylons calls global context, and `g` - is a way of naming it from Flask. As we are migrating to Flask, I just want to use flask-style names for variable

frafra:
Thank you for the clarification. I did not understand that they were equivalent and I was confused by the fact that using `g.user` is not enough, even if the previous patch set it.

I had to import `g`, as `helpers.py` uses `c` only.

PR #4578, opened 2018-12-06
Failing autocomplete tests on master

smotornyuk:
Apparently, previously(circleV1) we were using SOLRv4 and now(circleV2) we use SOLRv3 instead. They have little bit different implementation of NGram tokenizer(that one, I'm using for autocomplete) and, in addition, newer SOLR is more adaptive in terms of search parameters. Actually, I was passing `fl` as a list and SOLRv4 was able to handle it, whereas SOLRv3 strictly requires string.. anyway, it seems that I've fixed the problem

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1536,7 +1536,7 @@ def package_autocomplete(context, data_dict):
             'name:{0}',
             'title:{0}',
         ]).format(search.query.solr_literal(q)),
-        'fl': ['name', 'title'],
+        'fl': 'name,title',
         'rows': limit
     }
     query = search.query_for(model.Package)
--- a/ckan/tests/legacy/logic/test_action.py
+++ b/ckan/tests/legacy/logic/test_action.py
@@ -130,7 +130,7 @@ class TestAction(WsgiAppCase):
         assert_equal(res_obj['result'][0]['match_displayed'], 'warandpeace')
 
     def test_02_package_autocomplete_match_title(self):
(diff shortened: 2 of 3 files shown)


ckan/logic/action/get.py
```diff
@@ -1536,7 +1536,7 @@ def package_autocomplete(context, data_dict):
             'name:{0}',
             'title:{0}',
         ]).format(search.query.solr_literal(q)),
-        'fl': ['name', 'title'],
```
wardi: fl doesn't accept lists anymore?

ckan/tests/legacy/logic/test_action.py
```diff
@@ -130,7 +130,7 @@ def test_02_package_autocomplete_match_name(self):
         assert_equal(res_obj['result'][0]['match_displayed'], 'warandpeace')
 
     def test_02_package_autocomplete_match_title(self):
-        postparams = '%s=1' % json.dumps({'q':'a%20w', 'limit': 5})
```
wardi: what's the significance of this change?

smotornyuk: This is how an implementation of NGramTokenized differs in SOLR 3 and 4. newer SOLR was able to found record 'A Wonderful Story' with such query, but v3 can find it only if we are providing at least few symbols from a single word. So I've updated this test and now it's searching by part of the word, rather than by part of the whole phrase

ckan/logic/action/get.py
smotornyuk: It seems, that SOLRv3 doesn't accept lists, but all newer versions are ok with them

Issue #3327, opened 2016-11-22
Local set to 'fr' doesn't work correctly when visualising a dataset


stephane-lapointe:
### CKAN Version if known (or site URL)
CKAN version 2.6.0

### Please describe the expected behaviour
Dataset should adjust datetime when locale is set to 'fr'.

### Please describe the actual behaviour
An error is generated while accessing a dataset when local is set to fr. (http://my-domain/fr/dataset/dataset-name)

Logs show this message:

File '/usr/lib/ckan/default/src/ckan/ckan/lib/formatters.py', line 152 in localised_nice_date
.format(**details))
KeyError: u'heure'

The problem doesn't exist with locale different than 'fr'.

The problem can be fixed temporarily by changing the value in  ckan/i18n/fr/LC_MESSAGES/ckan.po. 

#: ckan/lib/formatters.py:148
msgid "{month} {day}, {year}, {hour:02}:{min:02} ({timezone})"
msgstr "{month} {day}, {year}, {**heure**:02}:{min:02} ({timezone})"

#: ckan/lib/formatters.py:148
msgid "{month} {day}, {year}, {hour:02}:{min:02} ({timezone})"
msgstr "{month} {day}, {year}, {**hour**:02}:{min:02} ({timezone})"

Note that the catalog is marked as fuzzy, so you'll have to force catalog compiling:
python setup.py compile_catalog -f --locale fr

amercader:
Thanks @stephane-lapointe 

Fixed on Transifex and backported to the dev-v2.6 branch, will be part of 2.6.1

wardi:
@amercader did you intend to merge this instead?

Issue #4545, opened 2018-11-05
Translation string issue


EricSoroos:
https://github.com/ckan/ckan/blob/master/ckan/i18n/nl/LC_MESSAGES/ckan.po#L1556

It appears that the translator has translated the template variables in this string, rather than arrange them in the proper order preserving the template variable names. 

### CKAN Version if known (or site URL)
2.7.2, master

### Please describe the expected behaviour
Date formatting works

### Please describe the actual behaviour
Date formatting returns a 500 error

File '/usr/lib/ckan/default/src/ckan/ckan/lib/base.py', line 78 in render_snippet
  renderer='snippet')
File '/usr/lib/ckan/default/src/ckan/ckan/lib/base.py', line 177 in render
  return cached_template(template_name, render_template)
File '/usr/lib/ckan/default/lib/python2.7/site-packages/pylons/templating.py', line 249 in cached_template
  return render_func()
File '/usr/lib/ckan/default/src/ckan/ckan/lib/base.py', line 131 in render_template
  return render_jinja2(template_name, globs)
File '/usr/lib/ckan/default/src/ckan/ckan/lib/base.py', line 88 in render_jinja2
  return template.render(**extra_vars)
File '/usr/lib/ckan/default/lib/python2.7/site-packages/jinja2/environment.py', line 1008 in render
  return self.environment.handle_exception(exc_info, True)
File '/usr/lib/ckan/default/lib/python2.7/site-packages/jinja2/environment.py', line 780 in handle_exception
  reraise(exc_type, exc_value, tb)
File '/usr/lib/ckan/default/src/ckan/ckan/templates/snippets/activity_item.html', line 8 in top-level template code
  <span class="date" title="{{ h.render_datetime(activity.timestamp, with_hours=True) }}">{{ h.time_ago_from_timestamp(activity.timestamp) }}</span>
File '/usr/lib/ckan/default/src/ckan/ckan/lib/helpers.py', line 1222 in render_datetime
  with_hours=with_hours)
File '/usr/lib/ckan/default/src/ckan/ckan/lib/formatters.py', line 147 in localised_nice_date
  .format(**details))
KeyError: u'maand'

### What steps can be taken to reproduce the issue? 
Visit an activity stream with CKAN_LANG set to 'nl': /dataset/activity/[[name-or-id]]

smotornyuk:
Yep, this issue was reported in #4508

Issue #4508, opened 2018-10-17
Invalid translation strings - causing server error


OriHoch:
Seems there are some invalid translation strings which cause a server error when viewing the page with the bad translation string

I ran a script to find them, here are the results (on master):

```
ckan/i18n/ru/LC_MESSAGES/ckan.po:211
msgstr "Поле {номер}."

ckan/i18n/ru/LC_MESSAGES/ckan.po:1574
msgstr "{месяц} {день}, {год}, {часов:02}:{минут:02} ({timezone})"

ckan/i18n/ru/LC_MESSAGES/ckan.po:1578
msgstr "{месяц} {день}, {год}"

ckan/i18n/ar/LC_MESSAGES/ckan.po:208
msgstr "حقل {رقم} "

ckan/i18n/mn_MN/LC_MESSAGES/ckan.po:1052
msgstr "буруу хайлтын бичлэг: {алдааны_мэдээлэл}"
```

This is the script:

```
from glob import glob
import re

for po in glob('ckan/i18n/*/LC_MESSAGES/ckan.po'):
    with open(po) as f:
        for i, line in enumerate(f.readlines()):
            m = re.findall(r'{[^a-z0-9]', line)
            if m:
                print(po + ':' + str(i))
                print(line)
```

wardi:
Let's add this to our tests. Instead of checking for non-ascii-numeric characters we should check that all the variables present in the original string are present in the translated strings with https://docs.python.org/2/library/string.html#string.Formatter.parse

PR #4580, opened 2018-12-07
Fix fatal errors due to broken Russian and Dutch locales (related to #3327)

tuffnatty:
Fixes 500 Internal server error on dataset view pages with ru or nl locales.

### Proposed fixes:
Maybe this should be done on Transifex? And backported to supported branches?

### Features:

- [ ] includes API changes


tuffnatty:
This also applies to #4545 and #4508 which I have only found today. However, #4508 reporter also found same kind of error in ar and mn_MN locales.

amercader:
@tuffnatty I've fixed yours on Transifex, which is the canonical source for our po files (https://www.transifex.com/okfn/ckan/) and will be included in today's patch releases. The rest of issues we'll sort of a wider change including tests to avoid these issues.

PR #4583, opened 2018-12-09
Return message from Postgres when there's a validation during datastore import

amercader:
Right now when there's a validation error during datastore_create you just get a generic message with no actual details (eg wrong type or wrong values). At least let's return Postgres message which gives the column and/or the wrong value.

eg in my case it went from this:

```
The data was invalid (for example: a numeric value is out of range or was inserted into a text field)

```

To this:

```
The data was invalid: invalid input syntax for type numeric: "SAN29"

```

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1054,11 +1054,10 @@ def upsert_data(context, data_dict):
 
         try:
             context['connection'].execute(sql_string, rows)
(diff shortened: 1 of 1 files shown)


wardi:
I'm working on some code adjacent to this so I'll merge now: I'm looking at a way to return a list of errors from datastore_create/upsert instead of just the first for validation failures written as triggers. So in ckan core at the API level I'll just be standardizing a way to return the column names along with the errors in each.

PR #4599, opened 2018-12-20
Fix delete button links

### LINKED ISSUE #4598: Removing user from group or organization raises 404
This occurs on _ckan 2.8.2_ but might be older

You should be able to remove a user from a _group_ or an _organization_.

I get a 404 and the user is not removed from _group_ or _organization_.

You need admin rights on a group or organization.
Get to the user list of the _group_ or _organization_.
Click on the wrench of the user you want to remove.
You should be redirected to the "Edit Member" page.
Click on the *delete* button.
You get a 404.
 
Note that the delete button on the user list of the _group_ or _organization_ does work. Only the one in the edit member page does not.

gt-novelt:
Fixes #4598

### Proposed fixes:

use `user.id` rather than `user_id` in template form.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates-bs2/group/member_new.html
+++ b/ckan/templates-bs2/group/member_new.html
@@ -55,7 +55,7 @@
     {{ form.select('role', label=_('Role'), options=roles, selected=user_role, error='', attrs=format_attrs) }}
     <div class="form-actions">
       {% if user %}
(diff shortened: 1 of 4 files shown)


tino097:
@gt-novelt before i merge this, i believe that you should add same changes to `templates-bs2` forms

p.s sorry for late review

gt-novelt:
@tino097 I did as requested.
Btw what are those `templates-bs2` for ?

tino097:
@gt-novelt those are templates with bootstrap2, we introduced `public-bs2` and `templates-bs2` during the migration to[ bootstrap3 ](https://github.com/ckan/ckan/pull/3547)

PR #4600, opened 2018-12-21
h.render_datetime renders datestamps now with seconds

davidread:
This is extracted from https://github.com/ckan/ckan/pull/3972

When a user drills into activity stream, each version is known by its timestamp, and
on occasions that a user makes multiple changes in the same minute, showing the
hours:minutes:seconds is really useful.

The impending #3972 has a replacement bit of HTML which calls h.render_datetime with the new
'with_seconds' option yet.

### Features:

- [ ] includes API changes


wardi:
don't see anyone objecting to this

PR #4605, opened 2018-12-24
Updating get.py

### LINKED ISSUE #4420: Dataset counts incorrect on Groups listing
After upgrading to CKAN 2.8.1, the dataset counts are incorrect on the group listing screen.

On the group listing screen (/group), the dataset count for each group is the full dataset count for the whole site.  The correct count is shown on the screen for each individual group (/group/somegroup)

CKAN 2.8.1
/group
/group/somegroup

iamarnavgarg:
Fixes #4420 
The number of datasets present in the group present is displayed properly.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1840,7 +1840,7 @@ def package_search(context, data_dict):
             data_dict['fl'] = ' '.join(result_fl)
 
         # Remove before these hit solr FIXME: whitelist instead
-        include_private = asbool(data_dict.pop('include_private', False))
+        include_private = asbool(data_dict.pop('include_private', True))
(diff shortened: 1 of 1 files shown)


iamarnavgarg:
Hi @amercader can you please guide me in these unsuccessful checks?

wardi:
I support changing the `include_private=False` default on `package_search` to `True`, but that's a larger change than what's required to fix #4420.

If you'd like to make that larger change I expect a number of tests will have to be updated to pass `include_private=False` because that's what they're expecting.

If instead you'd just like to fix the original issue then you can pass `include_private=True` from the controller that's giving the wrong dataset counts.

Finally, we accept PRs against the master branch not the release branches unless a change only applies to that release. Please submit a new PR against master. Thanks!

PR #4614, opened 2019-01-09
Webassets

smotornyuk:
Replace [fanstatic](http://www.fanstatic.org/en/latest/) with [webassets](https://webassets.readthedocs.io/en/latest/)

### Why?

- Webassets seems to be more popular than fanstatic
- webassets compiles to the real file and served as any other static file

### What should I know?

- bundles - it's like `groups` in fanstatic, predefined collection of static files, that has some name(like `ckan/main`) 

### What should I do?

- use `{% asset <bundle_name> %}` instead of `{% resource <resource_name> %}`
- stop including static files directly (e.g, `resource 'xxx/file.css'`) - define bundles itself
- instead of `resource.config` create `webassets.yml` - now it's required. This file MUST be created in every fanstatic directory.

### How it works?

During server startup all assets(included via `add_resource`) are compiled into `ckan/public/webassets` folder. After that, compiled files served directly, just like images, by new Flask route, defined during creation of Flask stack.
In addition, one can perform base tasks using CLI:

```
# build static files - in that case, they won't be recompiled during server's startup(if the source hasn't been changed, of course)
ckan -c config.ini asset build 

# remove compiled files - this will trigger recompilation during server's reload, even if the source hasn't been changed.
ckan -c config.ini asset clean

# watch bundles' source files and rebuild on change - i don't think that this command will be ever used and want to remove it, but let's see if anyone finds it useful
ckan -c config.ini asset watch
```

### webassets.yml

This file has to be created in the root of every "fanstatic" directory. It serves the same purpose as `resource.config` so I'll describe only it's content

- It uses [YAML](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html) syntax
- generally, it has a structure like below:
```yml
bundle_name:
    filters: rjsmin
    output: folder/%(version)s_bundle_name.js
    extra:
        preload:
            - vendor/vendor
    contents:
        - x.js
        - y.js
```
- As mentioned before, `bundles` are similar to fanstatic's `groups`. The main difference - bundle may contain either CSS files or js, but not both types at the same time. So, you'll need to define at least two bundles if your extension contains custom styles and custom scripts

Every bundle has a name consisting of two parts `EXTENSION_NAME/BUNDLE_NAME`(eg., base/main, ckanext-datatablesview/main). Inside webassets.yml one should define only BUNDLE_NAME, EXTENSION_NAME  will be added to definition automatically

- bundle_name - unique name in the scope of current webassets.yml. In different YML files, the same bundle name can be used without any restrictions.
- filters - compilation preprocessors. Generally, use `rjsmin` for js and omit section entirely for css. Details in webassets documentation
- output - file inside `ckan/public/webassets` where the compiled file should be stored. in order to avoid name conflicts, try to put extension's output into a folder named as extension itself. Also, you may notice `%(version)s` fragment in my example - this is the wildcard that will be replaced with file's hash during compilation. Thus, if the source was changed, file with a completely different name will be produced and you won't face any cache issues. In addition, you may set all possible cache headers on static responses, because there is no possible case when a file with the given name has its content updated. (Details here)[https://webassets.readthedocs.io/en/latest/expiring.html]
- extra - any additional information. I'm using extra in format `{preload:  [list of dependencies]}` in order to reproduce fanstatic's `[depends]` section. This is a completely custom solution, so don't try to find details in the documentation. Existing dependency resolving won't work in our case because we are using dynamically added extensions and multiple bundle configurations(webassets.yml). 
- contents - list of components for this bundle. DO NOT MIX css and js files into a single bundle!

Yep, that's all. I don't like to write documentation :sob:


tino097:
ive tested this locally, looks great

Good job @smotornyuk

ckan/config/middleware/flask_app.py
```diff
@@ -181,6 +182,16 @@ def hello_world():
     def hello_world_post():
         return 'Hello World, this was posted to Flask'
 
+    # WebAssets
+    public_folder = config.get('ckan.base_public_folder')
+    webassets_folder = os.path.join(
+        os.path.dirname(ckan.__file__), public_folder, 'webassets'
```
wardi: does this mean that webassets need to be stored under the ckan source directory? That feels wrong, especially for package installs.

ckan/lib/jinja_extensions.py
```diff
@@ -323,12 +324,33 @@ class ResourceExtension(BaseExtension):
 
     @classmethod
     def _call(cls, args, kwargs):
+        log.warn(
+            '`resource` tag is deprecated. '
+            'Use `assets`'
+            '<https://docs.ckan.org/en/latest/theming/webassets.html> instead')
```
wardi: maybe not worth spamming the logs with this one. I think marking deprecated in release notes is good enough

ckan/lib/webassets_tools.py
```diff
@@ -0,0 +1,139 @@
...
+    bundles = {
+        u'/'.join([name, key]): bundle
+        for key, bundle
+        in library.items()
+    }
+
+    # Unfortunately, you'll get an error attempting to register bundle
+    # with the same name twice. For now, let's just pop existing
+    # bundle and avoid name-conflicts
+    # TODO: make PR into webassets with preferable solution
```
wardi: can we link to an open issue on the webassets repo?

ckan/public/base/css/webassets.yml
```diff
@@ -0,0 +1,18 @@
...
+  contents: main.css
+main-rtl:
+  output: base/%(version)s_main-rtl.css
+  contents: main-rtl.css
+maroon:
+  output: base/%(version)s_maroon.css
+  contents: maroon.css
+red:
+  output: base/%(version)s_red.css
+  contents: red.css
```
wardi: do these themes still work on master?

ckan/public/base/javascript/resource.config
```diff
@@ -20,49 +19,4 @@ apply_html_class =
...
-    plugins/jquery.form-warning.js
-    plugins/jquery.images-loaded.js
-
-    sandbox.js
-    module.js
-    pubsub.js
-    client.js
-    notify.js
-    i18n.js
-    main.js
```
wardi: we need to keep these empty files for now?

ckan/templates/base.html
```diff
@@ -112,5 +117,10 @@
       {# defined in the config.ini under "ckan.template_footer_end" #}
       {{ g.template_footer_end | safe }}
     {%- endblock %}
+
+    {# render all assets included in scripts block and everywhere else #}
+    {# make sure there are no calls to `asset` tag after this point #}
+    {{ h.render_assets('style') }}
```
wardi: duplicated `h.render_assets('style')`?

ckan/templates/package/resource_edit_base.html
```diff
@@ -37,5 +37,6 @@ <h1 class="hide-heading">{% block form_title %}{{ _('Edit resource') }}{% endblo
 
 {% block scripts %}
   {{ super() }}
-  {% resource 'vendor/fileupload' %}
+  {#% resource 'vendor/fileupload' %#}
```
wardi: why leave the old tags as comments?

ckanext/example_theme_docs/v16_initialize_a_javascript_module/templates/snippets/package_item.html
```diff
@@ -4,15 +4,16 @@
   {{ super() }}
 
   {# Use Fanstatic to include our custom JavaScript module.
```
wardi: no longer Fanstatic

wardi:
Just a few comments above. Splendid work.

Do we need to document/enable support for http://flask.pocoo.org/docs/1.0/api/#flask.Flask.use_x_sendfile ? Need to make sure our web server is configured to support it. Even better (but less general) we can document serving the files straight from the web server.

ckan/config/middleware/flask_app.py
smotornyuk: Yes, I agree. I just chose the most simple way of serving files. But it may be actually served from any other place in the filesystem(like storage directory) as long as Flask has read access to it. We also may build assets before starting the server and serve them separately(not via flask), but it will add some configuration overhead. 
What you think about putting compiled assets into a storage folder? In that way, we avoid writing into CKAN filesystem, but storage directory wasn't required previously, so we need to change the installation process. Or, maybe there are other suggestions?

ckan/public/base/css/webassets.yml
smotornyuk: Yes

ckan/public/base/javascript/resource.config
smotornyuk: Yep. They are required for backward compatibility with extensions that depends on these files. It may look strange, why I removed all entries from them, but it's quite simple - if you leave them unchanged and some extension depends on them, two identical versions of javascript files will be included into page(one from web assets and one compiled as extension dependency) and this leads us to numerous jQuery conflicts and such kind of bugs.

ckan/templates/base.html
smotornyuk: No. Jinja parses/renders templates as a stream, so if some styles were included after `head` section, there is no way to go back and add styles into the header, so here is the place, where all "late" style goes. Fanstatic didn't include static files in realtime, instead, it allows to render the complete template and then rewrites it with regExps, but I try to get rid of these rewrites

ckan/templates/package/resource_edit_base.html
smotornyuk: I was sure that there will be some alterations before this PR merged into core, so I just left these lines as examples of migration from fanstatic to web assets. I'm going to remove them in next commit

smotornyuk:
Yep, I think we should describe all possibilities and migration details because otherwise there will be tons of questions in gitter

ckan/config/middleware/flask_app.py
smotornyuk: Finally, got some time for update. I'll use additional configuration variable `ckan.webassets.path` (with fallback to `ckan.storage_path` + `webassets` and RuntimeError if neither is specified) So, they can be placed near the resources and group images - it's reasonable enough, i suppose

ckan/lib/jinja_extensions.py
smotornyuk: Yes, probably so.

ckan/lib/webassets_tools.py
smotornyuk: Done

smotornyuk:
Updated. Fixed issues, mentioned by Ian. Add few config options:
```

## Webassets Settings
#ckan.webassets.use_x_sendfile = false
#ckan.webassets.path = /var/lib/ckan/webassets
```

`ckan.webassets.use_x_sendfile` - flask will add `X-Sendfile` header to webassets. 
`ckan.webassets.path` - desination for compiled webassets. If not provided, `webassets` folder inside `ckan.storage_path`  will be used. If `ckan.storage_path` not provided as well, use system temporary folder.


TODO(separate pull requests, good for contribution?)
1. Add new page to documentation.
2. Provide examples of configuration for X-Sendfiles option for different servers
3. Add ability to build links to external files, served from CDN/S3/etc.

smotornyuk:
@wardi @tino097 @amercader  It's ready for review

wardi:
Awesome. I think as long as our docs recommend setting up nginx in a way that supports X-SendFile the other documentation can come later.

Fix the last conflict and I'll merge :+1

amercader:
I apologize for not having properly reviewed this but can you summarize what are the implications in terms of backwards compatibility? ie will current extensions that use fanstatic and `{% resource %}` tags stop working on 2.9? Or both systems will work for a while?

And will extensions need to add `webassets.yml` files to their folders even if they are using fanstatic.

This feature is obviously superior but just want to understand what we need to communicate.

davidread:
@tino097 thinks fanstatic will be supported still - to check

smotornyuk:
@amercader , fanstatic is still supported - I've added commit with an example of webassets definition for existing fanstatic files - https://github.com/ckan/ckan/pull/4614/commits/d7ddb5308b5faa46d3933fc5688a00d9d556df7b

So, if you are using `resource` tags - it works as previously. If you want to add webassets support, there is no need in removing fanstatic's `resource.config` file, just add `webassets.yml` with bundle definitions. After this, anything required with `resource` will be loaded via fanstatic, anything required with `asset` will be loaded via webassets. As you may notice in the provided commit, you even can use both tags in the same file. 

The only serious issue - order of styles. Webassets will have higher priority because it loads styles at the end of `head` tag. Fanstatic, on the other hand, writes styles right after the opening of the `head` tag, so due to cascade nature of CSS, latter styles(webassets) will be used for the elements that have the same property defined in fanstatic and webassets. Example of a problematic case:

1. You have `.button {color: white}` inside default theme.
2. Previously you've used additional css file, that was required by fanstatic inside `styles` block after a call to `super`. inside this file, you wrote something like `.button{color:black}`.
3. In the generated output, you'll have `.button{color:white} .button{color:black}`. Latter wins, button uses black color.

After webassets:
1. All core styles are required by webassets.
2. Your extension includes styles with fanstatic.
3. Your, custom styles written first(fanstatics writes to the begining of the `head`) and default styles comes after.
4. Default styles win - button uses white color.

amercader:
Thanks so much @smotornyuk for the extra detail and for the example. I think it will make a huge difference that extensions have a whole version to update. I understand the limitations in using both systems at the same time but that's not the main use case anyway.

Big +1 from me to merge this!

PR #4615, opened 2019-01-09
Add docker compose ckan dev settings

ctownsen357:
Fixes #
N/A
### Proposed fixes:
N/A

### Features:
Adds functionality, commented out by default, that allows one to use the project `ckan` docker container as a development environment for interactive debugging with `pdb`.

Uncommenting the sections from the `docker-compose.yml` file as instructed in the update documentation, starting the containers, will then allow one to attach to the ckan container via a `docker attach ckan` command and interact with the code by using the `pdb` library and decorating any target code with `pdb.set_trace()` commands.  Additionally the mapping of the code from the ckan project directory into the target deployment volume on the container allows one to make changes to the code base and see the changes immediately in the running container after the service automatically restarts.

- [ ] includes API changes

### DIFF
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -38,6 +38,11 @@ services:
       - ckan_config:/etc/ckan
       - ckan_home:/usr/lib/ckan
       - ckan_storage:/var/lib/ckan
+    # Uncommenting the following 4 lines allows one to attach to the ckan container and interact with pdb sessions
+      #- ../../ckan/:/usr/lib/ckan/venv/src/ckan/ckan/
+    #stdin_open: true
+    #tty: true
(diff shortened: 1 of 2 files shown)


ctownsen357:
I've tried to go back through my documentation edits to make sure they are consistent.The warning from circleci that causes the test to fail doesn't give me any indication on what the problem is other than inconsistent indentation and I've tried to verify that where I am indenting is consistent with the indentation that occurs before my own. 

I'm not sure how to resolve the documentation test failure.

frafra:
This is a nice improvement, but what about using breakpoints with pdb? Using `set_trace` can work while developing, but it trickier to use that after the service has been deployed.

tino097:
This PR is old and some of the changes could not apply or have been fixed within other PR's

PR #4621, opened 2019-01-17
Ensure URL encoding for all requests

### LINKED ISSUE #4619: Handling incorrect URL encoding in requests
CKAN 2.8.2

Page renders normally

HTTP error 503 due to ascii encoding error

`curl -v "https://demo.ckan.org/dataset?q=ä"`, or opening the same page with Internet Explorer 10

### Commentary
This is similar to #2328. The client sends some non-escaped non-utf8 encoding (for example bare `\xE4` bytes when sending an 'ä' in ISO-8859-1 instead of the correct `%C3%A4`) without (or with incorrect) Content-Type which is assumed to be escaped utf8. When the url strings are used in eg. interpolation, things break.

I've hacked together a quick-and-dirty solution to the issue by using `chardet` library to detect the encoding of `PATH_INFO`, `REQUEST_URI` and `QUERY_STRING` in environ at the start of `ckan.middleware.AskAppDispatcherMiddleware.handle_i18n`, then (if they aren't detected as utf-8) decoding from that and encoding to utf-8, and finally `urllib.quote` each of them before overwriting the values back to environ. This also replaces the `urllib.quote` later when `qs` is being handled.

This is actually a client issue, but since IE is still widely used I think something needs to be done to fix this. My current fix is somewhat heavy and not really elegant, but it does at least seem to work.

bzar:
Fixes #4619 

### Proposed fixes:
For each incoming request, preprocess `PATH_INFO`, `REQUEST_URI` and `QUERY_STRING` variables in `environ` as follows:
- If the variable cannot be decoded as UTF-8, try to decode it as CP1252, then encode it as UTF-8
  - If both decodings fail, the request is considered invalid
- Ensure proper URL quoting of non-ascii characters for the variable with `urllib.quote`, but leave reserved URL characters `:;/?&=@` as-is (since certain fanstatic-URLs contain these), and avoid double-quoting by leaving `%` unquoted as well.
- Remove now unnecessary quoting from later in the code

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/__init__.py
+++ b/ckan/config/middleware/__init__.py
@@ -127,8 +127,21 @@ class AskAppDispatcherMiddleware(object):
             * CKAN_CURRENT_URL is set to the current application url
         '''
 
+        # Ensure URL encoding
+        for var in ('PATH_INFO', 'REQUEST_URI', 'QUERY_STRING'):
+            val = environ.get(var)
+            if val:
+                try:
+                    val.decode('utf-8')
+                except UnicodeDecodeError:
+                    val = val.decode('cp1252').encode('utf-8')
+
+                # Preserve reserved characters and avoid double-quoting
+                environ[var] = urllib.quote(val, '%:;/?&=@')
+
         # We only update once for a request so we can keep
         # the language and original url which helps with 404 pages etc
+
         if 'CKAN_LANG' not in environ:
             path_parts = environ['PATH_INFO'].split('/')
             if len(path_parts) > 1 and path_parts[1] in self.locale_list:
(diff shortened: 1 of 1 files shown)


ckan/config/middleware/__init__.py
```diff
@@ -127,8 +127,21 @@ def handle_i18n(self, environ):
...
+        for var in ('PATH_INFO', 'REQUEST_URI', 'QUERY_STRING'):
+            val = environ.get(var)
+            if val:
+                try:
+                    val.decode('utf-8')
+                except UnicodeDecodeError:
+                    val = val.decode('cp1252').encode('utf-8')
+
+                # Preserve reserved characters and avoid double-quoting
+                environ[var] = urllib.quote(val, '%:;/?&=@')
```
wardi: what's the quoting here for?

wardi: generally values inside the application to be Unicode and unquoted for any particular use. That makes handling them less surprising. When values are sent outside the application they should be encoded/quoted based on where they're being sent.

PR #4626, opened 2019-01-18
Activity stream fixes

### LINKED ISSUE #4625: Activity Streams don't show dataset deletions

Latest. The problematic line is: https://github.com/ckan/ckan/pull/3466/files#diff-43183fa02245f85a419fe7bd91d0b482R207 which went into CKAN 2.8.0, but maybe the issue goes back further.

When you delete a dataset in a group or org, the deletion of it should show up as an entry in the group/org's activity stream.

The group & org activity stream loses all previous entries for the dataset and doesn't create the expected deletion entry.

davidread:
Fixes #4625

### Proposed fixes:

* A deleted dataset's changes & deletion now show in its group & organization activity streams.
* Organization activity stream is produced with a more efficient db query, now it is not reusing the group one any more.
* Added a bunch of tests for the *display* of activity streams (in anticipation of mucking around with the templates)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2633,7 +2633,7 @@ def organization_activity_list(context, data_dict):
     org_show = logic.get_action('organization_show')
     org_id = org_show(context, {'id': org_id})['id']
 
-    _activity_objects = model.activity.group_activity_list(
+    _activity_objects = model.activity.organization_activity_list(
         org_id, limit=limit, offset=offset)
     activity_objects = _filter_activity_by_user(
         _activity_objects, _activity_stream_get_filtered_users())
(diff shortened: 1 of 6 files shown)


ckan/logic/action/delete.py
```diff
@@ -381,6 +382,28 @@ def _group_or_org_delete(context, data_dict, is_org=False):
...
+    else:
+        activity_type = 'deleted group'
+
+    activity_dict = {
+            'user_id': model.User.by_name(user.decode('utf8')).id,
+            'object_id': group.id,
+            'activity_type': activity_type,
+            'data': {
+                'group': dictization.table_dictize(group, context)
+                }
```
wardi: too much indent

wardi:
phew, that's a lot of updated tests. Looks good to me.

PR #4635, opened 2019-01-24
Run tests in Docker

frafra:
No excuses for not running tests!

Here is the code I am using to run the tests:

```bash
cd contrib/docker
docker-compose -f docker-compose-test.yml up --build -d
cd -
python3 -m venv venv
source venv/bin/activate
pip install -r requirements.txt -r dev-requirements.txt
python setup.py install
cd venv
nosetests --ckan --with-pylons=../test-core-docker.ini ckan ckanext
```

Stopping the container:
```bash
cd -
cd ckan/contrib/docker
docker-compose -f docker-compose-test.yml down
```

### DIFF
--- a/test-core-docker.ini
+++ b/test-core-docker.ini
@@ -22,11 +22,11 @@ testing = true
 
 
 # Specify the Postgres database for SQLAlchemy to use
-sqlalchemy.url = postgresql://ckan:pass@db/ckan
+sqlalchemy.url = postgresql://ckan_default:pass@db/ckan_test
 
 ## Datastore
-ckan.datastore.write_url = postgresql://ckan:pass@db/datastore
-ckan.datastore.read_url = postgresql://datastore_ro:pass@db/datastore
+ckan.datastore.write_url = postgresql://datastore_read:pass@db/datastore_test
+ckan.datastore.read_url = postgresql://datastore_write:pass@db/datastore_test
 ckan.datastore.sqlsearch.enabled = true
 ckan.datastore.sqlsearch.allowed_functions_file = %(here)s/ckanext/datastore/tests/allowed_functions.txt
 
--- a/test-core-docker.ini
+++ b/test-core-docker.ini
@@ -1,0 +1,134 @@
+#
+# ckan - Pylons testing environment configuration
+#
+# The %(here)s variable will be replaced with the parent directory of this file
+#
+[DEFAULT]
+# Uncomment and replace with the address which should receive any error reports
+# email_to = errors@example.com
+# smtp_server = localhost
+# error_email_from = ckan-errors@example.com
+
+[server:main]
+use = egg:Paste#http
+host = 0.0.0.0
+port = 5000
+
+[app:main]
+use = egg:ckan
+cache_dir = %(here)s/data
+debug = false
+testing = true
+
+
+# Specify the Postgres database for SQLAlchemy to use
+sqlalchemy.url = postgresql://ckan:pass@db/ckan
+
+## Datastore
+ckan.datastore.write_url = postgresql://ckan:pass@db/datastore
+ckan.datastore.read_url = postgresql://datastore_ro:pass@db/datastore
+ckan.datastore.sqlsearch.enabled = true
+ckan.datastore.sqlsearch.allowed_functions_file = %(here)s/ckanext/datastore/tests/allowed_functions.txt
+
+ckan.datapusher.url = http://datapusher.ckan.org/
+ckan.datapusher.formats = csv xls xlsx tsv application/csv application/vnd.ms-excel application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
+
+## Solr support
+solr_url = http://solr:8983/solr/ckan
+
+# Redis URL. Use a separate Redis database for testing.
+ckan.redis.url = redis://redis:6379/1
+
+ckan.auth.user_create_organizations = true
+ckan.auth.user_create_groups = true
+ckan.auth.create_user_via_api = false
+ckan.auth.create_user_via_web = true
+ckan.auth.create_dataset_if_not_in_organization = true
+ckan.auth.anon_create_dataset = false
+ckan.auth.user_delete_groups=true
+ckan.auth.user_delete_organizations=true
+ckan.auth.create_unowned_dataset=true
+ckan.auth.create_default_api_keys =true
+
+ckan.cache_enabled = False
+
+ckan.site_id = test.ckan.net
+ckan.site_title = CKAN
+ckan.site_logo = /images/ckan_logo_fullname_long.png
+ckan.site_description =
+licenses_group_url =
+ckan.site_url = http://test.ckan.net
+package_new_return_url = http://localhost/dataset/<NAME>?test=new
+package_edit_return_url = http://localhost/dataset/<NAME>?test=edit
+ckan.extra_resource_fields = alt_url
+
+# Change API key HTTP header to something non-standard.
+apikey_header_name = X-Non-Standard-CKAN-API-Key
+
+ckan.plugins =
+# ckan.views.default_views =
+
+# use <strong> so we can check that html is *not* escaped
+ckan.template_head_end = <link rel="stylesheet" href="TEST_TEMPLATE_HEAD_END.css" type="text/css">
+
+# use <strong> so we can check that html is *not* escaped, div is used for a11y compliance
+ckan.template_footer_end = <div role="region"><strong>TEST TEMPLATE_FOOTER_END TEST</strong></div>
+
+# mailer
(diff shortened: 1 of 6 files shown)


wardi:
You can simplify your ini file if you `use = config:test-core.ini` like the `test.ini` and `test-core-circleci.ini` files do

frafra:
Notes:
- `ckan.tests.config.test_environment.TestUpdateConfig.test_update_config_db_url_precedence` fails locally (the test calls `p.load()`, which tries to connect to the database and it fails, while it should not try to do it)
- I run `setup.py install` every time I want to run tests on CKAN, because running `nosetests` from the main directory fails: running it in a different one with ckan installed works

test-core-docker.ini
```diff
@@ -0,0 +1,16 @@
+#
+# ckan - Pylons testing environment configuration
+#
+
+[server:main]
+use = config:test-core.ini
+
+[app:main]
+use = config:test-core.ini
```
wardi: the test.ini has `faster_db_test_hacks = True` could that be the issue?

frafra: It still fails even if I set `faster_db_test_hacks = False`.

frafra:
I started to work on it again. Rebased on the current master, instructions updated accordingly.

frafra:
Tests are starting to run, finally!
Some notes:
- Cannot drop the all the tables due to PostGIS (CKAN Docker uses PostGIS as default, while CircleCI uses regular Postgres)
- Some tests still fail

frafra:
Ok, everything is working as expected right now. I am going to open a new issue for PostGIS.

frafra:
No known regression is seen. The only failing test is not related to this PR.

This PR is useful both to run test in a reproducible environment locally and to further improve Docker setup, as potential regression on the Docker setup could be spotted easily.

frafra:
I could have used a compose file similar to the regular one (which generates the configuration file automatically), but I hit some issues with the storage path and other environment variables that look to have higher priority. Having a static configuration for tests should not be a problem, I guess.

frafra:
PostGIS cannot be used (which is the default for Docker setups):

```
E       sqlalchemy.exc.InternalError: (psycopg2.errors.DependentObjectsStillExist) cannot drop table spatial_ref_sys because extension postgis requires it
E       HINT:  You can drop extension postgis instead.
E       
E       [SQL: 
E       DROP TABLE spatial_ref_sys]
E       (Background on this error at: http://sqlalche.me/e/2j85)

../../lib/python3.8/site-packages/sqlalchemy/engine/default.py:550: InternalError
```

frafra:
We can use `ckan/ckan-postgres-dev` as Postgres image, without having to alter the `Dockerfile` for PostgreSQL

frafra:
I made changed the Postgres test image and used `-c fsync=off -c full_page_writes=off` to try to speed up execution.

I see a couple of errors:
```
______________________ TestPackageNew.test_complete_package_with_two_resources ______________________

self = <ckan.tests.controllers.test_package.TestPackageNew object at 0x7ffa0602cb50>
app = <ckan.tests.helpers.CKANTestApp object at 0x7ffa006e9af0>, user_env = {'REMOTE_USER': 'fperry'}

    def test_complete_package_with_two_resources(self, app, user_env):
        url = url_for("dataset.new")
        response = app.post(url, environ_overrides=user_env, data={
            "name": "complete-package-with-two-resources",
            "save": "",
            "_ckan_phase": 1
        }, follow_redirects=False)
        location = _get_location(response)
        app.post(location, environ_overrides=user_env, data={
            "id": "",
            "url": "http://example.com/resource0",
            "save": "again"
        })
        app.post(location, environ_overrides=user_env, data={
            "id": "",
            "url": "http://example.com/resource1",
            "save": "go-metadata"
        })
        pkg = model.Package.by_name(u"complete-package-with-two-resources")
>       assert pkg.resources[0].url == u"http://example.com/resource1"
E       AssertionError: assert 'http://example.com/resource0' == 'http://example.com/resource1'
E         - http://example.com/resource1
E         ?                            ^
E         + http://example.com/resource0
E         ?                            ^

ckan/tests/controllers/test_package.py:199: AssertionError
___________________________ TestUserList.test_user_list_order_by_default ____________________________

self = <ckan.tests.logic.action.test_get.TestUserList object at 0x7ff9f7e00c10>

    def test_user_list_order_by_default(self):
        default_user = helpers.call_action("get_site_user", ignore_auth=True)
    
        users = [
            factories.User(fullname="Xander Bird", name="bird_x"),
            factories.User(fullname="Max Hankins", name="hankins_m"),
            factories.User(fullname="", name="morgan_w"),
            factories.User(fullname="Kathy Tillman", name="tillman_k"),
        ]
        expected_names = [
            u["name"]
            for u in [
                users[3],  # Kathy Tillman
                users[1],  # Max Hankins
                users[2],  # morgan_w
                users[0],  # Xander Bird
            ]
        ]
    
        got_users = helpers.call_action("user_list")
        got_names = [
            u["name"] for u in got_users if u["name"] != default_user["name"]
        ]
    
>       assert got_names == expected_names
E       AssertionError: assert ['tillman_k',...', 'morgan_w'] == ['tillman_k',..._w', 'bird_x']
E         At index 2 diff: 'bird_x' != 'morgan_w'
E         Full diff:
E         - ['tillman_k', 'hankins_m', 'morgan_w', 'bird_x']
E         ?                                      ----------
E         + ['tillman_k', 'hankins_m', 'bird_x', 'morgan_w']
E         ?                            ++++++++++

ckan/tests/logic/action/test_get.py:970: AssertionError
```

frafra:
https://github.com/ckan/ckan/blob/b47c0e53bb5f1f4c3417c7dcb39a599d32ec0799/ckan/model/package.py#L101-L119

Resource is appended, so it would make sense that the first resource would come first, but at the moment tests assume it is the opposite. It looks like there is no order enforced when fetching the resources related to a package.

Using `postgres:10-alpine` instead of `postgres:12-alpine` make the test work again. Different versions of Postgres produce a different order. This confirms that the order is not enforced at any level, so the test should not assume any specific order.

The second failing test looks at the order specifically. It looks like that regular Postgres images (built with glibc) pass the test, while the Alpine images (built with musl) make the test fails.

| Postgres version | Distribution | test_complete_package_with_two_resources | test_user_list_order_by_default |
| ---------------- | ------------ | ---------------------------------------- | ------------------------------- |
| 9.6 | Debian | ✓ | ✓ |
| 9.6 | Alpine | ✓ | ✗ |
| 10 | Debian | ✓ | ✓ |
| 10 | Alpine | ✓ | ✗ |
| 11 | Debian | ✓ | ✓ |
| 11 | Alpine | ✓ | ✗ |
| 12 | Debian | ✗ | ✓ |
| 12 | Alpine | ✗ | ✗ |
| 13 | Debian | ✗ | ✓ |
| 13 | Alpine | ✗ | ✗ |
| 14 | Debian | ✗ | ✓ |
| 14 | Alpine | ✗ | ✗ |

frafra:
I fixed the first test with a patch, by not assuming any specific order. I hope that other developers are doing the same too. (!)

The second is not trivial. Here is the commit which introduced it: https://github.com/ckan/ckan/commit/739e690799cc18a31cc0fc69e8e5e90a10d74ec2

```
$ docker run --rm alpine sh -c 'for name in Xander Max Kathy morgan; do echo $name; done | sort'
Kathy
Max
Xander
morgan
$ docker run --rm debian sh -c 'for name in Xander Max Kathy morgan; do echo $name; done | sort'
Kathy
Max
Xander
morgan
$ docker run --rm --entrypoint="" postgres:12 sh -c 'for name in Xander Max Kathy morgan; do echo $name; done | sort'
Kathy
Max
morgan
Xander
```

As you can see, only the `postgres:12` image, based on Debian, does not put `morgan` at the end. `Morgan` is sorted correctly, instead. That also shows that the problem is not related to the usage of musl.

I found that `postgres:12` locale is set to `en_US.utf8`, while `debian` uses `POSIX` as default. The locale determines if words should be sorted in a case-sensitive manner or not:
```
$ docker run --rm --entrypoint="" postgres:12 sh -c 'for name in Xander Max Kathy morgan; do echo $name; done | LC_ALL=POSIX sort'
Kathy
Max
Xander
morgan
$ docker run --rm --entrypoint="" postgres:12 sh -c 'for name in Xander Max Kathy morgan; do echo $name; done | sort'
Kathy
Max
morgan
Xander
```

I propose to make this test work with both locales.

frafra:
All the tests, within Docker using `postgres:12-alpine` and on Circle CI, are passing.

Usage is described here: https://github.com/ckan/ckan/pull/4635/#issuecomment-1033669915

frafra:
Should this PR be backported?

PR #4636, opened 2019-01-25
Password reset request - generally tighten it up

davidread:
Fixes #

* To reset a password, the user can now only specify username or email - not the looser search done by model.User.search(), which allowed: partial name, partial fullname (and if sysadmin: partial email address) etc
  (This was originally loose to be helpful to users if they forgot their username, but it is more normal these days to only allow resets with your email address. I left in the username option, as explained in the code)
* Don't confirm whether a user exists or not - this is [recommended](https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Authentication_and_Error_Messages)
* Logging added for audit purposes

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/user/request_reset.html
+++ b/ckan/templates/user/request_reset.html
@@ -14,7 +14,7 @@
       {% block form %}
         <form action="" method="post">
           <div class="form-group">
-            <label for="field-username">{{ _('Username') }}</label>
+            <label for="field-username">{{ _('Email or username') }}</label>
             <input id="field-username" class="control-medium form-control" name="user" value="" type="text" />
           </div>
           <div class="form-actions">
@@ -35,8 +35,9 @@
     {% block help_inner %}
     <h2 class="module-heading">{{ _('How does this work?') }}</h2>
     <div class="module-content">
-      <p>{% trans %}Enter your username into the box and we will send you
-      an email with a link to enter a new password.{% endtrans %}</p>
+      <p>{% trans %}Enter your email address or username into the box and we
+        will send you an email with a link to enter a new password.
+        {% endtrans %}</p>
     </div>
     {% endblock %}
   </section>
--- a/ckan/tests/controllers/test_user.py
+++ b/ckan/tests/controllers/test_user.py
@@ -9,7 +9,7 @@ import ckan.tests.helpers as helpers
 
 from ckan.lib.helpers import url_for
 from ckan import model
-from ckan.lib.mailer import create_reset_key
+from ckan.lib.mailer import create_reset_key, MailerException
 
 webtest_submit = helpers.webtest_submit
 submit_and_follow = helpers.submit_and_follow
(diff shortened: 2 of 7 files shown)


anotheredward:
Awesome work @davidread !
I'll use the time I have today to test and review this PR and close mine.
Thanks again :).

ckan/templates/user/request_reset.html
```diff
@@ -35,8 +35,9 @@ <h1 class="page-heading">{{ _('Reset your password') }}</h1>
     {% block help_inner %}
     <h2 class="module-heading">{{ _('How does this work?') }}</h2>
     <div class="module-content">
-      <p>{% trans %}Enter your username into the box and we will send you
-      an email with a link to enter a new password.{% endtrans %}</p>
+      <p>{% trans %}Enter your email address or username into the box and we
+        will send you an email with a link to enter a new password.
+        {% endtrans %}</p>
```
anotheredward: Allowing users to enter their username is sensible for the case you've specified below (multiple users with the same email), but with the user listing being enabled by default, it essentially means an attacker has similar information available to them (complete listing of all the users, rather than complete listing of emails).

To patch this hole we'll need to get the users listing page disabled by default.

ckan/tests/controllers/test_user.py
```diff
@@ -689,3 +690,68 @@ def test_user_page_sysadmin_user(self):
...
+
+        assert_in('A reset link has been emailed to you', response)
+        emailed_users = [call[0][0].name
+                         for call in send_reset_link.call_args_list]
+        assert_equal(emailed_users, [user_a['name'], user_b['name']])
+
+    def test_request_reset_without_param(self):
+        app = self._get_test_app()
+        offset = url_for('user.request_reset')
+        app.post(offset, params={}, status=400)
```
anotheredward: For this case I'd expect a flash message rather than a blank page.

screen shot 2019-01-28 at 5 32 37 pm

ckan/views/user.py
```diff
@@ -497,58 +497,79 @@ def _prepare(self):
...
+        self._prepare()
+        id = request.form.get(u'user')
+        if id in (None, u''):
+            base.abort(400, _(u'"user" parameter is required'))
+        log.info(u'Password reset requested for user "{}"'.format(id))
+
+        context = {u'model': model, u'user': g.user, u'ignore_auth': True}
+        user_objs = []
+
+        if u'@' in id:
```
anotheredward: Confirmed this check will work, as usernames cannot contain '@' symbols

ckan/views/user.py
```diff
@@ -497,58 +497,79 @@ def _prepare(self):
...
-                if len(user_list) == 1:
-                    # This is ugly, but we need the user object for the
-                    # mailer,
+        self._prepare()
+        id = request.form.get(u'user')
+        if id in (None, u''):
+            base.abort(400, _(u'"user" parameter is required'))
+        log.info(u'Password reset requested for user "{}"'.format(id))
+
+        context = {u'model': model, u'user': g.user, u'ignore_auth': True}
```
anotheredward: This may seem like an obvious question, but are you adding the `u` prefix to this so it's easier for us to port to Python 3 in the near future?

davidread: yup! :)

ckan/templates/user/request_reset.html
davidread: Agreed. Let's not couple this PR to that piece of work though.

ckan/views/user.py
```diff
@@ -497,58 +497,79 @@ def _prepare(self):
...
             except mailer.MailerException as e:
-                h.flash_error(_(u'Could not send reset link: %s') %
-                              text_type(e))
-        return self.get()
+                # SMTP is not configured correctly or the server is
+                # temporarily unavailable
+                h.flash_error(_(u'Error sending the email. Try again later '
+                                'or contact an administrator for help'))
+                log.exception(e)
+                return
```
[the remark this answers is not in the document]
davidread: Thanks for adding the redirect, which shows the flash message. I've added a test too now: `test_request_reset_but_mailer_not_configured`

ckan/tests/controllers/test_user.py
davidread: Fixed now with your PR :)

davidread:
I've merged @anotheredward 's additions and all is good. Over to you, @amercader for review!

amercader:
Looks good to me. Great work @davidread and @anotheredward!

PR #4638, opened 2019-01-27
Adding suffix to package name if exists another with same name  (issue #4637)

### LINKED ISSUE #4637: Adding suffix to package name if exists another with same name
If you want to create a package with a name that already exists for another dataset, an error is always raised.
This is a common issue if several datasets has the same title (probably from different orgs with same information) because the name is calculated using title.
Sometimes it is desired that name would been modified but no error would be raised. For instance when the datasets are created from an automatic process with no human intervention.

Current version

Yo can choose by defining an .ini property if you want to raises an error or regenerating a new modified dataset name (e.g. adding a suffix)
The default value for this new property could be raises always an error for backwards compatibility.

Always raises an error

Create two datasets with same name

dportoles:
Fixes #4637 

### Features:

- [ ] includes API changes


kmbn:
Thanks for working on this, @dportoles. I have a comment and a question.

Comment: As part of the Pylons to Flask and Python 2 to Python 3 transition, we need to stop using Paste. I just created an issue (#4639) so that there's something to refer to in the future—I know you couldn't have known about the plan in advance.

Question: I see you added a couple of commits to fix failing tests, and CircleCI is still reporting a failing test. Are you having trouble running the tests locally? If so, you're not alone :/ Please let me know—there are a couple of open issues regarding running the tests locally and we need to do a better job of addressing them.

dportoles:
@kmbn Thanks for your comment. I'll change paster reference
Now I'm trying to pass the tests locally. I've some trouble because I've change my laptop and now I'm config new environment.
ATM I don't know if tests are passed yet but I will keep you informed.

dportoles:
@kmbn I've changed calls from paster to ckan.plugins.toolkit? Tell me if now is ok.

Apart from that, my tests are failing locally in a fresh ckan environment :(

ckan/logic/action/create.py
```diff
@@ -166,13 +168,70 @@ def package_create(context, data_dict):
...
+                            else:
+                                # there are other errors despite of
+                                # changing its name => resign
+                                i = MAX_ATTEMPS+1
+                        else:
+                            data['name'] = current_name_temp
+                            data_dict['name'] = current_name_temp
+                            errors = errors2
+                            i = MAX_ATTEMPS+1
+
```
wardi: looping on the validation this way will be very slow on some sites.

How about updating the name validator to include this logic instead? It could make a single LIKE query to the DB to collect names with the same prefix then return the next available one. That should be a lot less code and perform much better.

dportoles: @wardi You're right. Doing this stuff in validator (using DB) is a much better way.
I've updated my code with this approach.
Thanks for your help!

PR #4649, opened 2019-02-05
add .ini path, jetty8 note, nose doc link

anotheredward:
While working through this document I got stuck in a couple of places, I've updated these parts so that others might find it easier in the future :).

### Proposed fixes:
* Explicit path for test-core.ini
* Note that `jetty` may be called `jetty8`
* Link to Nose documentation on running a subset of tests

### Features:
- [ ] includes API changes

### DIFF
--- a/doc/contributing/test.rst
+++ b/doc/contributing/test.rst
@@ -56,7 +56,7 @@ Create test databases:
 
 Set the permissions::
 
-    paster datastore set-permissions -c test-core.ini | sudo -u postgres psql
+    paster datastore set-permissions -c |virtualenv|/src/ckan/test-core.ini | sudo -u postgres psql
 
 This database connection is specified in the ``test-core.ini`` file by the
 ``sqlalchemy.url`` parameter.
@@ -105,7 +105,7 @@ To enable multi-core:
          <core name="ckan" instanceDir="/etc/solr/ckan" />
        </cores>
 
-5. Restart Solr by restarting Jetty (or Tomcat)::
+5. Restart Solr by restarting Jetty (may also be called jetty8) (or Tomcat)::
 
        sudo service jetty restart
 
@@ -133,7 +133,8 @@ drop and reinitialize the database before the run you can use the ``reset-db``
 option::
 
      nosetests --ckan --reset-db --with-pylons=test-core.ini ckan
-
+     
+You can find details on how to run a subset of tests in the `Nose Documentation <https://nose.readthedocs.io/en/latest/usage.html#selecting-tests>`_.
 
 
 .. _migrationtesting:


doc/contributing/test.rst
```diff
@@ -56,7 +56,7 @@ Create test databases:
 
 Set the permissions::
 
-    paster datastore set-permissions -c test-core.ini | sudo -u postgres psql
+    paster datastore set-permissions -c |virtualenv|/src/ckan/test-core.ini | sudo -u postgres psql
```
smotornyuk: In many cases, CKAN instance is configured in quite a unique way, so the assumption that `|virtualenv|/src/ckan/test-core.ini` exists won't be truthy all the time. In addition, `paster datastore`, without `--plugin=...` won't work from every directory, so I suggest to leave this place as it was before

doc/contributing/test.rst
```diff
@@ -105,7 +105,7 @@ To enable multi-core:
          <core name="ckan" instanceDir="/etc/solr/ckan" />
        </cores>
 
-5. Restart Solr by restarting Jetty (or Tomcat)::
+5. Restart Solr by restarting Jetty (may also be called jetty8) (or Tomcat)::
```
smotornyuk: Here service referred only in general, not by its internal name. If you are going to list all possible names of jetty/tomcat service it will take a lot of space. In addition, if you are going to use some other Linux distributive, not Ubuntu, the number of choices will increase, so it has almost no sense to mention `jetty8` separately. But you may add note about the fact, that service names may be different on different systems

doc/contributing/test.rst
```diff
@@ -133,7 +133,8 @@ drop and reinitialize the database before the run you can use the ``reset-db``
 option::
 
      nosetests --ckan --reset-db --with-pylons=test-core.ini ckan
-
+     
+You can find details on how to run a subset of tests in the `Nose Documentation <https://nose.readthedocs.io/en/latest/usage.html#selecting-tests>`_.
```
smotornyuk: Could you please format this line, because it don't pass tests currently

anotheredward: I'm not particularly familiar with .rst.
Are there any resources on running the tests I can look at so I can fix this issue @smotornyuk ?
Thanks!

doc/contributing/test.rst
smotornyuk: i think `your jetty executable may have a different name, eg: jetty8 or jetty9` - is perfect, because it doesn't make restriction to single possible alternative

doc/contributing/test.rst
smotornyuk: Yep, making it clear, that command have to be executed from particuar directory should be ok

doc/contributing/test.rst
smotornyuk: https://docs.ckan.org/en/2.8/contributing/test.html - here's `How to` for CKAN testing.

smotornyuk:
And one general note - we are merging PRs only into master branch, so could you please close this PR and create new one into ckan:master?

Issue #4459, opened 2018-09-13
datastore_search_sql issues with postgres 9.6+


tino097:
### CKAN Version if known (or site URL)

all

### Please describe the expected behaviour

return results
### Please describe the actual behaviour

throwing an error `TypeError: expected string or buffer`

### What steps can be taken to reproduce the issue? 

use postgres 9.6+ and use `datastore_search_sql` API call

https://github.com/ckan/ckan/blob/master/ckanext/datastore/helpers.py#L94
this line here, when using `psycopg2==2.4.5` is receiving `str` for `result['QUERY PLAN']` and if you upgrade the version of `psycopg2` it will be a type of `list`

That is causing https://github.com/ckan/ckan/blob/master/ckanext/datastore/helpers.py#L99
to throw the above error

michab23:
the same problem

tino097:
@michab23 for now you could just update the `psycopg` to 2.7.3.2

michaelbeilin:
@tino097 didn't work for me not with psycopg 2.7.3.2 and not with the latest  psycopg 2.7.7.
This is a workaround for the problem that worked in my case:
if isinstance(result['QUERY PLAN'], list):
   result_query_plan = json.dumps(result['QUERY PLAN'])
   query_plan = json.loads(result_query_plan)

Regards

tino097:
PR would be great @michaelbeilin

PR #4651, opened 2019-02-06
in case result['QUERY PLAN'] is list

michaelbeilin:
in case result['QUERY PLAN'] is list and not string as expected
https://github.com/ckan/ckan/issues/4459#issue-359865778
Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/helpers.py
+++ b/ckanext/datastore/helpers.py
@@ -96,7 +96,11 @@ def get_table_names_from_sql(context, sql):
                 sql.encode('utf-8'))).fetchone()
 
         try:
-            query_plan = json.loads(result['QUERY PLAN'])
+            if isinstance(result['QUERY PLAN'], list):
+                result_query_plan = json.dumps(result['QUERY PLAN'])
+            else:
+                result_query_plan = result['QUERY PLAN']
+            query_plan = json.loads(result_query_plan)
             plan = query_plan[0]['Plan']
 
             t, q = _get_table_names_queries_from_plan(plan)


amercader:
+1 we've been bitten by this and it would be good to future proof against future psycopg versions

wardi:
What's the root cause here, does https://github.com/ckan/ckan/blob/4171ce9abb8a0b2e77979022ef285f8ef95b7e5b/ckanext/datastore/backend/postgres.py#L123-L128 not work with some versions of psycopg2?

If that's true then there will be more places in the code that break like this.

We might be better off adding `::text` to the queries we want as text instead of json as http://initd.org/psycopg/docs/extras.html#json-adaptation suggests

amercader:
After further investigation, this happens on CKAN <= 2.7 + psycopg2 > 2.5, because the change @wardi pointed to (included in https://github.com/ckan/ckan/pull/3978) is not part of 2.7.

Although the psycopg2 version pinned on CKAN 2.7 is 2.4.x, sometimes is necessary to upgrade to work around issues like postgres client versions etc. So I agree that this should not go in master, but what's the best approach to patch 2.7?

Should be include this partial fix or backport https://github.com/ckan/ckan/pull/3978? Note that this implies changing the psycopg2 requirement.

wardi:
best fix is adding `::text` to the affected queries, that way psycopg version doesn't matter

pdelboca:
Closing this as is no longer relevant in newer versions of CKAN.

Issue #3441, opened 2017-02-17
Hide list of users by default


amercader:
This comes up regularly on security audits and I agree that is not standard practice to get a list of all user names in the system. I propose that we switch it to sysadmin only and people can override it if they want.

wardi:
:+1: from me. Need to think about how this affect user lists available other places, like in organization_show, and what details should be available if for anon users calling user_show with a known username.

mattfullerton:
I would be for a configuration option where organizations a user is not a member of are also not shown

tino097:
@amercader i will take look on this issue too

tino097:
@amercader 
currently i could set only authorized users to be able to access to 'user_list'

`def user_list(context, data_dict):
    # Users list is visible by default
    auth_user_obj = context.get('auth_user_obj')
    if auth_user_obj == None:
        return {'success' : False}
    return {'success': True}`

im not sure what other cases to take into consideration

wildcatzita:
We can use some cases for show user list for current login user:

1. For register user without organizations we can show only the same users
2. For user in organizations we can show only users from this organizations and registered users without organizations
3. For user in top organization we can show all users from top organization and child organizations
4. For sysadmins we show all users from cases above and sysadmins 

@amercader , @tino097 , @wardi what do you think about it?

anotheredward:
We're planning on either removing this and the user show feature, or implementing sysadmin only access in the near future, will open a PR once we have a fix.

Happy for future implementers to make something more complex, but removing or sysadmin only fixes the pressing security issue.

amiedes:
Hi, what's the status on a *sysadmin-only* access fix for this?

We have to do exactly this for one of our installations of CKAN and I was wondering if someone already implemented it somewhere or can we work on a PR to CKAN instead of doing it in one of our extensions.

anotheredward:
If you have made the change @amiedes , then it's probably worth submitting :)

anotheredward:
@amercader I'm starting on this feature now, will be locking down to sysadmin only for now.

anotheredward:
After investigation I'm looking at adding another flag, something like `ckan.auth.sysadmin_only_user_details`.
`ckan.auth.public_use_details` already exists, but I don't want to break the existing api.

I'm making the assumption that I won't need to modify anywhere except where the application already checks for the `ckan.auth.public_user_details` flag, is that correct?

[...]

PR #4654, opened 2019-02-09
default public access to false, restrict user list to sysadmins

anotheredward:
Fixes #
Hides the user listing page from all users except Sysadmins by default.
Swaps the default of the  `ckan.auth.public_user_details` flag so that user details are restricted to logged-in users by default, and can be made public by setting this flag to `true` (`true` was the previous default.

I'm still working on this PR, opening early for discussion.
Currently password reset is broken by these changes, currently investigating.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -111,9 +111,8 @@ class UserController(base.BaseController):
         limit = int(
             request.params.get('limit', config.get('ckan.user_list_limit', 20))
         )
-        try:
-            check_access('user_list', context, data_dict)
-        except NotAuthorized:
+
+        if (not asbool(config.get('ckan.auth.public_user_details', False))):
             abort(403, _('Not authorized to see this page'))
 
         users_list = get_action('user_list')(context, data_dict)


anotheredward:
@davidread @amercader I've started work on this PR in reference to #3441 .
I've gone with the following option:
1. Users must be logged in to see any user details by default
2. Only Sysadmins will be able to see the user listing page by default
3. If the `ckan.auth.public_user_details` flag is set to `true`, then the user listing and user detail pages will be publicly accessible.

I feel like this restricts the potential for user enumeration, ensures that very open installations of ckan work the same as before, and keeps the behaviour of the flag very similar, but opting for a secure default instead.

ckan/logic/auth/get.py
```diff
@@ -96,11 +96,8 @@ def tag_list(context, data_dict):
 
 def user_list(context, data_dict):
     # Users list is visible by default
-    if not asbool(config.get('ckan.auth.public_user_details', True)):
-        return restrict_anon(context)
-    else:
-        return {'success': True}
-
+    allow_public = asbool(config.get('ckan.auth.public_user_details', False))
+    return {'success': allow_public or context['user_is_admin']}
```
amercader: We don't need the `or context['user_is_admin']` bit, Sysadmins override these auth functions so they will get access anyway.

amercader:
@anotheredward thanks for the PR. Just to clarify you are implementing Option 2 in here right? https://github.com/ckan/ckan/issues/3441#issuecomment-460665708

Given the amount of tests failing I suspect that limiting the user lists on groups and orgs has undesired effects. We probably need to allow members of that org to list the users when requesting the object otherwise this breaks things in member management, activities, etc.

Actually after having another look it might be that limiting `user_show` could be also the one causing problems. I think sensitive properties from users are stripped now anyway so maybe we don't need to completely block user_show

I recommend doing one thing at a time and make sure that tests pass at each stage, and maybe we can merge just a subset of the changes for now:

1. user list
2. group/org show
3 user show

anotheredward:
@amercader Here's a PR for just locking down the user page.

davidread:
From the looks of this PR, sysadmins won't be able to see this page either. `check_access` always returns True for sysadmins, so you'd want that functionality here.

The other issue with this seems to be that the API is not protected - anyone can still call that, even in a browser to get the user list.

I think what @amercader is suggesting is still to do the change in ckan/logic/auth/get.py but let's try it only in user_list and then we can see what test breakages that causes and decide how to deal with that.

anotheredward:
That makes sense, thanks for the clarification.

The issue I was having was specifically with user_list:
https://github.com/ckan/ckan/commit/b31d5d13a492e050f6afb59f9f81bb32570360a0
Once I had locked down `user_list` then the `reset email` submission started failing as well.

I'm not sure why this is, but I expect I'll need to review what uses this permission, and test that that functionality still works as expected, when the user_list function is restricted to sysadmins only.

I'll take a look further into this when I next have time :).

ThrawnCA:
@anotheredward Locking down `user_list` to sysadmins means that any non-sysadmin will be unable to access the user autocomplete API, which is a problem when adding members to organisations. I'm not certain what further effects it would have, but I think that it would make more sense to allow any organisation admin to access `user_list`.

pdelboca:
Closing this PR due to inactivity. Thanks!

PR #4662, opened 2019-02-22
Fix hardcoded root paths

amiedes:
On a CKAN mounted on a path different than '/', the redirect made after requesting a password reset is not taking into account the path where it is mounted. We've solved it by replacing this paths.

### Proposed fixes:

Replace hardcoded '/' paths per the corresponding controller & action

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/user.py
+++ b/ckan/views/user.py
@@ -563,14 +563,14 @@ class RequestResetView(MethodView):
                 h.flash_error(_(u'Error sending the email. Try again later '
                                 'or contact an administrator for help'))
                 log.exception(e)
-                return h.redirect_to(u'/')
+                return h.redirect_to(u'home.index')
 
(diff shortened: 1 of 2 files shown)


ckan/views/user.py
```diff
@@ -563,14 +563,14 @@ def post(self):
                 h.flash_error(_(u'Error sending the email. Try again later '
                                 'or contact an administrator for help'))
                 log.exception(e)
-                return h.redirect_to(u'/')
+                return h.redirect_to(controller=u'home', action=u'index')
```
tino097: Can you please use calling the action as `'home.index'`

tino097:
@amiedes And if its ok for tou, can you please split the changes in two separate commits, one for the controller and one for the view, so we could easily cherry-pick for backporting

amiedes:
BTW changes adressed

PR #4664, opened 2019-02-22
Remove vdm - exploratory

davidread:
**NOTE: Closed in favour of https://github.com/ckan/ckan/pull/4784**

I'm aware @TkTech is [looking at this too](https://github.com/ckan/ckan/pull/3972#issuecomment-465523145). But I wanted to think about it in the meantime.

This is a quick hack to see the broad implications of removing revisioning / vdm. I've not tackled migration. This work is branched off 3484_revision_ui_removal2, since that removes the revision UI.

Observations:
* user_dict.number_of_edits is removed, as calculating that relied on revisions
* stats extension loses all the stats time-line, as calculating those relied on revisions. could be resurrected with activitie stream data.
* trash page removed (no revisions to trash)
* migrate_filestore won't work any more

Not working:
* fix up activities - they are not being created for some reason
* migration


davidread:
Closed in favour of #4784

PR #4627, opened 2019-01-20
Activity stream html

davidread:
This PR changes the way Activity Stream is displayed. Instead of the stream being rendered in a logic function that returns HTML, with this PR they are rendered with standard templates.

https://github.com/ckan/ckan/pull/4626 is merged to master, so this is ready to be reviewed!

Most of this PR was originally coded by @TkTech who said:
> Activities are now just jinja2 snippets meaning you can override them in your themes and adding new activities is much, much, much simpler.

I think these logic functions were originally designed to return HTML to make it easier to embed activity feeds in another website, but I don't think we've seen this happen.

These are the logic functions removed:

*    user_activity_list_html
*    package_activity_list_html
*    group_activity_list_html
*    organization_activity_list_html
*    recently_changed_packages_activity_list_html
*    dashboard_activity_list_html

They were deprecated in v2.7.0 changelog, so ok to remove.

This PR is an excerpt of the larger: https://github.com/ckan/ckan/pull/3972, which is critical for removing vdm & pylons and moving to python 3.

Note: This PR introduces `object_type` as parameter to `snippets/activity_stream.html`. It is so far unused, but will be needed in the follow-on PR: #3972. It was just simpler to leave it in for this PR.

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #4685, opened 2019-03-14
[#4684] Use get_action to call activity actions

### LINKED ISSUE #4684: Activity stream actions not overridable
Some calls to action functions in `action/get.py` call the functions directly instead of using `get_action`, meaning they can't be overridden from extensions.

amercader:
Fixes #4684 

Otherwise they can not be overridden.

This targets the next 2.8 patch release as it's no longer relevant in master after #4627


amercader:
Backport to 2.7 only

PR #4686, opened 2019-03-14
Fix typo in French ckan.po

fvanderbiest:


fvanderbiest:
Fixed on https://www.transifex.com/okfn/ckan/translate/#fr/2-7/114662740 & https://www.transifex.com/okfn/ckan/translate/#fr/2-8/138387730

PR #4691, opened 2019-03-15
Remove vdm from PackageExtra and GroupExtra

### LINKED ISSUE #4690: Remove vdm from PackageExtra and GroupExtra
This is part of the effort to get rid of vdm from CKAN

davidread:
Fixes #4690

* [x] PackageExtra no longer revisioned, but its revision table is kept (frozen) for the benefit of a future migration: [migrate_package_activity.py](#3972)
* [x] PackageExtra no longer stateful
* [x] GroupExtra no longer revisions, but its revision table is kept (frozen)
* [x] GroupExtra no longer stateful
* [x] State enum and StatefulObjectMixin are moved from vdm into ckan

Note: We're agreed that revisioning can go as soon as #3972 is merged and that should be soon, but maybe we want to confirm that that's going into this release of ckan before this one can be merged.

-- 

The original aim of this was to simply move the few lines of vdm.sqlalchemy.StatefulObjectMixin into CKAN, because I think we still want a package to be stateful - i.e. when you delete it, it only changes state=deleted. This allows an admin to still see it and look at its activity stream, see who deleted it etc.

However I didn't want to move into CKAN a whole load of vdm stateful stuff like the complicated vdm bits that are used by PackageExtras and GroupExtras:
* vdm.sqlalchemy.stateful.DeferredProperty
* vdm.sqlalchemy.stateful.StatefulDict
* vdm.sqlalchemy.stateful.OurAssociationProxy

So I figured that by making PackageExtra and GroupExtra not stateful, we don't need those vdm things at all. We don't need PackageExtra and GroupExtra to be stateful, because when we delete an extra, unlike with packages there's no benefit of it hanging around still - an admin can still see the history of it in the activity stream (as soon as #3972 is merged).

And when you remove the statefulness, you need a migration to delete the PackageExtra & GroupExtra objects which are state=deleted, otherwise they will suddenly appear to users. And to delete these objects, we need to delete the ForeignKey relation from PackageExtraRevision and GroupExtraRevision (continuity_id). So I needed to remove the revisioning from these objects too.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
+++ b/ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
@@ -1,0 +1,11 @@
+# encoding: utf-8
+
+
+def upgrade(migrate_engine):
+    migrate_engine.execute(
+        '''
+            ALTER TABLE "package_extra_revision"
+              DROP CONSTRAINT package_extra_revision_continuity_id_fkey;
+            DELETE FROM "package_extra" WHERE state='deleted';
+        '''
+    )
--- a/ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
+++ b/ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
@@ -6,6 +6,9 @@ def upgrade(migrate_engine):
         '''
             ALTER TABLE "package_extra_revision"
               DROP CONSTRAINT package_extra_revision_continuity_id_fkey;
+            ALTER TABLE "group_extra_revision"
+              DROP CONSTRAINT group_extra_revision_continuity_id_fkey;
             DELETE FROM "package_extra" WHERE state='deleted';
+            DELETE FROM "group_extra" WHERE state='deleted';
         '''
     )
--- a/ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
+++ b/ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
@@ -4,11 +4,11 @@
 def upgrade(migrate_engine):
     migrate_engine.execute(
         '''
-            ALTER TABLE "package_extra_revision"
-              DROP CONSTRAINT package_extra_revision_continuity_id_fkey;
-            ALTER TABLE "group_extra_revision"
-              DROP CONSTRAINT group_extra_revision_continuity_id_fkey;
-            DELETE FROM "package_extra" WHERE state='deleted';
-            DELETE FROM "group_extra" WHERE state='deleted';
+ALTER TABLE "package_extra_revision"
+    DROP CONSTRAINT IF EXISTS package_extra_revision_continuity_id_fkey;
+ALTER TABLE "group_extra_revision"
+    DROP CONSTRAINT IF EXISTS group_extra_revision_continuity_id_fkey;
+DELETE FROM "package_extra" WHERE state='deleted';
+DELETE FROM "group_extra" WHERE state='deleted';
         '''
     )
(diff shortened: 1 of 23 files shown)


TkTech:
Give me a poke when you think it's ready to review, I've got time set aside for anything related to vdm. Really happy to see this done.

davidread:
@TkTech Yes please to a review now! Just probably not for merging until we're sure we're getting #3972 into this release and can hack out revision code like this.

ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
```diff
@@ -0,0 +1,14 @@
+# encoding: utf-8
+
+
+def upgrade(migrate_engine):
+    migrate_engine.execute(
+        '''
+            ALTER TABLE "package_extra_revision"
+              DROP CONSTRAINT package_extra_revision_continuity_id_fkey;
```
TkTech: `DROP CONSTRAINT IF EXISTS` is probably a good idea here, I know some people have manually removed these for performance.

ckan/model/group.py
```diff
@@ -352,12 +357,11 @@ def all_related_revisions(self):
         this group. Ordered by most recent first.
         '''
         results = {}
-        from group_extra import GroupExtra
         for grp_rev in self.all_revisions:
             if not grp_rev.revision in results:
                 results[grp_rev.revision] = []
             results[grp_rev.revision].append(grp_rev)
-        for class_ in [Member, GroupExtra]:
+        for class_ in [Member]:  # GroupExtra is not revisioned any more
```
TkTech: Doesn't need to be a loop any longer, might as well simplify it while we're here.

ckan/migration/versions/088_delete_extras_which_are_deleted_state.py
davidread: 👍 Done

ckan/model/group.py
davidread: Tis true. However most revision bits, like this method, will be deleted in another PR (along the lines of https://github.com/ckan/ckan/pull/4664/files#diff-63e3d56154b1b0ef0681f88666b149c9L360 ) which should be merged shortly afterwards (in the same release of CKAN), so really this code just needs to not be in a broken state for the meantime.

davidread:
I checked that in this branch the model indeed removes those FK constraints and nothing else:
```
git checkout master
paster --plugin=ckan db clean
paster --plugin=ckan db create-from-model
sudo -u postgres pg_dump -s ckan_default |tee >master-model.sql
git checkout integrate-vdm-stateful
paster --plugin=ckan db clean
paster --plugin=ckan db create-from-model
sudo -u postgres pg_dump -s ckan_default |tee >new.sql
diff master-model.sql new.sql
< -- Name: group_extra_revision_continuity_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: ckan_default
< --
<
< ALTER TABLE ONLY public.group_extra_revision
<     ADD CONSTRAINT group_extra_revision_continuity_id_fkey FOREIGN KEY (continuity_id) REFERENCES public.group_extra(id);
<
<
< --
1145,1152d1136
< -- Name: package_extra_revision_continuity_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: ckan_default
< --
<
< ALTER TABLE ONLY public.package_extra_revision
<     ADD CONSTRAINT package_extra_revision_continuity_id_fkey FOREIGN KEY (continuity_id) REFERENCES public.package_extra(id);
<
<
< --
```
Also, test is now fixed, so ready for your consideration again @TkTech

TkTech:
👍

amercader:
This looks good to me. If @tktech doesn't have any more feedback I'm happy to merge it

davidread:
Yes please @amercader

PR #4700, opened 2019-03-26
CLI. Implement `color` command

smotornyuk:


requirements.txt
```diff
@@ -58,6 +58,7 @@ tzlocal==1.3
 unicodecsv==0.14.1
 urllib3==1.24             # via requests
 vdm==0.14
+webcolors==1.8.1
```
amercader: Should this be in requirements.in

smotornyuk: Done. My bad, I didn't even know before about requirements.in file :)

wardi:
dropping this feature as discussed

PR #4713, opened 2019-04-04
Remove confusing line from source install instructions

mattfullerton:


davidread:
Agreed, this dir was created with `mkdir -p ~/ckan/etc` so should be owned by the current user already. The previous command changes the owner of the default subdir, so no need for this line that is removed in the PR.

Issue #4719, opened 2019-04-06
email verification is not complete


sivang:
### CKAN Version if known (or site URL)
2.8.2

### Please describe the expected behaviour
When entering an email address on a field that has `email_validator` , it's expected that email addresses of the form:
- ...this..is..@example.com
- this.is..@example.com
etc..

Will be rejected as they represent non existent email addresses, e.g. when using gmail to try and email such addresses, it's rejected as being malformed.

### Please describe the actual behaviour
Those email addresses are accepted with testing made on CKAN 2.8.2

### What steps can be taken to reproduce the issue? 
- Have `ckan-scheming` installed
- Modify an input field to have `email_validator` via dataset.json
- Go add a new dataset from the portal
- input aforementioned email address
- Witness as it gets saved.

amercader:
Fixed in #4720

PR #4720, opened 2019-04-06
enable email validator to exclude malformed dots usage in email addre…

sivang:
…sses, fixes github #4719

Fixes #

### Features:

- [ ] includes API changes


amercader:
@sivang if we clearly mark which lines in the regexp come from html.spec.whatwg.org and which don't this is good to go

sivang:
dots lookup pattern taken from here:
* https://stackoverflow.com/questions/40718851/regex-that-does-not-allow-consecutive-dots

PR #4721, opened 2019-04-08
Retrieving the package after plugins run

hardingalexh:
### Proposed fixes:
Currently when a resource is deleted, the following process happens in `resource_delete`:

1. The package is retrieved
2. The resource and list of resources is passed to the `IResourceController.before_delete` function
3. The package is updated

Because the package is not retrieved or edited after the `before_delete` function runs in plugins, this causes any changes made to the package in the `before_delete` function to be overwritten.

By re-fetching the package after the plugins have run, any changes made to the package by plugins in `before_delete` are retained.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/delete.py
+++ b/ckan/logic/action/delete.py
@@ -182,6 +182,8 @@ def resource_delete(context, data_dict):
         plugin.before_delete(context, data_dict,
                              pkg_dict.get('resources', []))
 
+    pkg_dict = _get_action('package_show')(context, {'id': package_id})
+
     if pkg_dict.get('resources'):
         pkg_dict['resources'] = [r for r in pkg_dict['resources'] if not
                 r['id'] == id]
@@ -193,7 +195,7 @@ def resource_delete(context, data_dict):
 
     for plugin in plugins.PluginImplementations(plugins.IResourceController):
         plugin.after_delete(context, pkg_dict.get('resources', []))
-
+    
     model.repo.commit()
 
 


smotornyuk:
In the past we had a lot of issues because of modifying packages inside before_update/delete. Generally, all actions that are modifying the package shouldn't be re-entrant. I.e, you mustn't call anything, that will require a future call of `package_show` inside before_* hooks. Yep, sometimes it necessary, but you should try to update `data_dict` passed to hooks instead. Otherwise, you may schedule some updates(via background workers, redis or just putting something into context) so that they happen after "native" action. And we are emphasizing this idea by not updating `pkg_dict` here. So, PR is ok from a technical point of view, but it shouldn't be accepted because of ideology if you have no some strong reasons for this change

hardingalexh:
Understand that reasoning. Here is my use case, to illustrate why this feels inconsistent:

I'm working on a CKAN site that involves a lot of datasets with genomics data. One feature we have added in an extension is that when a resource is uploaded, the file is parsed and gene names are extracted and stored in the dataset `extras` field(s) so that they become searchable. They are stored as `key: 'resource_id_partnum', value: 'list of genes'`. 

When a resource is deleted, we would like to be able to delete those extras from the package so they are no longer searchable. In `before_delete`, I get the package from the resource being deleted, remove any extras that have the resource id, then update the package to delete the extras I want deleted. However, because of the problem addressed in this MR, the package is then re-updated with all of the extras I deleted, so they are deleted then re-added. This behavior took a day to track down, because our code was successfully deleting the extras, but then somehow CKAN was re-adding them.

That said, there are ways around this (like doing it asynchronously on the queue), but it doesn't change the fact that this is odd/inconsistent behavior. If this MR is not accepted (which I understand given your reasoning), I would think that we should update documentation of these interfaces to note that any changes you make to the package will be reset by the framework so that developers will be aware of this behavior if they try and use this functionality.

smotornyuk:
I think it's a quite good use case so that PR can be merged. But if you, in addition, will provide another one with an update for documentation if will be fantastic :sparkles: 

Merging this PR.

PR #4727, opened 2019-04-12
Fix Unicode Decode Error in error page

Zharktas:
### Proposed fixes:
If error messages have latin characters, there error template fails to render them with the following log output.

```
File "/usr/lib/ckan/default/src/ckan/ckan/templates/error_document_template.html", line 8, in block "primary"
{{ content}}
 UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
```
This pr fixes the decoding of the variable.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -446,7 +446,7 @@ def _register_error_handler(app):
     def error_handler(e):
         log.error(e, exc_info=sys.exc_info)
         if isinstance(e, HTTPException):
-            extra_vars = {u'code': [e.code], u'content': e.description}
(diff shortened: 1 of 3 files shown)


Zharktas:
The test is apparently failing on master and is not related to this pr, build probably passes after #4726 is merged.

wardi:
Let's not pass bytes in to templates in the first place, would you move the decode to the controller/view instead?

Zharktas:
Ah, true.  I'll fix it some point.

Zharktas:
Apparently this only affected flask routes, pylons already had handling for different character encodings.

Zharktas:
#4829 fixes the same thing, but from different angle, so I'm closing this.

PR #4729, opened 2019-04-13
Update datapusher dockerfile

usingsky:
Fixes #
Update datapusher dockerfile

### Proposed fixes:
The docker image of the currently guided datapusher is not being updated.
I have experienced the following issue (https://github.com/ckan/ckan-service-provider/issues/36)
and it seems to have been fixed in 0.0.8 of ckan-service-provider.
So I suggest building a datapusher docker image via git.

### Features:

- [ ] includes API changes

### DIFF
--- a/contrib/docker/.env.template
+++ b/contrib/docker/.env.template
@@ -42,4 +42,6 @@ POSTGRES_PORT=5432
 # Readwrite user/pass will be ckan:POSTGRES_PASSWORD
 # Readonly user/pass will be datastore_ro:DATASTORE_READONLY_PASSWORD
 DATASTORE_READONLY_PASSWORD=datastore
-
+#
+# Image: datapusher
+DATAPUSHER_VERSION=0.0.15
--- a/contrib/docker/datapusher/Dockerfile
+++ b/contrib/docker/datapusher/Dockerfile
@@ -1,0 +1,30 @@
+FROM python:2.7-stretch
+
+WORKDIR /usr/src
+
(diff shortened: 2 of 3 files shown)


contrib/docker/datapusher/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM python:2.7-stretch
+
+WORKDIR /usr/src
+
+# Clone datapusher
+RUN git clone https://github.com/ckan/datapusher.git
```
amercader: We should probably pin the datapusher version and clone a specific tag here. WDYT?

contrib/docker/.env.template
```diff
@@ -42,4 +42,6 @@ POSTGRES_PORT=5432
 # Readwrite user/pass will be ckan:POSTGRES_PASSWORD
 # Readonly user/pass will be datastore_ro:DATASTORE_READONLY_PASSWORD
 DATASTORE_READONLY_PASSWORD=datastore
-
+#
+# Image: datapusher
+DATAPUSHER_VERSION=0.0.15
```
usingsky: In order to use docker-compose, we need to set the .env file. How about putting the datapuser version environment variable in the .env file?
The user may be able to change the datapusher version.

pdelboca:
Thanks for this PR! I'm closing it since it is no longer relevant.

We now have a separated docker repository in [ckan/ckan-docker](https://github.com/ckan/ckan-docker)

PR #4739, opened 2019-04-19
Upgrade sqlalchemy-migrate and pbr

davidread:
Upgrades:
* sqlalchemy-migrate 0.10.0->0.12.0 (latest)
* pbr 1.10.0->5.1.3 (latest)

These are upgraded for the sake of being up to date.

sqlalchemy-migrate [releases](https://github.com/openstack/sqlalchemy-migrate/releases). There doesn't seem to be an up-to-date changelog, but 0.11 is mentioned [here](https://allmychanges.com/p/python/sqlalchemy-migrate/). This sort of annoyance is another reason to move to Alembic - work started [here](https://github.com/ckan/ckan/pull/4450).

pbr was quite out of date. It "manages setuptools packaging", and I think it is only used by sqlalchemy-migrate.

I've tested this like this:
* tests pass, including with `--ckan-migration` option
* with a database from CKAN 2.3, upgrading to this version of CKAN, running `paster db upgrade`  and then manually clicking around web interface ok.

### Features:

- [ ] includes API changes

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,6 @@ flask-babel==0.11.2
 flask==0.12.4
 formencode==1.3.1         # via pylons
 funcsigs==1.0.2           # via beaker
-html5lib==1.0.1           # via bleach
 idna==2.7                 # via requests
 itsdangerous==0.24        # via flask
 jinja2==2.10.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -17,7 +17,6 @@ flask-babel==0.11.2
 flask==0.12.4
 formencode==1.3.1         # via pylons
 funcsigs==1.0.2           # via beaker
-html5lib==1.0.1           # via bleach
```
davidread: This change (like other changes) to requirements.txt was made by `pip-compile`. html5lib is no longer an explicit dependency because it is now included as part of the bleach package - see https://github.com/mozilla/bleach/issues/386

PR #4740, opened 2019-04-20
add_ckan_admin_tab function with icon

usingsky:
Fixes #
fix add_ckan_admin_tab function with icon

### Proposed fixes:
I tried to add the admin tab using the toolkit api.
But, I can not display tab label with icon.
I modified add_ckan_admin_tab function so that icon can be displayed together.

toolkit.add_ckan_admin_tab(config, 'admin.config', 'Test Menu') (ok)
toolkit.add_ckan_admin_tab(config, 'admin.config', 'Test Menu', icon='trash') (ok)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -933,8 +933,11 @@ def build_extra_admin_nav():
     admin_tabs_dict = config.get('ckan.admin_tabs')
     output = ''
     if admin_tabs_dict:
-        for key in admin_tabs_dict:
-            output += build_nav_icon(key, admin_tabs_dict[key])
+        for k, v in admin_tabs_dict.iteritems():
+            if v['icon']:
+                output += build_nav_icon(k, v['label'], icon=v['icon'])
+            else:
+                output += build_nav(k, v['label'])
     return output
 
 
--- a/ckanext/example_iconfigurer/tests/test_example_iconfigurer.py
+++ b/ckanext/example_iconfigurer/tests/test_example_iconfigurer.py
@@ -63,22 +63,22 @@ class TestExampleIConfigurerBuildExtraAdminTabsHelper(helpers.FunctionalTestBase
         response = app.get('/build_extra_admin_nav')
         nosetools.assert_equal(response.body, expected)
 
-    @helpers.change_config('ckan.admin_tabs', {'ckanext_myext_config_one': 'My Label'})
+    @helpers.change_config('ckan.admin_tabs', {'ckanext_myext_config_one': {'label': 'My Label', 'icon': None}})
     def test_build_extra_admin_nav_one_value_in_config(self):
         '''
         Correct string returned when ckan.admin_tabs option has single value in config.
         '''
         app = self._get_test_app()
-        expected = """<li><a href="/ckan-admin/myext_config_one"><i class="fa fa-picture-o"></i> My Label</a></li>"""
+        expected = """<li><a href="/ckan-admin/myext_config_one">My Label</a></li>"""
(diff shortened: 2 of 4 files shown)


ckanext/example_iconfigurer/tests/test_example_iconfigurer.py
```diff
@@ -63,22 +63,22 @@ def test_build_extra_admin_nav_config_option_present_but_empty(self):
...
         nosetools.assert_equal(response.body, expected)
 
-    @helpers.change_config('ckan.admin_tabs', {'ckanext_myext_config_one': 'My Label'})
+    @helpers.change_config('ckan.admin_tabs', {'ckanext_myext_config_one': {'label': 'My Label', 'icon': None}})
     def test_build_extra_admin_nav_one_value_in_config(self):
         '''
         Correct string returned when ckan.admin_tabs option has single value in config.
         '''
         app = self._get_test_app()
-        expected = """<li><a href="/ckan-admin/myext_config_one"><i class="fa fa-picture-o"></i> My Label</a></li>"""
```
smotornyuk: Why this icon is not shown anymore? Does this mean, that existing portals may lose some icons as well?

ckan/lib/helpers.py
```diff
@@ -933,8 +933,11 @@ def build_extra_admin_nav():
     admin_tabs_dict = config.get('ckan.admin_tabs')
     output = ''
     if admin_tabs_dict:
-        for key in admin_tabs_dict:
-            output += build_nav_icon(key, admin_tabs_dict[key])
+        for k, v in admin_tabs_dict.iteritems():
+            if v['icon']:
```
smotornyuk: Doesn't `icon` has `None` as default value, so you can just safetuly call  `build_nav_icon(k, v['label'], icon=v['icon'])` without any conditional clause?

usingsky: Thank you for review
I felt that I have to use them separately because there are two functions in the ckan core.
build_nav_icon function may change to a policy that will have a default image in the future.
I'm sorry I do not know the policy of the ckan team. :)
But, I think that if I do not set the icon, the icon should not be displayed.

ckanext/example_iconfigurer/tests/test_example_iconfigurer.py
usingsky: The add_ckan_admin_tab function is not used in ckan core and has no effect.
By default, when I use the "add_ckan_admin_tab" function in extension, the icon does not appear.
In test-case, ckan_icon is set in the mapper, and the icon comes out via named_route.
And as I said above, if I do not set the icon in the "add_ckan_admin_tab" function, I think it is correct to just display the label without the icon. ('icon': None)

ckan/lib/helpers.py
usingsky: case 1) icon1(ckan_icon) exists in mapper, call `add_ckan_admin_tabs(..., icon=icon2) `
=> icon1 or icon2?

case 2) icon1(ckan_icon) exists in mapper, call `add_ckan_admin_tabs(..., icon=None) `
=> icon1 or just label without icon
-> If icon1 should be visible, then I should call the function as follows.

`if v['icon']:`
`    output += build_nav_icon(k, v['label'], icon=v['icon'])`
`else:`
`    output += build_nav_icon(k, v['label'])`


I would really appreciate your feedback

smotornyuk: Good point, you are absolutely right here, so I'm going to merge this PR

PR #4752, opened 2019-04-26
In Python identity is not the same thing as equality

### LINKED ISSUE #4674: use ==/!= to compare str, bytes, and int literals
[flake8](http://flake8.pycqa.org) testing of https://github.com/ckan/ckan on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics__
```
./ckan/lib/helpers.py:192:18: F632 use ==/!= to compare str, bytes, and int literals
    if uargs and len(uargs) is 1 and isinstance(uargs[0], string_types) \
                 ^
./ckan/plugins/toolkit.py:478:12: F632 use ==/!= to compare str, bytes, and int literals
        if len(endpoint) is 1:
           ^
./ckan/tests/legacy/logic/test_action.py:705:16: F632 use ==/!= to compare str, bytes, and int literals
        assert count is 0
               ^
./ckan/tests/legacy/functional/test_pagination.py:13:8: F632 use ==/!= to compare str, bytes, and int literals
    if object_type is not 'group_dataset':
       ^
4     F632 use ==/!= to compare str, bytes, and int literals
4
```

cclauss:
Fixes #4674

### Proposed fixes:
__is__ --> ==
__is not__ --> !=

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -189,7 +189,7 @@ def redirect_to(*args, **kw):
     _url = ''
     skip_url_parsing = False
     parse_url = kw.pop('parse_url', False)
-    if uargs and len(uargs) is 1 and isinstance(uargs[0], string_types) \
+    if uargs and len(uargs) == 1 and isinstance(uargs[0], string_types) \
             and (uargs[0].startswith('/') or is_url(uargs[0])) \
             and parse_url is False:
         skip_url_parsing = True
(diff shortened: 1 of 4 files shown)


smotornyuk:
I know, that we are using small numbers here, but i don't think that numbers should be compared with `is` because this works only for smal rang(as i remember, only for one or two bytes).
if you check something like
```
def f(): return 101010                                                                                                                                                                                                                                                  
f() is 101010 
```
You'll get `False`

amercader:
@cclauss Thanks! Can we add the flag that catches these to the flake8 tests?

PR #4756, opened 2019-05-01
Fix to allow multiple tags in a single dataset search

hardingalexh:
Fixes #
In the current version of the dataset search screen, applied tag facets are not displayed past the first facet applied.

This is because the params/args present on the request are `MultiDict`, and they are being accessed using `items()`, which only returns the first value for each key. Passing in the `multi=True` arg to `items()` allows for multiple series for a given key, which solves this problem.

Previous Behavior (check URL)
Screen Shot 2019-04-30 at 2 39 33 PM

Fixed behavior:
Screen Shot 2019-05-01 at 2 40 14 PM

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1834,7 +1834,7 @@ def add_url_param(alternative_url=None, controller=None, action=None,
     '''
 
     params_nopage = [
-        (k, v)for k, v in request.params.items(multi=True) 
+        (k, v)for k, v in request.params.items(multi=True)
         if k != 'page'
     ]
     params = set(params_nopage)
@@ -1873,7 +1873,7 @@ def remove_url_param(key, value=None, replace=None, controller=None,
 
     params_nopage = [
         (k, v) for k, v in request.params.items(multi=True)
-         if k != 'page'
+        if k != 'page'
     ]
     params = list(params_nopage)
     if value:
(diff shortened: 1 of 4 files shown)


amercader:
Thanks for this @hardingalexh. We should add a test to prevent the bug from creeping up again in the future. Something inspired in [this one](thttps://github.com/ckan/ckan/blob/master/ckan/tests/controllers/test_package.py#L1510) should be easy to write.
Also if you check the CircleCI failures there is a PEP8 formatting error: 

```
-------------------
ckan/lib/helpers.py
-------------------
E501 ln:1836 line too long (88 > 79 characters)
E501 ln:1871 line too long (88 > 79 characters)
```

Do you know the CKAN version this bug was introduced?

hardingalexh:
Thanks @amercader - fixed the pep8 failure.

Not entirely sure when this was introduced, I've only been using CKAN since ~January and I believe the bug has been around since then. Could do some detective work to step back through this if you'd like.

amercader:
@hardingalexh don't worry, I'll dig out the version it was introduced. If you can write the test I'll merge this.

hardingalexh:
@amercader added a test to make sure that the correct number of tag pills show up for a filtered query. Not sure what's going on with the travis build - any thoughts there?

amercader:
Don't worry about the travis failures, these are unrelated. Great work @hardingalexh , thanks for your patch

Issue #4736, opened 2019-04-17
Upgrade sqlalchemy


davidread:
It would be good to upgrade sqlalchemy from 1.1.11 to the latest (1.3.2 currently).

Currently vdm complains with this new version:
```
ValueError: VDM only works with SQLAlchemy versions 0.4 through 0.9, not: 1.3.2
```
(see [Travis for branch upgrade-to-sqlalchemy-1.3.2](https://circleci.com/gh/ckan/ckan/5136) )

vdm won't be required soon, so upgrading sqlalchemy should be easy. Although it might be useful to fix vdm anyway, for the benefit of older installs of CKAN wanting to keep up to date.

davidlday:
In addition, SQLAlchemy 1.1.11 falls on the list of versions for a couple of vulnerabilities. See: 

* SQLAlchemy [Issue 4481](https://github.com/sqlalchemy/sqlalchemy/issues/4481)
* [CVE-2019-7548](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-7548)
* [CVE-2019-7164](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-7164)

davidread:
@davidlday Indeed, these are a mild concern, and tracked on our private security site. However we have a policy of not sign-posting security concerns in public, so forgive me for deleting this comment.

PR #4763, opened 2019-05-07
pool_pre_ping defaults to True

### LINKED ISSUE #4734: Disconnect handling
2.8.1, Docker

Random db disconnection and 502 errors.

Try to open `/datasetterms` or wait a while and open a dataset.

### Workaround

`ckan.datastore.sqlalchemy.pool_pre_ping = True`

frafra:
Fixes #4734

### Proposed fixes:
Set `pool_pre_ping = True` as default for sqlalchemy connections. This should make connections to the database more reliable.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -289,6 +289,7 @@ def update_config():
 
     # CONFIGURATION OPTIONS HERE (note: all config options will override
     # any Pylons config options)
+    config.setdefault('pool_pre_ping', True)
 
     # Initialize SQLAlchemy
     engine = sqlalchemy.engine_from_config(config)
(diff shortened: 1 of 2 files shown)


amercader:
Thanks, I didn't know about [pessimistic disconnect handling](https://docs.sqlalchemy.org/en/13/core/pooling.html#pool-disconnects-pessimistic). Seems like the small overhead that the liveliness query adds is worth the robustness gained.

But unless I'm reading the docs incorrectly the `pool_pre_ping` option [was added in SQLAlchemy 1.2](https://docs.sqlalchemy.org/en/13/core/engines.html#sqlalchemy.create_engine.params.pool_pre_ping), and for now we are pinned to 1.1.11. So I don't think this code does anything at the moment.
I'm closing this for now but happy to revisit it once SA is upgraded (https://github.com/ckan/ckan/issues/4736)

PR #4768, opened 2019-05-08
[WIP] Tag removal fix

hardingalexh:
Fixes #

### Proposed fixes:
Fixes trailing slashes on local URL generation.

Using CKAN latest with docker, top-level URLs are given a trailing slash (`/dataset/`, for example). This causes some bugs with URL generation, notably removing tags. In the below screenshot, you can see that the helper function for removing params from the url `/dataset/?tags=tag3` redirects to `/dataset/dataset/`. This is because the user is at `/dataset/?tags=tag3` instead of `/dataset?tags=tag3`. Adding the same conditional to remove the trailing slash as we do with the root URL fixes this problem, and successfully removes the tag rather than being redirected to `dataset/dataset` which returns 500.

Screen Shot 2019-05-08 at 2 20 59 PM


hardingalexh:
Looks like a number of tests have hardcoded references to URL strings with a trailing slash. This may be too broad strokes a solution to the bug - interested to hear feedback on this/reasoning for other solutions.

PR #4772, opened 2019-05-09
Allows external database host with docker containers

ctownsen357:
Fixes #
This allows one to use an external PostgreSQL database rather than requiring the docker ckan container to point to the `db` service within the compose file. In our case this allows us to point our containers to an AWS RDS instance. This change programmatically determines the database host based on the required `CKAN_SQLALCHEMY_URL` environment variable rather than leaving it hard coded to the compose service name.

This change also allows the system to fail fast if required environment variables are not set rather than waiting for the database to be available before testing the environment variables.
### Features:

- [ ] includes API changes


pdelboca:
Thanks! Closing as we do not maintain the docker file in this repository anymore: https://github.com/ckan/ckan-docker

PR #4774, opened 2019-05-10
Cache paths of non-existing templates to reduce open() syscalls

### LINKED ISSUE #4773: Number of `open()` syscalls during template rendering scales with extensions
Any with `ckan_extends` jinja extension, most visible in form pages with lots of extensions.

Opening pages causes only a sensible amount of `open()` syscalls.

Currently templates with common imports like `macros/form.html` can incur significant overhead in template loading due to `open()` syscalls if many extensions are used. In our case organization edit page did over 2000 `open()` calls, totaling 400ms added page load time. Most of the calls are checking every extension for overriding templates. 

Use python profiling tools to instrument organization edit page in a setup with dozens of extensions. Monitor the number of `{open}` calls and their cumulative duration when opening the page.

bzar:
Fixes #4773 

### Proposed fixes:
- Cache template paths pointing to non-existing templates and skip trying to open them again
- Makes it impossible to add new template files on the fly, but is this a necessary feature?

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/jinja_extensions.py
+++ b/ckan/lib/jinja_extensions.py
@@ -19,8 +19,12 @@ import ckan.lib.base as base
 import ckan.lib.helpers as h
 from ckan.common import config
 
+from werkzeug.local import Local
+
 
 log = logging.getLogger(__name__)
+local = Local()
+local.missing_templates = set()
 
 
 def get_jinja_env_options():
@@ -195,8 +199,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         pieces = loaders.split_template_path(template)
         for searchpath in searchpaths:
             filename = path.join(searchpath, *pieces)
+            if filename in local.missing_templates:
+                continue
             f = open_if_exists(filename)
             if f is None:
+                local.missing_templates.add(filename)
                 continue
             try:
                 contents = f.read().decode(self.encoding)


bzar:
Seems there are still issues with this solution. I'll add commits here soon.

amercader:
That sounds like a good optimization @bzar.

Just trying to understand these limitation better:

> Makes it impossible to add new template files on the fly, but is this a necessary feature?

So this means that if I request say `/dataset/<id>`, which uses `package/read.html` and I have these extensions:

```
ckanext/myown/templates/
ckanext/pages/templates/package/read.html
```

If I then add a new template:
```
ckanext/myown/templates/package/read.html
ckanext/pages/templates/package/read.html
```
I will get the old version from pages because the one on myown will get skipped (unless I restart the server)

Is that correct?

bzar:
As far as I can tell, yes, assuming you've opened the package page between them, so the optimization has cached the template's non-existence.

There is a lighter version of this optimization, where we only cache the missing files per request. That would fix the massive amounts of `open()` we're seeing, but the number would still grow linearly with the number extensions, just not multiplied by the number of references as it is now.

amercader:
@bzar I think the request-based optimization you mention is a little bit safer. I really want to avoid people not understanding why the templates are not refreshing when reloading the browser, so I think this makes more sense even if at the expense of some performance.

bzar:
Changed the caching to request level. Not exactly sure if this is the correct way to store request level data though. Performance-wise it's somewhere between no caching and thread level caching:

caching | first load open calls | nth load open calls
----------|--------------------------|----------------------------
none | 2300 | 2300
request-level | 602 | 570
thread-level | 602 | 83

bzar:
Also it seems the patch breaks tests since there's no request context available. I would welcome input on where I should stash the request-scoped missing templates to be consistent and compatible with the existing code before trying again.

PR #4775, opened 2019-05-10
Update sqlalchemy to 1.3.3

frafra:
Fixes: #4736 
Related to: https://github.com/ckan/ckan/pull/4763#issuecomment-490235987


frafra:
https://github.com/okfn/vdm/ seems unmaintained. There is a branch for sqlalchemy 1.1, but not for 1.2.

amercader:
Thanks @frafra. Should we aim for SQLAlchemy 1.3 if we are upgrading?
And yes VDM is unmaintained but it will need to be updated to support SA 1.3 otherwise it will throw an error.
I've just uploaded vdm==0.15 to Pypi which should allow to run on SA 1.3. Can you update the requirement on `requirements.in` and `requirements.txt` and see how the tests look?

frafra:
Updated to 1.3.3.

amercader:
@frafra there is work being done in the same front by @iamarnavgarg on #4782 
Is it ok if I close this PR?

frafra:
Sure!

PR #4781, opened 2019-05-14
Add cache control headers to flask

Zharktas:
### Proposed fixes:
Cache-Control headers were not implemented for flask. This ports the caching logic for flask.

Most pages now days include some extra variables, so the old logic pretty much prevented caching. The only page I could find was about-page https://demo.ckan.org/about which was cacheable. This PR replaces extra variables chech with config option ckan.cache_enabled. This setting needs to be true for any caching to happen. If it is false or unset, the caching is the same as previous logic, as all most nothing was cached.

Documentation was changed as ckan.cache_enabled now has extra functionality.

### Features:

- [ ] includes API changes


ckan/lib/base.py
```diff
@@ -232,6 +230,44 @@ def set_pylons_response_headers(allow_cache):
     return render_template
 
 
+def _allow_caching(extra_vars, cache_force=None):
```
pdelboca, line 233: `extra_vars` is not needed as a parameter since we are not using it in the code.

Zharktas:
Updated to current master and converted new tests to pytest.

pdelboca:
Thanks @Zharktas!

doc/maintaining/configuration.rst
```diff
@@ -418,7 +418,7 @@ Example::
 
 Default value: ``None``
 
-Controls if we're caching CKAN's static files, if it's serving them.
+This sets ``Cache-Control`` header for unauthenticated users.
```
amercader, line 421: We should expand this to describe the actual behaviour. IIUC correctly if `ckan.cache_enabled = True`, the header `Cache-Control: public` will be set on requests where the user is not logged in.

Is there a reason why the default value is `None` and not `False`? This seems to be the case [here](https://github.com/ckan/ckan/pull/4781/files#diff-0aa5b9638fa87bb9c7fa9bca15b2c0dbR254)

ckan/views/__init__.py
```diff
@@ -96,6 +96,28 @@ def set_cors_headers_for_response(response):
...
 
+def set_cache_control_headers_for_response(response):
+
+    # no_cache should be None when caching is allowed
+    allow_cache = request.environ.get(u'__no_cache__')
+
+    if u'Pragma' in response.headers:
+        del response.headers["Pragma"]
+
+    if allow_cache is None:
```
amercader, line 107: I find this really confusing. I get that if `__no_cache__` is not on `request.environ` we should allow caching but maybe we can define this as
```python
allow_cache = request.environ.get(u'__no_cache__') is None
```
or

```python
allow_cache = u'__no_cache__' not in request.environ
```

so `allow_cache` is actually `True`

ckan/lib/base.py
```diff
@@ -233,6 +231,44 @@ def set_pylons_response_headers(allow_cache):
     return render_template
 
 
+def _allow_caching(cache_force=None):
+    # Caching Logic
+
+    allow_cache = True
+    # Force cache or not if explicit.
+    if cache_force is not None:
```
shubham-mahajan, line 239: @Zharktas What can be the case when the `cache_force` will not be None? 

In the case of `pylons`, it is mention in the method `_pylons_prepare_renderer`

amercader:
@Zharktas @shubham-mahajan I made the small review changes in 5948799d1. Let me know if you spot something odd.

ckan/lib/base.py
amercader: @shubham-mahajan I guess the only case would be if it's passed explicitly from another function in the code. I guess it can be removed but it can be done in future refactorings.

ckan/views/__init__.py
```diff
@@ -96,6 +96,28 @@ def set_cors_headers_for_response(response):
...
+        response.cache_control.public = True
+        try:
+            cache_expire = int(config.get(u'ckan.cache_expires', 0))
+            response.cache_control.max_age = cache_expire
+            response.cache_control.must_revalidate = True
+        except ValueError:
+            pass
+    else:
+        response.cache_control.private = True
```
ThrawnCA, line 116: Private browser-only caching is good for security, but it is not good enough for templates that need to be reloaded every time and should not be cached at all, such as `_logout`.

PR #4784, opened 2019-05-17
Remove VDM

### LINKED ISSUE #4783: Remove vdm
Now that the revisioning has been removed and activities expanded (#3972) removing VDM is the single major blocker to the Python 3 migration. We can't start working on running CKAN in Python 3 until it is removed, as it is used at a very low level and very early in the initialization code.

Current PR: https://github.com/ckan/ckan/pull/4784 (and @davidread previously did some exploratory work on https://github.com/ckan/ckan/pull/4664/, and @TkTech showed [interest](https://github.com/ckan/ckan/pull/3972#issuecomment-465523145).

The following issues have been flagged:

Observations:

* `user_dict.number_of_edits` is removed, as calculating that relied on revisions
* `stats` extension loses all the stats time-line, as calculating those relied on revisions. could be resurrected with activitie stream data.
* trash page removed (no revisions to trash)
* `migrate_filestore` won't work any more

The items above all seem fine to lose for me, including the stats extension which hasn't had any development in ages.

Not working:
* fix up activities - they are not being created for some reason
* migration

These need to be fixed as part of the work.

#### Approach

@davidread covered all the things involved in the actual removal of VDM and the revisioning: low level stuff when initializing the database, revision based columns and methods on the models, removing explicit revision creation calls, etc.

The second part of the work will be of course to make everything work as expected, fixing the activities problems, migration etc.

#### Story points

13

davidread:
Fixes #4783

### Proposed fixes:

* Removes vdm dependency from requirements.txt.
* Removes all mention of revisions.
    * deleted: `model.repo.new_revision()`
    * deleted revision_id columns - they have no use
* We still need some vdm and revision code for migrating from CKAN 2.8, so this is moved into `ckan/migration/revision_legacy_code.py`

Breaking changes:
* Actions like`package_show` don't return the `revision_id` key any more
* `user_dict.number_of_edits` is removed, as calculating that relied on revisions. `user_list` can no longer take parameter value `order_by=edits`
* stats extension loses all the stats time-line, as calculating those relied on revisions. could be resurrected with activitie stream data.
* migrate_filestore won't work any more (was v. old anyway)
* `revision_list`, `package_revision_list`, `group_revision_list`, 'organization_revision_list', `revision_show` actions is removed

To do:
* [x] fix up activities - they are not being created for some reason
* [x] migration
* [ ] check the 

(There was a previous PR for this same branch: https://github.com/ckan/ckan/pull/4664, but that was to merge the work into branch 3484_revision_ui_removal2, which was merged into master, so the PR now looks screwy, hence this fresh PR.)

### Features:

- [ ] includes API changes


davidread:
@amercader Help appreciated! I've not had an opportunity to work on this for a few months as you can see.

Much of the work is done, however I got stuck though thinking about how to do this line: https://github.com/ckan/ckan/pull/4784/files#diff-c0190fdd2bf6a4b258769b230dbb5396R48 Now that we don't have a revision which stores the user who is making a change, I'm not sure how to get this to put in the activity stream. I had thoughts about storing the user in the session, but maybe there is a simple solution... Anyway if you can crack that, most of the tests should pass and we should be pretty much there.

amercader:
An update on this PR. I started looking at the blocker mentioned above by @davidread and after some hours of VDM hell came up with a potential solution (https://github.com/ckan/ckan/commit/8c02842ebf60cd481ad1b0ad4f2d32b8a0c44b1b). Sadly the tests are still broken due to state issues with the legacy tests (all failing modules pass locally if run individually). I'll keep working on this.

davidread:
@amercader did you decide against that https://github.com/ckan/ckan/commit/8c02842ebf60cd481ad1b0ad4f2d32b8a0c44b1b change in the end?

amercader:
@davidread TL;DR:

* Yes, I went for a more explicit approach creating activities in the relevant actions
* I'm blocked on test failures arising from changes in `ckan/migration/revisions_legacy_code.py`


After discussion with @wardi and the rest of the team it felt like explicitly creating activities in the relevant action rather than relying on a SQLAlchemy session extension simplified things and reduced the amount of magic involved. I guess that the session extension approach made sense when we were creating more granular activities (like " added extra X", or "modified resource Y"), after your big activities refactor we only emit activities at the dataset level (ie all the previous ones are "dataset Z was updated").

Besides my changes introduced different test failures which suggested it was not robust enough.

I started the new approach on the `remove-vdm-5` branch (sigh). This commit has the gist of it https://github.com/ckan/ckan/commit/c003166ac4f2071d45696e3edd37a2f9ed143e21, but there were fixes before (this is a working branch, I'll clean everything up when we get out of this mess).

Sadly I've been battling for the last couple of weeks with a major blocker. There are test failures caused by the code used in `ckan/tests/migrations` so perhaps you can help me out on this one.

Basically it seems like the modifications done to the model in [revisions_legacy_code.py](https://github.com/ckan/ckan/blob/c003166ac4f2071d45696e3edd37a2f9ed143e21/ckan/migration/revision_legacy_code.py) disrupt the model somehow causing failures later on. This causes [two types a failures](https://circleci.com/gh/ckan/ckan/5814#tests/containers/1):

* Follower count related failures
* Create_test_data complaining that their datasets already exist (these might be caused by the former)

I used the following to reproduce it locally on `remove-vdm-5`:

```
nosetests --reset-db --ckan --nologcapture --with-pylons=test-core.ini -s -v ckan/tests/migration/test_revision_legacy_code.py ckan/tests/legacy/functional/api/test_follow.py:TestFollowerDelete
```
This fails with the following:

```
======================================================================
ERROR: test suite for <class 'ckan.tests.legacy.functional.api.test_follow.TestFollowerDelete'>
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/tests/legacy/functional/api/test_follow.py", line 823, in setup_class
    self.warandpeace['id'], self.testsysadmin['apikey'])
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/tests/legacy/functional/api/test_follow.py", line 42, in wrapped_func
    func(app, follower_id, apikey, object_id, object_arg, sysadmin_apikey)
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/tests/legacy/functional/api/test_follow.py", line 182, in follow_dataset
    assert follower_count_after == follower_count_before + 1
AssertionError

Ran 1 test in 36.160s

FAILED (errors=1)
```

I dug into the [follower model](https://github.com/ckan/ckan/blob/c003166ac4f2071d45696e3edd37a2f9ed143e21/ckan/model/follower.py) and after an unhealthy amount of debugging I found before the failing API call, that at this point:

https://github.com/ckan/ckan/blob/c003166ac4f2071d45696e3edd37a2f9ed143e21/ckan/model/follower.py#L52

```python
cls._get_followers(object_id).count() == 2   # Wrong
```

but 

```python
len(cls._get_followers(object_id).all())  == 1  # Right
```

This makes no sense of course, so I investigated a bit more and found the culprit a bit further down the stack:

https://github.com/ckan/ckan/blob/c003166ac4f2071d45696e3edd37a2f9ed143e21/ckan/model/follower.py#L76

`cls._object_class()` at this point is `ckan.model.Package`. But when aliasing, the following query is generated below:

```
SELECT user_following_dataset.follower_id AS user_following_dataset_follower_id, user_following_dataset.object_id AS user_following_dataset_object_id, user_following_dataset.datetime AS user_following_dataset_datetime, user_1.password AS user_1_password, user_1.id AS user_1_id, user_1.name AS user_1_name, user_1.fullname AS user_1_fullname, user_1.email AS user_1_email, user_1.apikey AS user_1_apikey, user_1.created AS user_1_created, user_1.reset_key AS user_1_reset_key, user_1.about AS user_1_about, user_1.activity_streams_email_notifications AS user_1_activity_streams_email_notifications, user_1.sysadmin AS user_1_sysadmin, user_1.state AS user_1_state, package_1.id AS package_1_id, package_1.name AS package_1_name, package_1.title AS package_1_title, package_1.version AS package_1_version, package_1.url AS package_1_url, package_1.author AS package_1_author, package_1.author_email AS package_1_author_email, package_1.maintainer AS package_1_maintainer, package_1.maintainer_email AS package_1_maintainer_email, package_1.notes AS package_1_notes, package_1.license_id AS package_1_license_id, package_1.type AS package_1_type, package_1.owner_org AS package_1_owner_org, package_1.creator_user_id AS package_1_creator_user_id, package_1.metadata_created AS package_1_metadata_created, package_1.metadata_modified AS package_1_metadata_modified, package_1.private AS package_1_private, package_1.state AS package_1_state, package.revision_id AS package_revision_id 
FROM user_following_dataset, "user" AS user_1, package AS package_1, package 
WHERE user_1.id = user_following_dataset.follower_id AND user_following_dataset.follower_id = user_1.id AND user_following_dataset.object_id = package_1.id AND user_1.state != %(state_1)s AND package_1.state != %(state_2)s AND package_1.id = %(id_1)s ORDER BY user_1.name

```

Note that the `package` table is referenced twice (`FROM user_following_dataset, "user" AS user_1, package AS package_1, package`) and that `revision_id` is requested from the non-aliased table (` package.revision_id AS package_revision_id `).

This suggest a link to the changes in the legacy revision model that the tests in `ckan/tests/migration` use but I've been incapable of finding a way around it. We are messing with the metadata adding new tables and columns to existing ones, so obviously there's some state leaking on the SQLAlchemy definitions. Skipping the tests in this folder makes all the rest pass.

I'd appreciate another pair of eyes and your thoughts as you are much more familiar with the legacy revision code and VDM by now.

Sorry for the long answer, I wanted to give as much context as possible. If you can

davidread:
I'm as foxed as you on this. However I don't think we care enough. It's just one test of code that is only used in migration. How about we just move it out of the regular tests and run it on its own? I've done this:

https://github.com/ckan/ckan/commit/4ec0ae36377f2e65ca320bd2fc151fbfdf701424

If that's acceptable, then I think my only remaining concern is to ensure that all the actions that should produce activities actually are.

amercader:
@davidread I think what you suggest is reasonable, let's not waste more time trying to get these tests to play nice.

Regarding not missing any activities, I'm pretty sure we aren't but I'll check the existing tests to see if they cover all cases.

Another question, is there a reason to keep the `revision_id` columns in the models? Shall we remove them to avoid confusion in the future?

amercader:
Forgot to say that I'll clean the working branch and rebase into this one.

wardi:
+1 on removing the revision_id values scattered everywhere (especially for things like extras and tags) 

For packages and other things that are tracked in activities, thoughts on storing the activity id instead?

amercader:
@davidread I added some further tests in 
9eba882  to make sure we are creating activities on different action, like add resource, update resource, add extra, etc.

I also did a final review of the code and as far as I'm concerned this is good to go once the migration script for the `revision_id` changes is done. I can help with that.
What should the migration script contain? I assume just dropping the revision_id columns and the foreign keys associated, but not dropping the actual tables right?

davidread:
@amercader thanks for all that.

> For packages and other things that are tracked in activities, thoughts on storing the activity id instead?

Might be nice, but I'm not sure I have the will to do more on this...

On the migration script, I started working out how to do this - https://github.com/ckan/ckan/pull/5047 - and I've done an initial try, just committed. So I've added to the alembic env.py the revision model, so that the auto-generated migration doesn't delete the tables. I've removed revision_id etc columns on the revision model - they are not useful afaict but I've not checked a migration still works, so that's todo.

The auto-generated migration drops the revision_id columns, but then is full of loads of other stuff which I don't recognise.

tino097:
Great job here @davidread and @amercader :fireworks: :tada:

amercader:
Thanks for your hard work on this one @davidread!

I think there are further cleanups we can do but let's start afresh.

@wardi your suggestion of including `activity_id` to the relevant object dict sounds really useful. What would be the best way to do that efficiently? At dictization time (one extra db call)?

davidread:
Phew! And thanks for pairing up on this to get it done @amercader 🙌

Chealer:
Thank you David,
As this causes functional changes, could something be added to CHANGELOG.rst as well? I don't understand very well everything in your list, but I am familiar with the stats extension, and I think at least the changes there warrant mention. The 2 options/statistics removed are the first 2. We use statistics on our catalogue, and in a few weeks these 2 options will be the only ones we will offer, which means we could end up with an empty page after upgrade.

Zharktas:
Changelogs are typically updated during the release process, so removing revisions would appear there after 2.9 release process is started https://docs.ckan.org/en/2.8/contributing/release-process.html. Sometimes the changelog is also updated before hand, but it will only be in the latest docs https://docs.ckan.org/en/latest/changelog.html, for example removing revisions is there already:
image

Chealer:
I am not convinced that is the most efficient and reliable process, but I see, thanks @Zharktas .

As a minimum [the stats extension's documentation](https://github.com/ckan/ckan/blob/master/doc/maintaining/stats.rst) should be updated to cope.

davidread:
@Chealer Yeah, the PR was a lot of work, and the priority was to merge rather than to tie up every loose end like this neatly, I'm afraid.

I think you make a good point about stats extension and encourage you to submit the PR to explain the change. There might also be a good argument to deprecate the stats extension entirely, now that the graphs with the main interest have been removed. But I'm not really involved right now to pursue this conversation myself.

Chealer:
I suppose the database still allows computing the rate of dataset creations? What would be missing is information to determine the rate of dataset deletions, right?

amercader:
Deleted datasets are kept in the database with state `deleted` until purged, but in any case there is an activity created for both deletion and creation so that's what a new version of the stats extension could use to calculate the necessary data.

Chealer:
Thank you @amercader 

Is there a reason why such a new version was not added? Could performance be expected to be on par with the 2.8 version?

PR #4788, opened 2019-05-22
Add support for base64 images for organization & group image

jeanpommier:
Fixes #

### Proposed fixes:
Add the possibility to enter the image for group or organization as base64 encoded data
(such data needs to be returned as-is to the templates, without pre-appending an url path)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -388,7 +388,7 @@ def group_dictize(group, context,
 
     image_url = result_dict.get('image_url')
     result_dict['image_display_url'] = image_url
-    if image_url and not image_url.startswith('http'):
+    if image_url and not image_url.startswith('http') and not image_url.startswith('data:'):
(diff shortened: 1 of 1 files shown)


ckan/lib/dictization/model_dictize.py
```diff
@@ -388,7 +388,7 @@ def get_packages_for_this_group(group_, just_the_count=False):
 
     image_url = result_dict.get('image_url')
     result_dict['image_display_url'] = image_url
-    if image_url and not image_url.startswith('http'):
```
wardi: this line of code was already wrong (`startswith('http')` is not how you check for a valid url).  Would you change this to use an existing library to check for a valid url instead?

jeanpommier: Yes, sure, I'll have a look at it.

jeanpommier: Does it suit you this way ?

PR #4790, opened 2019-05-22
fix remove_field in dateset page

### LINKED ISSUE #4823: After removing the tag the redirect web url is wrong and it causes an error
Hi, thanks for the wonderful CKAN. It makes me set up a open data portal very fast. 
When I search the data by tags, it works very quickly.
图片

But I am trying to remove the tags like the picture below
图片

 the redirect web url became` http://10.102.30.154:5000/en/dataset/dataset?tags=%E6%B5%8B%E8%AF%95`. The url has an extra `dataset`, which causes an error in the UI.

图片

图片

usingsky:
Fixes #4823
An error occurs when generating the url to delete the facet field.

image

### Proposed fixes:
fix remove_field function

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -84,13 +84,15 @@ def drill_down_url(alternative_url=None, **by):
 
 
 def remove_field(package_type, key, value=None, replace=None):
+    if not package_type or package_type == u'dataset':
+        url = h.url_for(u'dataset.search')
+    else:
+        url = h.url_for(u'{0}_search'.format(package_type))
     return h.remove_url_param(
         key,
         value=value,
         replace=replace,
-        controller=u'dataset',
-        action=u'search',
-        alternative_url=package_type
+        alternative_url=url
     )
 
 


ckan/views/dataset.py
```diff
@@ -83,14 +83,12 @@ def drill_down_url(alternative_url=None, **by):
...
-def remove_field(package_type, key, value=None, replace=None):
+def remove_field(key, value=None, replace=None):
     return h.remove_url_param(
         key,
         value=value,
         replace=replace,
-        controller=u'dataset',
-        action=u'search',
-        alternative_url=package_type
+        alternative_url=h.url_for(u'dataset.search')
```
smotornyuk: It won't work for custom dataset types, as i understand. This is the reason, why there was `package_type` argument

usingsky: Thank you for your explanation.
In my opinion, the root cause is likely to be due to the / at the end of the url path, unlike the previous release version. (e.g. localhost/dataset "**/**" , localhost/group "**/**" ....)

1. localhost/dataset **/**
2.  select facet -> localhost/dataset **/** ?res_format=csv
3.  remove facet ->  localhost/dataset **/** ?res_format=csv
4.  h.remove_url_param(alternative_url='dataset') ->  localhost/dataset **/** dataset

I would like to ask you if you know the relevant history.

PR #4810, opened 2019-05-24
Add test that shows definition of dataset_dict

davidread:
I think it is really useful having a test showing what a dataset_dict looks like.

### Features:

- [ ] includes API changes


davidread:
Let's delete this again if it breaks say 3 times in the next couple of months.

PR #4812, opened 2019-05-24
Upgrade PyYaml

davidread:
* PyYaml needed upgrading due to GitHub's security nag.
* PyYaml is used by webassets and it seems fine with the latest version, according to:
  https://github.com/miracle2k/webassets/blob/e3e82114324ffd6cf1a2877976a1de08c515eb10/requirements-dev.pip
* PyYaml was only just introduced to CKAN in webassets PR https://github.com/ckan/ckan/pull/4614
  However it was added to requirements.txt not requirements.in, so I fixed that too.
* Regenerating requirements.in adds a few minor things along the way.

CC @smotornyuk 

### Features:

- [ ] includes API changes

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -20,6 +20,7 @@ Pylons==0.9.7
 python-dateutil>=1.5.0
 pytz==2016.7
 pyutilib.component.core==4.6.4
+pyyaml  # needed by webassets. latest should be fine.
 repoze.who-friendlyform==1.0.8
 repoze.who==2.3
 requests==2.22.0
@@ -32,6 +33,7 @@ sqlparse==0.2.2
 tzlocal==1.3
 unicodecsv>=0.9
 vdm==0.14
+webassets==0.12.1
 WebHelpers==1.3
 WebOb==1.0.8
 WebTest==1.4.3  # need to pin this so that Pylons does not install a newer version that conflicts with WebOb==1.0.8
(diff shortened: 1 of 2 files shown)


davidread:
There is a warning added when running tests:
> /usr/lib/ckan/default/local/lib/python2.7/site-packages/webassets/loaders.py:162: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
>   obj = self.yaml.load(f) or {}

I'm not concerned because there is no user-created YAML being parsed. Hopefully webassets will fix this soon.

rufuspollock:
Discussed whether we should pin to an exact version with @davidread and sense was that it was ok to not have exact pinning for this one.

davidread:
In the discussion it was noted that we do pin everything in requirements.txt, but not necessarily in requirements.in.

The point being that pinning requirements.txt is helpful for consistency between everyone's installs, reproducing the same behaviour. But having requirements.in looser allows us to keep stuff more up to date - otherwise deps never get updated until a CVE comes out and we have a lot of work to do to catch up. I quite like the current balance where every couple of months or so we regenerate requirements.txt and get a batch of library updates and are pushed to sort out any little compatibility issues. And this is aided by the looser requirements.in file.

Issue #4813, opened 2019-05-25
Fix incorrect 'with_capacity' assignment in obj_list_dictize


LukeTully:
**Note:** Are you submitting a security related issue that could be a potential vulnerability? Please send it to security@ckan.org instead.

### CKAN Version if known (or site URL)
2.8.2 (Tested in 2.7.4 but the code hasn't changed)

### Please describe the expected behaviour
I don't know what the purpose of 'with_capacity' is exactly. It's not well documented and the code is old. The purpose of this line is to pass it along to the result_dict if it's present in context, but the current code neglects to actually retrieve the capacity value from the obj.

### Please describe the actual behaviour
This StackOverflow post by another user described the error
https://stackoverflow.com/questions/52992550/value-error-too-many-values-to-unpack-in-an-object-forloop
### What steps can be taken to reproduce the issue? 
I first encountered this issue when I was testing some code on a production instance in dev mode, but encountered a dataset that had been deleted from the db but not the index. That led behavior to occur that matched that of a package not existing in cache from what I can tell.

PR #4814, opened 2019-05-25
Fix assignment in obj_list_dictize

LukeTully:
with_capacity was incorrectly retrieved from its obj

Fixes https://github.com/ckan/ckan/issues/4813

### Proposed fixes:
Minimal fix to successfully retrieve the with_capacity value and pass it along

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/dictization/__init__.py
+++ b/ckan/lib/dictization/__init__.py
@@ -79,7 +79,7 @@ def obj_list_dictize(obj_list, context, sort_key=lambda x:x):
 
     for obj in obj_list:
         if context.get('with_capacity'):
-            obj, capacity = obj
+            capacity = context.get('with_capacity')
             dictized = table_dictize(obj, context, capacity=capacity)
         else:
             dictized = table_dictize(obj, context)


rufuspollock:
@wardi thinks this code path is not used anymore and we could just delete.

wardi:
yes let's remove this code instead of fixing it and adding tests

LukeTully:
@wardi In what sense is it not used, or in what version did the path stop being followed? Certainly I'm in support of it being removed if it's not used going forward, it was a little tricky to understand exactly how what it was for, but it was ultimately being hit in 2.7.

wardi:
I'm having trouble understanding this code as well, but it looks like this flag in the context object (shared mutable state that leaks from one part of code to the next) is supposed to change the parameter to obj_list_dictize to make it (obj, context) tuples instead of a simple list of objects. Why is capacity a special value that needs to be returned in a tuple instead of along with the objects? I don't know, but it's a really fragile way to write code. The same pattern is repeated all over model_dictize.py functions.

Doesn't look like it's safe to just remove, but the change you're suggesting will only copy the `with_capacity` True/False value into every dict returned, which is clearly not useful.

I suspect we have a context-object-with_capacity value leaking somewhere. Can you post your full traceback? If we can't clean up this code that might help us fix the problem you're seeing at least.

LukeTully:
I have an idea for replicating the stack trace and might be able to test that tonight. Fixing this bug solved my issue, and as per your guess, allowed me to find the real problem. This code was hit when a package existed in index but not the database if memory serves. Someone had created a dataset in draft and then deleted it, and for whatever reason wasn't cleared from index so when show_package was called everything blew up. 

Once I started looking into this a bit, I had the same thoughts regarding fragility of the code and otherwise that you mention, but upon stopping myself from yak shaving it seemed that this was just a typo given the context for which it was assigned and subsequently passed along. There is probably a leak of the value, I'm not sure if I was able to track down the origin or not

PR #4825, opened 2019-05-31
Use TextField instead of StrField due to hardcoded size limit in solr 6

Zharktas:
Using Solr6 indexing might fail with following log:

```
ckan.lib.search.common.SearchIndexError: Solr returned an error: (u'Solr responded with an error (HTTP 400): [Reason: Exception writing document id 4fb2245b8ebf2110d52eaf1899f72f8f to the index; possible analysis error: Document contains at least one immense term in field="coupled-resource" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped.  Please correct the analyzer to not produce such terms.  The prefix of the first immense term is: \'[91, 123, 34, 104, 114, 101, 102, 34, 58, 32, 91, 34, 104, 116, 116, 112, 58, 47, 47, 119, 119, 119, 46, 112, 97, 105, 107, 107, 97, 116]...\', original message: bytes can be at most 32766 in length; got 35362. Perhaps the document has an indexed string field (solr.StrField) which is too large]',)
```
image

### Proposed fixes:
Change StrField to TextField which does not have similar hardcoded limit.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/solr/schema.xml
+++ b/ckan/config/solr/schema.xml
@@ -27,7 +27,7 @@ schema. In this case the version should be set to the next CKAN version number.
 <schema name="ckan" version="2.9">
 
 <types>
-    <fieldType name="string" class="solr.StrField" sortMissingLast="true" omitNorms="true"/>
+    <fieldType name="string" class="solr.TextField" sortMissingLast="true" omitNorms="true"/>
     <fieldType name="boolean" class="solr.BoolField" sortMissingLast="true" omitNorms="true"/>
     <fieldtype name="binary" class="solr.BinaryField"/>
     <fieldType name="int" class="solr.TrieIntField" precisionStep="0" omitNorms="true" positionIncrementGap="0"/>


TkTech:
`StrField` and `TextField`do not have the same behaviour or storage characteristics. Are we sure we want to make this change? We should never be storing that much text in a string (which should used for things like keyword fields or other exact-matches)

Should we also make this change in the `multilingual/solr.xml` if we do?

Zharktas:
I think that the problem arises with catch all field https://github.com/ckan/ckan/blob/master/ckan/config/solr/schema.xml#L174 and use of spatial and it's xml fields, which get indexed, but i did not verify this yet.

amercader:
I don't think changing the type of a particular field type is the right fix. We already have a `TextField` field type. If there is something that gets indexed as string but it should be indexed as text then we need to change the field itself or the indexing logic.

@Zharktas if the problems come from a particular set of extras I guess you can remove them in `before_index` so they never get to Solr in the first place.

Zharktas:
We stumbled on this again while updating our Solr to newer version and experimented a bit and it seems that it would be enough to change the catch all fallback filed to TextField instead of all string fields.

Meaning this one https://github.com/ckan/ckan/blob/c7e946c0b144c1de7157b0cc3f6414ec8927e4c8/ckan/config/solr/schema.xml#L175

Any opinion on this or should the before_index function updated to remove large fields ?

Zharktas:
Looked into this again and it's not actually coordinate data, the error message just encodes it. In our case there are many URL's in the offending data and the correct solution would be removing it during harvesting. I'll close this and document in to https://github.com/ckan/ckan/wiki/Solr-and-search that there are limits of 32kb per field.

PR #4828, opened 2019-06-03
[#4827] More robust auth functions for resource_view_show

### LINKED ISSUE #4827: More robust auth functions for resource_view_show
Right now it relies on resource objects being present in the context. You should be able to call the auth function with the same parameters as the action (ie just the resource view `id`). This is not an issue in core but it can be problematic when extending auth from extensions.

amercader:
Fixes #4827.

Right now they rely on resource objects being present in the context. You should be able to call the auth function with the same parameters  as the action (ie just the resource view id). This is not an issue
 in core but it can be problematic when extending auth from extensions.

### DIFF
--- a/ckan/logic/auth/get.py
+++ b/ckan/logic/auth/get.py
@@ -163,11 +163,29 @@ def resource_show(context, data_dict):
 
 
 def resource_view_show(context, data_dict):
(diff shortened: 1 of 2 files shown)


amercader:
@wardi I pushed the change we discussed last time (not rely on the resource object in the context at all)

PR #4834, opened 2019-06-10
Fix button to not appear briefly

usingsky:
Fixes #
Sometimes, the go button **appears momentarily** on the organizations or group pages.
[ckan demo site]
image

### Proposed fixes:
The button's css file is shown below.
 - html
`
<button class="btn js-hide" type="submit">Go</button>
`
 - css
`
.js .js-hide {
  display: none;
}
`

However, the class name "js" is added in javascript. 
Therefore, if loading of javascript is delayed, it will cause an issue.

 - render .btn (css)
 - add class name "js" (javascript)
 - render .js .js-hide (css)

"js" is the class name that is added to all html by default. 
So, I think that it is not necessary to use it with the selector.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/css/fuchsia.css
+++ b/ckan/public/base/css/fuchsia.css
@@ -7218,10 +7218,10 @@ a.tag:hover {
 .nav-simple > .nav-btn .btn {
   display: inline-block;
 }
-.js .js-hide {
+.js-hide {
   display: none;
 }
-.js .js-hide.active {
+.js-hide.active {
   display: block;
 }
 .btn,
--- a/ckan/public/base/css/green.css
+++ b/ckan/public/base/css/green.css
@@ -7218,10 +7218,10 @@ a.tag:hover {
 .nav-simple > .nav-btn .btn {
   display: inline-block;
 }
-.js .js-hide {
+.js-hide {
   display: none;
 }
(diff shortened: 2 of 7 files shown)


smotornyuk:
Unfortunately, no. The idea of this(pretty old) hack - if the browser supports javascript and user hadn't disabled it intentionally, it will add `js` class to page and some interactive elements will be visually hidden but still available through js callbacks. 

If javascript is disabled for some reason, that means the user won't be able to search/do-something-else via js callbacks, so these elements must be visible. 

Generally, if you are facing this problem - the browser is too slow in page rendering. This can be caused by:
1. A lot of javascript -> reduce/minify/refactor you js
2. A lot of CSS -> rethink your theming/disable unused modules from your CSS framework
3. A lot of HTML -> try not to render so many tags in one template.
All cases above applied if you are using custom extensions and there is no such problem with pure ckan instance. But if you are facing a problem even without any extensions, it's probably caused by system performance and it cannot be helped. For your case, try to add local style patches(from admin UI, for example), but this particular PR cannot be merged into master.

usingsky:
Thanks for your feedback and explanation. It was a great help. This issue should be resolved on the my extention side.

PR #4835, opened 2019-06-10
fix TypeError keyword 'multi' in request.params.items

usingsky:
Fixes #
If "request.params.items(multi=True)" function is called with pylons request in ckan extention, TypeError will occur.

### Proposed fixes:
- Pylons request.params : webob.multidict.UnicodeMultiDict (There is no "**multi**" parameter in items())
- Flask request.params : werkzeug.datastructures.ImmutableMultiDict

In addition, I would like to ask if the 2.9.x version will support pylons or only flask.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1013,7 +1013,9 @@ def get_facet_items_dict(
     for facet_item in search_facets.get(facet)['items']:
         if not len(facet_item['name'].strip()):
             continue
-        if not (facet, facet_item['name']) in request.params.items(multi=True):
+        params_items = request.params.items(multi=True) \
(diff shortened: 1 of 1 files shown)


smotornyuk:
Merged. Thanks, @usingsky 
As for your question, in 2.9 we are going to switch all built-in actions to Flask's views, so there will be no use of controllers internally. But, for backward compatibility and in order to create some room for extension migration, controller's code won't be removed yet, so all extension that are using/extending existing controller classes shouldn't be broken. 
As for all new extensions - they should use blueprints, because controllers will be completely dropped as soon as migration to Python 3 completed(i.e, in CKAN 3.0)

PR #4838, opened 2019-06-12
Make reorder resources button translatable

Zharktas:
### Proposed fixes:
Reorder resources button has not been translatable since ckan 2.7 as message extraction does not pickup the javascript translation from variable. This PR also adds translatable help text to the module as its functionality is sometimes a little counter intuitive.

image

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/resource-reorder.js
+++ b/ckan/public/base/javascript/modules/resource-reorder.js
@@ -8,6 +8,7 @@ this.ckan.module('resource-reorder', function($) {
     },
     template: {
       title: '<h1></h1>',
+      help_text: '<p></p>',
       button: [
         '<a href="javascript:;" class="btn btn-default">',
         '<i class="fa fa-bars"></i>',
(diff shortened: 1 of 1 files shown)


Chealer:
Thank you for this change
Is it normal that as of today, the "Reorder resources" string is still missing on Transifex for CKAN 2.8, even though 2.8.3 was released last year?

Chealer:
And the "Reorder resources" string is not in ckan.pot in 2.8.3.

amercader:
@Chealer There are generally no new strings added to the patch releases so we hadn't updated Transifex yet. I pushed the latests strings now to the 2.8 resource, so you should see an untranslated "Reorder resources" string now (along some other ones). If you add translations in Transifex we will pull them before the 2.8.4 release in a couple of weeks.

Chealer:
No problem @amercader thank you very much for both the explanation and the update.

I (re)did the translation for French.
By the way, I added a comment in Transifex suggesting perhaps the English string could be improved (just "Order resources", or "Sort resources").

PR #4847, opened 2019-06-16
Build correct urls to relatively reffered files

### LINKED ISSUE #4844: 404s for minor recline assets
latest master

### Bug
When I preview a CSV on the resource page e.g. `/dataset/test/resource/0d534c55-879c-41c3-9e8f-26386b574158`
then it basically works, but there are some 404 errors fetching a few assets:

* /webassets/img/ajaxload-circle.gif
* /webassets/fonts/glyphicons-halflings-regular.woff
* /webassets/fonts/glyphicons-halflings-regular.ttf

There is a mismatch because there is an asset here: /img/ajaxload-circle.gif

I guess this is a side-effect from: https://github.com/ckan/ckan/pull/4614

smotornyuk:
Fixes #4844

When CSS file contains relative links to static files(eg. `background: url(../x.png)`), those paths are computed from webasset's path(`/webassets/...`) and(as webassets doesn't handle any files except for css/js) in UI we have a bunch of 404s. The most obvious solution here - applying `cssrewrite` filter so that your `webassets.yaml` looks like (attention to the third line)
```yaml
main-css:
  output: ckanext-reclineview/%(version)s_reclineview.css
  filters: cssrewrite
  contents:
    - vendor/bootstrap/3.2.0/css/bootstrap.css
    # ....
```
And it's exactly how one should fix such issues in future. For initial case i had to slightly update workflow for public paths registration. One of `Webassets`' CSS rewriter's rules - file that if referred by rewrote URL must exist in the filesystem and it MUST be available via public URL. So, we need to register all CKAN public URLs inside `webassets` environment so that it can verify existence of our files.

### DIFF
--- a/ckan/lib/webassets_tools.py
+++ b/ckan/lib/webassets_tools.py
@@ -58,7 +58,7 @@ def webassets_init():
     env.debug = config.get(u'debug', False)
     env.url = u'/webassets/'
 
-    env.append_path(base_path, u'/base/')
+    add_public_path(base_path, u'/base/')
 
     logger.debug(u'Base path {0}'.format(base_path))
     create_library(u'vendor', os.path.join(
@@ -159,3 +159,7 @@ def get_webassets_path():
             u'must be specified'
         )
     return webassets_path
+
+
+def add_public_path(path, url):
+    env.append_path(path, url)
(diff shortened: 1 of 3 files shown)


davidread:
Thanks @smotornyuk !

PR #4854, opened 2019-06-19
Blueprint name is 'feeds' not 'feed'

alexandru-m-g:
### Proposed fixes:
I think the Blueprint name is wrong in the html snippet. Double-checked on beta.ckan.org and the url that is generated is wrong:

image


amercader:
Thanks @alexandru-m-g

PR #4867, opened 2019-06-24
Deletion of resource before purging

### LINKED ISSUE #4705: Problems when deleting resources

2.9.0a.
 
iamarnavgarg:
Fixes #4705 

### Proposed fixes:
While purging of the dataset, the resources associated with them are deleted first and then the dataset. 

The procedure followed is as below:
a)We need to have/retrieve the package id.
b) Using the package id,we will then retrieve the list of resources associated the dataset.
c) Now, we will delete the resources associated with the datasets.
d)Now,when the resources are deleted, the flow of purging will resume its normal flow afterwards.

### DIFF
--- a/ckan/tests/views/test_admin.py
+++ b/ckan/tests/views/test_admin.py
@@ -1,10 +1,8 @@
+
 #encoding: utf-8
-
 from nose.tools import assert_true, assert_equal
-
 from ckan.lib.helpers import url_for
 from ckan.common import config
-
 import ckan.model as model
 import ckan.tests.helpers as helpers
 import ckan.tests.factories as factories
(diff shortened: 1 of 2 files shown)


amercader:
Hi @iamarnavgarg, thanks for submitting this. I'm not clear how this code helps with the original issue flagged in #4705. IIUC that issue states that if a dataset and its resources are purged, the actual data files hosted on the FileStore and the data tables hosted on the DataStore should be deleted as well.

But I don't see how your code implements that, although I might be missing something. Do you mind expand on the changes?

iamarnavgarg:
Hi @amercader,

please find the below explanation regarding the changes:
a) during purge operation of a dataset, I gather the package ID of a dataset through which I can now access the resources associated with that dataset which is in trash. 

c)Thus, from the package ID, I delete the resource associated with the datasets. By doing this, the resource is first deleted and then we could perform the purging of the datasets. 

d) After the resource has been completely deleted from the datastore and package is also updated. We can perform the normal flow/execution of purging of the dataset.

Chandradeep-NEC:
@amercader 
Hi, Please close the issue #4705 If there is no issue in PR.
Thanks.

Issue #4715, opened 2019-04-05
Docker setup is broken


frafra:
### CKAN Version if known (or site URL)
56d6a3fdb57eb3b08497f481316f95d77d2a00cb

### Please describe the expected behaviour
CKAN should work with the default Docker files.

### Please describe the actual behaviour
```
ckan          | db:5432 - no response
db            | CREATE EXTENSION
db            | CREATE EXTENSION
db            | CREATE EXTENSION
ckan          | db:5432 - no response
db            | CREATE EXTENSION
db            | Loading PostGIS extensions into ckan
db            | NOTICE:  extension "postgis" already exists, skipping
db            | CREATE EXTENSION
db            | CREATE EXTENSION
db            | CREATE EXTENSION
ckan          | db:5432 - no response
db            | CREATE EXTENSION
db            |
db            | waiting for server to shut down....2019-04-05 11:36:44.076 UTC [42] LOG:  received fast shutdown request
db            | 2019-04-05 11:36:44.078 UTC [42] LOG:  aborting any active transactions
db            | 2019-04-05 11:36:44.079 UTC [42] LOG:  background worker "logical replication launcher" (PID 49) exited with exit code 1
db            | 2019-04-05 11:36:44.087 UTC [44] LOG:  shutting down
ckan          | db:5432 - no response
db            | 2019-04-05 11:36:44.295 UTC [42] LOG:  database system is shut down
db            |  done
db            | server stopped
db            |
db            | PostgreSQL init process complete; ready for start up.
db            |
db            | 2019-04-05 11:36:44.411 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db            | 2019-04-05 11:36:44.411 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db            | 2019-04-05 11:36:44.421 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db            | 2019-04-05 11:36:44.454 UTC [106] LOG:  database system was shut down at 2019-04-05 11:36:44 UTC
db            | 2019-04-05 11:36:44.471 UTC [1] LOG:  database system is ready to accept connections
ckan          | db:5432 - accepting connections
db            | 2019-04-05 11:36:45.288 UTC [113] FATAL:  password authentication failed for user "postgres"
db            | 2019-04-05 11:36:45.288 UTC [113] DETAIL:  Role "postgres" does not exist.
[...]

PR #4876, opened 2019-06-26
Fix Postgres healthcheck

frafra:
Partial fix for #4715

### DIFF
--- a/contrib/docker/ckan-entrypoint.sh
+++ b/contrib/docker/ckan-entrypoint.sh
@@ -41,7 +41,7 @@ write_config () {
 }
 
 # Wait for PostgreSQL
-while ! pg_isready -h db -U postgres; do
+while ! pg_isready -h db -U ckan; do
   sleep 1;
 done
 
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -62,7 +62,7 @@ services:
     volumes:
       - pg_data:/var/lib/postgresql/data
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "postgres"]
+      test: ["CMD", "pg_isready", "-U", "ckan"]
 
   solr:
     container_name: solr


amercader:
Can you explain the reasoning behind this change @frafra ?

frafra:
Sure @amercader :-)

Here is the error you get from the health check if you run `docker-compose up`:
```FATAL: role "postgres" does not exist```

This happens because the users used by the health check is wrong: the user used by the ckan container is `ckan`, not `postgres`:
https://github.com/ckan/ckan/blob/aa1442cdea0d414a314285151d265b7f719bed81/contrib/docker/docker-compose.yml#L26

**Comment by davidread** [comment]
My understanding is that the default user is `postgres` so that should always work, but `ckan` should be fine too. Both work for me locally. So I'm not clear why this PR would help. Especially when the error message is about a *role*, rather than the *user* being changed in the PR. 🤷‍♂ 

BTW I tried docker-compose on this branch but get another unrelated error - a 404 for SOLR. But I'm not hugely familiar with this area, so I probably did something wrong...

**Comment by frafra** [comment]
I can confirm that this patch fixes the issue, while keeping "postgres" make it fail. Bear in mind that we could have been using different Postgres images/versions due to the fact that there is no tag specified at the moment on the Dockerfile.

**Comment by davidread** [comment]
I tried again and can reproduce this exact error, as per the issue:
```
cd contrib/docker
cp .env{.template,}
docker-compose up --build --exit-code-from ckan
```
 I'm still not sure why postgres user isn't there, but happy to just have it merged to fix this. I'll leave it to @amercader to decide as he's assigned.

We should really have CI if we're going to maintain this.

frafra:
Ok, it looks like there is a difference between the Docker image on Dockerhub and the one you can generate from the repository: the first uses postgresql as user, the latter uses ckan.

davidread:
Good find. [Dockerhub currently has a release](https://hub.docker.com/r/ckan/postgresql/builds) of: https://github.com/ckan/ckan/blob/60b98f932629acf67428c445a4e3d79fed495912/contrib/docker/postgresql/Dockerfile which is 3 years old. Sounds like someone with admin rights to https://hub.docker.com/u/ckan should setup autobuild, to ensure it doesn't diverge in future. This is probably a topic to take off line.

I'm keen to see this PR merged whatever.

markstuart:
@frafra @davidread what is needed to get this over the line? I'm keen to have this fix in as part of the work in #5381 , it'd be nice if the db logs weren't filling up with errors from the health check call :+1:

markstuart:
Thanks @amercader 😁

PR #4891, opened 2019-07-03
Adding additional step in installation useful for beginners

### LINKED ISSUE #4344: Additional step in installation from source would be useful for beginners

2.9.0a

According to the docs (http://docs.ckan.org/en/latest/maintaining/installing/install-from-package.html#install-and-configure-postgresql) PostgreSQL should list the existing databases after being installed

<pre>
    <strong>orphan:</strong>
    Check that PostgreSQL was installed correctly by listing the existing databases:
    
        sudo -u postgres psql -l
</pre>

(is **orphan:** an error message in this instance?)

Running the commands in the docs gives:

    chris@u1:~$ sudo -u postgres psql -l
    psql: could not connect to server: No such file or directory
            Is the server running locally and accepting
            connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?

It seems that PostgreSQL needs to be started after being installed with ``sudo service postgresql start`` before the above command can work. Can this be added to the docs?

(I tried to find it in the source, but 2.9.0a doesn't seem to be listed as a tag or a branch in the github repo, and the relevant rst file in the main branch also doesn't seem to correlate with 2.0.9a docs - what branch is it under?)

iamarnavgarg:
Fixes #4344 

Updated the ckan docs.

### DIFF
--- a/doc/maintaining/installing/install-from-package.rst
+++ b/doc/maintaining/installing/install-from-package.rst
@@ -105,13 +105,13 @@ CKAN:
 
 .. tip::
 
-.. note::
-     
-   The commands mentioned below are tested for Ubuntu system
-   
    You can install |postgres| and CKAN on different servers. Just
    change the :ref:`sqlalchemy.url` setting in your
    |production.ini| file to reference your |postgres| server.
+
+.. note::
+     
+   The commands mentioned below are tested for Ubuntu system
 
 Install |postgres|, running this command in a terminal::
 
--- a/doc/maintaining/installing/install-from-package.rst
+++ b/doc/maintaining/installing/install-from-package.rst
@@ -105,6 +105,10 @@ CKAN:
 
 .. tip::
 
+.. note::
+     
+   The commands mentioned below are tested for Ubuntu system
+   
    You can install |postgres| and CKAN on different servers. Just
    change the :ref:`sqlalchemy.url` setting in your
    |production.ini| file to reference your |postgres| server.
--- a/doc/maintaining/installing/postgres.rst
+++ b/doc/maintaining/installing/postgres.rst
@@ -2,8 +2,8 @@
 
 .. note::
 
-    The following commands should be executed only after running the postgres server. To run 
-    postgres server, execute the command- ``service postgresql start``. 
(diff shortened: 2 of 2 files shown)


doc/maintaining/installing/install-from-package.rst
```diff
@@ -109,6 +109,10 @@ CKAN:
    change the :ref:`sqlalchemy.url` setting in your
    |production.ini| file to reference your |postgres| server.
 
+.. note::
+     
+   The commands mentioned below are tested for Ubuntu system
```
smotornyuk: What's the point of continuous repetition if Ubuntu was mentioned as requirement in the very first paragraph of this page?

doc/maintaining/installing/postgres.rst
```diff
@@ -1,5 +1,11 @@
 :orphan:
 
+.. note::
+
+    The following commands should be executed only after running the postgres server. To run 
+    postgres server, execute the command- ``service postgresql start``. 
+
```
smotornyuk: Isn't PostgreSQL daemon started automatically on Ubuntu, right after installation? (but, explicit mention won't harm, so i don't have anything particular against this note, just curious)

smotornyuk: PS so mentioning the necessity of starting PostgreSQL should be in the form of 'if you are facing a problem because of Postgresql is not running, you have to..`. And, unless the user uses root account by default, you also should put `sudo` in the example, just for accordance with other commands. And this hyphen before the command can be omitted, right?

doc/maintaining/installing/install-from-package.rst
iamarnavgarg: Yes, Ubuntu has mentioned at the top, but still One issue has been raised by @wood-chris while he was trying to install on windows. 
If we add Ubuntu explicitly with the command so in this way it will be more clear for end user perspective

doc/maintaining/installing/postgres.rst
iamarnavgarg: Alright, I'll incorporate the relevant change.

Issue #4792, opened 2019-05-23
Write a Flask and Python 3 migration guide for extensions


amercader:
There is a minimal [page](https://docs.ckan.org/en/2.8/extensions/flask-migration.html) in the docs that needs to be expanded. Originally it was just targeted at Flask specific issues but I think it makes sense to expand the scope and include wider Python 3 support issues.

Things it should cover:

* Adding Flake8 tests / formatters to Travis tests to prepare code for Python 3 support, provide examples and update the extension template.
* Use the toolkit / ckantoolkit for imports
* `url_for` changes
* `IMapper` + `BaseController` migration to `IBlueprint` + Blueprints

Additionally we should update the [toolkit reference](https://docs.ckan.org/en/2.8/extensions/plugins-toolkit.html) to mark properties that will be deprecated (like `BaseController`).

#### Approach

Write the docs :) This is likely going to be done alongside the actual migration of some extensions.

#### Story points

3

smotornyuk:
_Just keeping notes in one place_

1) create py3 branch 
2) move plugin.py into the plugin/__init__.py. This won't break anything, as import path for plugin won't be changed. But you need to update relative paths. enable `absolute_imports` for python modules, and check, whether you registering any new directories(templates, resources, translations). For example, you'll likely make following change: 
```python
p.toolkit.add_template_directory(config, "templates")
# -> 
p.toolkit.add_template_directory(config, "../templates")

```
3) remove the version-dependent code from the original plugin(IMapper/IBlueprint)
4) create plugin/flask_plugin.py and plugin/pylons_plugin.py. Each of those files contains version-specific plugin-mixin(just ordinary CKAN plugin). In our example case, flask_plugin adds new routes via blueprint, and pytlons_plugin is doing it via before_map from IMapper.
5) Depending on current CKAN version, import one of the sub-plugins into main plugin 
```python
try:
    tk.requires_ckan_version("2.9")
except CkanVersionException:
    from ckanext.my_ext.plugin.pylons_plugin import MixinPlugin
else:
    from ckanext.my_ext.plugin.flask_plugin import MixinPlugin
```
6) Add version-specific plugin as a mixin to the main plugin class
```python
class MyPlugin(MixinPlugin, p.SingletonPlugin):
```
7) If plugin registers CLI commands, add cli.py with relevant logic. Follow  #5112  as manual for registering CLI commands.
8) Create webassets.yml for resources **(there is no documentation for webassets yet:( )**
[...]

Issue #4791, opened 2019-05-23
Migrate controllers from core extensions to blueprints


amercader:
As it has been done with the core ones. The extensions to be migrated are:

- [x] datatablesview
- [x] example_iconfigurer
- [x] resourceproxy
- [x] stats
- [x] datapusher
- [x] datastore

I think is very unlikely that someone is extending these controllers from
an extension so I propose that we remove the controllers as part of the migration to  
blueprints.

#### Approach

For each extension:

1. Add a blueprint.py file, create a blueprint object and recreate the routes defined in the `IRoutes.before_map` method of the `plugin.py` file.
2. Move the logic of each controller action to a new view function, referenced by the routes created.
3. Implement `IBlueprint.get_blueprint` in the `plugin.py` file and return the blueprint created above
4. Make sure that the extension is using the plugins toolkit whenever possible (eg no imports from `ckan.lib.base`, `pylons`, `logic`, etc)

Following is a full example (only showing relevant code). Note that the actual code inside the controller action / view function remains unchanged (unless there is some web framework specific thing, which is unlikely).

Before:

```python
# ckanext/datapusher/plugin.py

class ResourceDataController(base.BaseController):

    def resource_data(self, id, resource_id):

        if toolkit.request.method == 'POST':
            try:
                toolkit.c.pkg_dict = p.toolkit.get_action('datapusher_submit')(
                    None, {'resource_id': resource_id}
                )   
            except logic.ValidationError:
                pass

            core_helpers.redirect_to(
                controller='ckanext.datapusher.plugin:ResourceDataController',
                action='resource_data',
                id=id,
                resource_id=resource_id
            )

        try:
            toolkit.c.pkg_dict = p.toolkit.get_action('package_show')(
                None, {'id': id}
            )
            toolkit.c.resource = p.toolkit.get_action('resource_show')(
                None, {'id': resource_id}
            )
        except (logic.NotFound, logic.NotAuthorized):
            base.abort(404, _('Resource not found'))

        try:
            datapusher_status = p.toolkit.get_action('datapusher_status')(
                None, {'resource_id': resource_id}
            )
        except logic.NotFound:
            datapusher_status = {}
        except logic.NotAuthorized:
            base.abort(403, _('Not authorized to see this page'))

        return base.render('datapusher/resource_data.html',
                           extra_vars={
                               'status': datapusher_status,
                               'pkg_dict': toolkit.c.pkg_dict,
                               'resource': toolkit.c.resource,
                           })

[...]

Issue #4639, opened 2019-01-28
Replace Paste with Click


kmbn:
Replace all Paste/paster with Click if they're CLI commands or with another approach if they're used internally.

This is required for migrating from Pylons to Flask and for enabling Python 3 compatibility.

## Approach

User-wise we are going from this:

    paster search-index rebuild

To this:

    ckan search-index rebuild

The goal is to move all commands in `ckan/lib/cli.py` to the new `ckan/cli` folder. The already migrated commads in the `cli` folder should be self-explanatory but overall the process is:

* Pick a pending command from the list below and create a new file for it in `ckan/cli`
* Add a entrypoint using `@click.group`, eg:
    ```
    @click.group(name=u'db', short_help=u'Database commands')
    def db():
        pass
    ```
* Add individual subcommands. In most cases the actual code used in the current cli will remain unchanged, but of course feel free to add any improvements that you consider worth doing.
    ```
    @db.command(u'upgrade', short_help=u'Upgrade the database')
    @click.option(u'-v', u'--version', help=u'Migration version')
    def updatedb(version=None):
        # ...
    ```
* Add the command to the end of the `ckan/cli/cli.py` file

### Current status

 paster command | descritpion | status | ckan command 
  -------------------- | ----------------- | -------- | ---------------------
  check-po-files  |  Check po files for common mistakes |  migrated | `ckan translation`
  color             | Create or remove a color scheme.| drop it|
  config-tool      | Tool for editing options in a CKAN config file | migrated
  create-test-data |  Create test data in the database. | migrated | renamed to `seed`|
  datapusher      |  Perform commands in the datapusher | migrated |
  dataset         |  Manage datasets | migrated |
  datastore      |   Perform commands to set up the datastore | migrated|
  db          |      Perform various tasks on the database. | migrated | `ckan -c ckan.ini db` |
  front-end-build |  Creates and minifies css and JavaScript files | migrated
  jobs           |   Manage background jobs | migrated
  less         |     Compile all root less documents into their CSS counterparts | migrated
  minify       |     Create minified versions of the given Javascript and CSS files.| migrated
  notify        |    Send out modification notifications. | migrated
  plugin-info   |    Provide info on installed plugins. | migrated
  profile       |    Code speed profiler | migrated
  ratings       |    Manage the ratings stored in the db | drop it|
  rdf-export    |    Export active datasets as RDF | drop it, if useful will be ported to ckanext-dcat |
[...]

PR #4896, opened 2019-07-09
[WIP] Py 2-3 guide

### LINKED ISSUE #4855: Python 2/3 migration guide
# User story

As a {developer} working on Py2/3 migration for CKAN, I want a clear migration guide, so I can provide Py2/3 compatibility in a consistent way using a common set of tools.

# Use cases

- Contributing Py2/3 compatible code in CKAN Core
- Contributing Py2/3 compatible code in a CKAN Extension
- Starting a new CKAN Extension with Py2/3 compatibility

pwalsh:
Fixes #4855 

### Proposed fixes:

- Guide for developers contributing to Py2-3 codebases around CKAN.

### DIFF
--- a/doc/contributing/py-2-3-guide.rst
+++ b/doc/contributing/py-2-3-guide.rst
@@ -1,0 +1,117 @@
+=====================
+CKAN Python 2/3 Guide
+=====================
+
+This section documents the requirements for porting existing CKAN code for Python 3 support, and for creating new code that supports both Python 2 and Python 3.
+
+----------
+User story
+----------
+
+As a developer working on CKAN or CKAN Extensions, I want a clear guide, for writing code compatible with Python 2 and Python 3, so my code is consistent with common practice around CKAN, using a common set of tools.
+
+---------
+Use cases
+---------
+
+- Contributing Python 2/3 compatible code in CKAN Core
+- Contributing Python 2/3 compatible code in a CKAN Extension
+- Starting a new CKAN Extension with Python 2/3 compatibility
+
+--------
+Python 3
+--------
+
+From CKAN 2.9, CKAN will officially support Python 2 (≥ 2.6) and Python 3 (≥ 3.6).
+
+-----
+Flask
+-----
+
+CKAN provides Python 3 support by removing the underlying Pylons web framework, which is only compatible with Python 2, and integrating Flask as the web framework, which is compatible with Python 2 and Python 3. Flask is a very popular framework for writing web applications in Python; read more about Flask_.
+
+--------
+Approach
+--------
+
+CKAN provides support for both Python 2 and Python 3 using a compatible source approach. This enables us to continue supporting Python 2 in the current transitional period, and gracefully upgrade to a Python 3 only codebase in future versions of CKAN (from CKAN 3.0 onwards).
+
+----
+Tips
+----
+
+Future imports
+##############
+
+Import the following at the top of all modules::
+
+  from __future__ import absolute_import
+  from __future__ import division
+  from __future__ import print_function
+  from __future__ import unicode_literals
+
+String handling
+###############
+
+- Remove any ``u`` prefixes before unicode strings in existing Python 2 code.
+- Add a ``b`` prefix for any bytestrings
+- Review code to ensure only unicode strings are exposed as part of the public interfaces of the codebase, which will make string handling more consistent.
+  
+Ensure you read the references below for further background information on all these tips.
+
+Exception handling
+##################
+
+Python-Future provides a clean interface for exception handling compatibility::
+
+  from future.utils import raise_
+  traceback = sys.exc_info()[2]
+  raise_(ValueError, "dodgy value", traceback)
+
+Iterating over dicts
+####################
+
+Python-Future provides a clean interface for iterating over dicts (and for creating custom iterators - see the docs)::
+  from builtins import dict
+  from future.utils import listvalues
+  from future.utils import itervalues
+  from future.utils import listitems
+  from future.utils import iteritems
+  
+
+  heights = dict(Fred=175, Anne=166, Joe=192)
+
+  # value lists
+  valuelist = listvalues(heights)
+  valuelist = list(itervalues(heights))
+
+  # item lists
+  itemlist = listitems(heights)
+  itemlist = list(iteritems(heights))
+
+Flask-based API for extensions
+##############################
+
+TODO
+
+-------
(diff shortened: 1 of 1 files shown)


pwalsh:
@pdelboca @amercader 

Specially on Flask-based APIs for extensions, can @pdelboca contribute for this section, as I am not aware of the details, and you are both working on a codebase for this right now.

pdelboca:
Great to have this documented! There are some pieces of information in several tickets, and a Issue for Py 2-3 Guide for extensions (https://github.com/ckan/ckan/issues/4792)

I will leave here some of those interesting pieces of information that are suitable to be written down and merged in this PR:
- In https://github.com/ckan/ckan/issues/4791 there is a good approach on how to think the migration of extensions to Flask.
- There are already some core views migrated to Flask that can be use as reference and inspiration to design extensions (https://github.com/ckan/ckan/tree/master/ckan/views)
- There are some extensions already made in Flask like for example: https://github.com/okfn/ckanext-collaborators 
- There is also an old but good guide and documentation for the migration with lot of useful information in the Wiki: https://github.com/ckan/ckan/wiki/Migration-from-Pylons-to-Flask
- Not sure if related but it is worth noticing that there is ongoing work to migrate the command line tool (https://github.com/ckan/ckan/issues/4639) as it may also have an impact on how to design commands for the extensions.

doc/contributing/py-2-3-guide.rst
```diff
@@ -0,0 +1,117 @@
...
+Ensure you read the references below for further background information on all these tips.
+
+Exception handling
+##################
+
+Python-Future provides a clean interface for exception handling compatibility::
+
+  from future.utils import raise_
+  traceback = sys.exc_info()[2]
+  raise_(ValueError, "dodgy value", traceback)
```
wardi: -1 on python-future. I strongly dislike rewriting exceptions this way. I think the cases in ckan core where we raise exceptions that won't work in python 2/3 should just be fixed (if there are any)

doc/contributing/py-2-3-guide.rst
```diff
@@ -0,0 +1,117 @@
...
+
+  heights = dict(Fred=175, Anne=166, Joe=192)
+
+  # value lists
+  valuelist = listvalues(heights)
+  valuelist = list(itervalues(heights))
+
+  # item lists
+  itemlist = listitems(heights)
+  itemlist = list(iteritems(heights))
```
wardi: I don't see list and dict iteration differences in Python 3 as a big problem. For core we can just assume iterating will be lazy and pass to `list()` when we actually need a list.

doc/contributing/py-2-3-guide.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Future imports
+##############
+
+Import the following at the top of all modules::
+
+  from __future__ import absolute_import
+  from __future__ import division
+  from __future__ import print_function
+  from __future__ import unicode_literals
```
wardi: :+1: from me

doc/contributing/py-2-3-guide.rst
pwalsh: Hey, thanks.

As far as I know, raising exceptions with a traceback, and also reprising exceptions, across both 2 and 3 is only possible like this. Of course the standard approach of `Exception as e` is better for simple cases.

Some more info [here](https://python-future.org/compatible_idioms.html#raising-exceptions) and [here](https://franklingu.github.io/programming/2016/06/30/properly-reraise-exception-in-python/).

doc/contributing/py-2-3-guide.rst
pwalsh: I guess the issue is, the same syntax in Python 2 and Python 3 has different behaviour, and that make lead to either different behaviours or inconsistencies (or, not).

I'd only suggest such an approach for Py 2-3 codebases, and of course would also prefer the Py 3 only approach of lazy iteration being assumed.

Overall, fine by me if the core team decides iteration differences are not critical, and I can just update the doc to indicate that as a potential gotcha.

doc/contributing/py-2-3-guide.rst
```diff
@@ -0,0 +1,117 @@
...
+
+- Contributing Python 2/3 compatible code in CKAN Core
+- Contributing Python 2/3 compatible code in a CKAN Extension
+- Starting a new CKAN Extension with Python 2/3 compatibility
+
+--------
+Python 3
+--------
+
+From CKAN 2.9, CKAN will officially support Python 2 (≥ 2.6) and Python 3 (≥ 3.6).
```
amercader: CKAN 2.9 will support Python 2 (≥ 2.7) and Python 3 (≥ 3.6).

CKAN 3.0 will only support Python 3 (≥ 3.6)

doc/contributing/py-2-3-guide.rst
amercader: Shouldn't we add only the ones needed? I guess it makes sense `unicode_literals` is always present but for instance the print function is only needed in the cli module. I only could find divisions on 5 modules. And regarding the imports, the current codebase favours absolute imports anyway so perhaps we should recommend that and not load `absolute_import`

doc/contributing/py-2-3-guide.rst
amercader: I agree this seems overkill. I grepped the code for exceptions that attach tracebacks and couldn't find any (maybe some on the search index rebuild CLI). There are several instances where we catch exceptions and we reraise custom ones like, but most are at the validators and logic level and are later catch by the blueprints.

I think is best to deal with problematic cases as we find them

amercader:
What are people's views on imports from `urllib`? On https://github.com/ckan/ckan/pull/4045 we introduced the six based syntax in cli.py:

```python
from six.moves.urllib.error import HTTPError
from six.moves.urllib.parse import urljoin, urlparse
from six.moves.urllib.request import urlopen
```
I quite like it and I don't think the [alternatives](https://python-future.org/compatible_idioms.html#urllib-module) offer a bigger benefit.

What I would suggest is restrict the `urllib` usage to utilities like `urlparse` and `urlencode` and use requests for actual calls and exception handling.

Does this sound good?

davidread:
@amercader yeah I like this six syntax the best

amercader:
Closing in favour of https://github.com/ckan/ckan/issues/4792. End result: https://github.com/ckan/ckan/wiki/Python-3-migration-guide-for-extensions

PR #4904, opened 2019-07-12
Deleting a dataset now also deletes its resources

### LINKED ISSUE #4705: Problems when deleting resources

davidread:
Fixes #4705

@dassolkim is concerned that IResourceController.before_delete event is
not firing at any point when a dataset is deleted and then purged
(either with package_purge or the admin.Trash page). PR #4867 seeks to
implement the event in the admin.Trash page. But this PR proposes the
alternative, which is to actually delete the resource when the dataset
is deleted and fire the event then.

It seems an anomaly that we don't delete resources with the dataset. It
makes sense that the dataset and resources are both state=deleted.
However there is an argument that we shouldn't, because a sysadmin can still
see the deleted dataset, but with this PR it will have no resources.

Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/delete.py
+++ b/ckan/logic/action/delete.py
@@ -92,6 +92,29 @@ def package_delete(context, data_dict):
     rev.author = user
     rev.message = _(u'REST API: Delete Package: %s') % entity.name
 
+    # delete the dataset's resources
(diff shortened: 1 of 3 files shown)


davidread:
Closed in favour of https://github.com/ckan/ckan/pull/4905

PR #4905, opened 2019-07-12
Deleting a dataset sends IResourceController.before_delete

### LINKED ISSUE #4705: Problems when deleting resources

davidread:
Although it doesn't actually delete the resources, because when a
sysadmin views the deleted dataset it should show with those resource,
and they wouldn't show up if we set their state=deleted.

This fixes #4705 by sending the IResourceController.before_delete events
necessary for Datastore and FileStore to get rid of the associated data.

I think this is a better option compared to deleting the resources, which can be seen in alternate PR https://github.com/ckan/ckan/pull/4904

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/delete.py
+++ b/ckan/logic/action/delete.py
@@ -92,6 +92,29 @@ def package_delete(context, data_dict):
     rev.author = user
     rev.message = _(u'REST API: Delete Package: %s') % entity.name
 
+    # delete the dataset's resources
+    pkg_dict = _get_action('package_show')(context, {'id': id})
+    for res in pkg_dict['resources']:
+        for plugin in plugins.PluginImplementations(
+                plugins.IResourceController):
+            plugin.before_delete(context, res,
+                                 pkg_dict.get('resources', []))
+
+    def pop_resource(pkg_dict, res_id):
+        for res in pkg_dict.get('resources', []):
+            if res['id'] == res_id:
+                pkg_dict['resources'].remove(res)
+                return
+
+    pkg_dict = _get_action('package_show')(context, {'id': id})
+    for res in entity.resources:
+        pop_resource(pkg_dict, res.id)
+        res.delete()
+        for plugin in plugins.PluginImplementations(
+                plugins.IResourceController):
+            plugin.after_delete(context, pkg_dict.get('resources', []))
+
+    # delete the dataset
     for item in plugins.PluginImplementations(plugins.IPackageController):
         item.delete(entity)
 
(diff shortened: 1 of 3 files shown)


iamarnavgarg:
Hi @davidread , the code changes done by you seems to disassociate the resources with the plugins such as datastore. Can you please explain the reasons for keeping the resources even if the package is deleted and is likely to be purged.?

davidread:
> explain the reasons for keeping the resources even if the package is deleted and is likely to be purged.?

Yes, after you have done package_delete, dataset.state=deleted and resources.state=active. The dataset can still be viewed by a sysadmin, which is useful for understanding what was deleted.

If the resource is also deleted, i.e. dataset.state=deleted and resources.state=deleted, when the sysadmin views it, the dataset will show without any resources. I don't think this is helpful, and was the view of the tech team many years ago.

However this seems the right time to tell datastore etc that the resources are deleted. (Better than if a dataset/resources are purged.)

iamarnavgarg:
Hi @davidread, In the fix of #4867 (Line 186) we try to implement the similar behaviour as stated by you, wherein, the resource.state='deleted' are still visible when dataset.state=deleted. The criteria that the sysadmin should be able to view the resources associated with the dataset is fulfilled. Thus, when the purge button is pressed, then only the associated resources get deleted. So, as per my understanding this is expected behaviour. Please let me know your opinion on this @davidread ?

amercader:
I think the issue with keeping the resources in `state=active` is that then the Datastore won't delete the associated table, if I'm reading this code correctly:

https://github.com/ckan/ckan/blob/master/ckanext/datastore/plugin.py#L151:L154

I wonder if the potential issue of sysadmins looking at deleted datasets and not seeing the resources will get invalidated with the new activity streams? IIUC the new activities store the full dataset dict so that would contain the resources. So an admin could look at this old "revision" or version. Maybe there's not a UI for this or maybe that's not how it works.

amercader:
Also perhaps it also makes sense to call `IResourceController.before_delete/after_delete` on `dataset_purge`? AFAICT it's not calling `package_delete` but it would be also good to trigger the files/tables deletion on purge as well right?

iamarnavgarg:
> I think the issue with keeping the resources in `state=active` is that then the Datastore won't delete the associated table, if I'm reading this code correctly:
> 
> https://github.com/ckan/ckan/blob/master/ckanext/datastore/plugin.py#L151:L154
> 
> I wonder if the potential issue of sysadmins looking at deleted datasets and not seeing the resources will get invalidated with the new activity streams? IIUC the new activities store the full dataset dict so that would contain the resources. So an admin could look at this old "revision" or version. Maybe there's not a UI for this or maybe that's not how it works.

Can you please elaborate more on this ? And is it related to fix **4867** ?

pdelboca:

PR #4910, opened 2019-07-17
Fix TypeError in recline view

usingsky:
Fixes #

error occurs if you put a character or a blank value in a non-text filter such as numeric.
No error message and loading does not disappear.
image

### Proposed fixes:
add error handling.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/reclineview/theme/public/vendor/ckan.js/ckan.js
+++ b/ckanext/reclineview/theme/public/vendor/ckan.js/ckan.js
@@ -43,6 +43,11 @@ if (isNodeModule) {
   my.Client.prototype.datastoreQuery = function(queryObj, cb) {
     var actualQuery = my._normalizeQuery(queryObj);
     this.action('datastore_search', actualQuery, function(err, results) {
+      if(err || !results){
+        cb(err);
+        return;
+      }
+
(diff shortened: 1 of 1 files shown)


ckanext/reclineview/theme/public/vendor/ckan.js/ckan.js
```diff
@@ -43,6 +43,11 @@ if (isNodeModule) {
   my.Client.prototype.datastoreQuery = function(queryObj, cb) {
     var actualQuery = my._normalizeQuery(queryObj);
     this.action('datastore_search', actualQuery, function(err, results) {
+      if(err || !results){
```
smotornyuk: What is the possible case, when there will be no error but `results` appears to be logically `false`? Won't it be better to return empty array/object(whatever represents empty result here)?

usingsky: Thank you for your review.

I agree that 'results' are not false without errors.
But, It is because of the this that there is no 'result' attribute in the results variable.
so I added it as defensive code.
I do not know what errors will occur in datastore, I think that returning empty results can be considered to be no-error.

Anyway, I will check this PR more later

PR #4912, opened 2019-07-19
[#4796] Include repoze.who friendlyform in source

### LINKED ISSUE #4796: Replace repoze.who-friendlyform with a modern alternative
We use the [repoze.who-friendlyform](http://code.gustavonarea.net/repoze.who-friendlyform/) plugin during our authentication process. This plugin was last updated on 2010 and it's not py3 ready. More specifically we use it as [Indentifier](https://repoze.readthedocs.io/projects/repozewho/en/latest/plugins.html#identifier-plugins) (extract user name and password from the WSGI env) and as a [Challenger](https://repoze.readthedocs.io/projects/repozewho/en/latest/plugins.html#challenger-plugins) (present the user with a login/logout page). See the [who.ini](https://github.com/ckan/ckan/blob/master/ckan/config/who.ini) file.

The FriendlyForm plugin extends a default repoze.who one, [RedirectorPlugin](https://repoze.readthedocs.io/projects/repozewho/en/latest/plugins.html#repoze.who.plugins.redirector.RedirectorPlugin) with some extra functionality. The one that we use is

> Developers may define post-login and/or post-logout pages.

#### Approach

This is a pretty critical functionality so we need to come up with a good replacement strategy. It might be tempting to try to replace the whole repoze.who authenticating system with something like [Flask-login](https://flask-login.readthedocs.io/en/latest/), which has a much simpler and straight-forward integration and is better supported, but that is a big change, as we use some of the repoze.who assumptions at differents parts of the code (eg relying on the `REMOTE_USER` environ variable). So I'd suggest to plan that for a later stage and keep repoze.who in place for at least 2.9.

This means replacing or upgrading repoze.who-friendlyform. The plugin itself is not massive:

https://gist.github.com/amercader/0643f17f3d9641a49c3ed9f630d545df

So perhaps the best solution for now is to include this as part of the CKAN code base and upgrade it to run on Python 3

#### Story points

5

amercader:
Fixes #4796

In order to make repoze.who.friendlyplugin Python 3 friendly we include its source in the CKAN code base. I modified it to make it Python 3 ready and replaced usage of paster methods.

This will require an update in the who.ini file


amercader:
Add note to cahngelog about who.ini

amercader:
@smotornyuk all green now

smotornyuk:
I've played with this locally on weekend and finally clicking on merge button

Issue #3980, opened 2018-01-08
Error in datastore updating resource


jroigfer:
### CKAN Version if known (or site URL)
2.6.3

### Please describe the expected behaviour
Update resource uploading new file should be replace the file.

### Please describe the actual behaviour
Connecting through a vpn to the internal network and accessing the portal to update a resource, the resource always is updated correctly whitout errors, but accessing from the internal network without vpn, when trying to replace the file by updating the resource, an internal server error (500) appears, the datapusher if it updates the resource information well, but in the ckan logs an error appears when executing the sql SELECT alias_of FROM "_table_metadata" WHERE name = :id' in the datastore database:
```
[Mon Jan 08 18:12:50.968429 2018] [:error] [pid 29552] 2018-01-08 18:12:50,968 INFO [ckan.lib.base] /api/3/action/datastore_delete render time 0.110 seconds
[Mon Jan 08 18:12:50.983580 2018] [:error] [pid 29551] res_id:u'a7fc1c26-ac31-46e5-be04-a87499ea1b87'
[Mon Jan 08 18:12:50.983599 2018] [:error] [pid 29551] data_dict:{'connection_url': 'postgresql://xxxxxx:xxxx@localhost/xxxxx', 'limit': 0, 'resource_id': u'a7fc1c26-ac31-46e5-be04-a87499ea1b87'}
[Mon Jan 08 18:12:50.983609 2018] [:error] [pid 29551] results.count:0
[Mon Jan 08 18:12:50.988776 2018] [:error] [pid 29551] [remote ip_remote] Error - <class 'ckan.logic.NotFound'>: Resource "a7fc1c26-ac31-46e5-be04-a87499ea1b87" was not found.
[Mon Jan 08 18:12:50.993765 2018] [:error] [pid 29551] [remote ip_remote] URL: http://ckan_url/dataset/nuevo-dataset-01/resource/a7fc1c26-ac31-46e5-be04-a87499ea1b87
[Mon Jan 08 18:12:50.993776 2018] [:error] [pid 29551] [remote ip_remote] File '/usr/lib/ckan/client/lib/python2.7/site-packages/weberror/errormiddleware.py', line 171 in call
[Mon Jan 08 18:12:50.993778 2018] [:error] [pid 29551] [remote ip_remote] app_iter = self.application(environ, sr_checker)
[...]

Issue #4287, opened 2018-06-07
WebApp Error: <class 'ckan.logic.NotFound'>: Resource "...resource id...." was not found.


ply112:
### CKAN Version if known (or site URL)
2.6.1

### Please describe the expected behaviour
A package is successfully created and two resources are added.  If one wants to go back and add a description for one of the resources, he can simply click on "Explore" and then "Edit" to go back to that resource, add a description for that resource, and then click [Update Resource].  

### Please describe the actual behaviour
After adding a description and clicking [Update Resource], a "Server Error.  An internal error occurred" message is returned on the website.  In addition, an email with error messages is sent to the administrator.  Below are the error messages.  I didn't copy the rest of the error messages since they were quite long but I could provide that if necessary.  Please note that the resource description was successfully added despite receiving an error message from CKAN.

------------------------------------------------------------------------------------------------
Error Messages
------------------------------------------------------------------------------------------------
WebApp Error: <class 'ckan.logic.NotFound'>: Resource "a4743bb4-2ae5-4224-88c7-ada1b7d7552f" was not found.

Module ckan.controllers.package:1128 in resource_read
>>  return render(template, extra_vars=vars)
Module ckan.lib.base:174 in render
>>  return cached_template(template_name, render_template)
Module pylons.templating:249 in cached_template
>>  return render_func()
Module ckan.lib.base:128 in render_template
>>  return render_jinja2(template_name, globs)
Module ckan.lib.base:85 in render_jinja2
>>  return template.render(**extra_vars)
Module jinja2.environment:989 in render
>>  return self.environment.handle_exception(exc_info, True)
Module jinja2.environment:754 in handle_exception
>>  reraise(exc_type, exc_value, tb)
Module /usr/lib/ckan/default/src/ckan/ckan/templates/package/resource_read.html:3 in top-level template code
>>  {% set res = c.resource %}
Module /usr/lib/ckan/default/src/ckan/ckan/templates/package/base.html:3 in top-level template code
>>  {% set pkg = c.pkg_dict or pkg_dict %}
Module /usr/lib/ckan/default/src/ckan/ckan/templates/page.html:1 in top-level template code
>>  {% extends "base.html" %}
Module /usr/lib/ckan/default/src/ckan/ckan/templates/base.html:103 in top-level template code
>>  {%- block page %}{% endblock -%}
Module /usr/lib/ckan/default/src/ckan/ckan/templates/page.html:19 in block "page"
>>  {%- block content %}
[...]

PR #4918, opened 2019-07-23
handle missing resources in case we have a race condition with the Da…

ThrawnCA:
…taPusher, github #3980

It is possible for a resource to be deleted by the DataPusher at the point where we're trying to view it. This pull request handles the error and treats it like a non-datastore resource.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2331,7 +2331,11 @@ def resource_view_get_fields(resource):
         'limit': 0,
         'include_total': False,
     }
-    result = logic.get_action('datastore_search')({}, data)
+    import ckan.plugins as p
+    try:
+        result = logic.get_action('datastore_search')({}, data)
+    except p.toolkit.ObjectNotFound:
+        return []
 
     fields = [field['id'] for field in result.get('fields', [])]
 


boykoc:
@amercader @ThrawnCA I can confirm that this patch fixes my experience with issue #4287 

Although, I'm kind of feeling like this should be addressed somewhere else in datapusher or datastore. When a resource is re-submitted and being populated, why is the datastore_active flag set? Or whatever is deleting the record to re-submit it should set the proper "flags" to prevent things from trying to call it. (not sure yet exactly where the "delete" is happening though...)

It seems in the templates there's lots of checking for datastore_active or if there's a view and both seem to stay as True, which results in trying to load a non-existent datastore record for the resource.   The templates chase down from package/resource_read, to package/snippets/resource_view, to package/snippets/resource_view_filters.html until they crash out trying to load a datastore record that isn't in the datastore because it's deleted and being resubmitted.

amercader:
Thanks @ThrawnCA , merged and backported. I just added a small change to use core exceptions rather than the toolkit ones (493d601a8) can you confirm that still works as expected?

ThrawnCA:
@amercader Thanks for that. I won't be able to test until Monday GMT+10 at the earliest, but it looks like those are just different references to the same exception, so it should be fine.

SakshiSharma-India:
Hi @amercader ,
I have faced the similar problem in ckan version 2.7.2. 
When I keep on adding large number of resource files one by one into the dataset, I get the below error for some of the resource files:
error : "ERROR [ckan.logic] Could not find resource <resource_id> after all"
And observed that the resource files with resource id mentioned in above error log disappeared from ckan GUI.

Is it possible that above error occurs due to race condition? If yes, I would like to confirm if it is possible that this PR is applicable to ckan v 2.7.2?
We have raised the similar issue https://github.com/ckan/ckan/issues/5298

amercader:
@SakshiSharma-India you should upgrade to 2.7.7 which is backwards compatible and includes this patch. The latest patch release is the only one we support and contains important security fixes.

SakshiSharma-India:
Hi @amercader ,
Thank you so much for your quick response.

PR #4926, opened 2019-07-25
Add unified(dot) name for webassets endpoint(`webassets.index`)

smotornyuk:
Otherwise, it would be different from our convention `blueprint.action` and will produce numerous problems in the future. For example, `toolkit.get_endpoint` takes endpoint's name, splits it by dot and returns two items as a tuple. In case of the default naming for flask routes, it would return `('webassets', )` for current webassets definition.

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -512,6 +512,6 @@ def _setup_webassets(app):
 
     webassets_folder = get_webassets_path()
 
-    @app.route('/webassets/<path:path>')
+    @app.route('/webassets/<path:path>', endpoint='webassets.index')
     def webassets(path):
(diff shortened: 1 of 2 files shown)


amercader:
This was caused by my change [here](https://github.com/ckan/ckan/pull/4913/files#diff-2265d7cafb26f7e70aff0fb6b3d7b09aR501), as I thought this wouldn't be needed any more. Do you think is worth adding it back to prevent failures like these in the future?
Or shall we just assume dot notation in the endpoint (which should be the case 99% of times)

smotornyuk:
I prefer to use standardized names with a mandatory dot symbol. 
1. I see no reason, why not to use such names
2. There will be fewer `if`s

amercader:
Sounds good!

PR #4927, opened 2019-07-25
4898 fix dataset search show all types

Engerrs:


smotornyuk:
Currently there only two possible states of all-type-search - it either disabled or enabled only for `dataset` type's page.
Could you add two distinct options:
1. Enable/disable search for all types. `False` by default and allows to see all datasets on `/dataset` page when set to `True`.
2. Control, which dataset type is used to show all datasets. By default it can be `dataset`, but it's likely that someone wants only `dataset` packages to be shown on `/dataset` page and it quite logical to create some fallback, like `/search` or `/all` for showing all datasets separately.

frafra:
That is a nice PR! I hope it can move further :)

smotornyuk:
Closing this one, because original behavior now works as expected:

`ckan.search.show_all_types = false` - don't show all dataset types on any search page
`ckan.search.show_all_types = true` - show all dataset types only on `/dataset` page
`ckan.search.show_all_types = custom_type` - show all dataset types only on `/custom_type` pate

PR #4929, opened 2019-07-26
More readable & detailed Changes page

hayley-leblanc:
The current "Changes" page shows a JSON diff between two consecutive versions of a dataset, which is difficult to read and not particularly user-friendly. This PR provides a much more readable and detailed summary of each change on the "Changes" page. The JSON diff is still viewable for each change, but it is not automatically displayed. This PR also adds the ability to view the changes that were made in multiple consecutive revisions of a dataset (see image below). 

change_page

- [ ] includes API changes


hayley-leblanc:
A couple of new tests fail because in order to make the Changes page more useful, I removed the lines preventing activity about private datasets from being saved.

amercader:
@hayley-leblanc Thanks for this, really exciting

@davidread you seem like the best placed to lead on the review of this

ckan/lib/activity_streams_session_extension.py
```diff
@@ -70,8 +70,8 @@ def before_commit(self, session):
             # object is a package.
 
             # Don't create activities for private datasets.
-            if obj.private:
-                continue
+            # if obj.private:
```
amercader: That's quite a big change, was it intended? I've always thought that private datasets should have activity streams too, only hidden to non authorized users. 
I guess the original issue was that if the dataset was eventually made public the previously private activities would be made public too. I'm not sure I find that an issue but if it is we could add a functionality to give an option to editors to clear the previous activities when making a dataset public.

@ckan/core what are your views on activities and private datasets?

TkTech: Agree the simple solution is just to ask at the time the dataset has its visibility changed, but the default action should be to just not touch the visibility of activity events rather than delete them. Activity is important for auditing for many groups.

Quick solution, add a migration along with this that adds a `private` boolean to each activity event and populates it with the value from the dataset at the time of migration. Activity events use the same logic as the dataset to show or hide if `private` is set.

Editors would be able to show or hide activity events individually or in bulk, but culling would requirer sysadmin?

hayley-leblanc: The PR definitely doesn't need private datasets to have activity streams - my organization mostly uses private datasets, so I made the change with these in mind, but it isn't necessary. I think the migration solution sounds good but I will probably not be able to implement it soon so it would be fine to not save activity data for private datasets until then.

davidread: Cool, in which case I suggest this is change is removed from this PR and it can be tackled in a separate PR

ckan/lib/changes.py
```diff
@@ -0,0 +1,771 @@
...
+
+def _title_change(change_list, original, new):
+    '''
+    Appends a summary of a change to a dataset's title between two versions
+    (original and new) to change_list.
+    '''
+    s = u""
+    seq2 = (u"<a href=\"", url_for(qualified=True, controller=u"dataset",
+            action=u"read", id=new['name']), u"\">",
+            new['title'], u"</a>")
```
davidread: I guess there are some issues with building HTML in code:
* Firstly it is a security red-flag, putting user strings straight in. Using Jinja2 would escape any fragments of HTML. You could use `jinja2.Template(u'<a href="{{url}}">{{title}}</a>').render(url=x, title=y)` to fix that. However it would still suffer from:
* The strings are not translated (internationalization). You could add _('some string'), however the string would not be added to the ckan.pot file (i.e. 'message extraction' - created with `python setup.py extract_messages`). It normally picks up strings from .html files, although it would be a pain to add dozens of little .html files. Perhaps others (@smotornyuk @tino097 ) know a way to extract messages from code?

ckan/lib/helpers.py
```diff
@@ -2688,3 +2688,62 @@ def sanitize_id(id_):
     ValueError.
     '''
     return str(uuid.UUID(id_))
+
+
+@core_helper
+def compare_pkg_dicts(original, new, old_activity_id):
+    '''
+    Takes two package dictionaries that represent consecutive versions of
+    the same dataset and returns a list of detailed & formatted summaries of
```
davidread: Perhaps mention that it returns HTML

ckan/templates/package/changes.html
```diff
@@ -7,28 +7,47 @@
...
   {{ super() }}
   <li>{% link_for _('Changes'), named_route='dataset.activity', id=pkg_dict.name %}</li>
-  <li class="active">{% link_for activity_diff.activities[1].id|truncate(30), named_route='dataset.changes', id=activity_diff.activities[1].id %}</li>
+  <li class="active">{% link_for activity_diffs[0].activities[1].id|truncate(30), named_route='dataset.changes', id=activity_diffs[0].activities[1].id %}</li>
 {% endblock %}
 
 {% block primary %}
   <article class="module">
     <div class="module-content">
+      <div style="width:250px;position:absolute;right:30px;">
```
davidread: Let's put all style in CSS files so that it is overridable by extensions

ckan/templates/package/snippets/change_item.html
```diff
@@ -0,0 +1,42 @@
...
+<input type="button" onClick="show_metadata()" class="btn" value="Show metadata diff" id="metadata_button"></input>
+<div id="metadata_diff" style="display:none;">
+{% block package_changes_diff %}
+  <pre>
+    {{ activity_diff['diff']|safe }}
+  </pre>
+{% endblock %}
+</div>
+
+<script>
```
davidread: Again, let's put JS in .js files

ckan/templates/snippets/activities/changed_package.html
```diff
@@ -17,6 +17,8 @@
         <a href="{{ h.url_for('dataset.changes', id=activity.id) }}">
           {{ _('Changes') }}
         </a>
+        <br>
+        Activity ID: {{ activity.id }}
```
davidread: I'm not clear there is a use need to display an ID and would scare many users. I would have thought the timestamp is enough as a reference?

ckan/lib/changes.py
hayley-leblanc: I think I've fixed this with commit 7ccd68a12d100e6650225a52809efd6037bcd14d; let me know if that commit doesn't fix these issues.

ckan/templates/snippets/activities/changed_package.html
hayley-leblanc: Yup, that's left over from an older version where users chose the range of changes to view using the activity ID. It doesn't need to be there now.

wardi:
As written how easy is this to extend for custom metadata schemas? When this PR lands I'd like to extend its templates from scheming so that custom field changes show up nicely along with the default ones.

hayley-leblanc:
@wardi This PR already checks for changes to custom fields and has a generic format to display them alongside the changes to default metadata. It would definitely be possible to add in new templates for changes to specific custom fields if you want the changes to be displayed in a certain way, although it might require some editing of the CKAN core; I was mainly focused on changes to the default metadata when I wrote this PR.

davidread:
I reckon an extension point could easily be added into `helpers.py:compare_pkg_dicts()` by someone, to do further analysis on the changes to the pkg_dict and add them to `change_list`. The extension can then provide the additional templates as: `"snippets/changes/{}.html".format(change.type)`. This could be done in a subsequent PR.

ckan/lib/helpers.py
```diff
@@ -2688,3 +2688,66 @@ def sanitize_id(id_):
...
+    all of the default package dictionary keys, and also checks for additional
+    keys added by extensions and custom fields added by the user in the web
+    interface.
+
+    Returns a list of dictionaries, each of which corresponds to a change
+    to the dataset made in this revision. The dictionaries each contain a
+    string indicating the type of change made as well as other data necessary
+    to form a detailed summary of the change.
+    '''
+    from changes import _check_metadata_changes, _check_resource_changes
```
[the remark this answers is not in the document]
davidread: BTW I've changed this in: https://github.com/ckan/ckan/commit/45a48efecb8a12eb7e5c167b97cba907e311479b

ckan/lib/helpers.py
```diff
@@ -2688,3 +2688,66 @@ def sanitize_id(id_):
...
+    '''
+    select_list = []
+    for activity in pkg_activity_list:
+        entry = render_datetime(activity['timestamp'],
+                                with_hours=True,
+                                with_seconds=True)
+        if activity['id'] == current_activity_id:
+            select_list.append(
+                "<option value=\"" + activity['id'] +
+                "\" selected>" + entry + "</option>"
```
davidread: again, building HTML is a red flag. I've done an alternate version here: https://github.com/ckan/ckan/commit/c973f229d which you can cherry-pick

ckan/lib/changes.py
davidread: yep this is good!

ckan/templates/snippets/changes/resource_format.html
```diff
@@ -0,0 +1,37 @@
+<li>
+  <p>
+    {% if ah.method == "add" %}
+      {{ _('Set format of resource <a href="{resource_url}">{resource_name}
+          </a> to <a href="{format_url}">{format}</a> in
+          <a href="{pkg_url}">{dataset}</a>').format(
```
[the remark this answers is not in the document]
hayley-leblanc: It looks like if I use the `{% trans %}` tag instead of `_(`, the formatting of the string to translate might look a little bit nicer, but it still has a bunch of `<a href=...>` tags in. I could take those links out, but I thought that the ability to navigate to the referenced dataset/resource/tag/etc. from the summary of changes would be useful.

ckan/templates/package/snippets/change_item.html
```diff
@@ -0,0 +1,10 @@
+<strong>On {{ h.render_datetime(activity_diff.activities[1].timestamp, with_hours=True, with_seconds=True) }}, {{ h.linked_user(activity_diff.activities[1].user_id) }}:</strong>
+
+{% set changes = h.compare_pkg_dicts(activity_diff.activities[0].data.package, activity_diff.activities[1].data.package, activity_diff.activities[0].id) %}
+<ul>
+{% for change in changes %}
+  {% snippet "snippets/changes/{}.html".format(
+    change.type), ah=change %}
```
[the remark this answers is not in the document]
hayley-leblanc: I like this idea, except that I wonder if it might make dealing with things like changed tags, extra fields, etc. a bit more difficult? I guess for any case where we need to find a list of changed tags/fields/keys/etc. we could still do that in changes.py and still pass that information to the templates that need it. I'm going to keep it as is for now but I think this would be a good change to make in the future.

hayley-leblanc:
All the tests pass now, except for one (test_remove_org), which I cannot for the life of me get to pass. I can't replicate the issue either; the same case works just fine in my instance of CKAN.

smotornyuk:
I was able to reproduce fail locally and noticed that the one, that unsets dataset organization is failing. `update_package_schema` has `ignore_missing` as first validator and action just assumes that you don't want to update the organisation looking at its new value. Something, like direct model change would work:
```python
    @helpers.change_config(u'ckan.auth.create_unowned_dataset', True)
    def test_remove_org(self):
        changes = []
        old_org = Organization()
        original = Dataset(owner_org=old_org['id'])

        import ckan.model as model
        pkg = model.Package.get(original['id'])
        pkg.owner_org = None
        pkg.save()

        new = helpers.call_action(
            u'package_show', id=original['id'])

        check_metadata_changes(changes, original, new)

        eq(changes[0]['type'], u'org')
        eq(changes[0]['method'], u'remove')
```
but if there are no particular reason, why you put `owner_org` here, it would be easier to just test another field.

davidread:
Many thanks @smotornyuk ! I've added that test fix and opened a fresh PR to merge this (because that is simpler than asking @hayley-leblanc to add it to this one): https://github.com/ckan/ckan/pull/4969

davidread:
This is merged now 🎉  @hayley-leblanc many thanks for doing the hard work to meet all our CKAN standards - this is a really valuable contribution 👍

amercader:
Thanks @hayley-leblanc! Really exciting to see such a valuable contribution. And thanks @davidread for a proactive and thorough review

PR #4938, opened 2019-08-07
Additional info URLs can be clickable

xbian:
Fixes #

### Features:

- [ ] includes API changes


wardi:
This change looks better implemented in a theme plugin for your site. You can override just that `extras` block in a template using `{% ckan_extends %}`.

Making all extras that look like links clickable (and marked in rdf as the homepage, and opening a new window) isn't something we want as a default behaviour.

PR #4945, opened 2019-08-21
Upgrade werkzeug

davidread:
### Proposed fixes:
Just a few subtle test changes needed:

With this upgrade of werkzeug, `url_for` now returns unicode, rather than a str. I'm not clear exactly where that happens, but flask uses lots of bits of werkzeug underneath. This caused test failures like this:

```
  File "/vagrant/src/ckan/ckan/tests/controllers/test_admin.py", line 34, in _reset_config
    app.post(
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webtest/app.py", line 838, in post
    content_type=content_type)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webtest/app.py", line 810, in _gen_request
    expect_errors=expect_errors)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webtest/app.py", line 1102, in do_request
    res = req.get_response(app, catch_exc_info=True)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/request.py", line 1049, in get_response
    application, catch_exc_info=True)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webob/request.py", line 1022, in call_application
    app_iter = application(self.environ, start_response)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webtest/lint.py", line 151, in lint_app
    check_environ(environ)
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/webtest/lint.py", line 335, in check_environ
    % (key, type(environ[key]), environ[key]))
AssertionError: Environmental variable PATH_INFO is not a string: <type 'unicode'> (value: u'/ckan-admin/reset_config')
```

This exception occurs because WSGI requires environ.PATH_INFO to be a string (bytes, not unicode). This is being checked by webtest.lint (the newest version does the same check). PATH_INFO is the part of the WSGI request storing the URL's path. WebOb is the thing that saves it, but our old version, it keeps it in whatever encoding we give it (rather than encoding it so it is a str). The upshot of all this is that when we pass a URL to our test app, it needs to be an (encoded) str.

I'm not 100% sure that UTF8 is the correct encoding - it might need url (percent) encoding, but for these tests it doesn't make any difference.

### Features:

- [ ] includes API changes
- [x] includes sec fix for possible backport

### DIFF
--- a/ckan/tests/controllers/test_home.py
+++ b/ckan/tests/controllers/test_home.py
@@ -44,7 +44,7 @@ class TestHome(helpers.FunctionalTestBase):
         response = app.get(url=url_for('home.index'), extra_environ=env)
 
         assert 'update your profile' in response.body
-        assert url_for('user.edit') in response.body
+        assert str(url_for('user.edit')) in response.body
         assert ' and add your email address.' in response.body
 
     def test_email_address_no_nag(self):
--- a/ckan/tests/helpers.py
+++ b/ckan/tests/helpers.py
@@ -155,6 +155,10 @@ class CKANTestApp(webtest.TestApp):
             self._flask_app = self.app.apps['flask_app']._wsgi_app
         return self._flask_app
 
+    def post(self, url, *args, **kwargs):
+        url = url.encode('utf8')  # or maybe it should be url_encoded?
+        return super(CKANTestApp, self).post(url, *args, **kwargs)
+
 
 def _get_test_app():
     '''Return a webtest.TestApp for CKAN, with legacy templates disabled.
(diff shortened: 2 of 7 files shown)


davidread:
This important upgrade just needs a bit of test adjustments, so hopefully uncontroversial :)

amercader:
@davidread sorry, what was making me doubt is [this change](https://github.com/ckan/ckan/pull/4945/files#diff-4ced38bb014e9352e73434775fd7bb50R74) where the URL generated in the test itself was converted to `str`. We do that in dozens of other places in the tests so I was worried that we had to do the change across the board. But as the rest are passing I think we are good to go.

davidread:
Many thanks @amercader :)

Issue #4946, opened 2019-08-23
Maintainance release of 2.7.x


malthe:
A new release is required since 2.7.6 is not installable using pip (see issue #4271).

amercader:
A new release cycle for just this change seems unjustified (`pip install ckan` is not one of the standard supported installations). But don't worry, we will include the requirements-setuptools.txt change in the next patch release.

PR #4947, opened 2019-08-23
Include requirement-setuptools.txt in manifest

### LINKED ISSUE #4271: Unable to pip install ckan

2.8.0

I tried `pip install ckan` so that my linter would stop complaining (my actual ckan is in a VM) and I expected it to install CKAN.

It appears that requirement-setuptools.txt is not part of the pypi package.

```
 File "<string>", line 1, in <module>
      File "/private/tmp/pip-install-eaBfYZ/ckan/setup.py", line 31, in <module>
        with open(os.path.join(HERE, 'requirement-setuptools.txt')) as f:
    IOError: [Errno 2] No such file or directory: '/private/tmp/pip-install-eaBfYZ/ckan/requirement-setuptools.txt'
```

Try running `pip install ckan`

malthe:
When attempting to `pip install ckan` it fails as the setup.py requires the requirement-setuptools.txt to be present, and it is not.  This change makes sure that the pip install will not fail due to the absence of this file.

This PR does not include the requirements.txt as this would likely require further discussion about how requirements are loaded.

This fixes issues #4271, #4946.

### Features:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,9 @@ RUN apt-get -q -y update && apt-get -q -y upgrade && DEBIAN_FRONTEND=noninteract
 		python-dev \
         python-pip \
         python-virtualenv \
+        libffi-dev \
         libpq-dev \
+        libssl-dev \
         git-core \
 	&& apt-get -q clean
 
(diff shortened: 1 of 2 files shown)


amercader:
That sounds good. Can you create a new pull request targetting master, and without the unrelated Dockerfile changes? We will backport it to the previous release branches

malthe:
@amercader this _is_ the backport. The Dockerfile change is necessary for the build to succeed.

PR #4952, opened 2019-08-27
Do not reuse connection if invalidated on usage

malthe:
This exception handler ensures that a connection that's invalidated as a result of using it (for example, the connection may no longer be alive), is correctly closed and removed from the connection pool.

Previously, a defunct connection would be left in the pool and reused indiscriminately.

### Features:
 [ ] includes tests covering changes


malthe:
Actually, I realized that this error mode is something you'll probably want to handle within the WSGI stack (e.g. using [repoze.retry](https://repozeretry.readthedocs.io/en/stable/)).

PR #4961, opened 2019-08-29
DatastoreController -> Datastore blueprint

smotornyuk:


ckanext/datapusher/tests/test.py
```diff
@@ -23,11 +23,13 @@ class TestDatastoreCreate():
 
     @classmethod
     def setup_class(cls):
+        cls._original_config = dict(config)
```
amercader: This is a bit dodgy but this whole file needs to be rewritten anyway

PR #4962, opened 2019-08-29
[#4420] Fix Solr query for non-private datasets

### LINKED ISSUE #4420: Dataset counts incorrect on Groups listing

bellisk:
The assembled query was malformed, leading to Solr ignoring the main part of the query as passed in via the data_dict.

For instance, when searching for a group to display the number of datasets in it, the part of the query specifying the group was missed out. This meant that all groups displayed the total number of datasets, instead of the number in each group.

Fixes #4420 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1884,7 +1884,8 @@ def package_search(context, data_dict):
         include_drafts = asbool(data_dict.pop('include_drafts', False))
         data_dict.setdefault('fq', '')
         if not include_private:
-            data_dict['fq'] = '+capacity:public ' + data_dict['fq']
+            data_dict['fq'] = 'capacity:public' if not data_dict['fq'] \
+                else 'capacity:public +' + data_dict['fq']
         if include_drafts:
             data_dict['fq'] += ' +state:(active OR draft)'
 
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1888,6 +1888,8 @@ def package_search(context, data_dict):
                 else 'capacity:public +' + data_dict['fq']
         if include_drafts:
             data_dict['fq'] += ' +state:(active OR draft)'
+        if not include_drafts and not context.get('sysadmin'):
+            data_dict['fq'] += ' -state:draft'
 
         # Pop these ones as Solr does not need them
         extras = data_dict.pop('extras', None)
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1479,7 +1479,7 @@ def user_show(context, data_dict):
     if asbool(data_dict.get('include_datasets', False)):
         user_dict['datasets'] = []
 
-        fq = "+creator_user_id:{0}".format(user_dict['id'])
+        fq = "creator_user_id:{0}".format(user_dict['id'])
 
         search_dict = {'rows': 50}
 
--- a/ckan/tests/logic/action/test_get.py
+++ b/ckan/tests/logic/action/test_get.py
@@ -1498,7 +1498,7 @@ class TestPackageSearch(helpers.FunctionalTestBase):
 
         fq = "(creator_user_id:{0} AND +state:draft)".format(user['id'])
         results = logic.get_action('package_search')(
-            {'user': sysadmin['name']},
+            {'user': sysadmin['name'], 'sysadmin': True},
             {'fq': fq})['results']
 
         names = [r['name'] for r in results]
--- a/ckan/tests/logic/action/test_get.py
+++ b/ckan/tests/logic/action/test_get.py
@@ -1295,6 +1295,23 @@ class TestPackageSearch(helpers.FunctionalTestBase):
 
         eq(len(results), 0)
 
+    def test_package_search_with_fq_for_org_includes_only_org_datasets(self):
+        '''
+        Searching for datasets by organization returns only datasets from the
+        organization.
+        '''
+        user = factories.User()
+        org1 = factories.Organization(user=user)
+        org2 = factories.Organization(user=user)
+        dataset1 = factories.Dataset(user=user, owner_org=org1['name'], name="dataset1")
+        factories.Dataset(user=user, owner_org=org2['name'], name="dataset2")
+
+        fq = "owner_org:" + org1['id']
+        results = helpers.call_action('package_search', fq=fq)['results']
+
+        eq(len(results), 1)
+        nose.tools.assert_equal(results[0]['name'], dataset1['name'])
+
     def test_package_search_with_include_drafts_option_excludes_drafts_for_anon_user(self):
         '''
         An anon user can't user include_drafts to get draft datasets.


bellisk:
Hi @smotornyuk, could you review this please? From my pov it's ready.

bellisk:
Hi @smotornyuk is there anything else I need to do for this PR to be ready to review?

ckan/logic/action/get.py
```diff
@@ -1479,7 +1479,7 @@ def user_show(context, data_dict):
     if asbool(data_dict.get('include_datasets', False)):
         user_dict['datasets'] = []
 
-        fq = "+creator_user_id:{0}".format(user_dict['id'])
+        fq = "creator_user_id:{0}".format(user_dict['id'])
```
smotornyuk: There is no sense in ommitting `+` as we are searching for datasets of this particular user

ckan/logic/action/get.py
```diff
@@ -1884,9 +1884,12 @@ def package_search(context, data_dict):
         include_drafts = asbool(data_dict.pop('include_drafts', False))
         data_dict.setdefault('fq', '')
         if not include_private:
-            data_dict['fq'] = '+capacity:public ' + data_dict['fq']
+            data_dict['fq'] = 'capacity:public' if not data_dict['fq'] \
+                else 'capacity:public +' + data_dict['fq']
```
smotornyuk: Same here. What is the purpuse of `+` removal

ckan/logic/action/get.py
```diff
@@ -1884,9 +1884,12 @@ def package_search(context, data_dict):
...
         if include_drafts:
             data_dict['fq'] += ' +state:(active OR draft)'
+        if not include_drafts and not context.get('sysadmin'):
```
smotornyuk: It would be better to use `context['auth_user_obj']` instead of unneccessary `sysadmin` key

ckan/logic/action/get.py
```diff
@@ -1884,9 +1884,12 @@ def package_search(context, data_dict):
...
+            data_dict['fq'] += ' -state:draft'
```
smotornyuk: We already have default filter by active state  - https://github.com/ckan/ckan/blob/master/ckan/lib/search/query.py#L260 
How is this different from one from the link?

ckan/logic/action/get.py
```diff
@@ -1884,9 +1884,12 @@ def package_search(context, data_dict):
         include_drafts = asbool(data_dict.pop('include_drafts', False))
         data_dict.setdefault('fq', '')
         if not include_private:
-            data_dict['fq'] = '+capacity:public ' + data_dict['fq']
+            data_dict['fq'] = 'capacity:public' if not data_dict['fq'] \
```
smotornyuk: so, if `data_dict['fq']` is falsy, wouldn't it be safe to just use `'+capacity:public ' + data_dict['fq']` everytime?

ckan/tests/logic/action/test_get.py
```diff
@@ -1498,7 +1515,7 @@ def test_package_search_with_fq_for_create_user_id_will_include_drafts_for_other
 
         fq = "(creator_user_id:{0} AND +state:draft)".format(user['id'])
         results = logic.get_action('package_search')(
-            {'user': sysadmin['name']},
+            {'user': sysadmin['name'], 'sysadmin': True},
```
smotornyuk: There is no sense in new context parameters while they can be obtained from existing ones without overhead(`auth_user_obj`)

smotornyuk:
Hi. No, just was pretty busy recently, sorry. Now it's reviewed

ckan/logic/action/get.py
bellisk: This is the change that was necessary to fix the issue https://github.com/ckan/ckan/issues/4420. The old line created the query `+capacity:public [whatever was in the data dict]`, for example `+capacity:public group:operations`. The `+` at the start of the query is unnecessary because it's only needed to join parts of the query together, not in front of the whole query. There's also no `+` before the second part of the query, which meant it just got ignored. This is why searching for the number of datasets in a particular group returned the number of datasets in *all* groups instead.

ckan/logic/action/get.py
bellisk: The `+` at the start of the query was a workaround for the fact that because of the error at https://github.com/ckan/ckan/pull/4962#discussion-diff-325791829L1887, whatever you put into the `fq` here would be ignored. With the change there, the `+` is no longer needed (and in fact causes a malformed query with `++`).

ckan/logic/action/get.py
bellisk: No, because the `+` between the two parts of the query is necessary. We also can't use `'+capacity:public +' + data_dict['fq']` every time, because if there's nothing in `data_dict[fq]`, this causes a dangling `+` which also breaks the query.

ckan/logic/action/get.py
bellisk: I added this line because a test was breaking--I think it was https://github.com/ckan/ckan/blob/cf7f4cc04204a4bb14e140b0e7e9297787f35f6a/ckan/tests/logic/action/test_get.py#L1278, testing that you can't search with `fq = "state:draft"` and see draft datasets. Previously, this test passed, because the `state:draft` part of the query was just ignored as described above. Adding this line fixed it.

Looking at the test again, though, I think something is still wrong. The test doesn't include `'sysadmin': True` as in https://github.com/ckan/ckan/pull/4962/files/b1a31168334a54092938b1dd62a956ba475beac2#diff-c92cdbf4ff7391d90f31341452b90303R1518, and that's why the search in the test doesn't output the draft datasets.

ckan/logic/action/get.py
bellisk: Thanks, will do that instead.

ckan/logic/action/get.py
bellisk: (This error got into the code when someone tried to switch around the query to make `capacity:public` the first part of it instead of the second, and made a typo. https://github.com/ckan/ckan/commit/107bc894d05d8f2fc0fbcbeddc4b444c20267139)

ckan/logic/action/get.py
smotornyuk: Plus sign is just a signal for Solr that this filter `MUST` match occurrence. If it results in malformed query, we should look at the place where code assumes that there is no `+` in `fq` and adds second `+`. This line itself doesn't contain any contradiction and this change IS implicit fix of but that IS NOT created by this line.

ckan/logic/action/get.py
smotornyuk: Plus sign inside query doesn't have any `joining` behavior - it's Solr's operator that declares requirement for the following filter. The `+` at the start of the query IS NOT unnecessary - it filters out completely all results that do not contain given `field:value` but otherwise receive high match score because of other query parts.  By writing `else '...  +' ...` you are introducing bug with duplicated `+` that you have mentioned previously.
If there is no `+` before one of `fq` part, it just becomes rather desired, than required, but it's not ignored. Compare [example without plus sign](https://demo.ckan.org/dataset?q=%2Btitle%3ALondon+title%3AAir) and [with plus sign](https://demo.ckan.org/dataset?q=%2Btitle%3ALondon+%2Btitle%3AAir). Both are returning one dataset at the moment i'm writing this comment. But if you remove `title:Air` part completely ([example](https://demo.ckan.org/dataset?q=%2Btitle%3ALondon)) - you'll get two datasets. If parts without `+` are ignored, you would receive the same result in first link

bellisk: Thanks for the detailed explanation, I can see where I got things wrong here now. I will close this PR and make a new one with a proper fix in it.

PR #4966, opened 2019-09-03
4797 replace paste converters

### LINKED ISSUE #4797: Replace paste.deploy converters
This should be straightforward. We use paste.deploy's `asbool`, `asint`, `aslist` (specially `asbool`) in different parts of the code and they are exposed on the toolkit. Basically we need to implement our own versions of them and replace the references

#### Approach

* Write our implementation of the three functions based on the [current](https://github.com/Pylons/pastedeploy/blob/9cbc584a86bd1ae425bd0c0c3c00617d7e59475d/paste/deploy/converters.py) ones. These need to be imported early in the execution, so they should be probably placed in `ckan/common.py` or `ckan/lib/base.py`.
* Replace all imports in the code base from `paste.deploy.converters` to the new location.
* Replace the [references](https://github.com/ckan/ckan/blob/master/ckan/plugins/toolkit.py#L216:L230) in the toolkit module.

#### Story points

3

f-osorio:
Fixes #4797 
Need to replace external converters with something internal.

### Proposed fixes:
Recreate converters in CKAN and update import references.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -208,3 +208,37 @@ request = CKANRequest(_get_request)
 # Provide a `c`  alias for `g` for backwards compatibility
 g = c = LocalProxy(_get_c)
 session = LocalProxy(_get_session)
+
+# replacement for converters that previously came from paste.deploy.compat [#4797]
+truthy = frozenset(['true', 'yes', 'on', 'y', 't', '1'])
+falsy = frozenset(['false', 'no', 'off', 'n', 'f', '0'])
+
+def asbool(obj):
+    if isinstance(obj, basestring):
+        obj = obj.strip().lower()
+        if obj in truthy:
+            return True
+        elif obj in falsy:
+            return False
+        else:
+            raise ValueError("String is not true/false: {}".format(onj))
+    return bool(obj)
+
+def asint(obj):
+    try:
+        return int(obj)
+    except (TypeError, ValueError):
+        raise ValueError("Bad integer value: {}".format(obj))
+
+def aslist(onj, sep=None, strip=True):
+    if isinstance(obj, basestring):
+        lst = obj.split(sep)
+        if strip:
+            lst = [v.strip() for v in lst]
+        return lst
+    elif isinstance(obj, (list, tuple)):
+        return obj
+    elif obj is None:
+        return []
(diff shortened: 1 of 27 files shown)


ckan/common.py
```diff
@@ -208,3 +208,41 @@ def _get_session():
 # Provide a `c`  alias for `g` for backwards compatibility
 g = c = LocalProxy(_get_c)
 session = LocalProxy(_get_session)
+
+# replacement for converters that previously came from paste.deploy.converters
+# [#4797]
```
amercader: We can get this information from the commit message, so no need to add a comment to source code

ckan/common.py
```diff
@@ -208,3 +208,41 @@ def _get_session():
...
+truthy = frozenset([u'true', u'yes', u'on', u'y', u't', u'1'])
+falsy = frozenset([u'false', u'no', u'off', u'n', u'f', u'0'])
+
+
+def asbool(obj):
+    if isinstance(obj, basestring):
```
amercader: Can you change this to the following to ensure Python 3 compatibility?

```suggestion
    if isinstance(obj, six.string_types):
```

That should fix the Travis CI failure as well

ckan/common.py
```diff
@@ -208,3 +208,41 @@ def _get_session():
...
+
+def asint(obj):
+    try:
+        return int(obj)
+    except (TypeError, ValueError):
+        raise ValueError(u"Bad integer value: {}".format(obj))
+
+
+def aslist(obj, sep=None, strip=True):
+    if isinstance(obj, basestring):
```
amercader: Same as above

test-core.ini
```diff
@@ -16,11 +16,11 @@ testing = true
 
 
 # Specify the Postgres database for SQLAlchemy to use
-sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_test
+sqlalchemy.url = postgresql://ckan_default:ckan@localhost/ckan_test
```
amercader: Please don't change the defaults on this file as it will affect existing scripts

amercader:
All good, thanks a lot @f-osorio!

PR #4977, opened 2019-09-10
Stats Controller -> Blueprint

smotornyuk:
In addition, I've removed cache from `ckanext.stats.stats` as previously it relied on pylons caching. I think that this feature should be restored only after a decision about caching in CKAN v3.  We can continue using beaker as is in the future, but maybe it's a good moment to check other cache libraries or add an additional layer of abstraction over beaker in order to support tag-based/context-based cache. As far as I know, currently beaker supports only key-based cache with an optional expiration time, which isn't enough for implementing reliable and flexible caching in the portal.


smotornyuk:
Yes, I forgot about our convention to remove extension's controllers as they unlikely were extended by plugins.

PR #4978, opened 2019-09-10
DataTablesView Controller -> Blueprint

smotornyuk:


smotornyuk:
I've used incorrect names for asset groups during webassets migration - that's why JS was not loading on the preview. 
Now i removed custom version of bootstrap, that was loaded by datatablesview and added `super` call inside styles block instead, so that main bootstrap will be used. I suppose, originally it was there because we were using bs-2, but as i remember, we decided to remove bs-2 support in CKAN 2.9, so this fallback may be dropped. I left folder with custom bootstrap there, as it seems that some of JS lines relies on this particular minor version, but it would be cool to make cleanup here as well(or consider something more agnostic, like [ag-grid](https://www.ag-grid.com/)

wardi:
great work @smotornyuk !

PR #4980, opened 2019-09-10
Adding "require_auth" config variable

hardingalexh:
Fixes #

### Proposed fixes:

This change adds a `require_auth` config variable, which defaults to false.
If this variable is set to `true`, `NotAuthorized` is raised before any checks are passed to authz/custom auth.

This change also includes adding a link to login to the error template if the error code is 403, and additionally fixes a small bug in the `user` view where the `_` variable was being overwritten.

I'm submitting this now for feedback, but would appreciate review on the approach and what would be needed to include this, whether it be tests or additional documentation.

### Features:

- [ ] includes API changes


wardi:
The config variable name could be clearer, how about `allow_anonymous` instead?

Auth changes in particular need tests, take a look at https://github.com/ckan/ckan/blob/master/ckan/tests/controllers/test_package.py for examples of checking for a 403 response and use the `helpers.change_config` decorator to set your config variable

hardingalexh:
> The config variable name could be clearer, how about `allow_anonymous` instead?
> 
> Auth changes in particular need tests, take a look at https://github.com/ckan/ckan/blob/master/ckan/tests/controllers/test_package.py for examples of checking for a 403 response and use the `helpers.change_config` decorator to set your config variable

How about `allow_anonymous_access`? I can make that change and default `true`, then look into adding tests. What would you say is sufficient coverage? Seeing as how this should block a lot, I'm wondering what angle I should take for this. I can unit test the `check_access` function, but probably not everywhere it's called.

wardi:
`allow_anonymous_access` sounds good. We have an `auth_allow_anonymous_access` decorator for auth functions, we should document that this setting will override all use of that decorator.

https://github.com/ckan/ckan/blob/master/doc/maintaining/configuration.rst

How about a unit test for `check_access` then test a couple users to see that they are blocked properly, e.g. the dataset search page returning a 403, and the `user_list` action raising a `NotAuthorized` exception.

hardingalexh:
@wardi I've made the refactors discussed, added documentation and a few tests - but I'm not fully understanding why the two tests failing are failing. Mind taking a look?

hardingalexh:
@wardi @amercader are either of you available to help debug the tests here?

doc/maintaining/configuration.rst
```diff
@@ -643,6 +643,20 @@ Default value: ``False`` (however the default config file template sets it to ``
 Restricts access to 'view this version' and 'changes' in the Activity Stream pages. These links provide users with the full edit history of datasets etc - what they showed in the past and the diffs between versions. If this option is set to ``False`` then only admins (e.g. whoever can edit the dataset) can see this detail. If set to ``True``, anyone can see this detail (assuming they have permission to view the dataset etc).
 
 
+.. _ckan.auth.allow_anonymous_access:
+
+ckan.auth.allow_anonymous_access
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
boykoc: It seems one of the build errors is caused here at line 649. Can you remove the extra `^` so it's just 
```
ckan.auth.allow_anonymous_access
^^^^^^^^^^^^^^^^^^^^^^^
```

ckan/tests/logic/action/test_get.py
```diff
@@ -768,6 +768,12 @@ def test_user_list_default_values(self):
         assert 'email' not in got_user
         assert 'datasets' not in got_user
 
+    @helpers.change_config('ckan.auth.allow_anonymous_access', 'false')
+    def test_user_list_raises_not_authorized_if_config_allow_anon_access_false(self):
+        user = factories.User()
+        nose.tools.assert_raises(logic.NotAuthorized, helpers.call_action,
```
boykoc: user_list will also depend on the ckan.auth.public_user_details = False config. It defaults to true so this would have to be set first as well.

boykoc:
Same areas are still causing build errors.  Adding my notes to hopefully help other save some time when reviewing.

I tried to look at previous PRs on the configuration.rst file to see if these were ignored when merging or if the issue come up. It's flagging `AssertionError: Don't add any new warnings to the Sphinx build: [u'/root/project/doc/maintaining/configuration.rst:649: WARNING: duplicate label ckan.auth.allow_anonymous_access, other instance in /root/project/doc/maintaining/authorization.rst']`. But I checked the authorization.rst and only see that it includes the configuration.rst which is maybe causing the error?  So it's a bit of a false-positive if that's the case but others will know more.

Second one `AssertionError: NotAuthorized not raised` error at the new test `  File "/root/project/ckan/tests/logic/action/test_get.py", line 776, in test_user_list_raises_not_authorized_if_config_allow_anon_access_false
    'user_list')`

taking a closer look I think it should be booleans not strings based in:
```
    @helpers.change_config('ckan.auth.allow_anonymous_access', False)
    @helpers.change_config('ckan.auth.public_user_details', False)
```

PR #4982, opened 2019-09-11
4799 extension template documentation

### LINKED ISSUE #4799: Don't use paste.deploy to create the extension template
We use the following command to bootstrap a new extension:

```
paster --plugin=ckan create -t ckanext ckanext-iauthfunctions
```

As with the rest of paste we should look at options to replace it. The best option by far seems to be [Cookiecutter](https://cookiecutter.readthedocs.io/en/latest/index.html), with extensive support and lots of feature. This will allow us to host the template extension in a separate repo (like [this examples](https://cookiecutter.readthedocs.io/en/latest/readme.html#python)), so we can improve it separately from core.

#### Approach

Create a separate Cookiecutter recipe for CKAN extensions that covers the same parts as the current ones (and improves it if possible)

#### Story points

5

f-osorio:
Fixes #4799 

Provides a new template for extensions using [Cookiecutter](https://cookiecutter.readthedocs.io/en/latest/index.html) instead of `Paster`. 

This only updates the documentation for the extension tutorial. The cookecutter [template is here](https://github.com/f-osorio/cookiecutter-ckan-extension) -- I wasn't sure where it should go since part of the reason for moving to cookiecutter seems to be to decouple it from CKAN core. The documentation only has a place holder to refer to the location until it's finalized.

### DIFF
--- a/doc/extensions/tutorial.rst
+++ b/doc/extensions/tutorial.rst
@@ -440,8 +440,7 @@ group named ``curators``.
 
    If you've already created a ``curators`` group and want to test what happens
    when the site has no ``curators`` group, you can use CKAN's command line
-   interface to :ref`clean and reinitialize your database
-   <database management>`.
+   interface to :ref:`clean and reinitialize your database <database management>`.
 
 Try visiting the ``/group`` page in CKAN with our ``example_iauthfunctions``
 plugin activated in your CKAN config file and with no ``curators`` group in
--- a/doc/extensions/tutorial.rst
+++ b/doc/extensions/tutorial.rst
@@ -71,7 +71,7 @@ contains the CKAN git repo. Let's use the parent directory instead:
 
 Now run the ``paster create`` command to create your extension::
 
-    cookiecutter https://github.com/<user>/<repo>.git
+    cookiecutter https://github.com/ckan/ckan/<path-to-template>.git
 
 The command will present a few prompts. The information you give will
 end up in your extension's ``setup.py`` file (where you can edit them later if
@@ -79,7 +79,7 @@ you want).
 
 .. note::
 
-   The first prompt is for is the name of your next
+   The first prompt is for the name of your next
    extension. CKAN extension names *have* to begin with ``ckanext-``. This
    tutorial uses the project name ``ckanext-iauthfunctions``.
    The prompts for ``project_shortname`` and ``plugin_class_name`` can be
--- a/doc/extensions/tutorial.rst
+++ b/doc/extensions/tutorial.rst
@@ -44,14 +44,20 @@ Creating a new extension
    CKAN config file to activate the extension's features.
 
 
-You can use the ``paster create`` command to create an "empty" extension from
+You can use ``cookiecutter`` command to create an "empty" extension from
 a template. First, activate your CKAN virtual environment:
 
 .. parsed-literal::
 
    |activate|
 
-When you run the ``paster create`` command, your new extension's directory will
+Install the ``cookiecutter`` python module:
+
+.. parsed-literal::
+
+   pip install cookiecutter
+
+When you run ``cookiecutter``, your new extension's directory will
 be created in the current working directory by default (you can override this
 with the ``-o`` option), so change to the directory that you want your
 extension to be created in. Usually you'll want to track your extension code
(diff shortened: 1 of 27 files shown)


smotornyuk:
As mentioned in the original issue, one of the profs of separating template and core is separate maintenance and(implicitly) ability to quickly add fixes/backports/shims to extension template, without waiting for new CKAN release. But, maybe, it would be nice to have(in addition to the explicit `cookiecutter URL` example) built-in command for template generation? It's unlikely that devs will remember template's URL and they will constantly pick this value from documentation(which may be irritating for those, who prefer to execute a sequence of simple commands). For example, if we would have `ckan generate extension` CLI command with hardcoded URL for template and ability to specify the name of the custom template and few additional options, then devs can just inspect CLI options(via `ckan --help`) and find out everything they need. It will create a basis for the generation of other kinds of artifacts, like `actions` and `plugins`.

f-osorio:
I think having an option to use a CLI command makes a lot of sense. There's some limitation in `cookiecutter` that introduce some awkwardness for the user that this could smooth over. I can work on that, but should it be in this pull request or a new one?

wardi:
The trouble with using a separate repo for the template is that anything other than ckan/ckan gets much less attention and testing from the core team. It's easy for a change in core to lead to a change in the best practices for new extensions, so I think it's better for the extension template to stay in core so that the changes can be made at the same time on the same branch.

If someone would like to maintain a separate template for extensions that is designed to work across a wide range of ckan versions (e.g. by importing ckan/ckantoolkit and only using "safe" ways of accessing core features) I fully support that effort, but that's a different problem than the one the default extension templates should solve.

amercader:
I agree with @wardi. @f-osorio Here's my suggested plan for this:

1. Include your template in this PR. Current template is in `ckan/pastertemplates`. Perhaps is best to put the new one in `contrib/cookiecutter/ckan_extension`
2. Add cookiecutter to `dev-requirements.txt`
3. Update the docs if needed

This should be the MVP. If you want to build a dedicated command to make it easier for users that's great but not a blocker for merging. It can go in this PR or a separate one. It should be a ["new-style"](https://github.com/ckan/ckan/issues/4639) command, eg `ckan extension create`. There are details on that issue about how to set them up and plenty of already done examples in `ckan/cli`

amercader:
Forgot to say thanks a lot for taking this on @f-osorio!

davidread:
I tried it out and it works a treat, @f-osorio :)

A couple of v. minor things:
* Spelling of 'seperate' -> 'separate' https://github.com/f-osorio/cookiecutter-ckan-extension/blob/master/cookiecutter.json#L3
* Compatibility is python 2.7 only, for the time being https://github.com/f-osorio/cookiecutter-ckan-extension/blob/master/%7B%7Bcookiecutter.project%7D%7D/setup.py#L46

doc/extensions/tutorial.rst
```diff
@@ -65,18 +71,20 @@ contains the CKAN git repo. Let's use the parent directory instead:
...
 end up in your extension's ``setup.py`` file (where you can edit them later if
 you want).
 
+.. note::
+
+   The first prompt is for is the name of your next
+   extension. CKAN extension names *have* to begin with ``ckanext-``. This
+   tutorial uses the project name ``ckanext-iauthfunctions``.
+   The prompts for ``project_shortname`` and ``plugin_class_name`` can be
+   ignored. They will be filled in automatically.
```
davidread: The things mentioned in this note seem relatively clear to me during use. Does including the note just lengthen the docs unnecessarily?
```
$ cookiecutter https://github.com/f-osorio/cookiecutter-ckan-extension.git
project [Project's name (must begin with 'ckanext-')]: ckanext-hierarchy
keywords [A space seperated list of keywords]: tag1 tag2
description []: This is a great project!
author []: David
author_email []: david.read@hackneyworkshop.com
github_user_name [Github user or organization name]: davidread
project_shortname [hierarchy]:
plugin_class_name [HierarchyPlugin]:
```

f-osorio: If it's clear then I agree that it's unnecessary. I wasn't sure how clear it is partly because it seems to me to be a little unintuitive to get a prompt for something that should be ignored. But there's no way around that.

`Cookiecutter` doesn't currently support having templates in subdirectories of repos. But there is a [PR for it and it's mostly through review](https://github.com/cookiecutter/cookiecutter/pull/1224).

wardi: If cookiecutter doesn't support using a local path for a template then I'm fine with using another repo under the ckan org (kind of annoying for users to need internet access just to create an extension, but not worth maintaining a fork of cookiecutter)

f-osorio:
My mistake. I was thinking about someone using `cookiecutter` to get the template through a URL, but since the template should already be local it is possible to use a local path. I've updated the documentation to refer to this. 

There's also a CLI command in place. It's based on @smotornyuk suggestion, `ckan generate extension`, so that it can be expanded to generate other things should the need arise in the future.

The documentation right now goes through the process of using `cookiecutter <path/to/template>` but mentions the CLI. I wasn't sure how best to handle having two methods, and I couldn't find a place in the documentation that covers CLI.

There are some failures in the CI checks due to the syntax used by `cookiecutter`. I don't think there is anything that can be done about them within `cookiecutter`. The syntax is essential to how it works.

smotornyuk:
Perfect. As for thests, you can add template directory to whitelist - https://github.com/ckan/ckan/blob/dfad5d98a644a76939a57872073ef27dc7c68d86/ckan/tests/test_coding_standards.py#L32 . This should fix two of three failures. And third error should be fixed by listing problematic files here -https://github.com/ckan/ckan/blob/dfad5d98a644a76939a57872073ef27dc7c68d86/ckan/tests/legacy/test_coding_standards.py#L303

f-osorio:
I've updated the documentation to include using both `cookiecutter ...` and the CLI `ckan generate extension`. 

I added an exception to `.travis.yml` to ignore the directory `contrib/cookiecutter`. It seemed to be the only way to get the flake8 check to ignore the kind of weirdnesses that cookiecutter needs.

smotornyuk:
Cool! I don't see a problem with just ignoring this directory - this PR already does much more, that was supposed initially

Issue #3310, opened 2016-11-13
Replace nose with py.test


sirex:
[nose](https://nose.readthedocs.io/) is no longer supported:

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using [Nose2](https://github.com/nose-devs/nose2), [py.test](http://pytest.org/), or just plain unittest/unittest2.

It looks, that [py.test](http://pytest.org/) is the most active test runner project.

amercader:
> Can run unittest (including trial) and nose test suites out of the box

@sirex It looks like at least we would not have to rewrite all tests :)

The main blocker for that is that we are using the `--with-pylons` option on nose, which essentially loads up a CKAN environment. We would need to create a similar [plugin](http://docs.pytest.org/en/latest/writing_plugins.html#writing-plugins) for py.test.

If you or someone else is up for investigating a bit more please go ahead!

sirex:
@amercader if the team agrees to move to py.test, then I can look into that.

amercader:
@wardi @TkTech @davidread any objections? ^

TkTech:
None. nose works for us and will probably keep working, but it would be nice to get access to py.test plugins.

sirex:
If you plan to continue writing unittest style tests, then I guess it is not worth changing nose to py.test.

IMHO migration to py.test only worth the effort if tests gradually will be replaced with py.test style tests. For example this:

```python
class MyTest(unittest.TestCase):
    def setUp(self): # ...
    def tearDown(self): # ...
    def test_me(self):
        self.assertEqual(foo, 'bar')
```

should be replaced with this:

```python
def test_me(setup_teardown):
    assert foo == 'bar'
```

If tests will stay written in unittest style, then these important features won't work:

- powerful py.test fixtures
- better assertion error output
- integration with existing py.test plugins and fixtures

py.test does support unittest style tests, but it really shines only with py.test style tests.

wardi:
py.test style looks great to me

amercader:
I'm less excited now if this means that we will have tests written in two very different styles, with 90% of them written in the "old" nose style that should not be used for new tests. It's very unlickely that the current 1500 or so tests (or the half that are not in `legacy` rather) will be migrated to the new style. 
I guess it depends on how different are they though.

kmbn:
We decided to close old issues that are not actively worked on so that we can focus our effort and attention on issues affecting the current versions of CKAN.

[...]

PR #4996, opened 2019-09-24
nose -> pytest

smotornyuk:
Long-time ago there was discussion related to test-runners(#3310). `nose` is not supported and thus we have a reason for switching into something new. In addition, current test complexity and constant problems with config/app initialization during tests just freaking me out. Long story short, I've decided to configure `pytests`.


smotornyuk:
Yay, only 30 failures after initial run

amercader:
Really hoping this turns out well, :heart: pytest

amercader:
This is fanstastic work @smotornyuk. I'm still going through the main changes but one thing that caught my eye was that you are applying fixtures at the individual test level:

```python
  class TestResourceViewCreate(object):

      @pytest.mark.ckan_config("ckan.plugins", "image_view")
      @pytest.mark.usefixtures("clean_db", "with_plugins")  
      def test_resource_view_create(self):
          context = {} 
          params = self._default_resource_view_attributes()
  
          result = helpers.call_action("resource_view_create", context, **params)
  
          result.pop("id")
          result.pop("package_id")
  
          assert params == result

      @pytest.mark.ckan_config("ckan.plugins", "image_view")
      @pytest.mark.usefixtures("clean_db", "with_plugins")  
      def test_requires_resource_id(self):
          context = {} 
          params = self._default_resource_view_attributes()
          params.pop("resource_id")
  
          with pytest.raises(logic.ValidationError):
              helpers.call_action("resource_view_create", context, **params)

```

This is fine of course but given our reliance on test classes I think that being able to apply fixtures at the class level makes the tests more readable and avoids duplication:

```python
  @pytest.mark.ckan_config("ckan.plugins", "image_view")
  @pytest.mark.usefixtures("clean_db", "with_plugins")
  class TestResourceViewCreate(object):
  
      def test_resource_view_create(self):
          context = {} 
          params = self._default_resource_view_attributes()
  
          result = helpers.call_action("resource_view_create", context, **params)
  
          result.pop("id")
          result.pop("package_id")
  
          assert params == result
  
      def test_requires_resource_id(self):
          context = {} 
          params = self._default_resource_view_attributes()
          params.pop("resource_id")
  
          with pytest.raises(logic.ValidationError):
              helpers.call_action("resource_view_create", context, **params)

```

To make this work I had to slightly change the `ckan_config` fixture:

```diff
diff --git a/ckan/tests/pytest_ckan/fixtures.py b/ckan/tests/pytest_ckan/fixtures.py
index cc8521f56..216231ff9 100644
--- a/ckan/tests/pytest_ckan/fixtures.py
+++ b/ckan/tests/pytest_ckan/fixtures.py
@@ -56,7 +56,7 @@ def ckan_config(request, monkeypatch):
 
     """
     _original = config.copy()
-    for mark in request.node.own_markers:
+    for mark in request.node.iter_markers():
         if mark.name == u"ckan_config":
             monkeypatch.setitem(config, *mark.args)
     yield config

```

Of course we can combine both approaches depending on the context but perhaps we should favour marking tests at the class level. WDYT?

Of course I am assuming that changing all the references again would be a big task so maybe we can do it later

smotornyuk:
The more we'll do now, the less we need to rework in future :smile: I'll try to move fixtures to class level wherever it possible - it not so hard with regexps and macros

amercader:
@smotornyuk I had a good look at this and as far as I'm concerned this is good to go.
I changed a bit the docstrings and documentation, and added some examples in https://github.com/ckan/ckan/pull/4996/commits/3056beed7952c9957385bb0a3402601794216002#diff-fb126fa3f292fcb42a73fc636c96eb88R267, which I can add when merging.

Let me know when you are done with the class level changes.

BTW are [these changes](https://github.com/ckan/ckan/pull/4996/commits/3056beed7952c9957385bb0a3402601794216002#diff-fb126fa3f292fcb42a73fc636c96eb88R267) intended?

amercader:
On thing I've noticed though is that the tests are now noticeably slower (~6-7 mins with nose, ~10 min with pytest). We can investigate and improve it later, it might be related with the collection phase or the way we use fixtures.

smotornyuk:
>>> BTW are these changes intended?

I'm actively using search-replace-with-regexp(sed, awk, etc.) here, and it, unfortunately, breaks code style in a number of places. In order to solve this problem, I'm formatting tests with `black` so that I don't have to expand the PEP8 blacklist. That's why you can see changes that are not directly related to tests - anyway, I'm doing a lot of changes, so I decided to do it alongside with some formatting.

>>> tests are now noticeably slower

As for slow tests, unfortunately, it's not the collection phase but tests itself. Locally I'm using `pytest-randomly` in order to run tests in a random order every time. In this way, I discovered a fantastically huge number of stateful tests that will just fail if they won't be executed in a particular order. It prevents us from splitting tests between instances(to some degree) and does not allow(if someone will try to use this `pytest`'s feature in future) running tests in parallel. 
In order to make tests independent, I've added a lot of `db_clean` fixtures - they make tests a bit slower. And the main culprit here is `CreateTestData` - previously it often was called once per file and tests, that were executed in particular order, were mutating this test data. Now I'm creating test data for each individual test in order to preserve independence. I don't think that we should increase speed by using stateful tests - it would be much better to refactor/create only required data/rethink tests instead.

amercader:
Both things make sense. I totally agree that is better to have tests without state leaking that the speed gains. We can refactor them later.

Let me know when you are done with this PR and I'll merge

smotornyuk:
Done. I've merged fixes from your `pytest-2` branch into this one

amercader:
nose tests are no more, long live pytest

PR #4999, opened 2019-09-24
Add missing closing div tag to template

jrdh:
Fixes odd page rendering problem where the content of the page sits under the secondary content.

For example:

before

and with this fix:

after

### Proposed fixes:
Add closing `div` tag.

### Features:

- [ ] includes API changes


jrdh:
Oddly, this doesn't seem to be a problem in a fresh ckan install so must have something to do with our custom styling. Might as well still fix it though.

PR #5005, opened 2019-09-27
roles_that_cascade_to_sub_groups doc - fixes #5004

### LINKED ISSUE #5004: Clarification about ckan.auth.roles_that_cascade_to_sub_groups
I have some confusion about `ckan.auth.roles_that_cascade_to_sub_groups`.

The [documentation](https://docs.ckan.org/en/latest/maintaining/authorization.html#ckan-auth-roles-that-cascade-to-sub-groups) makes it sound like this applies to groups, i.e. what you use to create collections of datasets, e.g. for a common topic or related in another way - **not** organizations:

> Makes role permissions apply to all the groups down the hierarchy from the groups that the role is applied to.

> e.g. a particular user has the ‘admin’ role for group ‘Department of Health’. If you set the value of this option to ‘admin’ then the user will automatically have the same admin permissions for the child groups of ‘Department of Health’ such as ‘Cancer Research’ (and its children too and so on).

However, the observed behavior and [source code](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L699) seems to indicate that this actually applies to *organizations*, not *groups*.
Furthermore, groups don't even have a hierarchy (unlike organizations), or do they?

Is this correct, that this setting affects organizations? If it is, I think the documentation should be updated accordingly.

jeverling:
The documentation was referring to "groups", while this setting applies to "organizations".

Fixes #5004

### Features:

### DIFF
--- a/doc/maintaining/configuration.rst
+++ b/doc/maintaining/configuration.rst
@@ -607,9 +607,11 @@ Example::
 Default value: ``admin``
 
 
-Makes role permissions apply to all the groups down the hierarchy from the groups that the role is applied to.
+Makes organization permissions apply to all the organizations down the hierarchy from the organization that the role is applied to.
 
-e.g. a particular user has the 'admin' role for group 'Department of Health'. If you set the value of this option to 'admin' then the user will automatically have the same admin permissions for the child groups of 'Department of Health' such as 'Cancer Research' (and its children too and so on).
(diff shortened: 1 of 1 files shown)


amercader:
@jeverling while I agree that the main use case for hierarchical relations are organizations, this feature is also supported on groups, as you can see in the [relevant](https://github.com/ckan/ckan/blob/master/ckan/authz.py#L292:L296) [code](https://github.com/ckan/ckan/blob/master/ckan/model/group.py#L212). So something like this would be more accurate:

> Makes role permissions apply to all the groups or organizations down the hierarchy from the groups or organizations that the role is applied to.

amercader:
Done in cb0e838c1

PR #5008, opened 2019-10-01
Fix PostGIS base image tag

frafra:
Using `:latest` as tag (the default) is a bad idea, especially for a database like Postgres, where a manual upgrade of the database is required when moving to a major release. People sticking to a specific CKAN version could get a non-working database just because they rebuilt the Docker images (for security reasons) and they will end up with a different Postgres version. This should be avoided.

This patch fixes future problem by blocking automatic upgrades to new major releases of Postgres.

### Features:

- [ ] includes API changes

### DIFF
--- a/contrib/docker/postgresql/Dockerfile
+++ b/contrib/docker/postgresql/Dockerfile
@@ -1,5 +1,5 @@
 #FROM postgres:9.6
-FROM mdillon/postgis
+FROM mdillon/postgis:11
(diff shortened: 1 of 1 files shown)


frafra:
I would also suggest to backport that patch to 2.8, especially now that PostgreSQL 12 has been released and the PostGIS image will follow, causing potentially more issues related to that.

PR #5020, opened 2019-10-11
Bug caused by updated to how Flask handles the file upload field

shasan-marsdd:
… filename for the file_url field when I'm not actually uploading a new file, just changing other metadata (such as description).  This effectively kills kills the uploaded image link for the organization/groups.

Fixes #

### Proposed fixes:
Issue occurs only when I have an image for the organization/groups.  I need to be able to update other metadata but not uploading a new image.  It breaks due to the fact how old Pylons requests object not returning anything in the "image_upload" field when the user doesn't upload a new image.  Flask's FileStorage class has a boolean check instead where we can just do if "data_dict['image_upload']" or similar instead.  
I have kept the original if statement in place to be backwards compatible, additional check added in order to account for the Flask method.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/uploader.py
+++ b/ckan/lib/uploader.py
@@ -144,8 +144,9 @@ class Upload(object):
         if not self.storage_path:
             return
 
-        if isinstance(self.upload_field_storage, (ALLOWED_UPLOAD_TYPES)):
-            if self.upload_field_storage:
+        if isinstance(self.upload_field_storage, (ALLOWED_UPLOAD_TYPES,)):
+            if self.upload_field_storage \
+               and self.upload_field_storage.filename:
                 self.filename = self.upload_field_storage.filename
                 self.filename = str(datetime.datetime.utcnow()) + self.filename
                 self.filename = munge.munge_filename_legacy(self.filename)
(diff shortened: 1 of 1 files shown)


smotornyuk:
Given the fact that we are migrating away from Pylons, my comment may be ignored, but there is one issue with Pylons' controllers with this fix. You are checking file instance using implicit boolean casting, which has quite unexpected implementation for cgi.FieldStorage (https://bugs.python.org/issue19097) and you can get unexpected results for Pylons uploads unless you check with explicit `is not None`

shasan-marsdd:
Since I need to care about actual file upload, and in both cases I have checked the "filename" field instead, which is where the bug is actually occurring.  This ultimately should continue to work whether we use cgi.FieldStorage or FlaskFileStorage class.  It also reduces need to individually check based on each instance, keeping things cleaner (hopefully).

shasan-marsdd:
> Given the fact that we are migrating away from Pylons, my comment may be ignored, but there is one issue with Pylons' controllers with this fix. You are checking file instance using implicit boolean casting, which has quite unexpected implementation for cgi.FieldStorage (https://bugs.python.org/issue19097) and you can get unexpected results for Pylons uploads unless you check with explicit `is not None`

I did some changes, Not sure if that's ok with you?

smotornyuk:
It works pretty well for me, so i'm merging this PR

PR #5024, opened 2019-10-14
4673 allow reply to emails

### LINKED ISSUE #4673: RFE: allow Reply-To header on emails

2.7.3

We would like to send system emails that come from a real address (so we can catch bounces etc), but which don't reply to a real address when used by humans (ie use a 'no-reply' address as the Reply-To header).

Only the 'From' address is configurable.

f-osorio:
Fixes #4673

### Proposed fixes:

Add ability to set a 'Reply-to' value for out going system emails/notifications.

Update to the creation of the CKAN configuration file, `deployment.ini_tmpl`, which adds the option `smtp.reply_to`. By default it's empty and commented out like `smtp.mail_from`.

The value is used in `mailer.py`'s `_mail_recipient` function to set a `Reply-to`, if there is a non-blank value. Otherwise, no `Reply-to` is added and everything functions as it has been.

### Features:

### DIFF
--- a/ckan/config/deployment.ini_tmpl
+++ b/ckan/config/deployment.ini_tmpl
@@ -194,6 +194,7 @@ ckan.hide_activity_from_users = %(ckan.site_id)s
 #smtp.user = username@example.com
 #smtp.password = your_password
 #smtp.mail_from =
+#smtp.reply_to =
 
 ## Background Job Settings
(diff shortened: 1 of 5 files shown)


amercader:
Thanks @f-osorio for the patch and @ThrawnCA for the review

PR #5032, opened 2019-10-18
Fix resource.url being saved as "___"

### LINKED ISSUE #5031: Resource URL gets changed to ___
latest master

### The probem
When I create or edit a resource using the web form, whatever I set the resource.url to be, it gets saved as something like this: `http://myckansite.com/dataset/713c8c32-cbd8-46ce-90fb-86de1f0811d2/resource/59c0e90c-11e2-4d7f-af43-e098660745dc/download/___` (!)
It shouldn't change it. And the URL it sets it to gives a 404.

davidread:
Fixes #5031

[werkzeug upgrade 0.14 to 0.15](https://github.com/ckan/ckan/pull/4945/files) means that when you submit the resource form with a resource url (not an upload), and the uploader is enabled, then the 'upload' parameter value changes from being u'' to: FileStorage(filename=u''). This half-confused ResourceUpload into thinking that it should upload something and it stored the url as '___' (!). This is fixed by essentially testing bool(resource['upload']) because bool(FileStorage(filename=u'')) == False.

I was unable to test this using webtest - it didn't submit the form like my real browser does. So I just add a test for ResourceUpload that shows the issue and prevents it regressing in future.

- [ ] includes API changes


davidread:
The test failures show this is not a good enough fix. Perhaps someone with fresh eyes can take a look?

amercader:
@davidread @smotornyuk the fix was good enough, but some tests needed updating to use the Werkzeug based FileStorage wrapper. All green now.
I worked on this while looking at #5152 which is somehow related.

davidread:
Wonderful! Thanks v much @amercader

PR #5043, opened 2019-10-29
Translation

SerhiiRI:
### Features:
- added Polish translations to web interface.


amercader:
@SerhiiRI thanks a lot for this! We manage our translations in the [Transifex](https://www.transifex.com/okfn/ckan) platform. It is really easy to register and use online, and it simplifies the translations management on our side.
Can you upload your po file in the Polish translation of the 2.8 resource? This should automatically import all your translated strings to the Transifex platform, and we will then download the updated translations before the next release.

Screenshot_2019-10-29 CKAN Resources

Thanks again!

PR #5046, opened 2019-10-31
Fix wrong _ function reference in user blueprint

amercader:
This was causing an exception as it override the reference to the i18n `_` function below

```
  File "/srv/app/src/ckan/ckan/views/user.py", line 87, in before_request
    base.abort(403, _(u'Not authorized to see this page'))
TypeError: 'unicode' object is not callable
```


amercader:
@smotornyuk an easy one on the user blueprint

smotornyuk:
Yep, my bad - i've used to assign any garbage to underscore

PR #5054, opened 2019-11-05
webhelpers.feedgenerator -> feedgen

smotornyuk:


amercader:
@smotornyuk after merging this I realized that there still a `webhelpers.feedgen` import in views/feed.py because of the [`_FixedAtom1Feed(webhelpers.feedgenerator.Atom1Feed)`](https://github.com/ckan/ckan/blob/master/ckan/views/feed.py#L568) class definition. I assume this whole class can be deleted now right?

tino097:
i think it is safe

smotornyuk:
Yes, i didn't notice it, because it's not used anywhere, so nothing failed during checks:)

PR #5060, opened 2019-11-08
Implementation of one_of method in CKAN

### LINKED ISSUE #4800: Remove or replace formencode support
Although it's not explicitly listed as a requirement in `requirements.in` (because it's a Pylons requirement), formencode is used in two places:

* Its `OneOf` validator is added to the standard CKAN ones [here](https://github.com/ckan/ckan/blob/f43d6a572838c792193f3239827d04f9ffea9206/ckan/logic/__init__.py#L677).
* There is support for defining formencode validators in the CKAN schema ([docs](https://docs.ckan.org/en/2.8/extensions/adding-custom-fields.html?highlight=formencode#custom-validators), [code](https://github.com/ckan/ckan/blob/b9e45e2723d4abd70fa72b16ec4a0bebc795c56b/ckan/lib/navl/dictization_functions.py#L219:L233))

We need to remove the requirement, because it hasn't been updated in 3 years and its setup.py only lists support up until Python 3.5.

We use `OneOf` in core and someone might be using it in an extension so we need to implement it, but I doubt that anyone is using the second feature, and that we should drop it.

In the future we can add similar support for a better maintained validation library like [WTForms](https://wtforms.readthedocs.io/).

#### Approach

* Write our own implementation of `OneOf` in `validators.py` (and rename it to `one_of` for consistency, but keeping the old name for backwards compatibility).
* Remove all formencode bits from the code base.

#### Story points

2

iamarnavgarg:
Fixes #4800


ckan/logic/validators.py
```diff
@@ -858,3 +858,20 @@ def email_validator(value, context):
         if not email_pattern.match(value):
             raise Invalid(_('Email {email} is not a valid format').format(email=value))
     return value
+
+def one_of(value,context):
```
amercader: We can not use `context` to provide the iterable as there is no way to set it in the first place, and it is used internally by CKAN (in fact we don't need the context parameter at all). What we can do is to provide a wrapper that accepts a parameter, in the same way that for instance [`if_empty_same_as`](https://github.com/ckan/ckan/blob/master/ckan/lib/navl/validators.py#L37) does.

```python
def one_of(list_of_values):
    def callable(value):
         if value not in list_of_values:
             raise Invalid(_('Value must be one of {}'.format(list_of_values)))
         return value
```
This will be then use like this in the schemas:

```python
    {
         'some_custom_field': [ignore_missing, one_of(['bear', 'lion', 'tiger']), unicode_safe),
    }

```

ckan/logic/validators.py
```diff
@@ -858,3 +858,20 @@ def email_validator(value, context):
...
+   ''' Implementation of OneOf method
+       for Python Library Formencode
+   '''
+
+   if (value == '' or value == None):
```
amercader: You might have `''` or `None` in your list of allowed values, so this should not raise an exception.

ckan/logic/validators.py
```diff
@@ -858,3 +858,20 @@ def email_validator(value, context):
...
+      raise Invalid(_('A value must be provided'))
+
+   else:
+    if value:
+      for x in context:
```
amercader: It's better to use the `value in list` operator. See my example above.

ckan/tests/logic/test_validators.py
```diff
@@ -667,4 +667,31 @@ def call_validator(*args, **kwargs):
             errors[key] = []
             call_validator(key, {key: url}, errors, None)
 
+class TestOneOfValidator(object):
```
amercader: These are great, thanks for adding them. You'll need to rewrite them to use the callable described above instead. Something like:

```python
def test_value_in_list(self):
    func = validators.one_of(['a', 'b', 'c'])
    assert_equals(func('a'), 'a')

def test_value_not_in_list(self):
    func = validators.one_of(['a', 'b', 'c'])
    raises_Invalid(func)('d')
```

ckan/tests/logic/test_validators.py
```diff
@@ -1,3 +1,7 @@
+<<<<<<< HEAD
+
+=======
+>>>>>>> 5021e632b0c0e434e00f100e3c68a1ec4ab9558c
```
iamarnavgarg: Hi @amercader can you please suggest as to why this commit hash has been added into the file on its own? 
I did the git push command and while doing so, it showed some error and then again, I had added the file in git add and pushed.

amercader: @iamarnavgarg looks like a merge error. Just delete these four lines and push again

ckan/logic/validators.py
```diff
@@ -858,3 +858,14 @@ def email_validator(value, context):
...
             raise Invalid(_('Email {email} is not a valid format').format(email=value))
     return value
+
+def one_of(list_of_value):
+   ''' Implementation of OneOf method
+       for Python Library Formencode
+   '''
+   def callable(value):
+      if value not in list_of_values:
+	    raise Invalid(_('Value must be one of {}'.format(list_of_value)))
```
amercader: You need an indent here

ckan/tests/logic/test_validators.py
```diff
@@ -667,4 +671,17 @@ def call_validator(*args, **kwargs):
             errors[key] = []
             call_validator(key, {key: url}, errors, None)
 
+class TestOneOfValidator(object):
+
+      def test_val_in_list(self):
+        cont = [1,2,3,4]
```
amercader: Please check your editor configuration, we use 4 spaces for indentation

ckan/tests/logic/test_validators.py
```diff
@@ -667,4 +671,17 @@ def call_validator(*args, **kwargs):
...
+        func = validators.one_of(cont)
+        assert_equals(func(1),1)
+
+      @raises_Invalid
+      def test_val_not_in_list(self):
+        cont = [1,2,3,4]
+        y = validators.one_of(cont)
+	raises_invalid(func)(5)
```
amercader: Indent, plus `y` should be `func`

ckan/logic/__init__.py
```diff
@@ -674,7 +674,7 @@ def get_validator(validator):
         _validators_cache.update(validators)
         validators = _import_module_functions('ckan.logic.validators')
         _validators_cache.update(validators)
-        _validators_cache.update({'OneOf': formencode.validators.OneOf})
+        _validators_cache.update({'OneOf':'ckan.logic.validators.one_of'})
```
amercader: You need the actual reference to the function, not a string. So `{'OneOf': ckan.logic.validators.one_of}`

amercader:
Thanks @iamarnavgarg!
 I added a few improvements on ad8825046 and 2d5217c10

PR #5062, opened 2019-11-08
Replace usage of map unsupported on Python 3

amercader:
In Python 3, `map` returns a map object not a list. Replaced the occurrences I could find with list comprehensions.


amercader:
@cclauss do you know if there's a flake8 test to catch this type of code?

cclauss:
You did the right thing... manually convert to list comprehensions (or generators where applicable).

I do not think flake8 will help you detect these instances but you can try:
* __python-modernize --fix=map .__  # add -w to write the changes
* __pip install flake8-comprehensions__  # adds new rules to flake8 on when to use comprehensions

amercader:
Thanks for the pointers @cclauss !

cclauss:
Always happy to help ckan.

Issue #4801, opened 2019-05-23
Implement changes to allow running CKAN in both Python 2 and 3


amercader:
During our transition period the same code base needs to be able to run in both Python 2 and 3. Pylons will still be there to allow people to completely migrate their extensions, which means that there will be two different sets of requirements:

* `requirements.py2.in`: Including Pylons and whatever Pylons 2 only stuff that still needs to be there
* `requirements.in`: Python 3 only requirements, which will be the ones installed by default in the documentation for new instances

Once we have installed the Python 3 requirements in a py3 virtualenv then it's just a matter of running `ckan run` and keep fixing things until CKAN works :sweat_smile:. And then do the same until all tests pass.

We will find two main issues:

1. Python 3 incompatible code. Despite all the great work done so far there are still things that we have missed (eg relative imports, `map()` function returning an iterator rather than a list, etc). These should be easy to pick up and fix.
2. Conditional imports and execution.  There will be some imports and code that only need to be run on Python 2. We can decide the best way to define this conditional loading (the `if six.PY2` bit) but essentially it boils down to things like:

    ```python
    # ckan/common.py
    
    import six

    import flask

    if six.PY2:
        import pylons
        from pylons.i18n import (ugettext as pylons_ugettext,
                                 ungettext as pylons_ungettext)
        from pylons import response
    
    # ...
 
    def is_flask_request():
        '''
        A centralized way to determine whether we are in the context of a
        request being served by Flask or Pylons
        '''
        if six.PY3:
            return True
        
        # ...
    
    ```

Luckily as part of the Flask migration work a lot of imports for the most used objects were centralized in a few modules, so hopefully these hacks will be limited in scope (and remember that they should be removed in the next version).

I had a quick go a this and I quickly bumped into VDM, as it is used very early on the initialization, so VDM is definitely the major blocker on getting this important work started.

#### Approach

* Create two different sets of requirements for Python 2 / 3 envs - see #5063
* Load Python 2 requirements conditionally
* Fix Python 3 compatibility issues
* Get CKAN starting up when running `ckan run` (and to serve some pages) in Python 3
* Get all tests passing in Python 3

#### Story points

13

davidread:
Nice work @amercader and @smotornyuk!

Can this issue be closed then?

amercader:
Yes! Any new py3 compatibility issues can be created separately

Issue #4793, opened 2019-05-23
Flask and Python 3 support on selected extensions


amercader:
At least for the most used extensions on the ckan organization we should make sure they are ready for a Python 3 / Flask only version of CKAN. @davidread compiled a list of candidates and potential coordinators in #4681. It unlikely that the persons listed will be able to do all of them but if we do a couple of the critical ones first and document all the steps in the documentation (see #XXXX) that is a task where the wider community can definitely help us. I'd suggest focusing on ckanext-scheming and ckanext-dcat first, one because it's pretty critical and widely used and the other because I've already made some work on it.

- [x] [ckanext-scheming](https://github.com/ckan/ckanext-scheming) - 3 - @smotornyuk
- [x] [ckanext-dcat](https://github.com/ckan/ckanext-dcat) - 5 - @smotornyuk
- [x] [ckanext-xloader](https://github.com/ckan/ckanext-xloader) - 2
- [x] [ckanext-harvest](https://github.com/ckan/ckanext-harvest) - 5 - @smotornyuk
- [x] [ckanext-spatial](https://github.com/ckan/ckanext-spatial) - 5 - @smotornyuk
- [x] [ckanext-pages](https://github.com/ckan/ckanext-pages) - 3
- [x] [ckanext-geoview](https://github.com/ckan/ckanext-geoview) - 2 - @smotornyuk
- [x] [ckanext-googleanalytics](https://github.com/ckan/ckanext-googleanalytics) - 3 - @smotornyuk
- [ ] [ckanext-qa](https://github.com/ckan/ckanext-qa) - 5 - @shubham-mahajan
- [x] [ckanext-showcase](https://github.com/ckan/ckanext-showcase) - 5 - @smotornyuk
- [x] [ckanext-pdfview](https://github.com/ckan/ckanext-pdfview) - 2 - @smotornyuk
- [x] [ckanext-archiver](https://github.com/ckan/ckanext-archiver) - 5 - @Zharktas 
- [x] [ckanext-fluent](https://github.com/ckan/ckanext-fluent) - 2
- [x] [ckanext-hierarchy](https://github.com/ckan/ckanext-hierarchy) - 2 - @smotornyuk
- [x] [ckanext-report](https://github.com/davidread/ckanext-report) - 2 - @Zharktas 

TODO
add syndication - @smotornyuk
add cloudstorage - @smotornyuk

#### Approach

For each extension:

* Check if any extension requirement is Python 2 only, replace if so
[...]

PR #5063, opened 2019-11-08
Add requirements.txt for python 3

davidread:
Part of #4801

In this PR:
* Existing python 2 requirements.txt/.in are renamed to make room for python3 ones
* Install-by-source instructions is for python 3
* Adds initial python2->3 upgrade instructions
* Initial adjustments to setup.py and ckan/common.py for import in python3

I'm keen to do 'trunk based development' for python 3 work now. So python 3 is the default in the docs and requirements.txt. I realize that the devs who use master for other work might be a bit surprised/disrupted by these changes. However I think it is worth it because the focus now needs to be python3, and it's better to switch master's default now rather than just before release.

I added the pip upgrade, because I found that upgrading pip from v8.1.1 to latest (v19.3.1) gets rid of benign but worrying red messages like this during `pip install -r requirements.txt`:
```
  Failed building wheel for psycopg2
  Running setup.py clean for psycopg2
  Running setup.py bdist_wheel for pyyaml ... error
  Complete output from command /usr/lib/ckan/default/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-89zlo4xy/pyyaml/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" bdist_wheel -d /tmp/tmpebovlc0zpip-wheel- --python-tag cp35:
  usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: -c --help [cmd1 cmd2 ...]
     or: -c --help-commands
     or: -c cmd --help

  error: invalid command 'bdist_wheel'
```

I've tested install-from-source.rst on a fresh Ubuntu 16.04 image, and I can install requirements.txt ok and get as far as `paster make-config ckan /etc/ckan/default/development.ini` but we don't have paster any more, so that is the next thing to crack after this PR.

### Features:

- [ ] includes API changes


amercader:
@davidread This all looks good to me code-wise. I had made basically the same changes on my own py3 branch.

The only thing I want to get people's views on is on defaulting to Python 3 for CKAN 2.9 (I'm less worried about the disruption it will cause for people targetting master). Our plan is to make 2.9 py3 ready so by definition CKAN will work on Python 3, but will it be helpful if the ecosystem of extensions (or at least the critical ones listed on #4793) does not support Python 3 yet? I guess we can make #4793 an explicit requirement for 2.9.

amercader:
Team agrees is a good idea to default to Python 3 for 2.9. Merging.

davidread:
Thanks @amercader

PR #5064, opened 2019-11-08
fix code injection in autocomplete module

usingsky:
Fixes #
Prevent code injection into autocomplete module
e.g.)
- make group with name "<script>alert('injection')</script>
- change the group of dataset.
- injection code is performed by pressing the autocomplete input element

### Proposed fixes:
Check markup in select2 library before rendering

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/autocomplete.js
+++ b/ckan/public/base/javascript/modules/autocomplete.js
@@ -186,15 +186,20 @@ this.ckan.module('autocomplete', function (jQuery) {
      *
      * Returns a text string.
      */
-    formatResult: function (state, container, query) {
-      var term = this._lastTerm || null; // same as query.term
+    formatResult: function (state, container, query, escapeMarkup) {
+      var term = this._lastTerm || (query ? query.term : null) || null; // same as query.term
 
       if (container) {
         // Append the select id to the element for styling.
         container.attr('data-value', state.id);
       }
 
-      return state.text.split(term).join(term && term.bold());
+      var result = [];
+      $(state.text.split(term)).each(function() {
+        result.push(escapeMarkup ? escapeMarkup(this) : this);
+      });
+
+      return result.join(term && (escapeMarkup ? escapeMarkup(term) : term).bold());
     },
 
     /* Formatter for the select2 plugin that returns a string used when


usingsky:
Fixed to use default escapeMarkup function.
refer to [formatResult](https://github.com/select2/select2/blob/170c88460ac69639b57dfa03cfea0dadbf3c2bad/select2.js#L1123)

ckan/public/base/javascript/modules/autocomplete.js
```diff
@@ -186,15 +186,21 @@ this.ckan.module('autocomplete', function (jQuery) {
...
+    formatResult: function (state, container, query, escapeMarkup) {
       var term = this._lastTerm || null; // same as query.term
 
       if (container) {
         // Append the select id to the element for styling.
         container.attr('data-value', state.id);
       }
 
-      return state.text.split(term).join(term && term.bold());
+      var result = state.text.split(term).join(term && term.bold());
```
boykoc: This is awesome, thanks for submitting this fix.  I've tested and it seems to work as expected and escapes various tags, including script tags and helps prevent DOM side XSS. 

One thing I noticed. The bolding of the auto complete term is being escaped as well. A user will see `<b>search term</b>` in the autocomplete drop down. Maybe the bold styling can be removed, or if necessary it could be added after the term is escaped?

usingsky: Thank you for the review. 
In the autocomplete module of the group page, this._lastTerm is always null, so I've changed it to use query.term.  And I modified the code to call escapeMarkup first before bold markup the term.

boykoc: 👍 Tested on my end, seems to work.

I've added a couple more comments just to document some of the things I noticed, not necessarily needing any action.

ckan/public/base/javascript/modules/autocomplete.js
```diff
@@ -186,15 +186,20 @@ this.ckan.module('autocomplete', function (jQuery) {
...
-      var term = this._lastTerm || null; // same as query.term
+    formatResult: function (state, container, query, escapeMarkup) {
+      var term = this._lastTerm || (query ? query.term : null) || null; // same as query.term
 
       if (container) {
         // Append the select id to the element for styling.
         container.attr('data-value', state.id);
       }
 
-      return state.text.split(term).join(term && term.bold());
```
boykoc: This original line is a bit confusing to me. I'm not sure what it was intended to do. I understand it as:

It seems to always create a 2 element array with empty strings then joins those with the bolded term.
`state.text.split(term) => ['', '']`
`['', ''].join(term && term.bold()) => <b>apples</b>`

logic:
`state.text` and `term` seem to be identical values. So the `state.text` is split on itself (e.g. `'apples'.split('apples')`) which gives an array with 2 empty elements. Then the array is joined by the term and if the term exists, bold it (e.g. `['', ''].join('apples' && 'apples'.bold())`, if the term is null it returns null.

ckan/public/base/javascript/modules/autocomplete.js
```diff
@@ -186,15 +186,20 @@ this.ckan.module('autocomplete', function (jQuery) {
...
+      var result = [];
+      $(state.text.split(term)).each(function() {
+        result.push(escapeMarkup ? escapeMarkup(this) : this);
+      });
+
+      return result.join(term && (escapeMarkup ? escapeMarkup(term) : term).bold());
```
boykoc: With the above comment in mind, you could probably be able to get away with removing lines 197 to 200, because its escaping empty strings. And then changing line 202 to something like `return state.text.split(term).join(term && (escapeMarkup ? escapeMarkup(term) : term).bold());`

But test before making changes, i've looked at it too long lol

ckan/public/base/javascript/modules/autocomplete.js
```diff
@@ -186,15 +186,20 @@ this.ckan.module('autocomplete', function (jQuery) {
...
      * Returns a text string.
      */
-    formatResult: function (state, container, query) {
-      var term = this._lastTerm || null; // same as query.term
+    formatResult: function (state, container, query, escapeMarkup) {
+      var term = this._lastTerm || (query ? query.term : null) || null; // same as query.term
 
       if (container) {
         // Append the select id to the element for styling.
         container.attr('data-value', state.id);
```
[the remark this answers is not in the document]
usingsky: "term" is the query text you entered.
In the case of the this._lasterm variable, it will be updated when the query is executed as an ajax request (for example, autocomplete on an add memeber page).

However, in case of autocomplete of add group page, All option values are updated before page rendering, not ajax request.
As a result, the query text you entered is updated in the "query.term" variable.

If you test without the patch, you will see that the query text is not bolded.(on add group page)

ckan/public/base/javascript/modules/autocomplete.js
usingsky: "state.text" means full text and "term" is the query text you entered.

eg) state.text : "<script>alert('test');</script>", term : alert
    after split : ["<script>", "('test');</script>"]
    after join : "<script><b>alert</b>('test');</script>"


eg) state.text : "<script>alert('test');</script>", term : null
    after split : ["<script>alert('test');</script>"]
    after join : "<script>alert('test');</script>"

ckan/public/base/javascript/modules/autocomplete.js
usingsky: In my opinion, xss issue will not be resolved without escaping state.text.

PR #5065, opened 2019-11-08
'map' usage made python3 compatible

davidread:
Used automated tool:

    python-modernize --fix=map .

As suggested: https://github.com/ckan/ckan/pull/5062#issuecomment-551903816
And did a little manual tidying.

#### flake8-comprehensions
I also tried the [flake8-comprehensions](https://pypi.org/project/flake8-comprehensions/) linting @cclauss suggested. I guess rule C400 "Unnecessary generator - rewrite as a list comprehension" is most relevant because it's to do with generators, however there were no violations in ckan code.

Also of interest are C401 "Unnecessary generator - rewrite as a set comprehension" and C402 "Unnecessary generator - rewrite as a dict comprehension", however set & dict comprehensions are not compatible with python 2, so they will have to wait.

I'll cover the others flake8-comprehensions in another PR because they seem less related - see https://github.com/ckan/ckan/pull/5066

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/api.py
+++ b/ckan/controllers/api.py
@@ -19,6 +19,7 @@ import ckan.lib.munge as munge
 from ckan.views import identify_user
 
 from ckan.common import _, c, request, response
+from six.moves import map
 
 
 log = logging.getLogger(__name__)
@@ -287,7 +288,7 @@ class ApiController(base.BaseController):
             return out
 
         query = query.limit(limit)
-        out = map(convert_to_dict, query.all())
+        out = list(map(convert_to_dict, query.all()))
         return out
 
     @jsonp.jsonpify
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -6,6 +6,7 @@ import logging
 
 from sqlalchemy.orm import class_mapper
 from six import string_types
+from six.moves import map
 
 import ckan.lib.dictization as d
 import ckan.lib.helpers as h
@@ -419,7 +420,7 @@ def group_dict_save(group_dict, context, prevent_packages_update=False):
     package_ids.extend( pkgs_edited['added'] )
     if package_ids:
         session.commit()
-        map( rebuild, package_ids )
+        list(map(rebuild, package_ids))
 
     return group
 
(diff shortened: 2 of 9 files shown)


ckan/controllers/api.py
```diff
@@ -287,7 +288,7 @@ def convert_to_dict(user):
             return out
 
         query = query.limit(limit)
-        out = map(convert_to_dict, query.all())
+        out = list(map(convert_to_dict, query.all()))
```
cclauss: out = [convert_to_dict(q) for q in query.all()]  # list comprehension

https://docs.python.org/3.0/whatsnew/3.0.html#views-and-iterators-instead-of-lists
> map() and filter() return iterators. If you really need a list, a quick fix is e.g. list(map(...)), but a better fix is often to use a list comprehension (especially when the original code uses lambda), or rewriting the code so it doesn’t need a list at all. Particularly tricky is map() invoked for the side effects of the function; the correct transformation is to use a regular for loop (since creating a list would just be wasteful).

ckan/lib/dictization/model_save.py
```diff
@@ -419,7 +420,7 @@ def group_dict_save(group_dict, context, prevent_packages_update=False):
     package_ids.extend( pkgs_edited['added'] )
     if package_ids:
         session.commit()
-        map( rebuild, package_ids )
+        list(map(rebuild, package_ids))
```
cclauss: [rebuild(package_id) for package_id in package_ids]

ckan/lib/helpers.py
```diff
@@ -1512,7 +1512,7 @@ def date_str_to_datetime(date_str):
         microseconds = int(m.groupdict(0).get('microseconds'))
         time_tuple = time_tuple[:5] + [seconds, microseconds]
 
-    return datetime.datetime(*map(int, time_tuple))
+    return datetime.datetime(*list(map(int, time_tuple)))
```
cclauss: return datetime.datetime(*list(int(item) for item in time_tuple))

ckan/model/license.py
```diff
@@ -33,7 +34,8 @@ def __init__(self, data):
         for (key, value) in self._data.items():
             if key == 'date_created':
                 # Parse ISO formatted datetime.
-                value = datetime.datetime(*map(int, re.split('[^\d]', value)))
+                value = datetime.datetime(
+                    *list(map(int, re.split('[^\d]', value))))
```
cclauss: *list(int(item) for item in re.split('[^\d]', value))

PR #5069, opened 2019-11-10
[#4801] Transform relative imports into absolute ones

amercader:
So code can run in Python 3

### DIFF
--- a/ckan/controllers/admin.py
+++ b/ckan/controllers/admin.py
@@ -8,8 +8,7 @@ import ckan.lib.app_globals as app_globals
 import ckan.lib.navl.dictization_functions as dict_fns
 import ckan.model as model
 import ckan.logic as logic
-import ckan.plugins as plugins
-from home import CACHE_PARAMETERS
+from ckan.controllers.home import CACHE_PARAMETERS
 
 
 c = base.c
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -25,7 +25,7 @@ import ckan.plugins as p
 import ckan.lib.render
 
 from ckan.common import OrderedDict, _, json, request, c, response
-from home import CACHE_PARAMETERS
+from ckan.controllers.home import CACHE_PARAMETERS
 
 log = logging.getLogger(__name__)
 
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -1295,8 +1295,7 @@ class CreateTestDataCommand(CkanCommand):
 
     def command(self):
         self._load_config()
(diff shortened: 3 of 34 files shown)


amercader:
I could use some help fixing circular import issues in `ckan.model` in a way that works for both Py3 and Py2.

### Current master (Works on py2, not on py3)

Import stack:
```python

# ckan/config/environment.py
import ckan.model as model

# ckan/model/__init__.py
from ckan.model.core import (
    System,
    State,
)

# ckan/model/core.py
import domain_object

# ckan/model/domain_object.py
import meta
import core
```

This fails in Python 3: 
```
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/model/__init__.py", line 29, in <module>
    from ckan.model.core import (
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/model/core.py", line 3, in <module>
    import domain_object
ModuleNotFoundError: No module named 'domain_object'
```

### First try (Works on py3, not on py2)

I've refactored the imports as follows to make them work on Python 3:
Import stack:
```python

# ckan/config/environment.py
import ckan.model as model

# ckan/model/__init__.py
from ckan.model.core import (
    System,
    State,
)

# ckan/model/core.py
from ckan.model import domain_object

# ckan/model/domain_object.py
from ckan.model import meta, core
```

But then it fails on Python 2:
```
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/model/__init__.py", line 29, in <module>
    from ckan.model.core import (
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/model/core.py", line 3, in <module>
    from ckan.model import domain_object
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/model/domain_object.py", line 9, in <module>
    from ckan.model import meta, core
ImportError: cannot import name core
```

Is there a syntax that works on both versions? Do we need to refactor? Conditional imports (eugh)?

amercader:
@ckan/core @cclauss any ideas on the above?

davidread:
I believe you get absolute imports (ie python3 behaviour) by sticking at the top: ```from __future__ import absolute_import``` so maybe works as a quick fix?

davidread:
I've added a commit which removes a couple of the circular imports, which seems good in itself and solves the issues python2/3 differences.

I can now import ckan.model in both python2 and python3, assuming I comment out the pylons import in common.py, as done here: https://github.com/ckan/ckan/pull/5063/files#diff-354a68e29e2bd42febb189421ec481fd (so it would be great to get that one merged...:)

PR #5080, opened 2019-11-17
Migrate User CLI

tino097:
Migration of `User` command to use Click


ckan/cli/user.py
```diff
@@ -0,0 +1,143 @@
...
+    user = model.User.get(text_type(username))
+    click.secho(u'User: %s' % user)
+
+
+@user.command(u'setpass', short_help=u'Set password for the user')
+@click.argument(u'username')
+def set_password(username):
+    import ckan.model as model
+    if not username:
+        click.secho(u'Need name of the user.', fg=u'yellow')
```
davidread: error_shout?

tino097: Thanks @davidread

tino097:
@amercader @davidread because @smotornyuk is on vacation, please review the PR and merge it if ok

PR #5084, opened 2019-11-18
Fix broken Solr docs link in package_search (Action API reference)

eetumans:
https://docs.ckan.org/en/latest/api/index.html#ckan.logic.action.get.package_search

Solr RefGuide was moved from cwiki.apache.org to lucene.apache.org but
the link in Action API reference does not reflect this change and ends
up with 'Page Not Found'

I happened to stumble upon a broken link taking to the old Solr RefGuide that doesn't exist anymore http://wiki.apache.org/solr/CommonQueryParameters

https://cwiki.apache.org/confluence/display/solr/ states the following:
>**Solr Reference Guide**
This Confluence space was earlier used for the Solr Reference Guide. If you are looking for the RefGuide, please visit https://lucene.apache.org/solr/guide/

Thus I replaced the broken link with:
https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html

The Service is provided by the Population Register Centre (https://vrk.fi/en/).

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1545,8 +1545,8 @@ def package_search(context, data_dict):
 
     **Solr Parameters:**
 
-    For more in depth treatment of each paramter, please read the `Solr
-    Documentation <http://wiki.apache.org/solr/CommonQueryParameters>`_.
+    For more in depth treatment of each parameter, please read the `Solr
+    Documentation <https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html>`_.
 
     This action accepts a *subset* of solr's search query parameters:
 


eetumans:
Noticed some more broken links, fixed those and renamed the PR accordingly. I'm not confident enough in Git to rename the branch as I don't quite know what kind of a can of worms would that open...

Missed them initially as I searched the repo with "cwiki.apache" instead of "wiki.apache"... D'oh!

amercader:
Thanks @eetumans that's really helpful.

@wardi the only issue we flagged on the call was that the version linked (6.6) requires some changes in CKAN to work. But that's the oldest Solr version with docs published so perhaps that is better than broken links?

pdelboca:
Thanks for this PR! I'm closing it since Solr docs has been re-written.

PR #5085, opened 2019-11-18
CLI. sysadmin

smotornyuk:


tino097:
@smotornyuk you can change the code to use the new cli command

amercader:
@tino097 I updated the docs on 6d68cd9a9

PR #5088, opened 2019-11-19
Add draft datasets to dataset page

Zharktas:
### Proposed fixes:
Currently datasets in draft state are only displayed in the user profile, this makes hard to find them if the user creating the dataset is somewhat clueless and sysadmin tries to fix the problem.

This PR adds draft datasets to dataset search and group and organization pages similarly how private datasets are handled.

### Features:

- [x] includes API changes

The Service is provided by the Population Register Centre (https://vrk.fi/en/).

### DIFF
--- a/ckan/controllers/group.py
+++ b/ckan/controllers/group.py
@@ -334,6 +334,7 @@ class GroupController(base.BaseController):
                 'q': q,
                 'fq': fq,
                 'include_private': True,
+                'include_drafts': True,
                 'facet.field': facets.keys(),
                 'rows': limit,
                 'sort': sort_by,
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -282,6 +282,8 @@ class PackageController(base.BaseController):
                 'extras': search_extras,
                 'include_private': asbool(config.get(
                     'ckan.search.default_include_private', True)),
+                'include_drafts': asbool(config.get(
+                    'ckan.search_default_include_drafts', True)),
             }
 
             query = get_action('package_search')(context, data_dict)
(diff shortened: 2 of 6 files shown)


amercader:
I think this is a good change overall and will help users find drafts, which right now you can only do by going to your user page or dashboard.

The only thing to think about is whether drafts should be visible only to the creator user or to the members of their organization. From what I understand now, all org members can see draft datasets right now, although the fact that they need to explicitly go to a particular user page to find them makes this really obvious. This will all change if this PR is merged, all org members would see a new draft on top of the search results as soon as it was created.

I personally think that what most users would assume is that drafts should only be visible to the original submitter (and sysadmins). I know this is not what this PR covers but seems like a good time to raise it.

Zharktas:
In our instance, most common case for some dataset being in draft state, is due to person forgetting the whole thing after he initially creates the dataset and often that person is no longer involved with open data at all anymore. In general, we try to offload us much as possible of the administration of organisation to some responsible person within the said organization, so at least in our case it would make sense that other members in the organization would see the draft dataset. 

Draft dataset itself is kind of odd to our users, as one can't directly control the state from the UI. The dataset moves from draft state to active only when whole two page process (creating dataset and adding at least one resource) is finished in the UI. 

Other options for this would be that draft datasets would be visible to submitter and Sysadmin, which could be handled via permission labels, or removing the whole draft state concept all together, since creating datasets via api, they are active by default.

Zharktas:
@amercader how do you want handle this? I can make the PR for removal of draft states in form wizard thing, but do you want it as separate PR and in that case what you want to this one. I feel that this could be useful at least some instances as draft datasets are looked for every now and then and the PR adds configuration option for displaying the drafts which default to true, but it can be false too.

amercader:
@Zharktas having revisited this, I think it's a good idea to show drafts as long as we ensure only the original creator and sysadmins can see them via permission labels

PR #5090, opened 2019-11-19
[docker] rebuild search index in entrypoint

olivierdalang:
When rebuilding the docker-compose stack, solr's index isn't rebuilt, leading to a scary "0 datasets" messages.

### Proposed fixes:

Add `paster --plugin ckan search-index rebuild -c /path/to/config.ini` to the ckan's entrypoint. This will also fix 

Added benefit (not sure ?) : this will also make changes in `ckan.site_url` be applied on restart.
Drawback : this will make restarts slower, esp. for big datasets.

### Features:

- [ ] includes API changes

### DIFF
--- a/contrib/docker/ckan-entrypoint.sh
+++ b/contrib/docker/ckan-entrypoint.sh
@@ -69,4 +69,5 @@ fi
 
 set_environment
 ckan-paster --plugin=ckan db init -c "${CKAN_CONFIG}/production.ini"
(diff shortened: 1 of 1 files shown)


amercader:
I'm not sure that's a good approach for all sites. As you say for any site with a significant amount of datasets this will take a long time. Perhaps a better approach is to use a volume to persist the Solr data folder (and thus the search index), as done here: https://github.com/okfn/docker-ckan/blob/master/docker-compose.yml#L48

PR #5091, opened 2019-11-19
parseCompletions handles v3 api result

howff:
v2 api returns a ResultSet: { Result: { Name: xxx } }
but the v3 api returns result: [ xxx ]
so map the latter to the former.

Fixes #

https://github.com/ckan/ckanext-scheming/issues/224

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/client.js
+++ b/ckan/public/base/javascript/client.js
@@ -161,7 +161,10 @@
 
       var map = {};
       // If given a 'result' array then convert it into a Result dict inside a Result dict.
-      data = data.result ? { 'ResultSet': { 'Result': data.result.map(x => ({'Name': x})) } } : data;
+      // new syntax (not used until all browsers support arrow notation):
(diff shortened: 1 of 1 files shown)


smotornyuk:
1. We are not using any processing for javascript, so es6 arrow functions not acceptable for now, while we are trying to support IE10
2. As autocompletion doesn't expose any details to other extensions, we can safely change the representation of results. So instead of converting modern results into an obsolete format, won't it be better to convert v2 results into something similar to v3 API?

howff:
Thanks for your feedback, I didn't realise it would fail on IE10.

I agree it would be better to move to v3 instead of mapping v3 back to v2 format, but my first commit was intended to change the minimum amount of code, 1) to minimise the chance of me breaking something, 2) in case there are other effects that I'm not aware of. Would you like to propose a different implementation?

smotornyuk:
Nope, as far as i know, everything else is perfectly ok, so i'll merge your PR if you'll replace arrow function with anonymous. Switching to v3 response format can be implemented in future, as separate PR

PR #5095, opened 2019-11-21
Don't use `webhelpers` in new codebase

smotornyuk:
Replace `webhelpers.html` with [`dominate`](https://github.com/Knio/dominate) library(i just like the way it handles HTML creation and it quite popular)
`webhelpers.paginate.Page` copied(with few changes) into `ckan.lib.pagination` module and re-exported from `helpers`.

I left webhelpers imports in controllers - as controllers also will be removed before enabling support for Python 3(am I right?) - there is no sense in this cleanup.

That's it, we don't use webhelpers anymore outside of controllers.


amercader:
Thanks @smotornyuk, looks good. I don't think we need to keep the `file` and `submit` helpers but we can drop them later.

I added a license header on the paginate module as we've copied most of the code: b4d7606f3

Issue #2838, opened 2016-01-17
Improvements to User Management


mbellotti:
There are a couple things that bother me about the way CKAN handles Users. Emails are not validated and multiple accounts can be registered under one email. This encourages spam attacks. It also makes logging in by email impossible.

On HDX we rewrote a lot of the user backend to fix this problem and also to add the ability to extend users with extras same way one might with groups or datasets. Would definitely be interested in taking that work and rolling it into CKAN core.

Thoughts?

rossjones:
I'd be keen to see this and user-extras in CKAN core, although might involve some care that we don't lock out users in existing installations.

rossjones:
Sorry for the delay getting back, entirely my fault.

This was discussed at the meeting and all thought it may be best to split these into two, solely to avoid potential problems with existing instances that don't have the single email restriction.  

We thought it would be great to get email validation and single-email restriction into CKAN as a core-extension, which is a normal plugin that is distributed with CKAN at https://github.com/ckan/ckan/tree/master/ckanext - that way it is simple to activate and won't cause problems for those that have existing duplicate emails, they can migrate and activate when ready. Particularly if it were possible to activate them independently.

The user-extras, all believed would be easier to implement, and much more appropriate as a core part of CKAN.  Presumably it'd also need an IUserForm interface in the same way that they exist for datasets. 

Does that seem like a good approach?  
If you're happy with that, is there anything you might need to help you get started with it?  

I'd also suggest it would make sense to start the PR early, so that it's easier for others to spot potential issues early and save you some time, hopefully, in the long run.  Marking it as WIP will mean that it doesn't get merged prematurely.

mbellotti:
How far a long is the IAuthenticator Interface now? It is currently still listed as "experimental" in the docs and this made it pretty difficult to design these changes as an extension. It ended up being easier to add a little code in the core to allow login via email: https://github.com/OCHA-DAP/hdx-ckan/blob/65fb4f74bda4eed21fc8070b1e30493902cdcf07/ckan/lib/authenticator.py#L22

rossjones:
I doubt there would be objections to login-by-email, as long as the validation of re-used email addresses and validating emails was in the extension (presumably in the logic layer and a controller for the validation url).  So putting login-by-email into core with user-extras should be fine.

mbellotti:
Okay. I got I good idea of how I want to start this. One housekeeping question though, should the branch be based off master of some specific release branch?

rossjones:
[...]

PR #5100, opened 2019-11-26
[#2838] Email unique validator

mutantsan:
At the moment, there is no validation for the uniqueness of the email, which does not allow us to use login via email. And it's also can be a vulnerability to spam attacks.

This changes implements the is_email_unique validator, which prevents from creation two or more users with the same email. Also, it's a step to provide authentication via email. Since this changes doesn't touch the models, there are shouldn't be conflicts with existing portals.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -859,6 +859,26 @@ def email_validator(value, context):
             raise Invalid(_('Email {email} is not a valid format').format(email=value))
     return value
 
+def email_is_unique(key, data, errors, context):
+    '''Validate email is unique'''
+    model = context['model']
+    session = context['session']
+
+    users = session.query(model.User) \
+            .filter(model.User.email == data[key]).all()
+    # is there is no users with this email it's free
+    if not users:
+        return
+    else:
+        # allow user to update their own email
+        for user in users:
+            if user.id == data[("id",)]:
+                return
+
+    raise Invalid(
+                _('The email address \'{email}\' \
+                    belongs to a registered user.').
+                        format(email=data[key]))
 
 def one_of(list_of_value):
     ''' Checks if the provided value is present in a list '''
--- a/ckan/cli/db.py
+++ b/ckan/cli/db.py
@@ -4,9 +4,10 @@ import os
 import logging
 import ckan.migration as migration_repo
 import click
+from itertools import groupby
 
 from ckan.cli import error_shout
-
+import ckan.model as model
 log = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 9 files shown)


sivang:
@wardi @amercader would it make sense to have a script actually emailing the dups asking to consolidate the accounts to one account?

amercader:
@mutantsan yes, as @sivang points out having unique emails is definitely a desired behaviour. But as uniqueness has not been enforced until now, there might be an issue with existing duplicates. You will get an exception when trying to edit one of the users with the same email.

We need to provide guidance on how to migrate. Touching the actual emails is probably too risky so clear warnings in the changelog plus some helpers like the script @sivang suggests is probably safer, and we leave up to maintainers to decide how to fix the duplicates.

The script could be a new command `ckan db check-duplicate-emails` that returns the user details of those accounts with duplicated emails.
@mutantsan want to add this as part of this PR?

mutantsan:
> @mutantsan want to add this as part of this PR?
Yes, I will work on it.

mutantsan:
@wardi @amercader  I'm just thinking about what the command should do. I don't think that emailing to users is a great idea. Can we just update all duplicate emails to empty and let users change it, since they can just update their own profiles and even delete the email?

amercader:
@mutantsan sorry, I wasn't clear. The command should not do anything to the database or send any email. It should only show a list of users with duplicated emails so the site maintainers can decide what to with them

davidread:
This probably shouldn't be a top level cli command. Can it be moved to be under `db`, as @amercader suggested? And maybe calling it `duplicate_emails` or something is clearer than `email_check`.

mutantsan:
Sorry, this PR history looks like a mess. I've add _duplicate_emails_ subcommand under db.

davidread:
Many thanks @mutantsan! All LGTM, but I'll leave it to the assignees to confirm/merge.

ckan/logic/validators.py
```diff
@@ -861,6 +861,18 @@ def email_validator(value, context):
             raise Invalid(_('Email {email} is not a valid format').format(email=value))
     return value
 
+def email_is_unique(key, data, errors, context):
+    '''Validate email is unique'''
+    model = context['model']
+    session = context['session']
+    user = session.query(model.User).filter_by(email=data[key]).first()
```
davidread: Oh, if the existing user shows up first, then it won't detect any duplicates. I think you need to filter out the user in this query i.e. `.filter(model.User.name != data[('name',)])` rather than in the `if` in the next statement.

mutantsan: @davidread , thank you! I've changed the validator logic. But, since the validator used in two cases (registration and update), I've chose another approach.

ckan/logic/validators.py
```diff
@@ -861,6 +861,26 @@ def email_validator(value, context):
...
+    model = context['model']
+    session = context['session']
+
+    users = session.query(model.User) \
+            .filter(model.User.email == data[key]).all()
+    # is there is no users with this email it's free
+    if not users:
+        return
+    else:
+        # allow user to update his own email
```
davidread:
```suggestion
        # allow user to update their own email
```

amercader:
@mutantsan This all looks great. I was going to merge but after merging master I'm getting two failures in the tests involved in this PR. Would you mind having a look? https://circleci.com/gh/ckan/ckan/6311#tests/containers/1

mutantsan:
@amercader Hi, I've fixed the tests. `test_request_reset_when_duplicate_emails` was deleted, cause I think it's makes no sense to test reset with duplicate emails when we want to get away from it.

amercader:
Thanks a lot @mutantsan

PR #5096, opened 2019-11-21
PY3 general support - Part 1

amercader:
All commits are self-contained and should be self-explanatory.

PR #5107, opened 2019-11-28
PY3 General Support - Part 2

amercader:
#5096 Should be reviewed and merged first

WIP


davidread:
Looks good so far

amercader:
@ckan/core flagging this batch of changes as ready for review.

I've reached the milestone of serving the homepage without exceptions, but the static files are not loading.
@smotornyuk I'm going to need your help here as I'm not sure how the webassets ties into this or how to initialize it so the Flask app is the one serving the static files (right now it appears to be the Pylons one)

smotornyuk:
Looks cool. I'll take a look on static files

amercader:
@smotornyuk amazing thanks. If you try this out in py3 you'll need to comment this line https://github.com/ckan/ckan/blob/master/ckan/cli/sysadmin.py#L8 until https://github.com/ckan/ckan/pull/5114 is finished.

smotornyuk:
@amercader , here's a patch for serving static files from Flask. It requires additional library - `flask-multistatic`. Flask supports only a single static directory per blueprint/app, whereas we should provide access to multiple folders registered by extensions. The mentioned library extends the original `Flask` class, providing support for the list of static folders. It's pretty small, so we can implement the required functionality inside our `CKANFlask` if increasing number of dependencies is not desired. 

[static.txt](https://github.com/ckan/ckan/files/3972656/static.txt)

smotornyuk:
Looks good. I'm going to merge this PR if you aren't going to add more commits today.

amercader:
Go ahead @smotornyuk, I'll start Part 3 with the static files changes.

Issue #5108, opened 2019-11-29
Allow extensions to define Click commands


amercader:
The core ones are migrated but currently there is no way for extensions to add commands to the main `ckan` one (eg `ckan harvester`, `ckan saml2 initdb` etc).

This is the place where it should happen: [ckan/cli/cli.py](https://github.com/ckan/ckan/blob/master/ckan/cli/cli.py). In the same way that the core ones are added with the `ckan.add_command()` call, we should iterate all extensions and register any commands provided by them.

Extension paster commands use entrypoints defined in the extensions `setup.py` file (`[paste.paster_command]`). Perhaps for the `ckan` one we could follow the existing pattern of providing an interface (eg `ICLICommand`) with a `get_commands()` method that plugins can implement. This will return a reference to the click command defined in the extension (like the different modules on `ckan/cli/`).

Then on `ckan/cli/cli.py` we iterate over all plugins and load the provided commands (check how we get [extensions blueprints](https://github.com/ckan/ckan/blob/master/ckan/config/middleware/flask_app.py#L188) for an example).

If there are methods from the core cli module that would be helpful to extensions we can add them to the toolkit (eg `error_shout`)

mutantsan:
Hi, I've tried to implement this feature in this PR #5112 . It works, but probably needs some improvements.

PR #5112, opened 2019-12-03
[#5108] Allow extensions to define Click commands

mutantsan:
Currently, there is no way for extension to add commands
to the main ckan one

This changes implements the possibility for extensions
to do it. It's implements the new ICLICommands interface 
which extension can use to register click commands.

### Features:

- [ ] includes API changes


ckan/cli/cli.py
```diff
@@ -31,19 +32,33 @@
 log = logging.getLogger(__name__)
 
 
+class IClick(click.Group):
```
smotornyuk: I think, everyone who looks over the code will assume that `IClick` is some kind of interface(i.e. something that acts like contract and requires future implementation), whereas in your case it's Group that initializes the application and makes re-attempt to invoke the missing command

ckan/cli/cli.py
```diff
@@ -31,19 +32,33 @@
...
+    def get_command(self, ctx, name):
+        cmd = self.commands.get(name)
```
smotornyuk: `super(...).get_command` would be more idiomatic here, right?

ckan/cli/cli.py
```diff
@@ -31,19 +32,33 @@
...
+        if not cmd:
+            ckan.invoke(ctx)
```
smotornyuk: and here `self.invoke` is better, as it won't restrict us in future, when new entry points may be added to CLI interface

ckan/plugins/interfaces.py
```diff
@@ -39,6 +39,7 @@
     u'IBlueprint',
     u'IPermissionLabels',
     u'IForkObserver',
+    u'ICLICommands',
```
smotornyuk: It's my personal opinion - such sequences of capitalized letters are just ugly. There are some styles that suggest using title-case for capitalized sequences that consist of more than 3 letters, i/e `ICliCommand`. What do you think?

ckan/plugins/interfaces.py
```diff
@@ -1779,3 +1780,14 @@ def before_fork(self):
         u'''
         Called shortly before the CKAN process is forked.
         '''
+
+
+class ICLICommands(Interface):
+    u'''
+    Allow extensions to define click commands.
+    '''
+    def get_commands(self):
```
smotornyuk: as it called `get_commands` and just in case - maybe it would be better to allow returning list of commands instead of single one from this method? We have a similar experience with `IBlueprints` - it appears that people often want to create multiple entry points inside the extension.

amercader:
This PR should cover as well whether new methods or classes are needed in the toolkit (eg `load_config`?)

ckan/plugins/interfaces.py
wardi: I can't find the reference but I think ICLICommand is correct by python naming standards in the same way as HTMLParser is preferred over HtmlParser

mutantsan: What about ICommands or IClick then?

ckan/plugins/interfaces.py
mutantsan: Yes, it's definitely better to work with multiple commands at once.

ckan/plugins/interfaces.py
smotornyuk: :+1:

smotornyuk:
[You are checking IClick interface](https://github.com/ckan/ckan/pull/5112/files#diff-0e409a523668d377588e719582c66b22R60) while inside `interfaces` exists only [ICLICommands](https://github.com/ckan/ckan/pull/5112/files#diff-4a6382e3872769be8829f4e5968cb848R1785)

smotornyuk:
Cool :+1: 
I'm merging this PR, as it is pretty critical for the extensions' migration process. But it would be nice to have tests and examples of usage for this feature.. 
Anyway, I'm going to add those pieces while working on migration guide for extensions, but any help is highly appreciated :wink:

PR #5114, opened 2019-12-03
Use User CLI method 'user_add' instead of the old cli user method

tino097:
Fixes #5085

When we merged the sysadmin CLI command, it was still using the old `user_add` instead of the new `user.add_user` method

### DIFF
--- a/ckan/cli/sysadmin.py
+++ b/ckan/cli/sysadmin.py
@@ -5,7 +5,7 @@ from six import text_type
 
 import ckan.model as model
 from ckan.cli import error_shout
-from ckan.lib.cli import user_add
+from ckan.cli.user import add_user
 
 
 @click.group(
@@ -51,8 +51,7 @@ def add(username, args):
         if click.confirm(
             u"Create new user: %s?" % username, default=True, abort=True
         ):
-            # TODO: once, cli.user is merged, invoke `user.add_user` instead
-            user_add([username] + list(args))
+            add_user([username] + list(args))
             user = model.User.by_name(text_type(username))
 
     user.sysadmin = True


amercader:
Thanks @tino097 , I bumped into this just yesterday. But I think this needs a bit more work, because I'm getting:

```
  File "/home/adria/dev/pyenvs/ckan/local/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/cli/user.py", line 72, in add_user
    flask_app = ctx.obj.app.apps[u'flask_app']._wsgi_app
AttributeError: 'NoneType' object has no attribute 'app'
```

The new `user.add_user` function expects `ctx` as first argument.
To be honest I don't think we need the [` flask_app.test_request_context()`](https://github.com/ckan/ckan/blob/master/ckan/cli/user.py#L73) in there unless we are creating urls or doing something that requires a request context.
If we are, then we should do something like we did in the old CLI, where we were creating a default context on startup (like [here](https://github.com/ckan/ckan/blob/master/ckan/lib/cli.py#L245)). But first let's check if we really need those contexts in the user command.

smotornyuk:
I've tested it on upstream, and it seems, that issue is gone and `user_create` works without context:
```diff
diff --git a/ckan/cli/user.py b/ckan/cli/user.py
index d5a21d448..8be4b30b3 100644
--- a/ckan/cli/user.py
+++ b/ckan/cli/user.py
@@ -23,8 +23,7 @@ def user():
 @user.command(u'add', short_help=u'Add new user')
 @click.argument(u'username')
 @click.argument(u'args', nargs=-1)
-@click.pass_context
-def add_user(ctx, username, args):
+def add_user(username, args):
     u'''Add new user if we use ckan sysadmin add
     or ckan user add
     '''
@@ -69,10 +68,8 @@ def add_user(ctx, username, args):
             u'ignore_auth': True,
             u'user': site_user['name'],
         }
-        flask_app = ctx.obj.app.apps[u'flask_app']._wsgi_app
-        with flask_app.test_request_context():
-            user_dict = logic.get_action(u'user_create')(context, data_dict)
-            click.secho(u"Successfully created user: %s" % user_dict['name'],
+        user_dict = logic.get_action(u'user_create')(context, data_dict)
+        click.secho(u"Successfully created user: %s" % user_dict['name'],
                         fg=u'green', bold=True)
     except logic.ValidationError as e:
         error_shout(e)
```
I recall, that i'm also had some issues with app context previously, during `user_create`, so this issue probably was fixed by one of Adria's py3 PRs

tino097:
Closing in favor of #5129

Issue #4803, opened 2019-05-23
Replace paster make-config


amercader:
From @davidread input, plus some extra detail:

This is the command used to create new ini files (eg `development.ini` or `production.ini`). This is done as part of [installation](https://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html#create-a-ckan-config-file):

    paster make-config ckan /etc/ckan/default/development.ini

It generates it from a template: https://github.com/ckan/ckan/blob/master/ckan/config/deployment.ini_tmpl

I think it also has a feature that allows you to update your .ini file when you upgrade CKAN and the template has changed. However I'm not sure anyone uses that - we make sure additions are defaulted so that .ini files are forwards compatible.

#### Approach

Add a new CLI function that renders the template and saves it as a file. Now that we have the `ckan generate` command we can add a `ckan generate config` option (in addition to the existing `ckan generate extension`). This should accept one argument, the path to the new ini file.

Rendering in this case just means substituting variables like `${app_instance_secret}` which we might change to Jinja style and use the Jinja renderer. Or even python `.format()` - I don't think escaping is needed.

Variables like `%(ckan.site_id)s`  should be kept in the resulting ini file, as these are evaluated when reading the ini file, not when writing it.

#### Story points

3

davidread:
# Requirements

i.e. a way to create development.ini. This is done as part of [installation](https://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html#create-a-ckan-config-file):

    paster make-config ckan /etc/ckan/default/development.ini

It generates it from a template: https://github.com/ckan/ckan/blob/master/ckan/config/deployment.ini_tmpl

I think it also has a feature that allows you to update your .ini file when you upgrade CKAN and the template has changed. However I'm not sure anyone uses that - we make sure additions are defaulted so that .ini files are forwards compatible.

# How

A cli function that renders the template and saves it as a file. Rendering in this case just means substituting variables like `%(ckan.site_id)s` and `${app_instance_secret}` which we might change to Jinja style and use the Jinja renderer. Or even python `.format()` - I don't think escaping is needed.

mpolidori:
I'd like to take this one

amercader:
Great @mpolidori, thanks a lot! Ping us here if you need help. If you see you won't be able to work on it after all let us know so I we can assign it to someone else.

mpolidori:
@amercader Will do. Thanks!

mpolidori:
[...]

PR #5124, opened 2019-12-16
[WIP][fix][s]: replaces paster make-config - refs #4803

mpolidori:
Refs #4803 

It looks as though only 2 variables are replaced by `paster make-config`, `beaker.session.secret` and `app_instance_uuid`. It's a pretty straightforward replacement. I'm still working on avoiding the requirement to already have an existing `development.ini`. With my limited knowledge, I haven't discovered a way around it yet, but I'm still looking into it.


mpolidori:
@amercader I found a simple way to get around the CLI requiring an existing config file. Let me know if everything looks good, or if I need to change anything. I also check to make sure the output file is named `development.ini` or `production.ini`. Not sure if this is required or not.

ckan/cli/cli.py
```diff
@@ -72,6 +73,10 @@ def __init__(self, conf=None):
 @click_config_option
 @click.pass_context
 def ckan(ctx, config, *args, **kwargs):
+    # This is necessary to allow the user to create
+    # a config file when one isn't already present
+    if all(arg in sys.argv for arg in ['generate', 'config']):
```
smotornyuk: I think this idea is really nice, but there are two concerns from my side:

* Probably, the whole `generate` tree doesn't require config file, so i would only check if first sub-command is `generate`
* Also, there is a small possibility that some `generate` commands will either require config file, or behave differently when it provided. So i would add `config is None` check - if there is no config file AND it's a `generate` command - let's continue. But if there IS config file - let's use it- there is no harm in it

mpolidori: @smotornyuk Sorry for the delay. I made the changes. Let me know what you think. Thanks!

kowh-ai:
As CKAN 2.9+  is standardising on the configuration file being named 'ckan.ini' is it possible to reflect this requirement in this PR?

mpolidori:
@kowh-ai Ah, I didn't know that. I must've missed it. No problem.

kowh-ai:
@mpolidori hey no problem...I think the feeling from the Dev Core team is to consolidate to only one configuration file (ckan.ini) for simplicity sake. This is from 2.9 onwards

mpolidori:
@kowh-ai That makes sense. Unfortunately, some recent changes on master broke my workaround for the pre-existing config file. I'm working on it now.

mpolidori:
@kowh-ai Should be good to go now

kowh-ai:
Did you want me to bring up this PR at the Dev Meeting today Michael?

mpolidori:
@kowh-ai Between timezone differences and work this morning, I missed this comment. The Thursday meetings are the same time as my work Team meeting, but I'd love to sit in on some of the Tuesday meetings if I have time. You can definitely bring it up at the next one, if you want.

Does everything look good, or do you have any ideas on anything I can change? I thought about possibly implementing a check in the output directory, just in case a config already exists there, and the user doesn't want to accidentally overwrite it. This probably isn't a common case though.

kowh-ai:
@mpolidori - I actually brought this up at the Dev Meeting yesterday and Adria said he would take a look at the PR today...he's busy but hopefully he can get to this today!

mpolidori:
@kowh-ai Great! No rush on my end

ckan/cli/generate.py
```diff
@@ -77,3 +80,44 @@ def extension(output_dir):
 
     cookiecutter(template_loc, no_input=True, extra_context=context,
                  output_dir=output_dir)
+
+
+@generate.command(name=u'config', short_help=u'Create config from template.')
+@click.option(u'-o', u'--output-dir', help=u'Location to put the generated '
```
amercader: I know that the existing `extension` command uses the same pattern, but in my opinion options should be used to pass flags that modify the default behavior. For a parameter that is always required I would use arguments (ie  `@click.argument`). This feels more natural to me than the `-o` flag:

```
ckan generate config /etc/ckan/ckan.ini
```

This will also remove the need for checking if the parameter is provided as I assume click will take care of this.
Also can you change the helper text to "Path to the generated configuration ini file"

ckan/cli/generate.py
```diff
@@ -77,3 +80,44 @@ def extension(output_dir):
 
     cookiecutter(template_loc, no_input=True, extra_context=context,
                  output_dir=output_dir)
+
+
+@generate.command(name=u'config', short_help=u'Create config from template.')
```
amercader: Change helper text to "Generate a new CKAN configuration ini file"

ckan/cli/generate.py
```diff
@@ -77,3 +80,44 @@ def extension(output_dir):
...
+@click.option(u'-o', u'--output-dir', help=u'Location to put the generated '
+                                           u'template.')
+def make_config(output_dir):
+    if output_dir is None:
+        print('\nERROR: Try again with \'-o path/to/ckan.ini\'')
+        sys.exit(1)
+
+    name = 'ckan.ini'
+
+    if name not in output_dir:
```
amercader: We will default to `ckan.ini` on our installation instructions and deployment scripts but this name should not be enforced.
People might want to use the old `development.ini` / `production.ini` to support existing deployment pipelines, or use different ini files for different environments.

ckan/cli/generate.py
```diff
@@ -77,3 +80,44 @@ def extension(output_dir):
...
+        sys.exit(1)
+
+    # Output to current directory if no path is specified
+    if output_dir == name:
+        output_dir = os.getcwd() + '/' + output_dir
+
+    cur_loc = os.path.dirname(os.path.abspath(__file__))
+    os.chdir(cur_loc)
+    os.chdir('../config')
+    template_loc = os.getcwd() + '/deployment.ini_tmpl'
```
amercader: This can be accessed more concisely with 
```python
os.path.join(cur_loc, '..', 'config', 'deployment.ini_tmpl')
```

ckan/cli/generate.py
```diff
@@ -77,3 +80,44 @@ def extension(output_dir):
...
+    template_variables = {
+        'app_instance_uuid': uuid.uuid4(),
+        'app_instance_secret': ''.join(
+            random.SystemRandom().choice(string.ascii_letters + string.digits)
```
amercader: The original [paste implementation](https://github.com/cdent/pastescript/blob/e2741855d6250c8f6f942601b0e66e76f3dba564/paste/script/util/secret.py#L24) of this secret encodes it as base64. To be on the safe side in terms of compatibility let's do it as well before passing it to the template.

smotornyuk: How about using `secrets` library(build-in for py3 and backported as external `python2-secrets`) - it has `token_urlsafe` function that generates base64 encoded token in more secure way than systemRandom does and it's kind of de-facto standard when you are working with all kinds of tokens in py3

amercader: Sounds fantastic

mpolidori:
@amercader @smotornyuk Sounds good! I'll let you know if I have any questions.

ckan/cli/generate.py
mpolidori: That is much cleaner. I probably should have asked about that, as I was following the convention in the file, but it seemed that I didn't really need an `option`.

ckan/cli/generate.py
mpolidori: That's a good point. No problem!

ckan/cli/generate.py
mpolidori: That's much better. I always forget about `path.join`...

ckan/cli/generate.py
mpolidori: Great! I switched to `token_urlsafe`. 

I didn't think to ask before I pushed the changes, but does `beaker.session.secret` only take alphanumeric characters? In the paste implementation code, it looks like they're only avoiding characters in `'\n\r='`. From the few examples of `beaker.session.secret`s that I have seen, they seem to be alphanumeric though, not including `-` or `_` (which `token_urlsafe` can output).

mpolidori: @amercader @smotornyuk I pushed the new changes from the review. Let me know what you think.

requirements.txt
```diff
@@ -36,6 +36,7 @@ pysolr==3.6.0
 python-dateutil==2.8.1
 python-editor==1.0.4      # via alembic
 python-magic==0.4.15
+python2-secrets==1.0.5    # This can be removed after the Python 2->3 migration
```
smotornyuk: Put this line into `requirements-py2.txt` instead - it is used for such "Python 2" only requirements, that will no longer be used after migration finished

mpolidori: Yeah... I should have known better. Thanks for pointing it out!

davidread:
Great to see this one getting there!

I suggest that in this PR we also change the instructions. So in install-from-source.rst it changes from:
    paster make-config ckan |development.ini|
to:
    ckan generate config |development.ini|

Regarding switching from development.ini to ckan.ini, there are a bunch of references and it's a variable, so I think it would be better to decouple that from this PR.

ckan/cli/generate.py
```diff
@@ -77,3 +80,32 @@ def extension(output_dir):
 
     cookiecutter(template_loc, no_input=True, extra_context=context,
                  output_dir=output_dir)
+
+
+@generate.command(name=u'config',
+                  short_help=u'Path to the generated configuration ini file.')
```
davidread: This is the help to the config command, not its param:
```
$ ckan generate --help
Usage: ckan generate [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  config     Path to the generated configuration ini file.
  extension  Create empty extension.
```
How about
```suggestion
                  short_help=u'Create a ckan.ini file.')
```

mpolidori: @davidread Ah, I missed that. Okay, will do. Thanks!

mpolidori: @davidread Okay. I took care of those in my latest commit.

ckan/cli/generate.py
davidread: I think you're happy and I'll just agree that what you have done now is great. Read on for my workings, only in case that's of interest.

It looks like the old method included `[a-z][A-Z][0-9]/+`:
```
$ for i in {1..20}; do paster make-config ckan /tmp/$i.ini; done
$ for i in {1..20}; do grep 'beaker.session.secret' /tmp/$i.ini; done
beaker.session.secret = wwHtgWpAdXp/aGh2jCu09284V
beaker.session.secret = 01kwxI9G1umxh3qJx8ae+ksRz
beaker.session.secret = yMj5cEcVsmLgQhsdp71qEwFyG
beaker.session.secret = gFvxgpVOWvKMWCRAaYDnAfnXt
beaker.session.secret = C5AjGeTEGLnJ2BykE6XmjpcaS
beaker.session.secret = CPsdQg8eXyMGSRVaN2cbB+QUd
beaker.session.secret = 54+5SOpiSghC9fgexjHt4dMNw
beaker.session.secret = 8N/7y/i6j08otGKEpoUIvnCzo
beaker.session.secret = jhnZ9Flpxn9MdiCbO5tD1QppY
beaker.session.secret = WVJwrZcg6JU2AmUCy2xz8jdAL
beaker.session.secret = 8zcVTl/klx1Q8qRESe8Cll1T/
beaker.session.secret = zUXW1EsAbdRbvn7nmakwN+h/5
beaker.session.secret = +BjEzx63gcvSHpQFZMi8tF3mK
beaker.session.secret = CUHPuLDv4tqBb+apYMLQ2wWuQ
beaker.session.secret = eRBPUGNK+5iTG+IlS0z8mqY7u
beaker.session.secret = arYV8bSU3E8puzPCk344UJEaG
beaker.session.secret = mmK40LPvnngWxp1yv2zqQrowV
beaker.session.secret = orUEjHYRY4GqPnUwLi4upDF+9
beaker.session.secret = x5PDimCHWbewwSwsbKURFESSH
beaker.session.secret = 3RXkEjuQslPIi/M3OaO9w/gQ+
```
I really don't think it matters about what symbols are included. As long as its ascii and has no newlines it can be part of the config file. Beaker just wants any ["randomly generated string"](https://beaker.readthedocs.io/en/latest/configuration.html#session-options).

mpolidori: Thanks for the information! It's good to know.

Ah, so it looks like `token_urlsafe` will work, though I can easily go with the same or similar implementation paste has. I'll leave it for now, unless anyone thinks there's a reason to have the extra characters.

davidread:
The python2 test failure should be fixed if you cherry-pick these code style changes onto your branch: https://github.com/ckan/ckan/commit/37446c5c0bd2bda8e91f7151ac5df8d61518879b

The python3 test failure should be fixed by: https://github.com/ckan/ckan/pull/5181 so once that's merged, do rebase or merge from master.

Other than that I think this PR gets the 👍 from me and has covered all the comments by others.

mpolidori:
@davidread All checks are passing now, but master isn't up to date in the files changed tab (changes to `test_mailer.py` are showing there). I searched a bit, and read that you can temporarily change the branch you're merging into to fix/update it, but it didn't work. It sounds like it is a known issue, but it's also possible that I did something wrong (still have some learning to do in regards to `git`).

davidread:
Ah great, all looks good now. It's ready for final review from @amercader now.

> master isn't up to date in the files changed tab

So I guess the "Files changed" is the diff compared to the point where you branched off master (rather than latest master). That's acceptable because there are no conflicts. But yes it would be nicer if the diff didn't show the test_mailer changes. You can still merge from master to your branch, or possibly rebase your branch onto latest master, to tidy that up.

kowh-ai:
@davidread With regard to your comment on the documentation changes required ( ie: removal of paster commands), this is part of:  https://github.com/ckan/ckan/pull/5176 ...thanks!

amercader:
This looks good to go @mpolidori , only one teeny change needed, the `python2-secrets==1.0.5` line should be also added to `requirements-py2.in` as these are our base requirements (`requirements(-py2).txt` files are generated with `pip-tools`)

davidread:
I added the line in requirements-py2.in to simply get it merged... hope that's ok.

Great work @mpolidori

mpolidori:
@davidread I don't mind at all! Sorry for the delay. I was tied up with something else, and it slipped my mind. Thanks for taking care of it.

amercader:
Thanks for your work on this @mpolidori

PR #5127, opened 2019-12-18
4486 add organization and group purge

mutantsan:
Currently, there is no possibility to purge organizations and groups via GUI
Therefore, sysadmins cannot delete these objects permanently through the GUI,
which prevents the creation of objects with the same names.

This changes add possibility to purge any type of entities such as datasets,
organizations and groups via GUI.

### Features:

- [X] includes API changes


mutantsan:
I'm not entirely sure why the `test_mail_recipient_with_html` failed once. Seems like it's not related to this PR.

mutantsan:
@Zharktas Hi! Could you check it pls? I had to fix the `test_mail_recipient_with_html` to pass the tests. I wonder how it's works in other PR's.

Zharktas:
@mutantsan mind merging master into it ? Circle CI is kind of dumb by testing the branch itself and not what the resulting commit would like after merge, there might be an issue which has been since fixed in master since this branch was originally created a month ago.

mutantsan:
> @mutantsan mind merging master into it ? Circle CI is kind of dumb by testing the branch itself and not what the resulting commit would like after merge, there might be an issue which has been since fixed in master since this branch was originally created a month ago.

Maybe it's just me, but I see this code in master too. https://github.com/ckan/ckan/blob/master/ckan/tests/lib/test_mailer.py
```
assert test_email["headers"].keys()[0] in msg[3], msg[3]
assert test_email["headers"].values()[0] in msg[3], msg[3]
```
This asserts should always fail on py3 tests, according to the documentation:

> The objects returned by dict.keys(), dict.values() and dict.items() are view objects.

Zharktas:
https://circleci.com/gh/ckan/ckan/tree/master maste has been failing for the past three days, so it really is unrelated to this. I'll review this at the weekend at latest.

ckan/templates/admin/snippets/confirm_delete.html
```diff
@@ -0,0 +1,27 @@
...
+
+{% block main_content %}
+  <section class="module col-md-6 col-md-offset-3">
+    <div class="module-content">
+      {% block form %}
+      <p>
+        {% if ent_type == 'all' %}
+          {{ _('Are you sure you want to purge everything?') }}
+        {% else %}
+          {{ _('Are you sure you want to purge {}s?').format(ent_type) }}
```
smotornyuk: 1. `{}` placeholder doesn't make any hint about it's possible values for project translators.
2. `ent_type` pass directly into template, thus it's untranslatable. i.e., you'll get:
in russian: "Вы хотите удалить package?"
in japanese: "packageを削除しますか？"
See the problem?

ckan/views/admin.py
```diff
@@ -44,11 +44,13 @@ def _get_config_options():
     }]
 
     homepages = [{
-        u'value': u'1',
+        u'value':
+        u'1',
```
smotornyuk: I don't mind, but is this change mandatory?

ckan/views/admin.py
```diff
@@ -143,26 +144,63 @@ class TrashView(MethodView):
...
+            u'package': self.deleted_packages,
+            u'organization': self.deleted_orgs,
+            u'group': self.deleted_groups
+        }
 
     def get(self):
-        data = dict(deleted_packages=self.deleted_packages)
-        return base.render(u'admin/trash.html', extra_vars=data)
+        if request.params.get(u'name'):
+            ent_type = request.params.get(u'name').split(u'-')[-1]
```
smotornyuk: A bit of mystery here. Won't it be better if you just pass `name` as is instead of splitting it? And if you are going to change this line, try not to call `dict::get` twice.

ckan/views/admin.py
```diff
@@ -143,26 +144,63 @@ class TrashView(MethodView):
     def __init__(self):
         self.deleted_packages = model.Session.query(
             model.Package).filter_by(state=model.State.DELETED)
+        self.deleted_orgs = model.Session.query(model.Group).filter_by(
+            state=model.State.DELETED, is_organization=True)
+        self.deleted_groups = model.Session.query(model.Group).filter_by(
+            state=model.State.DELETED, is_organization=False)
+
+        self.data_type = {
```
smotornyuk: You mean, `data_type` stores not the actual type of data, but rather a collection of deleted items?

ckan/views/admin.py
```diff
@@ -143,26 +144,63 @@ class TrashView(MethodView):
...
+
+        vars = dict(data=self.data_type)
+        return base.render(u'admin/trash.html', extra_vars=vars)
 
     def post(self):
-        if (u'purge-packages' in request.form):
-            for pkg in self.deleted_packages:
-                logic.get_action(u'dataset_purge')(
-                    {u'user': g.user}, {u'id': pkg.id})
+        req_action = request.form.get(u'name', request.params.get(u'name'))
```
smotornyuk: Personally, I think that it very idiomatic to pass data inside request body while making POST requests, so `request.form.get` perfectly suits here. But are you actually put anything inside `request.form`? It seems, that all forms encode `name` in request url..

ckan/views/admin.py
```diff
@@ -143,26 +144,63 @@ class TrashView(MethodView):
...
+
+        if u'cancel' in request.form:
+            return h.redirect_to(u'admin.trash')
+
+        if req_action and (u'purge-all' in request.form.get(u'name',
```
smotornyuk: it's not clear enough - you are checking `req_action`(which represents `name` field from payload) and after that, you are picking `name` field from payload once again? Is it some kind of typo/debug code or i missed something?

ckan/views/admin.py
```diff
@@ -143,26 +144,63 @@ class TrashView(MethodView):
...
+                           request.params.get(u'name'))):
+            ent_types = (self.deleted_packages, self.deleted_groups,
+                         self.deleted_orgs)
+            func_names = (u'dataset_purge', u'group_purge',
+                          u'organization_purge')
+            for ent_type, func_name in zip(ent_types, func_names):
```
smotornyuk: `dict::items` instead of zipping two tuples?

mutantsan:
Huge thanks @smotornyuk !

ckan/public/base/javascript/modules/confirm-action.js
```diff
@@ -83,6 +94,12 @@ this.ckan.module('confirm-action', function (jQuery) {
         action: this.el.attr('href'),
         method: 'POST'
       });
+
+      // use parent to el form if data-module-with-data == true
+      if (this.options.withData) {
+        var form = $(this.el).closest('form');
```
smotornyuk: `this.el` is an instance of jQuery collection by default.

ckan/public/base/javascript/modules/confirm-action.js
```diff
@@ -13,6 +13,17 @@ this.ckan.module('confirm-action', function (jQuery) {
        */
       content: '',
 
+      /* By default confirm-action creates a new form and submit it
+       * But you can use closest to el form by setting data-module-with-data=true
+       *
+       *     <a href="..."
+       *        data-module="confirm-action"
+       *        data-module-with-data=true>
```
smotornyuk: It's HTML markup and setting attribute to literal `true` won't pass strict checks. Use double quotes

ckan/public/base/javascript/modules/confirm-action.js
```diff
@@ -13,6 +13,17 @@ this.ckan.module('confirm-action', function (jQuery) {
...
+       *     {{ _('Save') }}
+       *     </a>
+       */
+      withData: '',
```
smotornyuk: if you are using default values, prefer `false` because this value is treated as flag

ckan/templates/admin/snippets/confirm_delete.html
```diff
@@ -0,0 +1,28 @@
...
+
+{% block main_content %}
+  <section class="module col-md-6 col-md-offset-3">
+    <div class="module-content">
+      {% block form %}
+      <p>
+        {% if ent_type == 'all' %}
+          {{ _('Are you sure you want to purge everything?') }}
+        {% else %}
+          {% set message = 'Are you sure you want to purge {ent_type}s?'.format(ent_type=ent_type) %}
```
smotornyuk: Don't rely on basic grammar rules and use `ungettext` for proper handling of plural form. Take a look at `search_result_text.html` - it's pretty similar to current template in terms of wording.

ckan/tests/controllers/test_user.py
```diff
@@ -769,22 +769,6 @@ def test_request_reset_by_name(self, send_reset_link, app):
...
-        response = app.post(
-            offset, data=dict(user="me@example.com")
-        )
-
-        assert "A reset link has been emailed to you" in response
-        emailed_users = [
-            call[0][0].name for call in send_reset_link.call_args_list
-        ]
-        assert sorted(emailed_users) == sorted([user_a["name"], user_b["name"]])
-
```
smotornyuk: What's wrong with this test?

ckan/templates/admin/snippets/confirm_delete.html
mutantsan: I'm not entirely sure how to use ungettext here. The entity type in `message`  is always plural, but `ent_type` variable is in singular form for a logic purpose.

smotornyuk: If there is no difference between purging single entity and multiple entities, just use static messages without any substitution, like on lines 18, 24 and 30 - https://github.com/ckan/ckan/blob/master/ckan/templates/snippets/search_result_text.html

amercader:
@mutantsan is this ready for final review? Would love to get this into 2.9 (cc @smotornyuk )

mutantsan:
Hi, @amercader. Yes it is.  If there are anything else, I am about to fix it asap.

smotornyuk:
In that case, could you check comments from the last review and fix conflicts?

ckan/tests/controllers/test_user.py
mutantsan: This test is currently deleted from core, since we implemented validator for email uniqueness

mutantsan:
@smotornyuk I've fixed the mentioned issues and conflict.

PR #5128, opened 2019-12-18
5126 migrate datapusher&datastore cli commands

mutantsan:
Refactor the CLI commands for the datastore& datapusher extensions to use the new IClick interface rather than the old paster based one.

### Features:

- [X] includes API changes


mutantsan:
@amercader hi, may I get some help, please? I have problem with tests. We are using old paster datastore command in `config.yml` which I replaced by new one. But, I also deleted the entry point for datastore paster command, and new one accessible only when datastore plugin is "plugged".

Leave it unplugged - Error: No such command "datastore". Otherwise, I am getting problem with multiple tests. Should I try to edit that tests, or I chose the wrong approach?

mutantsan:
Tests passed, but for this I had to add another config file that runs the datastore plugin for `datastore set-permissions` commands.

test-core-cli-full-initialize.ini
```diff
@@ -0,0 +1,51 @@
+[server:main]
+use = config:test-core.ini
+
+[app:main]
+use = config:test-core.ini
+
+ckan.datastore.write_url = postgresql://datastore_write:pass@ckan-postgres/datastore_test
```
JGulic: Do we want to be consistent here with [test-core.ini ](https://github.com/ckan/ckan/blob/master/test-core.ini#L22)file and use localhost instead of ckan-postgres?

smotornyuk:
As discussed on tech team meeting, more preferable solution for disabled datastor plugin is linking datastore command globally(so that you'll be able to run `$ datastore set-permissions...` instead of `$ ckan datastore set-permission`

If you update [this place](https://github.com/ckan/ckan/blob/master/setup.py#L82) - you can register global `datastore` command and it still will be available under `ckan` group

PR #5130, opened 2019-12-19
[#4801] PY3 General Support - Part 3

amercader:
WIP


amercader:
@smotornyuk could you have a look at the changes so far? Most relate to the changes in the static files serving. Basically setting `static_url_prefix=''` in order to keep our current behaviour with extensions means that we now how a catch-all `/<filepath>` route that tries to serve the provided file from the public folder(s).

This is the exact same behaviour that we had up until know with the Pylons catch-all route, except that that controller looked for the provided path in the templates folder.
TBH I think it makes much more sense to serve automatically things in the public static folder than from the templates one, so I think that's a good move.

smotornyuk:
+1 for public files instead of `template` controller.

The only issue i found is - https://github.com/ckan/ckan/blob/py3-general-support-3/ckan/lib/helpers.py#L280 - `Py2` instead of `PY2`(uppercased). I didn't notice this reviewing previous PR

amercader:
Great thanks, will take care of the PY2 bit

amercader:
@ckan/core @kowh-ai @pdelboca Quick update on where I am on this.

App loads fine on py3, you can login and home, dataset search and dataset page work (and probably many more, I haven't tested it fully). The next step should be getting the tests running on py3. I've made the changes needed so the new pytest tests can run and now is a matter of starting fixing them! :)

If someone wants to pick up from where I left I probably won't be able to do much over the coming days so feel free to carry on. Just mention here what areas are you working on so nobody steps on each others toes.

We need to think of ways of running the test suite both on Py2 and Py3. I don't think the free plan we are on Circle CI will support that, so maybe we can investigate [GitHub Actions](https://github.com/features/actions) to see if it can be a replacement or we can combine it with Circle / Travis.

-- 

For reference, here's the setup I've been using to work on this so far. Two different virtualenvs, one running Py2, with the Py2 requirements installed (`requirements-py2.txt`) and the other running Py3 and its requirements (`requirements.txt`). Both use the same CKAN source repo.

In my case the CKAN source is installed in the py2 one, so I use the normal `paster serve` and tests commands on that venv.

To install this same CKAN source on the py3 one, just make sure to use the python executable from the relevant venv:

    ~/dev/pyenvs/ckan-py3/bin/python setup.py develop

And the same to install the requirements with pip:

    ~/dev/pyenvs/ckan-py3/bin/pip install -r requirements.txt
    ~/dev/pyenvs/ckan-py3/bin/pip install -r dev-requirements.txt 

Once this is in place, you can run CKAN in the activated py3 env with:

    ckan --config ~/dev/pyenvs/ckan/src/ckan/development.py3.ini run -p 5005

I use a custom ini file with `ckan.site_url=http://localhost:5000` and port number to be able to run both versions side by side.

To run the tests:

    pytest --ckan-ini=~/dev/pyenvs/ckan/src/ckan/test-core.ini ~/dev/pyenvs/ckan/src/ckan/ckan/tests/ -v -s

We are getting there, slowly but surely!

smotornyuk:
I'm a very lucky person and don't have any holidays this year(except for January 1st), so I'll take this branch and continue in direction of making all tests green. I haven't created PR yet, but here a few notes for anyone other, who is going to start from current PR:
* remove 'paste` and `paste-script` from requirements.py before installing it(python3) - otherwise you'll get a bunch of errors during startup
* comment `ckan.lib.cli.add_user` import inside `ckan.cli.sysadmin` - it relies on unavailable `paste-script` 
* if you are facing `TypeError: a bytes-like object is required, not 'str'` from `auth_tkt` - clean cookies. 

I'm going to create new WIP PR by the end of the day, so some additional detail may be found there

PR #5132, opened 2019-12-20
HTML emails

davidread:
Allows mailer to send HTML emails (i.e. multipart, with both a HTML and plain text fallback version). And adds emailing to the toolkit for plugins to use.

Justification: I'm currently using this in ckanext-subscribe, so I can add links without displaying the full URL. I'm sure others have asked about sending HTML in the past. HTML emails are the norm these days - so few people use text-based clients. So I think it would be a natural extension to use HTML in existing reset password and invite emails.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -6,6 +6,7 @@ import smtplib
 import socket
 import logging
 from time import time
+from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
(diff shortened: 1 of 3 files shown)


JGulic:
@amercader I've tested the PR, I think it can be merged

amercader:
Thanks for the patch @davidread and for the review @JGulic . I fixed the conflicts and merged it.

PR #5134, opened 2019-12-23
Green tests using Py3 - Part 1

smotornyuk:
Based on #5130


smotornyuk:
@ckan/core 
If anyone has time to review this PR - go ahead. It goes out of control - initially, I thought that there will be few encoding fixes and some conditional imports, but in the end, I jumped into flask_test_client as alternative to `WebTest`(latter has a number of problems related to flask's sessions - it was originally written for pylons, so there is no wonder it doesn't work well with 'pure' flask application)

Reviewing big PRs is a pretty hard thing to do, so as soon as i have green tests on Py2, i'm going to fork into separate PR(just as Adria did with `py3-general-support` series). 

Main features of current PR:
* `(test)app.get/post/etc` uses adapted `flask.testing.Client` instead of `webtest.TestApp`. In some cases, I've updated tests to use flask-style calls,  but there so many uses of `app.get/post` - and I've decided to imitate previous behavior(and obviously, I'll do some cleanup and standardization in future, but it may wait a bit, till we drop py2)
* circleCI config updated, so it possible to start Py3 testing workflow. I've commented this path because tests are not ready yet, anyway
* a bunch of `six.ensure_str/binary/text` when Py2/Py3 have different opinions about encodings
* dozen of `pytest.xfail` - I've marked places that throw `sqlalchemy.DetacedInstance` error. This error appears only in tests(i.e., if I'm doing the same via UI - there is no problem), so it related to application context and global variables. If someone will help with this - it would be super cool. But, generally, this issue has low priority
* general fixes, like `future.(print_function/absolute_imports)`, `list(dict.keys)`

smotornyuk:
Ready for review

ckan/lib/webassets_tools.py
```diff
@@ -14,6 +15,8 @@
 logger = logging.getLogger(__name__)
 env = None
 
+yaml.warnings({u'YAMLLoadWarning': False})
```
amercader: FYI I submitted a PR to webassets that got rid of the warning (https://github.com/miracle2k/webassets/pull/525) and it seems to be included in the latest version (2.0) so when we upgrade we can get rid of this.

amercader:
Fantastic work @smotornyuk . I have some minor comments but they might have been addressed on Part 2, so merging now :horse_racing:

PR #5135, opened 2019-12-26
4798 Replace paste appconfig

### LINKED ISSUE #4798: Don't use paste.deploy appconfig to parse the ini file
Right now we are using paste.deploy's `appconfig` to parse the INI file and get the `config` object that CKAN uses. It offers some functionality that we don't use (like the ability to parse an INI file hosted elsewhere via HTTP), but basically it's a wrapper around `ConfigParser`. The one feature that we need to keep for backwards compatibility is interpolation (eg things like `cache_dir = /tmp/%(ckan.site_id)s/`), which is also supported in the [standard library](https://docs.python.org/3/library/configparser.html#interpolation-of-values).

#### Approach

Write our own simpler version of the configuration file parser and replace the `appconfig` calls in the CLI code.

#### Story points

5

pdelboca:
Fixes #4798 

### Proposed fixes:
This PR replaces paste.deplopy's `appconfig` for our own configuration file parser. Not sure if this it what was expected so any feedback is welcome. I have tested it locally by running CKAN and executing cli methods like `add user`, `sysadmin`, `search-index rebuild_fast` and also added a test.

I have extracted the design from the [paste module](https://github.com/Pylons/pastedeploy/blob/9cbc584a86bd1ae425bd0c0c3c00617d7e59475d/paste/deploy/loadwsgi.py) to maintain the api structure of the `conf` object that is being call in other methods like `make_app` or `load_environment`. 

### Features:

- [ ] includes API changes


pdelboca:
Tests are failing because it is lacking the logic for when the `.ini` file uses other config file. I will look into it! 

The `appconfig` method from `paste.deploy` uses a set of recursive calls to solve this which will make our configuration parser more complex.

pdelboca:
Ok! 

All test are passing and the initial version of our own parser is ready. I took the idea from [this module](https://github.com/Pylons/pastedeploy/blob/9cbc584a86bd1ae425bd0c0c3c00617d7e59475d/paste/deploy/loadwsgi.py) and tried to be keep the current api of the config object (to be called with `config.local_conf` and `config.global_conf`) but this can also be simplified with further changes to simplify the code.

I will be nice to have some feedback about this issue to check if I'm on track and if this is what was expected!

ckan/cli/__init__.py
```diff
@@ -5,9 +5,60 @@
...
+        conf = AttrDict(self.global_conf)
+        conf.update(self.local_conf)
+        conf.local_conf = self.local_conf
+        conf.global_conf = self.global_conf
+        conf.context = self
+        return conf
+
+class AttrDict(dict):
+    """
+    A dictionary that can be assigned to.
```
sivang: @pdelboca can you please perhaps provide more info as to why we need this dict and why it's used? i.e. could you just use a local dict constructor within the `def config(self)` method and just call it conf; so something like:

```
def config(self):
        conf = dict(self.global_conf)
        conf.update(self.local_conf)
....
```
(- https://python-reference.readthedocs.io/en/latest/docs/functions/dict.html)

pdelboca: The current protocol of the config object is a mix between object and dictionary: 
 - sometimes their attributes are access using `conf.local_conf` 
 - sometimes using `conf.get('ckan.site_url', 'http://0.0.0.0')`
 - sometimes they are separated into local_conf and global_conf and each one of those should be dictionaries that can be accessed with `global_conf['__file__']`

This `AttrDict` class implementation was taken from [Pylon's pastedeploy](https://github.com/Pylons/pastedeploy/blob/9cbc584a86bd1ae425bd0c0c3c00617d7e59475d/paste/deploy/loadwsgi.py#L726) and it respect the current implementation.

Another option is to drop everything and use a simple conf `dict`, with no `local_conf` nor `global_conf` and refactor every call to it in the code. Maybe someone in the @ckan/core can give some insights on what is expected.

sivang: @pdelboca @amercader @Zharktas now it's clear, so we're wrapping it in an `object` to be able to assign to it. 

In my taste this is a bit wrong as composition should have been applied here to avoid ambiguity (but also to have readability :))  , so we should have a dict part that is explicitly accessible and an arbitrary namespace, explicitly accessible to set values into.

I propose we take this opportunity to clean up the usage of this, what do you think?

sivang: @pdelboca did you confirm the the `create()` method is redundant btw (https://github.com/ckan/ckan/pull/5135#pullrequestreview-337696537) ?

pdelboca:
I made some changes to simplify the code. One object containing `global_conf` and `local_conf` dictionaries to keep consistency whit what is being to initialize both Pylons and Flask apps.

Some info:
 - The parser assumes our `.ini` files will contain an `[app:main]` section and a `use=....` config.
 - For our `test.ini` files that has `use=config:test-core.ini` the parser will parse both files and apply interpolation of variables.
 - The `debug=true` config is part of the `[DEFAULT]` section in our `test.ini` file, however in the `test-core.ini` file the variable is changed inside the `[app:main]` section to `debug=false`. This will not change the value since the `local_conf` dictionary don't contain values already available in `global_conf`. This is the previous behaviour, should we keep this?

ckan/cli/__init__.py
```diff
@@ -5,18 +5,66 @@
 import click
 import logging
 from logging.config import fileConfig as loggingFileConfig
```
amercader: This one can be removed now

pdelboca:
@amercader I did both the small fixes you proposed!

Agree with the refactor of the `make_app` code. I made some changes locally but need to fix tests so It will take a little bit more. I will suggest to merge this and start working on those changes in another PR (I can do it) just to avoid unnecessary delays with this tasks that is required for the Python 3 migration.

amercader:
Thanks for your work on this @pdelboca !

amercader:
I created a new version for the refactoring discussed: https://github.com/ckan/ckan/issues/5197

pdelboca:
Great @amercader ! You can assign #5197 to me, I have some work done so I hope I can tackle it this week :)

PR #5138, opened 2020-01-02
Green tests using Py3 - Part 2

smotornyuk:
Follows #5134


amercader:
@smotornyuk can you merge master into this? For some reason Github is not picking up the changes from Part 1

smotornyuk:
Done

amercader:
All good, only one text failure after merging master

smotornyuk:
Updated - all green again

amercader:
:boom: let's get down to business!

Thanks for your work on this @smotornyuk

frafra:
@smotornyuk Do not use paste.fileapp c1c714d
Could you please comment on that?
I see that it is still used by CKAN 2.9 (dev): https://github.com/ckan/ckan/blob/da2da8e1b48c82815c411f70816f05f2c81064ff/ckan/controllers/package.py#L11
Should that be changed too?

smotornyuk:
No, all the `controllers` are not used in 2.9. They stay in the codebase just for reference. 

And as for this:
> Do not use paste.fileapp

do you understand that you are referring to the code, written two years ago, when `paste.fileapp` wasn't a problem:)

smotornyuk:
No, you can ignore it. Controllers are not used in 2.9 and just left in codebase in order for reference

frafra:
Wouldn't it make sense to remove controllers? It is a confusing reference, since it is code referring to an old version, which is no more supported.

smotornyuk:
They are already removed in master. In 2.9 we decided not to remove them in order to simplify backports of fixed issues. 
For example, I've fixed Flask's view and applied the same fix to the controller(even though it has no effect). When someone prepares a patch release, this person has to cherry-pick my commit, without any extra efforts spent on adapting my fix. 

Now we are trying to do backports as soon as the fix got merged into master, so such a measure has no value(and controllers are removed from master anyway). But as we are trying not to change much in the existing branch, nobody has removed controllers:) 
If you are working with 2.9 and controllers are bothering you, feel free to remove them. I don't think that somebody will mind merging a PR with such a change

Zharktas:
I wouldn't remove controllers from 2.9 as they are still used if you run ckan on python 2 and removing them would essentially drop python2 support.

frafra:
> I wouldn't remove controllers from 2.9 as they are still used if you run ckan on python 2 and removing them would essentially drop python2 support.

Ok, sorry, I missed that :)

Zharktas:
Or at least the extensions might rely on those controllers existing, if not in use directly by ckan :)

PR #5140, opened 2020-01-03
Add Webassets docs

smotornyuk:


doc/contributing/frontend/assets.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Libraries can be added to CKAN from extensions using a helper function
+within the toolkit. See below.
+
+In debug mode assets are served un-minified and unbundled (each asset is
+served separately). In non-debug mode the files are served minified and bundled
+(where allowed).
+
+.. Note::
+    .js and .css assets should be supplied as un-minified files.  Minified
```
amercader: @smotornyuk Two questions here:

1. Do you mean that when commiting files to this repo or an extension one you should not commit the minified ones?
2. With webassets do we still need to build the minified files with `ckan minify` or whatever before a release?

amercader:
This is great @smotornyuk , I added a few small changes in e1e47762f.
Can you clarify the two questions above? cheers

doc/contributing/frontend/assets.rst
smotornyuk: 1. Exactly - you need to commit only original source, without any preprocessing.
2. Generally, no. All JS files with `rjsmin` are minified automatically when `debug = false` (example: https://github.com/ckan/ckan/blob/master/ckan/public/base/javascript/webassets.yml#L2) As for css, right now styles are not minified as there is no built-in/default css-minifier provided by webassets. I think about including one additional dependency, so that styles will be handled in the same way as scripts, but i want to check first, whether there is some option that doesn't require updating dependencies

PR #5143, opened 2020-01-08
Run CKAN with Solr 8.4

smotornyuk:
While we are introducing some great changes, it may be a good time to add a few experimental upgrades. Available out-of-the-box Solr v4/5 is not supported anymore. From the point of view of existing portals, I don't want to do an aggressive switch to a newer version of solr. But there are constant issues that came from newcomers, who install the newest available Solr and cannot get CKAN in a more or less working state.

Previously, we discussed two separate schemas - #4387 (which wasn't a great idea, just some workaround). And now i suggest using new experimental built-in extension - `managed_search_schema`. While it is not enabled - everything works just as before(with solr 5/4). But, if you do the following things:
* install solr 8.4(just try docker image) and create new core(let's call it `ckan`)
* enable `managed_search_schema`
* run `ckan  search-schema create` 
* manually update `your-solr-data-dir/ckan/conf/managed-schema` by replacing `<uniqueKey>id` with `<uniqueKey>index_id` (while you are there, update `name` and `version` attributes of the root `schema` tag)
* restart solr
you'll get working CKAN instance with newest solr version.

What else needs to be done:
* tests - maybe just add additional docker image of solr to circle-ci config and run `test_search` twice - with default solr5 and with docker-solr8
* docs - it should be as clear as possible, that solr8 support is experimental and some extension won't work with(probably, ckanext-spatial)

Any ideas or suggestions?

PS i've added `ISearchSchema` interface, which provides a way for extension to define new and replace existing fields. It's used during `ckan search-schema create`


MurraySpeight:
Is this currently the best branch to get CKAN running with Solr 8?

smotornyuk:
I haven't a chance to work on it so it's maybe quite behind current master, but there are no alternative branches so you can give it a try

MurraySpeight:
> I haven't a chance to work on it so it's maybe quite behind current master, but there are no alternative branches so you can give it a try

Thanks Sergey. I'm new to CKAN so I opted for using Solr 6. Although it is end of life so not ideal, it is well documented and supported by CKAN.

adborden:
@smotornyuk I'm looking at the files changed, but it's showing no changes. Is it possible to restore your branch, it's a valuable set of changes!

smotornyuk:
Sure, sorry. I was going to work on a different branch with a bit more global solution, that's why I've closed this one. But if it has any value, I'll restore it, till working alternative is ready

smotornyuk:
@adborden , i've opened different PR #6530 that only changes schema.xml and few lines inside CKAN codebase. 
You can try these changes and, for example using dockerized Solr v8.0. At moment all the tests pass so it looks like it generally works:)

Docker solr recipe:

```
docker create --name solr80 -p 8983:8983 solr:8.0
docker start solr80

docker exec -it solr80 solr create -c ckan
docker cp ckan/config/solr/schema.xml solr80:/var/solr/data/ckan/conf/managed-schema
docker restart solr80

# Put the following line into your config: solr_url = http://127.0.0.1:8983/solr/ckan

```

PR #5145, opened 2020-01-08
bin/less -> gulp

smotornyuk:
Gulp is a little bit easier to read(at least, it has documentation) and is better adopts new tasks if we are going to refactor front-end files


Zharktas:
After this is merged, we probably should should install vendor javascript libraries like jquery through npm so that those would be updated more often and we could get automated security notifications of those.

smotornyuk:
And I thought about rewriting/updating CKAN.js(modules, clients, etc) with TypeScript so that we can generate definitions for objects and suggest some scenarios for using it in extension development. I found among my colleagues, that a lot of people prefer not to peek into source code and implement a lot of existing tools/methods from scratch(for example, I often saw manually written ajax requests to API instead of using `ckan.client.call`, etc.)

Issue #1212, opened 2013-08-27
API key stored as plain text


bakersemail:
Hi CKAN devs and OKFN,

Noticed that the API key is stored as plain text in the database. Since the API key is essentially being treated as a username/password for Data API functions (creating and changing datasets, resources, etc) the API is very sensitive.
I am concerned that if our database is leaked, we would need to update all the API keys ASAP or risk data being updated until we do so.

Should the API key be stored encrypted or hashed to increase the attack time potential if such a leak occurs?

LaurentGoderre:
I agree with @nigelbabu, if it acts like a password, it should be treated as a password

wardi:
Symmetric encryption of our current API keys in the DB is a great idea. Let's do this.

Removing the default creation of the current API keys for users is a good idea, but I'd want a configuration option for the behaviour, or sites that depend on new users having readable API keys will break.

When we add support for multiple API keys for users we could store them labelled (users give them names like "used by my import script im12345") and hashed so they can't be recovered after they are generated. It should still be possible to create a legacy API key for users that need them.

LaurentGoderre:
I like that idea

davidread:
This issue was closed due to inactivity. Feel free to reopen if you have more feedback or are interested it working on it

michab23:
Hi,
Is there any solution for this problem with the newest CKAN versions?
Does API key still is stored as plain text in DB?

wardi:
As far as I'm aware no one has worked on the code to better manage users API keys, but it'd certainly be a welcome contribution.

boykoc:
> Symmetric encryption of our current API keys in the DB is a great idea. Let's do this.
> 
> Removing the default creation of the current API keys for users is a good idea, but I'd want a configuration option for the behaviour, or sites that depend on new users having readable API keys will break.
> 
> When we add support for multiple API keys for users we could store them labelled (users give them names like "used by my import script im12345") and hashed so they can't be recovered after they are generated. It should still be possible to create a legacy API key for users that need them.

I like this approach. It in theory could be implemented incrementally in this way and start flagging for deprecation in future versions of CKAN.
- [ ] treat API keys like passwords (encrypt of current key setup in DB) with config option to prevent breakage (possibly show API key only once to user on generation, never store key unhashed/unencrypted and still allow re-generation of key)
- [ ] multiple keys for users
- [ ] ability to revoke individual keys (user or sysadmin)
- [ ] labeled keys for users

[...]

PR #5146, opened 2020-01-08
API Tokens 

smotornyuk:
Python 3 migration should be finished in the near future and I'll probably have some time to continue with "React.or.somethin.similar" frontend for CKAN. Back then, when I was actively working on it, I thought that it would be cool to have some kind of API authentication endpoint. I don't want to invest a lot of time into something unreasonably complex - we need to discuss the future of `repoze.who` and `flask-auth` first. Instead here just some simple solution. This endpoint is forbidden for users, except for cases, when config file contains explicit `ckan.auth.get_apikey_via_api` option, so there is no security issues. But when this action is enabled, it provides really simple way to obtain your api_key(as long as you know your credentials:) )


wardi:
-1 from me. For a JS front-end we should be using user name+password to get a session cookie or equivalent, just like the normal front-end.

API keys should only be used from scripts or for server-server communication. Giving API keys to client side JS makes XSS attacks much worse, because the credentials don't time out and don't let us check the referrer (which is all we can use until we get CSRF tokens on all our forms).

smotornyuk:
From my point of view, it this way has almost the same set of problems:
1. HTTP_REFFERER would be added to any ajax request by browser, so there should be no difference
2. Sessions have an expiration time, but it's unlikely that anyone would ever call `logout` - won't it create an enormously huge amount of session files(or whatever we use for storing session data) if someone implements request to login handler on every page_load?

I also think that it is not the best approach, rather some placeholder that may be useful while we don't have other options. Using session cookies in the modern web doesn't sound very cool and having any popular authorization system for AJAX would show CKAN from a nice way to the public and may bring us more users. 
BTW, having a discussion on this topic and maybe some agreement on implementation is much better, than merging this raw change into the core, so I mark it for discussion.

smotornyuk:
After some debates, we agreed on the implementation of token system(in addition to API keys).
New features:
1. API:
  * api_token_create - requires `user` and `name` and creates API Token for this user with the specified label
  * api_token_list - requires `user` and lists all API Tokens for this user
  * api_token_revoke - requires `token` and removes API Token by specified id
2. UI:
  * Under user profile add new tab `API Token`
  * Contains input field(required) for token's label and button `Generate new token`
  * Shows a table with the label, date of last access for token and "Revoke" button(with confirmation dialogue)
3. Interfaces:
  * IApiToken - contains two methods:
    * preprocess_api_token(token, original) - do anything with the token, before it used for used identification
    * postprocess_api_token(token, original) - do anything with the token, after it created and before it is shown to the user
4. DB:
  * new table 

API tokens can be used instead of API Keys. Actually, if CKAN fails to find a user by API-key, it assumes, that API-token was provided and makes an attempt to find the user by API token. In contrary to API-keys, a token is supposed to be used by frontend-clients. It stores the date of last use so that users can notice if a token was used by someone else. I don't store the IP for the last use of token, because it's likely a violation of privacy rules. And token can be removed(revoked) at any moment if misuse is discovered.

Developers are likely will miss IP recording, use-rate-limits, permission tags, etc. Because of this, I've added interface that allows modifying (thus, the addition of some info to the token) before it is shown to the user and pre-parsing after it received from used but before it was used to query the database. Token itself is used as `id` in the relevant DB table. It generated using the `secrets` module(built-in in py3 and installed as a polyfill for py2) and stored as-is. If extension modifies token before it is shown to the user, we don't know this modification and token will be lost if not copied right after creation(took this idea from GitHub). Theoretically, I can just re-run all post-processing from extensions, in order to show API key multiple times, but it will add idempotence as a restriction to post-processing and side-effects may become buggy. Instead, I clearly state that API token should be copied immediately after creation.

ThrawnCA:
> Giving API keys to client side JS makes XSS attacks much worse, because the credentials don't time out and don't let us check the referrer (which is all we can use until we get CSRF tokens on all our forms).

If XSS is possible, there are already multiple ways to retrieve the API key, notably the `user_show` API and the user profile page. And XSS is capable of forging headers, too, so referrer and even Origin checks are ineffective.

rufuspollock:
@smotornyuk let's sync on this as we've been doing work on this at Datopian - maybe we can some up with a solution together.

smotornyuk:
@rufuspollock , sounds really great - i'm in:)

rufuspollock:
@SergK great - let's catch up 😄

amercader:
@rufuspollock can you briefly summarize here your followed approach, constraints, etc? Would be really good to have detail from a working implementation on this.

rufuspollock:
@smotornyuk hey - let's sync again. We've now got something close to a working extension on both authentication and authorization and can push this. Obviously if we can get something into core that's even better!

rufuspollock:
@smotornyuk generally could i suggest we open an issue in e.g. ideas-and-roadmap repo and get clear job stories and spec there (it makes it easier to follow than this thread). wdyt?

smotornyuk:
Perfect - this will open the possibility to share ideas and suggestions from others, so it won't be solution that suits only two of us:)

boykoc:
@smotornyuk  and @rufuspollock was an roadmap issue created by chance? I'd like to link to it from here to help people find it. I didn't notice one but could have missed it.

Also, is this planned to help address #1212? From what i'm seeing I think it would (in which case just cross-linking to help referencing/documenting). It seems based on the comments above it will address the main issues of storing the token as raw text and displaying multiple times.

smotornyuk:
Hi. No, there is no roadmap yet.

There is one extension that solves security issues - https://github.com/datopian/ckanext-authz-service/ . It uses permission tokens that are created "on-fly" and not stored in the database. I would say that this extension is a must-use for anyone who is going to create a distributed system using independent services. 

Current PR is going to focus on multiple API keys(API tokens) that are stored in DB and its main auditory - monolith applications.  And yes, it should help in resolving of linked issue, because all the API Tokens will be encrypted and the owner will be able to revoke any of them. As for the main(initial) API Key, it won't be changed(encrypted) as a part of the current PR. But when this PR will be stable, we can decide future strategy and gradually apply changes to main API Key. If, for example, everyone start using **API Tokens** in CKAN v3.0, then encrypting the main API Key in CKAN v3.1 won't break anything.

rufuspollock:
@boykoc we did a nice presentation of the issues and approach and there's a zoom recording of the meeting - you can find more info here https://github.com/ckan/ideas-and-roadmap/issues/247

smotornyuk:
A lot has changed, so in order to reduce your efforts, I'll provide summary here. You can just ignore all comments above.

This PR introduces **API Tokens**, as an alternative to **API Key**. Probably, you've used different kinds of tokens in other APIs before and you have some picture in mind. From my experience, there is two common types of tokens:
* expirable, not controlled. The ones, used by GoogleAPI, FacebookGraphAPI. You are requesting such token either via a pop-up in UI, opening some one-time link, or providing your credentials. Token generated and can be used till expiration date(that's what I mean by uncontrolled - you cannot invalidate it at any moment). Good point here - such token will expire in an hour or two, so even if someone still it from you, very soon this token will become invalid. And such tokens are not stored in DB, so there is no way to get user's token if you have read access to a database(it's the problem with the current implementation of **API Key**). Current PR is not implementing such kind of token. There is already [existing implementation](https://github.com/datopian/ckanext-authz-service/) which I like, so I won't reinvent the wheel.
* controlled tokens. Think about GitHub tokens. You create one, copy its value and it can be used as long as it exists. You can revoke token using GitHub UI, that's what I mean by controlled. Current PR implements such kinds of tokens.

There are some issues, that may come in mind. Below i'll explain, how API Tokens solve the ones, that were discussed in previous comments, and if you see any other problem - feel free to ask questions:

### What if someone just read the database?

Tokens not stored inside the database. Only token id (`jti` claim from JWT) is stored inside the database. You can generate valid token only if you know secrets (they can be specified explicitly, or, by default, they are the same as session secret). Even more - you can use public/private keys for signing the token. Or, you can provide a custom encoding/decoding strategy via the ``IApiToken`` interface, if the default one is not sufficient. 

### What if someone steal my token?

There actually two ways to solve this problem. 

1. **API Tokens** tab under the user's profile shows the last access time for each token. If you notice that access time is different from what you know, you can revoke the token.
2. There is a built-in ``expire_api_token`` plugin. when it's enabled, every new token will have expiration time. By default, every token will expire in an hour. The default value is controlled by config file and there are few fields in UI that allow us to configure life-time for each token individually.



In addition, I tried to make this functionality configurable enough, so it should be possible to implement a custom decoding\encoding strategy, add some additional fields to the token itself, or perform side-effects during creation token or identifying the user by the token.

boykoc:
@smotornyuk at first review of your write up it looks good to me.  awesome work on this!

ckan/lib/api_token.py
```diff
@@ -0,0 +1,122 @@
...
+    if type_ == u'file':
+        with open(value, u'rb') as key_file:
+            value = key_file.read()
+    if not value:
+        log.warning(
+            u"Neither `%s` nor `%s` specified. "
+            u"Missing secret key is a critical security issue.",
+            config_key,
+            _config_secret_fallback,
+        )
```
wardi, line 45: Let's refuse to start ckan if there's no session secret instead of logging and continuing.

ckan/logic/action/delete.py
```diff
@@ -768,3 +769,31 @@ def job_cancel(context, data_dict):
         log.info(u'Cancelled background job {}'.format(id))
     except KeyError:
         raise NotFound
+
+
+def api_token_revoke(context, data_dict):
+    """Delete API Token.
+
+    :param string token: Token to remove(optional when `jti` specified).
+    :param string jti: Id of the token to remove(optional when `token` specified).
```
wardi, line 778:
```suggestion
    :param string token: Token to remove(ignored when `jti` specified).
    :param string jti: Id of the token to remove(optional if `token` specified).
```

ckan/lib/api_token.py
smotornyuk: ok. I hesitated whether it should be reported as an exception(personally I prefer to prevent any possibility of running instance in an obviously invalid state) or just notify that something is potentially broken.

wardi:
I think this is good to go. Any reservations @amercader ?

ckan/views/user.py
```diff
@@ -150,6 +151,92 @@ def read(id):
...
+            u'class': u'btn btn-default btn-xs',
+            u'data-module': u'copy-into-buffer',
+            u'data-module-copy-value': ensure_str(token)
+        })
+        h.flash_success(
+            _(
+                u'API Token created: <code style="word-break:break-all;">'
+                u'{token}</code> {copy}<br>'
+                u'Make sure to copy it now. '
+                u'It can be impossible to see it again!'
```
wardi, line 222:
```suggestion
                u'Make sure to copy it now, '
                u'you won't be able to see it again!'
```

changes/5146.feature
```diff
@@ -0,0 +1,9 @@
+API Tokens: an alternative to API keys. Tokens can be created and
+removed on demand(check :ref:`api authentication>`) and there is no
+restriction on the maximum number of tokens per user. Consider using
+it instead of them API key and prefer the creation of a separate token for
+each use-case instead of sharing the same token between multiple
+clients. In original implementation API Tokens are JWT, but the goal
```
wardi, line 6:
```suggestion
restriction on the maximum number of tokens per user. Consider using
tokens instead of API keys and create of a separate token for
each use-case instead of sharing the same token between multiple
clients. By default API Tokens are JWT, but the goal
```

wardi:
just a couple small text changes. Let's aim to disable generation of API keys in default installs of ckan 3.0

ckan/logic/action/create.py
```diff
@@ -1574,3 +1575,42 @@ def follow_group(context, data_dict):
...
 
     return model_dictize.user_following_group_dictize(follower, context)
+
+
+def api_token_create(context, data_dict):
+    """Create new API Token for current user.
+
+    You must provide your API key in the Authorization header.
+
+    :returns: a representation of the 'api_token'
```
amercader, line 1585: You need to document the input parameters in the data dict, `user` (user name or id?) and `name`.

Can you describe the returned key here? eg: Returns a dict with the key `token` containing the encoded token value.
Is it worth describing the internal decoded structure (ie `jti` and `iat` keys), I guess not as this is an internal implementation detail

ckan/views/user.py
```diff
@@ -150,6 +151,92 @@ def read(id):
...
+        return h.redirect_to(u'user.api_tokens', id=id)
+
+
+def api_token_revoke(id, jti):
+    context = {u'model': model}
+    try:
+        logic.check_access(u'api_token_revoke', context, {u'token': jti})
+    except logic.NotAuthorized:
+        base.abort(403, _(u'Unauthorized to revoke API tokens.'))
+    model.ApiToken.revoke(jti)
```
amercader, line 235: Should we wrap this into a `api_token_revoke` action? Even if it's a one line action. This will allow extensions to override it or chain it, otherwise it breaks our pattern of views  > action/auth > models

ckanext/expire_api_token/plugin.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
```
amercader, line 1: This sounds like a really useful feature, any reason to not include it in the core implementation and use a core extension instead?

doc/api/index.rst
```diff
@@ -251,9 +251,11 @@ functions and configuration as the web interface, so if a user is authorized to
...
 well.
 
-When calling an API function that requires authorization, you must authenticate
-yourself by providing your API key with your HTTP request. To find your API
-key, login to the CKAN site using its web interface and visit your user profile
+When calling an API function that requires authorization, you must
+authenticate yourself by providing your API key or API token with your
+HTTP request. To find your API key, login to the CKAN site using its
+web interface and visit your user profile page. To generate new API
+token visit **API Tokens** tab available from your puser profile's
```
amercader, line 258:
```suggestion
web interface and visit your user profile page. To generate a new API
token visit the *API Tokens* tab available on your user profile.
```

Also can you rename this section to "Authentication, API keys and API tokens" and replace "(Replacing XXX with your API key.)" with "(Replacing XXX with your API key or token.)"

If we want to encourage people to use them and deprecate the API key we should make them more prominent in this version docs.
Is it worth putting a `..note::`  saying that API keys will get deprecated/removed in the next version?

doc/maintaining/configuration.rst
```diff
@@ -771,6 +771,76 @@ Allows dataset collaborators to change the owner organization of the datasets th
...
+  api_token.jwt.encode.secret = file:/path/to/private/key
+
+Default value: same as ``beaker.session.secret`` config option with ``string:`` type.
+
+A key suitable for the chosen algorithm(``api_token.jwt.algorithm``):
+
+* for asymmetric algorithms: path to private key with ``file:`` prefix. I.e ``file:/path/private/key``
+* for symmetric algorithms: plain string, sufficiently long for security with ``string:`` prefix. I.e ``string:123abc``
+
+Value must have prefix, which defines it's type. Supported prefixes are:
```
amercader, line 806:
```suggestion
Value must have prefix, which defines its type. Supported prefixes are:
```

ckanext/expire_api_token/plugin.py
smotornyuk: I'm obsessed with extendibility in this PR because I have no ideas what others may expect/want to have. And, if you want to use a calendar widget for selecting the expiration date of token instead of specifying its lifetime, you can just create new extension(of even inherit from `expire_api_token`) with desired fields/validation/widget and use this extension instead of `expire_api_token`. 
If it becomes the part of default implementation, you will have to use existing format (`units` * `duration`) and that leads us to a number of hacks, like using JS and hidden fields for converting date from calendar into the required format

ckan/logic/action/create.py
smotornyuk: Updated. 
Yes, `jti` and `iat` are implementation-specific and generally, they won't be used by end-user directly. You are just copying encoded token and using it as is. Probably, the majority of users won't even know, that there are some data inside the token

ckan/views/user.py
smotornyuk: Yes, updated

doc/api/index.rst
smotornyuk: Done. I've added a note about deprecation of api keys. I don't think that we'll really remove them before 3.1/3.2, so the note is quite vague:)

ckanext/expire_api_token/plugin.py
wardi: I think it's ok as a core extension because it provides a non-toy example of extending API Tokens.

wardi:
> This is excellent work @smotornyuk! I added some comments and questions.
> I think we need a big notice on the changelog saying that API keys will no longer be generated on the next version, ie they are deprecated on CKAN 2.9 and will be removed on CKAN 3.0

We can add a configuration option to the next version that enables automatic creation of API keys default to false if not provided (or even this release if there's still time, default true for 2.9) API keys are fine for sites that are internal and tightly controlled and sites that want to maintain compatibility with existing third-party integrations so I wouldn't want to see the feature removed completely

amercader:
@wardi @smotornyuk I actually think that creation of API keys *on new users* should default to False on 2.9 but let's discuss on #5488

PR #5153, opened 2020-01-10
[#5152] Fix wrong resource URL after ValidationErrors

### LINKED ISSUE #5152: Validation errors during resource creation lead to wrong URLs
This was hard to track as it doesn't happen in a standard CKAN install, as the default resource form doesn't have required fields.

If an extension implements custom validation in the resource schema, and there are errors returned in the form the File upload field will be set to the previously uploaded file name (ie a string).

When submitting, CKAN will interpret that like an external URL, and create a bogus `http://<file_name>` URL. There will be no errors returned so it's easier for users to assume that the file they first selected was successfully uploaded.

To reproduce:
1. Modify the resource schema, eg adding the `not_empty` validator to `description`
2. Go to the create resource page, select a file to upload, eg `libraries.csv`, leave the description empty and submit
3. You get redirected to the form page, with the description error, and also with "libraries.csv" set on the File field:
Screenshot_2020-01-10 Add resource - Test URLs 4 - CKAN(1)

4. Submit the form. The resource is created, and has `http://libraries.csv` as URL.

The safest fix I can think of is resetting the upload field if there are validation errors. This will mean that users need to select the file again, but this is preferable to create a bad URL (we can not reset the field to the actual file previously selected because of browser security limitations). We can make this obvious to users with a message.

amercader:
Fixes #5152

### Proposed fixes:
Clears the `upload` and `url_type` fields if there is a validation error to avoid submitting the file name.
Adds a warning for users if they had previously selected a file for upload, as it's easy to miss it, creating an empty resource:

Screenshot_2020-01-10 Add resource - Test URLs 4 - CKAN

Happy to change the language being used if it doesn't seem clear enough

### DIFF
--- a/ckan/public/base/css/fuchsia.css
+++ b/ckan/public/base/css/fuchsia.css
@@ -8150,6 +8150,11 @@ fieldset[disabled] .control-custom.disabled .checkbox.btn.focus {
 .js .image-upload .btn-remove-url .icon-remove {
   margin-right: 0;
 }
+.js .image-upload .error-inline {
+  margin-top: 5px;
+  margin-left: 2px;
+  font-weight: bold;
+}
 .add-member-form .control-label {
   display: block;
 }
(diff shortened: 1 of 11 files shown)


amercader:
Ported an adapted version to `dev-v2.8` in 3607bcfdc. 2.7 was too complicated

Issue #5142, opened 2020-01-07
Default license when creating dataset via API


pdekraker-epa:
CKAN 2.8

For datasets created via the API if a license is not provided the resulting license_id and license_title are None. If you browse to the page associated with the dataset no license is displayed. If you then go to manage the dataset through the web interface the dropdown defaults to the first item in the list. When clicking update the license is changed, which may not have been intended. 

I think adding a default ('notspecified') in the package model would resolve this, but was unsure if that is the proper remedy.

PR #5154, opened 2020-01-10
[#5142] add default license when create dataset via API

mutantsan:
Fixes #
Set default license as `notspecified` when creating dataset via API. Currently, it's `None` if not provided and will be changed on first entity from select box after package update from UI. This behaviour may not be obvious for some users.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/templates/snippets/license.html
+++ b/ckan/templates/snippets/license.html
@@ -2,19 +2,18 @@
   {% if 'license_url' in pkg_dict %}
     <a href="{{ pkg_dict.license_url }}" rel="dc:rights">{{ pkg_dict.license_title }}</a>
   {% else %}
-    <span property="dc:rights">{{ pkg_dict.license_title }}</span>
+    {% if pkg_dict.license_id %}
+      <span property="dc:rights">{{ pkg_dict.license_title }}</span>
+    {% else %}
+      <span>{{ _('No License Provided') }}</span>
+    {% endif %}
   {% endif %}
 {% endmacro %}
 
 {% block license %}
   {% if text_only %}
-    {% if pkg_dict.license_id %}
-      {{ license_string(pkg_dict) }}
-    {% else %}
-      {{ _('No License Provided') }}
-    {% endif %}
+    {{ license_string(pkg_dict) }}
   {% else %}
-    {% if pkg_dict.license_id %}
       {% block license_wrapper %}
         <section class="module module-narrow module-shallow license">
           {% block license_title %}
@@ -34,6 +33,5 @@
           {% endblock %}
         </section>
       {% endblock %}
-    {% endif %}
   {% endif %}
 {% endblock %}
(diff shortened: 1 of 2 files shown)


smotornyuk:
The problem still there - after portal upgrade all datasets that were created without a license will take the first option from license dropdown on update.

We've discussed this problem on tech meeting - a much better solution would be to remove/map explicit `not specified` to an empty value and display empty license as `Not Specified`. Thus, two cases will be fixed:
* Dataset with an empty license will explicitly select `Not Specified` during update(because dataset has no license and `Not Specified` has empty value - probably you need to update HTML a bit)
* Datasets with explicit `not specified` still rendered correctly and will reset license_id on the first update

mutantsan:
@smotornyuk Hi! I've changed the approach, can you check it please?

ckan/views/dataset.py
```diff
@@ -754,6 +754,12 @@ def get(
         pkg = context.get(u"package")
         resources_json = h.json.dumps(data.get(u'resources', []))
 
+        # if dataset was created via API, license_id is None
+        # set it to nonspecified to avoid accidental update
```
smotornyuk: How it prevents accidental updates? From my point of view:
1. > dataset was created via API, license_id is None
2. Edit form show, `license_id` set to `notspecified`
3. User DOESN'T change anything and clicks **Save**.
4. `notspecidied` stored inside DB as new license value <- **ACCIDENTAL UPDATE**

smotornyuk: Does this really prevent accidental updates? What about cases, when custom licenses are used instead of built-in ones and there is no 'notspecified' option?

mutantsan: Yes, I understand now. So what is the best approach to fix it? Add additional field `notspecified`  if user list of licenses was provided? Fix in `select2.js` to prevent accidental choose? Make license mandatory in package_create API method? Or ..?

smotornyuk: As i said a month ago:
> We've discussed this problem on tech meeting - a much better solution would be to remove/map explicit `not specified` to an empty value and display empty license as Not Specified.

smotornyuk:
I think I see your problem. Let's stop on a simpler solution. Here're user stories:
1. When I'm updating dataset that already has non-epmty `license_id` - nothing should be changed.
2. When I'm updating dataset that has empty `license_id` - nothing should be changed.

In order to satisfy the first story, you don't need to do anything. For the second case, you need to add an empty option to license select box AND render an empty license in the same way as `notspecified`. This will solve problem for pure CKAN instances. Portals, that has required `license_id` won't be affected, as they won't be able to store empty choice. Portals, that has no `notspecified` won't be affected, as you just use custom representation of empty `license_id` instead of performing any actual change. How does it sound? If you see any problems with this solution, let's discuss it BEFORE you've wrote more code.

ckan/templates/snippets/license.html
```diff
@@ -2,19 +2,18 @@
   {% if 'license_url' in pkg_dict %}
     <a href="{{ pkg_dict.license_url }}" rel="dc:rights">{{ pkg_dict.license_title }}</a>
   {% else %}
-    <span property="dc:rights">{{ pkg_dict.license_title }}</span>
+    {% if pkg_dict.license_id %}
+      <span property="dc:rights">{{ pkg_dict.license_id }}</span>
```
smotornyuk, line 6: As i understand, it will print machine readable name of license. i.e, **cc-by** instead of **Creative Commons Attribution**

PR #5165, opened 2020-01-21
Fix file replacement bug

### LINKED ISSUE #5102: Uploading a resource with the same name does not update the file contents
**Note:** Are you submitting a security related issue that could be a potential vulnerability? Please send it to security@ckan.org instead.

The uploaded file should replace the previous file

The uploaded file is lost, and the previous file can only be downloaded

1) Upload a file called "test.txt"
2) Upload a file called "test.txt" with different contents
3) Download "test.txt" and inspect contents

g-rock:
Fixes #5102

### Proposed fixes:
Invalidates cache by using download attribute

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/package/resource_read.html
+++ b/ckan/templates/package/resource_read.html
@@ -33,7 +33,7 @@
             {% if res.url and h.is_url(res.url) %}
               <li>
                 <div class="btn-group">
-                <a class="btn btn-primary resource-url-analytics resource-type-{{ res.resource_type }}" href="{{ res.url }}">
+                <a class="btn btn-primary resource-url-analytics resource-type-{{ res.resource_type }}" href="{{ res.url }}" download>
                   {% if res.resource_type in ('listing', 'service') %}
                     <i class="fa fa-eye"></i> {{ _('View') }}
                   {% elif  res.resource_type == 'api' %}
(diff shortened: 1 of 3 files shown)


smotornyuk:
Won't this change make user forcibly download content even if the resource does not contain uploaded file, but just links to some external web page instead?

g-rock:
If a user uploads a resource using the "Link" button they will be shown the "Go to Resource" button and not the "Download" button.

g-rock:
Removed the attribute from `resource_view.html` because that button will display regardless of whether the resource is a file, or an URL to some external content.

smotornyuk:
I still disagree. 
1. [This attribute has bad support in ie 11](https://caniuse.com/#search=download). So there still be complains, even though this browser almost never used
2. You are adding this attribute to the link, that has `Download`, `View`, `API Endpoint`, `Go to resource` labels depending on different conditions. But having a `download` attribute will force downloads no matter what user expected to see. This is a bit frustrating. 

I think you should take a look at @mutantsan 's suggestion in the original issue. It's not ideal, as in such a way we'll disable any caching, but it much more stable and explicit. If only we can use something, like the last modification date instead of access time - it would be even better.

PR #5177, opened 2020-01-27
Add code examples

smotornyuk:
A bit more examples and boilerplate in generated extensions.


contrib/cookiecutter/ckan_extension/{{cookiecutter.project}}/ckanext/{{cookiecutter.project_shortname}}/cli.py
```diff
@@ -0,0 +1,20 @@
+import click
+
+__all__ = ['{{cookiecutter.project_shortname}}']
```
amercader: Do we really need the `__all__` directives?

smotornyuk: No, I'm just adding it in advance. If `blanket`(#5169) will be merged into core, then we can just replace almost the whole body of `ExtPlugin` with few class-level decorators. But those `__all__` attributes have no functional meaning currently, so I see no problem in removing them

amercader (commented):
@smotornyuk I'm conflicted with this. One hand the code snippets you added are a great way to get people started, on the other hand they add a lot of boilerplate that users might end up not removing. And they might be not needed for experienced users bootstraping new extensions. Is there a way to ask during the `ckan generate extension` command whether people want the helper code snippets?
Something like:

```
ckan generate extension


Extension's name [must begin 'ckanext-']: ckanext-myext
...
List of keywords (separated by spaces) [CKAN]: 
Do you want to include code examples? [Y/n]:
```

And then include or not these files depending on the answer.
This might be over-complicating things, so just want to know your opinion on it

smotornyuk:
Updated. I've disabled the creation of the code examples by default. From my experience, people who want to use such a thing will read all the steps carefully and will enable these code examples. While people who know what they want, skip all the steps and start to write code right away. So, for this "fast" option, disabling code examples sounds more appropriate.

amercader:
Thanks @smotornyuk !

PR #5189, opened 2020-02-06
Use current package_type for urls

smotornyuk:
Working with custom package types(especially, while using ckanext-scheming), I've noticed this quite a big problem in our templates/redirects. No matter, what kind of dataset you are creating, even though you are likely using URL like `/CUSTOM-PACKAGE-TYPE/new`, on progressing to the second step you'll be redirected to resource creation page on `/DATASET/name/resource/new`.We are just using `dataset.edit`, `resource.new`, etc. routes for URL generation, thus `package_type` is ignored.

Initially, i thought about hacking into `url_for` - this just requires few lines of code, but in this case, additional query into DB required and usage of `package_type` becomes implicit. Instead, even if it requires tons of changes, I'm leaning to direct updates of all `url_for` usages.  Yep, it requires more work, but it:
* explicit
* if you expect that `dataset.new` leads to `/dataset/new` - you are still right
* if you've provided some custom fix for this behavior inside your extension - it is still correct


smotornyuk:
I've updated everything single place I was managed to locate by now. As for tests - I'm not sure if just visiting all possible pages and checking all the URLs related to dataset/resource views would do the thing... any idea on this account?

MandanaMoshref:
Hi,

I updated my base.html accordingly, however, I receive this error:
GenerationException: url_for can only return a string, got unicode instead: agrihub_resource.read?id=parcel_llkkss8320000261

is this due to python version? or what?
How this can be solved for ckan 2.8?

Best regards
Mani

smotornyuk:
Hi, @MandanaMoshref .
This change won't work for CKAN v2.8 as it has quite a different implementation of controllers. Actually, there is only one `package` controller in v2.8, whereas in v2.9 it was replaced by two blueprints - `dataset` and `resource`. 

Thus, this PR neither can be cherry-picked nor backported into v2.8 and the only thing you can do - manually re-define a bunch of templates and `package` routes.


Generally, before v2.9, any trivial changes to metadata schema are assuming that you are still working with **Dataset** - i.e. it should be safe to use all the same routes and templates. If you want some entity, that behaves in a completely different way(ckanext-scheming) - you should define custom controller with custom routes in addition to changes to metadata schema. This means, for such purpose you shouldn't rely only on ckanext-scheming.
In v2.9 this ideas were re-thinked and we decided that different package types should be more distinct - that's why i'm working on this PR. BUth this is a big changes to the ideas of "what dataset is" and shuch changes cannot be simply applied to prevous CKAN versions because it will only produce more bugs.

ckan/templates/package/snippets/resource_view.html
```diff
@@ -37,26 +37,26 @@
           <p id="data-view-error" class="collapse"></p>
           <p>
             <a href="{{ resource.url }}" class="btn btn-default btn-lg resource-url-analytics" target="_blank">
-		<i class="fa fa-lg fa-arrow-circle-o-down"></i>
-		{{ _('Download resource') }}
+		          <i class="fa fa-lg fa-arrow-circle-o-down"></i>
+		          {{ _('Download resource') }}
```
wardi: mixing tabs and spaces?

wardi:
Does this change let plugins override the views for specific dataset types?

ckan/templates/package/snippets/resource_view.html
smotornyuk: Nice catch. We are using `.editorconfig` file, that sets indentation to 2 spaces, so this file wasn't updated a long time if there still tabs here. There are few other places with mixed spaces and tabs - I won't change those files unless I need to edit them in the scope of current PR. Otherwise, diffs are unnecessary big, because GitHub cannot notice that the only thing that was changed is indentation:(

smotornyuk:
Previously it didn't, but when you mentioned it, I've realized that change required for supporting extendible blueprints is quite trivial, so I've included it in the last commit. Here's the link, in case I'll add some style fixes before you'll have time to review changes: https://github.com/ckan/ckan/pull/5189/commits/c832174dd8d204ab360c44e49816742e4f5212f9

wardi:
@smotornyuk this is great work. Good to merge?

smotornyuk:
It will be really great if this PR will be merged and available in v2.9(we have a few projects that are currently using custom CKAN fork because this change is essential for their workflow). So, if you have no objections and or any ideas as to some additional improvements that can be useful in conjunction with ckanext-scheming - let's merge it

PR #5193, opened 2020-02-07
Dont fail with existing user sessions from py2

smotornyuk:
I think @tino097 remembers this issue: 
```
  File "/usr/lib/ckan/py3/lib/python3.8/site-packages/repoze/who/plugins/auth_tkt.py", line 34, in <lambda>
    'unicode': lambda x: utf_8_decode(x)[0],
TypeError: a bytes-like object is required, not 'str'
```

Basically, on Python2 `repoze.who` was adding something like `user_id_type=unicode` to the `auth_tkt` cookie and using this flag in order to utf8_decode cookie during authentication. By on Python3 such an attempt will raise an error because of changes inside bytes-str-unicode relationships. This may become an issue, as every upgraded portal is likely will have some users with old-style-authentication cookie.

This PR just drops `unicode` decoder. Internally, `repoze.who` will skip decoding if no decoder is registered for the data type, so it should fix the issue.

### DIFF
--- a/ckan/lib/repoze_plugins/auth_tkt.py
+++ b/ckan/lib/repoze_plugins/auth_tkt.py
@@ -4,6 +4,8 @@ import logging
 import math
 import os
 
+import six
+
 from ckan.common import config
 from repoze.who.plugins import auth_tkt as repoze_auth_tkt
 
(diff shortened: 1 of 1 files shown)


ckan/lib/repoze_plugins/auth_tkt.py
```diff
@@ -4,16 +4,29 @@
...
 _bool = repoze_auth_tkt._bool
 
 log = logging.getLogger(__name__)
 
+# With `unicode` decoder users that were logged in on CKAN+py2 will
+# get 500 on CKAN+py3 error because of to utf8_decode attempts applied
+# to str, while bytes-like object expected.
+_ckan_userid_type_decoders = (
+    repoze_auth_tkt.AuthTktCookiePlugin.userid_type_decoders.copy()
+)
```
wardi: minor style thing: I would write this as:
```python
_ckan_userid_type_decoders = dict(
    repoze_auth_tkt.AuthTktCookiePlugin.userid_type_decoders
```
to be more explicit

tino097:
@smotornyuk this looks great, merging it

PR #5195, opened 2020-02-08
Werkzeug dev server improvements

amercader:
* Upgrade Werkzeug to 0.16.1. **Note** that there is a Werkzeug 1.0.0 release, but we can not use it because Flask-babel does not support it and it fails on startup
* Use watchdog as file reloader, as [it is faster than the default one](https://werkzeug.palletsprojects.com/en/0.16.x/serving/#reloader).
* Add Werkzeug logger section to ini file template so the log level can be tweaked
* Reload server when there are changes to the ini file. It's easy to add additional files to watch. Any other files that might be useful to watch?

### DIFF
--- a/ckan/cli/server.py
+++ b/ckan/cli/server.py
@@ -5,6 +5,8 @@ import logging
 import click
 from werkzeug.serving import run_simple
 
+from ckan.common import config
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


smotornyuk:
I like the idea of extending list of watched files, but it rather separate issue, so let's merge this one

PR #5202, opened 2020-02-11
Fix search_url_params generation bug

mutantsan:
Currently, there is an error in `params_nopage` generating from `request.args`. Since dictionary key is unique, `request.args.items()` returns only one (key, value) pair for tag, res_format etc. In this case we will loose all values with duplicated keys.

### Features:

- [ ] includes API changes


mutantsan:
Seems like this one is still a problem. How to reproduce: select some facet filters. E.g a couple of formats like in screen below. And go to the next page. 
image

Issue #5197, opened 2020-02-10
Refactor and simplify the app factory function


amercader:
Our `make_app` function expects a `global_conf` and `local_conf` parameters, which come from how paster intialized the applications (which was unnecessarily complicated in our case).

Essentially we just need to update our app factory entry point from 

```python
def make_app(conf, full_stack=True, static_files=True, **app_conf):
```

to just

```python
def make_app(conf):
```

And update `load_environment()` to only consider a single conf object. If done on 2.9, for compatibility with the Pylons stack we can keep a `conf['global_conf']` copy to pass to the places where Pylons still expects it, like [here](https://github.com/ckan/ckan/blob/da100c82ce4da90f8267879ef11789c65d0da8d8/ckan/config/environment.py#L106).

amercader:
Done in #5204

PR #5204, opened 2020-02-11
5197 - Refactor and simplify the app factory function

pdelboca:
### Proposed fixes:

Refactor to the `make_app` and `load_environment` as described in #5197 

### Features:

- [ ] includes API changes


pdelboca:
I think there is only one thing left: the previous implementation of this feature [didn't override](https://github.com/Pylons/pastedeploy/blob/9cbc584a86bd1ae425bd0c0c3c00617d7e59475d/paste/deploy/loadwsgi.py#L432) in the `local_conf` configs present on `global_conf`. When looking at config files, this means that configs that where in the `[DEFAULT]` section (`global_conf`) were not override if they appear again but in `[app:main]` section.

Example: 
- our `test.ini` file contains in the `[DEFAULT]` section `debug = true` (and this was former `global_conf`) and our `test-core.ini` file contains `debug = false` in the `[app:main]` section.
- When we were using pylons implementation to read the config, it had `debug=true` even when `test-core.ini` "overrode" it.

What should we do? This is not well documented and may be confusing.

amercader:
@pdelboca There is a bug in the implementation which was already present in #5135, regarding the handling of the `%(here)s` variable when running a script from a non-core folder (eg tests in an extension folder with `pytest --ckan-ini=test.ini ckanext/harvest`)

Basically the `here` variable is set only once, on the first parsed config file, but it should be re-evaluated on each inherit file.

Consider the following ini files:

```
# ckan/test-core.ini
[app:main]
key1=%(here)s/core
key2=%(here)s/core
key4=core
```

```
# ckanext-harvest/test.ini

[app:main]
use = config:../ckan/test-core-custom.ini
key1=%(here)s/harvest
key3=%(here)s/harvest
key4=harvest
```

If you inspect the configuration objects at the start of [`load_environment()`](https://github.com/ckan/ckan/blob/da100c82ce4da90f8267879ef11789c65d0da8d8/ckan/config/environment.py#L43) in CKAN 2.8 you will get:

```python
# global_conf
{
  '__file__': '/home/adria/dev/pyenvs/ckan-28/src/ckanext-harvest/test.ini',
   'here': '/home/adria/dev/pyenvs/ckan-28/src/ckan',
}

# app_conf
{
   'key1': '/home/adria/dev/pyenvs/ckan-28/src/ckanext-harvest/harvest',
   'key2': '/home/adria/dev/pyenvs/ckan-28/src/ckan/core',
   'key3': '/home/adria/dev/pyenvs/ckan-28/src/ckanext-harvest/harvest',
   'key4': 'harvest',
}
```

On master, you will get:

```python
# conf
{
 u'__file__': u'/home/adria/dev/pyenvs/ckan/src/ckanext-harvest/test.ini',
 u'here': u'/home/adria/dev/pyenvs/ckan/src/ckanext-harvest',
 u'key1': u'/home/adria/dev/pyenvs/ckan/src/ckanext-harvest/harvest',
 u'key2': u'/home/adria/dev/pyenvs/ckan/src/ckanext-harvest/core',
 u'key3': u'/home/adria/dev/pyenvs/ckan/src/ckanext-harvest/harvest',
 u'key4': u'harvest',
}
```

Anything relying on `here` used the value of the first ini file parsed (harvest), and the end value is the harvest one as well. This causes an exception later on when we try to initialize repoze because the path to the who.ini file can not be found:

```
INTERNALERROR>   File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/config/middleware/__init__.py", line 58, in make_app
INTERNALERROR>     flask_app = make_flask_stack(conf)
INTERNALERROR>   File "/home/adria/dev/pyenvs/ckan/src/ckan/ckan/config/middleware/flask_app.py", line 285, in make_flask_stack
INTERNALERROR>     who_parser.parse(open(conf['who.config_file']))
INTERNALERROR> IOError: [Errno 2] No such file or directory: u'/home/adria/dev/pyenvs/ckan/src/ckanext-harvest/who.ini'
```

The fix should be easy, you just need to re-evaluate the `here` value every time you parse a new file. Here's a quick and dirty patch that gave me the expected values on master:

```diff
diff --git a/ckan/cli/__init__.py b/ckan/cli/__init__.py
index 490021276..6053d4601 100644
--- a/ckan/cli/__init__.py
+++ b/ckan/cli/__init__.py
@@ -17,12 +17,6 @@ class CKANConfigLoader(object):
         self.section = u'app:main'
         self.read_config_files(filename)
 
-        defaults = {
-            u'here': os.path.dirname(os.path.abspath(filename)),
-            u'__file__': os.path.abspath(filename)
-        }
-        self._update_defaults(defaults)
-
     def read_config_files(self, filename):
         '''Read and parses a config file.
 
@@ -33,12 +27,22 @@ class CKANConfigLoader(object):
         configs in extension's test.ini files will override the configs of
         test-core.ini.
         '''
+        defaults = {
+            u'here': os.path.dirname(os.path.abspath(filename)),
+            u'__file__': os.path.abspath(filename)
+        }
+        self._update_defaults(defaults)
         self.parser.read(filename)
 
         schema, path = self.parser.get(self.section, u'use').split(u':')
         if schema == u'config':
             path = os.path.join(
                 os.path.dirname(os.path.abspath(filename)), path)
+
+            defaults = {
+                u'here': os.path.dirname(os.path.abspath(path)),
+            }
+            self._update_defaults(defaults)
             self.parser.read([path, filename])
 
     def _update_defaults(self, new_defaults):

```

pdelboca:
Good catch @amercader! 

However, doing a little bit more of testing if we update the `here` conf before the second `self.parse.read([path, filename])` it will interpolate all variables with the second value, so it will find the `who.ini` file but variables from the extension will be interpolated with the `here` of `core` because is re-parsing both files with the new `here`.

The logic will be more tricky that I thought. I updated the tests with your examples to reflect better and clearly the expected behavior and will be working on this.

pdelboca:
I made some changes on the way we create the config and refactored the test to get same results that we get in 2.8 with `appconfig` from `paste.deploy`.

I executed this tests, with this files in CKAN 2.8.3 and get the following results for the `config` object:
```
ipdb> pp conf
{'__file__': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/ckanext-extension/test-extension.ini.tpl',
 'debug': 'false',
 'here': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data',
 'key1': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/ckanext-extension/extension',
 'key2': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/core',
 'key3': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/ckanext-extension/extension',
 'key4': 'extension',
 'smtp_server': 'localhost'}

ipdb> pp conf.global_conf
{'__file__': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/ckanext-extension/test-extension.ini.tpl',
 'debug': 'false',
 'here': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data',
 'smtp_server': 'localhost'}

ipdb> pp conf.local_conf
{'key1': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/ckanext-extension/extension',
 'key2': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/core',
 'key3': u'/home/pdelboca/Repos/ckan/ckan/tests/lib/data/ckanext-extension/extension',
 'key4': 'extension'}
```

So what I implemented is the following:
- The first file being read takes precedence in the values, this is, extension's config files will override the configs in `test-core.ini`
- Values in `DEFAULT` section will always be overriden for the values in the second file if they are also present in the `DEFAULT` section. (DEFAULTS configs always are overridden)
- `__file__` config, even when it is a `DEFAULT` has always the value of the first config (will keep the value of the extension config)
- `here` config is updated so it will have the value of the second config (`test-core.ini`) and this will fix #5211 
- The special key `global_conf` in the `conf` dictionary is to keep compatibility with pylons stack. This should be removed.

ckan/tests/cli/test_cli.py
```diff
@@ -74,18 +74,105 @@ def test_command_from_extension_is_available_when_all_requirements_satisfied(cli
...
-def test_ckan_config_loader_parse_files():
+def test_ckan_config_loader_parse_file():
     """
-    CKANConfigLoader should parse both 'test.ini.tpl' and 'test-core.ini.tpl'
-    files since test.ini.ptl has a use = config:test-core.ini.tpl config.
+    CKANConfigLoader should parse and interpolate variables in
+    test-core.ini.tpl file both in DEFAULT and app:main section.
     """
-    filename = os.path.join(os.path.dirname(__file__), u'data/test.ini.tpl')
+    data_dir = os.path.dirname(__file__) + u'/data/'
```
amercader: `data` is perhaps a confusing word here (and also ignored by our .gitignore file). What about `config` or `templates`?

pdelboca:
@amercader I did the change in the folder name, however there is this random test failing in Python 2.7. I think it is random because I saw it failing before and it pass in locals environment.

amercader:
Yeah, we need to look at that test failure, but merging this as it's important and the failure is unrelated

PR #4601, opened 2018-12-21
3994 fix background job timeout

scottdillon:
Fixes #3994

### Proposed fixes:
Pass a timeout keyword from the `ckan.lib.jobs.equeue` method through to the rq `Job` object. 

Includes tests for cases where timeout is:
- None
- -1
- 0
- 3600

### Features:

- [x] includes tests covering changes
- [x] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #5208, opened 2020-02-12
Allow passing arguments to the RQ enqueue_call function

amercader:
Right now you can only pass the `timeout` argument to the RQ `enqueue_call` invocation that we use on `toolkit.enqueue_job()` (added in https://github.com/ckan/ckan/pull/4601), but there are other useful arguments that can be passed, like `depends_on`, `ttl` etc

Rather than explicitly allow just one of them, it's better to just pass through all of them.

This is backwards compatible, so I proposed we backport it to at least 2.8


amercader:
Remove the timeout kw

PR #5219, opened 2020-02-17
Correct auth call in datastore_run_triggers

smotornyuk:
`ckanext.datastore.logic.auth.datastore_trigger_each_row` was [renamed](https://github.com/ckan/ckan/blame/master/ckanext/datastore/plugin.py#L117) to `datastore_run_triggers`, but corresponding action [uses](https://github.com/ckan/ckan/blob/master/ckanext/datastore/logic/action.py#L196) obsolete name


wardi:
mea culpa

PR #5238, opened 2020-02-25
SXSAGXZFZ-453: Fix format and memtype fields

styhar:


smotornyuk:
So, what's the background for this change? Not sure, how popular those `mimetype` and `mimetype_inner` fields are, but this if a change to the existing functionality, thus it should be inspired by some kind of bug or improve something.

PR #5241, opened 2020-02-26
Configure logging using ini file

smotornyuk:
Without `paste.appconfig` we don't have automatic applying of logger configuration anymore, so let's do this manually. The only thing, I'm not sure about is a warning from python docs:
> Warning: The fileConfig() function takes a default parameter, disable_existing_loggers, which defaults to True for reasons of backward compatibility. This may or may not be what you want since it will cause any non-root loggers existing before the fileConfig() call to be disabled unless they (or an ancestor) are explicitly named in the configuration. Please refer to the reference documentation for more information, and specify False for this parameter if you wish.

So, it will work just as before - if there is some logger that was created before config file loaded - it will be disabled. But is there any reason why we don't want to disable such loggers? If so, it's a good moment to change the default behavior.


smotornyuk:
There were tests that are checking logs inside `ckan/tests/lib/test_cli.py` - they become broken because of changes in `ckan.cli.load_config` and were rewritten to new style as a result.

PR #5255, opened 2020-03-04
Samesite auth cookie

amercader:
Allows setting the SameSite attribute on the auth cookie, and aligning with what browsers are moving towards to, default it to `Lax`.

Refactored the code to not rely on string manipulation for modifying the cookie attributes but rather the standard library module.

This causes some changes in the case of the attributes and their order which means updating the tests, but I think it's worth the extra work.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies

### DIFF
--- a/ckan/lib/repoze_plugins/auth_tkt.py
+++ b/ckan/lib/repoze_plugins/auth_tkt.py
@@ -6,6 +6,11 @@ import os
 
 import six
 
+try:
+    from http.cookies import SimpleCookie
+except ImportError:
+    from Cookie import SimpleCookie
+
 from ckan.common import config
 from repoze.who.plugins import auth_tkt as repoze_auth_tkt
 
@@ -27,9 +32,10 @@ class CkanAuthTktCookiePlugin(repoze_auth_tkt.AuthTktCookiePlugin):
 
     userid_type_decoders = _ckan_userid_type_decoders
 
-    def __init__(self, httponly, *args, **kwargs):
+    def __init__(self, httponly, samesite, *args, **kwargs):
         super(CkanAuthTktCookiePlugin, self).__init__(*args, **kwargs)
         self.httponly = httponly
+        self.samesite = samesite
 
     def _get_cookies(self, *args, **kwargs):
         u'''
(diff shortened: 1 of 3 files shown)


ckan/lib/repoze_plugins/auth_tkt.py
```diff
@@ -69,9 +92,15 @@ def make_plugin(secret=None,
     if timeout is not None and reissue_time is None:
         reissue_time = int(math.ceil(int(timeout) * 0.1))
     # Set httponly based on config value. Default is True
-    httponly = config.get(u'who.httponly', True)
+    httponly = _bool(config.get(u'who.httponly', True))
     # Set secure based on config value. Default is False
-    secure = config.get(u'who.secure', False)
+    secure = _bool(config.get(u'who.secure', False))
+    # Set samesite based on config value. Default is lax
```
ThrawnCA: This should be documented as a potentially breaking change (admittedly it *shouldn't* break sensible setups).

ckan/lib/repoze_plugins/auth_tkt.py
```diff
@@ -69,9 +92,15 @@ def make_plugin(secret=None,
     if timeout is not None and reissue_time is None:
         reissue_time = int(math.ceil(int(timeout) * 0.1))
     # Set httponly based on config value. Default is True
-    httponly = config.get(u'who.httponly', True)
+    httponly = _bool(config.get(u'who.httponly', True))
```
ThrawnCA: This should probably be deferred until after we check the attributes that might cause an early termination.

ckan/lib/repoze_plugins/auth_tkt.py
Zharktas: Changelogs are updated during the release process, so it would be documented there.

ckan/lib/repoze_plugins/auth_tkt.py
Zharktas: Boolean conversions were done in CkanAuthTktCookiePlugin call and those variables weren't used in anywhere else, so it makes sense to convert them right away.

Zharktas:
Note about the change and how to disable it should be added to the release notes.

amercader:
@Zharktas did you cherry-pick this into the dev-v2.* branches already? If not add the "Backport pending" label otherwise we assume the backport has been done.
This will require some manual work because on 2.6 to 2.8 the auth_tkt.py file is on `ckan/lib`, not in `ckan/lib/repoze_plugins`.

Zharktas:
@amercader ah right, missed the penging label, added now.

PR #5263, opened 2020-03-06
Datapusher cli migration

### LINKED ISSUE #5126: Migrate Datapusher and DataStore CLI commands to Click
Now that #5112 is merged, we can refactor the CLI commands for the datastore and datapusher extensions (`ckanext/datastore/cli.py` and `ckanext/datapusher/cli.py`) to use the new `IClick` interface rather than the old paster based one.

https://github.com/ckan/ckan/pull/5121 has some detail on how to implement this interface

davidread:
Fixes #5126 

### Proposed fixes:

These CLI commands were using the old `paster_click_group`. This PR converts them to use IClick. They are run directly:

`dataset set-permissions`

(i.e. not part of the new `ckan` command any more)

Also deletes the original paster versions of these CLI commands: ckan/cli/datastore.py & ckanext/datapusher/cli.py

### Features:

- [ ] includes API changes


amercader:
@davidread just checking status on this as part of the review of stuff we need for 2.9
Do you need someone else to pick it up?

Quick question, why have separate commands instead of `ckan datastore` and `ckan datapusher`? That's the intended behaviour if I'm not mistaken

amercader:
Sorry, just saw this https://github.com/ckan/ckan/pull/5128#issuecomment-571582830, ignore my last question

wardi:
I kind of prefer `ckan datastore`, `ckan datapusher`, `ckan <extension-name>` ... as well, what led to wanting to add a separate command for datastore at that meeting?

amercader:
I don't like having a separate command for datastore either.

If I understood correctly the issue described in #5128 , the problem is that with the new CLI for a plugin command to be able to be run, the plugin needs to be loaded (as opposed to the old Paster commands that just needed to be registered via the `setup.py` entrypoints). When we try to run the `datastore set-permissions` command in Circle CI during the setup phase, it fails because the datastore plugin is not loaded (because it's not added by default to `test-core.ini`)

Let's discuss on later on the call, but perhaps we can just load the plugin temporarily to run this command with some hack. Or maybe there is another option, I'm not sure what @smotornyuk means here:

> If you update this place - you can register global datastore command and it still will be available under ckan group

wardi:
I like the idea of registering CLI commands with an entrypoint again. I want to be able to hook the ckanapi tool's cli in to the ckan command (so that `ckan api <subcommand>` behaves like `ckanapi -c <ini-file> <subcommand>`) but I don't want people to have to add a do-nothing plugin to their ini file to use it.

amercader:
Summary of discussion:

* We want to keep the `ckan datastore set-permissions` syntax
* To support cases like this where commands *don't* need the plugin to be active we should also support entrypoint registration of commands (like the `datastore` one or external libraries like the `ckanapi` example that @wardi mentioned.
* This should be an advanced option for registering commands, the recommended and prominent one in the docs should still be `IClick`
* @smotornyuk can you help me spec this? Reading entrypoints is easy ([here](https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/processors.py#L78) is an example from ckanext-dcat). I guess they should be read at the same place `IClick` [is called](https://github.com/ckan/ckan/blob/3f2eefb2ef8fbbf8488c0b44d411f4a85a3b609c/ckan/cli/cli.py#L63). But what should the entrypoint point to? The command function or group?

smotornyuk:
Click works in such a way so there will be no difference, whether you are using command or group as entry point(basically, groups itself are commands). I don't think that there is any reason to stick just to one option and block another one.

I was just working on another fix for CLI so I've added an example of code, that parses entry points - https://github.com/ckan/ckan/pull/5329. If this option is acceptable, @davidread you can just use this example as a base and adapt it for your need - I'll remove it from my PR then

davidread:
I'm bowing out of this PR, I'm afraid. I just spotted the commits were done by @JGulic but weren't contributed back to CKAN, so I thought I'd create a PR. But I don't have the time for this one, I'm afraid.

smotornyuk:
This PR is covered by #5329, so i'm closing it

PR #5296, opened 2020-03-20
Correct faulty example and add another in a Jinja macro

mikkoi:
Signed-off-by: Mikko Johannes Koivunalho <mikko.koivunalho@iki.fi>

Fixes #

### Proposed fixes:

Correct faulty example and add another in a Jinja macro

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/macros/form.html
+++ b/ckan/templates/macros/form.html
@@ -79,7 +79,17 @@ options     - A list/tuple of fields to be used as <options>.
     Examples:
 
     {% import 'macros/form.html' as form %}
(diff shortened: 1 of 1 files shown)


Zharktas:
@mikkoi the builds fail as the there is now comment within a comment block and therefore it causes syntax issues, could you fix that first of all?

mikkoi:
Is there any way to have the comment in a sensible way (separate the comment from code), or just write it plainly?

Zharktas:
Plain text is fine in this case as it is part of multiline comment.

PR #5303, opened 2020-03-26
DevServer: threaded mode and extra files

smotornyuk:
* Allow running dev server in threaded mode(`-t/--threaded`) in order to handle each request in separate thread.
* Specify extra-watched files (`-e/--extra-files`). Something like `ckan -e xxx.json -e 'sub-path/**/*.json'` works as well


wardi:
in the dev call we discussed adding a multi-process option for the dev server as well. Werkzeug supports multi-process with a fixed number of processes.

We should also have configuration options for these as well, e.g.
```ini
ckan.devserver.threaded = false
ckan.devserver.multiprocess = 8
ckan.devserver.watch_patterns = mytheme/**/*.yaml mytheme/**/*.json
```

smotornyuk:
Updated as described above + documentation for new config options

PR #5304, opened 2020-03-26
Fixes problem with logs interfering with "ckan datastore set-permissions"

davidread:
When setting the datastore permissions, before this PR, the logging gets in the way:
```
$ ckan datastore set-permissions | sudo -u postgres psql --set ON_ERROR_STOP=1
You are now connected to database "datastore_test" as user "postgres".
ERROR:  syntax error at or near "2020"
LINE 1: 2020-03-26 17:18:05,311 CRITI [ckan.lib.uploader] Please specify a ckan.storage_path in your config for your uploads
```
You can solve this by adding in `2>/dev/null` however this is a complication for the user.

So this PR prevents logging happening when you are running this CLI command.

I thought about letting load_config() run and then removing the loggers it sets up. However by this time it has already done the log.critical message seen above. So I needed to prevent the loggers being setup in the first place with the `setup_logging` option.

--

Whilst in this area, the `2>/dev/null` doesn't work when you specify `-c test-core.ini` to setup for tests:

    $ ckan -c test-core.ini datastore set-permissions | sudo -u postgres psql --set ON_ERROR_STOP=1

That is because test-core.ini outputs logs to stdout, not stderr. So in this PR I've fixed that anomaly too (in test-core.ini).

I don't think any of this was an issue prior to CKAN 2.9

### Features:

- [ ] includes API changes


ckan/templates/macros/form.html
```diff
@@ -79,7 +79,17 @@
...
+
+    Or only with values if they are the same as text:
+    {{ form.select('year', label=_('Year'), options=[{'value':2010},{'value': 2011}], selected=2011, error=errors.year) }}
+
+    Complete example:
+    {{ form.select('the_data_type', id='the_data_type', label=_('Data Type'), options=[
+        {'value': '0', 'text': _('[Choose Data Type]')}
+        , {'value': '1', 'text': _('Private data')}
+        , {'value': '2', 'text': _('Not private data')}
+    ], selected=data.the_data_type if data.the_data_type else '0', is_required=true) }}
```
davidread: Ignore this file in this PR. I did a rebase and included this commit by mistake. The commit id is unchanged, but git thinks I coauthored it now... This [commit](https://github.com/ckan/ckan/pull/5304/commits/8fee4e8fbf2b81ad6c4c48470c823abf03f1edf6) is already merged to master 🤷‍♂

wardi:
I'm not sure I understand, are you copying and pasting the results from a terminal window into psql? Otherwise I can't see an issue with having some stderr output mixed with the sql on stdout.

davidread:
@wardi you're quite right. I have revamped the PR description - please take another look.

ckan/cli/cli.py
```diff
@@ -39,22 +39,29 @@
...
-        self.config = ckan_cli.load_config(conf)
+        self.config = ckan_cli.load_config(conf, setup_logging=setup_logging)
         self.app = make_app(self.config)
 
 
-def _init_ckan_config(ctx, param, value):
+def _init_ckan_config(ctx, param, config_filepath):
+    setup_logging = True
+    if len(sys.argv) > 1 and \
+            sys.argv[-2:] == [u'datastore', u'set-permissions']:
```
wardi, line 52: if someone runs the command like `ckan datastore set-permissions -c my-config.ini` then this won't work.

test-core.ini
```diff
@@ -140,7 +140,7 @@ level = WARNING
 
 [handler_console]
 class = StreamHandler
-args = (sys.stdout,)
+args = (sys.stderr,)
```
wardi, line 143: this part seems obviously correct to me, if we make this 1-line change and update the docs to add `2> /dev/null` to the command would that be good enough?

ckan/cli/cli.py
davidread: I'm not sure that is valid. Doesn't the param has to go before the subcommand?

test-core.ini
davidread: Good enough, sure.

ckan/cli/cli.py
```diff
@@ -39,22 +39,29 @@
 
 class CkanCommand(object):
 
-    def __init__(self, conf=None):
+    def __init__(self, conf=None, setup_logging=True):
         # Don't import `load_config` by itself, rather call it using
         # module so that it can be patched during tests
-        self.config = ckan_cli.load_config(conf)
+        self.config = ckan_cli.load_config(conf, setup_logging=setup_logging)
```
wardi, line 45:
```suggestion
        stdout = sys.stdout
        sys.stdout = os.devnull  # prevent logging from affecting cli stdout
        try:
            self.config = ckan_cli.load_config(conf)
        finally:
            sys.stdout = stdout
```
What do you think of this?

No need to check which cli command was issued: we should never be logging to stdout when running the cli. This way correctly configured logging can still work and we don't add a potentially fragile check in the `_init_ckan_config` function.

ckan/cli/cli.py
wardi: you're right, that doesn't work with the new click cli

PR #5322, opened 2020-04-02
#5320 Don't encode error detail before aborting.

### LINKED ISSUE #5320: Rendering error template fails for non-ascii error messages
2.8.3

When an error page is returned, it should include the text of the error, e.g. "Unauthorized to edit a user."

When the error text includes non-ascii characters, an "Internal server error" page with response code 500 is returned instead.

As a non-logged-in user, with language set to German, try to edit a user page: https://demo.ckan.org/de/user/edit/covenant. The error text that should be returned is, "Keine Berechtigung, Benutzer zu ändern," and it fails because of the ä. In English it works: https://demo.ckan.org/en/user/edit/covenant.

In `ckan.lib.base.abort`, the detail is utf8-encoded with a comment saying that this is to prevent an exception from WebOb 0.9.7.1, but this encoding now seems to create an exception instead. (It works if the detail is not encoded.) Is there anything speaking against just removing this line?

bellisk:
This line was added because WebOb < 1.0 caused an exception when receiving unicode. It now causes an exception when it doesn't receive unicode.

Fixes #5320

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -64,10 +64,6 @@ def abort(status_code=None, detail='', headers=None, comment=None):
     if is_flask_request():
         flask_abort(status_code, detail)
 
-    # #1267 Convert detail to plain text, since WebOb 0.9.7.1 (which comes
-    # with Lucid) causes an exception when unicode is received.
-    detail = detail.encode('utf8')
-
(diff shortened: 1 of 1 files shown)


Zharktas:
Fixed already in #4829

PR #5329, opened 2020-04-05
Cli extensions ++

smotornyuk:
Fix(first commit):
After recent changes to config loader, CLI commands from enabled plugin not shown in CLI help message. 

Feature(second commit):
As suggested in #5263, in addition to `IClick`, CLI commands can be registered via entry points in setup.py. Either commands or command groups can be registered as below:
```
    [ckan.click_command]
    xxx=ckanext.xxx.cli:xxx_command_or_group
```


wardi:
At today's meeting we discussed resolving conflicts between IClick and entrypoints for the same subcommand names:

1. IClick always wins (overrides) with the first plugin loaded defining the subcommand, the same as other plugin features.
2. conflicting entrypoint subcommands (when no matching IClick subcommand is present) should cause an error when running with the ckan cli indicating which packages have installed the same subcommand.

smotornyuk:
Updated. When there are multiple entry-points with the same name registered in different extensions, the user will get an error like the following:
image

smotornyuk:
`datastore` and `datapusher` added as entry points. With this change, #5263 can be closed.

PR #5337, opened 2020-04-09
Group type in UI

### LINKED ISSUE #5331: Calling dataset groups "groups" (rather than e.g. "categories") is unclear/confusing (terminology)
The CKAN interface offers to create something called "groups". This feature appears undocumented, but for many administrators, there is an expectation that groups are for grouping users (like Active Directory groups).

CKAN groups can in fact group datasets, and not users.

This is not a behavioral bug.

---

CKAN allows to organize datasets beyond the organization which publishes them. For example, an "Environmental investment" dataset published by the finance ministry can be added to an "Ecology" group, while an "Industrial forecasts" published by the same ministry can be added to an "Industry" group. The best description of groups I could find is in the answer to [this question](https://stackoverflow.com/questions/20963965/whats-the-difference-between-organizations-groups-in-ckan).
<br>

The province I work for created a CKAN instance (Données Québec) which has customized CKAN so that groups are called "catégories" (French for "categories"). In parallel, our province's biggest city has created [a CKAN instance](http://donnees.ville.montreal.qc.ca/group) which has customized CKAN so that groups are called "thématiques" (which I would translate to English as "themes" ([Wiktionary defines  a _thématique_](https://en.wiktionary.org/wiki/th%C3%A9matique) as a "structured group of themes").

Moreover, in just a few months of lurking on the ckan Gitter channel, I have seen 2 people state their intention to rename groups:

- [Abed khooli intends to rename groups to "categories" or "topics".](https://gitter.im/ckan/chat?at=5e841935c090ce1fde35d1ab)
- [olivierdalang intends to rename groups to "projects".](https://gitter.im/ckan/chat?at=5e84b7c155d0d07c2d2e69ab)

I consider the above clear signs that revising the term used to designate groups would be beneficial. A community poll would be required to identify the most consensual alternative, and to what degree users prefer it over "group". The above examples suggest alternatives:

1. "categories"
2. "themes"
3. "topics"
4. "projects"

In my humble opinion, the first 3 alternatives are better, while I would rather keep "groups" than changing for "projects". Another option may be to rename from just "Groups" to "Dataset groups".

There are 2 issues with the current term:

1. The term "group" is very vague. I experimented asking one adult to give me examples of groups, without giving any context, and her answers were all groups of humans. [Wiktionary lists 15 senses for the noun "group"](https://en.wiktionary.org/wiki/group).
2. CKAN's sense for "group" is potentially misleading. One sense of "group" is very likely in CKAN's context: a group of users. The above Wiktionary entry lists it in the computing context. The first 2 sections in [Wikipedia's disambiguation page on "group"](https://en.wikipedia.org/wiki/Group) are about groups of humans, then groups of animals. Even [Wikipedia's article about groups in the computing context](https://en.wikipedia.org/wiki/Group_(computing)) starts with "In computing, the term group generally refers to a grouping of users."

It is non-trivial to find an equally short term which would have no vagueness at all, but I believe "categories" would be much better. I was involved for years in a CMS project which has a similar concept, called "categories", and even though I was involved orders of magnitude more in that project and the product has many more instances than CKAN, I do not remember anyone ever customizing the term, nor even discussing that. The categories feature in that project is optional, but it is widely used.

<br>
I am afraid it is quite late in CKAN's development for such a change, but I see there was already a similar rename ("packages" to "datasets"), and I think this one might be equally important. I am merely proposing to rename in the interface (as was done for datasets); renaming in code/databases would surely be excessively costly.

Abed khooli suggested providing a configuration option for this. If there are several deployments who want to stick to "groups", it might be possible to give the choice between 2 terms, but that would be costly.

Meanwhile, I am planning to write a CKAN 2.8 extension to ease customization. I would write a .pot file and a French .po overriding the standard translations to use "catégories" instead.

smotornyuk:
Potentially fixes #5331

Similar to #5189, but for groups/organizations. Generate labels and URLs using `group_type` of the current group.

Heavily relies on `ckan.lib.helpers:default_group_type` (which uses `ckan.default.group_type` and `ckan.defaul.organization_type` config options). Using two corresponding config options, one can set custoum group/organization type as site-wide defaults. i.e, if you have custom `group` with type `theme` and custom `organization` with type `publisher`, by adding following lines into your config file 
```ini
ckan.default.group_type = theme
ckan.default.organization_type = publisher
``` 
you'll get following result without additional code modifications:

image

Right now I'm using `str.title` method when I need a human-readable representation of group type(exactly the same thing is done for packages and I prefer consistency) but in the future, maybe on the following iteration, it makes sense to introduce helper/interface/`IDataset/Groupform` method as an extension point for this behavior.

PS in the first version all those labels can be tweaked via custom translations, so using `str.title` it's just a fallback.


smotornyuk:
As I said, this PR is pretty similar to the one, related to package types. What is really magical here is [_extra_translations template](https://github.com/ckan/ckan/pull/5337/files#diff-88de8f8c164db6189ebd47555aa871f3). 
I want to use group type in different kinds of labels(**create group**, **list groups**, etc.), but as long as you are using non-static strings(i.e, static - `"create group"`, non-static = `"create " + group.type`), it won't be extracted into .pot file for translations. Using `_("create {type").format(type=group.type)` won't work either. Such placeholders work for numbers or non-translatable terms, but due to grammar in different languages, it may be quite tricky to correctly insert random noun into a sentence. Thus, I've extracted most expected strings into a separate file and they will be extracted for translations. 
**Group** and **Organization** versions are expected. But what are those **Category**, **Theme**, **Project**, **Topic**? I've just picked terms that are likely to be used instead of **Groups** from the issue, referenced in this PR. Probably, it would be better to implement custom group type and translate it inside custom extension, but translations of few additional lines to all supported languages as part of the CKAN project are much more effective, that translation of custom extension. I don't think, that there is some extension, that supports all the languages, that are supported by CKAN core.
Another problem, which is not so obvious - plural form. For example, in breadcrumbs, we are using something like `Home / Groups / Xxx`. Note, **Groups** fragment is in plural form. It's merely simple for the organization or group types - just add `s` suffix. But it won't work for irregular nouns (**child**/**children**) and those y-ending nouns (**Category**). Right now I'm just adding `s` to any word. One can convert **Categorys** into **Categories** using custom translations. In CKAN we don't have such edge cases, but it may be quite problematic for extensions. Mabe anyone knows some small library for converting English nouns into plural form?

ckan/controllers/group.py
```diff
@@ -314,8 +314,14 @@ def pager_url(q=None, page=None):
 
             facets = OrderedDict()
 
-            default_facet_titles = {'organization': _('Organizations'),
-                                    'groups': _('Groups'),
+            org_type = h.default_group_type(u'organization') + u's'
+            org_label = h.humanize_entity_type(org_type).capitalize()
+
+            group_type = h.default_group_type(u'group') + u's'
+            group_label = h.humanize_entity_type(group_type).capitalize()
```
wardi, line 321: Not `.titile()` as you used for package_type?  Even better, maybe make it part of `humanize_entity_type`?

ckan/controllers/group.py
```diff
@@ -314,8 +314,14 @@ def pager_url(q=None, page=None):
 
             facets = OrderedDict()
 
-            default_facet_titles = {'organization': _('Organizations'),
-                                    'groups': _('Groups'),
+            org_type = h.default_group_type(u'organization') + u's'
+            org_label = h.humanize_entity_type(org_type).capitalize()
+
+            group_type = h.default_group_type(u'group') + u's'
```
wardi, line 320: Let's leave pluralization in helpers that can be overridden by sites (like humanize_entity_type) because it's not the same for all words/languages

ckan/controllers/group.py
```diff
@@ -314,8 +314,14 @@ def pager_url(q=None, page=None):
...
+            group_label = h.humanize_entity_type(group_type).capitalize()
+
+            default_facet_titles = {'organization': _(org_label),
+                                    'groups': _(group_label),
```
wardi, line 324: Here's my suggestion for the humanize_entity_type interface. This will let us cover all the common needs for customizing pluralizations, capitalizations etc in a plugin. 

```suggestion
            org_label = h.humanize_entity_type(
                entity_type=u'organization',
                object_type=h.default_group_type(u'organization'),
                purpose=u'facet label') or _('Organizations')
            group_label = h.humanize_entity_type(
                entity_type=u'group',
                object_type=h.default_group_type(u'group'),
                purpose=u'facet label') or _('Groups')

            default_facet_titles = {'organization': org_label,
                                    'groups': group_label,
```

wardi: I left the gettext calls in above to handle the helper returning nothing and to support `extract_messages` use

ckan/lib/helpers.py
```diff
@@ -896,7 +897,7 @@ def link_to(label, url, **attrs):
 
 
 @core_helper
-def file(name, value=None, id=None, **attrs):
+def file(name, value='', id=None, **attrs):
```
wardi, line 900: choose a different name so we don't shadow the builtin `file` object

ckan/templates/_extra_translations.html
```diff
@@ -0,0 +1,148 @@
...
+  {{ _('There are no topics associated with this dataset') }}
+  {{ _('There are no projects associated with this dataset') }}
+
+  {{ _('Are you sure you want to delete this Organization? Note*: Deleting cannot be performed while public or private datasets belong to this organization.') }}
+  {{ _('Are you sure you want to delete this Group? Note*: Deleting cannot be performed while public or private datasets belong to this group.') }}
+  {{ _('Are you sure you want to delete this Category? Note*: Deleting cannot be performed while public or private datasets belong to this category.') }}
+  {{ _('Are you sure you want to delete this Theme? Note*: Deleting cannot be performed while public or private datasets belong to this theme.') }}
+  {{ _('Are you sure you want to delete this Topic? Note*: Deleting cannot be performed while public or private datasets belong to this topic.') }}
+  {{ _('Are you sure you want to delete this Project? Note*: Deleting cannot be performed while public or private datasets belong to this project.') }}
+</pre>
```
wardi, line 148: An artificial translation file like this will go out of sync with the actual messages really easily

ckan/templates/group/base_form_page.html
```diff
@@ -1,14 +1,16 @@
...
 
+{% set group_type_label = h.humanize_entity_type(group_type).capitalize() %}
+
 {% block page_primary_action %}{% endblock %}
 
 {% block breadcrumb_content %}
-  <li>{{ h.nav_link(_('Groups'), named_route=group_type+'.index') }}</li>
-  <li class="active">{% block breadcrumb_link %}{{ h.nav_link(_('Add a Group'), named_route=group_type+'.new') }}{% endblock %}</li>
+  <li>{{ h.nav_link(_(group_type_label ~ 's'), named_route=group_type+'.index') }}</li>
+  <li class="active">{% block breadcrumb_link %}{{ h.nav_link(_('Add a ' ~ group_type_label), named_route=group_type+'.new') }}{% endblock %}</li>
```
wardi, line 9:
```suggestion
  <li>{{ h.nav_link(h.humanize_entity_type(
          'group', group_type, 'breadcrumb') or _('Groups'), named_route=group_type+'.index') }}</li>
  <li class="active">{% block breadcrumb_link %}{{ h.nav_link(h.humanize_entity_type(
          'group', group_type, 'add link') or _('Add a Group'), named_route=group_type+'.new') }}{% endblock %}</li>
```

ckan/templates/group/base_form_page.html
```diff
@@ -1,14 +1,16 @@
...
 {% endblock %}
 
 {% block primary_content_inner %}
-  <h1 class="{% block page_heading_class %}page-heading{% endblock %}">{% block page_heading %}{{ _('Group Form') }}{% endblock %}</h1>
+  <h1 class="{% block page_heading_class %}page-heading{% endblock %}">{% block page_heading %}{{ _(group_type_label ~ ' Form') }}{% endblock %}</h1>
```
wardi, line 13:
```suggestion
  <h1 class="{% block page_heading_class %}page-heading{% endblock %}">{% block page_heading %}{{
      h.humanize_entity_type('group', group_type_label, 'form label') or _('Group Form')
    }}{% endblock %}</h1>
```

wardi:
@smotornyuk what do you think of my suggestions? I think this would let us have customizable labels for all entity types by overriding a single helper function and it lets us keep the core messages in the right places for `extract_messages` use.

ckan/lib/helpers.py
```diff
@@ -1126,8 +1127,25 @@ def _make_menu_item(menu_item, title, **kw):
...
+    human-readable form.
+
+    Example::
+
+        >>> humanize_entity_type('group')
+        'group'
+        >>> humanize_entity_type('custom_group')
+        'custom group'
+    """
+    return type_.replace('_', ' ')
```
wardi, line 1148:
```suggestion
def humanize_entity_type(entity_type, object_type, purpose):
    """Convert machine-readable representation of package/group type into
    human-readable form.

    Example::

        >>> humanize_entity_type('group', 'custom group', 'add link')
        'Add Custom Group'
        >>> humanize_entity_type('group', 'custom_group', 'breadcrumb')
        'Custom Groups'
    """
    if entity_type == object_type:
        return  # use the default text included in template
    
    name = object_type.replace('_', ' ').capitalize()
    if purpose == 'add link':
        return _('Add {object_name}').format(object_name = name)
    # ... other purposes with custom text
    return name + 's'
```

wardi: I think it's better to *not* have an exhaustive list of translations for each object_type + purpose. If a site needs to override specific translations they can do that by overriding this helper (and they get their translation text in code that can be picked up by `extract_messages` too)

ckan/lib/helpers.py
smotornyuk: I think, that `file` was completely removed in Python 3 in favor of `open`. And this helper doesn't seem to be very popular - I don't see it in the core, neither is saw it inside any of extensions. So i prefer to remove it completely instead of renaming

wardi: even better

ckan/lib/helpers.py
smotornyuk: Yep, I like this approach.

wardi:
> @smotornyuk what do you think of my suggestions? I think this would let us have customizable labels for all entity types by overriding a single helper function and it lets us keep the core messages in the right places for `extract_messages` use.

to be more specific: with a helper like this ckanext-scheming could allow overriding all these text snippets from the schema yaml file, in multiple languages, right with the rest of the schema definition.

smotornyuk:
+1 to this approach.  But If it implemented as a plain helper, then you'll be able to redefine it only once. I think you'll get an exception if two extensions try to redefine the same helpers. At least it was true some time ago.

so, instead, maybe we can use a bit more complex option, like:
```python
# helpers
def humanize_entity_type(...):
    translator = create_entity_humanizer()
    ...

_humanizer_hierarchy = []
_EntityHumanizer = None

def create_entity_humanizer():
    global _EntityHumanizer
    if not _EntityHumanizer:
        _EntityHumanizer = type('EntityHumanizer', tuple(reversed(_humanizer_hierarchy)), {})
    return _EntityHumanizer()


# extension/plugin.py
class Plugin(...):
    def update_config(...):
        tk.add_humanizer(CustomHumanizerClass)
```
In that way, every extension can provide a custom implementation of the humanization class. All the custom implementations will be used to create the final class, that inherits all of them. As alternative, a new interface can be used, in order to avoid global variables. But without global variable, we will iterate through the implementation of the interface for each humanization call - i'm not sure how big can be performance overhead in this case

smotornyuk:
Updated.

wardi:
@smotornyuk ah, we have `@chained_action` and `@chained_auth_function` but no `@chained_helper`. Let's add `@chained_helper` instead of inventing a new way of doing this.

ckan/lib/humanize.py
```diff
@@ -0,0 +1,141 @@
...
+        term = self._prepare_term(term)
+        if handler:
+            return handler(term, *args, **kwargs)
+        log.debug(u'%s has no handlers for "%s"', self.__class__, purpose)
+        return self._default(term, **kwargs)
+
+
+class EntityTypeHumanizer(BaseHumanizer):
+    def add_link(self, term, entity_type):
+        return _(u"Add {}".format(term))
```
wardi, line 83: These need to be written as `_(u"Add {}").format(term)` so that extract_messages can pick them up.

wardi:
I like having a single function with if statements instead of an interface class with a list of methods because it's less work to override just a couple humanized forms (with @chained_helper`). It's also easier for ckanext-scheming to do a quick lookup in the schema yaml based on the `purpose` passed that will work with any future purpose we add to the helper.

ckan/lib/humanize.py
smotornyuk: This option works only with names\numbers and won't work with terms. 
Let's use `Add {type}` as example. In translation to japanes it will turn into ` {type}を追加`, which becomes `Camel photosを追加` after interpolation with `camel_photos` type.

Instead, let's extract just default messages from templates and leave those humanized labels. user will clearly see `Add Camel photos` in japanese locale and will manually add into .po file following lines:
```
msgid "Add Camel photos"
msgstr " 駱駝の写真を追加"
```

wardi: It's ok that the translations with placeholders don't work for every language. The default humanize_entity_type is *best effort*. When you want an exact translation in your language you can name your type something that makes sense in your preferred language, or to support multiple languages override the helper or use an extension like ckanext-scheming and put all the message translation in your schema next to the rest of the descriptions of those schemas.

smotornyuk: Ok, makes sense. But i expect complains from people, who think that such functionality must be available out of the box :)

smotornyuk:
Using chained_* was so obvious that I hadn't even thought about it xD
Updated

ckan/lib/helpers.py
```diff
@@ -2834,9 +2872,21 @@ def load_plugin_helpers():
...
     for plugin in reversed(list(p.PluginImplementations(p.ITemplateHelpers))):
-        helper_functions.update(plugin.get_helpers())
+        for name, func in plugin.get_helpers().items():
+            if _is_chained_helper(func):
+                chained_helpers[name].append(func)
+            else:
+                helper_functions[name] = func
+    for name, func_list in chained_helpers.items():
+        if name not in helper_functions:
+            raise Exception(
```
wardi, line 2885: Let's use `NotFound` instead of the base exception class here like we do for chained actions

ckan/lib/helpers.py
```diff
@@ -2834,9 +2872,21 @@ def load_plugin_helpers():
 
     helper_functions.clear()
     helper_functions.update(_builtin_functions)
+    chained_helpers = defaultdict(list)
 
     for plugin in reversed(list(p.PluginImplementations(p.ITemplateHelpers))):
```
wardi, line 2877: we don't reverse the plugins for chained helpers because we want "first plugin wins" like for templates, actions and other plugins.

ckan/plugins/interfaces.py
```diff
@@ -1059,6 +1059,22 @@ def get_helpers(self):
         Function names should start with the name of the extension providing
         the function, to prevent name clashes between extensions.
 
+        By decorating a registered helper function with the
+        ``ckan.plugins.toolkit.chained_helper`` decorator you can
+        create a chain of helpers that are called in a sequence. This
+        chain starts with the last chained helper to be registered and
```
wardi, line 1065: *first chained helper

ckan/controllers/home.py
```diff
@@ -57,9 +57,19 @@ def index(self):
...
+            group_label = h.humanize_entity_type(
+                u'group',
+                h.default_group_type(u'group'),
+                u'facet label') or _(u'Groups')
+
             c.facet_titles = {
-                'organization': _('Organizations'),
-                'groups': _('Groups'),
+                'organization': _(org_label),
+                'groups': _(group_label),
```
wardi, line 72: don't need to `_()` again

ckan/controllers/package.py
```diff
@@ -252,9 +252,19 @@ def pager_url(q=None, page=None):
...
+            group_label = h.humanize_entity_type(
+                u'group',
+                h.default_group_type(u'group'),
+                u'facet label') or _(u'Groups')
+
             default_facet_titles = {
-                'organization': _('Organizations'),
-                'groups': _('Groups'),
+                'organization': _(org_label),
+                'groups': _(group_label),
```
wardi, line 267: here too

ckan/lib/helpers.py
```diff
@@ -1126,8 +1127,55 @@ def _make_menu_item(menu_item, title, **kw):
...
+        u'save label': _(u"Save {object_type}"),
+        u'sidebar label': _(u"{object_type}"),
+        u'my label': _(u"My {object_type}s"),
+        u'no any objects': _(
+            u"There are currently no {object_type}s for this site"),
+        u'description placeholder': _(
+            u"A little information about my {object_type}..."),
+    }
+    if purpose not in templates:
+        return
```
wardi, line 1176: do you want to use this default path for the no-op humanize things like "sidebar label" or "default label" so they don't show up in the dict above with somewhat meaningless `_("{object_type}")` "translations". Also if our default just shows the object type then a new purpose won't return `None`

smotornyuk: Sorry, didn't get it

This is sidebar label - 
image

When dataset belongs to the organization with type `publisher` it has the sense to show `Publisher` as a sidebar label and that's what happens now. It's not a problem to remove those keys from dict and return `None`. In that case, the sidebar label will be `Organization`(default from template) but as long as we can redefine helper, it can be fixed.

Returning object type instead `None` as fallback won't work. There are a number of long messages, that are not added to the dict ([delete confirmation](https://github.com/ckan/ckan/pull/5337/files#diff-ea325fbc1a770c95b6cffacc7f23ecb7R39) ) and we have to return `None` in order to use template's default and not put such rare but big messages into dict.

Does it correlates somehow with your question?

wardi: of course, you're right I forgot about the default outside the caller.

ckan/lib/helpers.py
```diff
@@ -1126,8 +1127,53 @@ def _make_menu_item(menu_item, title, **kw):
...
+        u'save label': _(u"Save {object_type}"),
+        u'my label': _(u"My {object_type}s"),
+        u'no any objects': _(
+            u"There are currently no {object_type}s for this site"),
+        u'description placeholder': _(
+            u"A little information about my {object_type}..."),
+    }
+    if purpose not in templates:
+        return
+    type_label = object_type.replace(u"_", u" ").capitalize()
```
wardi, line 1175:
```suggestion
    type_label = object_type.replace(u"_", u" ").capitalize()
    if purpose not in templates:
        return type_label
```

smotornyuk: We've discussed it in [previous thread](https://github.com/ckan/ckan/pull/5337#discussion_r451040241) - it won't allow using default message from template. And in that way we have to put into `templates` dictionary every single message, even if it very long and unlikely to be changed, like [this one](https://github.com/ckan/ckan/pull/5337/files#diff-ea325fbc1a770c95b6cffacc7f23ecb7R39)

wardi: @smotornyuk I think this is required because you removed a couple keys from the dict above, and it seems like a sensible default for unknown purpose values

smotornyuk: ok then, let's go with it

smotornyuk: I've updated docstring a bit, mentioning default value

amercader:
@smotornyuk @wardi You mentioned this is a candidate for 2.9. If that still the case we should wrap it up as soon as possible because I see that the translated strings are an important part of this changeset, so it would be great if we could create the 2.9 Transifex resource with these strings included from the start (ie this cherry-picked into the dev-v2.9 branch before we create and upload the new pot file)
If you need some more time is fine, we can create the translations now and announce it, but mention that there will be some new strings coming up eg next week, but if we can freeze the strings before then great

wardi:
@amercader fixed and merged, I'm +0 on backporting to 2.9 because I'd like to see #5466 updated to use the same helper and maybe it doesn't make sense to have this feature for only groups/org labels in 2.9. @smotornyuk WDYT should we wait and push them both as part of 3.0 or cherry-pick just this one into 2.9?

smotornyuk:
Let's postpone it to 3.0. I agree that the dataset and group humanization must be delivered in the same release. But I don't want to make any delays in 2.9 release(everyone is waiting for it too long already). And I definitely don't want to create new bugs in core trying to things in a hurry and cherry-picking updated and untested #5466 into dev-2.9.

Even though it's quite a nice feature to have, we can wait for one more release.

amercader:
That sounds like a good plan!
Just to be clear https://github.com/ckan/ckan/pull/5358 is not related to this and should go into 2.9 right?

smotornyuk:
Exactly,  #5358  is a completely different change

PR #5339, opened 2020-04-10
[#4968] Fix multiline translations - Take 2

### LINKED ISSUE #4968: i18n: Multiline translation strings not translated
2.8.3

All translated strings in the CKAN templates should be translatable into all available languages.

Multi-line translated strings in the CKAN templates that are specified with `{% trans %}` tags are not translated.

e.g., this works: `{{ _('Update Member') }}`
This works: `{% trans %}Update Member{% endtrans %}`

This doesn't work, and the text remains in English when a different language is selected:
```
{% trans %}
    Some really long string that is so long
    we put it on multiple lines in the template.
{% endtrans %}
```

Bring up a new CKAN instance. Go to the Organizations page and see that the help text in the left column (under "What are Organizations?") does not change when a different language is selected.

amercader:
Fixes #4968 

This builds on #5338 but it goes beyond with what I think is a better approach.

For reference, this is a string extracted from a Jinja2 multiline block (a `msgid`) in the pot file in current master:

```po
#: ckan/templates/group/snippets/helper.html:8
msgid ""
" You can use CKAN Groups to create and manage collections of datasets. This "
"could be to catalogue datasets for a particular project or team, or on a "
"particular theme, or as a very simple way to help people find and search your"
" own published datasets. "
msgstr ""
```

As @smotornyuk correctly pointed in #5338, using the standard Jinja2 Babel extractor save us some custom code that we were using in our extractor. But as it is, #5338 extracts messages as:

```po
#: ckan/templates/group/snippets/helper.html:8
msgid ""
"\n"
"        You can use CKAN Groups to create and manage collections of datasets."
"\n"
"        This could be to catalogue datasets for a particular project or team,"
"\n"
"        or on a particular theme, or as a very simple way to help people find"
"\n"
"        and search your own published datasets.\n"
"      "
msgstr ""
```

If we use the above msgid in the po files, the string gets correctly translated. But by essentially keeping all existing whitespace including line breaks in the msgid, it makes them much more fragile as any whitespace change in the templates will result in a new msgid, and the existing translation will be lost.

I propose we use the [`ext.i18n.trimmed`](https://jinja.palletsprojects.com/en/2.10.x/api/#ext-i18n-trimmed) policy both when extracting messages (https://github.com/ckan/ckan/commit/da39d7ebe7eba2d3d562a82bc4cff5514ccc60fd) and when parsing translated blocks (https://github.com/ckan/ckan/commit/4cf6e0a64907133ef025892d4d320db6d96b1bf5) to avoid any whitespace differences.

This is how the string gets extracted with this policy:

```po
#: ckan/templates/group/snippets/helper.html:8
msgid ""
"You can use CKAN Groups to create and manage collections of datasets. This "
"could be to catalogue datasets for a particular project or team, or on a "
"particular theme, or as a very simple way to help people find and search your"
" own published datasets."
msgstr ""
```

Now, this gives us more robust msgids going forward, but as @smotornyuk correctly pointed out in #4968 if we update them now in the PO files, the current existing translations (`msgstr`) will be lost. To avoid this I wrote a small command that updates the msgids in the PO files with the corresponding new msgid in the POT file. You can see in the PO files in the PR that only the multiline strings were updated.

If the suggested approach makes sense to all, I'd like to backport this to 2.8. Here's the intended process:

* Backport 036972c, da39d7e and 4cf6e0a
* Adapt f69d086 to paster CLI
* Pull the latest translations from Transifex to get the latest PO files
* Extract new messages (update POT file)
* Run sync command (update PO files msgids)
* Commit and push to Transifex

@smotornyuk @wardi Does this make sense?


wardi:
backporting might break theme plugin translation overrides that expect the current whitespace, but maybe it's fine to just mention this in the release notes?

smotornyuk:
+1 for backporting. Eventually, every portal will be upgraded and the maintainer will have no choice but to do something with custom overrides. So adding clear notice about translation update would be ok - there are not so many multiline strings, so, I hope, it won't take much effort to do and update inside your extension while applying patch releases.

amercader:
There are 12 strings that were broken because of this, but some of them are quite prominent like the Groups and Orgs helper text (which are the ones that people will have customized if they wanted to rename them).

It is true that we can break some existing translation but I lean towards backporting and documenting because:

1. What we are fixing is more severe (users can't translate these core strings right now without messing with their own po files)
2. Nothing will break immediately after update. It would only happen after users have extracted the strings again and compiled the po files

I'll add the notice in the changelog as part of this PR so you can check if it is informative enough

amercader:
@smotornyuk @wardi if any of you has some time can you review and merge if happy? Thanks!

smotornyuk:
:heavy_check_mark: reviewed. Need to fix conflicts and then it's ready to merge

amercader:
Thanks @smotornyuk!

Chealer:
This was **not** backported in CKAN 2.8.4.

amercader:
I decided not to backport this to 2.8 at this stage. The Jinja2 version that 2.8 uses (`jinja2==2.8`) does not have support for policies, which were used to define the `trimmed` behaviour in da39d7ebe7eba2d3d562a82bc4cff5514ccc60fd and 4cf6e0a64907133ef025892d4d320db6d96b1bf5. Perhaps there is a lower-level way of achieving the same effect but it feels risky to try to hack something around this in a patch release.

PR #5342, opened 2020-04-14
fix order_by param in user_list action

### LINKED ISSUE #5323: user_list action ignores order_by param
2.8

```py
users = toolkit.get_action('user_list')(context, {
    'order_by': 'number_created_packages'
})
```
sorts users by `number_created_packages`

```py
users = toolkit.get_action('user_list')(context, {
    'order_by': 'number_created_packages'
})
```
sorts users by name

see above

---

Bug is here:
https://github.com/ckan/ckan/blob/64774670b722ed1bd3808901137fa80753d3c772/ckan/logic/action/get.py#L750:L757

Related discussion: https://github.com/okfn/ckanext-unhcr/pull/268#discussion_r395679730

chris48s:
Fixes #5323

### Proposed fixes:

Respect `order_by` param in `user_list` action.

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -747,14 +747,31 @@ def user_list(context, data_dict):
     if email:
         query = query.filter_by(email=email)
 
+    order_by_field = None
     if order_by == 'edits':
         raise ValidationError('order_by=edits is no longer supported')
-    query = query.order_by(
-        _case([(
-            _or_(model.User.fullname == None,
-                    model.User.fullname == ''),
-            model.User.name)],
-            else_=model.User.fullname))
+    elif order_by == 'number_created_packages':
+        order_by_field = order_by
+    elif order_by != 'name':
+        try:
+            order_by_field = getattr(model.User, order_by)
+        except AttributeError:
+            pass
+    if order_by == 'name' or order_by_field is None:
+        query = query.order_by(
+            _case(
+                [(
+                    _or_(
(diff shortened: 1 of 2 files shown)


ckan/logic/action/get.py
```diff
@@ -747,14 +747,31 @@ def user_list(context, data_dict):
...
-            model.User.name)],
-            else_=model.User.fullname))
+    elif order_by == 'number_created_packages':
+        order_by_field = order_by
+    elif order_by != 'name':
+        try:
+            order_by_field = getattr(model.User, order_by)
+        except AttributeError:
+            pass
+    if order_by == 'name' or order_by_field is None:
```
chris48s, line 760: @amercader I've implemented this as you suggested in https://github.com/okfn/ckanext-unhcr/pull/268#discussion_r402169184

I wonder if it would actually be better to only do this "sort by fullname if it exists else fall back to name" behaviour when `order_by_field is None` and just sort by `name` if the user has explicitly specified `order_by == 'name'`?

amercader: Sounds good. You should update the docs though to reflect that behaviour. Also it can definitely not "be any user field".

chris48s: I've updated this as suggested. Having learned a bit more about CKAN since originally submitting this PR, I now realise the correct terminology is to label that behaviour ordering by `display_name` :bulb: 

Docs are also updated.

PR #5345, opened 2020-04-15
Allow initial values in group form

smotornyuk:
Package create form picks values from the query string as an initial value for fields. This PR replicates that behavior for a group create form


smotornyuk:
Example of this feature for packages(you need to be logged in on demo.ckan.org) - https://demo.ckan.org/dataset/new?title=a&name=b&notes=c&url=d&author=e

Group form just ignores query string - https://demo.ckan.org/group/new?title=a . But with this PR you'll see that title field contains **a**

PR #5346, opened 2020-04-15
Collaborators: dataset-level permissions

amercader:
Screenshot_2020-04-14 Collaborators - Dataset - CKAN

## What is this and why it's useful

From the docs:

> In addition to traditional organization-based permissions, CKAN instances can also enable
the dataset collaborators feature, which allows dataset-level authorization. This provides
more granular control over who can access and modify datasets that belong to an organization,
or allows authorization setups not based on organizations. It works by allowing users with
appropriate permissions to give permissions to other users over individual datasets, regardless
of what organization they belong to.

## Naming

This was a tricky one as always with naming things, but I decided to go with "Collaborators" for the user facing naming of this feature. Although they resemble quite closely organization "Members" it felt more natural to call them collaborators as a dataset is not meant to be a collection of users. But I'm not a native English speaker so if people feel "Dataset members" work fine I'm happy to do a big search and replace.

Internally and at the API level though I thought it was important to be aligned with all the existing member based names (`Member` models and `member_*` actions) so models are named `PackageMember` and API actions `package_member_*`.

## Feature overview

Authorization settings are a delicate thing that require some planning on the maintainers and implementers side so this is feature is turned off by default (at least on 2.9)

Enabling it adds a new tab on the dataset Manage section to manage collaborators. Authorized users (more on this later) can add any existing user in the instance (not invite for now) as collaborator of the dataset, with different roles. By default, these rolese are:

* Member: can read private datasets
* Editor: can edit or delete datasets (including changing the organization)

So by default, organization admins retain full control over who can edit their organization datasets. If they want to allow collaborators to add new collaborators they can enable the Admin role via config:

* Admin: can also add, update or remove collaborators.

Another scenario enabled by this is non-org based CKAN where datasets don't need to belong to an organization. From the docs:

> For instance on instances where datasets don't need to belong to an organization (both :ref:`ckan.auth.create_dataset_if_not_in_organization` and :ref:`ckan.auth.create_unowned_dataset` are ``True``), the user that originally created a dataset can also add collaborators to it (allowing admin collaborators or not depending on the ``ckan.auth.allow_admin_collaborators`` setting). Note that in this case though, if the dataset is assigned to an organization, the original creator might no longer be able to access and edit, as organization permissions take precedence over collaborators ones.

So to sum, in this implementation a user can manage collaborators if: 

1. Is an administrator of the organization the dataset belongs to
2. Is a collaborator with role "admin" (assuming `ckan.auth.allow_admin_collaborators` is set to True)
3. Is the creator of the dataset and the dataset does not belong to an organization (requires `ckan.auth.create_dataset_if_not_in_organization` and `ckan.auth.create_unowned_dataset` set to True)

By default, collaborators can not change the owner organization of a dataset unless they are admins or editors in both the source and destination organizations. To allow collaborators to change the owner organization even if they don't belong to the source organization, set `ckan.auth.allow_collaborators_to_change_owner_org` to `True`.
## Implementation overview

This adds a new `package_member` table to store the relations. I tried to use the existing `member` one but it has a `group_id` foreign key, and rather to do some complex pg migration I just went for the simpler approach of having another table.

There are four new actions, the first three simplified versions of their `member` counterparts.

* `package_member_create` - Create / update collaborators
* `package_member_delete` - Remove
* `package_member_list` - Return the list of all collaborators for a given package
* `package_member_list_for_user` - Return a list of all packages the user is a collaborator in

Auth works as described in the previous section.

The critical parts of the existing authorization logic where it hooks are:

* [`package_update`](https://github.com/ckan/ckan/pull/5346/files#diff-7348b3443b4245b7471d3b278a5b6e13): Organization-based auth always take precedence. Only if org-based auth fails we check if the user is a collaborator
* [Default permission labels](https://github.com/ckan/ckan/pull/5346/files#diff-cf66d4c076e7091ab114f9e5bdb31c4c): there are labels added for the dataset collaborators

## Tests

Loads of them! as it touches auth stuff I rather be cautious and maybe a bit redundant. They could use some parametrization for sure but I still don't know pytest that well, and we can always improve them later.

## TODO

- [x] Admin collaborators tests
- [x] Tests for unowned datasets
- [x] Handle organization changes


pdekraker-epa:
@amercader one item with editor collaborator is the potential to alter dataset ownership and/or visibility. Blocking updates that modify those parameters by non-owners would probably be sufficient.

On a similar note, requiring an owner org admin to add collaborators seems a little restrictive. It is probably a good default, but enabling editors to modify collaborators would be handy.

pdekraker-epa:
Should protection similar to the owner_org be applied to package visibility too?  

One expansion on the collaborator idea would be to allow organizations or even groups to be collaborators too. I mention it now because It might be good to consider how it might affect this implementation. Extending this thought, maybe a CKAN 3 level idea, you could convert the owner org to a package admin and have everything with permissions / visibility handled through one interface.

amercader:
This is finally ready for review! :turtle: 
If someone from the @ckan/core team can help @pdelboca with the review that would be great!

amercader:
@pdekraker-epa I don't think the visibility change protection should be part of core. Preventing the change of owner organization is a sane default to prevent dataset owners getting locked out of their datasets but the rest of fields would depend on the specific needs of the site, so it's probably better placed in an extension.
And in any case I think this PR is already too big :) We can expand the feature in future versions.

ckan/authz.py
```diff
@@ -397,6 +404,47 @@ def get_user_id_for_username(user_name, allow_none=False):
...
+
+    if (not owner_org
+            and check_config_permission('create_dataset_if_not_in_organization')
+            and check_config_permission('create_unowned_dataset')
+            and pkg.creator_user_id == user_id):
+        # User is the creator of this unowned dataset
+        return True
+
+    if has_user_permission_for_group_or_org(
+            owner_org, user_id, 'membership'):
```
smotornyuk, line 434: Probably, `ROLE_PERMISSIONS['admin']` should be updated with this new `membership` permission? Or, you should explicitly check `admin` permission, if you are relying on it to bypass unexisting `membership` check. But I think it would be better just to add `membership` to the `ROLE_PERMISSIONS['admin']`  - it can be used as an example for anyone who extends `ROLE_PERMISSIONS` dictionary.

ckan/logic/action/create.py
```diff
@@ -619,6 +619,77 @@ def member_create(context, data_dict=None):
...
+            _('Role must be one of "{}"').format(', '.join(
+                allowed_capacities)))
+
+    package = model.Package.get(package_id)
+    if not package:
+        raise NotFound(_('Dataset not found'))
+
+    user = model.User.get(user_id)
+    if not user:
+        raise NotFound(_('User not found'))
```
smotornyuk, line 667: During last year we've merged few PRs which were trying to prevent exposing of the user account to anonymous request. For example, you won't see **User not found** when resetting the password, you'll see **email was sent to your mailbox**  even if you misspelled username. But this error comes before access check and creates a way to verify, whether the account exists on the portal.

amercader: Good point. What error would you suggest returning here? Something generic like "Object not found" for both missing datasets and users is not very useful as it's trivial to check if a (public) dataset exists. So maybe raise `NotAuthorized` if the user is not found?

ckan/authz.py
amercader: Sounds good, I copied the check from another auth function but what you propose makes sense

rufuspollock:
Late in the day to ask, so more for clarity than to propose a change: why do this in core vs an extension? If some core functionality is needed what is it and could we just add a new plugin point? What are the trade-offs overall?

(Motivation: keeping core "surface area" minimal so that the codebase is more maintanable and extensible).

ckan/logic/action/create.py
smotornyuk: I think it's optimal solution, so +1 for `NotAuthorized`

ckan/authz.py
```diff
@@ -397,6 +404,47 @@ def get_user_id_for_username(user_name, allow_none=False):
...
+        # User is an administrator of the organization the dataset belongs to
+        return True
+
+    if check_config_permission('allow_admin_collaborators'):
+        collaborators = p.toolkit.get_action('package_member_list')(
```
pdelboca, line 439: Any reason to use plugin toolkit's  `get_action(...)` instead of `ckan.logic` one?

ckan/migration/alembic.ini
```diff
@@ -43,7 +43,7 @@ script_location = %(here)s
 # functionality is considered as internal, so don't rely on it,
 # because it may be removed in future. Use official CKAN commands
 # rather than direct interactions with alembic
-sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_default
+sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_default_3
```
pdelboca, line 46: Is `ckan_default_3` a valid value? Or is something for your local instance?

ckan/templates/package/collaborators/collaborator_new.html
```diff
@@ -0,0 +1,76 @@
...
+              value="" class="control-medium" data-module="autocomplete"
+              data-module-source="/api/2/util/user/autocomplete?q=?">
+            {% endif %}
+          </div>
+        </div>
+      </div>
+    </div>
+
+    {% set format_attrs = {'data-module': 'autocomplete'} %}
+    {{ form.select('capacity', label=_('Capacity'), options=capacities, selected=user_capacity, error='', attrs=format_attrs) }}
```
pdelboca, line 41: To be consistent with the other parts of the UI this label should be Role and not Capacity.

```suggestion
    {{ form.select('capacity', label=_('Role'), options=capacities, selected=user_capacity, error='', attrs=format_attrs) }}
```

ckan/lib/helpers.py
```diff
@@ -2863,3 +2863,56 @@ def activity_list_select(pkg_activity_list, current_activity_id):
...
+    if (user
+            and authz.check_config_permission('allow_dataset_collaborators')
+            and user.id in [
+                co[0] for co in get_collaborators(package_dict['id'])
+            ]):
+
+        # User is a collaborator, allow to change the owner_org depending
+        # on config
+        return authz.check_config_permission(
+            'allow_collaborators_to_change_owner_org')
```
wardi, line 2916: short circuit a few lines of code by moving this authz check above `user = model.User.get(c.user)`

ckan/lib/plugins.py
```diff
@@ -623,15 +623,24 @@ class DefaultPermissionLabels(object):
     - everyone can read public datasets "public"
     - users can read their own drafts "creator-(user id)"
     - users can read datasets belonging to their orgs "member-(org id)"
+    - users can read datasets where they are collaborators "collaborator-(dataset id)"
```
wardi, line 626:
```suggestion
    - users can read datasets where they are collaborators "collaborator-(user id)"
```

ckan/logic/action/create.py
```diff
@@ -619,6 +619,77 @@ def member_create(context, data_dict=None):
     return model_dictize.member_dictize(member, context)
 
 
+def package_member_create(context, data_dict):
```
wardi, line 622: apis are pretty hard to rename, should we go with `package_collaborator_create` et.al instead?

ckan/logic/action/create.py
wardi: `ValidationError` seems more apt for a problem with one parameter of an API call, but I agree it should be done after the `_check_access` call below

ckan/logic/action/get.py
```diff
@@ -233,6 +233,115 @@ def translated_capacity(capacity):
...
+        capacity and the last modified date
+    :rtype: list of dictionaries
+
+    '''
+
+    model = context['model']
+
+    user_id = _get_or_bust(data_dict, 'id')
+
+    user = model.User.get(user_id)
```
wardi, line 310: user info leaking here as well, move this after _check_access?

ckan/logic/auth/update.py
```diff
@@ -34,10 +33,24 @@ def package_update(context, data_dict):
...
+            # if org-level auth failed, check dataset-level auth
+            # (ie if user is a collaborator)
+            user_packages = logic.get_action(
+                'package_member_list_for_user')(
+                    context, {'id': user})
+            for user_package in user_packages:
+                if (user_package['package_id'] == package.id
+                        and user_package['capacity'] in ('admin', 'editor')):
+                    success = True
+                    break
```
wardi, line 49: Query the table directly instead of doing this check in python so that users with large collaborator lists don't slow down package_update calls.

ckan/logic/validators.py
```diff
@@ -49,6 +49,27 @@ def owner_org_validator(key, data, errors, context):
...
 
+    if (authz.check_config_permission('allow_dataset_collaborators')
+            and not authz.check_config_permission('allow_collaborators_to_change_owner_org')):
+
+        package = context.get('package')
+        if package and user and not user.sysadmin:
+            try:
+                user_packages = logic.get_action(
+                    'package_member_list_for_user')(
+                            {'ignore_auth': True}, {'id': user.id})
```
wardi, line 60: same comment as above, if you'd like to avoid db calls all over the place maybe add a helper just for checking the collaborator status for a single package id + user id.

ckan/migration/versions/097_f789f233226e_add_package_member_table.py
```diff
@@ -0,0 +1,35 @@
...
+    op.create_table(
+        u'package_member',
+        sa.Column(u'id', sa.UnicodeText(), nullable=False),
+        sa.Column(u'package_id', sa.UnicodeText(), nullable=True),
+        sa.Column(u'user_id', sa.UnicodeText(), nullable=True),
+        sa.Column(u'capacity', sa.UnicodeText(), nullable=False),
+        sa.Column(u'modified', sa.DateTime(), nullable=True),
+        sa.ForeignKeyConstraint([u'package_id'], [u'package.id'], ),
+        sa.ForeignKeyConstraint([u'user_id'], [u'user.id'], ),
+        sa.PrimaryKeyConstraint(u'id')
```
wardi, line 30: package_id + user_id would be a better primary key and db would prevent us from adding the user as a collaborator on the same dataset multiple times. I can't think of a reason to have a separate id field on this table.

ckan/logic/action/create.py
amercader: @wardi the problem I see with `ValidationError` is that is generally tied to one or more fields values (eg "id is missing", "value not allowed", etc) and I'm not sure what error message we would include with one of them (`ValidationError: User not authorized` seems wrong). I went for a NotAuthorized but happy to change it if you have an alternative.

ckan/authz.py
amercader: Yes, to avoid circular import problems (logic imports authz, so if we import logic from authz, :boom: )

ckan/lib/helpers.py
amercader: We want to apply this check only to users that are collaborators right? That's what the lines above check. Otherwise users that are not admin/editors nor collaborators might be allowed to change the owner_org if `allow_collaborators_to_change_owner_org` is True.

amercader: Sorry, I see what you mean, no need to check the rest if the config is False

ckan/logic/auth/update.py
amercader: I think I addressed this in 7d270b739

ckan/authz.py
amercader: But thanks for flagging, as this needed refactoring (bdd3b93ba)

ckan/logic/action/create.py
amercader: Sounds good!
I didn't rename the models (`PackageMember`) because there are not used widely and we might want to rename the `Package` bit at some point anyway :)

amercader:
@smotornyuk @pdelboca @wardi Thanks a lot for your review comments. I think I addressed them all now.

PR #5357, opened 2020-04-26
Deployment doc - little improvements

davidread:
* fixed instruction of reloading nginx, when it should be restarting -
you need to start nginx before it will let us reload it. (It doesn't
start when you install it, because apache is listening on port 80 by
default when you installed it)
* service -> systemctl - because systemd replaced upstart in ubuntu
since 16.04, and people are very unlikely to be installing on 14.04 now.
* rst syntax corrections

Fixes #

### Features:

- [ ] includes API changes


amercader:
These look good @davidread , just one warning added to the Sphinx build:
``
u'/root/project/doc/maintaining/installing/deployment.rst:203: WARNING: Undefined substitution referenced: "restart_nginx".'
``

davidread:
Looks good now

davidread:
Ready for re-review @amercader

PR #5358, opened 2020-04-27
Type in templates

smotornyuk:
Adds argument to `*_template` methods of `IDatasetForm`/`IGroupForm`. 
New argument is a context dictionary, with additional info, used for more precise template selection. Right now it contains the type of corresponding entity, allowing to do something like:

```python
def read_template(self, options):
    return 'package/{type}/read.html'.format(type=options['package_type'])
```

In order to preserve backward compatibility, this extra argument is passed only to methods, that are expecting it (by checking the declared number of arguments).

### DIFF
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -53,7 +53,8 @@ def _setup_template_variables(context, data_dict, package_type=None):
 
 def _get_pkg_template(template_type, package_type=None):
     pkg_plugin = lookup_package_plugin(package_type)
-    return getattr(pkg_plugin, template_type)()
+    method = getattr(pkg_plugin, template_type)
+    return method()
 
 
 def _encode_params(params):
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -54,12 +54,10 @@ def _setup_template_variables(context, data_dict, package_type=None):
 def _get_pkg_template(template_type, package_type=None):
     pkg_plugin = lookup_package_plugin(package_type)
     method = getattr(pkg_plugin, template_type)
-    if method.__code__.co_argcount == 1:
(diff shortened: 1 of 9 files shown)


wardi:
as discussed in call: please change parameter to `package_type` and handle exception for missing parameter instead of inspecting function parameters

smotornyuk:
Updated

ckan/views/group.py
```diff
@@ -41,16 +41,13 @@
...
-
-
-def _form_to_db_schema(group_type=None):
-    return lookup_group_plugin(group_type).form_to_db_schema()
+def _get_group_template(template_type, group_type=None):
+    group_plugin = lookup_group_plugin(group_type)
+    method = getattr(group_plugin, template_type)
+    try:
+        return method(group_type)
+    except TypeError:
```
wardi, line 49:
```suggestion
    except TypeError as err:
        if 'takes no arguments' not in str(err):
            raise
```

smotornyuk: Yep, good point

ckan/views/dataset.py
```diff
@@ -53,7 +53,11 @@ def _setup_template_variables(context, data_dict, package_type=None):
 
 def _get_pkg_template(template_type, package_type=None):
     pkg_plugin = lookup_package_plugin(package_type)
-    return getattr(pkg_plugin, template_type)()
+    method = getattr(pkg_plugin, template_type)
+    try:
+        return method(package_type)
+    except TypeError:
+        return method()
```
wardi, line 60: more specific exception check here too please

smotornyuk:
Done. There are quite a lot similar lines inside dataset and group views - probably it worth doing some code-extraction after release 2.9, when pylons code removed

PR #5360, opened 2020-04-30
Make ckan accessible, part 1

Zharktas:
Fixes ckan/ideas-and-roadmap#238

### Proposed fixes:

Adds [cypress-axe](https://github.com/avanslaars/cypress-axe) plugin to cypress which runs a11y axe validation tool on the pages tested with cypress. 

Adds cypress tests starting with the basic pages and moving on to more complex user interactions while fixing validation issues which are reported.

As I am not accessibility professional, this will mostly only cover issues which can be discovered with automated tools. It would require professional to fix the more complex ones. 

This is an ongoing effort to make ckan accessible, I'll try to finish as much as I can before 2.9 release, but most likely it will not cover everything.

Automated tests only cover pages in the main nav, as individual datasets and organizations would require pre-populating them in the database for cypress tests. 

Forms are also not covered by this PR, they affect far less in the user base anyway, so those would be fixed later.

There are user visible changes, as some colors are changed to make them have enough contrast.

This PR also updates Cypress to 4.8.0.

### Features:

- [ ] includes API changes


Zharktas:
Weirdly there is some color difference between python 2 and 3, might be a bug somewhere.

smotornyuk:
Great work. No problems detected, so i'm merging it

PR #5367, opened 2020-05-05
Add make_resource fixture

smotornyuk:
Provide a new fixture for imitating resources with the uploaded files. With it, we can remove a dozen of lines from tests and it really helps in cases when one need to upload file in order to test some reaction of extension. I was indexing content of resource's file for some of my projects and used this fixture in extension for a while - it's pretty stable.


ckan/tests/pytest_ckan/fixtures.py
```diff
@@ -248,3 +250,55 @@ def with_extended_cli(ckan_config, monkeypatch):
...
+    content_type = None
+
+    def __init__(self, stream, filename):
+        self.stream = stream
+        self.filename = filename
+        self.name = u"upload"
+
+
+@pytest.fixture
+def make_resource(clean_db, ckan_config, monkeypatch, tmpdir):
```
amercader, line 265: I'd call it `make_resource_with_upload` or `make_uploaded_resource`.
Or in fact what about `create_with_upload` and allow to pass a parameter for the upload field name. The defaults would be `action=resource_create` and `upload_field_name=upload` so you create a resource by default, but this way we can also use it to create users, orgs and groups with uploaded images.

Issue #4147, opened 2018-03-23
Plugin Extras


wardi:
Extending #3072 to implement `plugin_extras` fields on:
* Users
* Groups and Organizations (same underlying model)
* Datasets
* Resources

`plugin_extras` is a JSON field that is not exposed through the user interface or API (by default.) It's intended as a generic way for plugins to attach data to Users, Groups, Organizations, Datasets and Resources. 

By convention each plugin will store its own data in a separate JSON object, e.g. `my_user.plugin_extras['funky_auth']` would be the place a `ckanext.funky_auth` extension could store its data related to `my_user`.

For API-level migration of data from one CKAN instance to another, sysadmin users may request `plugin_extras` data by passing `include_plugin_extras=true` to the appropriate `*_show` API, and `plugin_extras` values passed to `*_update` and `*_create` APIs will be accepted when called by a sysadmin user. By default (and for all other users) this key is ignored and not returned.

Plugin extras are **not** revisioned and **not** copied into activity records. Values stored are private and hidden to normal users.

Plugins may choose to expose values they store in `plugin_extras` and/or allow them to be updated from the API and front end by chaining actions or customizing validation rules, **not** through any new interface introduced by this change.

This PR only creates the `plugin_extras` fields to enable future plugins to do something actually useful, e.g. https://github.com/ckan/ideas-and-roadmap/issues/211

torfsen:
I'd love to see the `plugin_extras` field on as many CKAN entities as possible. For example, I currently need a way to attach plugin-related extras to resource views (see also #2655).

Aside from the entities already mentioned by @wardi, universal support could include:

* Views
* Tags
* Relationships (but see #4212)
* Vocabularies
* Activities

Perhaps there is a way to implement this mechanism in some kind of "base  class" so that all current and future entities (in CKAN core and in extensions) are automatically supported?

rufuspollock:
Repeating comment from https://github.com/ckan/ideas/issues/211 (naturally we have ended up with roughly same item in both ideas and here):

Would it not be better to store workflow related information in a table dedicated to workflow rather than adding stuff to the dataset object (or other core objects)? This creates a cleaner setup where the direction of dependency is better: the workflow system depends on dataset not the other way round (you aren't changing the dataset object because of the workflow system).

I note that where this started which was with User Extras it makes more sense #16  - though even there i would be concerned to limit the use of this (it should really be specific stuff added to the user and other extensions can create their own tables).

[...]

PR #5382, opened 2020-05-14
User plugin_extras

amercader:
These changes follow the plan outlined in this issue https://github.com/ckan/ckan/issues/4147 and basically add a new JSON field to the user model to store arbitrary data. 

```
   {
        "name": "test-user",
        "plugin_extras": {
            "gdx": {
                "team": "value"
                "title": "value"
            }
        }
    }
```

Only sysadmins can set or edit this field, and only they can access it via the API if passing the `include_plugin_extras` parameter.


rufuspollock:
@amercader as per comment here https://github.com/ckan/ideas/issues/211#issuecomment-628883843 i think this could be something of an anti-pattern. Why are we storing extra info onto the user? If the use case is to support some extension why can't the extension store relevant info in its own table somewhere?

wardi:
@rufuspollock I replied to your comment on #4147 and I'd like to add: I'm really looking forward to this change, it will support enable free-form extras, scheming-style validated extras without conflict and with much better performance than the strings we store package_extras-style for datasets, groups and orgs. Once it's merged we can use code like it to reimplement package, group and resource extras transparently to existing extensions. :+1:

amercader:
@rufuspollock adding extra fields to objects as `extras` is a concept as old as CKAN so very much a pattern I'd say. The name `plugin_extras` came about to differentiate them from traditional extras in packages and groups as the plan is to add them to these objects as well, and eventually replace the existing ones.

rufuspollock:
@amercader

>  adding extra fields to objects as extras is a concept as old as CKAN so very much a pattern I'd say.

Being the "way we did things" does  not mean it has turned out to be a good idea or what we should keep doing 😉 (and i say this as the person who designed it in in the first place). We could change moving forward. Of course, whether we should is the open question.

I agree that distinguishing as plugin extras is useful but then why not just have each plugin have its own table. That's potentially cleaner ... (or if that is really a problem a central table for "plugin extras" with good clean namespacing). 

Here's why I think think that it could be better to keep "plugin attributes" explicitly separated in DB / ABPI:

1. There is a real distinction between actual extra "metadata" and non-metadata housekeeping info we are sticking in as "metadata". A concrete example: datastore_active should probably not in the core descriptor of a dataset. It's housekeeping information *about the datastore*.
2. Keeping things separate makes it cleaner to introduce and remove things. 
3. We avoid increasingly fat (and complex APIs) for the core objects. Instead we expect clients to fetch the extra meta they need.

cf https://martinfowler.com/articles/schemaless/ which i find very useful in thinking about "schemaless" setups.

wardi:
I believe plugin extras does all the above.
1. It's not part of the metadata exposed with the API (unless a plugin copies it in a validator) Totally agree fields like datastore_active and hash (as its used by datapusher) should be separated with a mechanism like this from the core fields
2. It's labelled "plugins" because it's for plugin use, not core. It's namespaced per-plugin with top-level keys to avoid conflicts and allow for plugins to manage their own data, migration etc.
3. No change to existing APIs but plugins are free to expose their fields with validators.

rufuspollock:
@wardi 

> It's not part of the metadata exposed with the API (unless a plugin copies it in a validator) Totally agree fields like datastore_active and hash (as its used by datapusher) should be separated with a mechanism like this from the core fields

Great. Am I right in assuming that this stuff gets stored into some separate table like `PluginExtras` (or `PluginExtrasPackage`, `PluginExtrasGroup` etc)?

> It's labelled "plugins" because it's for plugin use, not core. It's namespaced per-plugin with top-level keys to avoid conflicts and allow for plugins to manage their own data, migration etc.

👍 and cf this thread about how to display those sorts of properties in a datapackage.json https://github.com/frictionlessdata/specs/issues/663

> No change to existing APIs but plugins are free to expose their fields with validators.

My point here would be that i think it would be nice to start separating APIs more per plugin rather than ending up with e.g. "fat" package objects in the API. Yes, we might want, graphql like, to have somewhere you can get "everything" but the default should be an API per major feature and clients pick and mix as they need.

amercader:
@wardi After using this for a while in live environments and building the proper root level fields on top of plugin extras I was bitten two times by an implementation detail that I think should be addressed in core.

It relates with tracking changes in the JSON field we use to store the plugin extras (or "mutation tracking" in SQLAlchemy lingo) and having them reflected in the database. I pushed a test in ad010f419639418b45aeceec91ae710e03e4be4d that illustrates a the issue, but basically if you start from a user like this:

```json
{
    "name": "test_user",
    "plugin_extras": {
         "plugin1": { 
             "key1": "value1",
             "key2": "value2",
       }
    }
}
```
If you modify one of the values directly (user_dict['plugin_extras']['plugin1']['key1'] = 'value2') when saving the model the change won't be reflected on the database. We are using the [`MutableDict`](https://github.com/ckan/ckan/pull/5382/commits/74b528c5b94d22b7fcc5d5f1831c5124aded170f#diff-82a2d330f46d2d29019ac12e726015feR39) class for this field, but according to [SQLA docs](https://docs.sqlalchemy.org/en/13/orm/extensions/mutable.html#sqlalchemy.ext.mutable.MutableDict):

> Note that MutableDict does not apply mutable tracking to the values themselves inside the dictionary. Therefore it is not a sufficient solution for the use case of tracking deep changes to a recursive dictionary structure, such as a JSON structure. 

Which means that it only tracks changes on the values of the first level. But given that we want to namespace every plugin (so one extra level) this essentially means that on each update right now you need to provide a copy of the existing extras in order to get the whole dict marked as changed eg:

```
user_dict['plugin_extras'] = {'plugin1': {'key1': 'value2'}
``
To add to the confusion, the models and dict returned will have the change reflected and is not until you go to the database that you will notice that the value is not updated.

To address this the docs say:

> To support this use case, build a subclass of MutableDict that provides appropriate coercion to the values placed in the dictionary so that they too are “mutable”, and emit events up to their parent structure.

That's easier said than done, I had a quick go but didn't go far. Luckily someone else has dealt with it before:

**Comment by wardi** [comment]
we could add a utility function to the model to help with updating values or use a more verbose way of updating plugin extras:
```python
user.plugin_extras['plugin1']['key1'] = 'value2'
# becomes:
user.plugin_extras['plugin1'] = dict(
    user.plugin_extras.get('plugin1', {}),
    key1='value2')
```
^ this or a similar utility function has the benefit of working even if the base level of the plugin extra for this plugin hasn't been set yet.

amercader:
@wardi I'd really like the handling of plugin extras to be as transparent as possible for plugin authors, ie if you update the plugin_extras dict in any way and save it via `user_update` the changes should be stored. So I'm not convinced about relying on users using the verbose option (this or just doing a copy of the whole dict is what I'm currently using in my project).

As alternative to using the external library we could create a copy of the plugin_extras just after pulling them from the database, which would mean that any further change to them would be picked up by SQLAlchemy, this is one line change:
```diff
diff --git a/ckan/lib/dictization/model_dictize.py b/ckan/lib/dictization/model_dictize.py
index 7a92ac0f5..08f6025da 100644
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -11,7 +11,7 @@ The basic recipe is to call:
 
 which builds the dictionary by iterating over the table columns.
 '''
-
+import copy
 import six
 from six.moves.urllib.parse import urlsplit
 
@@ -526,8 +526,8 @@ def user_dictize(
             result_dict['password_hash'] = password_hash
 
         if include_plugin_extras:
-            result_dict['plugin_extras'] = plugin_extras
-
+            if plugin_extras:
+                result_dict['plugin_extras'] = copy.deepcopy(plugin_extras)
 
     model = context['model']
     session = model.Session
```

wardi:
You shouldn't need a deepcopy to force detection of a change, just returning `dict(plugin_extras)` should force sqlalchemy to re-serialize the contents, right?

amercader:
@wardi `dict()` will only update the reference of the dict itself, and use the existing values for the root keys, so eg `plugin_extras['plugin1']` will still be the same object:

```
ipdb> pp plugin_extras
{u'plugin1': {u'key1': 'value2'}}
ipdb> id(plugin_extras)
140016662774752
ipdb> id(plugin_extras['plugin1'])
140016679925520
ipdb> xx = dict(plugin_extras)
ipdb> id(xx)
140016651413840
ipdb> id(xx['plugin1'])
140016679925520
ipdb> yy = copy.deepcopy(plugin_extras)
ipdb> id(yy)
140016652010160
ipdb> id(yy['plugin1'])
140016651440784

```

wardi:
agreed, but doesn't calling `user_update` with a new dict trigger sqlalchemy's dirty flag on the field? Or does MutableDict compare the identity of the values contained in the new dict passed?

wardi:
Oh I see it must do a comparison of the dicts to check for changes. Makes sense, deepcopy would be the only way.

amercader:
@wardi I went for the copy approach as it's simpler. Happy to discuss it and replace it on the review process.

This is now ready to review.

ckan/logic/validators.py
```diff
@@ -888,3 +889,16 @@ def callable(value):
...
+
+def json_object(value):
+    ''' Make sure value can be serialized as a JSON object'''
+    if value is None or value == '':
+        return
+    try:
+        if not json.dumps(value).startswith('{'):
+            raise Invalid(_('plugin_extras should be valid a JSON object'))
+    except ValueError as e:
+        raise Invalid(_('Could not parse plugin_extras as a valid JSON object'))
```
wardi, line 902: the field name is already included as the key of the errors dict, no need to repeat it here (and removing plugin_extras from the error lets us use this validator elsewhere)

wardi:
just one minor comment above otherwise LGTM. Hope one of us gets time to work on IUserForm soon to make it as easy to customize the user forms as the dataset forms.

amercader:
Thanks @wardi , I've changed the validator message

wardi:
@amercader looks like the JSONB field bumps our postgresql version requirement to 9.4

amercader:
@wardi Yes, that's right. We should mention that in the changelog and documentation. Looks like 9.4 is not supported anymore so we should probably recommend 9.5 or higher:
https://www.postgresql.org/support/versioning/

PR #5386, opened 2020-05-16
Revert "Override datapusher site url" in ckan 2.7

### LINKED ISSUE #5385: Datapusher requests failing in 2.7.7 when using uwsgi sockets
2.7.7

Datapusher requests to ckan should succeed when using nginx proxy and uwsgi sockets on the ckan server

Datapusher fails with an error in the UI, and uswgi error in the ckan container logs: 
```
invalid request block size: 21327 (max 4096)...skip
```

Run ckan with datapusher, ckan running uwsgi with flags:
```
--protocol uwsgi  --socket 0.0.0.0:5000
```

So this appears to have been introduced in https://github.com/ckan/ckan/commit/9a41f5a791bbc551959d3608103340ce545785c0

@amercader I'm confused as to why this url would be hardcoded like this? In our particular setup it almost works since we have our ckan container as a named route in the docker-compose file and we're running on port 5000, but since it bypasses the nginx proxy it is trying to talk http to uwsgi socket. Looking at the master branch it seems to be still in the same format as prior to your commit, `h.url_for('/', qualified=True)`.

markstuart:
Fixes #5385

### Proposed fixes:
Reverts hardcoded site url, forcing datapusher requests to go via proxy (if present) rather than direct to ckan server.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datapusher/logic/action.py
+++ b/ckanext/datapusher/logic/action.py
@@ -60,7 +60,7 @@ def datapusher_submit(context, data_dict):
 
     datapusher_url = config.get('ckan.datapusher.url')
 
-    site_url = 'http://ckan:5000'
+    site_url = h.url_for('/', qualified=True)
 
     callback_url_base = config.get('ckan.datapusher.callback_url_base')
     if callback_url_base:


markstuart:
See #5393 for fix for Travis build failing

PR #5394, opened 2020-05-18
Update CKANTestClient

smotornyuk:
Ignore legacy param `expect_errors` in TestClient requests. Right now it used in [ckanapi](https://github.com/ckan/ckanapi/blob/master/ckanapi/testappckan.py#L54) and can be potentially used in all extensions that are supported in CKAN <=2.8

### DIFF
--- a/ckan/tests/helpers.py
+++ b/ckan/tests/helpers.py
@@ -240,13 +240,11 @@ class CKANTestApp(object):
 
 class CKANTestClient(FlaskClient):
     def open(self, *args, **kwargs):
+        kwargs.pop('expect_errors', None)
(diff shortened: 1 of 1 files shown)


wardi:
don't think this class gets much use but happy to see it improved!

wardi:
@smotornyuk on second thought it would be nice to add a comment to this line explaining why this one keyword argument is filtered out, WDYT?

smotornyuk:
Yes, a good idea. In a few years, this line will become quite intriguing without comment :) I'm was going to check and update some of the dev requirements/add `ipdb` for debugging as dev-requirement, so I'll put comment inside new PR

PR #5400, opened 2020-05-19
Configure CodeClimate

smotornyuk:
This small change adds free [codeclimate reports](https://codeclimate.com/github/DataShades/ckan) (because CKAN is open-source, yay :partying_face: ). I think they will be quite useful when we start removing legacy code. It doesn't require any additional integration with GitHub/circleCI, it's free(at least,  it looks so) and it helps to make CKAN better - it's an absolute win.


smotornyuk:
I've updated `.codeclimate.yml` - now it contains commented examples of different options for analysis control. 
Any check can be disabled completely(now we don't have restrictions on file length) or changed(in our case allowed the number of arguments in functions and cognitive complexity were increased).
In addition, we can just ignore files/folders matching the specified pattern and now all the JS files and tests are excluded from checks.

PR #5402, opened 2020-05-21
Add towncrier

smotornyuk:
Improve changelog management. Instead of periodical updating of the single **CHANGELOG.rst**, any new feature/bugfix/deprecation/etc supplied with a file inside **changes/** directory. Each such file is following naming format `{issue number}.{change type}`. The issue number will be used to generate a link to the corresponding PR/issue and change type define group, under which message from the file will be shown. For now, I have defined 4 groups - `migration`, `removal`, `bugfix` and `misc`. But we can add as many groups as required.
When a release is ready, just run `towncrier --yes`. This command will remove (if you omit `--yes`, you can keep those files) all the files from **changes/**, but their content will be rendered inside **CHANGELOG.rst**. And there is also  `towncrier --draft` option, that only renders content to standard output, but doesn't perform any real change.

I think it can improve the general process. Anyone can just create a changelog fragment(file inside **changes/**) while working on PR and those files will be picked when doing backports. And there will be fewer chances to miss some new features.


smotornyuk:
@amercader, I've fixed conflict and added a few notes to documentation

<details><summary>Here's result of <strong>changes</strong> compilation with <code>towncrier</code></summary> 

```rst

v.2.9.0a 2020-05-28
===================

Migration
---------

- A full history of dataset changes is now displayed in the Activity Stream to
  admins, and optionally to the public. By default this is enabled for new
  installs, but disabled for sites which upgrade (just in case the history is
  sensitive). When upgrading, open data CKANs are encouraged to make this
  history open to the public, by setting this in production.ini:
  ``ckan.auth.public_activity_stream_detail = true`` (`#3972
  <https://github.com/ckan/ckan/pull/3972>`_)
- When upgrading from previous CKAN versions, the Activity Stream needs a
  migrate_package_activity.py running for displaying the history of dataset
  changes. This can be performed while CKAN is running or stopped (whereas the
  standard `paster db upgrade` migrations need CKAN to be stopped). Ideally it
  is run before CKAN is upgraded, but it can be run afterwards. If running
  previous versions or this version of CKAN, download and run
  migrate_package_activity.py like this::

    cd /usr/lib/ckan/default/src/ckan/
    wget
  https://raw.githubusercontent.com/ckan/ckan/3484_revision_ui_removal2/ckan/migration/migrate_package_activity.py
    wget
  https://raw.githubusercontent.com/ckan/ckan/3484_revision_ui_removal2/ckan/migration/revision_legacy_code.py
    python migrate_package_activity.py -c /etc/ckan/production.ini

  Future versions of CKAN are likely to need a slightly different procedure.
  Full info about this migration is found here:
  https://github.com/ckan/ckan/wiki/Migrate-package-activity (`#4784
  <https://github.com/ckan/ckan/pull/4784>`_)
- This version requires changes to the ``who.ini`` configuration file. If your
  setup doesn't use the one bundled with this repo, you will have to manually
  change the following lines::

       use = ckan.lib.auth_tkt:make_plugin

  to::

       use = ckan.lib.repoze_plugins.auth_tkt:make_plugin

  And also::

       use = repoze.who.plugins.friendlyform:FriendlyFormPlugin

  to::

       use = ckan.lib.repoze_plugins.friendly_form:FriendlyFormPlugin

  Otherwise, if you are using symbolinc link to ``who.ini`` under vcs, no
  changes required. (`#4796 <https://github.com/ckan/ckan/pull/4796>`_)


Bugfixes
--------

- Action function "datastore_search" would calculate the total, even if you set
  ``include_total=False``. (`#4448 <https://github.com/ckan/ckan/pull/4448>`_)


Removals and deprecations
-------------------------

- Revision and History UI is removed: `/revision/*` & `/dataset/{id}/history`
  in favour of `/dataset/changes/` visible in the Activity Stream.
  ``model.ActivityDetail`` is no longer used and will be removed in the next
  CKAN release. (`#3972 <https://github.com/ckan/ckan/pull/3972>`_)
- ``c.action`` and ``c.controller`` variables should be avoided.
  ``ckan.plugins.toolkit.get_endpoint`` can be used instead. This function
  returns tuple of two items(depending on request handler):
  1. Flask blueprint name / Pylons controller name
  2. Flask view name / Pylons action name
  In some cases, Flask blueprints have names that are differs from their
  Pylons equivalents. For example, 'package' controller is divided between
  'dataset' and 'resource' blueprints. For such cases you may need to perform
  additional check of returned value:

  >>> if toolkit.get_endpoint()[0] in ['dataset', 'package']:
  >>>     do_something()

  In this code snippet, will be called if current request is handled via
  Flask's
  dataset blueprint in CKAN>=2.9, and, in the same time, it's still working for
  Pylons package controller in CKAN<2.9 (`#4319
  <https://github.com/ckan/ckan/pull/4319>`_)
- Logic functions removed:
    ``dashboard_activity_list_html`` ``organization_activity_list_html``
    ``user_activity_list_html`` ``package_activity_list_html``
    ``group_activity_list_html`` ``organization_activity_list_html``
    ``recently_changed_packages_activity_list_html``
    ``dashboard_activity_list_html`` ``activity_detail_list`` (`#4627
  <https://github.com/ckan/ckan/pull/4627>`_)
- Extensions that add CLI commands should note the deprecation of
  ``ckan.lib.cli.CkanCommand`` and all other helpers in ckan.lib.cli.
  Extensions should instead implement CLIs using the new IClick interface.
  (`#5112 <https://github.com/ckan/ckan/pull/5112>`_)


Minor changes
-------------

- For navl schemas, the 'default' validator no longer applies the default when
  the value is False, 0, [] or {} (`#4448
  <https://github.com/ckan/ckan/pull/4448>`_)
- If you've customized the schema for package_search, you'll need to add to it
  the limiting of ``row``, as per default_package_search_schema now does.
  (`#4484 <https://github.com/ckan/ckan/pull/4484>`_)
- Several logic functions now have new upper limits to how many items can be
  returned, notably ``group_list``, ``organization_list`` when
  ``all_fields=true``, ``datastore_search`` and ``datastore_search_sql``.
  These are all configurable. (`#4562
  <https://github.com/ckan/ckan/pull/4562>`_)

```

</details>

amercader:
Looks great @smotornyuk, let's give it a go!
One thing I like about the current CHANGELOG is the summary points on top:

```
* This version requires a requirements upgrade on source installations                                                                                   
* This version requires a database upgrade                                                                                                         * This version requires a Solr schema upgrade
```

I guess we can just add them manually before the release

PR #5408, opened 2020-05-25
extras_require in setup.py

smotornyuk:
Allow installation of requirements without any additional actions using pip(fetch requirements.txt, etc).
For example:
```bash
pip install ckan[requirements]
pip install ckan[setuptools,requirements-py2]
pip install ckan[requirements,dev]
```

For testing, try to use my fork:
```
pip install -e "git+https://github.com/DataShades/ckan.git@extras-require#egg=ckan[requirements,dev]"
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -195,8 +195,9 @@ _extras_groups = [
     ('requirements', 'requirements.txt'), ('requirements-py2', 'requirements-py2.txt'),
     ('setuptools', 'requirement-setuptools.txt'), ('dev', 'dev-requirements.txt'),
 ]
+
 for group, filepath in _extras_groups:
-    with open(filepath, 'r') as f:
+    with open(os.path.join(HERE, filepath), 'r') as f:
         extras_require[group] = f.readlines()
 
 setup(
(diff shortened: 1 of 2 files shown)


wardi:
does this break later upgrading any dependencies the way `install_requires` does?

smotornyuk:
Just checked it - no, if some other package will override dependencies specified in CKAN's `extras_require`, no errors will be produced. So it behaves more like requirements file

setup.py
```diff
@@ -190,6 +190,15 @@
     ],
 }
 
+extras_require = {}
+_extras_groups = [
+    ('requirements', 'requirements.txt'), ('requirements-py2', 'requirements-py2.txt'),
+    ('setuptools', 'requirement-setuptools.txt'), ('dev', 'dev-requirements.txt'),
+]
+for group, filepath in _extras_groups:
+    with open(filepath, 'r') as f:
```
wardi, line 199: this assumes you're running setup.py from the same working directory. better to use the path of the current module to be safe

smotornyuk: Updated. I'm usually using `pip install -e ...` which implicitly changes directory to the package root before installation, so I didn't notice the problem, even though I've installed it from outside of CKAN's repo.

PR #5426, opened 2020-06-07
[#5272] User profile pictures uploads

### LINKED ISSUE #5272: Please allow to disable gravatar.com service (user privacy)
I would like to disable the gravatar.com service in my CKAN instance. This is an issue because now I have to include gravatar.com in my site's privacy policies.

2.8.3

It would be nice if I could disable gravatar via the CKAN [.ini file](https://docs.ckan.org/en/2.8/maintaining/configuration.html#ckan-gravatar-default), e.g.
```
ckan.gravatar_default = disabled
```

I cannot disable this feature. I also cannot set a user-defined image.

amercader:
Fixes #5272

This PR adds the ability to upload or link to a custom user profile picture.

Screenshot_2020-06-07 Manage - Adrià Mercader - Users - CKAN

If there is no upload provided by default we fallback to gravatar, but if desired, gravatar can be completely disabled by setting `ckan.gravatar_default = disabled`. In that case a placeholder image is shown instead. The image can be customized by overriding the `templates/user/snippets/placeholder.html` template.

Screenshot_2020-06-07 Adrià Mercader - Users - CKAN

The implementation is almost identical to the group and organization `image_url` fields, with the new field being added to the user table and model, and a `image_display_url` field being populated at dictization time with the site url if it's an upload.

On the templates there is a new `user_image()` helper that centralizes all the related logic and falls back to gravatar or the placeholder depending on the value of `ckan.gravatar_default`


amercader:
For some reason I thought this had been merged. I'd love to get this one into 2.9, @pdelboca do you think you'll have time to review this one?

pdelboca:
I had the same thought and today I saw it in my queue of PR during the meeting so I updated my ToDo. I will have it reviewed today/tomorrow morning @amercader :)

ckan/lib/helpers.py
```diff
@@ -1520,11 +1511,39 @@ def gravatar(email_hash, size=100, default=None):
         default = quote(default, safe='')
 
     return literal('''<img src="//gravatar.com/avatar/%s?s=%d&amp;d=%s"
-        class="gravatar" width="%s" height="%s" alt="Gravatar" />'''
+        class="user-image" width="%s" height="%s" alt="Gravatar" />'''
```
pdelboca, line 1514: This css class doesn't have any effect since, by specificity, is override by `.context-info .image img` (see my final comment)

amercader:
@pdelboca many thanks for the review! I believe the gravatar user profile images should now look fine after my changes. Can you re-try?
Also from your screenshots it looks like you need to rebuild the local less files to get the updated styles included in this PR. You can do so running (you might need to run `npm install` first):

```
ckan less
```
Here's how it should look:

Screenshot_2020-07-01 Activity Stream - Adrià Mercader - Users - CKAN

pdelboca:
@amercader merging master again introduced a new alembic migrations that it is pointing to the same head and it is causing an error when initializing the Database.

There are two 97 scripts both pointing to the same `down_revision`: https://github.com/ckan/ckan/tree/5272-user-pictures-uploads/ckan/migration/versions

pdelboca:
According to the docs, cd into the `ckan/migration` folder and running `alembic merge heads` fixes the conflict by creating a new migration file: https://alembic.sqlalchemy.org/en/latest/branches.html

Tried it locally and it works.

pdelboca:
Regarding the alembic issue, the changes looks good and now the gravatar image is in good quality :+1:

amercader:
All green now!
@wardi @smotornyuk can you give a final quick look and approval to this after @pdelboca review if you have a moment?

ckan/lib/helpers.py
```diff
@@ -1520,11 +1511,39 @@ def gravatar(email_hash, size=100, default=None):
...
+    gravatar_default = config.get('ckan.gravatar_default', 'identicon')
+
+    if user_dict['image_display_url']:
+        return literal('''<img src="{url}"
+                       class="user-image"
+                       width="{size}" height="{size}" alt="{alt}" />'''.format(
+            url=user_dict['image_display_url'],
+            size=size,
+            alt=user_dict['name']
+        ))
```
smotornyuk, line 1538: Not a change-request. Just want to mention that i've added `dominate` library during migration to py3 and it allows following constructions:

```
img = literal(dom_tags.img(
    src="...", cls="...", width=100, height=100
))
# or, when you have more complex nested structer, `with` blocks can be used
with dom_tags.img() as img:
    dom_tags.attr(src="...")
    dom_tags.attr(cls="...")
    dom_tags.attr(width=100, height=100)
img = literal(img)
```

smotornyuk:
Looks fine for me. +1 for merge

ckan/lib/helpers.py
amercader: Good to know thanks!

amercader:
Thanks for the review @pdelboca and @smotornyuk

PR #5431, opened 2020-06-10
Prevent password reset exposing account presence

markstuart:
This work was done in CKAN 2.7.7 by @smotornyuk 
just porting to other releases here.

### Proposed fixes:

The password reset behaviour currently behaves differently when there is no account found matching the username or email address entered in the form. This can give a malicious agent information about which accounts exist in the system and which don't.

This work makes the response to the user the same for all cases.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -461,46 +461,56 @@ class UserController(base.BaseController):
 
         if request.method == 'POST':
             id = request.params.get('user')
-
+            if id in (None, u''):
+                h.flash_error(_(u'Email is required'))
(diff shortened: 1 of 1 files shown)


markstuart:
@amercader we discussed this via email in late May, just got back to it now sorry. I also have a branch with the same commit targeting dev-v2.8 if you want me to raise a PR for that too?

smotornyuk:
I'm merging this PR - it will be backported as a part of usual patch-release process, so there is no need in creating PR into 2.8

PR #5432, opened 2020-06-10
Prevent account presence exposure when ckan.auth.public_user_details = false

markstuart:
### Proposed fixes:

This gives the access check the opportunity to fail with a NotAuthorized first. This means that when ckan.auth.public_user_details = false the auth check supercedes the NotFound, since we don't want to expose information about whether the user account exists or not.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1280,14 +1280,15 @@ def user_show(context, data_dict):
     if id:
         user_obj = model.User.get(id)
         context['user_obj'] = user_obj
-        if user_obj is None:
-            raise NotFound
     elif provided_user:
         context['user_obj'] = user_obj = provided_user
     else:
         raise NotFound
 
     _check_access('user_show', context, data_dict)
+
+    if not bool(user_obj):
+        raise NotFound
 
     # include private and draft datasets?
     requester = context.get('user')


markstuart:
This changes the behaviour so that profile lookups for non-existent accounts and existing accounts all respond consistently. However, the behaviour when an account is not found seems very odd to me.

If you navigate to `/user/non-existing-user` you get redirected to `/user/login` with a flash message. This seems unintuitive, especially if you were already logged in. It looks like this behaviour comes from https://github.com/ckan/ckan/blob/master/ckan/controllers/user.py#L84

Any ideas if that behaviour is what is still wanted/expected?

markstuart:
I also can raise PRs against the dev-v2.8 and dev-v2.7 branches whenever we're happy with this one @amercader :+1:

markstuart:
Any updates on this one @smotornyuk ?

smotornyuk:
No, I just was busy with other issues/PRs. Don't bother yourself with creating PRs into 2.8\2.7 - I'll mark this PR as backport target and it will be cherry-picked by the person working on patch-releases.

As for redirect to the login page when the user does not exist - this isn't expected behavior for me too and using normal 404 page would be much more appropriate

markstuart:
Thanks a lot @smotornyuk... I can raise a PR for the redirect to 404 page instead of login if that's a wanted change?

smotornyuk:
It would be perfect if you have a time for this. Thanks!

markstuart:
I'll try to get to it today 👍

PR #5437, opened 2020-06-11
[2.8] i18n: add test of ckan.i18n_directory config option

risicle:
This is the 2.8-branch equivalent of https://github.com/ckan/ckan/pull/5436

For me, the pylons test passes but the flask test fails.


smotornyuk:
Covered by #5500

Issue #5016, opened 2019-10-10
Problem loading datasets with IE and Edge


Aaron-M:
**Note:** Are you submitting a security related issue that could be a potential vulnerability? Please send it to security@ckan.org instead.

### CKAN Version if known (or site URL)
2.72

### Please describe the expected behaviour
Create a dataset, load a resource file (e.g. named my data file.xlsx), and a resource is created with a short URL pointing to a file called my-data-file.

### Please describe the actual behaviour
If using IE or Edge, then instead of just the filename of the resource being uploaded being used, the entire path is used, so if you have a heaviliy nested folder structure to get to your file upload, you end up with a very long URL and a meaningless resource name.

E.g. https://datastore.landcareresearch.co.nz/dataset/reports-special-reviews-of-tb-in-nz/resource/4533b4cf-1162-46c7-aab4-c9aad110244b   (look at the URL for the resource, it is a concatenation of all the subfolders truncated at max length so the actual file name is missing) - e.g. this one I manually fixed https://datastore.landcareresearch.co.nz/dataset/reports-special-reviews-of-tb-in-nz/resource/e827b2ec-389f-441b-85af-98bfe8363491

### What steps can be taken to reproduce the issue? 
Put a test file inside a very long nested folder structure.  Using IE or Edge create a dataset and upload a resource.  See what the resulting URL to the resource is.

mutantsan:
@Aaron-M Hi, I am trying to reproduce this bug, but I can't. I've tested it on 2.7.2 and 2.7.6 (current CKAN version on your portal) in Edge browser. Do you still have this issue, or it's gone? The problem, probably, was in _image-upload.js_, but it could be fixed already.

Aaron-M:
I have just tested and uploaded a file and the same behaviour is occuring (in CKAN v 2.7.6)

The file I chose to upload was  C:\Program Files\Inkscape\lib\gdk-pixbuf-2.0\2.10.0\loaders.cache.  In IE this is what results, URL points to https://datastore.landcareresearch.co.nz/dataset/b7804c9c-24b8-4258-8e18-e6c011dc37ed/resource/6635654d-5c8e-42e2-88cd-37fc0099d82e/download/cprogram-filesinkscapelibgdk-pixbuf-2.02.10.0loaders.cache

![image](https://user-images.githubusercontent.com/4842751/71856399-a7bfe980-3148-11ea-8f4b-a349c738f6ba.png)

[...]

PR #5438, opened 2020-06-12
[#5016] fix resource upload filename fetching in IE

mutantsan:
There is a problem in IE when you are trying to upload a resource with 'Include local directory path when uploading files to a server' option enabled. This little hack tries to fetch filename from this 'path' if we detect that user uses IE or Edge browser.

Also, I don't know how to write test for this code, if it's really necessary. So if someone can help me, I'll appreciate it.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/image-upload.js
+++ b/ckan/public/base/javascript/modules/image-upload.js
@@ -204,6 +204,15 @@ this.ckan.module('image-upload', function($) {
      */
     _onInputChange: function() {
       var file_name = this.input.val().split(/^C:\\fakepath\\/).pop();
+
+      // Internet Explorer 6-11 and Edge 20+
+      var isIE = false || !!document.documentMode;
+      var isEdge = !isIE && !!window.StyleMedia;
+      // for IE/Edge when 'include filepath option' is enabled
+      if (isIE || isEdge) {
+        (file_name=file_name.match(/[^\\\/]+$/)) && file_name[0];
+      }
+
       this.field_url_input.val(file_name);
       this.field_url_input.prop('readonly', true);
 
--- a/ckan/public/base/javascript/modules/image-upload.js
+++ b/ckan/public/base/javascript/modules/image-upload.js
@@ -211,7 +211,7 @@ this.ckan.module('image-upload', function($) {
       // for IE/Edge when 'include filepath option' is enabled
       if (isIE || isEdge) {
         var fName = file_name.match(/[^\\\/]+$/);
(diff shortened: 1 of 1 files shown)


ckan/public/base/javascript/modules/image-upload.js
```diff
@@ -204,6 +204,15 @@ this.ckan.module('image-upload', function($) {
      */
     _onInputChange: function() {
       var file_name = this.input.val().split(/^C:\\fakepath\\/).pop();
+
+      // Internet Explorer 6-11 and Edge 20+
+      var isIE = false || !!document.documentMode;
```
smotornyuk, line 209: You can remove `false` and the result won't change, right?

ckan/public/base/javascript/modules/image-upload.js
```diff
@@ -204,6 +204,15 @@ this.ckan.module('image-upload', function($) {
...
+      var isEdge = !isIE && !!window.StyleMedia;
+      // for IE/Edge when 'include filepath option' is enabled
+      if (isIE || isEdge) {
+        (file_name=file_name.match(/[^\\\/]+$/)) && file_name[0];
```
smotornyuk, line 213: * Second part of conjunction does nothing;
* even though match object will be converted to string bellow, in `val` method, it would be better to make things explicit and avoid potential bugs if the surrounding code will be modified.

ckan/public/base/javascript/modules/image-upload.js
```diff
@@ -206,11 +206,12 @@ this.ckan.module('image-upload', function($) {
...
 
       // Internet Explorer 6-11 and Edge 20+
-      var isIE = false || !!document.documentMode;
+      var isIE = !!document.documentMode;
       var isEdge = !isIE && !!window.StyleMedia;
       // for IE/Edge when 'include filepath option' is enabled
       if (isIE || isEdge) {
-        (file_name=file_name.match(/[^\\\/]+$/)) && file_name[0];
+        var fName = file_name.match(/[^\\\/]+$/);
+        file_name = fName ? fname[0] : file_name;
```
smotornyuk, line 214: ` fname[0]` should be ` fName[0]`

smotornyuk, line 214: fname[0] should be fName[0]

boykoc:
@smotornyuk any chance this can be backported to 2.8.x?

smotornyuk:
Sure, i'l mark it as backport for 2.6-2.8

PR #5439, opened 2020-06-12
fix readthedocs for v2.8

### LINKED ISSUE #5380: docs.ckan.org for 2.6, 2.7 and 2.8 haven't been updated since 2018
https://docs.ckan.org/en/2.8/, https://docs.ckan.org/en/2.7/ and https://docs.ckan.org/en/2.6/ should have latest docs for each version.

The docs are generated for 2.8.2, 2.7.5 and 2.6.7

OriHoch:
Fixes #5380

a bit hackish, but it works - https://orihoch-ckan.readthedocs.io/en/dev-v2.8-fix-readthedocs/changelog.html

it uses .readthedocs.yaml for the readthedocs configuration, so you may need to modify it according to the settings currently in the readthedocs advanced settings UI, see https://docs.readthedocs.io/en/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,27 @@
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: doc/conf.py
+
+# Build documentation with MkDocs
+#mkdocs:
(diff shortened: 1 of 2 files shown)


amercader:
Thanks @OriHoch! Given that these are the 2.8 docs I think the hack is fine. Going forward we probably want to look at sphinx-autoapi as suggested [here](https://github.com/readthedocs/readthedocs.org/issues/6894#issuecomment-613392787) to avoid having to install ckan to build the docs.

PR #5444, opened 2020-06-16
[#5443] unflatten allow nesting >1 level

### LINKED ISSUE #5443: Make unflatten match flatten_dict > 1 level deep
all

`unflatten(flatten_dict(x)) == x` for any `x` with more than 1 level of dict/list/dict nesting 

unflatten returns dict with numeric keys past one level instead of lists

see test in PR

wardi:
Fixes #5443

### Proposed fixes:

simplify unflatten, allowing more than one level of dict/list/dict nesting

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/navl/dictization_functions.py
+++ b/ckan/lib/navl/dictization_functions.py
@@ -410,23 +410,18 @@ def unflatten(data):
     for flattend_key in sorted(data.keys(), key=flattened_order_key):
         current_pos = unflattened
 
-        if (len(flattend_key) > 1
-                and not flattend_key[0] in convert_to_list
-                and not flattend_key[0] in unflattened):
-            convert_to_list.append(flattend_key[0])
-
         for key in flattend_key[:-1]:
             try:
                 current_pos = current_pos[key]
-            except KeyError:
+            except IndexError:
                 new_pos = {}
+                current_pos.append(new_pos)
+                current_pos = new_pos
+            except KeyError:
+                new_pos = []
                 current_pos[key] = new_pos
                 current_pos = new_pos
         current_pos[flattend_key[-1]] = data[flattend_key]
-
-    for key in convert_to_list:
-        unflattened[key] = [unflattened[key][s]
-                            for s in sorted(unflattened[key])]
 
     return unflattened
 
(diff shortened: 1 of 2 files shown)


smotornyuk:
I'm merging this PR - no problem detected. 

But I'm wondering why we are accepting only structure where dict/lists are rotated? i.e, the following is not possible:
```JSON
{
  "resources": [{
    "point": {"x": 1, "y": 2}
  }]
}

{
  "resources": [{
    "matrix": [
        [1,2,3],
        [4,5,6],
        [7,8,9]
    ]
  }]
}

```

I think, checking the type of the following key would be pretty safe - for `str` keys default container is `dict`, while for `int` key container is 'list'. And similar changes should be done on the other side, in `flatten_*` functions. Because previously such format was invalid, nothing will be broken, no one ever notices this change(ok, there is one edge case - if you are using numbers as keys for resource fields - `('resouces', 0, 0): 9` currently unflattens into `{"resources":[{0: 9}]}`, while with my suggestion it'll become `{"resources": [[9]]}`. But it creates more possibilities in schema configuration, which may be quite useful when one is implementing some existing standard, with strict distinction between single value(`{"point":{"x":1}}`) and list with a single element (`{"point":[{"x":1}]}`)

wardi:
AFAICT the dict-list-dict pattern comes from the way navl schemas are implemented. In the navl schemas a dict represents a sub-schema for a list of dicts (e.g the resource schema within the package schema) so it makes sense to continue this pattern.

We can and do support lists in lists and dicts in dicts with custom validators, it's just the navl schema won't be automatically applied to the inner objects.

I am using nested navl schema for validation of the repeating subfields I've added to ckanext-scheming because that lets me use all the existing supported field types as subfields while maintaining compatibility with their custom validators. e.g. ckanext-fluent fields work just fine as repeating subfields without modification with this change.

Thanks @smotornyuk !

EricSoroos:
This appears to have broken the datastore data dictionary form.

In ckan 2.8.4, it works, in 2.8.5, there's an off by one error where incorrect fields are edited (as compared to the form).

I think the root cause is that the datastore forms are run off of the 1 based index in the template, and this change in conversion affects that.

PR #5448, opened 2020-06-17
Fix collapsible menu in mobile view

smotornyuk:
The last security update of bootstrap exposes a bug in jQuery. it reproducible on master.ckan.org - switch to mobile layout, open dev console, and click on the button that expands header navigation.  Header won't expand and you'll see an error in console: `Cannot convert object to primitive value`.

This problem can be fixed by upgrading jQuery from 3.5.0 to 3.5.1.

In addition, I've added `package-lock.json` to git, so that we can use `npm ci` instead of `npm i` and rely on repeatable builds.


amercader:
@smotornyuk @Zharktas I need to do this same process on an existing 2.8.5 site (and while I'm at it I can do the backport to the dev-v2.7 and dev-2.8 branches). How should I go about it? Is it a case of manually updating all the individual files? Or did you use an npm command or similar?
Also I see that this PR includes changes in the minified versions of Bootstrap but not the un-minified ones, is that expected?

Zharktas:
@amercader I've been doing the same, just haven't had the time to do the backporting. Manually updating ckan/public/base/vendor/jquery.js and running minification for it is enough as fanstatic serves the minified version when running in non-debug mode. Npm cannot be used as ckan 2.7 and 2.8 doesn't have package.json. 

This PR updates also other files as there was no lock file, and patch updates were installed. Now that lock file is in version control, they libraries don't update as often.

PR #5449, opened 2020-06-17
Do not ignore ckan.i18n_directory in flask

smotornyuk:
Fixes #5436 and  #5437

There is one controversial point in this PR. Looking at [docs of ckan.i18n_directory](https://docs.ckan.org/en/2.8/maintaining/configuration.html?highlight=ckan.i18n_directory#ckan-i18n-directory) i would expect this option to contain path including `i18n` dir itself. I.e, with default CKAN installation, following option in config file should replicate default behavior(nothing should be changed):
```
ckan.i18n_directory = /user/lib/ckan/default/src/ckan/ckan/i18n
```

But, current implementation expects that last `i18n` fragment is implied. I.e, currently, if you want to set default CKAN translations directory via `ckan.i18n_directory`, you should put following in the config file
```
ckan.i18n_directory = /user/lib/ckan/default/src/ckan/ckan
```

It seems really incorrect for me because option name and documentation sound like this it should specify the path to `i18n` directory itself, not **the path to the folder which contains i18n directory**.


risicle:
Testing this now - I would suggest _not_ including the `i18n/` change in any _backports_ though as people won't expect to have to update their configurations for stable updates.

risicle:
WFM :+1:

smotornyuk:
Yes, I also thought about this, so I've added extra check:

1. If there is `ckan.pot` file inside the specified translation directory(new style) - use this directory as translations folder
2. Otherwise, search for `i18n` folder inside the specified directory(old behavior)

wardi:
IIUC the ckan.pot file isn't strictly necessary for translations to work, you just need the language directories and .mo files. What do you think of checking for the i18n directory instead of the ckan.pot file? I can't think of a reason to have an i18n directory at the same level as the language directories so it should be safe.

smotornyuk:
Sounds reasonable - i'll update it

amercader:
Thanks for doing the compatibility changes @smotornyuk I think this is safe for backporting

PR #5453, opened 2020-06-18
[#5452] allow lists in resource extras

### LINKED ISSUE #5452: lists not allowed as resource extras
all

Resource extras are very liberal, almost any key may be used and complex nested JSON is supported as values natively.

Resource extras with almost any JSON type are supported, but list values are silently removed .

Pass an extra with a list value to the `resource_create` API.

wardi:
Fixes #5452 

### Proposed fixes:
remove two lines of code from resource_dict_save. This allows ckanext-scheming and other extensions to have repeating subfields and other list values in their resource schemas without hacks like converting to/from JSON (resource extras are already stored in the DB as JSON)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -907,7 +907,8 @@ def json_object(value):
 def extras_valid_json(extras, context):
     try:
         for extra, value in iteritems(extras):
-            json.dumps(extra)
+            json.dumps(value)
     except ValueError as e:
-        raise Invalid(_('Could not parse the value as valid JSON'))
+        raise Invalid(_(u'Could not parse extra \'{name}\' as valid JSON').
+                format(name=extra))
     return extras
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -9,7 +9,7 @@ import mimetypes
 import string
 import json
 
-from six import string_types
+from six import string_types, iteritems
 from six.moves.urllib.parse import urlparse
 
 import ckan.lib.navl.dictization_functions as df
@@ -902,3 +902,12 @@ def json_object(value):
         raise Invalid(_('Could not parse the value as a valid JSON object'))
 
     return value
+
+
+def extras_valid_json(extras, context):
+    try:
+        for extra, value in iteritems(extras):
+            json.dumps(extra)
+    except ValueError as e:
+        raise Invalid(_('Could not parse the value as valid JSON'))
+    return extras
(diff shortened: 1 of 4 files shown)


wardi:
One more small change was added here. In current versions of ckan dicts passed as resource extras were being stored as their python repr, e.g. `{u'mykey': u'myvalue'}` even though the lower level was JsonDict was able to store the actual JSON. I've added an `extras_valid_json` validator instead of the validator that was turning objects into their python repr.

wardi:
| resource extra value (JSON) | returned value before | returned value after |
| --- | --- | --- |
| `"text"` | `"text"` :heavy_check_mark:  | `"text"` :heavy_check_mark: |
| `42` | `"42"` :x: | `42` :heavy_check_mark: |
| `{"test": "value"}` | `"{u'test': u'value'}"` :x: | `{"test": "value"}` :heavy_check_mark: |
| `true` | `"True"` :x: | `true` :heavy_check_mark: |
| `null` | `"None"` :x: | `null` :heavy_check_mark: |
| `[1, 2, 3]` | (silently removed) :x: | `[1, 2, 3]` :heavy_check_mark: |

I think this can be considered a bug fix, the current behaviour is very strange for a JSON API

ckan/logic/validators.py
```diff
@@ -902,3 +902,12 @@ def json_object(value):
         raise Invalid(_('Could not parse the value as a valid JSON object'))
 
     return value
+
+
+def extras_valid_json(extras, context):
+    try:
+        for extra, value in iteritems(extras):
+            json.dumps(extra)
```
smotornyuk, line 910: I didn't get it. As i see this construction, for dictionary `{"a": [1,2], "b": {"x": 1}}`, you are converting key to json string without assigning value:
```
extra, value = ("a", [1,2])
json.dumps(extra) == json.dumps("a") == '"a"'

extra, value = ("b", {"x": 1})
json.dumps(extra) == json.dumps("b") == '"b"'
```

 Is `ValueError` ever possible here?

wardi: ValueError is possible from within ckan or an extension calling the action with something not json compatible, or if you upload a file as an extra field.

wardi: oh, I see, I should be dumping value not extra, will fix

amercader:
@wardi I had to tweak this a bit to adapt it to the old way validators where handled in the schema in 2.7, do you mind having an extra quick look at these to make sure nothing seems off? thanks

https://github.com/ckan/ckan/compare/85656ba458a34e74ddc172a7db150827329ca2c1...8fc5ceb378b858699487184ec6f0fcde8bc56da4

wardi:
LGTM

PR #5464, opened 2020-06-24
Always 404 on non-existing user lookup

markstuart:
Moved specific redirect to login when accessing dashboard into the action for dashboard.
See #3224 for original work around this.

### Proposed fixes:

This is roughly as discussed in #5432 @smotornyuk , though I found that the redirect to login had been implemented intentionally in #3224. I believe that my implementation here solves both problems:
- redirect to login if anonymous/unauthenticated user accesses `/dashboard` or `/user/me`
- display 404 page for `/user/non-existing-user`

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/controllers/user.py
+++ b/ckan/controllers/user.py
@@ -80,8 +80,7 @@ class UserController(base.BaseController):
         try:
             user_dict = get_action('user_show')(context, data_dict)
         except NotFound:
-            h.flash_error(_('Not authorized to see this page'))
-            h.redirect_to(controller='user', action='login')
+            abort(404, _('User not found'))
         except NotAuthorized:
             abort(403, _('Not authorized to see this page'))
 
@@ -147,10 +146,6 @@ class UserController(base.BaseController):
         return render('user/read.html')
 
     def me(self, locale=None):
-        if not c.user:
-            h.redirect_to(locale=locale, controller='user', action='login',
-                          id=None)
-        user_ref = c.userobj.get_reference_preferred_for_uri()
         h.redirect_to(locale=locale, controller='user', action='dashboard')
 
     def register(self, data=None, errors=None, error_summary=None):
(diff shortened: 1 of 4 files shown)


markstuart:
I'm kinda expecting tests to fail, but it'd be great to get feedback on the approach, and I'll fix up the tests soon.

smotornyuk:
These changes also have to be done inside [users view](https://github.com/ckan/ckan/blob/master/ckan/views/user.py). In CKAN 2.9 UserController(and all the other controllers) was finally rewritten as Flask view. As result, we have [controllers](https://github.com/ckan/ckan/tree/master/ckan/controllers) for compatibility with existing extensions(in case someone extends controller and registers new routes)  and [views](https://github.com/ckan/ckan/tree/master/ckan/views) which now handle all the routes available out-of-the-box. So your fix will work for 2.8 and prior versions but will be ignored in the upcoming 2.9 release.

markstuart:
OK, great info, thanks. I'll look into what's need to mirror these changes into the users view, and possibly the dashboard view.

markstuart:
Hopefully this is what you were looking for @smotornyuk ?

smotornyuk:
Yes, exactly. And only two failed tests - thats quite a good change

markstuart:
I'll have to pick up the test fixes tomorrow, thanks for the feedback 👍

markstuart:
Alright, I *think* we're all good here @smotornyuk, let me know if you need anything else :smiley:

markstuart:
I know you said previously not to worry about contributing PR's for backport to other 2.8/2.7 @smotornyuk, but this one feels a little different as only some of the work here is applicable to those releases (specifically not all the change in the views directory).

If you want, I am happy to raise PR's against `dev-v2.7` and `dev-v2.8`

smotornyuk:
Yep, great. I'm merging it. 
As for backports, it become a bit more tricky with these view changes, so yes, please, add PR into dev-2.{7,8} branches when you have time

Issue #5451, opened 2020-06-18
Toolbar dataset type naming is inconsistent


frafra:
### CKAN Version if known (or site URL)
2.8.4

### Please describe the expected behaviour
`/custom-dataset` and `/custom-dataset/new` toolbar should refer to the dataset type.

### Please describe the actual behaviour
They don't. The first URL shows the custom dataset name, while the latter uses "Dataset" as prefix.

### What steps can be taken to reproduce the issue? 
Try to add a custom dataset.

frafra:
https://github.com/ckan/ckan/blob/7f11188b106a428f41540cc9394dbc6a763ba5b0/ckan/templates/package/base.html#L23-L24

smotornyuk:
Hi, @frafra . It's quite a major change because there are a lot of places that have dataset type hardcoded. In addition, you can observe the same problem in URLs(i.e, it's `/dataset/edit/...` instead of `/protocol/edit/...`). It will be solved in 2.9 - there is one merged PR(#5189) and one active, created for this issue(#5466). But I don't think it will be backported to prior versions, because it requires not only template changes, but updating controllers as well in a way, which may cause incompatibility issues with existing extensions.

PR #5466, opened 2020-06-24
[#5451] incorrect dataset type in some labels

smotornyuk:
Fix a few labels missed in the previous pr(#5189) and add records for towncrier

### DIFF
--- a/ckan/templates/package/base.html
+++ b/ckan/templates/package/base.html
@@ -21,6 +21,6 @@
     <li{{ self.breadcrumb_content_selected() }}>{% link_for dataset|truncate(30), named_route=pkg.type ~ '.read', id=pkg.id if is_activity_archive else pkg.name %}</li>
   {% else %}
     <li>{% link_for _(dataset_type.title()), named_route=dataset_type ~ '.search' %}</li>
-    <li class="active"><a href="">{{ _('Create Dataset') }}</a></li>
+    <li class="active"><a href="">{{ h.humanize_entity_type('package', dataset_type, 'create label') or _('Create Dataset') }}</a></li>
   {% endif %}
 {% endblock %}
--- a/ckan/templates/package/snippets/stages.html
+++ b/ckan/templates/package/snippets/stages.html
@@ -14,15 +14,17 @@ Example:
 #}
 {% set s1 = stages[0] or 'active' %}
 {% set s2 = stages[1] or 'uncomplete' %}
+{% set dataset_type = dataset_type or h.default_package_type() %}
+
 {% if s1 != 'uncomplete' %}{% set class = 'stage-1' %}{% endif %}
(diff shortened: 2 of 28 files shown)


changes/5189.misc
```diff
@@ -0,0 +1,4 @@
+Dataset type is used when building URLs(eg. `/showcase/edit/NAME`
+intead of `/dataset/edig/NAME` when `dataset['type']=='showcase'`)
```
wardi, line 2:
```suggestion
intead of `/dataset/edit/dataset` when `dataset['type']=='showcase'`)
```

ckan/templates/package/base.html
```diff
@@ -21,6 +21,6 @@
     <li{{ self.breadcrumb_content_selected() }}>{% link_for dataset|truncate(30), named_route=pkg.type ~ '.read', id=pkg.id if is_activity_archive else pkg.name %}</li>
   {% else %}
     <li>{% link_for _(dataset_type.title()), named_route=dataset_type ~ '.search' %}</li>
-    <li class="active"><a href="">{{ _('Create Dataset') }}</a></li>
+    <li class="active"><a href="">{{ _('Create ' ~ dataset_type.title()) }}</a></li>
```
wardi, line 24: this kind of string can't be picked up by `extract_messages` but I guess we've done the same in the earlier PR.

ckan/templates/package/snippets/stages.html
```diff
@@ -20,9 +20,9 @@
 <ol class="stages {{ class }}">
   <li class="first {{ s1 }}">
     {% if s1 != 'complete' %}
-      <span class="highlight">{{ _('Create dataset') }}</span>
+      <span class="highlight">{{ _('Create ' ~ dataset_type) }}</span>
```
wardi, line 23: You used `.title()` above, what's the difference here?

ckan/templates/package/base.html
wardi: See my suggestions in #5337 I think we should do the same here

ckan/templates/package/snippets/stages.html
smotornyuk: I'm just trying to leave labels as they were before. In previous example **Dataset** was originally in title-case and here it was in lower-case before my change

smotornyuk:
Hi, @wardi 

This PR is finished and can be reviewed. It does exactly the same things as #5337, but for datasets

wardi:
@smotornyuk is there a front-end test that needs updating?

smotornyuk:
Nope. CircleCI is failing because of some updates to the accessibility guide. Basically, it thinks that the contrast between text and background is not big enough:)
This test fails periodically on all the PRs and I think that @Zharktas has recently created fix for it

PR #5477, opened 2020-06-30
[#5259] allow returning response from IAuthenticator.identify() for f…

### LINKED ISSUE #5259: Allow IAuthtenticator methods to modify the response
The Pylons standard way of modifying the outgoing response (eg adding cookies, issuing redirects etc) is to either edit a global `response` object or call global functions like `redirect_to`.

Flask views on the contrary always expect a response-like object to be returned so if you want to modify it, it needs to be in the context of that view function. Take for instance the `login` function of the `user` view:

```python
  def logout():
      # Do any plugin logout stuff
      for item in plugins.PluginImplementations(plugins.IAuthenticator):
          item.logout()
      url = h.url_for(u'user.logged_out_page')
      return h.redirect_to(
          _get_repoze_handler(u'logout_handler_path') + u'?came_from=' + url,
          parse_url=True)
```
It's quite common for `IAuthenticator.login()` or `IAuthenticator.logout()` to need to clear cookies, issue redirects etc. In Pylons, plugins could just do `toolkit.response.cookies = ...` or `toolkit.redirect_to()` and the web response would be immediately affected.

On Flask though the current code just calls the plugin methods and carries on with the normal view function execution, so extensions can't do any of these actions.

One potential way I've solved it for this particular case is checking if extensions are returning a response object, and if so just returning this directly on the view function, as it assumes that the extension has done all that's needed.

Happy to hear alternatives.

pazepaze:
Fixes #5259 

### Proposed fixes:

Allow returning response from IAuthenticator.identify() for flask. This allows plugins to send e.g. a redirect like they can do with pylons.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -359,7 +359,7 @@ def ckan_before_request():
 
     # Identify the user from the repoze cookie or the API header
     # Sets g.user and g.userobj
-    identify_user()
+    response = identify_user()
 
     # Provide g.controller and g.action for backward compatibility
     # with extensions
@@ -367,6 +367,8 @@ def ckan_before_request():
 
     set_ckan_current_url(request.environ)
     g.__timer = time.time()
+
+    return response
 
 
 def ckan_after_request(response):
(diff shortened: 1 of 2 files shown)


amercader:
Closing this in favour of the more comprehensive https://github.com/ckan/ckan/pull/5640. @pazepaze would be great if you could test the new PR and see if it works for you. Cheers.

Issue #5468, opened 2020-06-25
Page view tracking of datasets is not working if ckan is running at a subpath


wrinklenose:
### CKAN Version if known (or site URL)
2.8.4

### Please describe the expected behaviour
Page view tracking works as describend in https://docs.ckan.org/en/2.8/maintaining/tracking.html as well if ckan is operated at subpath.

### Please describe the actual behaviour

Page view tracking works as expected if ckan runs at standard root path "/".
But we're running ckan on a subpath ('/ckan') using the ckan.root_path property in the production.ini.
In this case the page view tacking of datasets **always shows zero hits**.

After some investigation into the db, I figured out that the hits are successfully reflected in table 'tracking_raw', but not updated during a 'paster --plugin=ckan tracking update' in table 'tracking_summary'.

I further noticed, that tracking_summary.package_id keeps the value '\~\~not\~found\~\~' for all datasets at this system. (Unlike to a system running on '/' where correct package-ids are set).

When I compared the stored urls of both systems, i also noticed that the leading url path is reflected in the stored url. So you'll get:
' /ckan/dataset/test' or ' /ckan/de/dataset/test'(both not working) im comparision to ' /dataset/test' or ' /de/dataset/test' (both working).

I've tried to track the problem in the code. For my understanding the problem could be located in **/cli/tracking.py** - function **update_tracking**.
Below the comment "get ids for dataset urls" happens the construction of a sql, using a regex.

```
# get ids for dataset urls
sql = u'''UPDATE tracking_summary t
            SET package_id = COALESCE(
                (SELECT id FROM package p
                WHERE p.name = regexp_replace
                  (' ' || t.url, '^[ ]{1}(/\\w{2}){0,1}' || %s, ''))
                ,'~~not~found~~')
            WHERE t.package_id IS NULL
            AND tracking_type = 'page';'''
engine.execute(sql, PACKAGE_URL)
```

This regex does not reflect a leading path part ('/ckan' in our example), which leads to an failing UPDATE statement.

### What steps can be taken to reproduce the issue? 
- configure ckan under a sub path like mentioned in the docs https://docs.ckan.org/en/2.8/maintaining/configuration.html#ckan-root-path
- enable page view tracking like mentioned in the docs https://docs.ckan.org/en/2.8/maintaining/tracking.html
- create a dateset and visit it
- update the page view tracking statistic like mentioned in the docs https://docs.ckan.org/en/2.8/maintaining/tracking.html (I did it manually running the paster command)
[...]

PR #5483, opened 2020-07-02
[#5468] Fixed page view tracking of datasets when running ckan on sub…

wrinklenose:
…path

Used the configuration value 'ckan.root_path' to expand the existing regular expression to strip the root path when matching package names updating the tracking_summary.

Fixes #

### Proposed fixes:

Page view tracking of datasets is now working also when running ckan on a subpath url.

### Features:

- [ ] includes API changes

### Remarks:
We're running ckan 2.8.4 as mentioned before. We needed a fix for this version, so I made the shown changes originally in lib/cly.py of version 2.8.4. The affected code seems to be moved to cli/tracking.py in actual master but seemed to be unchanged in other ways, so I just copied the original code to the new location.
I can promise the changed code works on lib/cly.py of 2.8.4 but haven't run any tests for the actual master version.

### DIFF
--- a/ckan/cli/tracking.py
+++ b/ckan/cli/tracking.py
@@ -4,9 +4,11 @@ import ckan.model as model
 import click
 import datetime
 import csv
+import re
 import ckan.logic as logic
 from collections import namedtuple
 from ckan.cli import error_shout
+from ckan.common import config
 
 _ViewCount = namedtuple(u'ViewCount', u'id name count')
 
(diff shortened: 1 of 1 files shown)


wardi:
instead of including the root path in the regex, can we remove it from the urls stored in the first place? That should be a simpler and would work even if a root path changes on a live site.

wrinklenose:
I agree, storing the urls without the subpath would be a more flexible and easier way.

But I think we have to balance between flexibility and potential use of the accesibility of additional information in future.
Another question of this approach would be: Should the clearing of the supath already happen in tracking_raw (e. g. for consistency) or only in the tracking_summary (e. g. to keep the original information available for potential future use).
Are there any side effects changing those urls?

I'm sorry to say that, but I'm not sure if I can effort any further progress in this issue in the next weeks due to other work load. So if anyone else likes to take over, feel free to do so...

pdelboca:

PR #5486, opened 2020-07-02
Test for multiple heads in migrations

pdelboca:
## Proposed fixes:

This PR adds a test to check that there aren't multiple heads in the migration files that may cause errors or undesired results on deployment. Having multiple heads is common when working with branching since merging `master` in our branch can silently bring another migration script that will cause a conflict in heads.

If two heads are in the migration folder when running the alembic migrations it will fail with:
```
alembic.util.exc.CommandError: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '@head' to narrow to a specific head, or 'heads' for all heads
```

This test will raise with:
```
alembic.util.exc.CommandError: The script directory has multiple heads (due to branching).Please use get_heads(), or merge the branches using alembic merge.
```

Solution based on this [blogpost](https://blog.jerrycodes.com/multiple-heads-in-alembic-migrations/) given some issues we had when reviewing #5426 

### Features:

- [ ] includes API changes


amercader:
Update: make the `ckan db init` and `ckan db upgrade` commands fail if we have multiple alembic heads

amercader:
No need to backport this as there will be no new migrations in the 2.9 branch

PR #5488, opened 2020-07-06
[#5146] Make creation of API keys configurable

amercader:
Follow up from #5146. This allows site admins to disable the creation of API keys by default on all users

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -479,7 +479,7 @@ CONFIG_PERMISSIONS_DEFAULTS = {
     'allow_dataset_collaborators': False,
     'allow_admin_collaborators': False,
     'allow_collaborators_to_change_owner_org': False,
-    'create_default_api_keys': True,
+    'create_default_api_keys': False,
 }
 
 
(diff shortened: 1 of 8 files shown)


amercader:
@wardi @smotornyuk I implemented defaulting to True as mentioned in #5146 for now, but I think API keys creation by default should be turned off by default on 2.9:
* API tokens are obviously superior from a security and functional point of view, so we want to encourage *new* users to use them
* Any existing API key will keep working as currently, as we are not touching these, so we are not breaking any existing script.
* If set to False (ie no API key in new users) and for some reason someone really needs an API key instead of an API token they can use the "Regenerate API key" button to create a new one.

The only scenario I can think of where this might break something is a system where the API key of newly created users is automatically used on some external script. In that case maintainers only need to set `ckan.auth.create_default_api_keys=True` to keep the current behavior.

I'm happy to be convinced otherwise though in case I'm missing something

wardi:
+1 for defaulting api key generation off for 2.9

smotornyuk:
Is it worth adding some kind of `--with-api-key` flag to `ckan user add` CLI command?

And, as an idea for separate PR, we probably need `token add/revoke/list` subcommands under `ckan user` group.

amercader:
@smotornyuk I think these are good ideas for another PR

PR #5495, opened 2020-07-09
WIP Respect ckan.search.show_all_types config setting

chris48s:
### Proposed fixes:

Respect `ckan.search.show_all_types` config setting

### Features:

- [ ] includes API changes

---

ping @amercader
refs https://github.com/okfn/ckanext-unhcr/issues/292

I have some follow up questions about this inline.

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -230,7 +230,7 @@ class PackageController(base.BaseController):
             # types any search page. Potential alternatives are do show them
             # on the default search page (dataset) or on one other search page
             search_all_type = config.get(
-                'ckan.search.show_all_types', 'dataset')
+                'ckan.search.show_all_types', False)
             search_all = False
 
             try:
--- a/ckan/tests/controllers/test_package.py
+++ b/ckan/tests/controllers/test_package.py
@@ -1430,10 +1430,13 @@ class TestResourceDelete(object):
         assert 200 == response.status_code
 
 
-@pytest.mark.usefixtures("clean_db", "clean_index", "with_request_context")
+@pytest.mark.ckan_config("ckan.plugins", u"example_idatasetform_v5")
+@pytest.mark.usefixtures(
+    "clean_db", "clean_index", "with_plugins", "with_request_context"
+)
 class TestSearch(object):
     def test_search_basic(self, app):
-        dataset1 = factories.Dataset()
+        dataset1 = factories.Dataset(type="dataset")
 
         offset = url_for("dataset.search")
         page = app.get(offset)
@@ -1441,7 +1444,7 @@ class TestSearch(object):
         assert helpers.body_contains(page, dataset1["name"])
 
     def test_search_language_toggle(self, app):
-        dataset1 = factories.Dataset()
+        dataset1 = factories.Dataset(type="dataset")
 
         with app.flask_app.test_request_context():
             offset = url_for("dataset.search", q=dataset1["name"])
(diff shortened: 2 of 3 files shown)


ckan/controllers/package.py
```diff
@@ -230,7 +230,7 @@ def pager_url(q=None, page=None):
             # types any search page. Potential alternatives are do show them
             # on the default search page (dataset) or on one other search page
             search_all_type = config.get(
-                'ckan.search.show_all_types', 'dataset')
+                'ckan.search.show_all_types', False)
```
chris48s, line 233: I've applied the same change here in the `ckan/controllers/package.py` controller as I did in the blueprint in `ckan/ckan/views/dataset.py`. The blueprint is the the one that actually gets called under test. I'm not sure about this one. Is it just to make back-porting to 2.8 easier? Is there some circumstance in 2.9 where the old controller still gets called?

ckan/tests/controllers/test_package.py
```diff
@@ -1503,9 +1506,15 @@ def test_search_plugin_hooks(self, app):
...
-        factories.Dataset(name="dataset-three", title="Dataset Three")
+        factories.Dataset(
+            name="dataset-one", title="Dataset One", type="dataset"
+        )
+        factories.Dataset(
+            name="dataset-two", title="Dataset Two", type="dataset"
+        )
+        factories.Dataset(
+            name="dataset-three", title="Dataset Three", type="dataset"
+        )
```
chris48s, line 1517: In order to get the tests to pass after I made this change, I have to explicitly set type="dataset", on every dataset we create in this test suite. Would it be better to just move this to the base implementation of `factories.Dataset`?

ckan/tests/controllers/test_package.py
```diff
@@ -1752,6 +1799,115 @@ def test_search_with_extra_params(self, app, monkeypatch):
...
+        search_url = url_for("dataset.search")
+        search_response = app.get(search_url)
+
+        search_response_html = BeautifulSoup(search_response.data)
+        ds_titles = search_response_html.select(
+            ".dataset-list " ".dataset-item " ".dataset-heading a"
+        )
+        ds_titles = [n.string for n in ds_titles]
+
+        assert False
```
chris48s, line 1909: TODO: what should the assertion be here?
I read https://docs.ckan.org/en/2.8/maintaining/configuration.html#ckan-search-show-all-types but I still don't really understand what the desired behaviour is in this situation, or whether the code as it stands is doing it.

ckan/controllers/package.py
amercader: Core doesn't use these controllers at all, and in fact they will be removed soon, but we might backport this to 2.8 so it will make things easier to have this change here

ckan/tests/controllers/test_package.py
amercader: Yes please, add a `type = dataset` property on the Dataset factory

ckan/tests/controllers/test_package.py
amercader: These docs are terrible. I think what they are trying to say is that when this config setting is "fancy_type" if you go to the "/fancy_type" page (custom dataset type page like /harvest or /showcase) you should get all dataset types listed.

chris48s: OK. That isn't what happens now if I set `ckan.search.show_all_types = "fancy_type"` I'm also now unclear on what `/fancy_type` should show if `ckan.search.show_all_types = False` and what should be there if `ckan.search.show_all_types = True` ?

Lets ignore what the code does at the moment and run over each case to work out what I'm aiming for. I think probably the **intention** is:

* `ckan.search.show_all_types = False` (or not set):
    * `/datasets`: show only `type=dataset`
    * `/fancy_type`: show only `type=fancy_type`
* `ckan.search.show_all_types = True`:
    * `/datasets`: show both `type=dataset` and `type=fancy_type`
    * `/fancy_type`: show only `type=fancy_type` (or should this also show both `type=dataset` and `type=fancy_type` in this case?)
* `ckan.search.show_all_types = "fancy_type"`:
    * `/datasets`: show only `type=dataset`
    * `/fancy_type`: show both `type=dataset` and `type=fancy_type`

does that seem like what we should be aiming for, or am I wrong?

PR #5504, opened 2020-07-16
Action for current user to get id or details #5490

### LINKED ISSUE #5490: There is no way to determine the user ID when connecting through the API
**CKAN version**
2.8

**Describe the bug**
It is not possible to determine a users ID when connecting through the API. In my opinion, this should be the default behavior of `user_show`, but the documentation states that *either the `id` or the `user_obj` parameter must be given*. I think it would be user-convenient to change this default behavior.

**Steps to reproduce**
Connect via API using an API key and call `user_show` without arguments.

**Expected behavior**
The information about the currently logged-in user is shown.

**Actual behavior**
```json
{"help": "https://dmoain.tld/api/3/action/help_show?name=user_show", "success": false, "error": {"message": "Not found", "__type": "Not Found Error"}}
```

shubham-mahajan:
Fixes #5490

### Proposed fixes:
New Action that will return the id of the current logged in user or by API-KEY and can return details if `include_details` is passed explicitly.

### Steps to Test:-
http://{host}:5000/api/3/action/current_user_show - **Without  API-Key/Logged In (Should give Authorization Error)**
http://{host}:5000/api/3/action/current_user_show - **API-KEY/Logged In (Should return only id of user)**
http://{host}:5000/api/3/action/current_user_show?include_details=true - **API-KEY/Logged In (Should return details of user)**

### Features:

- [x] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1488,9 +1488,9 @@ def current_user_show(context, data_dict):
     model = context['model']
     # Get the current user that is making the request
     # raise NotAuthorized if the requestor is not found
-    requester = context.get('user')
-    if requester:
-        user_obj = model.User.get(requester)
+    requestor = context.get('user')
+    if requestor:
+        user_obj = model.User.get(requestor)
         context['user_obj'] = user_obj
     else:
         raise NotAuthorized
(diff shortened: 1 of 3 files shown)


wardi:
I don't agree with having a second API call with duplicated code for returning user details.

If you need a single call like this then then let's add an optional `current_user` parameter to user_show that returns the current user details instead of making a slightly different copy of the user_show code.

PR #5514, opened 2020-07-28
Fix for Issue5512

### LINKED ISSUE #5512: Password validator does not include validation for at least 1 upper case, 1 lower case, 1 number and 1 special character.
The built in "user_password_validator" function does not validate 1 uppercase, 1 lowercase, 1 digit and 1 special character. If customized password validator is applied then the "user_invite" function crashes with "Internal Sever Error".  

The "user_invite" function does not generate the password with at least 1 special character.

**CKAN version**
   2.8.4

**Actual behavior**
Existing password validator validates the password to be string and must be 8 character or long.

Gauravp-NEC:
Fixes #5512 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1136,9 +1136,15 @@ def user_invite(context, data_dict):
     # Choose a password. However it will not be used - the invitee will not be
     # told it - they will need to reset it
     while True:
-        password = ''.join(random.SystemRandom().choice(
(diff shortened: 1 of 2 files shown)


wardi:
We're not going to change ckan's default password validator, this can be done from an IValidators plugin.

For generating new passwords we should make passwords optional (invited users have no password) instead of hoping that a particular random password will pass whatever validation rules are configured on a particular ckan site.

PR #5515, opened 2020-07-28
Set host based on site_url on middleware

amercader:
The recently introduced HostHeaderMiddleware used to remove the `HTTP_HOST` key completely. But that breaks setups where CKAN (uWSGI) is sitting behind a reverse proxy like Nginx, as the redirect is then issued to the internal CKAN host (eg localhost:8080)

### DIFF
--- a/ckan/config/middleware/common_middleware.py
+++ b/ckan/config/middleware/common_middleware.py
@@ -5,11 +5,12 @@ import hashlib
 import cgi
 
 import six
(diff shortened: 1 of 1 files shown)


amercader:
Props to @shubham-mahajan who correctly suggested this change first

Zharktas:
Tests passed by rerunning them, so merging this.

PR #5520, opened 2020-07-28
DataPusher fixes for 2.9

amercader:
Two fixes:

* Replace py2-only syntax for exception handling
* Use site user API key to send DataPusher job

The second one is the one worth discussing. As we recently did changes in the way API keys are created in #5488, by default users don't have an API key created so services can no longer rely on users having one.
We can think of a more generic solution for this (like creating a token for them) but for now it makes sense to use the site user API key, which as of now always has an API key. I checked and Xloader uses the same approach.


amercader:
@ckan/core a review on this one would be appreciated if someone has a spare moment

PR #5521, opened 2020-07-29
Conditianally allow missing config in CLI

smotornyuk:
Changes a bit way of reporting errors from #5516. Instead of using `exit` immediately, throw an exception, which will be caught inside the decorator, that provides config object for CLI commands. Thus, two problems will be solved:

* we can potentially write tests for CLI and intentionally skip config file. Using `exit` interrupts test while raising an exception naturally works as a part of a test workflow
* `config-tool` and `generate` branches work without `--config` parameter.


amercader:
Thanks @smotornyuk , I should have tested #5516 properly. I fixed a failing test and coding standards on 711ae284b

PR #5523, opened 2020-07-29
Accessibility enhancement

bblwq:
### Issues
1. Fixed the __Broken skip link__ on the landing page. The issue was identified using WAVE Chrome extension (as shown below).
image

1. Fixed the __Multiple form label__ error on the metadata edit page. The issue was identified using WAVE Chrome extension (as shown below).
image

1. Fixed the __Missing form label__ error on the dashboard page. The issue was identified using WAVE Chrome extension (as shown below).
image

1. [__WIP__] Empty link error to remove filter(s) on the dataset search result page.
image

### Proposed fixes:
1. Added tag id to the default layouts.
1. Removed the redundant label for the custom field key.
1. Added input label to the Followee search field on the dashboard page.
1. Added "Remove Filter" aria-label.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/home/index.html
+++ b/ckan/templates/home/index.html
@@ -8,7 +8,7 @@
 
 {% block content %}
   <div class="homepage layout-{{ homepage_style }}">
-    <div class="container">
+    <div id="content" class="container">
       {{ self.flash() }}
     </div>
     {% block primary_content %}
--- a/ckan/templates/user/snippets/followee_dropdown.html
+++ b/ckan/templates/user/snippets/followee_dropdown.html
@@ -20,8 +20,7 @@
     <div class="popover-header">
       <div class="input-group">
         <span class="input-group-addon"><i class="fa fa-search"></i></span>
-        <label for="field-followee-search" style="display: none;">Search Followees</label>
-        <input id="field-followee-search" type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off">
+        <input type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off" aria-label="Search followee">
       </div>
     </div>
     {% if followees %}
--- a/ckan/templates/user/snippets/followee_dropdown.html
+++ b/ckan/templates/user/snippets/followee_dropdown.html
@@ -20,7 +20,8 @@
     <div class="popover-header">
       <div class="input-group">
         <span class="input-group-addon"><i class="fa fa-search"></i></span>
-        <input type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off">
+        <label for="field-followee-search" style="display: none;">Search Followees</label>
+        <input id="field-followee-search" type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off">
       </div>
     </div>
     {% if followees %}
--- a/ckan/templates/user/snippets/followee_dropdown.html
+++ b/ckan/templates/user/snippets/followee_dropdown.html
@@ -20,7 +20,7 @@
     <div class="popover-header">
       <div class="input-group">
         <span class="input-group-addon"><i class="fa fa-search"></i></span>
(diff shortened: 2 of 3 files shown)


ckan/templates/user/snippets/followee_dropdown.html
```diff
@@ -20,7 +20,8 @@
     <div class="popover-header">
       <div class="input-group">
         <span class="input-group-addon"><i class="fa fa-search"></i></span>
-        <input type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off">
+        <label for="field-followee-search" style="display: none;">Search Followees</label>
```
Zharktas, line 23: `display: none;` is not recommended as it hides the element also from screen readers [ref](https://accessibility.18f.gov/hidden-content/). Either use .sr-only class which is available in [bootstrap](https://getbootstrap.com/docs/3.4/components/#callout-inputgroup-form-labels) or add aria-label to the input field, which I would recommend as it is largely used in the current codebase.

ckan/templates/user/snippets/followee_dropdown.html
```diff
@@ -20,7 +20,7 @@
     <div class="popover-header">
       <div class="input-group">
         <span class="input-group-addon"><i class="fa fa-search"></i></span>
-        <input type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off">
+        <input type="text" class="form-control" name="q" placeholder="{{ _('Search list...') }}" value="{{context.q}}" autocomplete="off" aria-label="Search followee">
```
Zharktas, line 23: One small issue I didn't notice before, could you add translation tags to the aria label, eg. _('Search followee') so it would be translated too if the user uses some other language than english.

bblwq: > One small issue I didn't notice before, could you add translation tags to the aria label, eg. _('Search followee') so it would be translated too if the user uses some other language than english.

Good point. In addition to adding a translation tag, I also extracted all messages and updated the entire dictionary.

Zharktas: However, translations are extracted and compiled during release process as it makes harder to see what has actually changed in pull requests and produces unnecessary conflicts if PR's modify the compiled translations.

bblwq: > However, translations are extracted and compiled during release process as it makes harder to see what has actually changed in pull requests and produces unnecessary conflicts if PR's modify the compiled translations.

True, that's a good point. I guarantee I never modified the compiled translations though. How about now?

Zharktas: Now it's all good :)

ckan/templates/home/index.html
```diff
@@ -8,7 +8,7 @@
 
 {% block content %}
   <div class="homepage layout-{{ homepage_style }}">
-    <div class="container">
+    <div id="content" class="container">
```
boykoc, line 11: Any reason why this was added here when there is already an `id="content"` in the page.html? https://github.com/ckan/ckan/blob/0e3f501e9fe36289cee4fc59647db80feb0e2f97/ckan/templates/page.html#L21

But this div doesn't seem to get rendered (at least not in 2.8).  If it's supposed to be here, maybe we should remove that other id. But would be nice to know why it wasn't working to start with.

boykoc, line 11: This is good for the home page.  It seems for most other pages this is already set in [page.html](https://github.com/ckan/ckan/blob/0e3f501e9fe36289cee4fc59647db80feb0e2f97/ckan/templates/page.html#L21).

One thing that seems to be missing is making the skip link keyboard accessible.  Right now it seems that you can't navigate to it and click on it. It looks like the current way it's being [hidden is an old way of just using `display:none;.`](https://github.com/ckan/ckan/blob/18adddcf1150f912b65bac5aa439e5bd647a52ee/ckan/templates/page.html#L6).

Would be good to switch this to Bootstraps built in handling of this as [outline here.](https://getbootstrap.com/docs/3.3/getting-started/#skip-navigation). 

As is, it'll pass an automated test but not a manual one.

boykoc: @Zharktas including you for your input and not sure if you'd get notified on a close PR?

Zharktas: @boykoc yeah, it should use sr classes instead of hide. Want to make pr for it ?

Issue #5542, opened 2020-08-14
Purging some datasets lead to 500 error


Engerrs:
**CKAN version**
CKAN 2.9

**Describe the bug**
Purging Datasets using Trash admin page, leads to 500 errors

**Expected behavior**
Purging Datasets without errors.

**Additional details**
It seems that this happens because some columns in revision tables are still having foreign key. I will create new migration removing those relationships.

amercader:
Thanks @Engerrs , can you fix the coding standards issues?

@smotornyuk @Zharktas I know we don't normally include db changes in patch releases but this seems worth including in 2.9.1, and asking in the readme to run `ckan db upgrade. WDYT?

Zharktas:
@amercader I'm in favor of releasing it in 2.9.1 as broken functionality is not desirable to wait until 3.0 or 2.10.0 or whatever next big release is.

shasan-marsdd:
> @amercader I'm in favor of releasing it in 2.9.1 as broken functionality is not desirable to wait until 3.0 or 2.10.0 or whatever next big release is.

whole heartedly agree, we need this fix for our use case as well, and it does cause the whole instance of the app to start returning 502 errors constantly until it is shut down and restarted.  I realize purge isn't a often used functionality but if an admin can break the whole app, which requires waking up the server admins to fix, it's not really something we can just let go of that easily.  Only workaround right now is I do this manually to the db, which isn't ideal either, because we'll have to bring the migrations up to date manually afterwards as well.

amercader:
Fixed in #5533

PR #5533, opened 2020-08-07
Fixing issues with purge options when revisions have data for the dat…

shasan-marsdd:
Fixes #

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [X] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #5543, opened 2020-08-14
[#5542] Remove foreign keys relationships in revision tables

Engerrs:


Zharktas:
This duplicates #5533

PR #5587, opened 2020-09-07
[#5586] Don't cache license translations across requests

### LINKED ISSUE #5586: Incorrectly Cached License Translations
**CKAN version**

2.8.5->master.

Probably introduced in #4594.

**Describe the bug**

ckan.models.license:LicenseRegister.load_licenses translates the license text, which is then cached as a classmember in the package class. (as called from ckan.lib.helpers.license_options)

This cache is for the lifetime of uwsgi process, not the request. 

The license options should be translated at the helper or the template level, and not cached between multiple requests, unless the cache is keyed to the request language. 

**Steps to reproduce**

On a site with multiple localizations, restart ckan to clear any caches, then request `/[lang]/dataset/edit/[name]` several times to make sure all wsgi processes have loaded the licenses.

Then visit ` `/en/dataset/edit/[name]` and the licenses will be in the previous localization. 

**Expected behavior**

Each license popup would be rendered in the correct localization.

EricSoroos:
Fixes #5586

### Proposed fixes:

* Licenses are read and cached in the package model in english
* Translate the license in the helper for each request.

### Features:

- [ ] includes API changes

(note, unclear how to test this as it's tricky to trigger in isolation)

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2840,7 +2840,7 @@ def license_options(existing_license_id=None):
         license_ids.insert(0, existing_license_id)
     return [
         (license_id,
-         register[license_id].title if license_id in register else license_id)
+         _(register[license_id].title) if license_id in register else license_id)
         for license_id in license_ids]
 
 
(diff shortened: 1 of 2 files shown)


amercader:
Thanks @EricSoroos sorry it took a while

PR #5596, opened 2020-09-09
Removed hardcoded references to db host and users

mabah-mst:
Fixes #
The docker-compose.yml file describes that you can modify it to use your own postgres database. Currently this is not possible, since the contrib/docker/ckan-entrypoint.sh is hardcoded check the host db for the database being up. I consider the hardcoded reference to "db" in ckan-entrypoint.sh a bug.

It is also currently not possible to modify the names of the database users in the .env file. This patch enables this, while keeping the existing calues (ckan and datastore_ro) as defaults. If running on on Azure's Postgress PAAS offering it is currently not possible to have as user named ckan as it postfixes the servername to the username. If an Azure Postgres server is hosted on myhost.postgres.database.azure.com the ckan user on that server would have to log in with the username "ckan@myhost". This patch makes it easy to use such a setup in a production environment without bothering any users without this need.

### Features:

- [ ] includes API changes

### DIFF
--- a/contrib/docker/.env.template
+++ b/contrib/docker/.env.template
@@ -34,12 +34,18 @@ CKAN_SMTP_MAIL_FROM=ckan@localhost
 # Image: db
 POSTGRES_PASSWORD=ckan
 #
+# POSTGRES_USER must be set to the database user used by CKAN, typically named ckan
+POSTGRES_USER=ckan
+#
 # POSTGRES_PORT must be available on the host: sudo netstat -na | grep 5432
 # To apply change: docker-compose down && docker rmi docker_db docker_ckan && docker-compose build
 POSTGRES_PORT=5432
 #
 # The datastore database will be created in the db container as docs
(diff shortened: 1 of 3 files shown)


kowh-ai:
@mabah-mst - I have just  used the changes in your PR to install  a fresh CKAN docker environment - all looks good - thanks for this

kowh-ai (approved):
I have installed a fresh CKAN docker environment with these new changes - works fine

mabah-mst:
> I have installed a fresh CKAN docker environment with these new changes - works fine

My current build process checks out tag ckan-2.9.0, applies the fix from https://github.com/ckan/ckan/commit/878dbf6bf2a058de4efa0d8fce9b66e2ef5b7fa3 and the changes in this pull request.
I'm hoping both changes will be included in tag ckan-2.9.1 which would simplify my build process a little.

mabah-mst:
I have reverted the PR to draft status while verifying the update to use a configurable username in the pg_isready checks

PR #5611, opened 2020-09-21
[#5602] unflatten: skip empty items regression + test

### LINKED ISSUE #5602: Data Dictionary Update Bug
**CKAN version**
2.8

**Describe the bug**
Data dictionary not updating properly.

**Steps to reproduce**
Example of inputting labels and descriptions:
image

Output:
image

The description from the first field is being inputted into the description in the second field.

Which code is the logic of the editing of the field inputs? The front end code seems fine...

In addition, where in the postgresql database are the label and description being stored? I can't seem to find them.

wardi:
Fixes #5602 

### Proposed fixes:

fix regression in unflatten that caused:
```python
unflatten({("items", 1, "name"): "first", ("items", 1, "value"): "v1",}) == {
    "items": [
        {"name": "first"},
        {"value": "v1}
    ]
}
```
when it should be:
```python
unflatten({("items", 1, "name"): "first", ("items", 1, "value"): "v1",}) == {
    "items": [
        {"name": "first", "value": "v1}
    ]
}
```

This regression appears in 2.9.0 and 2.8.5. This PR is should replace the earlier partial fix #5546 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/navl/dictization_functions.py
+++ b/ckan/lib/navl/dictization_functions.py
@@ -405,7 +405,7 @@ def unflatten(data):
     '''
 
     unflattened = {}
-    convert_to_list = []
+    clean_lists = {}
 
     for flattend_key in sorted(data.keys(), key=flattened_order_key):
         current_pos = unflattened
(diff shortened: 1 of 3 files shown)


Zharktas:
I tested this and this also makes https://github.com/ckan/ckanext-scheming/pull/255 unnecessary.

PR #5623, opened 2020-09-28
Split new_user_form.html into blocks

chris48s:
Now that the `User` model has `plugin_extras` it is quite useful for plugin authors to be able to customise the signup form. This is currently a bit awkward because `new_user_form.html` is all one block, so you have to copy & paste the whole thing. This PR splits the form into blocks so it is easier for plugin authors to customise it.

- [ ] includes API changes


chris48s:
I've updated this to apply similar changes to `edit_user_form.html`

PR #5629, opened 2020-10-01
Listing users or sysadmins error

### LINKED ISSUE #5621: CKAN 2.9.0 can't create a sysadmin user – Command 'sysadmin' not known
**CKAN version**: 2.9.0 + patch for Docker install (PR #5381)

**Describe the bug**
When first installing CKAN and trying to create the first sysadmin user, I get a message "Command 'sysadmin' not known (you may need to run setup.py egg_info)" instead.

**Steps to reproduce**
I'm following the instructions in the documentation for [installing with Docker Compose](https://docs.ckan.org/en/2.9/maintaining/installing/install-from-docker-compose.html). I initially came across bug #5572, as the CKAN container wouldn't work, then applied the patch [as suggested by mabah-mst](https://github.com/ckan/ckan/issues/5572#issuecomment-693263908) to fix it:

```
git checkout tags/ckan-2.9.0
#Apply fix to get CKAN to start at all in the docker-compose: https://github.com/ckan/ckan/pull/5381
git diff 9abeaa1b7d2f6539ade946cc3f407878f49950eb^ 9abeaa1b7d2f6539ade946cc3f407878f49950eb | git apply
```

Then, skip step 3, as we don't need to set up and enable the data store and data pusher.

We get to step 4, [create CKAN admin user](https://docs.ckan.org/en/2.9/maintaining/installing/install-from-docker-compose.html#create-ckan-admin-user).

However, when running the suggested command, we get the error message saying "Command 'sysadmin' not known".

```
$ docker exec -it ckan /usr/local/bin/ckan-paster --plugin=ckan sysadmin -c /etc/ckan/production.ini add johndoe
Command 'sysadmin' not known (you may need to run setup.py egg_info)
Known commands:
  create       Create the file layout for a Python distribution
  exe          Run #! executable files
  help         Display help
  make-config  Install a package and create a fresh config file/directory
  points       Show information about entry points
  post         Run a request for the described application
  request      Run a request for the described application
  serve        Serve the described application
  setup-app    Setup an application, given a config file
```

**Expected behavior**
A new sysadmin user is expected to be created, instead of displaying this error.

**Additional information***
I also tried entering the container and enabling the Python environment first, but the result is the same.

```
docker exec -it ckan bash
cd /usr/lib/ckan/
source venv/bin/activate
/usr/local/bin/ckan-paster --plugin=ckan sysadmin -c /etc/ckan/production.ini add johndoe
```

tino097:
When we are listing users or sysadmins but there is no records in the database, error stack is displayed. Proposed fix is to display an error message and maybe we could invoke add user command 

Fixes #5621 
### Proposed fixes:
Display an error message 
`"There are no users in the database. Maybe you should create one!"
`

### Features:

- [ ] includes API changes
- [x ] includes bugfix for possible backport

### DIFF
--- a/ckan/cli/sysadmin.py
+++ b/ckan/cli/sysadmin.py
@@ -2,6 +2,8 @@
 
 import click
 from six import text_type
+from sqlalchemy.exc import ProgrammingError
+
 
 import ckan.model as model
 from ckan.cli import error_shout
@@ -24,6 +26,9 @@ def sysadmin(ctx):
 
 @sysadmin.command(name=u"list", help=u"List sysadmins.")
 def list_sysadmins():
+    """
+    List all the sysadmins
+    """
     click.secho(u"Sysadmins:")
     sysadmins = model.Session.query(model.User).filter_by(
         sysadmin=True, state=u"active"
@@ -40,8 +45,8 @@ def list_sysadmins():
                     sysadmin.id,
                 )
             )
-    except Exception:
-        error_shout("There are no users in the database. Maybe you should create one!")
+    except ProgrammingError:
+        error_shout(u'The database is not created. Please initialize database first')
 
 
 @sysadmin.command(help=u"Convert user into a sysadmin.")
--- a/ckan/cli/sysadmin.py
+++ b/ckan/cli/sysadmin.py
@@ -54,19 +54,22 @@ def list_sysadmins():
 @click.argument(u"args", nargs=-1)
 @click.pass_context
 def add(ctx, username, args):
-    user = model.User.by_name(text_type(username))
-    if not user:
-        click.secho(u'User "%s" not found' % username, fg=u"red")
-        if click.confirm(
-            u"Create new user: %s?" % username, default=True, abort=True
-        ):
-            ctx.forward(add_user)
-            user = model.User.by_name(text_type(username))
+    try:
+        user = model.User.by_name(text_type(username))
+        if not user:
+            click.secho(u'User "%s" not found' % username, fg=u"red")
+            if click.confirm(
+                u"Create new user: %s?" % username, default=True, abort=True
+            ):
+                ctx.forward(add_user)
+                user = model.User.by_name(text_type(username))
 
-    user.sysadmin = True
-    model.Session.add(user)
-    model.repo.commit_and_remove()
-    click.secho(u"Added %s as sysadmin" % username, fg=u"green")
+        user.sysadmin = True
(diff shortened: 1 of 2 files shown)


ckan/cli/sysadmin.py
```diff
@@ -27,17 +28,20 @@ def list_sysadmins():
...
+                u"%s name=%s email=%s id=%s"
+                % (
+                    sysadmin.__class__.__name__,
+                    sysadmin.name,
+                    sysadmin.email,
+                    sysadmin.id,
+                )
             )
-        )
+    except Exception:
```
amercader, line 43: Let's use a more specific exception here

tino097:
@amercader should this be marked for backporting ?

amercader:
Having looked at this, I wonder if we can implement something generic at a higher level, because this same issue (DB not initialized) will happen with many other commands (eg `ckan dataset`, ` ckan views`, `ckan seed`...)
Perhaps we can add a `requires_db` decorator on these commands that performs a quick check before executing them, or some sort of allowed or disallowed list like we have for commands that don't require config (although I suspect it would be more tedious to find all the ones that require db access)
Any thoughts on this @smotornyuk ?

smotornyuk:
TLDR; +1 to `requires_db` decorator

I agree that using a separate decorator is the most simple and flexible way. As for maintaining the separate list of commands that relies on DB, we'll definitely forget to add something to that list, so I really don't want to use such an approach:)

As an alternative, we can define global error handler, but it just too general and it will rely on psycopg2 internals, so it's not the best option either: 
```python
diff --git a/ckan/cli/cli.py b/ckan/cli/cli.py
index cc61b9783..8b9206523 100644
--- a/ckan/cli/cli.py
+++ b/ckan/cli/cli.py
@@ -134,6 +134,17 @@ class CustomGroup(click.Group):
         u'entry_point': u'Entry points',
     }
 
+    def invoke(self, ctx):
+        from sqlalchemy.exc import ProgrammingError
+        from psycopg2.errors import UndefinedTable
+        try:
+            super(CustomGroup, self).invoke(ctx)
+        except ProgrammingError as e:
+            if isinstance(e.orig, UndefinedTable):
+                p.toolkit.error_shout(e.orig.diag.message_primary)
+                ctx.abort()
+            raise
+
     def format_commands(self, ctx, formatter):
         # Without any arguments click skips option callbacks.
         self.parse_args(ctx, [u'help'])
```

PR #5639, opened 2020-10-02
reloader option fixed

### LINKED ISSUE #5632: No way to disable reloader when starting the development server
**CKAN version**
2.9.0

**Describe the bug**
When trying to start the development server without a reloader i encountered problems with the `--reloader` argument.
The reloader option requires a TEXT argument, therefore i expected that --reloader False disables the reloader.

**Steps to reproduce**
Start ckan with following command:

`ckan -c [PATH_TO_CONFIG] run --host 0.0.0.0 --reloader False`

**Expected behavior**
Server starts without reloader

**Additional details**

Currently the `reloader` option is passed as string and if it's not provided it defaults to the boolean value `True`

So we have two cases when the `run_simple` method is called:
1. `--reloader` argument is not provided --> reloader=True
2. `--reloader` argument is provided --> some string is passed as reloader argument to the `run_simple` method, which evaluates to true in the if statement distinguishing whether the reloader should be used or not.

So the `--reloader` argument does not affect anything.

_My suggestion:_ rename the argument to `disable-reloader` and turn it into a boolean flag. This enables the user to disable the reloader and the default behaviour (i.e. the dev server starts with a reloader) stays the same.

fwoerister:
Fixes #5632 

### Proposed fixes:
Introduces a new `--disable-reloader`option for the ckan run command to enable users to run the dev-server without reloader.
The default behavior (server running **with** reloader) remains unchanged.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/server.py
+++ b/ckan/cli/server.py
@@ -5,8 +5,8 @@ import logging
 import click
 from werkzeug.serving import run_simple
 
-from ckan.common import config
 import ckan.plugins.toolkit as tk
+from ckan.common import config
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


pdelboca:
Looks good! Thanks for the contribution @fwoerister, this is ready to merge @amercader :+1:

PR #5648, opened 2020-10-06
Remove jinja2 blocks from robots.txt

smotornyuk:
Previously, robots.txt was rendered as a Jinja2 template and had few blocks for extensions. Now it served as a static file and doesn't need those blocks, because an extension can just contain its own robots.txt in the public folder, which will be served instead

### DIFF
--- a/ckan/public/robots.txt
+++ b/ckan/public/robots.txt
@@ -1,11 +1,6 @@
 User-agent: *
-{% block all_user_agents -%}
 Disallow: /dataset/rate/
 Disallow: /revision/
 Disallow: /dataset/*/history
 Disallow: /api/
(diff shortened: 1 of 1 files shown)


mtearle:
What happens when multiple extensions need to update the robots.txt file?

Zharktas:
It was done as part of python3 migrations https://github.com/ckan/ckan/commit/8ee21484d3ceb6c0f0d9eb8d239a3a97baf2e5ac for simplicity sake. However it can be reimplemented as flask view if someone has the time for it.

mtearle:
#6930 notes the issue, with a pull request offering a fix in #6953

Issue #4802, opened 2019-05-23
Update the WSGI entrypoint to not use paste.deploy


amercader:
TODO

paste.deploy loadapp

#### Story points

davidread:
We want to get rid of PasteDeploy's `loadapp` and instead use Werkzeug.

It looks like @JGulic has done this for the Profile command - https://github.com/ckan/ckan/pull/5244.

The heart of this ticket is to change the .wsgi file we use to deploy CKAN into production: https://github.com/ckan/ckan/blob/master/doc/maintaining/installing/deployment.rst#4-create-the-wsgi-script-file 

A basic Werkzeug .wsgi is documented here: https://werkzeug.palletsprojects.com/en/1.0.x/deployment/mod_wsgi/#creating-a-wsgi-file

Our [make_app()](https://github.com/ckan/ckan/blob/master/ckan/config/middleware/__init__.py#L50) also needs a config parameter. So I guess we need to do: CKANConfigLoader(filename).get_config()
(whereas loadapp we just passed in the filename)

The existing apache.wsgi also configures logging:
```
from paste.script.util.logging_config import fileConfig
    fileConfig(config_filepath)
```
We want to get rid of PasteScript in py3 requirements. Logging is not configured from the ini when running tests or the click cli, so this is a new thing to ckan. It looks the built-in python3 logging module has what we want: [logging.config.fileConfig](https://docs.python.org/3.5/library/logging.config.html#logging.config.fileConfig)

davidread:
☝️ I wrote this with a bit of research - I'd appreciate @amercader or @smotornyuk checking this is right. And if it is then it should be easier for whoever gets there first to contribute a PR and test it out with Apache.

kowh-ai:
I will also be able to test the new (Werkzeug)  .wsgi file with the new uWSGI and gunicorn "servers/entrypoints" as I have now finished these new configurations. They both work with the PasteDeploy loadapp file.  I'm guessing Apache will still be around for CKAN 2.9 however I could add these 2 new configs to the CKAN 2.9 documentation. I guess we will be phasing out Apache completely by CKAN 3.0...

davidread:
@kowh-ai I'm very happy to see https://github.com/ckan/ckan/issues/4991 progressing 👍 

When we have a PR for this issue, it doesn't really matter which server it is tested against - I just said Apache as it was the current default - so I'm also very happy to see your offer of testing it 🙏

PR #5659, opened 2020-10-09
Upgrade requirements

pdelboca:
Fixes #5638 and #5656 

### Notes:

- `python2-secrets==1.0.5` was repeated in `requirements-py2.txt`
- Under Python 3 `pip-compile` adds `# via -r requirements.in` automatically
- `requests` package doesn't have a hard dependency on `urllib` so upgrading it doesn't upgrade what we need, changed it manually as suggested in #5638 

@amercader: 
- if `requirements.txt` are output from commands, shouldn't be better to have all the manual changes of dependencies documented in `requirements.in`?
- There is a manually kept requirement: `pastedeploy==2.0.1        # manually kept - remove when #4802 is complete` that maybe we can remove since #4802 is already closed?
- `pip-compile` allows the `--upgrade` parameter that I assume it will safely upgrade all the dependencies to the latest version without creating any conflict. I did a test and it updates correctly `urllib` (and another ones to newer versions)


amercader:
@pdelboca You are right, we shouldn't manually update `requirements.txt` files. If we need to pin a dependency dependency we should add it to `requirements.in`. But in any case I run `pip-compile --upgrade` with requests upgraded to 2.24.0 in `.in` and didn't see any problematic changes so let's go with that (as this will upgrade urllib3 to 1.25.11)

amercader:
About pastedeploy if you include the following patch in this PR we can remove the requirement line altogether:

```diff
diff --git a/ckan/lib/helpers.py b/ckan/lib/helpers.py
index 69d8b591a..f96787708 100644
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -18,7 +18,7 @@ import uuid
 import functools
 
 from collections import defaultdict
-from paste.deploy import converters
+from ckan.common import asbool
 
 import dominate.tags as dom_tags
 from markdown import markdown
@@ -2891,8 +2891,8 @@ core_helper(localised_filesize)
 core_helper(i18n.get_available_locales)
 core_helper(i18n.get_locales_dict)
 core_helper(literal)
-# Useful additions from the paste library.
-core_helper(converters.asbool)
+# Useful additions from CKAN converters.
+core_helper(asbool)
 # Useful additions from the stdlib.
 core_helper(urlencode)
 core_helper(include_asset)
```

pdelboca:
Great @amercader ! I did the changes you suggested :+1:

PR #5673, opened 2020-10-15
Re-enable tracking data

mpolidori:
### Proposed re-implementation:

This brings back `tracking_summary` data by default and allows the flame icon to appear next to popular (views/downloads > 10) datasets and resources when `tracking_enabled` is set in the config.

### Features:

- [x] includes updated documentation **(documentation was out of date for this and assumed that this functionality still existed by default)**
- [x] includes user-visible changes **(when `tracking_enabled = true` in config, flame icon appears on popular datasets/resources)**
- [x] includes API changes **(changes output of `package_show`/`package_search` to include `tracking_summary`)**
- [ ] includes bugfix for possible backport **(this was the working code in previous CKAN versions)**

### DIFF
--- a/ckan/controllers/package.py
+++ b/ckan/controllers/package.py
@@ -345,7 +345,7 @@ class PackageController(base.BaseController):
         context = {'model': model, 'session': model.Session,
                    'user': c.user, 'for_view': True,
                    'auth_user_obj': c.userobj}
-        data_dict = {'id': id, 'include_tracking': True}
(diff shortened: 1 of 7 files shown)


mpolidori:
@wardi mentioned to me that the reason this was removed because it was thought to cause performance issues (but that this isn't actually the case).

mpolidori:
I need to fix the tests. Should have time over the weekend.

pdelboca:

PR #5693, opened 2020-10-27
Set default group_type to group in group_create

Zharktas:
### Proposed fixes:
When using custom group schemas, group_create does not use the schema unless type is explicitly defined as group. The models sets it to [group](https://github.com/ckan/ckan/blob/master/ckan/model/group.py#L117) if it not defined, so group_update would use the custom schema.  Organization_create sets it to type [organization](https://github.com/ckan/ckan/blob/master/ckan/logic/action/create.py#L928) so it is incoherent to demand that group_create must explicitly define it.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -702,7 +702,7 @@ def _group_or_org_create(context, data_dict, is_org=False):
     upload.update_data_dict(data_dict, 'image_url',
                             'image_upload', 'clear_upload')
     # get the schema
-    group_type = data_dict.get('type', 'group')
+    group_type = data_dict.get('type', 'organization' if is_org else 'group')
     group_plugin = lib_plugins.lookup_group_plugin(group_type)
     try:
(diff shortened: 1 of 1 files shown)


ckan/logic/action/create.py
```diff
@@ -702,7 +702,7 @@ def _group_or_org_create(context, data_dict, is_org=False):
     upload.update_data_dict(data_dict, 'image_url',
                             'image_upload', 'clear_upload')
     # get the schema
-    group_type = data_dict.get('type')
+    group_type = data_dict.get('type', 'group')
```
wardi, line 705:
```suggestion
    group_type = data_dict.get('type', 'organization' if is_org else 'group')
```

wardi:
Travis: `anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit` guess it's time to figure out how to authenticate to docker.

PR #5695, opened 2020-10-27
Allow user to follow deleted entities

mutantsan:
Currently, if you try to follow any deleted object you'll get the internal error. The follow counter is always 0, because it's doesn't count the followings if object is deleted.
This behavior is weird, because we can visit the deleted obj page and we can restore it. Moreover, we can press the 'Follow' button and it's not hidden.

### Features:

- [ ] includes API changes


mutantsan:
still a problem, updated

wardi:
@mutantsan the admin UI and it sounds like the sites you work on make deleted objects a lot more visible than vanilla ckan so I'm really interested in your thoughts on https://github.com/ckan/ckan/issues/7368

mutantsan:
I decided to take a look at some old PRs and realized this big boy is about to turn 6 years old, and it's still an issue. 😄

Let me know if you'd prefer a different fix, such as removing the Follow button and restricting the action.

PR #5703, opened 2020-10-29
Keep get parameters while redirecting

mutantsan:
I had a problem with pagination at harvester source. I found out that harvester using source.id to create a path for pagination (_package_list_for_source_ func) '.
In this part of code (dataset.read:491) we redirect the user using pkg.name if he using pkg.id. And in process we are loosing the get parameters. In this changes I'm trying to keep the get parameters during redirect.

### Features:

- [ ] includes API changes


mutantsan:
still a problem, updated
try something like `http://127.0.0.1:5000/dataset/f5ad3d81-ef8b-4f68-81f7-3fa071f25bab?test=1` to test

PR #5712, opened 2020-11-03
Remove old Pylons controllers and routing (and with it, support for Python 2)

amercader:
I think now that 2.9 has been out for a while and we had a patch release out it's a good time to start cleaning up for good, but happy to discuss if someone doesn't agree.

I wanted to start with the old Pylons controllers, which are obviously tied to the old routing. Because of how we initialize the environment this essentially means that the code base can no longer be run in Python 2.

Which is good, and what our original plan was for post-2.9. And tests now magically run in half the time! :sparkles:


jqnatividad:
Hi @amercader.  I'm currently working on an implementation using CKAN 2.9.1 running on Python 2.7 - primarily, thinking I'll need to do so to use the following required extensions - ckanext-ldap, ckanext-pages & ckanext-issues. 

[ldap](https://github.com/NaturalHistoryMuseum/ckanext-ldap) has been recently updated to support python 3.

pages has been [shortlisted for migration](https://github.com/ckan/ckan/wiki/Python-3-migration-guide-for-extensions), and it works for me right now, I gather, because I'm still on Python 2.7.

As for issues, I just started working on it to make it work per @davidread's comment (https://github.com/ckan/ckanext-issues/issues/80). 

To "future-proof" this implementation (cleanly upgrade to CKAN 3; work post pylons/python 2), I have a few questions:

1) what is the status of ckanext-pages python 3 migration? Am I right to assume its working on my installation as I'm still on Python 2.7?
2) To minimize rework for ckanext-issues, I'm thinking of following the ["mixin plugin"](https://github.com/ckan/ckan/wiki/Python-3-migration-guide-for-extensions#create-mixin-plugin-implementations-for-pylons--flask) implementation per the migration guide.  Should I still follow that guidance?

What's your recommendation?  Thanks in advance!

amercader:
@jqnatividad 
> 
    what is the status of ckanext-pages python 3 migration? Am I right to assume its working on my installation as I'm still on Python 2.7?

AFAIK pages still has not been formally updated to support 2.9 so I'd double check that the functionality works as expected. It definitely won't support py3 as it is now. I project I'm working on has its migration scheduled, but it will take at least a couple of weeks to get started.

>     To minimize rework for ckanext-issues, I'm thinking of following the "mixin plugin" implementation per the migration guide. Should I still follow that guidance?

Yes, for extensions that need to support different versions the mixin approach @smotornyuk pioneered is the most straight-forward, and is working well on harvest and dcat for instance if you need examples.

smotornyuk:
No problems detected

jqnatividad:
Thanks for the tip @amercader!  I'm working on ckanext-issues using the mixin approach right now, and will be sure to contribute it back.

PR #5718, opened 2020-11-05
Docs: Update uswgi example to support non-root path installations

Zharktas:
### Proposed fixes:

When running ckan on some other path than root, like for example /data, uwsgi file needs to modifed to mount ckan in different path. This PR updates the example with required modification while still mounting ckan to root. The user only needs to update the mount point e.g. mount = /data=wsgi:application 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan-uwsgi.ini
+++ b/ckan-uwsgi.ini
@@ -1,15 +1,17 @@
 [uwsgi]
 
-http            =  127.0.0.1:8080
-uid             =  www-data
-guid            =  www-data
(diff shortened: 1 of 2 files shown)


Zharktas:
ping @wardi can this be advanced some where :D

Zharktas:
i can take a look at it how it should be configured to allow those paths, I wonder how apache was configured to allow those, the old docs aren't saying a thing about it and I have never tried to run CKAN in any other than `/foo/{{ LANG }}` paths

PR #5734, opened 2020-11-10
Add allow_partial_update to fix losing users in group and organization patch actions

Zharktas:
### Proposed fixes:
When public user details is set to false, group and organization show actions do not include users by default. Calling group or organization patch actions will lose users unless developer sets allow_partial_update to the context. This is documented in [here](https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L769). This PR set allow_partial_update to the context so that the developer doesn't need to know about it and enables patch functions to work properly over http. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/patch.py
+++ b/ckan/logic/action/patch.py
@@ -137,7 +137,7 @@ def organization_patch(context, data_dict):
     patched = dict(organization_dict)
     patched.pop('display_name', None)
     patched.update(data_dict)
-    return _update.group_update(dict(context, allow_partial_update=True), patched)
+    return _update.organization_update(dict(context, allow_partial_update=True), patched)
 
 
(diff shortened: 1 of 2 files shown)


ckan/logic/action/patch.py
```diff
@@ -107,6 +107,7 @@ def group_patch(context, data_dict):
     patched = dict(group_dict)
     patched.pop('display_name', None)
     patched.update(data_dict)
+    context['allow_partial_update'] = True
     return _update.group_update(context, patched)
```
wardi, line 111:
```suggestion
    return _update.group_update(dict(context, allow_partial_update=True), patched)
```

Zharktas: Done

PR #5750, opened 2020-11-20
Datastore cli purge

### LINKED ISSUE #5589: Purging deleted packages does not delete associated filestore/datastore resources
**CKAN version**
2.9

**Describe the bug**
After deleting and purging dataset packages, the associated datastore/filestore resources are orphaned and not deleted.

**Steps to reproduce**

1. Delete datasets
2. Go to sysadmin/trash
3. Select "Purge All"
4. Check datastore and filestore.  Associated resources are still there.

**Expected behavior**
Dataset resources in the filestore and the datastore are deleted when the associated package resources are purged.

**Additional details**
This is related to #4705 and a lot of work has been done on PRs #4867 and #4905, but neither have been merged.

jqnatividad:
Fixes #5589

### Proposed fixes:
- [X] implement datastore purge cli command.
  Do this by scanning `_table_metadata` for orphaned resources (i.e. there is an entry in _table_metadata, but no matching resource), and dropping these orphaned tables by calling `datastore_delete` for that resource_id without filters - which results in a drop table cascade.
- [x] update documentation

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/cli.py
+++ b/ckanext/datastore/cli.py
@@ -138,8 +138,9 @@ def purge():
             if record[u'alias_of']:
                 continue
 
-            # we need to do this in the loop to trigger resource_show auth function
-            site_user = logic.get_action(u'get_site_user')({u'ignore_auth': True}, {})
+            # we need to do this to trigger resource_show auth function
+            site_user = logic.get_action(u'get_site_user')(
+                {u'ignore_auth': True}, {})
             context = {u'user': site_user[u'name']}
 
             logic.get_action(u'resource_show')(
--- a/ckanext/datastore/cli.py
+++ b/ckanext/datastore/cli.py
@@ -133,9 +133,14 @@ def purge():
     resource_id_list = []
     for record in result[u'records']:
         try:
-            # ignore 'alias' records
+            # ignore 'alias' records (views) as they are automatically
+            # deleted when the parent resource table is dropped
             if record[u'alias_of']:
                 continue
+
+            # we need to do this in the loop to trigger resource_show auth function
+            site_user = logic.get_action(u'get_site_user')({u'ignore_auth': True}, {})
+            context = {u'user': site_user[u'name']}
 
             logic.get_action(u'resource_show')(
                 context,
--- a/ckanext/datastore/cli.py
+++ b/ckanext/datastore/cli.py
@@ -137,13 +137,20 @@ def purge():
                 context,
                 {u'id': record[u'name']}
             )
-            click.echo(u"Resource '%s' found" % record[u'name'])
-        except logic.NotFound:
+        except logicfNotFound:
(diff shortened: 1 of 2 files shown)


amercader:
@jqnatividad is this ready to review?

jqnatividad:
@amercader , if you can give some initial feedback, I'd appreciate it! 

Was planning on adding documentation first before removing draft, but then, I stumbled on #5752 and #5755. :)

Also, can you confirm that cli commands don't require tests?

ckanext/datastore/cli.py
```diff
@@ -111,3 +112,48 @@ def _parse_db_config(config_key=u'sqlalchemy.url'):
...
+            click.echo(u"Resource '%s' found" % record[u'name'])
+        except logic.NotFound:
+            resource_id_list.append(record[u'name'])
+            click.echo(u"Resource '%s' orphaned - queued for drop" %
+                       record[u'name'])
+        except KeyError:
+            continue
+
+    # Drop the orphaned datastore tables. When datastore_delete is called
+    # without filters, it does a drop table cascade
```
amercader, line 149: Does this mean it deletes the record on `_table_metadata`?

jqnatividad: Effectively yes.  As `_table_metadata` is a view of all the existing tables in the datastore.

https://github.com/ckan/ckan/blob/5156a4024a1aea7af6ad927b4af8a70aab8597ee/ckanext/datastore/set_permissions.sql#L53-L71

So dropping a resource table removes the corresponding record on `_table_metadata`.

jqnatividad:
After further testing, decided to add a confirmation prompt just to be on the safe side.

jqnatividad:
@amercader this PR is now ready for review.

ckanext/datastore/cli.py
```diff
@@ -111,3 +112,63 @@ def _parse_db_config(config_key=u'sqlalchemy.url'):
...
+            context,
+            {u'resource_id': resource_id, u'force': True}
+        )
+        click.echo(u"Table '%s' dropped)" % resource_id)
+        drop_count += 1
+
+    click.echo(u'Dropped %s tables' % drop_count)
+
+
+def get_commands():
```
smotornyuk, line 173: Looks fine to me. The only question, is this function something that you've used locally? As I remember, we don't export these functions because they registered via `setup.py` [directly](https://github.com/ckan/ckan/blob/master/setup.py#L54) (without IClick interface) - in that way they are available even when the corresponding CKAN plugin is not enabled

jqnatividad: Hi @smotornyuk. its only meant to be used thru the datastore CLI. I'll defer to the core team if there is any additional stuff I need to implement to make it available even if the datastore plugin is not enabled.

Issue #4597, opened 2018-12-19
Attributes on chained actions not always exposed


alycejenni:
### CKAN Version if known (or site URL)
Current master branch (latest commit 8f0cb74)

### Issue
**EDIT:** since submitting the issue some things have _stopped_ working so I'm having to change it.

Decorators for chained actions have to be defined in a specific order to ensure that attributes can be used later, e.g.:

This results in the `side_effect_free` attribute not being exposed:
```py
@toolkit.chained_action
@logic.side_effect_free
def datastore_search(prev_func, context, data_dict):
    ...
```

Example of where this needs to be handled: `ckan/default/src/ckan/ckan/views/api.py:263`

```py
side_effect_free = getattr(function, u'side_effect_free', False)  # this returns False for chained actions
```

wardi:
let's come up with a list of action + auth attributes that we'd like to maintain across action function decorators and then copy them all when we wrap an action function. Am I missing any:
```python
ACTION_FUNCTION_ATTRIBUTES = [
    'side_effect_free',
    'auth_audit_exempt',
]
AUTH_FUNCTION_ATTRIBUTES = [
    'auth_sysadmins_check',
    'auth_allow_anonymous_access',
]
```

Also I'm not sure why these functions are wrapping the action functions they decorate instead of adding their attributes directly to the function.

iamarnavgarg:
Hi @wardi I would like to work on this issue?Can you please guide me further in it?

bzar:
Hey, we ran into this when implementing a chained action on `organization_list`. I cobbled together a quick fix for our issue in bzar/ckan@8cc74a. Is there any progress on the more general solution described here?

amercader:
Fixed in #5873

PR #5756, opened 2020-11-23
bugfix: persisiting auth attributes

### LINKED ISSUE #5751: Chained auth functions lose anonymous access
**CKAN version**
All
**Describe the bug**
If an auth function is marked as chained, it is transformed into a functools partial. In doing so, no function attributes are persisted to the new function - leaving `auth_allow_anonymous_access` flags out. This means that there is no way for developers to specify that an auth function is both chained and allowed to be accessed anonymously.

**Steps to reproduce**
Create a plugin implementing `IAuthFunctions`, and use both the `@toolkit.auth_allow_anonymous_access` and `@toolkit.chained_auth_function` decorators. The function will be chained, but will return `403` before ever running your function.

**Expected behavior**
The function doesn't automatically get 403'd.

**Additional details**
The error is in the `_build` function for the `AuthFunctions` class in `ckan/ckan/authz.py`. Specifically the `auth_allow_anonymous_access` attributes are managed [here](https://github.com/ckan/ckan/blob/612e0f9ab01b61516399e39a803d8cb66cd3c9cf/ckan/authz.py#L79), but then the chained auth functions are transformed into partials [here](https://github.com/ckan/ckan/blob/612e0f9ab01b61516399e39a803d8cb66cd3c9cf/ckan/authz.py#L118). I don't have my dev environment set up for contributing back at the moment (it's in a private fork) but have this fixed in my local 2.8 version and will get this into master this weekend.

hardingalexh:
Fixes #
Fixes #5751 and (partially) #4597. 

### Proposed fixes:
Because chained authentication functions are wrapped into partials, they lose access to any attributes that may have been set on the functions themselves (like allowing anonymous access). This fix defines a list of attributes, checks them against the original function, then applies them to the new partial function.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -116,7 +116,7 @@ class AuthFunctions:
                     prev_func = self._functions[name]
                 
                 new_func = (functools.partial(func, prev_func))
-                # persisting attributes to the new partiall function
+                # persisting attributes to the new partial function
                 attributes = list(func.__dict__.keys())
                 for attribute in attributes:
                     value = getattr(func, attribute)
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -117,9 +117,7 @@ class AuthFunctions:
                 
                 new_func = (functools.partial(func, prev_func))
                 # persisting attributes to the new partial function
-                attributes = list(func.__dict__.keys())
-                for attribute in attributes:
-                    value = getattr(func, attribute)
+                for attribute, value in six.iteritems(func.__dict__):
                     setattr(new_func, attribute, value)
                 
                 fetched_auth_functions[name] = new_func
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -116,16 +116,9 @@ class AuthFunctions:
                     prev_func = self._functions[name]
                 
                 new_func = (functools.partial(func, prev_func))
-                attributes_list = [
-                    'auth_allow_anonymous_access',
-                    'auth_disallow_anonymous_access',
-                    'auth_sysadmins_check',
-                    'chained_auth_function'
-                ]
-                for attribute in attributes_list:
-                    if hasattr(func, attribute):
-                        attval = getattr(func, attribute)
-                        setattr(new_func, attribute, attval)
+                # persisting attributes to the new partial function
+                for attribute, value in func.__dict__.iteritems():
+                    setattr(new_func, attribute, value)
                 
                 fetched_auth_functions[name] = new_func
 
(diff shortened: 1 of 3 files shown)


smotornyuk:
This is error prone in the long term because we need to maintain this list as source code changes. And attributes from custom decorators, that can be introduced by 3rd-party code are still ignored. 
What do you think about using `func.__dict__` instead of a predefined list

Zharktas:
This also need to be done on chained actions and then there are two places to maintain decorator lists which is somewhat undesirable.

smotornyuk:
Probably it should be done in separate PR, but just FYI: we also have [chained_helper](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L147) with a similar block of code [here](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L2921-L2923)

hardingalexh:
@smotornyuk I've rewritten it to use `func.__dict__` rather than a manual list, that's a good call. Should I also add this pattern to actions/helpers in this PR, or should I make separate PRs for each of them? It's pretty simple to add this snippet to both.

smotornyuk:
Probably, it will be faster if you do it as a part of this PR, as long as you don't mind for some reasons:)

And another notice. You've used `dict.iteritems` which is not available in Python3. We use `six` library for handling such cases, so could you rewrite `dict.iteritems` to `six.iteritems(dict)`?

hardingalexh:
Rewrote to be more python-version-agnostic. A little more verbose, but native. Apologies for that, my CKAN dev environment is stuck in 2.8/python 2.7 land for a little while longer. If you're happy with this syntax, I'll add it to actions/helpers and we can get this going!

hardingalexh:
I've added this functionality to the other uses of `functools.partial`. This should be ready to merge unless anyone objects to syntax/has other concerns. @smotornyuk @Zharktas

Zharktas:
I would have just used six.iteritems(dict) which does dict.iteritems on python2 and iter(dict.items) on python3. It would make it easier to understand what is going on at a glance. Tests would be nice as clearly chaining actions, auths or helpers aren't tested.

hardingalexh:
Rewritten to use six. How would I go about adding tests for this? I'm not really sure the best way to intercept a chained function to check if it has the right attributes.

Zharktas:
For chained auth_functions you could import [_AuthFunctions](https://github.com/ckan/ckan/blob/master/ckan/authz.py#L123) from ckan.auths and call get('some_chained_action') and test the attributes. For actions you could just call get_action and for helpers I think by importing helper_functions from ckan.lib.helpers. 

Might require crafting custom plugin which creates chained functions in the first place though.

Zharktas:
Superseded by #5873 which adds tests.

PR #5774, opened 2020-12-04
Don't add --ckan-ini pytest option if already added (by pytest-ckan)

amercader:
We are recommending extensions to migrate to pytest when adding 2.9 support and for that they require the
[pytest-ckan](https://github.com/datopian/pytest-ckan) package. If that package is installed in your venv and you try to run the CKAN core tests, you get a ValueError:

```
    File
"/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/_pytest/config/argparsing.py",
        line 304, in addoption
    raise ValueError("option names %s already added" % conflict)
ValueError: option names {'--ckan-ini'} already added
```

This changes captures the exception and allows execution to continue


amercader:
@wardi comment added

@smotornyuk The reason this can't be implemented in pytest-ckan is that plugins registered via setup entry points (ie installed) are evaluated before the conftest.py ones (ie ckan core): https://docs.pytest.org/en/2.7.3/plugins.html?#plugin-discovery-order-at-tool-startup

PR #5784, opened 2020-12-08
Update organization_show package limit docs

### LINKED ISSUE #5775: organization_show returns 10 packages instead of the advertised 1000
**CKAN version**

2.9.1

**Describe the bug**

/api/3/action/organization_show?include_datasets=True returns a max of 10 packages.

api/3/action/help_show?name=organization_show says 

>     .. note:: Only its first 1000 datasets are returned

**Steps to reproduce**

Create an org and add 11 packages.
Hit the organization_show endpoint with id=your-org
Note that only 10 packages are returned.

**Expected behavior**

I expected up to 1000 packages to be returned. 

**Additional details**

Looks like it might have been related to this change: https://github.com/ckan/ckan/commit/bd01afe8139eb8cb75277503089b96a8247beb7e#diff-479ec62a04a4a945f75aa6d8644b5b609ddc678b4ee8f56328b50b9d748420b9R394

I see logic that looks like it should allow overriding the search default limit of 10, but I'm not sure where/how `context['limits']['packags']` is set: https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L331

mac9416:
Fixes #5775 

### Proposed fixes:

An update was made to let package_search set the package limit for organization_show. As it turns out, package_search always uses 10 in this case. Thus the docs update.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1277,7 +1277,7 @@ def organization_show(context, data_dict):
 
     :rtype: dictionary
 
-    .. note:: Only its first 1000 datasets are returned
+    .. note:: Only its first 10 datasets are returned
     '''
     return _group_or_org_show(context, data_dict, is_org=True)
 


amercader:
Thanks @mac9416 , that was quick!

mac9416:
@amercader you bet, thanks for confirming the issue!

PR #5805, opened 2020-12-21
Docker compose fix

syats:
Fixes #

Updated the Dockerfile and the docker-compose file to use a Python 3 version of Ckan. These files where quite outdated / broken, using some 11-year old versions of python2 libraries (eg pylons) that we causing problems. Documentation was also updated to take into account new instructions, but many of the steps previously covered have not been tested, so it has been greatly reduced.  I am not an expert in CKAN, I just wanted to give it a try using docker.

### Proposed fixes:

Adding extensions, migrating users and data, have not been added to the docker-compose section of the docs yet. The previous version used paster, so now these instructions need updates to use the ckan cli.

### Features:

Current master can now be launched with docker-compose.

- [ ] includes API changes


kowh-ai:
I have just tested this PR and it works. Thank you for your efforts. However, this (seperate) repo: https://github.com/ckan/ckan-docker is in the process of being updated to become the replacement to the current one. There is still some way to go but it will be available for use at some point in the near future.

EricSoroos:
Activity is now on https://github.com/ckan/ckan-docker

PR #5809, opened 2020-12-24
Use CKANRequest object implicitly in templates

mutantsan:
Currently, we have a [wrapper](https://github.com/ckan/ckan/blob/599ff35f9c289bab674f544367d5acdb1d2c9423/ckan/common.py#L162) for **request** object, to support pylons request along with the flask request.
We've added `params` property and it's seems to work.

But, there are two places where we are using **request object**. In python code and inside the templates. There are not so much places where we are appeal to `requests.params`, so this bug wasn't noticed. To really get an error you should fetch `params` from **request** and try to do something with it. An example from our extension:

`{% set enabled = h.or_facet_or_enabled(name, request.params) %}`

You can try to render `{{ request.params }}` inside any template and jinja won't render anything. I've tried to pass the result of **request.params** inside a helper function and got **Undefined**

You can try to pass the **request object** itself into the helper and you'll get an exception trying to get `params` from it. But, you can import request from ckan.common and it's works like a charm. Seems like the **request object** inside the template is not the same we are using in other places. Maybe, flask provides it implicitly somehow. However, the straightforward decision is to provide **request object** with **context_processor** to template explicitly.

### Features:

- [ ] includes API changes


amercader:
Small test needed @mutantsan

amercader:
@mutantsan do you have time to write a small test?

mutantsan:
@amercader sorry, I'll find time soon!

amercader:
@mutantsan I've started backporting stuff into 2.9.3, do you want to clean this up and fix the tests?

PR #5815, opened 2021-01-07
[#5814] textview: escape text formats

### LINKED ISSUE #5814: text_view renders html markup
**CKAN version 2.9.1**

**Describe the bug**
The text_view renders html elements if they occur in a text file.

**Steps to reproduce**
Create resource and upload the following text file:

foo.txt with content
```
<input type="text"/>
```

The text view actually displays this as an input field:
image

**Expected behavior**
Text view just renders everything as plain text

**Additional details**
I didn't test what else can be injected that way. Could this even allow XSS attacks?

wardi:
Fixes #5814

### Proposed fixes:

escape HTML tags in text previewed with textview. Appreciate review from an experienced JS developer, used a textarea element for the escaping because it seemed to be the simplest approach

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/textview/theme/public/text_view.js
+++ b/ckanext/textview/theme/public/text_view.js
@@ -58,7 +58,9 @@ ckan.module('text_view', function (jQuery) {
           if (p.language) {
             highlighted = hljs.highlight(p.language, data, true).value;
           } else {
-            highlighted = '<pre>' + data + '</pre>';
+            var escape_ = document.createElement('textarea');
+            escape_.textContent = data;
+            highlighted = '<pre>' + escape_.innerHTML + '</pre>';
           }
 
           self.el[0].innerHTML = highlighted;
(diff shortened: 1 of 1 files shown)


ckanext/textview/theme/public/text_view.js
```diff
@@ -58,7 +58,9 @@ ckan.module('text_view', function (jQuery) {
           if (p.language) {
             highlighted = hljs.highlight(p.language, data, true).value;
           } else {
-            highlighted = '<pre>' + data + '</pre>';
+            var escape_ = document.createElement('textarea');
+            escape_.textContent = data;
+            highlighted = '<pre>' + escape_.innerHTML + '</pre>';
```
Zharktas, line 63:
```suggestion
            highlighted = $('<pre></pre>').text(data)[0].outerHTML;
```
Textarea is not really needed, any element would do. Instead of creating new elements I would just create the pre element, insert the data as text and return outerHTML to keep the functionality as the same as before.

PR #5825, opened 2021-01-12
Allow to run the development server using SSL

amercader:
Running CKAN using SSL locally is very useful sometimes to replicate live setups, and mandatory in some situations like when developing a SAML2 authentication based instance (which requires CKAN to run on eg https://localhost:5000).

With paster with did that adding the `ssl_pem=/path/to/cert.pem` config option to the ini file. On Werkzeug we need to use the [`ssl_context`](https://werkzeug.palletsprojects.com/en/1.0.x/serving/#werkzeug.serving.run_simple
) param on the call to run the server that we invoke in `ckan run`.

This adds supports for this, both as CLI parameters and as config options, replicating existing `ckan.devserver.*` options. In addition to provide your self-signed cert and key you can also pass `adhoc`, which will make Werkzeug create one for you (on each reload though)

I'm proposing this to be backported to 2.9 as it is really useful and doesn't add any breaking changes.

### DIFF
--- a/doc/maintaining/configuration.rst
+++ b/doc/maintaining/configuration.rst
@@ -158,7 +158,7 @@ Example::
 
 Default value: False
 
-Controls, whether development server handle each request in a separate
+Controls whether the development server should handle each request in a separate
 thread.
 
 .. _ckan.devserver.multiprocess:
@@ -172,7 +172,7 @@ Example::
 
 Default value: 1
 
-If greater than 1 then handle each request in a new process up to this
+If greater than 1 then the developmente server will handle each request in a new process, up to this
 maximum number of concurrent processes.
 
 .. _ckan.devserver.watch_patterns:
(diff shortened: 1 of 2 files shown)


amercader:
Also includes this extreme productivity booster! :chart_with_upwards_trend: 

Screenshot from 2021-01-13 13-11-48

doc/maintaining/configuration.rst
```diff
@@ -172,7 +172,7 @@ Example::
 
 Default value: 1
 
-If greater than 1 then handle each request in a new process up to this
+If greater than 1 then the developmente server will handle each request in a new process, up to this
```
kowh-ai, line 175: typo should be `development`

doc/maintaining/configuration.rst
```diff
@@ -186,8 +186,48 @@ Example::
...
+Example::
+
+  ckan.devserver.ssl_key = path/to/host.key
+
+Default value: None (SSL disabled)
+
+Path to a certificate file that will be used to enable SSL (ie to serve the
+local development server on https://localhost:5000). See :ref:`ckan.devserver.ssl_cert`
+for more details. This option also supports the ``adhoc`` value, with the same caveat.
+
```
kowh-ai, line 230: I'm wondering if a full command line example should be here: `ckan -c /path/to/ckan.ini run --ssl-cert=/path/to/host.cert --ssl-key=/path/to/host.key` _or_ maybe somewhere else in the documentation

kowh-ai (approved):
Looks great Adria! - generated a cert and key as per your doco and brought up a CKAN development HTTPS server listening on port 5000 using `ckan -c /etc/ckan/default/ckan.ini run --ssl-cert=/home/brett/host.cert --ssl-key=/home/brett/host.key`

amercader:
Thanks @kowh-ai, I added your suggestions

PR #5790, opened 2020-12-10
[#5789] Adds typechecks for the max_limit and limit field causing the issue

blazhovsky:
Fixes #5789 

Currently when querying the `organization_list` endpoint with a limit
parameter set the server returns an error as described in the issue.

The fix is just converting the types before comparing

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [X] includes bugfix for possible backport

PR #5830, opened 2021-01-15
[#5789] Fix exception when passing limit to organization_list

### LINKED ISSUE #5789: Organization list crashes when called with a limit parameter
**CKAN version**
2.9.1

**Describe the bug**
When querying/listing organizations or groups with an added limit parameter there is a server error.

**Steps to reproduce**
Calling `curl` on the `organization_list` with the limit field set should recreate it.
E.g.
```
curl http://localhost:5000/api/3/action/organization_list\?limit\=31
{"help": "http://localhost:5000/api/3/action/help_show?name=organization_list", "error": {"__type": "Internal Server Error", "message": "Internal Server Error"}, "success": false}% 
```

**Expected behavior**
The server should return a list of organizations/groups.

amercader:
Fixes #5789 

Based on @blazhovsky work on #5790 

Just extended a bit to ensure we have ints everywhere and added a test

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -350,11 +350,18 @@ def _group_or_org_list(context, data_dict, is_org=False):
 
     if all_fields:
         # all_fields is really computationally expensive, so need a tight limit
-        max_limit = config.get(
-            'ckan.group_and_organization_list_all_fields_max', 25)
+        try:
+            max_limit = int(config.get(
+                'ckan.group_and_organization_list_all_fields_max', 25))
+        except ValueError:
+            max_limit = 25
     else:
-        max_limit = config.get('ckan.group_and_organization_list_max', 1000)
-    if limit is None or limit > max_limit:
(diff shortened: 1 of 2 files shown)


tomeksabala:
I have came up with a very similar fix for a slightly different issue. In my case after having configured custom max limits in ckan.ini:
```
ckan.group_and_organization_list_all_fields_max = 1000
ckan.group_and_organization_list_max = 1000
```
When calling organization list with integer limit I have received `TypeError` as `max_limits` in `get.py` were of type `str`. This is fixed in this PR but you might want to use regression tests I've created to reproduce the bug.
https://github.com/ckan/ckan/pull/5856

pdelboca:
We're good to go @amercader with this one. Although it would be nice to add some extra test coverage as suggested in @tomeksabala PR!

amercader:
Thanks @tomeksabala , I included your tests in 0bcffdb75

PR #5838, opened 2021-01-19
Update highlight.js in textview

Zharktas:
### Proposed fixes:
Highlight.js library in textview is ancient and not under security maintenance. This PR updates the library to the latest version and adds it to package.json so that it might be updated later more easily. It uses cdn-assets package as that contains pre-built assets with common languages. The resulting JavaScript file is larger than the old one as it has more languages. If we want to support fewer languages, we could build our own highlight.js. 

Theme is changed to a11y-light for a11y compliance.

default.css is removed as it is not actually used.

### Features:

- [ ] includes API changes


pdelboca:
Tested it locally, LGTM! @Zharktas @amercader

PR #5846, opened 2021-01-20
return more datastore info

### LINKED ISSUE #5831: datastore_info does not return all available datastore info
**CKAN version**
2.9.1

**Describe the bug**
`datastore_info` _["returns information about the data imported, such as column names and types."](https://docs.ckan.org/en/2.9/maintaining/datastore.html#ckanext.datastore.logic.action.datastore_info)_.  For example:

```
$ ckanapi action datastore_info id=5ab4b4de-c970-4619-ab55-ce4338535b24 -r https://data.boston.gov
{
  "meta": {
    "count": 1268020
  },
  "schema": {
    "Agency": "text",
    "DateEntered": "text",
    "Developer": "text",
    "GeneralContractor": "text",
    "MINOR": "text",
    "ProjectAddress_1": "text",
    "ProjectName": "text",
    "ProjectNeigborhood": "text",
    "RESIDENT": "text",
    "Race_Desc": "text",
    "SEX": "text",
    "SubContractor": "text",
    "SubContractorAddress_1": "text",
    "SubContractorAddress_2": "text",
    "TotalHours": "text",
    "Trade": "text"
  }
}
```
However, it doesn't return other relevant available datastore information like aliases and the data dictionary.

Currently, to get an alias, you need to [query a special view `_table_metadata`](https://docs.ckan.org/en/2.9/maintaining/datastore.html#resource-aliases). To [get the data dictionary](https://docs.ckan.org/en/2.9/maintaining/datastore.html#data-dictionary), you need to call `datastore_search`.

**Expected behavior**
Instead of fetching datastore information for a given resource from three different areas, expand 
`datastore_info` so that it also returns the resource's aliases and its data dictionary in addition to what it returns now,

Having the actual data type as stored in the database, and the logical datatype as described by the data dictionary is also useful for Datastore API users.  Returning if a column is indexed is also useful for developers who are using `datastore_search_sql` so they can create performant SQL queries.

**Additional details**
This will facilitate implementing #5801.  It will also help CKAN Datastore API users who are using the datastore to create efficient reports/visualizations and extensions.

### LINKED ISSUE #5832: datastore_info does not work for aliases
**CKAN version**
2.9.1

**Describe the bug**
`datastore_info`  does not work for resource aliases.

**Steps to reproduce**
1. Create a resource
2. Use `datastore_create` to create aliases for the resource
3. Calling `datastore_info` with a valid alias results in a `NotFound` error

**Expected behavior**
`datastore_info` should also work for aliases.  Further, it should returned the aliased resource id.

jqnatividad:
Fixes #5831 and resolves #5832.

### Proposed fixes:
- `datastore_info` also works with aliases and can be used to dereference aliases.
- expands data returned by `datastore_info`
   - expanded `meta` dictionary.
     beyond rowcount, added aliases, resource_id (for dereferencing aliases), dbsize, resource size,  indices size, and table type (BASE TABLE, VIEW, FOREIGN TABLE, MATERIALIZED VIEW)
  - returns the data dictionary as a list of dictionaries - `fields` 
    - moved `schema` into `fields`, at the same time, adding the following keys - native_type (in addition to logical type above), index_name,  is_index, notnull, uniquekey
    -  this was done to align the results of `datastore_info` with `datastore_search`    

**EDIT**: Update proposed fixes after review & subsequent changes. Shorter example.

### Features:

- [X] includes API changes

### Example:
Note that ckanapi returns API results as sorted, pretty-printed JSON.
```
$ ckanapi action datastore_info id=test_alias -r=https://myckan.example.com
{
  "fields": [
    {
      "id": "e_field_id",
      "info": {
        "label": "equipment field id",
        "notes": "",
        "type_override": ""
      },
      "schema": {
        "native_type": "text",
        "index_name": null,
        "is_index": false,
        "notnull": false,
        "uniquekey": false
      },
      "type": "text"
    },
    {
      "id": "data_source_id",
      "info": {
        "label": "",
        "notes": "data_source_id information can be found in the data_source_master table",
        "type_override": ""
      },
      "schema": {
        "native_type": "numeric",
        "index_name": null,
        "is_index": false,
        "notnull": false,
        "uniquekey": false
      },
      "type": "numeric"
    },
    {
      "id": "adjustment_factor",
      "info": {
        "label": "",
        "notes": "",
        "type_override": ""
      },
      "schema": {
        "native_type": "numeric",
        "index_name": null,
        "is_index": false,
        "notnull": false,
        "uniquekey": false
      },
      "type": "numeric"
    },
    {
      "id": "start_date",
      "info": {
        "label": "",
        "notes": "",
        "type_override": ""
      },
      "schema": {
        "native_type": "timestamp without time zone",
        "index_name": null,
        "is_index": false,
        "notnull": false,
        "uniquekey": false
      },
      "type": "timestamp"
    },
    {
      "id": "end_date",
      "info": {
        "label": "",
        "notes": "",
        "type_override": ""
      },
      "schema": {
        "native_type": "timestamp without time zone",
        "index_name": null,
        "is_index": false,
        "notnull": false,
        "uniquekey": false
      },
      "type": "timestamp"
    }
  ],
  "meta": {
    "aliases": [
      "test_alias"
    ],
    "count": 1000027,
    "dbsize": 22266180255,
    "id": "3d9805b0-4eed-480c-8984-2ba2858dfab4",
    "idxsize": 165134336,
    "size": 304619520,
    "table_type": "BASE TABLE"
  }
}
```


wardi:
Thanks @jqnatividad 

Where possible would you align the keys returned to the ones that are returned by `datastore_search`? I don't love having duplicated functionality but at least if the keys are the same that reduces some strangeness in our API

jqnatividad:
Thanks @wardi 
Tweaked it to align the keys - using `native_type` for the native db-level data type and `table_type` for db table type.
Also changed some of the keys to use underscore spacing.
Let me know if there's anything else...

wardi:
in `package_search` the data dictionary entries are included in the `fields` list, can you do the same here instead of putting them in a separate `data_dictionary` list?

wardi:
Would you move the schema information into the fields list too? That way you don't need the `column_no` keys and all the information about a column would be in one place.

jqnatividad:
IMHO, leaving the two separate is best not only for backward compatibility but for "forward" compatibility as well.

The schema dictionary focuses on the "physical", native database metadata that's specific to the underlying datastore backend - in this case - postgres.

The data dictionary fields are focused on the "logical" field information that "should" stay the same even with a different backend.  

Also, even though I don't know of an extension that extends the Data Dictionary form template by leveraging how you used postgres comments to store the data dictionary JSON, wouldn't mixing the logical/physical info tie ourselves to postgres?

Right now, postgres is our primary backend, but I know of several initiatives to develop alternate backends.

Your thoughts?

wardi:
Not concerned about mixing data at different levels (we can group each in its own dict). Don't like that fields are represented as a list of objects and a dictionary with column names as keys at the same time.

As far as different datastore back ends it's up to them to implement as much of the API as they like.

jqnatividad:
Got it,

Will move schema info into the fields list, so:

```
{
  "fields": [
    {
      "id": "field",
      "info": {
        "label": "field name",
        "notes": "",
        "type_override": ""
      },
      "type": "text",
      "schema": {
           "data_type": "text",
           "index_name": null,
           "is_index": false,
           "notnull": false,
           "uniquekey": false
    },
...
    {
      "id": "stddev",
      "info": {
       "label": "Standard Deviation",
       "notes": "",
       "type_override": ""
    },
    "type": "numeric",
    "schema": {
        "data_type": "numeric",
        "index_name": null,
        "is_index": false,
        "notnull": false,
        "uniquekey": false
    },
...
```

Is this OK?

wardi:
Yes that sounds good, we'll just need a note in the changelog about the change in format of what this endpoint returns

wardi:
there's a test failure building the docs: `/root/project/ckanext/datastore/logic/action.py:docstring of ckanext.datastore.logic.action.datastore_info:24: WARNING: Field list ends without a blank line; unexpected unindent.` otherwise this looks good to me

jqnatividad:
Hi @wardi,
The apidoc is now fixed and I also aligned the result as we discussed earlier - https://github.com/ckan/ckan/pull/5846#issuecomment-765663693

wardi:
@amercader this looks good to me, but it does represent a change in API response format for `datastore_info`. Mentioning this change in the changelog should suffice, WDYT?

amercader:
I think the changelog should be enough, but add it as part of this PR or we might forget

wardi:
we're doing that with https://pypi.org/project/towncrier/ now, right? I don't see a `newsfragments` directory in our source

amercader:
We are putting the changelog entries here: https://github.com/ckan/ckan/tree/master/changes (or not putting rather, we haven't been following this, so we might need to enforce it with an action or re-think it)

jqnatividad:
Added the towncrier entries.

FWIW, +1 to using towncrier. To ensure compliance, may I suggest adding it to the ["Making a Pull Request"](https://docs.ckan.org/en/latest/contributing/pull-requests.html) checklist?

jqnatividad:
Pinging to check when/if this PR will be merged.
Was hoping to leverage the additional info in revamped datatatables view to:
* show resource aliases
* to get index info to help extended version of datatables_view that uses datastore_search_sql instead of datastore_search

wardi:
Love it @jqnatividad . Maybe even worth creating a place for this information in a "storage details" tab or something on the resource pages

PR #5853, opened 2021-01-25
Check for valid postgres identifer length

### LINKED ISSUE #5834: Postgres check for valid identifiers does not check for length
**CKAN version**
2.9.1

**Describe the bug**
The  postgres backend has several functions/helpers that check for valid postgres identifiers - for valid table, column, index and view (alias) names - that's used by the different datastore API actions.

However, it doesn't detect when an identifier is greater than postgres' maximum identifier length (63 by default - `NAMEDATALEN-1` - unless postgres is compiled with a different NAMEDATALEN setting).

https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

This results in truncated identifiers when database objects are created with overly long identifiers, sometimes resulting in errors.

**Steps to reproduce**
1. Create a resource
2. Create an alias that's longer than 63 characters with `datastore_create`
3. Check `_table_metadata` and the alias is truncated at 63 characters

or 

1. Prepare a CSV file with several column names longer than 63 characters
2. Upload it to CKAN
3. Datapusher error because `datastore_create` fails

**Expected behavior**
The datastore API should not fail silently and pass the long identifiers to postgres which blindly truncates at 63 characters. It should explicitly return an "identifier too long" error.

jqnatividad:
Fixes #5834 

### Proposed fixes:
* add len check to is_valid_field_name

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/helpers.py
+++ b/ckanext/datastore/helpers.py
@@ -26,10 +26,12 @@ def is_valid_field_name(name):
     * can't start with underscore
     * can't contain double quote (")
     * can't be empty
+    * can't be longer than 63 characters
     '''
     return (name and name == name.strip() and
             not name.startswith('_') and
-            '"' not in name)
+            '"' not in name and
+            not len(name) > 63)
 
 
 def is_valid_table_name(name):
--- a/ckanext/datastore/helpers.py
+++ b/ckanext/datastore/helpers.py
@@ -26,12 +26,10 @@ def is_valid_field_name(name):
     * can't start with underscore
     * can't contain double quote (")
     * can't be empty
-    * can't be longer than 63 characters
     '''
     return (name and name == name.strip() and
             not name.startswith('_') and
-            '"' not in name and
-            not len(name) > 63)
+            '"' not in name)
 
 
 def is_valid_table_name(name):
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -47,6 +47,9 @@ _pg_types = {}
 _type_names = set()
 _engines = {}
 
+_max_identifier_length = 63
+_db_enconding = 'UTF8'
+
 _TIMEOUT = 60000  # milliseconds
 
 # See http://www.postgresql.org/docs/9.2/static/errcodes-appendix.html
@@ -656,6 +659,20 @@ def _is_valid_pg_type(context, type_name):
         else:
             return True
 
+def is_valid_pg_identifier(name):
+    '''
+    Check that identifier is valid:
+    * can't start or end with whitespace characters
+    * can't start with underscore
+    * can't contain double quote (")
+    * can't be empty
+    * can't be longer than max_id_len per _db_enconding
+    '''
+    return (name and name == name.strip() and
+            not name.startswith('_') and
+            '"' not in name and
+            not len(name.encode(db_encoding)) > _max_identifier_length)
+
 
 def _execute_single_statement(context, sql_string, where_values):
     if not datastore_helpers.is_single_statement(sql_string):
@@ -757,7 +774,7 @@ def check_fields(context, fields):
                 'fields': [u'"{0}" is not a valid field type'.format(
                     field['type'])]
             })
-        elif not datastore_helpers.is_valid_field_name(field['id']):
+        elif not is_valid_pg_identifier(field['id']):
             raise ValidationError({
                 'fields': [u'"{0}" is not a valid field name'.format(
                     field['id'])]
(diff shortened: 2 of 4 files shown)


ckanext/datastore/helpers.py
```diff
@@ -26,10 +26,12 @@ def is_valid_field_name(name):
     * can't start with underscore
     * can't contain double quote (")
     * can't be empty
+    * can't be longer than 63 characters
     '''
     return (name and name == name.strip() and
             not name.startswith('_') and
-            '"' not in name)
+            '"' not in name and
+            not len(name) > 63)
```
wardi, line 34: Does postgres count character length the same way we do in python?

Does the encoding of the datastore database affect this calculation?

jqnatividad: I believe so.

In my testing, I encountered this problem when I was automatically creating aliases when a resource was created using the template "ownerorg-packagename-resourcename"(https://github.com/ckan/ideas/issues/264), and sometimes, it will create aliases longer than 63 characters and the resulting aliases are truncated at 63 characters by postgres.

Also, there is another area where the datastore checks for 63 characters already - 
https://github.com/ckan/ckan/blob/cc000d3acf7401957d8321be53e8cabc2d9ebf3a/ckanext/datastore/backend/postgres.py#L863-L867

However, it only does it for field names. Not table or alias names.

wardi: Both might be wrong. It sounds like the identifier length is limited to 63 bytes, so we either need to know the encoding of the DB  or we should do a  `select octet_length(...)` on the strings we plan to use field or alias names.

jqnatividad: Since postgres UTF8 encoding is already required, and the code has the `# encoding utf-8` magic comment, won't doing `len()` be good enough?

If not, I'd go for `len(name.encode("utf-8"))` instead of querying postgres for `octet_length(...)`.

wardi: where is postgres utf-8 encoding required? There's nothing preventing a user from setting the datastore encoding to cp1252 to increase performance or space-efficiency for their data is there?

jqnatividad: The docs **_imply_** a utf-8 requirement:

https://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html?highlight=UTF#setup-a-postgresql-database
https://docs.ckan.org/en/latest/maintaining/datastore.html?highlight=UTF#set-up-the-database

There is also a hardcoded utf8 assumption in the postgres backend.
https://github.com/ckan/ckan/blob/cc000d3acf7401957d8321be53e8cabc2d9ebf3a/ckanext/datastore/backend/postgres.py#L555-L557

Regardless, the docs should be updated to state if utf8 encoding is actually not required, and maybe even the tests updated for alternate encodings.

wardi: This `encode` is used to generate a hexadecimal hash, it doesn't affect any postgres encoding.

In other issues we're talking about letting users create their own materialized views or foreign data wrappers or other objects. It's not unreasonable for a user to want an encoding optimal for their use case.

```sql
postgres=# create database cp1252test encoding 'win1252' lc_ctype='C' lc_collate='C' template=template0;
CREATE DATABASE
cp1252test=# create table t1 ( ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ text );
NOTICE:  identifier "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ" will be truncated to "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþ"
```

vs

```
utf8test=# create table t1 ( ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ text );
NOTICE:  identifier "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ" will be truncated to "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞ"
CREATE TABLE
```

jqnatividad: Point taken!  Especially as someone who's asking for more advanced postgres support. :)

And speaking of advanced postgres support, perhaps, instead of hardcoding it, get it from postgres directly with `pg_control_init()` now that postgres v9.5 is about to [get EOLed Feb 11, 2021](https://endoflife.software/applications/databases/postgresql).

https://pgpedia.info/p/pg_control_init.html

wardi:
the postgresql encoding names don't match the ones in Python
```sql
cp1252test=# SELECT character_set_name FROM INFORMATION_SCHEMA.CHARACTER_SETS;
 character_set_name 
--------------------
 WIN1252
(1 row)
```

```python
>>> 'hi'.encode('WIN1252')
  File "<stdin>", line 1, in <module>
LookupError: unknown encoding: WIN1252
```

Doing a `select octet_length(...) < max_identifier_length` for each column name would be the safest. You should be able to pass them as an array and get the result back with a single query.

pdelboca:
Closing due to inactivity.

Feel free to reopen it if consider that it is still relevant.

PR #5856, opened 2021-01-26
Cast group and orgs max_limit config values to ints.

tomeksabala:
Fixes # Type error while comparing integer limit with string max_limit values provided in ckan.ini config.
I've encountered an error coming from core ckan code when accessing organizations list with http://ckan:5000/organization:
```
    File "/usr/lib/ckan/venv/src/ckan/ckan/logic/action/get.py", line 567, in organization_list
    return _group_or_org_list(context, data_dict, is_org=True)
    File "/usr/lib/ckan/venv/src/ckan/ckan/logic/action/get.py", line 374, in _group_or_org_list
    if limit is None or limit > max_limit:
    TypeError: '>' not supported between instances of 'int' and 'str'
```
The relevant piece of action/get.py:374 looks like this:
```
# ckan/ckan/views/group.py
    if all_fields:
        # all_fields is really computationally expensive, so need a tight limit
        max_limit = config.get(
            'ckan.group_and_organization_list_all_fields_max', 25)
    else:
        max_limit = config.get('ckan.group_and_organization_list_max', 1000)
    if limit is None or limit > max_limit:
        limit = max_limit
```
With debugger I've double checked that max_limit is actually of type `str`.
After I've casted max values to ints organization list loads withour any errors. Everything works fine too if used with python2, bug is present only with python3.
My ckan.ini config for those variables looks like this:
```
# ckan.ini
ckan.group_and_organization_list_all_fields_max = 1000
ckan.group_and_organization_list_max = 1000
```
### Proposed fixes:
Cast `config.get` values to integers. Also I've created two tests for regression.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -350,10 +350,11 @@ def _group_or_org_list(context, data_dict, is_org=False):
 
     if all_fields:
         # all_fields is really computationally expensive, so need a tight limit
-        max_limit = config.get(
-            'ckan.group_and_organization_list_all_fields_max', 25)
+        max_limit = int(config.get(
+            'ckan.group_and_organization_list_all_fields_max', 25))
     else:
-        max_limit = config.get('ckan.group_and_organization_list_max', 1000)
+        max_limit = int(config.get(
+            'ckan.group_and_organization_list_max', 1000))
     if limit is None or limit > max_limit:
         limit = max_limit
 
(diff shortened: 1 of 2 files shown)


amercader:
I think this is a duplicate of #5830 Can you confirm it @tomeksabala ?

tomeksabala:
@amercader Yes it is, thanks for pointing it out. I have left a comment as I think it might be useful to copy over my regression tests to #5830

tomeksabala:
duplicate of #5830

PR #5864, opened 2021-01-29
[#5670] Return content-type header on downloads if mimetype is present

### LINKED ISSUE #5670: Embed file in another site problem with mime type
**CKAN version**
ckan_version": "2.9"

**Describe the bug**
Embed file in another site from ckan force to download because mime type is incorrect, always is embed with Content-Type: application/octet-stream

**Steps to reproduce**
Load PDF in ckan and use embed to display in another site, when you load page with embed file try download and not display pdf.

**Expected behavior**
Display embed pdf

amercader:
Fixes #5670

### DIFF
--- a/ckan/tests/controllers/test_package.py
+++ b/ckan/tests/controllers/test_package.py
@@ -1064,6 +1064,7 @@ class TestResourceNew(object):
                 status=403,
             )
 
+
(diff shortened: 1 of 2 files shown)


pdelboca:
Tested it locally and changes looks good @amercader ! :)

amercader:
This worked as expected on 2.7 / 2.8

Issue #5855, opened 2021-01-26
Add datastore_register to Datastore API


jqnatividad:
**CKAN version**
2.9.1

**Describe the bug**
With the datastore, CKAN automatically creates tables for tabular data.

With `datastore_create`, you can even declare indices, aliases, triggers and primary keys, and the API transparently handles all the underlying SQL DDL to create/alter/delete & maintain the table.

However, this limits the datastore from exploiting underlying postgres database management features:
* everything in the datastore is in one tablespace
* data type guessing is not bullet-proof
* cannot use advanced clauses in the SQL DDL for tables/index/view creation/deletion/maintenance
* cannot take advantage of advanced features like declarative table partitioning, foreign tables, materialized views, etc. 

This limits the CKAN Datastore to "medium-sized" (by enterprise standards) databases, as there is no way to use database administration techniques on the datastore.

**Expected behavior**
A new endpoint is added to the Datastore API - `datastore_register`.  API users can choose to use the classic `datastore_create` or `datastore_register` to create a datastore resource.

The main difference is that the postgres DDL is created ahead of time by the API user, before calling `datastore_register` to "register" an existing postgres object (may it be a view, a materialized view, a foreign table, or a partitioned table, etc, so long as that it looks like a table) with CKAN. 

For the most part, all other datastore API actions will work the same as before (except instead of calling `datastore_delete`, the user calls `datastore_unregister` for previously registered postgres objects).

Here's the prospective `datastore_register` apidoc:

> ckanext.datastore.logic.action.datastore_register(context, data_dict)
> Registers an existing "table-like" postgres object with the Datastore.
> 
> Parameters:
> * resource_id (string) - id of the postgres object to register as a resource.
> * force (bool (optional, default: False)) - set to True to edit a read-only resource
> * resource (dictionary) - resource dictionary that is passed to `resource_create()`. Use instead of `resource_id` (optional)
> * aliases (list or comma separated string) - names for read only aliases of the resource (optional)
> * fields (list of dictionaries) - fields/column and their extra metadata (optional)
> * calculate_record_count (bool (optional, default: False)) – updates the stored count of records, used to optimize datastore_search in combination with the total_estimation_threshold parameter. If doing a series of requests to change a resource, you only need to set this to True on the last request.

Issue #5847, opened 2021-01-21
Overly aggressive indexing strategy greatly increases datastore storage requirements


jqnatividad:
**CKAN version**
2.9.1

**Describe the bug**
The postgres datastore creates several indices when a resource is uploaded.

A unique index for the `_id` internal field, an FTS index using the `_full_text` internal field (which stores all the column values in one field for FTS searches, effectively doubling the width of the table), and **_one index per text column._**

Though this is great for ad-hoc queries using `datastore_search` and `datastore_search_sql`, and for interactive filtering on the UI, it greatly increases the datastore's storage requirements.

For example, a table with 1.37 million rows and 9 columns is:

- **60 mb** as a csv file
- **302 mb** for the database table
- **569 mb** for the 8 indices (unique index for `_id`, 1 fts index, and 6 indices for the 6 text columns. It did not create indices for the one timestamp and two numeric columns) 

I got this info while testing the [expanded `datastore_info` PR.](https://github.com/ckan/ckan/pull/5846)

**Expected behavior**
The team should consider giving finer granular control to the CKAN administrator beyond [specifying the index method](https://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-datastore-default-fts-index-method).

Some indexing "knobs" to consider:
* turn on/off FTS index
* leverage the data dictionary to explicitly declare columns that should be indexed. As an added benefit, this should also allow indices for non-text columns
* or alternatively, instead of creating a mini DBA interface on CKAN (a non-trivial task), just read the postgres system catalog and  selectively show some indexing info on the data dictionary (leveraging the expanded `datastore_info`), and let the DB/CKAN administrator use robust tools like PgAdmin to manage datastore indices manually.  This is consistent with the keeping the Datastore API _["as thin as possible to allow you to use the features you would expect from a powerful database management system."](https://docs.ckan.org/en/2.9/maintaining/datastore.html#the-datastore-api)_

wardi:
:+1: to using the data dictionary interface to enable indexes and having the per-column indexes disabled by default

wardi:
1. Update `datastore_create` to default to not add any indexes when the `index` list of fields is empty or not provided.
2. Show which fields are indexed in the data dictionary and make it possible to add or remove indexes for field types that support indexing. Future: we might want to be able to select which columns are included in the full text search index here too.
3. For large datasets creating indexes will take a long time, we should use a background job to run the datastore_create call so that the web UI doesn't time out returning a response to the user.

[...]

PR #5867, opened 2021-02-01
Check for min postgres version

### LINKED ISSUE #5854: Track Postgres EOL lifecycle
**CKAN version**
2.9.1

**Describe the bug**
The current minimum postgres version for CKAN is 9.2.  Postgres 9.2 is no longer supported.

**Expected behavior**
The current release should ideally always specify a supported Postgres version.  Similar to how CKAN tracks the Ubuntu LTS version, perhaps, there should also be a CKAN tracking policy consistent with Postgres' EOL lifecycle.

https://endoflife.software/applications/databases/postgresql

jqnatividad:
Partly resolves #5854.

### Proposed fixes:
Check for minimum postgres version in datastore postgres backend.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1742,6 +1742,10 @@ class DatastorePostgresqlBackend(DatastoreBackend):
         if rows_max is not None:
             int(rows_max)
 
+        if not _pg_version_is_at_least(self.read_engine, "9.5"):
+            error_msg = 'At least PostgreSQL v9.5 is required.'
+            raise DatastoreException(error_msg)
+
     def datastore_delete(self, context, data_dict, fields_types, query_dict):
         query_dict['where'] += _where_clauses(data_dict, fields_types)
         return query_dict
--- a/doc/maintaining/datastore.rst
+++ b/doc/maintaining/datastore.rst
@@ -47,7 +47,7 @@ Setting up the DataStore
 .. versionchanged:: 2.6
 
    Previous CKAN (and DataStore) versions were compatible with earlier versions
-   of |postgres|.
+   of |postgres|. Currently, at least |postgres| version 9.5 is required.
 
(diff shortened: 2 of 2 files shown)


Zharktas:
I was thinking that it is up to the codebase to enforce minimum requirements if the new features aren't needed? The codebase should support newer versions so it would not limit upgrading, but if there aren't anything that requires the newer version, should it block upgrading ckan ?

jqnatividad:
Good point @Zharktas . Still, what is the team's thinking on Postgres v9.5, which will be EOLed by next week (Feb 11)?

Should CKAN support EOLed Postgres servers? 

Along those lines, IMHO, the team should actually consider specifying Postgres v10.15 (the bundled version with Ubuntu LTS 18.04) in master, and skip v9.6 altogether, as its scheduled for EOL Nov 2021.

I've been doing a lot of work on the datastore of late because the project I'm working on has very large datastore resources, and it has become apparent that the current datastore, though well suited to "open data" publishing use cases, is ill suited for enterprise deployments (#5847) and streaming data (for IoT and Smart City use cases).

I'm actually looking forward to Postgres v10, as I can envision a fairly straightforward way to take advantage of declarative table partitioning and query parallelism with the Datastore.

At the same time however, one motivation for `datastore_register` (#5855) is to "decouple" the Datastore from Postgres a bit more so the CKAN sysadmin can independently take advantage of advanced Postgres features and apply DBA management techniques.  Even db features beyond postgres v10.

So yes, going back to your question, I feel that the codebase should always specify a supported version of Postgres even if the new features are not currently used, as I do plan to take advantage of postgres v10 features as its "necessary" for my current implementation (using time-based partitioning for large datasets, and using foreign tables instead of pushing very large tables into the datastore).

Also, as we migrate to Python 3, Flask, etc. I think it's the best time to modernize Postgres as well :)

PR #5878, opened 2021-02-05
fix email validation regexp pattern

mutantsan:
I was working on some issue with inviting user to group and found out that email validation is weird. `test@example` - this email is invalid, but it passes the validation.

### Features:

- [ ] includes API changes


wardi:
We can't validate all possible invalid emails so we went with a regexp that matches the HTML5 email input validation instead. Also it is possible to have internal email addresses without a dot depending on the DNS settings in your environment.

PR #5879, opened 2021-02-05
purge user if invitation fails

mutantsan:
Curently, if an error occurs during the invitation to group, the user state will be changed to 'deleted. The problem will appear on second try. Because user was already created we will face the validation error: `The email address 'examle@gmail.com' belongs to a registered user.`

I propose to purge these users, because there is no sense to keep them. As I understand, there is no data associated with such user that could prevent us from purging them. Also, the validation error message states that the user was not created, but it was actually :)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1168,10 +1168,9 @@ def user_invite(context, data_dict):
     try:
         mailer.send_invite(user, group_dict, data['role'])
     except (socket_error, mailer.MailerException) as error:
-        # Email could not be sent, delete the pending user
-
-        _get_action('user_delete')(context, {'id': user.id})
-
+        # Email could not be sent, purge the pending user
+        user.purge()
+        user.commit()
         msg = _('Error sending the invite email, ' +
                 'the user was not created: {0}').format(error)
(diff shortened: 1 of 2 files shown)


pdelboca:
Looks good @amercader @wardi!

pdelboca:
When calling `_get_action('user_delete')` we are also deleting all memberships of this user. 

This implementation, even when it purges the user correctly, it doesn't delete the memberships. This leaves the group page with a lot of non-user members:
image

So we should find a way to delete this membership if failing. Maybe implement a `user_purge` action?

@mutantsan @amercader @wardi

pdelboca:
Are you able to keep working on this @mutantsan ? The goal will be to implement a `user_purge` action that not only purges the user but also cleans all the memberships.

mutantsan:
@pdelboca, sorry,  we have a little problem in our country. I would like to continue, but I'm not ready yet.

pdelboca:
@mutantsan don't even worry about this PR. I was reviewing my assigned tickets and wanted to do a quick follow up. Didn't knew you were in Ukraine.

All my best wishing of peace :peace_symbol: and reach out if you need anything!

Issue #5849, opened 2021-01-21
Data Dictionary doesn't update field type permanently


Engerrs:
**CKAN version**
2.9.1

**Describe the bug**
Due to the fact that after an update to CKAN 2.9, when User uploads file, all column that are send to Datastore are being saved as "text" (CSV files example). If we setup to use Messytables for that, it parses the file incorrectly if the file has some additional "," in one of the columns, as a result we have broken types setup for column.
That's why the most appropriate feature to use is Data Dictionary, but the logic of an update can be a headache. For example, if User uploads CSV file (1000mb) and wants to update one of the columns type, according to the current logic, he/she should re-upload the whole file (not to re-upload to Datastore) and there is a tricky moment that the file should be in someway modified as ckanext-xloader compares file hash and is case if files match, it won't do anything at all. Also re-uploading to large files take a big amount of time.

**Expected behavior**
When updating column type in Data Dictionary, changes should appear permanently.

Note:
I already prepared a PR to CKAN, will upload it after issue created.

Engerrs:
Hi @wardi , as far as I understand, you are the person who worked with indexes in DB. Can you please review my PR that is above this message.
Note:
Before changing column from text type, I remove index from the column. If User changes back to text type, CKAN according to the logic will bring the index back.

PR #5886, opened 2021-02-08
[#5849] Update for Data Dictionary save logic

Engerrs:


kowh-ai:
Hi @Engerrs  - Could you please provide a description for this PR please?

Engerrs:
Hi @kowh-ai , there is actually a link to the described issue and this PR is a solution for it. (https://github.com/ckan/ckan/issues/5849)

kowh-ai:
Thanks @Engerrs. @wardi are you able to review Yan's code in this PR. Iv'e taken a look and it's really a little beyond my level of understanding to formally review - thanks

wardi:
Converting the types of fields in-place can be a very expensive operation for large tables, this could even time out making it impossible to submit changes to the data dictionary. In-place type conversions could cause errors that aren't handled here. Successive in-place type conversions could also lose data making the data loaded into the datastore not match what would be there if loaded from csv again.

Those are some of the reasons type_override is implemented in xloader/datapusher, not in datastore itself. xloader/datapusher are already set up to run as separate jobs, always use the source csv so they won't lose data, and can handle and log errors gracefully.

I do like the feature. Maybe it can be implemented as a new API call or background job?

Engerrs:
Hi @wardi , well I can think about it more. New API call sounds reasonable, DEV's can implement their custom view with form that can call the new API endpoint.

wardi:
altering the table might also make sense when setting the types on the columns to a `datastore_create` call instead of the `type_override` in the data dictionary. Right now the types are ignored on existing columns which is a little weird. Altering the table would be better behavior.

An extension could make that call in the background when a user changes the `type_override` in the data dictionary to get the same functionality without the timeout issues.

PR #5892, opened 2021-02-09
Fix edit organization member bug

styhar:
There is a bug when user tries editing any member of existing organization.
Hitting "edit" button near the organization member gives "Add new member" functionality instead of "Edit member" functionality. This change fixes it.


kowh-ai (approved):
I have just tested this change and it looks good!

Screen Shot 2021-02-09 at 4 33 54 pm

PR #5910, opened 2021-02-22
[#5909] fix view templates with custom resource_view auth

### LINKED ISSUE #5909: fix view templates with custom resource_view auth
**CKAN version**
master

**Describe the bug**
If the `resource_view_create` and `resource_view_delete` auth is restricted with an IAuthFunctions plugin the UI still presents create/edit forms to users without access.

**Steps to reproduce**
create an IAuthFunctions plugin like
```python
     def get_auth_functions(self):
        return {
            'resource_view_create': auth.resource_view_create,
            'resource_view_update': auth.resource_view_update,
            'resource_view_delete': auth.resource_view_delete,
        }
```

and restrict view editing to sysadmin only
```python
def resource_view_create(context, data_dict):
    return {'success': False}

def resource_view_update(context, data_dict):
    return {'success': False}

def resource_view_delete(context, data_dict):
    return {'success': False}
```

**Expected behavior**
Create/edit view should be hidden in the UI

**Additional details**
UI is present but leads to server error when submitted (a separate issue, or should it be addressed here too?).

wardi:
Fixes #5909

### Proposed fixes:
use `check_access` to hide view editing UI when user doesn't have permission

### Features:

- [ ] includes API changes


wardi:
@Zharktas fixed the test failure

PR #5914, opened 2021-02-24
Ccw patch#1

wood-chris:
Fixes #

### Proposed fixes:

Currently, `/api/3/action/status_show` returns a 500 error if there are no extensions, because `split()` is called on `None`. This fixes that issue.

Caveat: it's fixed in `ckan/logic/action/get.py`, which is the NoneType is exposed - the return from `load(*plugins)` in `ckan/plugins/core.py` is

```
[<Plugin DomainObjectModificationExtension 'domain_object_mods'>, <Plugin SynchronousSearchPlugin 'synchronous_search'>]
```

so it seems these 2 plugins are removed from the output somewhere, but I didn't know where. It might be better to fix the issue at this point, rather than in `get.py`

### Features:

- [X] includes API changes


amercader:
@wood-chris looks like your local master was out of date. I doubt this change involves 448 files :)

wood-chris:
Yeh, sorry - I realised this when I PRd. I'll fix soon :)

wood-chris:
Yes, sorry - I realised that when I submitted the PR! I'll fix it and resubmit :)

wood-chris:
Yes, sorry - I realised that when I submitted the PR! I'll fix later and resubmit :)

PR #5916, opened 2021-02-24
Fix status_show when no extensions are installed

wood-chris:
Fixes #

### Proposed fixes:

Currently, `/api/3/action/status_show` returns a 500 error if there are no extensions, because `split()` is called on `None`. This fixes that issue.

Caveat: it's fixed in `ckan/logic/action/get.py`, which is where the NoneType is exposed; however, the return from `load(*plugins)` in `ckan/plugins/core.py` is actually

```
[<Plugin DomainObjectModificationExtension 'domain_object_mods'>, <Plugin SynchronousSearchPlugin 'synchronous_search'>]
```

so it seems these 2 plugins are removed from the output somewhere, but I didn't know where. It might be better to fix the issue at this point, rather than in `get.py`. Feel free to reject if you think this would be a better option

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2416,7 +2416,7 @@ def status_show(context, data_dict):
     '''
 
     plugins = config.get('ckan.plugins') 
-    extensions = "" if plugins == None else plugins.split()
+    extensions = plugins.split() if plugins else [] 
 
     return {
         'site_title': config.get('ckan.site_title'),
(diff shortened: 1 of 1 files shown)


ckan/logic/action/get.py
```diff
@@ -2414,14 +2414,18 @@ def status_show(context, data_dict):
     :rtype: dictionary
 
     '''
+
+    plugins = config.get('ckan.plugins') 
+    extensions = "" if plugins == None else plugins.split()
```
wardi, line 2419:
```suggestion
    extensions = plugins.split() if plugins else [] 
```
we should always store a list

PR #5926, opened 2021-03-03
fix: replace nonexistent UWSGI option "guid" with "gid"

paulmueller:
The "guid" option does not exist. I also added the "strict" option to make sure such things don't happen again.

Reference:

https://uwsgi-docs.readthedocs.io/en/latest/Options.html#gid

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan-uwsgi.ini
+++ b/ckan-uwsgi.ini
@@ -2,7 +2,7 @@
 
 http            =  127.0.0.1:8080
 uid             =  www-data
-guid            =  www-data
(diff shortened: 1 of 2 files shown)


amercader:
Thanks @paulmueller , can you also update the documentation to reflect this. I think this is only mentioned here: https://github.com/ckan/ckan/blob/master/doc/maintaining/installing/deployment.rst

paulmueller:
Yes, Done!

PR #5940, opened 2021-03-05
Remove unused action call in download blueprint

pdelboca:
This call to `get_action(u'package_show')` is not used for anything in the download blueprint.

### DIFF
--- a/ckan/views/resource.py
+++ b/ckan/views/resource.py
@@ -165,7 +165,6 @@ def download(package_type, id, resource_id, filename=None):
 
     try:
         rsc = get_action(u'resource_show')(context, {u'id': resource_id})
-        get_action(u'package_show')(context, {u'id': id})
     except (NotFound, NotAuthorized):
         return base.abort(404, _(u'Resource not found'))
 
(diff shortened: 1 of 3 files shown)


ckan/views/resource.py
```diff
@@ -165,7 +165,6 @@ def download(package_type, id, resource_id, filename=None):
 
     try:
         rsc = get_action(u'resource_show')(context, {u'id': resource_id})
```
wardi, line 167: package_show was at least checking that the id passed is a valid dataset that the user has access to, this would be even better and less expensive:

```suggestion
        rsc = get_action(u'resource_show')(context, {u'id': resource_id})
        if rsc['package_id'] != id:
            raise NotFound
```

pdelboca: Great! I thought on a similar idea! I added a more explicit error raise with clear message.

ckan/views/resource.py
```diff
@@ -165,6 +165,11 @@ def download(package_type, id, resource_id, filename=None):
...
+        if rsc['package_id'] != id:
+            return base.abort(
+                    404,
+                    _(u'The resource found does not belong to the package.')
+                )
```
wardi, line 172: if we're adding a new message it makes sense to make the behaviour consistent across all the views that take both a package id/name and a resource id

PR #5952, opened 2021-03-09
Datatablesview revamp reprise

jqnatividad:
Thanks for merging the original datatablesview revamp!

After more testing, fixed some bugs and introduced new features.

### Proposed fixes:
 - responsive view (list mode) now works properly.
 - localStorage keys now expire properly using the `ckan.datatables.state_duration` configuration setting
 - reformatted JS file to https://standardjs.com guidelines
 - updated DataTables i18n files 
 - removed yacdf plugin which was causing UI/UX issues in column filters. Implemented column filters using DataTables primitives
 - more robust, graceful error handling when there are FTS query errors. DataTables expects server-side errors to be relayed in a specific way (https://datatables.net/manual/server-side)
 - filtered downloads only works properly on first invocation as the hidden div where parms where being passed to the server was not being cleared after each call (this bug existed before this revamp)
 - filtered downloads now also works with column filters
 - column filters did not work properly when data dictionary labels were used
 - bumped DataTables library to latest release - [v.1.10.24](https://cdn.datatables.net/1.10.24/) to fix problem with sorting.

### New Features:
 * **Better State Saving**  
   When `ckan.datatables.state_saving` is true (default):
 	- the view's settings are persisted for `ckan.datatables.state_duration` (default: 7200 seconds - 2 hours) for that resource view:
 	  - search keywords
 	  - column order/visibility
 	  - row selections
 	  - page settings
 	- "Share current view" button is enabled. This is actually a persistent _deeplink_ into the view. The state is persisted by Base64 encoding the state settings and passing it as a URL parm.
 	- "Reset" button is enabled. This "resets" the state (forget search keywords, row and page selections). There is now a subtle visual indicator with the button turning dark red when a state is active.

 * **View Modes:**
   - __'table' mode__ - the default DataTable grid layout, with column filters.
   - __'list' mode__ - aka Responsive mode, now also with column filters.  While fixing responsive view, actually discovered an additional use case to treat this as an alternate view mode optimized for viewing/copying/printing _record details_. 
   - view mode can be toggled at anytime using new 'Table/List toggle' button. 

 * **Data Dictionary Integration**
   - used column type to inform DataTables sorting (sorting is done on server-side, but still useful for DT to know data type)
   - use Label for column header if defined
   - for columns with Data Dictionary entries, a tooltip is displayed next to the Column name in the table header.

 * **UI/UX improvements**
   - i18n strings throughout
   - used FontAwesome for language-independent visual cues
   - User can now reorder Columns (drag-n-drop)
   - `_id` column behaves like Excel's row id (fixed column, darker shade, always available)
   - removed input placeholders and reset buttons for cleaner UI and to be more i18n friendly
   - all search fields now have input type search instead. This allows them to behave more like Google (search keywords are saved; cancel button, esc key works to clear input - eliminating need for separate reset button) in Webkit/Blink-based browsers (Chrome, Edge, Safari). In Firefox, search fields behave just like normal input fields.
   - Footer now has more info
     - __Resource Info__. Package and resource name linked to CKAN. Info button with hover info with resource info localised and using `ckan.display_timezone` settings (Data last updated, Metadata last updated, Created, file format and file size)
     - __Sort Info__. As DataTables supports multi-column sorting, this shows the field/sort order, and a tooltip to let user know how to do multi-column sorting.
   - __'Autofit columns' button__.
     DataTables does a good job auto-fitting column width, but sometimes, you need to manually fit it after a search (as it just auto-fits the column based on what's in memory)
   - __Smart resize__. On load and only when the viewport size is changed, the view will optimize page length to fit the current viewport so there is no unused whitespace between the tablebody and the footer.
  
     Further, the Header and Footer sections resize to be visible at all times, and are pinned to the top/bottom respectively. 
     This is especially useful when embedding the view, and viewing a deeplink share outside CKAN, while giving the user the ability to always click back to the portal with the footer Resource Info. 

   - **Filter visual cues** - Global filters and column filters are currently mutually exclusive (limitation because [`datastore_search`](https://docs.ckan.org/en/latest/maintaining/datastore.html#ckanext.datastore.logic.action.datastore_search)  `q` parameter does not support full FTS query **AND** per column search). This is indicated to the user by shading filter input boxes as necessary. When a filter is active, an information icon is displayed with a tooltip showing the filters in place.  Double-clicking the info icon is also a handy way of clearing filters.
   - **Filter and sort info** are included when printing/copying to clipboard. 

* **Under the hood, server-side improvements**
  - moved from legacy key names to latest DT api key names (i.e. iTotalRecords to recordsTotal, etc.)
  - convert data response from an [array to a JSON object](https://datatables.net/manual/server-side#Example-data).  This enables additional functionality in DataTables (e.g. Column reordering, being able to save row states, and lays groundwork to be able to use advanced DT features like [SearchPanes ](https://datatables.net/extensions/searchpanes/)and [SearchBuilder ](https://datatables.net/extensions/searchbuilder/)(once that is server-side ready))
  - add ajax timeout of 60 seconds

** updated Mar 15 to reflect latest changes now that it is ready for review

cc @smotornyuk @wardi 

- [ ] includes API changes


wardi:
comments from the meeting:
- style of buttons should match rest of ckan
- datatables column view overrides should be in the view configuration, not mixed with data dictionary description text field

Also can word wrapping be enabled by default for all columns to avoid overlapping text?

jqnatividad:
@wardi thanks for capturing the comments.

 * Will use bootstrap instead of DataTable's default style so the UI is more consistent with the rest of CKAN
 * Will add the DataTables configuration to the view configuration instead of the data dictionary

As for word-wrapping, I'll add it to the view configuration as well, with a default ckan.ini configuration value that can be overridden on a per resource view basis.

Was thinking of something like:
`ckan.datatables.wordwrapping=50`

which it will apply to all text columns, with the value being where the number of characters word-break check is done. 

setting it to 0 turns off word-wrapping if people want to optimize for more compact per row display.  And -1 if folks want DataTables to handle it altogether and decide for itself.

wardi:
the old datatables we're using on our site wraps to the column width, maybe that's -1 in the new configuration?

jqnatividad:
Yes. -1 is like the old datatablesview, which was letting DT decide, which is primarily on a column width basis.

I found it played it too safe though, resulting in fewer rows displayed per page, that's why I went for the compact display.

The problem with the compact layout, however,  is that DT's column width heuristics is sometimes thrown off by the variable spacing of proportional fonts.

Anyway, the good thing with your suggestion of storing this in the view configuration is the data owner can fine-tune it on a view-per-view basis, based on the display characteristics of the underlying resource, so they have the flexibility to stay with the default wrapping configuration, or override it for a specific column.

jqnatividad:
@amercader, as to your privacy concern about using localstorage to persist view settings and for the deeplink share (which is basically a base64 encoded version of the localstorage entry sent as a url parameter), I just wanted to confirm that its not saving column names - its using column index to send the datatables view configuration metadata.

```
{time: 1616079142202, start: 0, length: 20, order: [[0, "asc"]],…}
ColReorder: [0, 1, 2, 3, 4, 5, 6, 7]
columns: [{visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}},…]
0: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
search: {search: "", smart: true, regex: false, caseInsensitive: true}
caseInsensitive: true
regex: false
search: ""
smart: true
visible: true
1: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
2: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
3: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
4: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
5: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
6: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
7: {visible: true, search: {search: "", smart: true, regex: false, caseInsensitive: true}}
keyTable: {row: 4, column: 0, columnVisible: 0}
column: 0
columnVisible: 0
row: 4
length: 20
order: [[0, "asc"]]
0: [0, "asc"]
0: 0
1: "asc"
page: 0
pagelen: 20
search: {search: "", smart: true, regex: false, caseInsensitive: true}
caseInsensitive: true
regex: false
search: ""
smart: true
selected: [4]
start: 0
time: 1616079142202
viewmode: "list"
```

jqnatividad:
Confirming that this also fixes #5989 with the addition of the div removes at line 32-33

https://github.com/ckan/ckan/blob/4deceb560557e2f6eb7b8fe351f8e19f13d96368/ckanext/datatablesview/public/datatablesview.js#L22-L34

smotornyuk:
Whoa, it was quite a bit of reading and clicking. But it looks really stable so i'll click on merge button

PR #5956, opened 2021-03-11
add solr_timeout option

mutantsan:
We've recently encountered a problem with SOLR timeout. And seems like we don't have a possibility to configure timeout for SOLR connection. The default value in pysolr is 60 sec. So I've decided to add config option. I've added it to default production.ini template and to documentation also.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/search/common.py
+++ b/ckan/lib/search/common.py
@@ -9,6 +9,8 @@ import simplejson
 from six import string_types
 from six.moves.urllib.parse import quote_plus
 
+import ckan.plugins.toolkit as tk
+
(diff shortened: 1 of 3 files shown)


ckan/lib/search/common.py
```diff
@@ -77,11 +79,13 @@ def make_connection(decode_dates=True):
                                        quote_plus(solr_password),
                                        solr_url)
 
+    timeout = tk.asint(tk.config.get('solr_timeout', 60))
```
amercader, line 82: For core code is better to use core imports, in this case `from ckan.common import config, asint` and

PR #5969, opened 2021-03-18
Make datastore search whitelist case insensitive

ThrawnCA:
### Proposed fixes:

Make the datastore search SQL function whitelist case-insensitive, since PostgreSQL function calls are typically case-insensitive.

If case-sensitivity is desired, it can be achieved by enclosing a whitelist entry in "double quotes".

### Features:

- [X] includes bug fix for possible backport

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1588,7 +1588,10 @@ def search_sql(context, data_dict):
         context['check_access'](table_names)
 
         for f in function_names:
-            if f not in backend.allowed_sql_functions:
+            for name_variant in [f.lower(), '"{}"'.format(f)]:
+                if name_variant in backend.allowed_sql_functions:
+                    break
+            else:
                 raise toolkit.NotAuthorized({
                     'permissions': [
                         'Not authorized to call function {}'.format(f)]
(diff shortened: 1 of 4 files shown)


ThrawnCA:
I took the liberty of making the test module pass Flake8 checks while I was at it.

amercader:
@ThrawnCA many thanks for this. We don't merge new branches to old release ones, all pull requests should target `master`. Do you mind resubmitting it?

ckanext/datastore/backend/postgres.py
```diff
@@ -1511,7 +1511,10 @@ def search_sql(context, data_dict):
         context['check_access'](table_names)
 
         for f in function_names:
-            if f not in backend.allowed_sql_functions:
+            for name_variant in [f, f.lower(), '"{}"'.format(f)]:
```
amercader, line 1514: Isn't is better to normalize the input function name using `format_entry()` in here as well?

ThrawnCA: I think that would work, and might perhaps be slightly faster.

ThrawnCA:
> @ThrawnCA many thanks for this. We don't merge new branches to old release ones, all pull requests should target `master`. Do you mind resubmitting it?

Done. I've added a checklist entry for possible backporting.

amercader:
Thanks @ThrawnCA

PR #5970, opened 2021-03-18
Remove fanstatic

smotornyuk:
Remove fanstatic library from source. Without it `minify` and `front-end-build` CLI commands also won't work, so they are dropped as well. Actually, those commands had no effect, because webassets minifies JS and aggregates CSS by itself when compiling files into `webassets` folder.

As a future improvement, we can add an extra filter for CSS minification and experiment with JS filter, checking options other than rjsmin. But this if outside of this PR's scope


ckan/config/middleware/flask_app.py
```diff
@@ -65,6 +66,16 @@ def __call__(self, environ, start_response):
         return self.app(environ, start_response)
 
 
+class RepozeAdapterMiddleware(object):
+    def __init__(self, app):
+        self.app = app
+
+    def __call__(self, environ, start_response):
+        request = webob.Request(environ)
```
tino097, line 74: cant we use the werkzeug Request  instead webob?

amercader: +1, webob should be dropped soon

What does this middleware do anyway? AFAICT it just forwards the request

smotornyuk: Because we are using repoze.who which doesn't play well with werkzeug. And this adapter is required(previously Fanstatic had done it) in order to avoid BadRequests when our repoze.who interrupts unauthorized requests, dropping request payload while werkzeug tries to work with it.

PR #5976, opened 2021-03-22
replace uses of @helpers.change_config with @pytest.mark.ckan_config

chris48s:
### Proposed fixes:

Replace a couple of remaining uses of `@helpers.change_config` with `@pytest.mark.ckan_config`
Should the `change_config` function itself also be removed in this PR?

### Features:

- [ ] includes API changes


chris48s:
Gah this passed locally.
I will see if I can work out what is up with this and update the PR

chris48s:
Well that seems to get it passing, but I'm not 100% sure why that works :confused:

PR #5983, opened 2021-03-24
Upgrade select2 to 4.0 for issue 5697

### LINKED ISSUE #5697: Upgrade to Select2 4.x
It would allow to have custom selection (using `tags: true`) and other features: https://select2.org/upgrading/new-in-40
It would be also nice to include it as a git submodule, instead of copying the files over the ckan repository.

Sunny-NEC:
Fixes #5697 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/vendor/select2/bower.json
+++ b/ckan/public/base/vendor/select2/bower.json
@@ -1,6 +1,6 @@
 {
     "name": "select2",
-    "version": "3.5.4",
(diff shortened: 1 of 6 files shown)


Zharktas:
Changing version numbers within the library doesn't actually update anything in the library itself. What you need to is update the files themselves with newer version without modifying anything in them. And then do the migration to relevant parts in CKAN.

PR #5987, opened 2021-03-25
Allow to set path to INI file in the WSGI script

amercader:
The wsgi script that we include as example to be used with uwsgi or gunicorn expects a `ckan.ini` file in the same folder. This adds support for the `CKAN_INI` env var which is already supported when running the CLI.

### Features:

### DIFF
--- a/wsgi.py
+++ b/wsgi.py
@@ -4,9 +4,17 @@ import os
 from ckan.config.middleware import make_app
 from ckan.cli import CKANConfigLoader
 from logging.config import fileConfig as loggingFileConfig
-config_filepath = os.path.join(
-    os.path.dirname(os.path.abspath(__file__)), u'ckan.ini')
-abspath = os.path.join(os.path.dirname(os.path.abspath(__file__)))
-loggingFileConfig(config_filepath)
-config = CKANConfigLoader(config_filepath).get_config()
+
+if os.environ.get('CKAN_INI'):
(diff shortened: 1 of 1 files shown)


kowh-ai:
@amercader - This change seems to work OK when I run  `/usr/lib/ckan/default/bin/uwsgi -i /etc/ckan/default/ckan-uwsgi.ini` after setting the CKAN_INI environment to "somewhere else"...and in my case `/tmp/ckan.ini`. However it then has a problem trying to find `who.ini` in `/tmp` which I presume is fine (and expected). I'm just wondering if the uwsgi server is started via supervisor, the `CKAN_INI` environment variable would need to be included (somehow) in the shell running the command...Perhaps this should be included in the documentation somewhere also

kowh-ai (approved):
Looks good to go!

amercader:
Thanks @kowh-ai 


> However it then has a problem trying to find who.ini in /tmp which I presume is fine (and expected).

The path to who.ini is defined using `%(here)s` by default so you need to move the who.ini next to the ckan.ini file location or modify the path in the ini file


>  if the uwsgi server is started via supervisor, the CKAN_INI environment variable would need to be included

If you need to pass an env var to the supervisor process you can do so using the `environment` supervisor setting.

```
environment=CKAN_INI=/etc/ckan/ckan.ini
```

We can definitely add this to the deploy docs

PR #5995, opened 2021-03-26
Include extra files into ckanext distribution

smotornyuk:
Currently, 'webassets.yml` and all the alembic migration files are missing when source distribution created using the default setup.py\manifest.in. This PR includes those files and creates webassets.yml with commented basic examples of asset definition


amercader:
I literally had to fix this today in a couple of extensions! Great stuff

PR #6000, opened 2021-04-02
Allow multi-level config inheritance

smotornyuk:
Currently `use = config:...` option in the main config file allows us to parse referenced config file. But if this referenced file also contains a reference to yet another config file, the latest one will be ignored. I.e, it impossible to create chains of config files.

This PR provides the ability to extend config with an arbitrary depth(unless there is a recursion, in which case we stop as soon as we see a file that was already parsed)

### DIFF
--- a/ckan/tests/cli/test_cli.py
+++ b/ckan/tests/cli/test_cli.py
@@ -110,3 +110,19 @@ def test_ckan_config_loader_parse_two_files():
     assert conf[u'global_conf'][u'__file__'] == filename
     assert conf[u'global_conf'][u'here'] == tpl_dir
     assert conf[u'global_conf'][u'debug'] == u'false'
+
+
+def test_recursive_loading():
+    """Load chains of config files via `use = config:...`.
+
+    Make sure we still remember main config file.
+    If there are circular dependencies, don't fall into infinite recursion.
+    """
+    filename = os.path.join(
+        os.path.dirname(__file__), u'data', u'test-one.ini')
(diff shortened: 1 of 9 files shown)


pdelboca (approved):
Looks good @smotornyuk @amercader ! 

As a small optional improvement, I would add a debug to notify the user of the circular dependency before breaking the loop. It could become one of those silent failures hard to debug.

amercader:
I agree. In fact, isn't it better to fail with a clear exception if there is a circular dependency?

smotornyuk:
Yep, completely agree. Was just too lazy initially to write error message :)

ckan/tests/cli/test_cli.py
```diff
@@ -110,3 +111,26 @@ def test_ckan_config_loader_parse_two_files():
...
+    assert conf[u'key3'] == u'three'
+
+
+def test_recursive_loading():
+    """ Make sure we still remember main config file.
+
+    If there are circular dependencies, make sure the user knows about it.
+    """
+    filename = os.path.join(
+        os.path.dirname(__file__), u'data', u'test-one-recursive.ini')
```
amercader, line 134: Wrong name or missing file here

smotornyuk: Yeah, we have all *.ini files ignored by git and I forgot about it

amercader:
Nice one! :six: :zero: :zero: :zero:

PR #6008, opened 2021-04-09
[#4413] Fix Chinese locales

### LINKED ISSUE #4413: Flask-Babel does not translate zh_TW and zh_CN

≧ 2.8.0

When switching to ``zh_TW`` or ``zh_CN`` languages on the pages written in Flask (ex. ``/`` and ``/user``), the pages should be shown in corresponding languages.

It shows English instead of ``zh_TW`` or ``zh_CN``.

Switch language to Chinese on the above-mentioned pages.

amercader:
Fixes #4413

There's a fair amount of discussion in the issue but the main points are:

* Visiting a URL with the `zh_CN` and `zh_TW` locales does not translate the content.
* Babel / flask-babel expects folders named `zh_Hans_CN` and `zh_Hant_TW`. 
* We want to fix this, but keep old URLs like http://localhost:5000/zh_CN working (new one will be http://localhost:5000/zh_Hans_CN)
* Transifex expects `zh_CN` and `zh_TW`

So this PR does:

1. Rename the locale folders to what Babel expects
2. Sets up redirect routes for the old legacy locale names (eg http://localhost:5000/zh_CN/dataset ->
http://localhost:5000/zh_Hans_CN/dataset)
3. Adds an alias to the Transifex config file to be able to push/pull normally


spectatorcharlie:
Problem solved! Thanks for your help.

1. cp -r zh_CN zh_Hans_CN
2. cp -r zh_TW zh_Hant_TW
3. Go to localhost:5000
4. UI displays in correct language

PR #6020, opened 2021-04-14
Delete resource views if type changes

pdelboca:
Currently when a resource is updated and it changes it format, it triggers the creation of new default resources views but it doesn't clean the current ones.

This may cause that a resource that changed from PDF to CSV to maintain 2 non-compatible views.

### Proposed fixes:
Remove the current resource_views if the format changes since it is fair to assume that different resource types will show different types of views.

I tried to create a new action to delete all the views of a resource (so it can be reused) but `resource_view_clean` is already taken and I cannot define a good name for it :facepalm:. Suggestions are welcome!

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -110,11 +110,21 @@ def resource_update(context, data_dict):
     resource = _get_action('resource_show')(context, {'id': id})
 
     if old_resource_format != resource['format']:
-        _get_action('resource_create_default_resource_views')(
-            {'model': context['model'], 'user': context['user'],
-             'ignore_auth': True},
-            {'package': updated_pkg_dict,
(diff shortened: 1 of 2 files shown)


wardi:
Suggest we keep incompatible views in case they have been customized by the user. Instead we hide views on the resource page that don't apply to the current resource format. In the resource edit page we should also indicate that a view is hidden because of the format type to reduce confusion.

Also, what if a customized grid view exists and the format changes (e.g. csv->xlsx) will we get a new default grid view added that might not be expected?

If we had a way to detect that a view has not been customized it would be better to delete it, but that seems like a harder problem.

pdelboca:
I'm closing this PR since it fixes a specific use case in our instance that may not be easily migrated to Core.

PR #6044, opened 2021-04-22
Fix error when `search_facets_limits` not set in the global object

aramboi:
### Proposed fixes:

It seems that `search_facets_limits` is not always set in the global
object, as such `has_more_facets` crashes when it tries to access it.
This change implements the same check as in the `get_facet_items_dict`
function in order to deal with this issue.

### Features:

- [ ] includes API changes


wardi:
Looks like this could be hiding a larger problem.

Why does this helper use the global object to pass the limits? Could it be passed as a normal parameter instead? Is this helper only expecting to be called from a specific flask view, and if so why?

PR #6048, opened 2021-04-27
Do not rely on error message when selecting validator signature

smotornyuk:
The current implementation of validators relies on error messages when validators applied. 

Instead, it will be more obvious to check the number of arguments that are expected by the validator - after migration to py3 it's possible via `inspect.signature`.

### DIFF
--- a/ckan/lib/navl/dictization_functions.py
+++ b/ckan/lib/navl/dictization_functions.py
@@ -2,7 +2,6 @@
 
 import copy
 import json
-import inspect
 
 import six
 from six import text_type
--- a/ckan/lib/navl/dictization_functions.py
+++ b/ckan/lib/navl/dictization_functions.py
@@ -220,8 +220,12 @@ def augment_data(data, schema):
 
 
 def convert(converter, key, converted_data, errors, context):
-    signature = inspect.signature(converter)
-    nargs = len(signature.parameters)
+    try:
+        nargs = converter.__code__.co_argcount
+    except AttributeError:
+        raise TypeError(
+            f"{converter.__name__} cannot be used as validator "
+            "because it is not a user-defined function")
     if nargs == 1:
         params = (converted_data.get(key),)
     elif nargs == 2:
@@ -230,7 +234,8 @@ def convert(converter, key, converted_data, errors, context):
         params = (key, converted_data, errors, context)
     else:
         raise TypeError(
-            f"Wrong number of arguments(expected 1, 2 or 4): {nargs}")
+            "Wrong number of arguments for "
+            f"{converter.__name__}(expected 1, 2 or 4): {nargs}")
     try:
         value = converter(*params)
         # 4-args version sets value internally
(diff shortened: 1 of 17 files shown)


ckan/lib/navl/dictization_functions.py
```diff
@@ -219,35 +220,22 @@ def augment_data(data, schema):
...
+    nargs = len(signature.parameters)
+    if nargs == 1:
+        params = (converted_data.get(key),)
+    elif nargs == 2:
+        params = (converted_data.get(key), context)
+    elif nargs == 4:
+        params = (key, converted_data, errors, context)
+    else:
+        raise TypeError(
+            f"Wrong number of arguments(expected 1, 2 or 4): {nargs}")
```
wardi, line 233: we should indicate which function had the wrong number of arguments because it won't be obvious from the traceback

ckan/lib/navl/dictization_functions.py
```diff
@@ -219,35 +220,22 @@ def augment_data(data, schema):
...
-    except Invalid as e:
-        errors[key].append(e.error)
-        return
-    except TypeError as e:
-        # hack to make sure the type error was caused by the wrong
-        # number of arguments given.
-        if converter.__name__ not in str(e):
-            raise
-
+    signature = inspect.signature(converter)
```
wardi, line 223: does `inspect.signature` do a lot of work that we don't need to just get the number of arguments (i.e. `converter.__code__.co_argcount`)? We call this function many times when running validation, have you measured the performance impact of this change?

smotornyuk: I haven't used __code__ object because it's not available for built-ins. But apparently signature has the same limitations - it can't inspect functions like `str`. So I'll search for other options

smotornyuk: Hmm, it seems that my solution won't work because all the introspection tools work only win python-fuctions. Even though it may be rare and we can update all such cases in CKAN, in 3rd-party extension people still may want to use `str/list/dict` as validators, so we have to rely on the old way with TypeErrors till python will provide even-more stable option to inspect functions :(

wardi: dunder names like `__code__` aren't internals, they're just special names designed not to conflict with anything a user might add.  I'm happy to prevent things like `str` from being used as a validator because they can raise exceptions we're not expecting.

smotornyuk: Cool, in that case i'll replace such functions with simple wrappers and switch to `__code__` for checks.

BTW, `__code__` x3 times faster than `try\except` version

ckan/lib/navl/dictization_functions.py
```diff
@@ -2,6 +2,7 @@
 
 import copy
 import json
+import inspect
```
wardi, line 5: seems to be unused

Issue #5765, opened 2020-11-29
User following an Orgazation can't load any page


Engerrs:
**CKAN version**
CKAN 2.9

**Describe the bug**
When A User follows an Organization/Group with a lot of Datasets related to it, he/she unable to load the page. The issue start appearing at least when Org have 100 and more Datasets.

**Expected behavior**
As on CKAN 2.8 it worked well without any loading issues.

**Additional details**
I know that the activity model has been modified in CKAN 2.9, to be more exact here is the line that makes to most time consume https://github.com/ckan/ckan/blob/master/ckan/model/activity.py#L229

Couldn't prepare a PR to this one as I'm not sure how it should work properly so I suggest the author to look into to it.

alexandru-m-g:
At the link mentioned above there is a left outer join with an OR condition: 

1. `model.Package.id == model.Member.table_id` OR
1. `model.Package.owner_org == group_id`

Is (2) really needed ? What is the case where the owner_org in the package table is not reflected in the member table ? 

Also, isn't it a bit weird to have a join condition that doesn't connect to any other table ? We are looking for data from Activity table but how can (2) help with that if it's not connected to it ? And doesn't this just multiply the rows of the subquery ?

I'm asking because on our database this query never finishes. Removing (2) seems to solves the issue.

PR #6057, opened 2021-04-29
[#6028] Fix performance bottleneck in activities queries

### LINKED ISSUE #6028: Activity queries getting slow (~15 seconds) when following entities.
## CKAN version
Tested in 2.9.2 but probably also happening in master

## Describe the bug
`<lib/helpers.py:2567(new_activities)>` spikes to ~15 seconds if the user is following 2 or 3 organizations (with around 200 datasets each) in a ~1000+ dataset system. This makes the system unusable since this helper is called in almost every request to the UI.

**Without following organizations (50 ms):**
Calls | Total Time (ms) | Per Call (ms) | Cumulative Time (ms) | Per Call (ms) | Function
-- | -- | -- | -- | -- | --
1 | 0.005 | 0.0050 | 52.439 | 52.4390 | <lib/helpers.py:2567(new_activities)>
1 | 0.006 | 0.0060 | 52.374 | 52.3740 | <logic/action/get.py:3216(dashboard_new_activities_count)>
1 | 0.016 | 0.0160 | 50.164 | 50.1640 | <ckanext/gdx/logic/action.py:418(dashboard_activity_list)>

**Following 3 organizations with 100 datasets each (1500 ms):**
Calls | Total Time (ms) | Per Call (ms) | Cumulative Time (ms) | Per Call (ms) | Function
-- | -- | -- | -- | -- | --
1 | 0.01 | 0.0100 | 1610.417 | 1610.4170 | <lib/helpers.py:2567(new_activities)>
1 | 0.018 | 0.0180 | 1610.26 | 1610.2600 | <logic/action/get.py:3216(dashboard_new_activities_count)>
1 | 0.031 | 0.0310 | 1605.896 | 1605.8960 | <ckanext/gdx/logic/action.py:418(dashboard_activity_list)>

## Steps to reproduce
- Create 3 organizations
- Add ~100 datasets to each one (enough to test)
- Login with a user
- Open the home page
- Using flask debug toolbar's profiler measure the response time of the activity related methods: `<lib/helpers.py:2567(new_activities)>`
- It should be around `50 ms`
- Now follow the 3 organizations
- Refresh the web page
- The call to `<lib/helpers.py:2567(new_activities)>` it's now in 1500 ms (it will scale to ~20 seconds with a bigger database)

## Expected behavior
The response time of the new activities feature will not impact on the UX of the site. (A delay time of ~100 ms?)

## Additional details

The issue seems to be related to two main factors:
1. How the overall logic to get the new activities count is being made
2. How the internal queries to the activity models are being made

**1. New Activities logic:**

The logic queries everything from the database and then uses some business logic in the action layer to filter results. Even when this approach may be fine in most cases, it is creating a bottleneck in the database for this feature.

Example: Even when activities from the users [are not counted](https://github.com/ckan/ckan/blob/abfda8b401ec370c9c74fc67678dcb38f8c72a26/ckan/logic/action/get.py#L3235), it [queries them](https://github.com/ckan/ckan/blob/abfda8b401ec370c9c74fc67678dcb38f8c72a26/ckan/model/activity.py#L413) from the database, and then adds some logic to the returned dictionary to [filter the count](https://github.com/ckan/ckan/blob/abfda8b401ec370c9c74fc67678dcb38f8c72a26/ckan/logic/action/get.py#L3218).

**2. Internal queries**

The logic to get activities from the database is a little bit complicated and not optimized. For each type of entity it:
- Get every followee object (user following X object)
- Creates and executes a select query for each of the previous objects
- Creates a union of all the previous queries

https://github.com/ckan/ckan/blob/abfda8b401ec370c9c74fc67678dcb38f8c72a26/ckan/model/activity.py#L346-L358

I played a little bit with queries. Instead of doing a union of select results, if I do a select filtering by id and the time of the query is reduced to a half (although It's not enough to reduce the time for a good UX).

```python
def _activities_from_datasets_followed_by_user_query(user_id, limit):
    '''Return a query for all activities from datasets that user_id follows.'''
    import ckan.model as model

    # Get a list of the datasets that the user is following.
    follower_objects = model.UserFollowingDataset.followee_list(user_id)
    if not follower_objects:
        # Return a query with no results.
        return model.Session.query(model.Activity).filter(text('0=1'))

    packages_id = set([f.object_id for f in follower_objects])

    return model.Session.query(model.Activity).\
        filter(model.Activity.object_id.in_(packages_id)).\
            limit(limit)
```

amercader:
Fixes #6028, #5765

As it happens often with these kind of issues, many hours of debugging and false leads end up condensed in a single line change :man_shrugging: . 
The fix is in 73c1ffc. The issue was located in `_group_activity_query()`, which apart from activities in the group itself looks for activities in datasets belonging to the group / org. When performing the join between the activity + member and package tables we were joining both the `package.id`  with the `member.table_id` (which makes sense and it's fast) but also `package.owner_org` with the provided `group_id`, which is *really* slow and also unnecessary, because we are adding an explicit member.group_id filter later on.

I fixed a test which was supposed to be testing that you got activities from datasets belonging to a group you were following, and added another specific one for datasets in organizations you are following.

Testing the `dashboard_activity_list` endpoint with this patch drastically reduced the load times from several seconds to milliseconds.

To reproduce:

```
# Create a test org, a test user and assign it as editor in the org
ckanapi action organization_create name=my-org
ckanapi action user_create name=my-user email=hi@example.com password=testtest
ckanapi action member_create id=my-org object=my-user object_type='user' capacity=editor

# Follow the org
ckanapi action follow_group -u my-user id=my-org

# Create loads of datasets belonging to the org (from @wardi in #2008)

(for a in `seq 3000` 
     do echo '{"name": "test-dataset-'$a'", "owner_org": "my-org"}'
done) | ckanapi load datasets -u my-user
```

You can then use a tool like [siege]() to measure loading time, eg

```
siege http://localhost:5010/api/action/dashboard_activity_list -H "Authorization:`cat api_token_my-user.txt`" -d1 -c10 -r1
```

| version | lowest dashboard_activity_list time | vs. master |
| --- | --- | --- |
| ckan master | 4.36s | 1.00 |
| PR #6028 | 0.05s | 0.0115 |

### DIFF
--- a/changes/6028.bugfix
+++ b/changes/6028.bugfix
@@ -1,0 +1,1 @@
+Fix performance bottleneck in activity queries
--- a/ckan/model/activity.py
+++ b/ckan/model/activity.py
@@ -225,8 +225,7 @@ def _group_activity_query(group_id, include_hidden_activity=False):
     ).outerjoin(
         model.Package,
         and_(
-            or_(model.Package.id == model.Member.table_id,
-                model.Package.owner_org == group_id),
+            model.Package.id == model.Member.table_id,
             model.Package.private == False,
         )
     ).filter(
--- a/ckan/tests/logic/action/test_get.py
+++ b/ckan/tests/logic/action/test_get.py
@@ -4338,14 +4338,44 @@ class TestDashboardNewActivities(object):
         user = factories.User()
         another_user = factories.Sysadmin()
         group = factories.Group(user=user)
+        helpers.call_action(
+            "follow_group", context={"user": user["name"]}, **group
+        )
         _clear_activities()
         dataset = factories.Dataset(
             groups=[{"name": group["name"]}], user=another_user
         )
         dataset["title"] = "Dataset with changed title"
         helpers.call_action(
-            "follow_dataset", context={"user": user["name"]}, **dataset
+            "package_update", context={"user": another_user["name"]}, **dataset
         )
+
+        activities = helpers.call_action(
+            "dashboard_activity_list", context={"user": user["id"]}
+        )
+        assert [
+            (activity["activity_type"], activity["is_new"])
+            for activity in activities[::-1]
+        ] == [("new package", True), ("changed package", True)]
+        assert (
+            helpers.call_action(
+                "dashboard_new_activities_count", context={"user": user["id"]}
+            )
+            == 2
+        )
+
+    def test_activities_on_a_dataset_in_a_followed_org(self):
+        user = factories.User()
+        another_user = factories.Sysadmin()
+        org = factories.Organization(user=user)
+        helpers.call_action(
+            "follow_group", context={"user": user["name"]}, **org
+        )
+        _clear_activities()
+        dataset = factories.Dataset(
+            owner_org=org['id'], user=another_user
+        )
+        dataset["title"] = "Dataset with changed title"
         helpers.call_action(
             "package_update", context={"user": another_user["name"]}, **dataset
         )


amercader:
While working on this I started refactoring the activity queries to make them easier to understand and refactor. Below is as far as I got and I won't have time to properly follow up on this but if this looks like an improvement I can send a separate PR:

```diff
diff --git a/ckan/model/activity.py b/ckan/model/activity.py
index 12ac4b9c6..d2410ce2e 100644
--- a/ckan/model/activity.py
+++ b/ckan/model/activity.py
@@ -201,7 +201,7 @@ def package_activity_list(
     return _activities_at_offset(q, limit, offset)
 
 
-def _group_activity_query(group_id, include_hidden_activity=False):
+def _group_activity_query(group_ids, include_hidden_activity=False):
     '''Return an SQLAlchemy query for all activities about group_id.
 
     Returns a query for all activities whose object is either the group itself
@@ -210,11 +210,6 @@ def _group_activity_query(group_id, include_hidden_activity=False):
     '''
     import ckan.model as model
 
-    group = model.Group.get(group_id)
-    if not group:
-        # Return a query with no results.
-        return model.Session.query(model.Activity).filter(text('0=1'))
-
     q = model.Session.query(
         model.Activity
     ).outerjoin(
@@ -236,17 +231,17 @@ def _group_activity_query(group_id, include_hidden_activity=False):
         # desired but is consistent with legacy behaviour.
         or_(
             # active dataset in the group
-            and_(model.Member.group_id == group_id,
+            and_(model.Member.group_id.in_(group_ids),
                  model.Member.state == 'active',
                  model.Package.state == 'active'),
             # deleted dataset in the group
-            and_(model.Member.group_id == group_id,
+            and_(model.Member.group_id.in_(group_ids),
                  model.Member.state == 'deleted',
                  model.Package.state == 'deleted'),
                  # (we want to avoid showing changes to an active dataset that
                  # was once in this group)
             # activity the the group itself
-            model.Activity.object_id == group_id,
+            model.Activity.object_id.in_(group_ids),
         )
     )
 
@@ -307,7 +302,7 @@ def group_activity_list(group_id, limit, offset, include_hidden_activity=False):
     etc.
 
     '''
-    q = _group_activity_query(group_id, include_hidden_activity)
+    q = _group_activity_query([group_id], include_hidden_activity)
     return _activities_at_offset(q, limit, offset)
 
 
@@ -337,9 +332,19 @@ def _activities_from_users_followed_by_user_query(user_id, limit):
         # Return a query with no results.
         return model.Session.query(model.Activity).filter(text('0=1'))
 
-    return _activities_union_all(*[
-        _user_activity_query(follower.object_id, limit)
-        for follower in follower_objects])
+    user_ids = set([f.object_id for f in follower_objects])
+
+    q = model.Session.query(model.Activity).\
+        filter(
+            or_(
+                model.Activity.object_id.in_(user_ids),
+                model.Activity.user_id.in_(user_ids)
+            )
+        )
+
+    q = _activities_limit(q, limit)
+
+    return q
 
 
 def _activities_from_datasets_followed_by_user_query(user_id, limit):
@@ -352,9 +357,14 @@ def _activities_from_datasets_followed_by_user_query(user_id, limit):
         # Return a query with no results.
         return model.Session.query(model.Activity).filter(text('0=1'))
 
-    return _activities_union_all(*[
-        _activities_limit(_package_activity_query(follower.object_id), limit)
-        for follower in follower_objects])
+    dataset_ids = set([f.object_id for f in follower_objects])
+
+    q = model.Session.query(model.Activity).\
+        filter(model.Activity.object_id.in_(dataset_ids))
+
+    q = _activities_limit(q, limit)
+
+    return q
 
 
 def _activities_from_groups_followed_by_user_query(user_id, limit):
@@ -373,9 +383,13 @@ def _activities_from_groups_followed_by_user_query(user_id, limit):
         # Return a query with no results.
         return model.Session.query(model.Activity).filter(text('0=1'))
 
-    return _activities_union_all(*[
-        _activities_limit(_group_activity_query(follower.object_id), limit)
-        for follower in follower_objects])
+    group_ids = set([f.object_id for f in follower_objects])
+
+    q = _group_activity_query(group_ids)
+
+    q = _activities_limit(q, limit)
+
+    return q
 
 
 def _activities_from_everything_followed_by_user_query(user_id, limit):

```

PR #6089, opened 2021-05-12
reorder resource view button: allow translation

wardi:
Fixes untranslatable "Reorder resource view" string

### Proposed fixes:

Removes `labelText` from the module options list so that it can be captured by `extract_messages` and added to the .pot file so that it is exported as a translation available to JS.

Note: this is my first foray into updating a JS module for a translated string, so I've probably done everything wrong. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/i18n/ckan.pot
+++ b/ckan/i18n/ckan.pot
@@ -2451,6 +2451,10 @@ msgstr ""
 msgid "Select a field"
 msgstr ""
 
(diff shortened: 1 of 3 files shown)


u10313335:
Could we have this fix in the future 2.9 releases?

smotornyuk:
Yes, this can be safely backported. I added corresponding labels

PR #6099, opened 2021-05-18
cast query string keys/values to string before passing to `quote()`

chris48s:
### Proposed fixes:

Cast query string keys/values to string before passing to `quote()`

This issue can cause us to throw `AttributeError: 'bool' object has no attribute 'rstrip'` calling something like

```py
url_for("/foo/bar", __no_cache__=True)
```

I've been hitting this problem on CKAN 2.9 so this should definitely be backported to 2.9. It might also be applicable to 2.8.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -8,6 +8,7 @@ available to Controllers. This module is available to templates as 'h'.
 import email.utils
 import datetime
 import logging
+import numbers
 import re
 import os
 import pytz
@@ -383,6 +384,14 @@ def url_for(*args, **kw):
     return _local_url(my_url, locale=locale, **kw)
 
 
+def _coerce_for_querystring(value):
+    if value is None:
+        return 'null'
+    if isinstance(value, numbers.Number):
+        return str(value)
+    return value
+
+
 def _url_for_flask(*args, **kw):
     '''Build a URL using the Flask router
 
@@ -442,10 +451,18 @@ def _url_for_flask(*args, **kw):
                     if isinstance(val, (list, tuple)):
                         for value in val:
                             query_args.append(
-                                u'{}={}'.format(quote(key), quote(value)))
+                                u'{}={}'.format(
+                                    quote(_coerce_for_querystring(key)),
+                                    quote(_coerce_for_querystring(value))
+                                )
+                            )
                     else:
                         query_args.append(
-                            u'{}={}'.format(quote(key), quote(val)))
+                            u'{}={}'.format(
+                                quote(_coerce_for_querystring(key)),
+                                quote(_coerce_for_querystring(val))
+                            )
+                        )
                 my_url += '&'.join(query_args)
         else:
             raise
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -8,7 +8,6 @@ available to Controllers. This module is available to templates as 'h'.
 import email.utils
 import datetime
 import logging
-import numbers
 import re
 import os
 import pytz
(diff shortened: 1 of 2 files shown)


amercader:
@chris48s thanks for this. It was discussed in the dev meeting that rather than cast everything to `str` and show Python's representation, the best behaviour if the passed value is not a string would be to handle specifically the `None` (output as `null`) and boolean (output as `true`, `false`) cases and fail in any other type (lists, dicts, etc)

chris48s:
Predumably the numeric case should also cast to string.
i.e: `{'id': 1}` should give us `?id=1` not throw an exception.

amercader:
Yes, sorry. Numbers should be supported

ckan/lib/helpers.py
```diff
@@ -383,6 +384,14 @@ def url_for(*args, **kw):
     return _local_url(my_url, locale=locale, **kw)
 
 
+def _coerce_for_querystring(value):
+    if value is None:
+        return 'null'
+    if isinstance(value, numbers.Number):
```
chris48s, line 390: note we don't need to explicitly check for boolean here because bool is a subclass of int

```
>>> import numbers
>>> isinstance(True, numbers.Number)
True
>>> isinstance(False, numbers.Number)
True
```

ckan/lib/helpers.py
```diff
@@ -383,6 +384,14 @@ def url_for(*args, **kw):
...
+        return str(value)
```
wardi, line 391: To keep with the JSON style `null`:
```suggestion
        return str(value).lower()  # render True/False as true/false
```

chris48s: Having churned this over slightly more over the weekend, it seems to me that really in this fallback case `except FlaskRouteBuildError:` we should really be trying to match the behaviour of named routes as closely as possible.

i.e: what we should be aiming for is that

`h.url_for("dataset.read", id="my_dataset", somebool=True)` and
`h.url_for("/dataset/my_dataset", somebool=True)`

should do the same thing, as should

`h.url_for("dataset.read", id="my_dataset", empty=None)` and
`h.url_for("/dataset/my_dataset", empty=None)`

etc


FWIW:
```
h.url_for("dataset.read", id="my_dataset", param=True) == '/dataset/my_dataset?param=True'
h.url_for("dataset.read", id="my_dataset", param=None) == '/dataset/my_dataset'
h.url_for("dataset.read", id="my_dataset", param={}) == '/dataset/my_dataset?param=%7B%7D'
```

Shall I alter it so the fallback matches the named route behaviour? @amercader @wardi ? It seems the most logical on reflection.

amercader: Yes, @chris48s this makes sense. It's good to have `url_for` having a consistent output regardless of how it is invoked

chris48s: OK. I've updated to implement that behaviour

amercader:
Thanks @chris48s 

Just so it's clear as I was confused by this. In this PR we are sticking with `some_param=True` instead of `some_param=true` because the former is the default behaviour by Flask, eg:

```python
from flask import url_for
url_for('home.about', some_param='cat', another=True, dogs=[], rabbits=None)

# '/about?some_param=cat&another=True'

```

And in this PR we are just fixing our fallback code for when Flask can't build a URL, eg because we are using the deprecated `h.url_for('/dataset/new')`

Does that sound good @wardi ?

wardi:
@amercader sounds good

chris48s:
Just to flag, I think this PR should be backported to the next 2.9 release (it is a bugfix not a feature) but it doesn't have the "Backport 2.9.4" label applied to it.

PR #6109, opened 2021-05-21
render activity timestamps with title= attribute

chris48s:
### Proposed fixes:

In CKAN 2.8, activites rendered timestamps with a `title=` attribute showing the full formatted timestamp. This meant that you would see the "fuzzy" timestamp (e.g: `2 days ago`) but in most browsers you can also hover over it to view the formatted datetime (e.g: `May 19, 2021, 11:17 (UTC)`). This was useful, but it was removed in CKAN 2.9 (presumably though oversight rather than intent?). In this PR I propose adding it back in.

### Features:

- [ ] includes API changes
- [x] includes bugfix for possible backport (2.9)

### DIFF
--- a/ckan/templates/snippets/activities/added_tag.html
+++ b/ckan/templates/snippets/activities/added_tag.html
@@ -7,7 +7,7 @@
       tag=ah.tag(activity)
     )|safe }}
     <br />
-    <span class="date">
+    <span class="date" title="{{ h.render_datetime(activity.timestamp, with_hours=True) }}">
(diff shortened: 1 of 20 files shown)


amercader:
TODO: check if there's a way to access this info without hovering.

amercader:
I did a quick audit and we are using `title` to convey extra information all over the place in the templates. I think we can merge this as it adds an existing feature that was removed during the 2.9 refactor, but it's something we need to address at some point.
I've created a separate issue for this where hopefully we can leverage the community expertise:

PR #6114, opened 2021-05-25
[#6113] Fix user email validator when using name as id parameter

### LINKED ISSUE #6113: Can not update user via `user_update` when using `name` for `id` parameter
**CKAN version**
2.9

**Describe the bug**
CKAN 2.9 introduced the `email_is_unique` validator to prevent duplicates. But the current implementation only works when the `id` param contains the actual UUID not the name, eg:

     http POST localhost:5000/api/action/user_update Authorization:{API_TOKEN} id=2e2dafdf-a7a8-495e-a9da-84ed4a1051a0 email=adria.mercader@okfn.org about=hola

Works fine, but this:

     http POST localhost:5000/api/action/user_update Authorization:{API_TOKEN} id=adria-mercader email=adria.mercader@okfn.org about=hola

Fails with a `ValidationError: "The email address 'adria.mercader@okfn.org' belongs to a registered user.`

The validator just needs to account for this scenario here:

https://github.com/ckan/ckan/blob/bcda3f6bfec5c68ac339070886702437df1c6ff2/ckan/logic/validators.py#L952-L953

amercader:
CKAN 2.9 introduced the `email_is_unique` validator to prevent duplicates. But the current implementation only works when the id param contains the actual UUID not the name.

Fixes #6113

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -936,20 +936,21 @@ def dict_only(value):
         raise Invalid(_('Must be a dict'))
     return value
 
+
 def email_is_unique(key, data, errors, context):
     '''Validate email is unique'''
     model = context['model']
     session = context['session']
 
     users = session.query(model.User) \
-            .filter(model.User.email == data[key]).all()
+        .filter(model.User.email == data[key]).all()
     # is there is no users with this email it's free
     if not users:
         return
     else:
         # allow user to update their own email
         for user in users:
-            if (user.name == data[("name",)]
+            if (user.name in [data[("name",)], data[("id",)]]
                     or user.id == data[("id",)]):
                 return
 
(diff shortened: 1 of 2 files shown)


ckan/logic/validators.py
```diff
@@ -936,20 +936,21 @@ def dict_only(value):
...
+
 def email_is_unique(key, data, errors, context):
     '''Validate email is unique'''
     model = context['model']
     session = context['session']
 
     users = session.query(model.User) \
-            .filter(model.User.email == data[key]).all()
+        .filter(model.User.email == data[key]).all()
     # is there is no users with this email it's free
```
kowh-ai, line 947: Maybe comment could be: ` # if there are no users with this email address then it's available`

kowh-ai (approved):
Logic looks good to me @amercader...although I'm definitely a lightweight when it comes to understanding the complexities...

amercader:
Thanks @kowh-ai !

PR #6123, opened 2021-05-28
Allow filter activities list

### LINKED ISSUE #6106: Allow `*_activity_list` actions to filter by `activity_type`
**CKAN version**: 2.9

It would be helpful if the `*_activity_list` actions had the ability to optionally accept a list of `activity_type`s to filter by.

e.g:

```py
get_action('package_activity_list')(context, {
    'id': 'my-dataset',
    'activity_types': ['changed package'],
})
```

- The default would be that all activities associated with `id` are returned (as it is now).
- Passing `activity_types` would filter to only the specified activity types.

This would be particularly useful for plugin authors leveraging custom activity types.

avdata99:
Fixes #6106

### Proposed fixes:
Adding `include_activity_types` and `exclude_activity_types` filters for activities list

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2534,8 +2534,8 @@ def package_activity_list(context, data_dict):
         NB Only sysadmins may set include_hidden_activity to true.
         (default: false)
     :type include_hidden_activity: bool
-    :param include_activity_types: A list of activity types to include in the response
-    :type include_activity_types: list
+    :param activity_types: A list of activity types to include in the response
+    :type activity_types: list
 
     :param exclude_activity_types: A list of activity types to exclude from the response
     :type exclude_activity_types: list
@@ -2547,9 +2547,12 @@ def package_activity_list(context, data_dict):
     # authorized to read.
     data_dict['include_data'] = False
     include_hidden_activity = data_dict.get('include_hidden_activity', False)
-    include_activity_types = data_dict.pop('include_activity_types', None)
+    activity_types = data_dict.pop('activity_types', None)
     exclude_activity_types = data_dict.pop('exclude_activity_types', None)
-    
+
+    if activity_types is not None and exclude_activity_types is not None:
+        raise Exception('You can\'t use activity_types and exclude_activity_types at the same time' )
+
     _check_access('package_activity_list', context, data_dict)
 
     model = context['model']
@@ -2565,7 +2568,7 @@ def package_activity_list(context, data_dict):
     activity_objects = model.activity.package_activity_list(
         package.id, limit=limit, offset=offset,
         include_hidden_activity=include_hidden_activity,
-        include_activity_types=include_activity_types,
+        activity_types=activity_types,
         exclude_activity_types=exclude_activity_types
     )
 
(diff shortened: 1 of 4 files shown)


wardi:
Can we name `include_activity_types` differently? In many other actions `include_foo` means "I want field foo in each of my results" so it could be confusing. How about just `activity_types`?

avdata99:
> Can we name `include_activity_types` differently? In many other actions `include_foo` means "I want field foo in each of my results" so it could be confusing. How about just `activity_types`?

@wardi, this come from [this comment](https://github.com/ckan/ckan/issues/6106#issuecomment-845131026) from @amercader.
Maybe we can use something like `filter_activity_types` as a dict with all the required information. Something like
```
filter_activity_types = {
    'types': ['a', 'b', 'c'],
    'action': 'exclude'
}
```
but this is more complex than I expected.

We can also just rename `include_activity_types` to `show_activity_types` and `exclude_activity_types` to `hide_activity_types`

wardi:
Passing a dict parameter (like we do with `datastore_search` and `datastore_delete`) is more difficult to validate with our normal validators and it's harder for users to pass with tools like curl or requests when using url parameters.

My vote is to keep the interface the way you have it just change the name. IMO it's clear what "activity_types" or even just "types" means from the action name. "exclude_activity_types" or "hide_activity_types" are fine because callers will use them less frequently.

Similar search APIs we have are `package_search` and `resource_search` which have "q" and "query" fields respectively with complicated logic behind them. I don't know if we want to go down that path.

avdata99:
I just renamed the parameter and added an exception if users try to use both parameters together
https://github.com/ckan/ckan/pull/6123/commits/bb3abeb4b2a5dcd6545a373cf9b4ecb68ae89d30
@wardi

smotornyuk:
In this case [ValidationError](https://github.com/ckan/ckan/blob/master/ckan/logic/__init__.py#L68) is more suitable than base Exception

avdata99:
> In this case [ValidationError](https://github.com/ckan/ckan/blob/master/ckan/logic/__init__.py#L68) is more suitable than the base Exception. [Here's](https://github.com/ckan/ckan/blob/master/ckan/logic/__init__.py#L68) a good example of its usage.
> 
> PS I may be subjective in this question, but I think it's better not to wrap actual error message into translation function(`_`). i.e, `` raise ValidationError({'include_filters': ['Cannot be used together with `exclude_filters`]}) ``

I [just updated](https://github.com/ckan/ckan/pull/6123/commits/2d83379c1fb205f85de05c152a7d8b79c12aa19d) the exception type @smotornyuk

ckan/logic/action/get.py
```diff
@@ -2542,6 +2547,12 @@ def package_activity_list(context, data_dict):
     # authorized to read.
     data_dict['include_data'] = False
     include_hidden_activity = data_dict.get('include_hidden_activity', False)
+    activity_types = data_dict.pop('activity_types', None)
+    exclude_activity_types = data_dict.pop('exclude_activity_types', None)
+
+    if activity_types is not None and exclude_activity_types is not None:
+        raise ValidationError({'activity_types': ['Cannot be used together with `exclude_filters']})
```
wardi, line 2554: exclude_filters?

avdata99: This was thew previous version of the field name
This is proposed to fix @wardi: https://github.com/ckan/ckan/pull/6144

PR #6132, opened 2021-06-02
[#5975] Add hydbrid search to Datastore search

### LINKED ISSUE #5975: Datastore_search doesn't allow "hybrid" searches
**CKAN version**
master

**Describe the bug**
The `datastore_search` q parameter can either do a table-wide FTS query OR a per-field query, but not both.  Per the [API doc](https://docs.ckan.org/en/2.9/maintaining/datastore.html#ckanext.datastore.logic.action.datastore_search):

> q (string or dictionary) – full text query. If it’s a string, it’ll search on all fields on each row. If it’s a dictionary as {“key1”: “a”, “key2”: “b”}, it’ll search on each specific field (optional)

**Expected behavior**
A table-wide FTS query combined with per-field query should be possible. Right now, the validator prevents the internal "_full_text" field from being added as a per-field query which should make this possible without changing the API interface.

steveoni:
Fixes #5975 

### Features:

- [X] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -383,7 +383,8 @@ def _where_clauses(data_dict, fields_types):
 
     # add full-text search where clause
     q = data_dict.get('q')
-    if q:
+    full_text = data_dict.get('full_text')
+    if q and not full_text:
         if isinstance(q, string_types):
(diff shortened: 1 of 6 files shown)


steveoni:
@amercader

PR #6133, opened 2021-06-03
make output of duplicate emails command clearer

chris48s:
### Proposed fixes:

Currently if you run `ckan db duplicate_emails`  and there are >0 active users and no email duplicates, the command prints `Searching for accounts with duplicate emails.` and then exits, leaving it slightly ambiguous what happened.
This PR adds a message making it clearer that no duplicates were found.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/db.py
+++ b/ckan/cli/db.py
@@ -117,18 +117,20 @@ def duplicate_emails():
         .filter(model.User.email != u"") \
         .order_by(model.User.email).all()
 
-    if not q:
-        log.info(u"No duplicate emails found")
+    duplicates_found = False
     try:
         for k, grp in groupby(q, lambda x: x[0]):
             users = [user[1] for user in grp]
             if len(users) > 1:
+                duplicates_found = True
                 s = u"{} appears {} time(s). Users: {}"
                 click.secho(
                     s.format(k, len(users), u", ".join(users)),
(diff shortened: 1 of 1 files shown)


smotornyuk:
`click.secho(message, fg="green")` - usually CKAN CLI reports success in this way. BTW, don't you think that this command suits for `ckan user` more than for `ckan db`?

chris48s:
> `click.secho(message, fg="green")` - usually CKAN CLI reports success in this way.

Updated :+1:

> don't you think that this command suits for `ckan user` more than for `ckan db`?

I don't have a strong view on it. Looking back at the original PR where this was added two maintainers asked for it to be added under `db` so that's what the author did https://github.com/ckan/ckan/pull/5100 . Maybe moving it is another issue.

ckan/cli/db.py
```diff
@@ -129,6 +129,7 @@ def duplicate_emails():
                     fg=u"green", bold=True)
     except Exception as e:
         tk.error_shout(e)
+    click.secho(u"No duplicate emails found", fg=u"green")
```
smotornyuk, line 132: Hm, this line is going to be executed every time, even if there **are** duplicates, right?

chris48s: :facepalm:

PR #6139, opened 2021-06-07
List(and apply) pending DB migrations / apply migrations from tests

smotornyuk:
* Test fixture `migrate_db_for`, which applies DB migrations defined in plugins ( `ckan generate migration -p PLUGIN_NAME`)
* Registers `ckan db pending-migrations [--apply]`, which shows all the enabled plugins that have unapplied migrations and, when `--apply` flag is provided, runs those migrations


amercader:
Looks good @smotornyuk, I think it's worth adding this new command and fixture to 2.9, what do you think?

PR #6149, opened 2021-06-10
Fix object list validation output

amercader:
This PR includes 3 changes that all combined fix the output of the navl validation against lists of objects (eg resources, extras, etc in a dataset dict):

* A test provided by @chris48s to reproduce the issue
* The reversion of commits a63207806c9474ed736c1a4fc774155042a45a51 and 7b6702b6c17efbe0b2b7eb46ab46da86e387c55c. These two commits were introduced by yours truly in #5107 during the massive Python 3 rewrite. While working on fixing problems caused by changing dicts during iterations on these validation functions I decided to simplify the output of the validation without taking into account implications down the line. It's clear now that this should be reverted.
* At the same time, the original code I introduce to addressed the changing dicts during iteration had a bug (`dict_to_process_copy = copy.deepcopy(dict_to_process)` instead of the correct `dict_to_process_copy = copy.copy(dict_to_process)`) which added to the confusion.

I believe the code works as expected now, and we are back to getting a list of equal size to the original objects, with empty dicts when they pass validation and a dict with the error keys if the don't. Eg for a dataset with tthree resources, where the second resource has a validation error in the field `position`:

```
    {
      'resources': [
        {}, {'position': [u'Missing value']}, {}
      ]
    }
```


amercader:
@wardi This one is also green now and would be good to have it ready for the next patch release

Issue #6012, opened 2021-04-13
Bootstrap 5 Upgrade


Zharktas:
The release of Bootstrap 5 is fast approaching and at the time of this writing beta3 is already released. There is old unfinished PR #5423 and it's probably smarter to upgrade straight to Bootstrap 5 instead of finishing the upgrade to 4 and then upgrade again in the near future.

This requires migrating from less to sass which has been used by bootstrap since bootstrap 4.
* Convert all less files to sass, some tool probably can be used to do this.
* Replace build tool from [Less](https://www.npmjs.com/package/less) to [Sass](https://www.npmjs.com/package/sass) 
* Update [Gulpfile](https://github.com/ckan/ckan/blob/master/gulpfile.js) to compile sass instead of less.
* Update documentation related to less.

We want to keep compatibility with Bootstrap 3 at least for one ckan version, so upgrading templates should have a flag in ckan configuration which version to use, similar to [Bootstrap3 upgrade](https://docs.ckan.org/en/2.8/maintaining/configuration.html#ckan-base-public-folder) had.
* Copy current templates to templates-bs3.
* Copy current public libraries to public-bs3.
* Add configuration flag for choosing which templates to use.
* Upgrade templates in templates folder.
* Update Gulpfile to copy bootstrap files to both public-bs3 and public.
* Update documentation to reflect upgrade of bootstrap and the new configuration flag.

amercader:
@Zharktas Could point 1 (less -> sass) be done independently with the current Boostrap 3 templates? Maybe using something like https://github.com/twbs/bootstrap-sass in the interim until we don't move to 5. Or is it not worth it and makes sense to do it all in one go?

Zharktas:
@amercader yes, sass conversion can be done with bootstrap3. There is a official port of bs3 for sass https://github.com/twbs/bootstrap-sass

Zharktas:
And I think it would make sense to make it as separate PR, so it would merged faster to master.

sagargg:
@Zharktas  I'm interested in this,  are we planning to migrate bootstrap 3 less ==> sass independently and then upgrade bootstrap 5 in a separate PR? Why we shouldn't do direct upgrading?

anuveyatsu:
@sagargg I believe this is to make reviewing and merging PRs easier.

Zharktas:
Bootstrap upgrade in itself is already a major effort, migrating to sass can be separated and as such should be so that something is eventually finished.

[...]

PR #6175, opened 2021-06-20
migrate css preprocessors less to scss 

sagargg:
Fixes:  #6012

### Proposed fixes/changes:
- Migrated css preprocessor less to SCSS.

### Features:

- [ ] includes API changes


Zharktas:
Whitelist was actually added in 3.4.1 as a security feature, not deprecated https://getbootstrap.com/docs/3.4/javascript/#js-sanitizer

sagargg:
I think it is removed and added its own HTML sanitizer on the later version.  
https://github.com/twbs/bootstrap/blob/v3.3.7/js/popover.js#L24
https://github.com/twbs/bootstrap/blob/v3.3.7/js/tooltip.js#L33-L47

Zharktas:
The whitelist in dashboard.js is literally from the Bootstrap 3.4 example as they changed it due to possible XSS.

image

Zharktas:
Oh right, you used bootstrap 3.3.7 as base, we have been using 3.4.1 for couple of years now.

Zharktas:
Bootstrap sass files should be installed using npm https://github.com/ckan/ckan/blob/master/package.json even though the files are copied in directories in gulp script

sagargg:
@Zharktas  Yes, already done here. 
https://github.com/ckan/ckan/blob/63cc3df3b568c11753f4a1db29dbede1e84d4a09/package.json

Zharktas:
👍 just did not notice it as the changelog is huge :D

Zharktas:
@sagargg can you resolve the conflict, forms.less had few lines removed https://github.com/ckan/ckan/commit/f868af00b3091dbf1ffe2c30ffbf2542114f4807#diff-fb75921ea5046f141dcf9308302fc071acf60e674518bdad56c13dec1d8327b3, i'll try to review and merge this soonish.

sagargg:
@Zharktas  fixed conflict. Let me know if you any confusion while reviewing.

ckan/public/base/scss/_bootstrap.scss
```diff
@@ -0,0 +1,56 @@
+/*!
+ * Bootstrap v3.3.7 (http://getbootstrap.com)
```
Zharktas, line 2: This file is based on the wrong version, probably doesn't have any other differences than the version number but for consistency sake it should be 3.4.1

ckan/public/base/scss/main-rtl.scss
```diff
@@ -0,0 +1,6 @@
+// Bootstrap 3.3.7
```
Zharktas, line 1: Wrong version number here too.

package.json
```diff
@@ -5,12 +5,13 @@
...
+    "bootstrap-sass": "^3.4.1",
     "font-awesome": "^4.7.0",
     "gulp": "^4.0.2",
     "gulp-if": "^3.0.0",
     "gulp-less": "^4.0.1",
     "gulp-rename": "^2.0.0",
+    "gulp-sass": "^4.1.0",
     "gulp-sourcemaps": "^2.6.5",
     "jquery": "^3.6.0",
     "less": "^3.13.1",
```
Zharktas, line 17: Probably merge conflict miss, less should be sass.

package.json
```diff
@@ -5,12 +5,13 @@
...
     "@highlightjs/cdn-assets": "^10.7.3",
     "blueimp-file-upload": "^10.31.0",
-    "bootstrap": "^3.4.1",
+    "bootstrap-sass": "^3.4.1",
     "font-awesome": "^4.7.0",
     "gulp": "^4.0.2",
     "gulp-if": "^3.0.0",
     "gulp-less": "^4.0.1",
     "gulp-rename": "^2.0.0",
+    "gulp-sass": "^4.1.0",
```
Zharktas, line 14: gulp-sass 5.0.0 was released a 20 days ago, would be nice to update to that directly. Requires setting the compiler yourself https://www.npmjs.com/package/gulp-sass#user-content-migrating-to-version-5

Zharktas (changes requested):
Some minor things I noticed.

package.json
```diff
@@ -9,13 +9,12 @@
...
     "gulp-if": "^3.0.0",
-    "gulp-less": "^4.0.1",
     "gulp-rename": "^2.0.0",
-    "gulp-sass": "^4.1.0",
+    "gulp-sass": "^5.0.0",
     "gulp-sourcemaps": "^2.6.5",
     "jquery": "^3.6.0",
-    "less": "^3.13.1",
     "moment": "^2.29.1",
+    "node-sass": "^6.0.1",
```
Zharktas, line 17: It would be better to use sass instead of node-sass, libraries used by node-sass are deprecated already:

image

Zharktas (changes requested):

package.json
sagargg: @Zharktas There was some warning message thrown for `sass` so, I've added `node-sass. I think bootstrap 3.4 is compatible only with `node-sass.  We can change it later while upgrading the bootstrap 5.

Zharktas: They are only deprecation warnings, it still compiles division as before. The warnings should go away when upgraded to bootstrap 5. I would be okay with the warnings compared to using deprecated versions.

sagargg: Sounds good. Replaced with  `sass`.

sagarggh: Hi @Zharktas,   is there anything else to improve on this ?

Zharktas:
Otherwise this seems fine, only thing missing is the changelog entry. Can you add file named 6175.feature to changes directory which has the content "Migrated preprocessor from LESS to SCSS" or something similar.

sagarggh:
@Zharktas  Due to forgotten 2fa codes I'm locked out  of my GitHub account and  it is irrecoverable 😢 .  I can't  push another commit so, can you add that by yourself once it's merged?

Zharktas:
I could modify your pull request so I added it :)

Zharktas (approved):

PR #6182, opened 2021-06-22
Remove legacy tests

### LINKED ISSUE #5158: Bug in `ckan jobs cancel` command
While completing #5110, I'm creating tests for new click CLI commands and found an error in `ckan jobs` command.

https://github.com/EvgeniiaVak/ckan/commit/a3cbca96505aae0398534e02da7ed9ac60b03a72

### Expected behaviour

(Probably) `ckan jobs cancel` for a not existing job exits with non-zero code.

### Actual behaviour

The [new TestJobsCancel.test_cancel_not_existing test](https://github.com/EvgeniiaVak/ckan/blob/fix/remove-paster-cli/ckan/tests/test_cli.py#L249) shows that it exits with code 0. This test was created as an update from [the old paster test](https://github.com/EvgeniiaVak/ckan/blob/fix/remove-paster-cli/ckan/tests/lib/test_cli.py#L258). 
At the same time, the [new TestJobsCancel.test_cancel_missing_id](https://github.com/EvgeniiaVak/ckan/blob/fix/remove-paster-cli/ckan/tests/test_cli.py#L259) passes ok - the command does exit with a non-zero code if it's not given a job id at all.

### Steps to reproduce 

1. checkout my branch: https://github.com/EvgeniiaVak/ckan/tree/fix/remove-paster-cli
2. run  
```bash
pytest --ckan-ini=test.ini ckan/tests/test_cli.py::TestJobsCancel
```

### LINKED ISSUE #6152: Fragile API token test on master
The test `ckan.tests.cli.test_user.TestApiToken.test_revoke` fails randomly on master on Circle CI.
Maybe it's a race condition or caused by splitting the tests.

```
assert 2 == 0
  +2
  -0
self = <ckan.tests.cli.test_user.TestApiToken object at 0x7f0ee7f675d0>
cli = <ckan.tests.helpers.CKANCliRunner object at 0x7f0ee6ef6890>

    def test_revoke(self, cli):
        user = factories.User()
        call_action(u"api_token_create", user=user[u"id"], name=u"first token")
        tid = model.Session.query(model.ApiToken).first().id
        args = [
            u"user",
            u"token",
            u"revoke",
            tid,
        ]
        result = cli.invoke(ckan, args)
>       assert result.exit_code == 0
E       assert 2 == 0
E         +2
E         -0

ckan/tests/cli/test_user.py:96: AssertionError
```

smotornyuk:
Move legacy tests to the main test tree. I think it will reduce a bit test execution time and finally there will be no `legacy` inside CKAN's test suite. 

Fixes #6152 (not actually fixes, but adds more verbosity to the failed CLI tests) and fixes #5158


ckan/lib/config_tool.py
```diff
@@ -235,7 +236,7 @@ def insert_new_sections(new_sections):
                         (key, existing_option.value, section)
 
             if changes_made:
-                print(changes_made)
+                log.debug(changes_made)
```
amercader, line 239: Other CLI commands output to stdout with `click.secho`, shouldn't we do the same here? I found useful to know what was changed in the ini file when using this command. In any case if we want to use logging instead this should be probably be a `log.info()`

smotornyuk: This code is outside `ckan.cli` module, so I don't want to tie it to the `click`. But using `log.info` is good alternative

pyproject.toml
```diff
@@ -38,12 +38,13 @@ line-length = 79
...
+    "ignore::bs4.GuessedAtParserWarning"  # using lxml as default parser
 ]
 markers = [
     "ckan_config: patch configuration used by other fixtures via (key, value) pair.",
     "ckan_pytest: test case that is explicitely was rewriten from `nose` style"
 ]
 
 testpaths = ["ckan", "ckanext"]
-addopts = "--strict-markers  --pdbcls=IPython.terminal.debugger:TerminalPdb"
+addopts = "--strict-markers --pdbcls=IPython.terminal.debugger:TerminalPdb -p no:ckan"
```
amercader, line 50: Can you explain what this does?

amercader (approved):
Great work @smotornyuk!

This is truly the end of an era :sauropod:

pdelboca:
Great work @smotornyuk ! This cleaning is awesome :+1:

pyproject.toml
smotornyuk: `-p` flag controls pytest plugins that are enabled during the test execution. `-p xxx` means "enable pytest-xxx", `-p no:xxx` means "disable pytest-xxx". If there is no such plugin `-p no:xxx` just silently ignored.

As all the new extension has `pytest-ckan` added to the `dev-requirements.txt`, as long as you have a custom extension installed, you'll get warnings every time you run CKAN core tests. `pytest-ckan` just duplicates original CKAN fixtures and this causes those warnings.

The flag above disables `pytest-ckan` plugin(which is not used by CKAN itself, anyway) if it's present and thus removes noisy warnings

amercader: This makes sense thanks

PR #6192, opened 2021-06-26
Predictable variables in chained config

smotornyuk:
At the moment chained config files evaluate variables in the order they are parsed. This means that the following case 
```
# a.ini
output = %(var)s

# b.ini
use = config:a.ini
var = B

# c.ini
use = config:b.ini
var = C
```
will produce `output == B`, because `b.ini` is parsed after `c.ini` (as dependency) and thus `var` from `b.ini` overrides `c.ini` version. 

As a result, one can't use inside `c.ini` any variable that is defined in `a.ini` or `b.ini` (because these files are not parsed yet) 

I think that python-like behavior is more appropriate in that case. When `class C(B): ...` is defined, we expect `C` to override values from `B` and not vice versa. The only exception is the `here` variable - it must never be overridden and must always refer to the file where it used. 
Keeping this special case in mind, I've updated the logic. Now configuration files are parsed starting from the last file in the chain (`a.ini` in the example above), values that contain `%(here)s` are evaluated immediately and all other values are evaluated when parsing is finished. 
Note: that means that now `[DEFAULTS]` from the top-level config (`c.ini`) has higher precedence. Previously `[DEFAULTS]` from the very last file (`a.ini`) were used. Thus it's a breaking change, but I think it still can be backported to v2.9.4


ckan/cli/__init__.py
```diff
@@ -34,45 +34,63 @@ def _read_config_file(self, filename):
...
-                self._read_config_file(use_config_path)
-                self._update_config()
-            else:
-                break
+        chain = self._unwrap_config_chain(self.config_file)
+        for filename in reversed(chain):
+            self._read_config_file(filename)
+            # The global_config key is to keep compatibility with Pylons.  It
+            # can be safely removed when the Flask migration is completed.
+            self.config[u'global_conf'] = self.parser.defaults().copy()
```
pdelboca, line 89: I think we can safely remove this since we don't need to maintain compatibility with Pylons anymore.

amercader:
This looks good @smotornyuk, can you add a changelog entry for the change in behaviour (it's unlikely people will get hit by it but better to be thorough)

smotornyuk:
Done

PR #6193, opened 2021-06-26
Make `CKAN_*` envvars available as variables in config

smotornyuk:
Add all `CKAN_*` environment variables to the `[DEFAULT]` section of the config. We already doing something similar [for the subset of config options](https://github.com/ckan/ckan/blob/master/ckan/config/environment.py#L97-L115). But existing functionality is aggressive and unconditionally replaces config values with the ones that came from the environment. Probably it was done mainly for docker.

This PR provides this sort of functionality for any env variable with the name `CKAN_*` but in a less aggressive manner. Variable can be specified explicitly in the config file and then can be overridden with the static value if the current config file is extended by another one (#6192)

### DIFF
--- a/changes/6192.misc
+++ b/changes/6192.misc
@@ -1,0 +1,1 @@
+Environment variables prefixed with `CKAN_` can be used as variables inside config file via `option = %(CKAN_***)s`
--- a/ckan/cli/__init__.py
+++ b/ckan/cli/__init__.py
@@ -18,7 +18,10 @@ class CKANConfigLoader(object):
         self.config = dict()
         self.parser = ConfigParser()
         self.section = u'app:main'
-        defaults = {u'__file__': os.path.abspath(self.config_file)}
+        defaults = dict(
+            (k, v) for k, v in os.environ.items()
+            if k.startswith("CKAN_"))
+        defaults['__file__'] = os.path.abspath(self.config_file)
         self._update_defaults(defaults)
         self._create_config_object()
 
--- a/ckan/tests/cli/data/test-env-var.ini
+++ b/ckan/tests/cli/data/test-env-var.ini
@@ -1,0 +1,3 @@
+[app:main]
+use = egg:ckan
+var = %(CKAN_TEST_ENV_VAR)s
--- a/ckan/tests/cli/data/test-no-env-var.ini
+++ b/ckan/tests/cli/data/test-no-env-var.ini
@@ -1,0 +1,3 @@
+[app:main]
+use = egg:ckan
+var = %(TEST_ENV_VAR)s
--- a/ckan/tests/cli/test_cli.py
+++ b/ckan/tests/cli/test_cli.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+from configparser import InterpolationMissingOptionError
 import pytest
 import os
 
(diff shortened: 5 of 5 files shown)


wardi:
I like this. It makes pulling env vars into site configurations explicit and is less magical than ckanext-envvars

pdelboca:
While trying to implement the Python Monitor for CKAN core, I came across this PR.

The issue is that the command `ckan config undeclared` will output all `CKAN_*` environment variables as not declared. This created some noise in my head: "How should we treat this dynamic variables in our code?"
 - If using strict mode, should we force a declaration? 
 - Should we exclude them from the undeclared logic?
 - Which `CKAN_` variables should we exlude given that we have declared `CKAN_` ones.
 - If we exclude them from the undeclared statement, we are running on `strict` mode but allowing variables to not be declared.
 
The following test fails:
```python
self = <ckan.tests.cli.test_config.TestUndeclared object at 0x7f7a73fc2250>
command = <function command.<locals>.invoke at 0x7f7a6e0a4dd0>

    def test_no_undeclared_options_by_default(self, command):
        result = command("undeclared", "-idatapusher", "-idatastore")
>       assert not result.output
E       AssertionError: assert not 'CKAN_REDIS_URL\nCKAN_SQLALCHEMY_URL\nCKAN_SOLR_URL\nCKAN_DATASTORE_WRITE_URL\nCKAN_DATASTORE_READ_URL\n'
E        +  where 'CKAN_REDIS_URL\nCKAN_SQLALCHEMY_URL\nCKAN_SOLR_URL\nCKAN_DATASTORE_WRITE_URL\nCKAN_DATASTORE_READ_URL\n' = <Result okay>.output
```
If we set the CD/CI like this:
```yaml
jobs:
  test:
    runs-on: ubuntu-latest
    container:
      image: openknowledge/ckan-dev:master
    services:
      solr:
        image: ckan/ckan-solr:master
      postgres:
        image: ckan/ckan-postgres-dev:master
        env:
          POSTGRES_USER: postgres
          POSTGRES_PASSWORD: postgres
          POSTGRES_DB: postgres
        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
      redis:
          image: redis:3
    env:
      CKAN_SQLALCHEMY_URL: postgresql://ckan_default:pass@postgres/ckan_test
      CKAN_DATASTORE_WRITE_URL: postgresql://datastore_write:pass@postgres/datastore_test
      CKAN_DATASTORE_READ_URL: postgresql://datastore_read:pass@postgres/datastore_test
      CKAN_SOLR_URL: http://solr:8983/solr/ckan
      CKAN_REDIS_URL: redis://redis:6379/1
```

smotornyuk:
> If using strict mode, should we force a declaration?
No. The strict mode only controls declared variables. Even though enforcing declaration for every config option may guarantee better code quality, we cannot just disallow undeclared config options. People are free to write bad code. If something is declared, we can control it in strict mode. If something is not declared - people will see a warning as a hint, that code is not ideal.

> Should we exclude them from the undeclared logic?
I don't think so. Every config option that we are expecting in the config file, should be declared. The problem here is that we are using a finite set of `CKAN_*` options([here's the full list](https://github.com/ckan/ckan/blob/master/ckan/config/environment.py#L86-L102)), but I forgot to add them to declarations. I think, [here's the good place for them](https://github.com/ckan/ckan/blob/master/ckan/config/config_declaration.yaml#L29-L46)

> Which CKAN_ variables should we exlude given that we have declared CKAN_ ones.
> If we exclude them from the undeclared statement, we are running on strict mode but allowing variables to not be declared.
Answered in the previous reply.

So, basically, I missed these options while declaring everything in core. As soon as they will get their declaration, there will be no problem. If somebody is going to use a different `CKAN_*` environment variable, it's a good idea to declare it in the custom extension.

PR #6205, opened 2021-06-30
Remove old resource preview system

amercader:
Before the current resource views (CKAN 2.3) there was another implementation for a more rudimentary system (known as "previews", as opposed to the current "views"), which was still supported. This removes this long unused code, simplifying a bit the resource read template, removing unused templates and helpers.


wardi:
@amercader would you mind fixing up the conflict?

amercader:
Done @wardi

amercader:
@wardi This is now fixed and green again

PR #6206, opened 2021-06-30
Upgrade node dependencies to latest minor versions

amercader:
This addresses some security alerts

@Zharktas I could not for the live of me upgrade `glob-parent`, for which there is dependabot alert. My npm skills only go so far so maybe you have a better shot.

```
ckan@1.0.0 /home/adria/dev/pyenvs/ckan-py3/src/ckan
└─┬ gulp@4.0.2
  ├─┬ glob-watcher@5.0.3
  │ └─┬ chokidar@2.1.8
  │   └── glob-parent@3.1.0 
  └─┬ vinyl-fs@3.0.3
    └─┬ glob-stream@6.1.0
      └── glob-parent@3.1.0 
```


Zharktas:
@amercader the problem lies in gulp and its glob watcher, they can't update before they fix breaking changes. And as they maintain them mainly on volunteer basis like us, the work is slow. If we want something financially maintained, we will need to replace it with something else. The security alerts aren't that bad as the dependencies are only used in dev environments and only with script that compiles less and copies vendor javascript libraries from node_modules to those used in version control. So it shouldn't be too hard to replace it with something maintained or something that doesn't have as much dependencies.

amercader:
@Zharktas thanks that makes sense. Any ideas why the front-end test broke because of this change?

Zharktas:
I'm guessing that ancient node.js 0.10 has something to with it and it should be updated.

amercader:
Also updated the node version we use on CircleCI from 10 to 14

Zharktas (approved):

PR #6218, opened 2021-07-02
remove unused resource-upload-field js+css

wardi:
### Proposed fixes:

resource-upload-field.js has been untouched for years and doesn't seem to be used by anything in core

### Features:

- [ ] includes API changes


ckan/public/base/css/main.css
```diff
@@ -8095,34 +8095,6 @@ fieldset[disabled] .control-custom.disabled .checkbox.btn.focus {
...
-  margin: 0;
-  border: solid transparent;
-  border-width: 100px 0 0 200px;
-  cursor: pointer;
-  direction: ltr;
-  -moz-transform: translate(-300px, 0) scale(4);
-}
-.resource-upload-field.loading {
-  display: inline-block;
-  background: url("../../../base/images/loading-spinner.gif") no-repeat center right;
```
amercader, line 8123: I guess we can remove this image too!

wardi: I see it referenced in ckan/public/base/javascript/modules/popover-context.js but it doesn't seem like that's used either (ckan/templates/snippets/context/dataset.html has a class with the same name but that shouldn't use the module IIUC)

Should I remove this .js file too?

amercader: Yes, I think this can be removed as well

PR #6237, opened 2021-07-09
[2317] fix default ordering consuming resouce content from datastore

### LINKED ISSUE #2317: Unreliable ordering of datastore results
Hi,

In the absence of an "ORDER BY" clause, PostgreSQL does not make any promises regarding the ordering of rows. In particular if you use "OFFSET/LIMIT" clauses the ordering might actually **change** between two subsequent queries. This is from the PostgreSQL documentation:

> The query planner takes LIMIT into account when generating a query plan, so you are very likely to get different plans (yielding different row orders) depending on what you use for LIMIT and OFFSET. Thus, using different LIMIT/OFFSET values to select different subsets of a query result will give inconsistent results unless you enforce a predictable result ordering with ORDER BY. This is not a bug; it is an inherent consequence of the fact that SQL does not promise to deliver the results of a query in any particular order unless ORDER BY is used to constrain the order. 

http://www.postgresql.org/docs/current/static/sql-select.html

This issue affects datastore queries (I have witnessed it). While this is not technically a CKAN bug (the problem can be moved upwards to the API consumers), it is an unexpected behaviour. I would suggest the best place to address this is in the datastore itself, to always ensure queries include an ORDER BY term on a unique key (eg. _id)

I can do a pull request if people agree this is a good thing to do.

wrinklenose:
Fixes #2317 

### Proposed fixes:

Fixed default ordering when consuming resource content from datastore

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/2317.bugfix
+++ b/changes/2317.bugfix
@@ -1,0 +1,1 @@
+Fixed default ordering when consuming resource content from datastore
--- a/changes/2317.bugfix
+++ b/changes/2317.bugfix
@@ -1,1 +1,1 @@
-Fixed default ordering when consuming resource content from datastore
+Stable default ordering when consuming resource content from datastore
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1296,7 +1296,7 @@ def search_data(context, data_dict):
 
     query_dict = {
         'select': [],
-        'sort': ['_id'],
+        'sort': [],
         'where': []
     }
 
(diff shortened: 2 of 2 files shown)


ckanext/datastore/backend/postgres.py
```diff
@@ -1319,6 +1319,9 @@ def search_data(context, data_dict):
     else:
         distinct = ''
 
+    if not sort and not distinct:
```
wardi, line 1322: Is there a reason you don't want the results ordered when distinct is applied?

wardi: nevermind, I see the failing error now

changes/2317.bugfix
```diff
@@ -0,0 +1 @@
+Fixed default ordering when consuming resource content from datastore
```
wardi, line 1:
```suggestion
Stable default ordering when consuming resource content from datastore
```

PR #6238, opened 2021-07-09
Upgrade Python requirements

amercader:
Upgrade the base requirements, except the following ones which can be done in separate PRs:

* Flask + Wekzeug + Click (There's a 1.0 -> 2.0 jump, I didn't check the changelog in detail)
* Jinja2: Upgrade seems fine, but we need to upgrade Flask first
* SQLAlchemy: See https://github.com/ckan/ckan/issues/6243
* pysolr: latest version pysolr 3.9.0 won't work with the ancient Solr version (3.6.0) shipped by Ubuntu. We need to update our docs and the CI setup to ensure a more recent version is installed.
* ~~unicodecsv: @wardi this seems only used for the DataStore dump functionality and IIUC should no longer be needed in Python 3 right?~~
* Flask-Babel: FYI @TkTech upgrading from 1.0.0 to 2.0.0 broke the [`ITranslation` interface tests](https://app.circleci.com/pipelines/github/ckan/ckan/1902/workflows/8ef66c69-6def-4d1b-8bbf-1678e93f56aa/jobs/9630)


amercader:
Moving discussion for this failure from Gitter to here:

```
  File "/root/project/ckan/cli/cli.py", line 123, in _init_ckan_config
    _add_ctx_object(ctx, value)
  File "/root/project/ckan/cli/cli.py", line 132, in _add_ctx_object
    ctx.obj = CtxObject(path)
  File "/root/project/ckan/cli/cli.py", line 53, in __init__
    self.config = ckan_cli.load_config(conf)
  File "/root/project/ckan/cli/__init__.py", line 122, in load_config
    config_loader = CKANConfigLoader(filename)
  File "/root/project/ckan/cli/__init__.py", line 26, in __init__
    self._create_config_object()
  File "/root/project/ckan/cli/__init__.py", line 54, in _create_config_object
    self._update_config()
  File "/root/project/ckan/cli/__init__.py", line 41, in _update_config
    value = self.parser.get(self.section, option)
  File "/usr/local/lib/python3.7/configparser.py", line 791, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'ckan_datastore_postgres_read_user' in section: 'app:main'
```
From @smotornyuk :
* in pyutillib v5.7 [config parser was patched](https://github.com/PyUtilib/pyutilib/blob/5.7.3/pyutilib/component/config/plugin_ConfigParser.py#L31). After v5.8 this patch is [not called globally](https://github.com/PyUtilib/pyutilib/blob/master/pyutilib/component/config/plugin_ConfigParser.py#L31)
* We are [updating internal _defaults property](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/cli/__init__.py#L30) of the config parser directly
* Without the pyutilib patch, keys are stored in the default way, which is lower casing them.

I guess the fix here is to set the same option in `ConfigParser` ourselves (b5dfb746e)

---

On a kind of related note to https://github.com/ckan/ckan/pull/6193, I noticed that these changes, apart from making `CKAN_*` env vars available for interpolation in the ini file, they also add the env var key to the config object. So for instance if I set:

    export CKAN_MY_CUSTOM_SETTING=test

and add this to my ini file:

   ckan.some.setting = %(CKAN_MY_CUSTOM_SETTING)s

Then the resulting config object will be:

```
{
    "CKAN_MY_CUSTOM_SETTING": "test",
    "__file__": "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan.ini",
    "ckan.some.setting": "test",
    [...]

}
```

Was that intended?

smotornyuk:
> ... apart from making CKAN_* env vars available for interpolation in the ini file, they also add the env var key to the config object. So for instance if I set: ...

Yes. These `CKAN_*` variables are added to the `DEFAULT` section of the config object. If the same key exists under `app:main` section, it will be used. But when such key is missing, we are taking this "default"-env option. When you printing the config object, it lists values from both `app:main` and `default` sections as if they are defined on the same level. But it's internal representation is more like:
```
collection.ChainMap(APP_MAIN,  DEFAULTS_WITH_ENV)
```

wardi:
@amercader yes unicodecsv can be dropped

amercader:
I'll leave this PR as is and we can deal with the outstanding libraries in separate PRs. Once we get all requirements up to speed with the latest versions I suggest configuring and turning on [Dependabot updates](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically) to make this maintenance easier in the future.

PR #6239, opened 2021-07-09
Refactor to use Flask's builtin flash method

pdelboca:
This PR refactors the helper functions to use [Flask's built-in flash()](https://flask.palletsprojects.com/en/2.0.x/api/#message-flashing) method to display flash messages: [Message Flashing](https://flask.palletsprojects.com/en/2.0.x/patterns/flashing/)

[Flask's implementation](https://github.com/pallets/flask/blob/7a73171edc3bcffef96ef6367977ab3ae9af9350/src/flask/helpers.py#L367-L439) of flash messages is quite similar to CKAN's one (objects stored in session object) so the migration is quite straight forward.

I added some tests just to make sure we're not changing the current behavior

This PR maintains the current helpers API for flash messages: `flash_*(message, allow_html=False):` so it shouldn't impact on other extensions

For the `allow_html` feature, I'm using Flask's recommended approach to use Markups before flashing messages: https://flask.palletsprojects.com/en/2.0.x/templating/#controlling-autoescaping.

### __no_cache__
I deleted the following snippet of code in `redirect_to`:
```python
    ...
    if are_there_flash_messages():
        kw['__no_cache__'] = True
   ...
```

I would like to debate why are we forcing a `__no_cache__` rule if we have flash_messages in the session object. It doesn't make too much sense and is one of those hidden behaviors that makes debugging hard and causes some not desired side effects.

Looks like an old line of code when we were implementing some sort of built-in cache. Some commits for reference:
 - https://github.com/ckan/ckan/commit/44812132f0837bf196a03a6414452d57077935a5
 - https://github.com/ckan/ckan/commit/187c77840788ee0c68b83b202d64791b07e25597
 - https://github.com/ckan/ckan/commit/ebfb75eacf1e1d6f3cbc4f1fe4d2584611c970d4

It can be easily implemented with this new approach but I would like to review if it is truly necessary.


pdelboca:
@amercader this one has been sitting here for a while. Let me know if I can help reviewing it!

pdelboca:
@amercader I have updated this PR with the latest changes in master (BS5). The failing test seems random but let me know if it looks okay!

PR #6241, opened 2021-07-09
Clean lib/base.py

pdelboca:
This PR cleans the `base.py` module:
- Fixes docstring on `abort` method
- Clean old deprecated arguments from `render`
- Removes `render_jinja2` function since it depends on the no longer existing `pylons.app_globals` config
- Cleans unused imports
- Edit `custom_emails` example to use `render` instead of `render_jinja2`

### DIFF
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -5,29 +5,18 @@
 Provides the BaseController class for subclassing.
 """
 import logging
-import time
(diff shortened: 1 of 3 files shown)


amercader:
@pdelboca FYI these tests you update weren't actually run and when they were, they were failing. I fixed them in 471eccb2d

PR #6252, opened 2021-07-14
fix for render_datetime()

### LINKED ISSUE #5946: render_datetime helper does not respect ckan.display_timezone configuration
**CKAN version**
N/A

**Describe the bug**
[`ckan.display_timezone`](https://docs.ckan.org/en/latest/maintaining/configuration.html#ckan-display-timezone) sets the timezone to use when displaying dates on the frontend.  

The `render_date` template helper however,  still uses UTC even if `ckan.display_timezone` is set to a non-UTC timezone.

**Expected behavior**
`render_datetime` should respect the display_timezone setting.

TomeCirun:
Fixes #5946

### Proposed fixes:
passing rebase=False to format_datetime() will make render_datetime to respect ckan.display_timezone configuration

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -247,6 +247,7 @@ def make_flask_stack(conf):
     app.config[u'BABEL_TRANSLATION_DIRECTORIES'] = ';'.join(i18n_dirs)
     app.config[u'BABEL_DOMAIN'] = 'ckan'
     app.config[u'BABEL_MULTIPLE_DOMAINS'] = ';'.join(i18n_domains)
+    app.config[u'BABEL_DEFAULT_TIMEZONE'] = str(helpers.get_display_timezone())
 
     babel = CKANBabel(app)
 
--- a/ckan/lib/formatters.py
+++ b/ckan/lib/formatters.py
@@ -51,7 +51,7 @@ def localised_nice_date(datetime_, show_date=False, with_hours=False,
     if with_seconds:
         return format_datetime(datetime_, format or 'long')
     elif with_hours:
-        fmt_str = "MMMM d, YYYY, HH:mm (ZZZZ)"
+        fmt_str = "MMMM d, YYYY, HH:mm (z)"
         return format_datetime(datetime_, format or fmt_str, rebase=False)
     else:
         return format_date(datetime_, format or 'long')
--- a/ckan/lib/formatters.py
+++ b/ckan/lib/formatters.py
@@ -51,8 +51,8 @@ def localised_nice_date(datetime_, show_date=False, with_hours=False,
     if with_seconds:
         return format_datetime(datetime_, format or 'long')
     elif with_hours:
-        fmt_str = "MMMM d, YYYY, HH:mm (z)"
-        return format_datetime(datetime_, format or fmt_str)
+        fmt_str = "MMMM d, YYYY, HH:mm (ZZZZ)"
+        return format_datetime(datetime_, format or fmt_str, rebase=False)
     else:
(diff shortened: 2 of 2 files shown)


TomeCirun:
@wardi what's your opinion over this?

I have already put my research [here](https://github.com/ckan/ckan/issues/5946)

Thanks

wardi:
@smotornyuk you updated this code in https://github.com/ckan/ckan/pull/5376 WDYT?

At a minimum it doesn't make sense to pass this to only one of the `format_date`/`format_datetime` calls.

TomeCirun:
@wardi Thanks for you reply.

In the template we are passing with_hourse=True https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/templates/snippets/local_friendly_datetime.html#L13
as only this code is going to execute
https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/lib/formatters.py#L53  
https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/lib/formatters.py#L55
thats why i pass rebase=False only in that place

If you have any suggestion, I m glad to fix it if needed

Thanks.

smotornyuk:
It doesn't sound quite right for me.. you are trying to fix consequences, not the source of the problem. Can you just set babel's default locale to the one, specified by `ckan.display_timezone`? Using `BABEL_DEFAULT_TIMEZONE` option of flask-babel

And yes, I think that all routes of `localised_nice_date` should return localized representation, not only `with_hours=True`.

TomeCirun:
hey @smotornyuk  thanks for your reply

I just set the babel's default timezone to the one specified by `ckan.display_timezone`... working good when providing a timezone(e.g Europe/Zurich) 
image


.. but when i set the value:  `server`  

image


i got this error `pytz.exceptions.UnknownTimeZoneError: 'server'`

Any suggestion?
Thanks

PS The source of the problem is because we didnt provide `BABEL_DEFAULT_TIMEZONE` (by default is set to UTC when no timezone provided)...so when we are calling `format_datetime()` by default rebasing happens and babel's `get_timezone()` returns UTC 

So .. to prevent this...we can: 
-pass `rebase=False` to all `format_datetime() ` 
or
-provide `BABEL_DEFAULT_TIMEZONE` value to the one specified by `ckan.display_timezone` (just keep on mind that when `ckan.display_timezone` is set to `server` we get this error: `pytz.exceptions.UnknownTimeZoneError: 'server'` )

smotornyuk:
There is a helper, `get_display_timezone`, that solves our issue with `server` value. [in this special case](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L1578-L1579) it returns `tzlocal.get_localzone()` object. As any tzinfo object, this one has `zone` property that can be passed to the Babel's  config.

So you can "unwrap" value of `ckan.display_timezone` [here](https://github.com/ckan/ckan/blob/master/ckan/config/middleware/flask_app.py#L247-L249) , right before babel is initialized

TomeCirun:
@smotornyuk Thanks for your help.

TomeCirun:
hey @wardi can you check my last PR? I did it as @smotornyuk suggest.

Thanks

TomeCirun:
Hey @wardi it's me again 😊 can you please check my PR ?

Thanks

wardi:
@smotornyuk I'll have to rely on you again, does this seem like a sensible flask config change?

smotornyuk:
Yep. I would say that it's the only right way to configure flask-babel timezone in CKAN)

PR #6253, opened 2021-07-14
filter activities that user should not be authorized to view

### LINKED ISSUE #6107: Expose `include_data` flag on `*_activity_list` actions
**CKAN version**: 2.9

In CKAN 2.9 the `*_activity_list` actions hard-code

```py
data_dict['include_data'] = False
```

https://github.com/ckan/ckan/blob/ckan-2.9.2/ckan/logic/action/get.py#L2479
https://github.com/ckan/ckan/blob/ckan-2.9.2/ckan/logic/action/get.py#L2530
etc

which causes the data object to be stripped down to an empty object or just the `title` attribute (if one exists).

This is a behaviour change from CKAN 2.8 where the `*_activity_list` functions returned the full contents of the data object - a feature which some plugins may rely on.

This has been done to prevent information in the data object being exposed to users who aren't authorized to see it :+1: This is a very hard problem to fully solve because `data` is an arbitrary field which may contain different objects depending on the activity type and could contain a mix of content with different auth requirements. This gets even harder to fully address once there are userland plugins potentially inserting arbitrary data via custom activity types.

A useful first step would be to allow `include_data` to be set True on the `data_dict` when:

- The user invoking the action is a sysadmin **or**
- The action is invoked with `context['ignore_auth']: True`

This would allow plugin authors to leverage more of the `data` object content while still preventing unwanted information leakage to unprivileged users via the JSON API. `context['ignore_auth']: True` would provide some migration path for authors who have developed plugins relying on content from the `data` object based on the CKAN 2.8 behaviour, especially when combined with https://github.com/ckan/ckan/issues/6106 and delegate responsibility for ensuring correct authorization to the plugin author.

TomeCirun:
Fixes #6107

### Proposed fixes:
I filtered the activities that user should not be authorized to read: 
-user can see activity list  for himself only
-only user who is member in particular organization/group should be able to see the activity stream for that organization.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -17,6 +17,7 @@ import ckan.lib.dictization
 import ckan.logic as logic
 import ckan.logic.action
 import ckan.logic.schema
+import ckan.lib.helpers as h
 import ckan.lib.dictization.model_dictize as model_dictize
 import ckan.lib.jobs as jobs
 import ckan.lib.navl.dictization_functions
@@ -2469,6 +2470,11 @@ def user_activity_list(context, data_dict):
     # FIXME: Filter out activities whose subject or object the user is not
     # authorized to read.
     data_dict['include_data'] = False
+
+    auth_user_obj_ = context['auth_user_obj']
+    if auth_user_obj_ is None or not auth_user_obj_.sysadmin:
+        data_dict['include_data'] = True
+
     _check_access('user_activity_list', context, data_dict)
 
     model = context['model']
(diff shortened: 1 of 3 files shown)


TomeCirun:
Hey @wardi can you please tell me where am i wrong?
Thanks

wardi:
@TomeCirun yes if the behavior has changed please update the tests. If there is new functionality add a test to cover that too.

TomeCirun:
@wardi i will do so 
Many Thanks

PR #6262, opened 2021-07-16
Clean old LazyViews

pdelboca:
This PR cleans old LazyViews created years ago for backward compatibility.

### DIFF
--- a/ckan/views/__init__.py
+++ b/ckan/views/__init__.py
@@ -21,24 +21,6 @@ APIKEY_HEADER_NAME_KEY = u'apikey_header_name'
 APIKEY_HEADER_NAME_DEFAULT = u'X-CKAN-API-Key'
 
 
-class LazyView(object):
-
-    def __init__(self, import_name, view_name=None):
-        self.__module__, self.__name__ = import_name.rsplit(u'.', 1)
-        self.import_name = import_name
-        self.view_name = view_name
-
-    @cached_property
-    def view(self):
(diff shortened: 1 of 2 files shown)


amercader:
I know I said it might be worth keeping support for these old endpoints, but they are pretty terrible to be honest:

/dataset/new_resource/<dataset_id>
/dataset/<dataset_id>/resource_edit/<resource_id>

The new ones are: 

/dataset/<dataset_id>/resource/new
/dataset/<dataset_id>/resource/<resource_id>/new

Any extension using `url_for` should get the new endpoint, so I think the removal of complexity outweighs some potential compatibility.

This might be unrelated @smotornyuk but do you know why we are registering two resource blueprints here?

https://github.com/ckan/ckan/blob/2dd94c19c7514d1ffd9230525d1648b1b8f50d4e/ckan/views/resource.py#L37-L48

smotornyuk:
It's unrelated, yes.

The second one, `resource` is the basic resource view for naive usage.
The first one, `dataset_resource` is there for compatibility with the custom dataset types. Every package type registers its own set of rules (i.e, `package[type=study]` will register `study_resource` blueprint, `package[type=publication]` will register `publication_resource` blueprint and so on). That means, we have to use `h.url_for(f"{package_type}_resource.download")` constructions in order to get the best URL for the particular package type. And in order to make this style of url-generation compatible with the `package[type=dataset]` we have to create this `dataset_resource` blueprint

PR #6263, opened 2021-07-17
Clean _make_menu_item helper

pdelboca:
This PR aims to clean the helpers to build navigation bars and some old code and configs that are now obsolete.

- Removing `config['routes.named_routes']` should be straight forward since we don't require a mapping anymore. (assuming that all templates are updated to use flask routes when creating nav links._
- `_make_menu_item` will no longer raise an exception if the helper (in the template) is missing a required argument for the endpoint. I think this is an acceptable tradeoff aiming to clean lot of legacy code that it is still tied to old Pylons logic.

### DIFF
--- a/changes/6263.removal
+++ b/changes/6263.removal
@@ -1,0 +1,2 @@
+`build_nav_main()`, `build_nav_icon()` and `build_nav()` helpers no longer support
+Pylons route syntax. eg use `dataset.search` instead of `controller=dataset, action=search`.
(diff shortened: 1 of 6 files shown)


amercader:
This looks good @pdelboca but we should add a note to the CHANGELOG (a `changes/6263.removal` file) about dropping support for old pylons-based routes on the helpers like `build_nav_main()` and `link_to()`, eg use `dataset.search` instead of `search` or `dataset_search`.

pdelboca:
Done @amercader ! Let me know if that's okay :)

frafra:
Will this be backported to 2.9?

PR #6264, opened 2021-07-17
Migrate url_for calls to Flask syntax

pdelboca:
This PR migrates some leftovers of `url_for` method calls using the old Pylons syntax.


tino097 (approved):
LGTM

PR #6267, opened 2021-07-19
[#4807] Remove usage of six

### LINKED ISSUE #4807: Remove all remaining Pylons code in core
This should be done in separate pull requests to make reviews easier

- [x] Remove `requirements.py2.*` files and update documentation to remove py2 mentions
- [x] Remove py2 specific code. Look for `if six.PY2:` and remove what's inside!
- [x] Remove py3 checks. Look for `six.PY3` and make that the standard code run (remove the check if any)
- [ ] Remove usage of six. We should not need the compatibility layer any more. Make all code standard Python 3

This should definitely be done separately

- [ ] Remove unicode literals (eg `my_conf[u"test_key_1"] = u"Test value 1"` -> `my_conf["test_key_1"] = "Test value 1"`

steveoni:
Fixes #4807 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -9,7 +9,7 @@ import pytz
 import six
 import sqlalchemy
 
-from six.moves.urllib.parse import urlparse
+from urllib.parse import urlparse
 
 import ckan.model as model
 import ckan.plugins as p
--- a/ckan/config/middleware/common_middleware.py
+++ b/ckan/config/middleware/common_middleware.py
@@ -4,7 +4,7 @@
 import hashlib
 
 import six
-from six.moves.urllib.parse import unquote, urlparse
+from urllib.parse import unquote, urlparse
 
 import sqlalchemy as sa
 
--- a/ckan/lib/cli.py
+++ b/ckan/lib/cli.py
@@ -9,7 +9,7 @@ import paste.script
 import routes
 from paste.registry import Registry
 from six.moves import input
-from six.moves.urllib.parse import urlparse
+from urllib.parse import urlparse
 
 from ckan.config.middleware import make_app
 from ckan.cli import load_config as _get_config
--- a/ckan/lib/datapreview.py
+++ b/ckan/lib/datapreview.py
@@ -7,7 +7,7 @@ Functions and data structures that are needed for the ckan data preview.
 
 import logging
 
-from six.moves.urllib.parse import urlparse
+from urllib.parse import urlparse
 
 from ckan.common import config
 
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -13,7 +13,7 @@ which builds the dictionary by iterating over the table columns.
 '''
 import copy
 import six
-from six.moves.urllib.parse import urlsplit
+from urllib.parse import urlsplit
 
 from ckan.common import config
 from sqlalchemy.sql import select
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -31,7 +31,7 @@ from ckan.lib import i18n
 
 import six
 from six import string_types, text_type
-from six.moves.urllib.parse import (
+from urllib.parse import (
     urlencode, quote, unquote, urlparse, urlunparse
 )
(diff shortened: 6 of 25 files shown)


steveoni:
@amercader this PR is ready for review.

between, some `six` functions such as `ensure_binary`, `ensure_string` etc. are not removed, caused they involve writing functions to check `binary_type` and `test_type`, and instead of duplicate the code as it is here https://github.com/benjaminp/six/blob/3974f0c4f6700a5821b451abddff8b3ba6b2a04f/six.py#L898 we can make use of the one available in `six` as it is currently in-use.

amercader:
Many thanks @steveoni , this looks great.

I think we can merge this as is if you fix the conflict but we should aim to remove all six usage, including `six.ensure_binary()` and `six.ensure_text()` (again, this can be a separate PR to make it easier to review)

We were using `six.ensure_binary()` and `six.ensure_text()` to handle situations were the str/bytes behaviour was different on py2 and py3. We should no longer need these and we should aim to use the expected py3 syntax.

Rather than using `six.ensure_binary()` we can:

* Mark a string as bytes with `b` for instance [here](https://github.com/ckan/ckan/blob/75fc5d206632ef316398d755862f1ab4c0552576/ckan/tests/controllers/test_api.py#L35):
```diff
diff --git a/ckan/tests/controllers/test_api.py b/ckan/tests/controllers/test_api.py
index 1615caeac..ac8fb987f 100644
--- a/ckan/tests/controllers/test_api.py
+++ b/ckan/tests/controllers/test_api.py
@@ -32,7 +32,7 @@ class TestApiController(object):
         )
         env = {"REMOTE_USER": six.ensure_str(user["name"])}
 
-        content = six.ensure_binary('upload-content')
+        content = b'upload-content'
         upload_content = BytesIO(content)
         postparams = {
             "name": "test-flask-upload",
```

* Use the `.encode()` function eg [here](https://github.com/ckan/ckan/blob/75fc5d206632ef316398d755862f1ab4c0552576/ckan/tests/model/test_user.py#L27):
```diff
diff --git a/ckan/tests/model/test_user.py b/ckan/tests/model/test_user.py
index 960b8d2fb..bd8a70995 100644
--- a/ckan/tests/model/test_user.py
+++ b/ckan/tests/model/test_user.py
@@ -24,11 +24,11 @@ def _set_password(password):
     #     password_8bit = password
 
     salt = hashlib.sha1(os.urandom(60))
-    hash = hashlib.sha1(six.ensure_binary(password + salt.hexdigest()))
+    hash = hashlib.sha1((password + salt.hexdigest()).encode())
     hashed_password = salt.hexdigest() + hash.hexdigest()

```

* Or use the `bytes('some_string', 'utf8')` function, which should have the same effect

All these three methods should give you a bytes object which is what `six.ensure_binary()` did.

Similarly, we can replace `six.ensure_text()` with:

* Using the `str(bytes_obj,  'utf8')`
* Use the `.decode()` method on a bytes object

steveoni:
ok. i will do that; resolve the conflict and create a new PR for removing `six.ensure_binary` and `six.ensure_text`

steveoni:
@amercader the conflict is resolved now.

But one of the test is actually failing, but the test is not failing locally.

```
assert not 2
 +  where 2 = <Result SystemExit(2)>.exit_code
self = <ckan.tests.cli.test_user.TestApiToken object at 0x7f9c73fe7fd0>
cli = <ckan.tests.helpers.CKANCliRunner object at 0x7f9c74510250>

    def test_revoke(self, cli):
        user = factories.User()
        call_action(u"api_token_create", user=user[u"id"], name=u"first token")
        tid = model.Session.query(model.ApiToken).first().id
        args = [
            u"user",
            u"token",
            u"revoke",
            tid,
        ]
        result = cli.invoke(ckan, args)
>       assert not result.exit_code, result.output
E       AssertionError: Usage: ckan user token revoke [OPTIONS] ID
E         Try 'ckan user token revoke --help' for help.
E         
E         Error: no such option: -b
E         
E       assert not 2
E        +  where 2 = <Result SystemExit(2)>.exit_code

ckan/tests/cli/test_user.py:96: AssertionError
```

PR #6274, opened 2021-07-21
changes view for organizations now show changes in extras

### LINKED ISSUE #6037: Changes view for organizations does not show changes in extras
**CKAN version**
master

**Describe the bug**
Changes view was implemented in #5985, it does not however show changed in extras.

**Steps to reproduce**
Modify organization custom fields, view changes.

**Expected behavior**
The change in extras should be show in changes view.

TomeCirun:
Fixes #6037

### Proposed fixes:
i pass extras=data['extras'] in table_dictize()

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -731,7 +731,7 @@ def _group_or_org_update(context, data_dict, is_org=False):
                 'deleted organization' if is_org else 'deleted group'
     if activity_dict is not None:
         activity_dict['data'] = {
-                'group': dictization.table_dictize(group, context)
+                'group': dictization.table_dictize(group, context, extras=data['extras'])
                 }
         activity_create_context = {
             'model': model,


TomeCirun:
@Zharktas can you please tell me where am i wrong?  Thanks

Zharktas:
data doesn't have extras, at least not in all cases, which results in failing tests. And I think that update.py is wrong place for this change as changes view should not call update functions.

TomeCirun:
hey @Zharktas thanks for your reply. 

If we look at `views/group` at line https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/views/group.py#L524
we can see that we are calling `activity_diff` from `logic/action/get` which calls the `get` method from Activity model https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/logic/action/get.py#L3326
at this point we dont have extras.. because when we create/update the activity for organization/group we are not including extras https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/logic/action/create.py#L754

Thanks

TomeCirun:
Here is how it looks like on my screen 

image

Gauravp-NEC:
@TomeCirun , As I can see that you have pass extras=data['extras'] in table_dictize() do you have also called extra field function for the same?

TomeCirun:
Hey @amercader , @Zharktas here is my research:

https://github.com/ckan/ckan/blob/220fafd9b49efb3c26352383eb99fb80d9372ef1/ckan/logic/action/update.py#L732-L743

As we can see, exactly here in update we are creating the activity for organization.. till point 732 in the code we have group.extras after that we are calling table_dictize with 2 arguments, the one we are interested in, is group object, we are passing it with extras included, but what we get is only the fields from Group table and this is why: 

https://github.com/ckan/ckan/blob/220fafd9b49efb3c26352383eb99fb80d9372ef1/ckan/lib/dictization/__init__.py#L28-L62

As we can see at line 36, we are passing the group model, iterating over Group model and retrieving the group fields only ... and later we are writing down at line 50 in result_dict and at line 69 we are returning that dict. I think that's why we have no extras...

As we know different functions are calling table_dictize, so we cannot modify only for this purpose... but thats why we have **kw so we can pass extra arguments and update result_dict with that arguments we pass

https://github.com/ckan/ckan/blob/220fafd9b49efb3c26352383eb99fb80d9372ef1/ckan/lib/dictization/__init__.py#L62

Thanks

TomeCirun:
### hey @zharktas here is the changes i have made:

-added template for extra_fields
-included custom fields when creating an organization
-changed table_dictize function with group_dictize (i think group_dictize fits better than table_dictize.. because table_dictize gives us only the fields from the group table... this would be good if we have the group_extras in group table but unfortunately we dont, on other hand... group_dictize actually call table_dictize within + adds more fields optionally to it like tags, users,groups, packages etc..  )
-added _extra_fields() on check_metadata_org_changes() in ckan.lib.changes.py

Please, if you have any suggestion on how to do it better.. i will be glad to do it.
Thanks.

TomeCirun:
@Zharktas can you please review my PR ?
Thanks

PR #6276, opened 2021-07-21
Clean breadcrumb code

pdelboca:
This PR is a first step to clean the breadcrumb code overall the application for a modern approach on building it.

Changes:
 - Replace `h.nav_link()` for `{% link_for %}` to have a more consistent approach across the templates (Most of them currently are using the snippet
 - Clean old `breadcrumb.html` template in development folder.
 - Clean `breadcrumb_content_selected` block. Supposedly it is use to detect and set the `class="active"` but its hard to follow and it's currently buggy (Is not working for example in the dataset changes template). I cleaned it and replaced some `super()` call with duplicated code. This may seem counter intuitive but [we are currently using it](https://github.com/ckan/ckan/blob/53b9442509b46525d653f2f705e98319752ceb2d/ckan/templates/user/edit.html#L5-L9) and also it's the way [Django implements some of it's breadcrumbs](https://github.com/django/django/blob/012f38f9594b35743e9ab231757b7b62db638323/django/contrib/admin/templates/admin/delete_confirmation.html#L12-L20). I think is a good tradeoff to reduce complexity.

Finally:

This PR removes all the calls to `build_nav(...)` in the templates. It's still used in some other places but it's a good step forward to clean up the helpers library.

### DIFF
--- a/ckan/public/base/less/toolbar.less
+++ b/ckan/public/base/less/toolbar.less
@@ -19,26 +19,6 @@
     border: none;
     background: none;
     font-size: @font-size-large;
-    // line-height: 1.3;
-}
-
-// .toolbar .breadcrumb li:after {
-//   content: " / ";
-// }
-// .toolbar .breadcrumb li.active:after {
-//   content: "";
-// }
-// Needs to be a separate declaration as IE chokes on the last-of-type.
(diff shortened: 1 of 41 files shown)


pdelboca:
I'm closing this one until we migrate to a newer version of Bootstrap.

Issue #6159, opened 2021-06-16
Chained actions, auth functions and helpers are not documented


Zharktas:
**CKAN version**
2.7, 2,8, 2.9, master

**Describe the bug**

Chained actions are missing docstring completely:
https://github.com/ckan/ckan/blob/e89ca125dc68ddb9fe9bad68a401404146ba90c7/ckan/logic/__init__.py#L334-L336
There is a mention of chained actions within get_actions:
https://github.com/ckan/ckan/blob/e89ca125dc68ddb9fe9bad68a401404146ba90c7/ckan/plugins/interfaces.py#L922-L935

Chained auth functions have minimal docstring:
https://github.com/ckan/ckan/blob/e89ca125dc68ddb9fe9bad68a401404146ba90c7/ckan/logic/__init__.py#L634-L639

Chained helpers have minimal docstring:
https://github.com/ckan/ckan/blob/e89ca125dc68ddb9fe9bad68a401404146ba90c7/ckan/lib/helpers.py#L141-L145

Changelogs have minor mentions to them. `chained_action` was added in 2.7 https://github.com/ckan/ckan/pull/3494 and 2.9 docs have a mention to them:
![image](https://user-images.githubusercontent.com/830663/122172865-df78a800-ce89-11eb-852f-966809ddc474.png)

`chained_auth_function` was added in 2.8 https://github.com/ckan/ckan/pull/3995 and docs have a small mention of them:
![image](https://user-images.githubusercontent.com/830663/122173107-19e24500-ce8a-11eb-92d1-46f2967480e3.png)

`chained_helper` is only in master https://github.com/ckan/ckan/pull/5337 and are not mentioned in the changelog. 

There is a minor example how to use these in:
https://github.com/ckan/ckan/blob/e89ca125dc68ddb9fe9bad68a401404146ba90c7/ckanext/chained_functions/plugin.py#L32-L46

PR #6284, opened 2021-07-26
Updated Docstrings for after_update/after_delete

EricSoroos:
Fixes / Related to #6159

### Proposed fixes:

The after_update and after_delete callbacks aren't always called
when the underlying dataset changes. This enumerates the cases where
it happens, and what to do about it.

Not having this documentation led to some headscratching on a site where we're doing some special work with permissions. 

- [ ] includes API changes

### DIFF
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -494,6 +494,11 @@ class IPackageController(Interface):
         u'''
         Extensions will receive the validated data dict after the dataset
         has been updated.
+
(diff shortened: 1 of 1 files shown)


EricSoroos:
Add case for Datapusher / ckanext-xloader

EricSoroos:
Rebased to current master.

EricSoroos:
@wardi This is a doc only fix that should still be valid for 2.11

pdelboca:
Thanks @EricSoroos ! I'm merging this since the new docs seems okay!

PR #6289, opened 2021-07-27
[#6288] Replace python3.6 with python3.8 as 3.6 is not supported by Ubuntu Focal

### LINKED ISSUE #6288: Wrong Python version in Dockerfile
**CKAN version**
master

**Describe the bug**
Dockerfile tries to install python 3.6 on Ubuntu Focal image, but apt-get installs 3.8 instead. On Ubuntu 20.04 apt-get just ignores that package python3.6 is missing and is installing python3.8 as a dependency to package python-dev. Using `apt` instead of `apt-get` would result in an error.

**Steps to reproduce**
Run `docker build .`  on master branch and watch output

**Expected behavior**
Replace python3.6 package with python3.8.

**Additional details**
Output of apt-get on Ubuntu 20.04
```➜  ~ sudo apt-get -y install python3.6 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'python3.6-2to3' for regex 'python3.6'
Note, selecting 'libpython3.6-stdlib' for regex 'python3.6'
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

fulior:
Fixes #6288

### Proposed fixes:
Use right Python version on Ubuntu 20.04, as current Dockerfile is misleading, it's already using Python 3.8 but Dockerfile suggests it's 3.6

### Features:

- [ ] includes API changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN update-locale LANG=${LC_ALL}
 RUN apt-get -q -y update \
     && DEBIAN_FRONTEND=noninteractive apt-get -q -y upgrade \
     && apt-get -q -y install \
-        python3.6 \
+        python3.8 \
         python3-dev \
         python3-pip \
         python3-venv \


kowh-ai (approved):
I have just tested this PR and it works - Python **3.8.10** is now installed in the CKAN container (as opposed to **3.8.5**)

PR #6290, opened 2021-07-28
Fix unpriviledged users being able to access bulk process

f-osorio:
Any user was able to access and view the bulk process page for organization
management. This fix checks access and returns unauthorized exception
if theuser shouldn't be there.

Fixes #

### Proposed fixes:
Added `check_access` for `group update` and a catch statement for `NotAuthorized`

### DIFF
--- a/ckan/views/group.py
+++ b/ckan/views/group.py
@@ -856,6 +856,9 @@ class BulkProcessView(MethodView):
         try:
             group_dict = _action(u'group_show')(context, data_dict)
             group = context['group']
+            check_access(u'group_update', context)
+        except NotAuthorized:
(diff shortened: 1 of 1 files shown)


smotornyuk:
I see that the post-handler of the same view uses the `bulk_update_public` auth function. Can you do the same in the get-handler? Or even extract this auth-check and put it into the `_prepare` method.

f-osorio:
> I see that the post-handler of the same view uses the `bulk_update_public` auth function. Can you do the same in the get-handler? Or even extract this auth-check and put it into the `_prepare` method.

Right, I'll get that in soon.

PR #6293, opened 2021-07-28
cors: "no content"+access-control-allow-headers response

### LINKED ISSUE #6292: CORS headers for swagger (open api) et. al.
**CKAN version**
all

**Describe the bug**
When developing a customized swagger UI for our CKAN sites [1] I was unable to include POST requests as part of the interface

**Steps to reproduce**
Try using a cross-domain AJAX POST request to a ckan site with `ckan.cors.origin_allow_all` set to `true`

**Expected behavior**
Call should succeed

**Additional details**
POST call is blocked because an initial OPTIONS call is made to the serve and the response the wrong response code and has no `Access-Control-Allow-Headers` header line

[1]: https://open-data.github.io/OpenCanadaAPI/index-en.html (in progress)

wardi:
Fixes #6292 

### Proposed fixes:

respond with "no content" on empty responses and include access-control-allow-headers to make post requests to the API from other sites possible

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -399,6 +399,12 @@ def ckan_after_request(response):
 
     log.info(' %s render time %.3f seconds' % (url, r_time))
 
+    # 204 "No content" response required for CORS OPTIONS calls
+    if not response.data and response.status_code == 200:
+        response.status_code = 204
+        response.headers.pop('Content-Type', None)
+        response.headers['Access-Control-Allow-Headers'] = 'Authorization, Content-Type, Accept'
+
     return response
 
 


wardi:
probably not best place for the change. Might be better to specifically have a blueprint for API OPTIONS calls. Should have a test too.

smotornyuk:
Just in case.. won't [flask-cors](https://github.com/corydolphin/flask-cors/blob/master/flask_cors/core.py#L203) solve this problem?

wardi:
@smotornyuk not sure if it adds the response code I needed. For my case it might be simpler to add an IBlueprint plugin that does something like the extension you linked.

pdelboca:
Closing due to inactivity. Feel free to reopen it if consider that it is still relevant.

PR #6297, opened 2021-07-31
Clean old banner logic

pdelboca:
This PR cleans the logic for an old banner. 

Currently there is no logical call to it, the few that exists on templates default to false and no view set that template variable.


pdelboca:
@Zharktas I fixed the conflict in this one! Feel free to merge whenever you have time :)

pdelboca:
@Zharktas I think we are safe to merge this right?

pdelboca:
@Zharktas polite reminder :)

Zharktas:
@pdelboca sorry for taking so long with this one. There are now old and new templates which both have the old banners, would you remove them from public-bs3 and templates-bs3 as well :)

Zharktas (approved):

PR #6299, opened 2021-08-01
Refactor truncate logic to use jinja's built-in

pdelboca:
**Summary:**
This PR migrates truncate helper from CKAN's `h.truncate(text, 120)` to jinja2's `text|truncate(120)`. ([docs truncate](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.truncate))

**String truncate logic is inconsistent across templates:**
 - Sometimes we use jinja filter and other times our custom helper
 - There is no clear default for the length, it varies across templates.
 - Some templates defines custom `truncate` and `truncate_title` parameters and others don't
 - Some templates defines custom `truncate` variables but never uses it (like `group.html`)

**This PR is to simplify the logic:**
 - Removes inconsistent and complex behaviour beind `truncate` and `truncate_title` snippets parameters.
 - Defines a default directly when calling the `truncate` filter
 - Normalizes the use of jinja's biult-in filter in all the templates
 - Adds a deprecation warning in our custom helper


smotornyuk:
+1 for marking `h.truncate` as deprecated.

Doubts:

* should you add the `break_long_words=False` parameter to the `shorten` call in order to mimic `whole_word` of `h.truncate`?
* what is the purpose of removing `truncate` \ `truncate_title`? It will make snippets less flexible. In addition, this will change the appearance of the portals after the update. Are there some reasons to do it?

pdelboca:
Hey, thanks for the comments @smotornyuk !  I'm updating this to draft to make some changes and extra tests.

Regarding 1), I'll review and add the parameter. 

`shorten(...)` has a different behavior than our `truncate(...)` in which it doesn't split words in the same way if they're too long.

```python
In [13]: shorten('abc defghij', 8, break_long_words=False, placeholder='...')
Out[13]: 'abc...'

In [14]: shorten('abc defghij', 8, break_long_words=True, placeholder='...')
Out[14]: 'abc...'

In [15]: h.truncate('abc defhhij', 8, '...')
Out[15]: 'abc d...'
```

So I will review this and add more details and some tests.

Regarding 2), the purpose is to gain simplicity and readability in our templates. I think that we need to refactor this because:
  - `truncate` and `truncate_title` names are confusing since looks like boolean when reading them (even if they are documented). So if maintained a more descriptive name should be given like `title_max_length` and `description_max_lenght`
  - If flexibility is desired: Why don't we use them in all templates? And what's the criteria for having them in one templates and not in others?
  - I don't think templates are more flexible for being able to choose the length of the truncate since most of the usecase will not have any difference between a text of 150 and 190, 240, etc. The logic behind truncate is mostly to not break CSS rather than provide a "useful truncated length"
  - There are some unnecessary logic like:
```jinja
{% if truncate == 0 %}
  <p>{{ h.markdown_extract(group.description)|urlize }}</p>
{% else %}
  <p>{{ h.markdown_extract(group.description, truncate)|urlize }}</p>
{% endif %}
```

But it's just a matter of preference and it's up to debate (I'm ok if we decide to keep them) but I rather have more simple templates. I just feel it adds more noise than value.

pdelboca:
@smotornyuk I have been digging a little bit about the `break_long_words` parameters and it seems that it cannot be used to mimic our `whole_word`. `shorten` is a word-based shorten while our `truncate` its a character-based shorten (and also `jinja2`). This means that `shorten` will remove **words** to fit the width while `truncate` will remove **characters**.

This means that the behavior of jinjas `truncate()` and `h.markdown_extract()` will be a little bit different since the second will remove whole words. (Which I think it's not a big deal)

PR #6303, opened 2021-08-02
Replace glyphicons with fontawesome

pdelboca:
Bootstrap 4 dropped the Glyphicons icon font: https://getbootstrap.com/docs/4.6/migration/#components

This are the only uses I found in the codebase, so one less thing to do when migrating Bootstrap to a newer version.


pdelboca:
Can we merge this one @Zharktas ?

wardi:
:+1: I've found fontawesome icons load faster too

pdelboca:
@Zharktas polite reminder :)

Zharktas (approved):

PR #6327, opened 2021-08-16
[#6070] IUserForm

### LINKED ISSUE #6070: Spec: IUserForm
Create a new plugin interface IUserForm that allows extending the user create/update form and validation for users the same way IDatasetForm does for packages and IGroupForm does for orgs and groups.

The user model has plugin_extras #5382 to store extra metadata fields. plugin_extras are namespaced at the top level so that separate plugins could each add internal fields they need. For metadata we're adding to the form and public API let's use a common namespace e.g. `"public"`. Extra user metadata fields can be accessed at the model level using `user.plugin_extras["public"]["field-name-here"]`. We'll need new converters to store and load these values during validation:

```python
def store_public_extra(key, data, errors, context):
    """Copy data[key] to plugin_extras["public"][key] for storing in the db"""

def load_public_extra(key, data, errors, context):
    """Copy pugin_extras["public"][key] to data[key] for exposing to forms and API"""
```

These converters should behave like `convert_to_extras` and `convert_from_extras` do for package and group/org extra fields.

Like IDatasetForm and IGroupForm, IUserForm should allow the user form and display snippets to be overridden so that extra fields can appear the same as core fields.

Users do not have a `type` field so there is no need to support "default" or multiple simultaneous IUserForm plugins in the initial implementation.

steveoni:
Fixes #6070 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/converters.py
+++ b/ckan/logic/converters.py
@@ -220,3 +220,23 @@ def remove_whitespace(value, context):
     if isinstance(value, string_types):
         return value.strip()
     return value
+
+
+def store_public_extra(key, data, errors, context):
+    """Copy data[key] to plugin_extras["public"][key] for storing in the db"""
+    model = context['model']
+    user = context['user']
+    user = model.User.get(user)
+    if user.plugin_extras:
+        user.plugin_extras["public"][key] = data[key]
+    else:
+        user.plugin_extras = {"public": {}}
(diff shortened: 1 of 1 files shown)


pdelboca:
Closing due to inactivity.

Feel free to reopen it if consider that it is still relevant.

Thanks!

PR #6338, opened 2021-08-24
Self user details for logged in user

### LINKED ISSUE #5490: There is no way to determine the user ID when connecting through the API

shubham-mahajan:
Fixes #5490 

### Proposed fixes:
As discussed in https://github.com/ckan/ckan/pull/5504#issuecomment-760974628, we are not creating an extra action but adding the functionality in the current `user_show` action. So added the check to validate whether the user is logged in and if logged in return the user details.

### Features:

- [x] includes API changes

### DIFF
--- a/changes/5490.bugfix
+++ b/changes/5490.bugfix
@@ -1,1 +1,0 @@
-Add functionality for the user_show to fetch own details when logged in without passing id or user_obj
--- a/changes/5490.bugfix
+++ b/changes/5490.bugfix
@@ -1,0 +1,1 @@
+Add functionality for the user_show to fetch own details when logged in without passing id or user_obj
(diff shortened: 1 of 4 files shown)


shubham-mahajan:
@wardi Anything needed for this PR?

PR #6339, opened 2021-08-24
[#5727] Fix datapusher trigger in case of resource_update 

### LINKED ISSUE #5727: Datapusher being triggered before file is uploaded (update)
**CKAN version**
v2.9, 2.8, 2.7

**Describe the bug**
The Datapusher is triggered before the file is uploaded and causing the file to use the old URL and the new data is not uploaded to filestore until clicked manually

**Steps to reproduce**
1. Create a resource
2. Check the preview
3. Update the resource with a new file.
4. The preview is same.
5. Click `Upload to Datapusher`. The resource will be updated

**Expected behavior**
The data from the new file should be updated to the datastore

**Additional details**

We can fix this by removing the `notify` method and using the `after_update` hook of the resource.

`notify` is triggered as soon the URL is changed and updated in DB and not committed to SOLR yet and the `resource_show` method use the SOLR to fetch the data so the datapusher is fetching from the old URL instead of new URL or a new resource

shubham-mahajan:
Fixes #5727 

### Proposed fixes:
The notify method in the IResourceUrlChange will be triggered only when the URL is changed or if we do the resource update via the API, _submit_to_datapusher will not be triggered and cause the old preview to be displayed.

To fix that we need to add the `after_update` IResourceController to trigger `_submit_to_datapusher`.

### Features:

- [x] includes API changes


pdelboca:
Thanks @shubham-mahajan ! Looks good to me :)

cc @amercader

PR #6346, opened 2021-08-27
[#6290] prevent unauth user accessing bulk process [test fix]

smotornyuk:
Adds small fix to #6290


amercader:
This was backported to 2.9. AFAICT this is not an issue in [previous versions](https://github.com/ckan/ckan/blob/8d30a099d25b3b4c171fa74c3c92605ac46805f0/ckan/controllers/group.py#L388)

PR #6352, opened 2021-08-31
removed deprcated API-keys

### LINKED ISSUE #6247: Remove support for deprecated API keys
As discussed in one of the last dev meetings, old legacy API keys will no longer be supported in the next version. [API tokens](https://docs.ckan.org/en/2.9/api/index.html#api-authentication) are a much more secure and flexible alternative, and migrating is just a matter or switching one for the other when sending the `Authorization` header.

Tasks:

* Remove [the logic that checks for API keys](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/views/__init__.py#L228:L233) on incoming requests (maybe rename methods/vars from api_key -> api_token for consistency as well)
* Remove the API key snippet and "Regenerate API key" button from the user profile templates
* Remove the endpoint used to regenerate the API key and the `user_generate_apikey` action and auth function
* Remove `apikey` from the user schema
* Update documentation to make sure API keys are no longer mentioned or used in examples
* Some tests rely on test users' API keys (like [this one](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckanext/datastore/tests/test_search.py#L498)). This should be refactored to use a different auth method or to create an API token and use that instead.

I think for the next version it is fine to keep the model property and the database field and not add any migration to remove it, in case users want to use it for migrating.

TomeCirun:
Fixes #6247

### Proposed fixes:
-Renamed methods/vars from api_key -> api_token 
-Removed the  API key snippet and "Regenerate API key" button from the user profile templates
-Removed the endpoint used to regenerate the API key and the `user_generate_apikey` action and auth function
-Removed apikey from the user schema
-Instead of [api key](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckanext/datastore/tests/test_search.py#L498) i created API Token and used it for authorization.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/factories.py
+++ b/ckan/tests/factories.py
@@ -401,6 +401,32 @@ class SystemInfo(factory.Factory):
         return obj
 
 
+class APIToken(factory.Factory):
+    """A factory class for creating CKAN API Tokens"""
+
+    class Meta:
+        model = ckan.model.ApiToken
+
+    name = "first token"
+    user = factory.LazyAttribute(lambda _: User()["name"])
+
+    @classmethod
+    def _build(cls, target_class, *args, **kwargs):
+        raise NotImplementedError(".build() isn't supported in CKAN")
+
+    @classmethod
+    def _create(cls, target_class, *args, **kwargs):
+        if args:
+            assert False, "Positional args aren't supported, use keyword args."
+
+        context = {"user": kwargs["user"]}
+
+        token_create = helpers.call_action(
+            'api_token_create', context=context, **kwargs
+        )
+        return token_create["token"]
+
+
 def validator_data_dict():
     """Return a data dict with some arbitrary data in it, suitable to be passed
     to validator functions for testing.
--- a/ckan/tests/factories.py
+++ b/ckan/tests/factories.py
@@ -408,7 +408,6 @@ class APIToken(factory.Factory):
         model = ckan.model.ApiToken
 
     name = "first token"
-    user = factory.LazyAttribute(lambda _: User()["name"])
 
     @classmethod
     def _build(cls, target_class, *args, **kwargs):
@@ -419,10 +418,11 @@ class APIToken(factory.Factory):
         if args:
             assert False, "Positional args aren't supported, use keyword args."
 
-        context = {"user": kwargs["user"]}
+        context = {"user": _get_action_user_name(kwargs)}
+        data_dict = {"name": kwargs["name"], "user": kwargs["user"]["name"]}
 
         token_create = helpers.call_action(
-            'api_token_create', context=context, **kwargs
+            'api_token_create', context=context, **data_dict
         )
         return token_create["token"]
 
--- a/ckan/tests/factories.py
+++ b/ckan/tests/factories.py
@@ -418,11 +418,12 @@ class APIToken(factory.Factory):
         if args:
             assert False, "Positional args aren't supported, use keyword args."
 
-        context = {"user": _get_action_user_name(kwargs)}
-        data_dict = {"name": kwargs["name"], "user": kwargs["user"]["name"]}
+        target_user_name = _get_action_user_name(kwargs)
+        context = {"user": target_user_name}
+        kwargs["user"] = target_user_name
 
         token_create = helpers.call_action(
-            'api_token_create', context=context, **data_dict
+            'api_token_create', context=context, **kwargs
         )
         return token_create["token"]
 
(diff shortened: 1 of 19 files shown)


amercader:
@TomeCirun is this ready to review?

TomeCirun:
Yes @amercader, i thought it was better to make it from scratch.
Please let me know  if you have any suggestion for improvement.
Thanks

ckan/tests/lib/test_base.py
```diff
@@ -18,31 +20,45 @@ def test_comment_absent_if_debug_false(app):
...
+    }
 
     app.get("/dataset/new", headers=request_headers)
 
 
 @pytest.mark.usefixtures("clean_db", "with_request_context")
-def test_apikey_in_x_ckan_header(app):
+def test_apitoken_in_x_ckan_header(app):
     user = factories.Sysadmin()
+    user_token = h.call_action(
```
amercader, line 48: This pattern looks like it's going to be used a lot across the tests, so why don't we add a helper to make life easier for developers. Why don't we create a factory for them, eg `user_token = factories.APIToken(user=user_id)`

If you look at `factories.py` it should be easy to adapt an existing one calling the approapiate action.

amercader (commented):
This looks great @TomeCirun . Just one minor change to reduce boilerplate

ckan/tests/lib/test_base.py
TomeCirun: Yep, you are right, i will do it.
Thanks for your suggestion.

ckan/tests/factories.py
```diff
@@ -401,6 +401,32 @@ def _create(cls, target_class, *args, **kwargs):
...
+    @classmethod
+    def _build(cls, target_class, *args, **kwargs):
+        raise NotImplementedError(".build() isn't supported in CKAN")
+
+    @classmethod
+    def _create(cls, target_class, *args, **kwargs):
+        if args:
+            assert False, "Positional args aren't supported, use keyword args."
+
+        context = {"user": kwargs["user"]}
```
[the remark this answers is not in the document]
TomeCirun: Actually that func gives me TypeError.. that's why i have to pass it explicitly.. but i can make another helper func (e.g `def _get_user_from_kwargs(kwargs):` ) to make this work

https://github.com/ckan/ckan/blob/604e86ddf055a6c457d6d8a0c59deb1b4ddd9542/ckan/tests/factories.py#L65-L66
This is where i get the TypeError.

Thanks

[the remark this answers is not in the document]
amercader: It gives you `TypeError` because you should pass the full user dict as the `user` param in a factory, not just the name, eg change this:

```
user_token = factories.APIToken(user=user["name"])
```

with:

```
user_token = factories.APIToken(user=user)
```

(I actually think this is confusing and we should match the `context['user']` behaviour, but it's the current behaviour so let's keep it like that)

TomeCirun:
hey @amercader these are the changes i had to make in order the code to work:
-i added `data_dict` to separate `token_name` and `user_name` so i can pass it to `api_token_create` func... (please have in mind that now  `**kwargs` contains the whole user_dict ... we cannot pass it like that).
-Instead of user obj i replace them with factories one (in order to make `_get_action_user_name(kwargs)` work.)
-I adjusted `set_url_type()`  to work with user_dict (previously worked with user obj).
changes have to be done  also in `datapusher/tests/test.py` as this module calls `set_url_type`  from `datastore`

Whats your opinion over this?
Thanks

ckan/tests/factories.py
```diff
@@ -401,6 +401,32 @@ def _create(cls, target_class, *args, **kwargs):
...
+    def _create(cls, target_class, *args, **kwargs):
+        if args:
+            assert False, "Positional args aren't supported, use keyword args."
+
+        context = {"user": _get_action_user_name(kwargs)}
+        data_dict = {"name": kwargs["name"], "user": kwargs["user"]["name"]}
+
+        token_create = helpers.call_action(
+            'api_token_create', context=context, **data_dict
+        )
```
amercader, line 426: We still want to allow the `kwargs` to be passed to the action as data dict, but I see that we need to replace the `user` property, so can we try:

```suggestion

        target_user_name = _get_action_user_name(kwargs)
        context = {"user": target_user_name}
        kwargs["user"] = target_user_name

        token_create = helpers.call_action(
            'api_token_create', context=context, **kwargs
        )
```

The changes in the DataPusher tests are fine

TomeCirun: Thanks for your help @amercader

amercader:
@TomeCirun , I changed the changelog entry in f6e8a223e as users don't care about the internal changes, but how this affects scripts

Thanks for your work on this!

PR #6361, opened 2021-09-03
removed h.link_to

### LINKED ISSUE #6272: Remove helpers that output HTML
Replace them with Jinja blocks and snippets that can be easily overridden from plugins.

For instance:

```html+jinja
<td>{{ h.link_to(tag.name, h.url_for('dataset.search', tags=tag.name)) }}</td>
```

Doesn't look that much readable or easy to write than:
```html+jinja
<td><a href="{{ h.url_for('dataset.search', tags=tag.name)) }}"> {{ tag.name }} </td>
```

For higher level ones like `build_nav_icon()`:
https://github.com/ckan/ckan/blob/d513351db72731df48a38ce3c1621a5058f502c5/ckan/templates/user/read_base.html#L18-L24

we can add helper snippets snippets, but I suspect that following a similar approach just using the actual markup on the templates and calling the necessary helpers would be the easiest.

I think this can be a gradual task and we can assess what's the best replacement on a case by case basis.

Candidates:
* [`link_to()`](https://github.com/ckan/ckan/blob/d513351db72731df48a38ce3c1621a5058f502c5/ckan/lib/helpers.py#L867)
* [`nav_link()`](https://github.com/ckan/ckan/blob/d513351db72731df48a38ce3c1621a5058f502c5/ckan/lib/helpers.py#L888)
* [`build_nav_main()`](https://github.com/ckan/ckan/blob/d513351db72731df48a38ce3c1621a5058f502c5/ckan/lib/helpers.py#L913)
* [`build_nav_icon()`](https://github.com/ckan/ckan/blob/d513351db72731df48a38ce3c1621a5058f502c5/ckan/lib/helpers.py#L937)
* [`build_nav()`](https://github.com/ckan/ckan/blob/d513351db72731df48a38ce3c1621a5058f502c5/ckan/lib/helpers.py#L956)

TomeCirun:
Fixes #6272

### Proposed fixes:
I changed all h.link_to (and some h.nav_link that came my way) with appropriate HTML

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/admin/snippets/data_type.html
+++ b/ckan/templates/admin/snippets/data_type.html
@@ -18,7 +18,9 @@
     {% for entity in entities %}
       {% set title = entity.title or entity.name %}
         <li>
-          {{ h.link_to(title|truncate(80), h.url_for(entity.type + '.read', id=entity.name)) }}
+          <a href="{{ h.url_for(entity.type + '.read', id=entity.name) }}">
+            {{ title|truncate(80) }}
+          </a>
         </li>
       {% else %}
       <p>
--- a/ckan/templates/organization/bulk_process.html
+++ b/ckan/templates/organization/bulk_process.html
@@ -73,7 +73,9 @@
                         {{ _('Edit') }}
                       </a>
                       <h3 class="dataset-heading">
-                        {{ h.link_to(title|truncate(80), h.url_for(package.type ~ '.read', id=package.name)) }}
+                        <a href="{{ h.url_for(package.type ~ '.read', id=package.name) }}">
+                          {{ title|truncate(80) }}
+                        </a>
                         {% if package.get('state', '').startswith('draft') %}
                           <span class="label label-info">{{ _('Draft') }}</span>
                         {% elif package.get('state', '').startswith('deleted') %}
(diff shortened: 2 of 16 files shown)


TomeCirun:
hey @amercader i checked the tests and it seems that they have failed because of the line breaks.

Do you have any suggestion about this?
Thanks

amercader:
@TomeCirun tests that rely on whitespace in the markup are fragile so I think that what's best is to refactor the failing ones. For instance here:
https://github.com/ckan/ckan/blob/ad405b580332e5f73e933da8db6ed36a90adfcdf/ckan/tests/controllers/test_package.py#L2234-L2237

Instead we should use BeautifulSoup to parse the HTML response and get the required tags like in [here](https://github.com/ckan/ckan/blob/ad405b580332e5f73e933da8db6ed36a90adfcdf/ckan/tests/controllers/test_package.py#L444)
Alternatively, we can be pragmatic and just do:
```python
assert '<a href="/dataset/{}">'.format(dataset["id"]) in response
assert dataset['title'] in response
```

For failures like [these one](https://app.circleci.com/pipelines/github/ckan/ckan/2256/workflows/264a131f-1a82-40a4-aeee-29b1105a8a2b/jobs/9997/tests#failed-test-6) the test needs to `.strip()` spaces before doing the comparison

amercader:
@TomeCirun can you merge latest master to make sure all tests are green?

TomeCirun:
> That should fix it :)

Thank you @Zharktas

PR #6371, opened 2021-09-06
Remove control characters from package.notes

kentsanggds:
## What

Control characters are causing a problem for lxml parsing as strings must be XML compatible.
Control characters cause the `feedgen` module to throw an error leading to an Internal Server error.

Fixes #

Problem with control characters in package.notes which will throw this error - 

`ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters` which is called by `writeStriing` function in `views/feed.py`

Test it out by reverting back to the original https://github.com/ckan/ckan/pull/6371/files#diff-9e26466d662dfbb08832a2f76bd00a0ef6f52c3bfbc7edf6b2eca0115e3b345fL181 and then run the test to see the error raised when control characters are not removed.

### Proposed fixes:

Remove control characters from the package notes before pushing it to `feedgen` module.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/controllers/test_feed.py
+++ b/ckan/tests/controllers/test_feed.py
@@ -26,11 +26,18 @@ class TestFeeds(object):
         ), res
 
     def test_general_atom_feed_works(self, app):
-        dataset = factories.Dataset()
+        dataset = factories.Dataset(notes="Test\x0c Notes")
         offset = url_for(u"feeds.general")
         res = app.get(offset)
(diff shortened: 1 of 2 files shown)


amercader:
Please target the `master` branch on any new patch. If necessary, it will be backported to 2.9

kentsanggds:
ok, I'll do that

PR #6375, opened 2021-09-07
Add pre-commit

smotornyuk:
* Initialize basic [pre-commit](https://pre-commit.com/) hooks (no trailing spaces in line, file ends with the newline, yaml files are valid)
* Update contributing guide
* Update dev-requirements

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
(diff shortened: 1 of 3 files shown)


smotornyuk:
We discussed it and decided, that it's too early for CKAN to apply such massive automatizations. We need to go through a long path of refactoring and create a number of GitHub actions before enforcing such requirements via git hooks. 
We won't be there for, at least, a year, so I prefer to close this PR. In the future, there may appear better tools for doing this job. If not, we can always reopen it0

PR #6378, opened 2021-09-07
Remove control characters from package.notes

kentsanggds:
## What

Control characters are causing a problem for lxml parsing as strings must be XML compatible.
Control characters cause the `feedgen` module to throw an error leading to an Internal Server error.
Also have to pass empty string to the regex when no notes in the dataset as when there are no notes it returns a `None` not the expected ''.

Fixes #

Problem with control characters in package.notes which will throw this error - 

`ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters` which is called by `writeStriing` function in `views/feed.py`

Test it out by reverting back to the original https://github.com/ckan/ckan/pull/6371/files#diff-9e26466d662dfbb08832a2f76bd00a0ef6f52c3bfbc7edf6b2eca0115e3b345fL181 and then run the test to see the error raised when control characters are not removed.

### Proposed fixes:

Remove control characters from the package notes before pushing it to `feedgen` module.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/controllers/test_feed.py
+++ b/ckan/tests/controllers/test_feed.py
@@ -24,11 +24,18 @@ class TestFeeds(object):
         ), res
 
     def test_general_atom_feed_works(self, app):
-        dataset = factories.Dataset()
+        dataset = factories.Dataset(notes="Test\x0c Notes")
         offset = url_for(u"feeds.general")
         res = app.get(offset)
-
         assert helpers.body_contains(res, u"<title>{0}</title>".format(dataset["title"]))
+        assert helpers.body_contains(res, u"<content>Test Notes</content>")
+
+    def test_general_atom_feed_works_with_no_notes(self, app):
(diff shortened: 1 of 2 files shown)


smotornyuk:
Looks good, but can you try using [unicodedata](https://docs.python.org/3.6/library/unicodedata.html#unicodedata.category) and character categories ([wiki](https://en.wikipedia.org/wiki/Template:General_Category_(Unicode)), [anoter source](https://www.compart.com/en/unicode/category))

And you have two style issues: 
```
E           ------------------
E           ckan/views/feed.py
E           ------------------
E           W293 ln:147 blank line contains whitespace
E           E501 ln:184 line too long (80 > 79 characters)
E           
```

kentsanggds:
I've added another commit to use `unicodedata` instead of the regex and fixed style issues.

smotornyuk:
Perfect, thanks!

PR #6382, opened 2021-09-09
removed pg_buffercache from datastore_search

### LINKED ISSUE #6354: `_table_metadata` view contains references to PostgreSQL extension's tables.
**CKAN version**
master

**Describe the bug**
Datastore's `_table_metadata` view contains references to PostgreSQL extension's tables. 

**Steps to reproduce**
1. Setup an instance with datastore
2. Connect to the datastore database
3. Install an extension. eg: `datastore=# CREATE EXTENSION pg_buffercache;`
4. Execute a select statement or run `get_all_resources_ids_in_datastore()`

```
datastore=# select * from _table_metadata;

       _id        |                 name                 |  oid   | alias_of 
------------------+--------------------------------------+--------+----------
 c2f1f7c1455f528d | pg_buffercache                       | 193578 | 
 cb62df6b7321f939 | 3702ef71-6575-40ac-80af-fea0bb9f5d87 | 149899 | 
 3cb4d6d32ac884da | 83ab6651-1e54-4b20-a0ce-ca3ea5e55559 | 149465 | 
(3 rows)
```

**Expected behavior**
`_table_metadata` VIEW doesn't return system related tables.

**Additional details**
This bug can create non desired deletes or errors when running commands with logic like `datastore purge` that queries all the ids and then executes a `datastore_delete` on the result.

TomeCirun:
Fixes #6354

### Proposed fixes:
I  installed few more extensions and i saw they are not included in _table_metadata... somehow only pg_buffercache is..  and i think thats because it has reference in pg_class, pg_depend and pg_rewrite as we query it when creating the table.

image

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1317,6 +1317,7 @@ def search_data(context, data_dict):
     sort = query_dict['sort']
     limit = query_dict['limit']
     offset = query_dict['offset']
+    where = query_dict['where']
 
     if query_dict.get('distinct'):
         distinct = 'DISTINCT'
@@ -1325,6 +1326,10 @@ def search_data(context, data_dict):
 
     if not sort and not distinct:
         sort = ['_id']
+
+    if not where:
+        where_clause = 'WHERE %s' % (''.join("name != 'pg_buffercache'"))\
+            .replace('%', '%%')
 
     if sort:
         sort_clause = 'ORDER BY %s' % (', '.join(sort)).replace('%', '%%')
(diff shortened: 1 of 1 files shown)


pdelboca:
Hey @TomeCirun ! Thanks for the contribution. We've discussed this in the Dev Technical Meeting (you're always welcome to join!) but this implementation is not what we are aiming for.

The problem is that it doesn't fix nor refactor the `_table_metadata` VIEW but rather it filters the error later. It doesn't fix the error but rather patches the outcome of it. It would be a good patch for an custom install but not for the main branch.

The goal would be to either refactor this VIEW or create a new one. (Maybe the good for contribution tag is optimistic on this.)

Please let me know if you have further questions. We're gonna aim to add more detail to the main issue #6354.

TomeCirun:
hey @pdelboca  thanks for your reply.

I  investigated a little bit over this, actually this check causes pg_buffercache to be on the _table_metadata https://github.com/ckan/ckan/blob/77425a04ff2a956e8dfe04c1364c12f1d342874b/ckanext/datastore/set_permissions.sql#L66-L68

because pg_buffercache is a VIEW... and we are getting resource ids from pg_tables.

I tried to remove 
image
and this actually works ... but i m not sure what else we have in pg_views.. do we actually need it?

The question is why we need _table_metadata ? If only for resource ids, then we dont needed pg_views, we are getting the ids from pg_tables.

Thanks

Issue #6390, opened 2021-09-10
Users can register with an "empty" string for their email


michael-long88:
**CKAN version**
2.9.3

**Describe the bug**
Users are able to register with " " for the email.

**Steps to reproduce**
1. Go to registration page
2. Fill out information, entering " " for the email
3. Submit

**Expected behavior**
The system does not allow the user to register and provides an error explaining why

**Additional details**
I believe the issue originates here: https://github.com/ckan/ckan/blob/11f151bb13103913b4099c0d7cc37263f27ef28f/ckan/logic/schema.py#L417
The `user_new_form_schema`  "inherits" from the `default_user_schema` which has the appropriate validators for `email`. However, in `user_new_form_schema`, the `email` key gets overridden with just the `email_is_unique` validator. We've verified this by trying to register another user with " " as the email, and rather than producing an error saying the email is invalid, it says the email is already in use.

amercader:
Fixed in https://github.com/ckan/ckan/pull/6399

PR #6399, opened 2021-09-15
[#6390] fix user create/edit email validators

mutantsan:
Fix for #6390

### Features:

- [ ] includes API changes


amercader:
@mutantsan can you explain your changes here? It seems like you are removing the validator to ensure email uniqueness which might not be what we want right?

PR #6403, opened 2021-09-15
add tqdm progressbar helper and use for search-index progressbar

mutantsan:
Replaces search-index rebuild progressbar with an external module `tqdm`. It has many useful features, including the duration estimate that could be useful dealing with thousands of datasets.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -17,6 +17,7 @@ import copy
 import uuid
 import functools
 
+from tqdm import tqdm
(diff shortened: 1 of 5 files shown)


amercader:
Discussed on the dev meeting: maybe not worth adding a new requirement for this but it could definitely be added to the CLI documentation as example of how to implement commands in extensions using external libraries. Thanks for the submission @mutantsan

PR #6414, opened 2021-09-17
h.remove_url_param fail with minimal set of params

smotornyuk:
`h.remove_url_param` with the minimal set of parameters: 
```
h.remove_url_param("q")
```
will throw an error on the `organization.read` page(not only here). Because internally it gets current blueprint and view (`g.controller` and `g.action`) and tries to build the URL `h.url_for(blueprint + "." + view)`. And as long as the given blueprint requires extra arguments (`package_type` for dataset pages, `group_type`/`is_organization` for group pages, `id` for organization.read page, etc), this will fail. User has to do something like
```
h.remove_url_param("q", extras={"is_organization": group_dict.type, id=group_dict.id, "group_type": group_dict.type})
```

While it is perfectly ok from the clarity point of view, it requires every developer to know CKAN internals and these extra parameters. And it makes the given helper too fragile. In addition, the `extras` parameter is optional, so there may be some expectations, like 
**whenever I'm using `h.remove_url_param("q")` in the simplest form, I expect `q` to be removed from the CURRENT URL, whatever it is**

This PR takes all the extra parameters required to build the current URL and adds them **only when neither `controller` nor `action` nor `extras` argument is present**. I.e, only when the user trying to modify the current URL.

PS. And it also updates `toolkit.get_endpoint`. Now it won't raise RuntimeError in CLI. Instead, its returns `None, None`, which is quite safe, I think.

### DIFF
--- a/changes/6414.bugfix
+++ b/changes/6414.bugfix
@@ -1,0 +1,1 @@
+h.remove_url_param fail with minimal set of params
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1947,16 +1947,25 @@ def follow_count(obj_type, obj_id):
     return logic.get_action(action)(context, {'id': obj_id})
 
 
-def _create_url_with_params(params=None, controller=None, action=None,
-                            extras=None):
-    ''' internal function for building urls with parameters. '''
+def _create_url_with_params(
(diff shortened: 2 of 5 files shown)


frafra:
Will this be backported to CKAN 2.9?

PR #6439, opened 2021-09-28
concurent-safe resource updates

### LINKED ISSUE #6420: concurrent-safe resource updates
**CKAN version**
master

**Describe the bug**
The only way to safely update multiple resources in the same dataset at the same time is to use `package_revise` exclusively.

Any concurrent calls to `package_update`, `resource_create`, `resource_update`, `resource_delete`, `package_patch` or `resource_patch` can cause data loss because changes made in calls happening at the same time aren't visible.

**Steps to reproduce**
Attempt to modify multiple resources in the same dataset concurrently

**Expected behavior**
Possibly slower performance but no data loss

**Additional details**
Each of these APIs need to use `for_update=True` when calling `package_show` like https://github.com/ckan/ckan/blob/d20703f261c2f42fe034bbac7370943784b74e07/ckan/logic/action/update.py#L469-L474 to prevent concurrent updates losing data

TomeCirun:
Fixes #6420

### Features:

- [ ] includes API changes


TomeCirun:
hey @amercader  @smotornyuk it seems after merging of  #6335  i have a lot of  failed tests ... i can see that  `titles`, `passwords`, `names`, etc. are random now(using `faker`) and yet we have tests that needed to be changed to work that way ... i will show you one example :
https://github.com/ckan/ckan/blob/de31140c4860aacf38c10aeed379751f75cc012e/ckan/tests/controllers/test_group.py#L932-L939

here we can see that  Mr. Test User  and Test Dataset are hard-coded, and both of them are now created in а generic way.

What's your opinion over this ?
Thanks

amercader:
Mark as backport but we need to double check if it affects performance

ckan/logic/action/patch.py
```diff
@@ -69,6 +70,7 @@ def resource_patch(context, data_dict):
         'session': context['session'],
         'user': context['user'],
         'auth_user_obj': context['auth_user_obj'],
+        'for_update': True
```
wardi, line 73: resource_show attaches a resource model to the context before calling package_show with for_update in the context, so the resource object in our context might differ from the one returned by package_show. I guess this doesn't matter because we throw away the show_context, but it's the kind of thing I can see breaking as changes are made in the future.

I think some tests would be the best way to mitigate this and cover this all the new functionality in general. Can you think of a way to test this feature?

ckan/logic/action/update.py
```diff
@@ -545,7 +547,9 @@ def package_resource_reorder(context, data_dict):
     if len(set(order)) != len(order):
         raise ValidationError({'order': 'Must supply unique resource_ids'})
 
-    package_dict = _get_action('package_show')(context, {'id': id})
+    package_show_context = dict(context, for_update=True)
+    package_dict = _get_action('package_show')(
+        package_show_context, {'id': id})
```
wardi, line 552: good catch.

smotornyuk:
Yep, my bad. I haven't merged upstream into #6335 for quite a long time. Because of this it passed tests at the moment it was created, but created a mess after the merge.

Here #6453 is the fix for master branch

ckan/logic/action/patch.py
amercader: What sort of tests would we need here @wardi? a test with concurrent resource updates? or that reuse the same context between `resource_patch` calls?

wardi: kind of tricky to write a test of concurrent updates actually being serialized by postgres. Maybe good enough is a test that verifies the `for_update==True` reaches the dictization layer.

TomeCirun: Wardi, can you give some pointers so I can start making the test 
Thanks

amercader: @wardi I tried to write an example for Tome but I'm a bit lost with how the `for_update` flag works. You mention the dictization layer but AFAICT this never reaches that level and it is instead used at the [model level](https://github.com/ckan/ckan/blob/36839d57d443bd189056bc83cf3e684c72a1f5d5/ckan/model/package.py#L87). Is this what you meant?

wardi: yes, I misremembered.

wardi: @amercader you're right, the model level not dictization. I think we could mock that method just to verify that the actions do in fact pass the flag properly.

wardi:
@amercader WDYT should we merge this without tests covering the new flag? It's a pretty important change for preventing accidental data loss.

amercader:
@wardi I agree it's an important change that I can see could break in subtle ways in the future so I thought about spending some time coming up with a tests. Turns out is *extremely* difficult to properly mock actions but I finally came up with this:

```diff
diff --git a/ckan/tests/logic/action/test_patch.py b/ckan/tests/logic/action/test_patch.py
index c1479c9ad..75910febb 100644
--- a/ckan/tests/logic/action/test_patch.py
+++ b/ckan/tests/logic/action/test_patch.py
@@ -1,7 +1,9 @@
 # encoding: utf-8
 """Unit tests for ckan/logic/action/patch.py."""
+from unittest import mock
 import pytest
 
+from ckan.logic.action.get import package_show as core_package_show
 from ckan.tests import helpers, factories
 
 
@@ -190,3 +192,30 @@ class TestPatch(object):
 
         assert user2["fullname"] == "Mr. Test User"
         assert user2["about"] == "somethingnew"
+
+    def test_package_patch_for_update(self):
+
+        dataset = factories.Dataset()
+
+        mock_package_show = mock.MagicMock()
+        mock_package_show.side_effect = lambda context, data_dict: core_package_show(context, data_dict)
+
+        with mock.patch.dict('ckan.logic._actions', {'package_show': mock_package_show}):
+
+            helpers.call_action('package_patch', id=dataset['id'], notes='hey')
+
+            assert mock_package_show.call_args_list[0][0][0].get('for_update') is True
+
+    def test_resource_patch_for_update(self):
+
+        dataset = factories.Dataset()
+        resource = factories.Resource(package_id=dataset['id'])
+
+        mock_package_show = mock.MagicMock()
+        mock_package_show.side_effect = lambda context, data_dict: core_package_show(context, data_dict)
+
+        with mock.patch.dict('ckan.logic._actions', {'package_show': mock_package_show}):
+
+            helpers.call_action('resource_patch', id=resource['id'], description='hey')
+
+            assert mock_package_show.call_args_list[0][0][0].get('for_update') is True
```

This tests that the first call we make to `package_show` after calling `package_patch` / `resource_patch` has `{"for_update": True}` in the `context` object. At first I tried mocking `model.Package.get()`, which is the function that ultimately gets the `for_update` flag but it gets called loads of times in a single action, and the one we care about (the one that gets `for_update`) is not always the first one.
The `package_show` approach is not rock solid, we could potentially introduce a new `package_show` call before the one we are testing but I think it is very unlikely and at least we have some sort of testing.

If you are happy with this approach perhaps @TomeCirun can replicate the same tests for `resource_update` and `resource_delete`

wardi:
@amercader LGTM!

amercader:
@TomeCirun `resouce_delete` does another `package_show` call before the one we need so the tests needed to be tweaked a bit. I've finished off the tests in 311a84a79 and af685d483
Thanks for your work on this!

wardi:
Let's add a note in the changelog about the possible performance impact on some sites. We should have lots of room to improve `package_update` performance to address those issues while still having safe updates.

amercader:
2.9 version was backported here: 53adf09c2

PR #6466, opened 2021-10-12
[5690] User last active

### LINKED ISSUE #5690: There is no way to find out when a user last logged in
**CKAN version**
2.9.1

**Describe the bug**
The `user` table stores the `created` timestamp, but not last login.  This information is needed to:
 - purge idle users (#5658)
 - enforce password aging policies
 - help track utilization
 - help implement brute force lockout? (perhaps, by adding another column - `last_login_attempt`?)

**Steps to reproduce**
Register a user.

**Expected behavior**
A `last_login` column is added to the `user` table, similar to the `last_accessed` column in the new `api_token` table.

TomeCirun:
Fixes #5690

### Proposed fixes:
I added:
-`last_active` column  to `user` table
-`set_user_last_login` instance method
-`helper function`
-`migration file`

### Features:

- [ ] includes API changes


amercader:
@TomeCirun @smotornyuk as I mentioned on the [issue](https://github.com/ckan/ckan/issues/5690#issuecomment-717279141) I think that rather than tracking logins we should track the last time a user has been active, ie identified in a request. Otherwise we lose all API access. So basically have a `last_active` field instead of `last_login` and hook it to [`identify_user`](https://github.com/ckan/ckan/blob/04cc84540ffd7b00e70a04de6f84f1b0721ca838/ckan/views/__init__.py#L122).

This does mean a DB write on each request so maybe we can store the current last_active value in the session and only write it if its bigger than X mins (which can be configurable).

If all this seems to complicated and we want to stick with actual UI logins I'm fine with it.
But then @smotornyuk the signal you suggest would probably miss logins from alternative authentication methods like SAML2, OAuth etc right?

TomeCirun:
@amercader do you have any idea why this happens? 
Thanks

amercader:
@TomeCirun I don't know about these failures, but they are unrelated to your changes. I've created an issue to track them if you want to investigate in a separate PR: https://github.com/ckan/ckan/issues/6537

Thanks for your PR, we'll review it as soon as we have availability

TomeCirun:
Hey, @amercader is there anything else that can be done in this PR?
Thanks

amercader:
> is there anything preventing us to merge this PR?

Only time, and the relentless onslaught of the world against our attention.

Thanks for this, it's a very valuable contribution

PR #6469, opened 2021-10-13
Faster tests

smotornyuk:
Before:
image

After:
image

The last 16-min container is the one, that runs frontend tests. I've extracted them into separate jobs inside different PR

Main changes:
* `non_clean_db` - a fixture that guarantees that DB is available, but doesn't clean it. it means that you have to rely on random IDs and names, generated by factories:
  ```
  # instead of this
  name = "a"
  factories.Dataset(name=name)
  # use this
  name = factories.Dataset()["name"]
  ```
* Split tests into more or less equal chunks based on their execution time


jj-crypto (approved):

ckan/common.py
```diff
@@ -10,7 +10,7 @@
 from __future__ import annotations
 
 import logging
-from collections import MutableMapping
+from collections.abc import MutableMapping
```
wardi, line 13: If this was broken is it unused? can it be removed?

CHANGELOG.rst
```diff
@@ -522,7 +522,7 @@ v.2.8.10 2022-01-19
 
 Fixes:
 
-* Add timeouts to requests calls (see :ref:`ckan.requests.timeout`) (`#6408 <https://github.com/ckan/ckan/pull/6408>`_)
+* Add timeouts to requests calls (see `ckan.requests.timeout`) (`#6408 <https://github.com/ckan/ckan/pull/6408>`_)
```
wardi, line 525: Is removing this link supposed to be part of this PR?

wardi:
Looks good to me too, just a couple questions above about seemingly unrelated changes

ckan/common.py
smotornyuk: It works, but an import from `collections` is deprecated since python 3.7 and it's recommended to import from `collections.abc` instead. This class is used as a superclass for the config object.

CHANGELOG.rst
smotornyuk: There is no corresponding reference in the documentation and that leads to a warning during documentation build, which prevents tests from passing. 

So this problem is not related to this particular PR but without this change none of the new PRs can pass the tests

smotornyuk:
All issues solved / ready to final review

amercader:
Fantastic work @smotornyuk 

I've created https://github.com/ckan/ckan/issues/6700 so we don't forget to document these changes

PR #6477, opened 2021-10-15
Allow strict types for user/group uploads

smotornyuk:
User/Group images do not have any restriction on actual mime-type when the user uploads a file.

This PR adds the possibility to configure allowed types per `Uploader.object_type` (`user`, `group`, etc.). By default, the current behavior is unchanged. But one can allow only particular file types via `ckan.upload.<OBJECT_TYPE>.types` or `ckan.upload.<OBJECT_TYPE>.mimetypes`

Examples:

```
ckan.upload.user.types = image

ckan.upload.group.mimetypes = image/png image/svg+xml

ckan.upload.media.types = video audio
```

I think it also should be backported to 2.9


jj-crypto (approved):

hagaiwech:
@amercader Any updates on this?

ckan/lib/uploader.py
```diff
@@ -192,6 +194,25 @@ def upload(self, max_size=2):
             except OSError:
                 pass
 
+    def verify_type(self):
+        if not self.filename:
```
amercader, line 198: I suggest skipping all this code entirely if the `ckan.upload.{self.object_type}.types` config options are not set.

ckan/config/config_declaration.yaml
```diff
@@ -461,6 +461,26 @@ groups:
         type: int
         default: 2
 
+  - annotation: Uploader Settings
+    options:
+      - key: ckan.upload.user.types
```
amercader, line 466: Can you document these in `doc/configuration.rst`? IIRC this is still not automatically generated right?

changes/6477.bugfix
```diff
@@ -0,0 +1 @@
+Type of uploads for group and user image can be restricted via `ckan.upload.user.types` and `ckan.upload.group.types`
```
amercader, line 1:
```suggestion
Type of uploads for group and user imagec can be restricted via the `ckan.upload.{object_type}.types` and `ckan.upload.{object_type}.mimetypes` config options (eg `ckan.upload.group.types`, `ckan.upload.user.mimetypes`)
```

If you can link to the `doc/configuration.rst` docs with `:ref:` that would be ideal

amercader (commented):
Looks good some minor comments.

ckan/config/config_declaration.yaml
smotornyuk: Yes. We can use tokens that will be replaced declared defaults, but it must be done manually

amercader:
Nice one @smotornyuk

PR #6490, opened 2021-10-21
Skip db init for additional CKAN containers

frafra:
https://github.com/ckan/ckan/discussions/6489

### Proposed fixes:
If `CKAN_EXTRA` is set to `true`, then the CKAN container skips the database initialization. Useful for running additional components, such as ckanext-harvest containers.

### Features:

- [ ] includes API changes


frafra:
It would be better if CKAN had done the initialization of the database in a single transaction, actually, but it would require a bigger change. This is trivial instead

frafra:
Meeting notes (05/04/2022): look into alternatives (like healthchecks).

frafra:
Indeed, that can be achieved with healthchecks when using Docker compose.
Superseeded by #6812.

PR #6504, opened 2021-10-28
Fake data using CLI

smotornyuk:
Register new command `ckan generate fake-data [type] [--factory-class=IMPORT_PATH]` that creates entities using test factories.

I don't really like `ckan seed ..` CLI, which generates test data, because it's a one-time tool and one cannot just create any number of datasets using it. In addition, it uses hardcoded values and does not allow to creation of entities of any type. So I hope that eventually `ckan seed` will be removed in favor of `ckan generate fake-data`

### DIFF
--- a/ckan/cli/generate.py
+++ b/ckan/cli/generate.py
@@ -242,7 +242,8 @@ def fake_data(ctx, category, factory_class, fake_count):
          ckan generate fake-data dataset
          ckan generate fake-data dataset  --title="My test dataset"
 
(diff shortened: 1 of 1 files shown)


ThrawnCA:
Well, that's annoying. Configuring our automated tests to use either `paster create-test-data` or `ckan seed` was simple enough, but replicating the effects with `ckan generate fake-data` is going to be more of a nuisance, firstly because it means digging into the details of exactly what test data was being created, and secondly because it will mean extra "if this version do this, otherwise do something completely different" checks.

PR #6507, opened 2021-10-29
Sqlalchemy: clean deprecated mapper.order_by

### LINKED ISSUE #6243: Refactor model to support SQLAlchemy 1.4
Our current usage of SQLAlchemy prevents us from upgrading to [version 1.4](https://docs.sqlalchemy.org/en/14/changelog/changelog_14.html).

* We use the [`SessionExtension`](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/model/meta.py#L13) and [`MapperExtension`](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/model/extension.py#L27) classes, which are deprecated in 1.3 and removed in 1.4. They need to be replaced by the [event listener system](https://docs.sqlalchemy.org/en/14/core/event.html). Might be a good opportunity to assess if these extensions are actually used.
* The dictization functions use [`RowProxy`](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/lib/dictization/__init__.py#L13) which is not available on 1.4 ([`Row`](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Row) seems to be a potential replacement).
* `mapper()` no longer supports `order_by` (eg [here](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/model/tag.py#L285)) or `extension` (eg [here](https://github.com/ckan/ckan/blob/94194d43667d122bc5a55a45e16b07b4430f3405/ckan/model/package.py#L553))
* A couple of warnings coming up:
```
/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/model/user.py:62: SAWarning: relationship 'Package.package_tag_all' will copy column package.id to column package_tag.package_id, which conflicts with relationship(s): 'Package.package_tags' (copies package.id to package_tag.package_id), 'PackageTag.package' (copies package.id to package_tag.package_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="package,package_tags"' to the 'Package.package_tag_all' relationship.
  query = meta.Session.query(cls).autoflush(False)
/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/model/user.py:62: SAWarning: relationship 'PackageTag.pkg' will copy column package.id to column package_tag.package_id, which conflicts with relationship(s): 'Package.package_tags' (copies package.id to package_tag.package_id), 'PackageTag.package' (copies package.id to package_tag.package_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="package,package_tags"' to the 'PackageTag.pkg' relationship.
  query = meta.Session.query(cls).autoflush(False)
```

```
/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/model/activity.py:130: SAWarning: TypeDecorator JsonDictType() will not produce a cache key because the ``cache_ok`` flag is not set to True.  Set this flag to True if this type object's state is safe to use in a cache key, or False to disable this warning.
  return _activities_limit(q, limit, offset).all()
```

pdelboca:
Fixes #6243 

Part of the upgrade to SQLAlchemy 1.4 is to remove the deprecated `mapper.group_by(...)` method: https://docs.sqlalchemy.org/en/14/changelog/migration_11.html?highlight=mapper%20order_by#mapper-order-by-is-deprecated

This change will remove default order by in some objects queries. 

#### For example:

A current query that returns a list of packages order by name:
```
ipdb> [p.name for p in model.Session.query(model.Package).all()]
['a-package', 'b-package']
```

Will return the values as they appear in the database (Review):
```
ipdb> [p.name for p in model.Session.query(model.Package).all()]
['b-package', 'a-package']
```


jj-crypto (approved):

amercader:
@jj-crypto unless you have actual knowledge of the changes introduced on the pull requests please refrain from submitting more reviews, as it creates noise for the maintainers

pdelboca:
This is now ready for review!

The failing test is deprecated and have been removed in this PR: https://github.com/ckan/ckan/pull/6573 so once it's merged I can update this PR with the changes.

PR #6508, opened 2021-10-30
Upgrade cypress and axe

Zharktas:
### Proposed fixes:
Upgrades Cypress and axe as they weren't previously updated due to incompatibilities.

**Fixes build warnings in docs.**

### Features:

- [ ] includes API changes


jj-crypto (approved):

amercader:
@Zharktas do you know why the `package.lock` file grew so much with these changes?

amercader:
@Zharktas looks like this changed the `"lockfileVersion": 2`, probably because of using one of the latest node versions.
@EricSoroos suggested to perhaps wait on this as it can lead to reproducibility errors (let me know if I didn't get that right). Do you have any thoughts on this?

In any case it looks like if we go for it we should do the minimum Node version / package.lock format upgrade in a separate PR

Zharktas:
@amercader I've upgraded the package.lock version as our github actions upgrade it every time as example here: https://github.com/ckan/ckan/actions/runs/1411142934

image

Should have probably mention it though.

amercader:
@Zharktas fair enough. Does this mean that we need to document a minimum Node version for running the tests locally?

Zharktas:
The new version is however backwards compatible https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile, so it should not matter.

Zharktas:
@amercader I removed the documentation fixes as they where fixed elsewhere, this would be good to merge unless you feel that lock-file version change should be done in separate PR. I can rerun this with older npm if you wish.

amercader:
Let's  merge this as is, if that causes issues we can revisit it. Thanks @Zharktas

PR #6509, opened 2021-10-31
Clean requests params

pdelboca:
This PR replaces all the calls to `requests.params` for `requests.args` and set a deprecation warning in the method.

Additionally, it cleans some old `_is_flask_request()` code.


jj-crypto (approved):

ckan/common.py
```diff
@@ -119,22 +114,17 @@ class CKANRequest(LocalProxy):
...
-
-    LocalProxy will forward to Flask or Pylons own request objects depending
-    on the output of `_get_request` (which essentially calls
-    `is_flask_request`) and at the same time provide all objects methods to be
-    able to interact with them transparently.
     '''
+
+    @maintain.deprecated(since="2.10.0")
     @property
     def params(self):
```
amercader, line 121: Let's keep this for a while to support extensions

PR #6517, opened 2021-11-03
document 'ckan.download_proxy' config value

ThrawnCA:
- This is intended to be used by extensions that download external URLs, such as XLoader, Archiver, and Validation.

See https://github.com/ckan/ckanext-xloader/pull/127

### Features:


doc/maintaining/configuration.rst
```diff
@@ -1284,6 +1284,23 @@ Format as a space-separated list of the plugin names. The plugin name is the key
...
+ckan.download_proxy
+^^^^^^^^^^^^^^^^^^^
+
+Example::
+
+  ckan.download_proxy = http://proxy:3128
+
+Specifies a HTTP proxy to be used by extensions such as XLoader or Archiver
+when downloading remote files. This will not be used by CKAN core.
+
```
amercader, line 1299: Shall we give more details about why this setting might be useful?

```suggestion
This might be useful if you want to restrict extensions from downloading files from unauthorized locations by setting up and configuring an HTTP proxy like `Squid <http://www.squid-cache.org/>`__. 
```

ThrawnCA: Reasonable, but perhaps it would be even better to link directly to a page about how to use Squid for security, eg https://feeding.cloud.geek.nz/posts/restricting-outgoing-webapp-requests-using-squid-proxy/ ?

PR #6519, opened 2021-11-04
Use get_action in patch actions to allow custom logic

amercader:
In most cases, if you are chaining some `*_update` action in your plugin to add some custom logic you also want it applied to the `*_patch` actions. Because the core `*_patch` actions called the core update actions directly, plugins need to explicitly override both the update and the patch actions.

These updates the calls to follow the convention of never calling core actions directly.
I've marked this as good for backporting as existing logic should not be affected.


amercader:
Failures seem unrelated

Zharktas:
Changelog warnings are due to fragment referencing issue number and not pull request number. Fixed here https://github.com/ckan/ckan/pull/6508/commits/cdf89e901b9feb660c00a5ce1920f5d2c7cd38a6 and here https://github.com/ckan/ckan/pull/6508/commits/b428bf82a5ec04ce909a7a594033fe4065f67527

amercader:
TODO: Add changelog entry

amercader:
Added changelog entry and merged latest master @Zharktas

PR #6530, opened 2021-11-11
Switch to Solr v8.0 (with the minimal code changes and no new features)

smotornyuk:
It looks like there will be quite a few parties working on the Solr improvements. At the moment it looks like:

* @amercader is going to move forward search Index&Queries Abstraction layers
* @TkTech did a huge work related to the Lucene query parser and created django-orm-like module in order to interact with search engine 
* @wardi and @smotornyuk are going to work on support for managed schemas

Probably it will be the best option, for now, to start using solr v8.0 as soon as possible. In the future, if it has sense, we can bring back support of older Solr versions via a "custom backends"-like feature.


smotornyuk:
Here's the Dockerfile that was used for CircleCI Solr image
```
FROM solr:8.0
EXPOSE 8983
COPY ./schema.xml ./server/solr/configsets/_default/conf/managed-schema
CMD ["solr-precreate", "ckan"]
```

.github/workflows/cypress.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         ports:
           - 6379:6379
       ckan-solr:
-        image: solr:6.5
+        image: solr:8.0
```
EricSoroos, line 30: Recommend either `solr:8` or `solr:8.10`, as there's no point in pinning to an old minor release.

ckan/lib/search/query.py
```diff
@@ -366,6 +366,8 @@ def run(self, query, permission_labels=None, **kwargs):
             query['mm'] = query.get('mm', '2<-1 5<80%')
             query['qf'] = query.get('qf', QUERY_FIELDS)
 
+        query.setdefault("df", "text")
```
EricSoroos, line 369: I think that if it's possible to do this in solr config, that would be ideal, as it reduces the impact on ckan core.

EricSoroos (commented):
Also, should probably update:
`<schema name="ckan" version="2.9">`

.github/workflows/cypress.yml
smotornyuk: So, `solr:8` will pin it to the latest v8 release, right? I prefer this option in that case

ckan/lib/search/query.py
smotornyuk: Have no objections, but it's up to Adria :)

In order to remove these lines, we need to update `solrconfig.xml` inside the core's `conf` folder and set the following `defaults` for the request handler
```
<str name="df">text</str>
<str name="q.op">AND</str>
```

.github/workflows/cypress.yml
EricSoroos: Yes, currently, this is the (top of) the Solr docker tags:
image

So solr:8 will follow the latest solr 8.x.y release, where solr:8.0 would be solr 8.0.0:
image

ckan/lib/search/query.py
EricSoroos: The key though is that then CKAN can work, unmodified, with either solr backend.  This decouples the CKAN upgrade path from the solr upgrade path, at least for the current set of upgrades.  It also allows backports on the CKAN side without requiring that you update existing infra in a point release.

EricSoroos:
On the call I mentioned the date processing bits -- That can be done with https://solr.apache.org/docs/8_8_0//solr-core/org/apache/solr/update/processor/ParseDateFieldUpdateProcessorFactory.html, in the updateRequestProcessorChain in solrconfig.xml:
```
    <processor class="solr.ParseDateFieldUpdateProcessorFactory">
      <arr name="format">
        <str>yyyy-MM-dd'T'HH:mm:ss.SSSZ</str>
        <str>yyyy-MM-dd'T'HH:mm:ss,SSSZ</str>
        <str>yyyy-MM-dd'T'HH:mm:ss.SSS</str>
        <str>yyyy-MM-dd'T'HH:mm:ss,SSS</str>
        <str>yyyy-MM-dd'T'HH:mm:ssZ</str>
        <str>yyyy-MM-dd'T'HH:mm:ss</str>
        <str>yyyy-MM-dd'T'HH:mmZ</str>
        <str>yyyy-MM-dd'T'HH:mm</str>
        <str>yyyy-MM-dd HH:mm:ss.SSSZ</str>
        <str>yyyy-MM-dd HH:mm:ss,SSSZ</str>
        <str>yyyy-MM-dd HH:mm:ss.SSS</str>
        <str>yyyy-MM-dd HH:mm:ss,SSS</str>
        <str>yyyy-MM-dd HH:mm:ssZ</str>
        <str>yyyy-MM-dd HH:mm:ss</str>
        <str>yyyy-MM-dd HH:mmZ</str>
        <str>yyyy-MM-dd HH:mm</str>
        <str>yyyy-MM-dd</str>
      </arr>
    </processor>
```

I'll also note -- the official solr 8 images are built differently than the solr 7 and below ones, so there's some monkeying with volumes that makes it so that you can't preload cores in your docker images, they have to be done at initial runtime with volumes.

ckan/config/solr/schema.xml
```diff
@@ -24,29 +24,29 @@
 <!-- We update the version when there is a backward-incompatible change to this
 schema. In this case the version should be set to the next CKAN version number.
 (x.y but not x.y.z since it needs to be a float) -->
-<schema name="ckan" version="2.9">
+<schema name="ckan" version="2.10">
```
amercader, line 27: I get this when trying to use this schema:
```
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/config/environment.py", line 176, in update_config
    search.check_solr_schema_version()
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/lib/search/__init__.py", line 318, in check_solr_schema_version
    version = tree.documentElement.getAttribute('version')
ckan.lib.search.common.SearchError: SOLR schema version not supported: 2.1. Supported versions are [2.8, 2.9, 2.10]
```
Did you get that working at all? I'm curious to know how this could work on your end and the tests.

The reason is failing is because Solr must be parsing this value like a float and when you request the schema using its endpoint it returns `<schema name="ckan" version="2.1">`

I know what you are thinking, let's just add a `v` and be done with it:
```
<schema name="ckan" version="v2.10">
```
That doesn't work and makes Solr crash. In fact, it seems like [we shouldn't have used this field at all](https://cwiki.apache.org/confluence/display/solr/schemaxml#SchemaXml-Schemaversionattributeintherootnode). Adding custom fields or attributes to the schema doesn't work because solr strips everything it doesn't know of away when requesting the schema.

I guess the only safe option is to use the `name` attribute, which on the Solr side is just used for displaying purposes, with a convention like `ckan-{version_number}`, eg `<schema name="ckan-2.10" version="1.6">`. And modify our schema version check here:

https://github.com/ckan/ckan/blob/157aa9b5c78da005f0938eb2037b67bd774f5cce/ckan/lib/search/__init__.py#L317

To use the `name` attribute. We can fallback to the `version` attribute if using a 2.8/2.9 schema.

ckan/lib/search/query.py
amercader: I'm not opposed to shifting these changes to the Solr config side (although I have a to say that after going through a manual install of Solr it is not the most friendly process for a beginner user). I think having a clearer picture of the steps involved in installing and setting up Solr and write the install instructions for users will help me decide what we should do, so bear with me.

I guess that if we want to support both versions of Solr now, the alternative is to query the Solr version at startup time (like we do to get the schema version) using [this endpoint](http://localhost:8983/solr/admin/info/system) and conditionally execute these lines but is pretty ugly tbh and probably adds too much complexity for saving users to edit one file.

@EricSoroos can you clarify what you mean with this?
>  It also allows backports on the CKAN side without requiring that you update existing infra in a point release.

If you want to run CKAN 2.9 against Solr 8 you will need to do the changes outlined in this PR, so you'll need to update your existing infra right?

amercader (commented):
@smotornyuk some initial comments, just started playing with this

.circleci/config.yml
```diff
@@ -57,7 +57,7 @@ defaults:
     name: ckan-redis
 
   solr_image: &solr_image
-    image: ckan/ckan-solr-dev:master
+    image: smotornyuk/ckan-solr:latest
```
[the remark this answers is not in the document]
smotornyuk: Oh, i'll try to push it myself then.

Yes, sound safe, let's go with this option

ckan/config/solr/schema.xml
EricSoroos: The comment just above that line points to the issue:  `(x.y but not x.y.z since it needs to be a float) -->`

Let's just go with 3.0? Aren't we on 3.0 at this point?

ckan/lib/search/query.py
EricSoroos: It Allows. It doesn't Require.

* I think that requiring a solr8 upgrade in a point/bugfix release is a non-starter.
* I think that providing the option to upgrade solr outside of a minor/major release cycle is a good idea. 

So. IFF we can manage CKAN running against solr6 or solr8 with no changes on the CKAN side, then it's something that we should prioritize because it simplifies the backport process. Bring up a new solr server, configure it, point ckan at it, and rebuild indexes. If we can't, we can't. (but I don't think that's the case)

FWIW, I'm running solr8.8 in two instances, both of them are dev/testing, but they're working with unmodified ckan 2.9. I'm not going to claim 100% compatibility due to a lack of exhaustive testing, but I haven't noticed anything of an issue other than setting things up in the first place. 

For reference, my standard solr install in 6 is essentially: Pull solr6:, slap in some default config files in the docker image, then add a volume for data. Potentially bind mount in a new schema.xml on particular instances, depending.

For solr 8, because of the upstream docker image layout changes, you can't actually put a set of core config files in the image, you have to add them as a volume later. So I have something like an initdb_solr stage in the makefile, similar to the initdb stages that are required for postgres.  Inlining a config file in addition to the schema here is just one more file to have in the volume that gets added.

ckan/config/solr/schema.xml
Zharktas: 3.0 would have major breaking changes, current master does not have those compared to 2.9 so most likely next release will be 2.10

EricSoroos: Wasn't python2 support removed?

Zharktas: Most of the breaking changes are already in 2.9, but anyway there isn't an official decision what the next version is called, it is 2.10 in the code base for now.

EricSoroos: There's a PR in master removing python 2 support (#5172), which was not backported to 2.9. If that's not a major breaking change, I don't know what is.

Zharktas: But still, the code base says that the next version is 2.10. https://github.com/ckan/ckan/blob/f96337c172c4111e6228f8e12bdea82a80139814/ckan/__init__.py#L3 This PR is not really the place to decide otherwise :p

EricSoroos: Ok, then I guess we don't need to update the schema version, if it's compatible with unmodified ckan.

amercader: @smotornyuk regardless of what we end up using here, I still want to understand why CKAN is running at all in this PR and not failing with `SOLR schema version not supported: 2.1. Supported versions are [2.8, 2.9, 2.10]` :)

smotornyuk: Hm... I think that I've built docker image **before** I've changed the version inside `schema.xml`, thus it still uses `2.9` which is parsed with no problems.

amercader: Yes, that's probably what happened, the Cypress tests failed because they used the updated version. 
I've pushed some changes (9fcac69) to switch to storing the version in the `name` attribute. I think even if CKAN would work without changing the version we should do it as the syntax is Solr 8 only. We can backport it to 2.9.5 to allow 2.9 to run on this version of the schema.

We can have the 2.10 vs 3.0 discussion elsewhere, it should be easy to change the numbers when we settle on a version

amercader:
OK, I've spent long enough on this to be able to propose a path forward that I think on balance has the most benefits.

In summary:
* For installing Solr 8 we will document both a Docker based approach and a basic manual one in case users can't or don't want to use Docker (TODO). Our [ckan/ckan-solr-dev](https://github.com/ckan/ckan-solr-dev) image will be renamed to `ckan/ckan-solr`. I'd appreciate a review on https://github.com/ckan/ckan-solr-dev/pull/2 as Solr 8 really makes difficult to pre-package things so there might be a better way to do it.
* The new Solr 8 schema should have a new version (even if it could work on CKAN 2.9). I really think we need to change they way we are storing the schema version (see for https://github.com/ckan/ckan/pull/6530#discussion_r753097219 context) and now it's the perfect time to do it.
* This means a small backport to 2.9.5 (9fcac69 and `SUPPORTED_SCHEMA_VERSIONS`) but I think it's reasonable to ask and we should encourage regular patch updates
* I'd like to keep these [two small changes](https://github.com/ckan/ckan/pull/6530/files#diff-33eba33435c8e3defcf43d1cedb9ee96a9f3079394db8615b4c7a1c778af2f0dR369) to set up default request params. AFAICT they are backwards compatible across Solr versions and will save users editing the `solrconfig.xml` file, which I think is worth doing to simplify some already complex setup steps. Happy to discuss the implications I've missed though

I believe with this leaves in a place where both CKAN 2.9 and 2.10/3.0 can work against Solr 6 and Solr 8:

| -- | Solr 6 | Solr 8 |
| -- | -- | -- |
| **CKAN 2.9** | :heavy_check_mark:  Works as currently | :heavy_check_mark: Works if upgraded to 2.9.5 |
| **CKAN 2.10** | :heavy_check_mark: Code unchanged except for the two default params, so no burden to support | :heavy_check_mark: Works after this PR is merged |

Keen to hear your thoughts on this. I'll keep working on the installation docs.

amercader:
Apologies I've now realized that I've break all master builds since I last pushed the `ckan/ckan-solr-dev:master` image

I've reverted that tag and created a new one for solr 8:

| Docker image | Solr version |
| --- | --- |
| ckan/ckan-solr-dev:master | Solr 6 |
| ckan/ckan-solr:master | Solr 8 |

amercader:
@ckan/core @kowh-ai  @EricSoroos

This PR and the Docker images at [ckan/ckan-solr](https://github.com/ckan/ckan-solr) are ready for a final review before it can be merged to master and update the docs.

I'd love for people to go through the documentation and give feedback on: 
1. [Docker images documentation](https://github.com/ckan/ckan-solr)
2. The updated Solr install docs, which can be reviewed here: http://docs.ckan.org/en/solr-8.0/maintaining/installing/install-from-source.html#setup-solr

Summary of changes:

* CKAN 2.10 (master) supports Solr 8. CKAN 2.9 (and almost certainly 2.8) will support Solr 8 if running at least 2.9.5 and 2.8.10 (next patch releases)
* There are pre-configured Docker images for all the combinations above (see the [Docker images documentation](https://github.com/ckan/ckan-solr))
* Install docs describe both Docker and manual installs.
* I've updated the default value for `solr_url` from `http://localhost:8983/solr` to `http://localhost:8983/solr/ckan`

doc/conf.py
```diff
@@ -129,8 +127,9 @@
...
 def get_release_tags():
     git_tags = subprocess.check_output(
-        ['git', 'tag', '-l'], stderr=subprocess.STDOUT).split()
-    release_tags_ = [tag for tag in git_tags if tag.startswith(six.b('ckan-'))]
+        ['git', 'tag', '-l'], stderr=subprocess.STDOUT).decode('utf8')
+    git_tags = git_tags.split()
+    release_tags_ = [tag for tag in git_tags if tag.startswith('ckan-')]
 
     # git tag -l prints out the tags in the right order anyway, but don't rely
     # on that, sort them again here for good measure.
```
wardi, line 135: this comment reminds me that the following line will be broken for 2.10

wardi:
I don't know enough about what `query.setdefault("df", "text")` does to argue about it, but the rest of this PR looks good to me.

amercader:
@wardi 

This was a change in Solr 7. Our old schema defined the default search field when querying [in the schema.xml file](https://github.com/ckan/ckan/blob/36839d57d443bd189056bc83cf3e684c72a1f5d5/ckan/config/solr/schema.xml#L178), but this is no longer supported:

> The defaultSearchField parameter in the schema is no longer supported. Use the df parameter instead. This option had been deprecated for several releases. See the section Standard Query Parser Parameters for more information.

(From [the Solr Changelog](https://solr.apache.org/guide/8_1/major-changes-in-solr-7.html#other-deprecations-and-removals)).

We need to explicitly pass a `df` parameter when querying instead.

kowh-ai:
@amercader  - Iv'e been through the Docker images documentation (https://github.com/ckan/ckan-solr) and built the (Pre-configured) Solr Docker images. Looks good. I'm having a problem with building the **multi image** though for Sol 6. The `make build-multi` command has a problem with the `ALL_CKAN_VERSIONS="2.6 2.7 2.8 2.9"` parameter when it tries to build the docker image `invalid argument "ckan/ckan-solr-dev:2.6 2.7 2.8 2.9" for "-t, --tag" flag: invalid reference format` I could be doing something wrong here though

amercader:
Thanks @kowh-ai! I fixed that in https://github.com/ckan/ckan-solr/commit/cada2e87d9833b538c5256f530381ce7d15f3eb8 and push the image. I've now updated all `ckan/ckan-solr:*` images to pull the latest base Solr images in order to include the fix for the log4j 2 issue.

amercader:
Thanks all for you feedback and @smotornyuk for the initial work on this. I'm going to merge now and we can tweak things later

.github/workflows/cypress.yml
```diff
@@ -27,10 +27,9 @@ jobs:
         ports:
           - 6379:6379
       ckan-solr:
-        image: solr:6.5
+        image: ckan/ckan-solr:master
```
florianm, line 30: Should the docker-compose SolR Dockerfile also use `ckan:ckan-solr:master`? 
The [first two lines](https://github.com/ckan/ckan/blob/master/contrib/docker/solr/Dockerfile#L1-L2) would become (with an updated `LABEL`): 
```
FROM ckan/ckan-solr:master
LABEL maintainer="Open Knowledge"
```

amercader: @florianm no, I think we should completely remove the solr Dockerfile in this repo and call the `ckan/ckan-solr` image directly from the docker-compose file:

```diff
diff --git a/contrib/docker/docker-compose.yml b/contrib/docker/docker-compose.yml
index 3fe96ddee..7687f9b46 100644
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -65,9 +65,7 @@ services:
 
   solr:
     container_name: solr
-    build:
-      context: ../../
-      dockerfile: contrib/docker/solr/Dockerfile
+    image: ckan/ckan-solr:2.10
     volumes:
       - solr_data:/opt/solr/server/solr/ckan/data
     restart: unless-stopped
```

Maybe the volume path needs updating, I'm not sure.

Want to submit a PR with these changes?

florianm: Ah, calling `ckan/ckan-solr` from the compose file makes a lot of sense.

I'm just working through the docker-compose approach and will likely have some fresh ideas by Feb next year (xmas break is looming). It appears that the docker-compose approach would benefit from the patch release and all the SolR changes from this PR.

I also feel that mounted folders would provide a smoother workflow for plugin development after all, which means re-working the docker-compose approach.

All of this means there's more work to do before I can submit a comprehensive PR for the docker-compose install, and I'll be slow about it too (Feb '22).

Did you want me to submit a quick PR for just the changes above for now? (Not sure about the volume path, strapped for time to figure out the correct path)

PR #6533, opened 2021-11-11
Add resource_revision index

jbrown-xentity:
Not related to a current bug, found in the wild when using CKAN on US data.gov system.

### Proposed fixes:
When updating a package, the CKAN system searches the current resources from [here](https://github.com/ckan/ckan/blob/2.8/ckan/lib/dictization/model_dictize.py#L137).
When you have many records, this query can take multiple seconds to run which
slows down the update considerably. This index increased
speed for the given query by > 90%, and helped cut time for package updates.

This goes in line with [resource index](https://github.com/ckan/ckan/blob/master/ckan/migration/versions/092_01afcadbd8c0_resource_package_id_index.py), which we also had to implement.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/migration/versions/101_7ab0d120f411_add_resource_indexes.py
+++ b/ckan/migration/versions/101_7ab0d120f411_add_resource_indexes.py
@@ -1,0 +1,28 @@
+"""add resource indexes
+
+Revision ID: 7ab0d120f411
+Revises: ccd38ad5fced
+Create Date: 2021-11-11 14:12:09.088843
+
+"""
+from alembic import op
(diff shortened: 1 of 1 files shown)


wardi:
If we're still using the revision tables that's a bug. We stopped using vdm (the code responsible for managing revision tables) in 2.9

jbrown-xentity:
My apologies, this was discovered on a CKAN2.8 build and I didn't see/find any documentation on that change for 2.9. If there is any, could you point me to it?
Closing this as it's not valuable for other users...

PR #6535, opened 2021-11-12
Support file attachments in emails, refactor email module

amercader:
Allow `mail_user()` or `mail_recipient()` to support a new `attachments` parameter. This should be a list of tuples in the form:

```
    [
        (file_name, file_object, media_type)
    ]
```

While working on this and to make the changes easier I refactored the email sending function to use the `email.message` module instead of the legacy `email.mime`, which simplifies significantly the code.

Note that the existing base64 encoding has been kept on, although in the future we might consider adding an option to
support the default "quoted-printable" one which is more human friendly.

### DIFF
--- a/changes/6535.misc
+++ b/changes/6535.misc
@@ -1,0 +1,1 @@
+Support including file attachments when sending emails
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -5,6 +5,7 @@ import os
 import smtplib
 import socket
 import logging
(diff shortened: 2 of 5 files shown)


amercader:
Current failure is unrelated: https://github.com/ckan/ckan/issues/6537

smotornyuk:
Perfect, that's exactly what I need on my project :fire:

jj-crypto (approved):

PR #6538, opened 2021-11-12
Simplify toolkit / do not use tk in CKAN core

smotornyuk:
1. Stop using the toolkit inside CKAN code. As the toolkit's documentation says, it developed only for extensions
2. After completing the previous item it's possible to simplify the toolkit and stop using `sys.modules["toolkit"] = _Toolkit()` hack. The simpler, the better, right?

### DIFF
--- a/ckan/plugins/__init__.py
+++ b/ckan/plugins/__init__.py
@@ -1,6 +1,21 @@
 # encoding: utf-8
+import warnings
 
 from ckan.plugins.core import *  # noqa: re-export
 from ckan.plugins.interfaces import *  # noqa: re-export
 
-from ckan.plugins import toolkit  # noqa: re-export
+
+def __getattr__(name):
+    if name == 'toolkit':
+        import ckan.plugins.toolkit as t
+        from ckan.exceptions import CkanDeprecationWarning
+        msg = ("`toolkit` attribute of the `ckan.plugins` module will be"
+               "removed in future CKAN release. Import `ckan.plugins.toolkit`"
+               " directly instead:\n\t"
+               "import ckan.plugins.toolkit as toolkit")
+        warnings.warn(
+            msg, CkanDeprecationWarning, stacklevel=2
+        )
+
+        return t
+    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
--- a/ckan/plugins/__init__.py
+++ b/ckan/plugins/__init__.py
@@ -1,5 +1,4 @@
 # encoding: utf-8
-import warnings
 
 from ckan.plugins.core import *  # noqa: re-export
 from ckan.plugins.interfaces import *  # noqa: re-export
@@ -8,14 +7,6 @@ from ckan.plugins.interfaces import *  # noqa: re-export
 def __getattr__(name):
     if name == 'toolkit':
         import ckan.plugins.toolkit as t
-        from ckan.exceptions import CkanDeprecationWarning
-        msg = ("`toolkit` attribute of the `ckan.plugins` module will be"
-               "removed in future CKAN release. Import `ckan.plugins.toolkit`"
-               " directly instead:\n\t"
-               "import ckan.plugins.toolkit as toolkit")
-        warnings.warn(
-            msg, CkanDeprecationWarning, stacklevel=2
-        )
-
         return t
+
     raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
--- a/ckan/plugins/__init__.py
+++ b/ckan/plugins/__init__.py
@@ -2,3 +2,5 @@
 
 from ckan.plugins.core import *  # noqa: re-export
 from ckan.plugins.interfaces import *  # noqa: re-export
+
+from ckan.plugins import toolkit  # noqa: re-export
--- a/ckan/plugins/__init__.py
+++ b/ckan/plugins/__init__.py
@@ -2,5 +2,3 @@
 
 from ckan.plugins.core import *  # noqa: re-export
 from ckan.plugins.interfaces import *  # noqa: re-export
-
-from ckan.plugins import toolkit  # noqa: re-export
--- a/ckan/plugins/toolkit.py
+++ b/ckan/plugins/toolkit.py
@@ -10,9 +10,9 @@ Functions/objects should only be removed after reasonable deprecation notice
 has been given.
 
 """
-import ckan  # noqa: re-export
-import ckan.lib.base as base  # noqa: re-export
-from ckan.lib.base import render, abort  # noqa: re-export
+import ckan
+import ckan.lib.base as base
+from ckan.lib.base import render, abort
 
 from ckan.logic import (  # noqa: re-export
     get_action,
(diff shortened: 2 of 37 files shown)


ckan/tests/test_coding_standards.py
```diff
@@ -243,38 +243,20 @@ def test_building_the_docs():
...
-    new_warnings = [
-        warning for warning in warnings if warning not in warnings_to_remove
-    ]
-
-    if new_warnings:
+    warnings = [line for line in output_lines if "WARNING" in line]
+    if warnings:
         assert False, (
-            u"Don't add any new warnings to the Sphinx build: "
-            u"{warnings}".format(warnings=new_warnings)
```
wardi, line 277: seems like an unrelated but worthwhile change

ckan/plugins/__init__.py
```diff
@@ -2,5 +2,3 @@
 
 from ckan.plugins.core import *  # noqa: re-export
 from ckan.plugins.interfaces import *  # noqa: re-export
-
-from ckan.plugins import toolkit  # noqa: re-export
```
wardi, line 6: doesn't this make
```python
import ckan.plugins
ckan.plugins.toolkit
```
stop working?

ckan/plugins/toolkit.py
```diff
@@ -1,189 +1,303 @@
 # encoding: utf-8
+"""This class is intended to make functions/objects consistently available to
+plugins, whilst giving core CKAN developers the ability move code around or
+change underlying frameworks etc. This object allows us to avoid circular
+imports while making functions/objects available to plugins.
```
wardi, line 5:
```suggestion
"""This module is intended to make functions/objects consistently available to
plugins, whilst giving core CKAN developers the ability move code around or
change underlying frameworks etc. 
```

ckan/plugins/toolkit.py
```diff
@@ -1,189 +1,303 @@
...
+    """
+    import inspect
+    import os
+    from ckan.lib.webassets_tools import create_library
+
+    # we want the filename that of the function caller but they
+    # will have used one of the available helper functions
+    # TODO: starting from python 3.5, `inspect.stack` returns list
+    # of named tuples `FrameInfo`. Don't forget to remove
+    # `getframeinfo` wrapper after migration.
```
wardi, line 178:
```suggestion
```

ckan/plugins/__init__.py
smotornyuk: Yes, you are right. Taking it back

smotornyuk: Oops, that's the problem. This particular line creates a lot of circular references. 
We are using `ckan.plugins` a lot(particularly because of interfaces). Which means that `ckan.logic` loads `IAction` and gets `toolkit` loaded. That prevents  `toolkit` from using `ckan.logic.get_action`

smotornyuk: Do we really have to keep this object inside `ckan.plugins`? I mean, we have `ckan.plugins.toolkit` as a variable inside `ckan.pluins` and `ckan.plugins.toolkit` as a module.. maybe it's a time to get rid of such duplication? And, even though this is a breaking change, the fix for it is quite straightforward and backward compatible. 

Instead of 
```
import ckan.plugins as p
toolkit = p.toolkit
```

developers have to write
```
import ckan.plugins as p
import ckan.plugins.toolkit as toolkit
```

smotornyuk: There is another option, but it will be a hack. Smaller than previous implementation of toolkit, but still a hack. I think something like this will solve the problem:

```
# ckan/plugins/__init__.py

from ckan.plugins.core import *  # noqa: re-export
from ckan.plugins.interfaces import *  # noqa: re-export

def __getattr__(self, name):
    if name == "toolkit":
        from . import toolkit
        return toolkit
    super().__getattr__(name)
```

But my original intention was to get rid of magic, not to replace it with smaller amount of magic:)

smotornyuk: PS [circular import can be fixed by importing `ckan.cli.cli` inside `ckan.__init__.py`](https://github.com/ckan/ckan/pull/6538/files#diff-eacc8c68ea0e9371c3294a7133871cd90f9248595c523016d27a39ec848ae6e7R23). And that's what I've done at the moment. But this is not **right** solution neither :)

smotornyuk:
Fixed

doc/conf.py
```diff
@@ -284,6 +284,7 @@ def get_min_setuptools_version():
 
 
 def config_defaults_from_declaration():
+    import ckan.cli.cli  # noqa: solves circular imports introduced by toolkit
```
wardi, line 287: that's. wow. ok, really?

smotornyuk: YapXD 
When importing toolkit inside `ckan.plugins` we are breaking the rule "do not use the toolkit in CKAN core". For example, `ckan.logic` uses `ckan.plugins.IActions` interface. This means the toolkit is imported as a side-effect when `ckan.logic` is loaded. But toolkit uses `ckan.logic.get_action` and we have a circular dependency.


Import of `ckan.cli.cli` fixes it it the following way:

* load cli
* cli loads config
* config loads `ckan.plugins`
* `ckan.plugins` loads interfaces first(they are available at this moment) and then loads toolkit
* toolkit loads `ckan.logic`
* `ckan.logic` uses interfaces that are already available, so `ckan.plugins` is not actually touched and we don't have a circular dependency


As an alternative, we can recommend stop using `toolkit` as object from `ckan.plugins` and import it exlicitely instead. But it's a breaking change. We can in addition add a patch like bellow to `ckan.plugins`, that solve circular dependency. But this will be a hack, anyway.

```
def __getattr__(name):
    if name == 'toolkit':
        import ckan.plugins.toolkit as t
        return t
     raise AttributeError(<message>)
```

smotornyuk: It works when you run `ckan run` CLI because cli loads `ckan.plugins` first and issues is solved. But it just came to my mind that using `uWSGI` will likely have the same issues...

So, what do you think, which one is better:
* stop using `import ckan.plugins as p; p.toolkit` (maybe too extreme)
* Use `__getattr__` hack with an extra deprecation notice, suggesting to import toolkit directly
* solve the issue in a similar manner for uWSGI script - just organize imports, so that `ckan.plugins` loads before `ckan.logic`
* close this PR

wardi: I feel like the `__getattr__` hack for for ckan.plugins might be the least disruptive (and is still a lot smaller than the whole ckan.plugins.toolkit using it)

jj-crypto (approved):

smotornyuk:
Updated

ckan/plugins/__init__.py
```diff
@@ -1,6 +1,21 @@
...
+
+def __getattr__(name):
+    if name == 'toolkit':
+        import ckan.plugins.toolkit as t
+        from ckan.exceptions import CkanDeprecationWarning
+        msg = ("`toolkit` attribute of the `ckan.plugins` module will be"
+               "removed in future CKAN release. Import `ckan.plugins.toolkit`"
+               " directly instead:\n\t"
+               "import ckan.plugins.toolkit as toolkit")
```
wardi, line 15: I don't think this code is hurting anything. do we need to remove it in the future?

smotornyuk: Nope. I'm just feeling guilty for adding a new hack into core :laughing:  I'll drop this warning

PR #6544, opened 2021-11-16
Delete a resource in dataset from the `Explore` drop-down button

### LINKED ISSUE #5159: Add ability to delete a resource view from the resource view list page
**Note:** Are you submitting a security related issue that could be a potential vulnerability? Please send it to security@ckan.org instead.

ALL

We should provide additional way to the users so they can delete the resource view. Maybe add a delete button on the resource_views_list page (specifically on resource_view_list_item) and add a delete button, so the user is able to delete the view, without having to go to the view form page.
It would be nice if we can also somehow limit the creation of incorrect views, that cause a Server Error, during resource creation.

Sometimes users create wrong views for the resource, and CKAN doesn't let them know during creation that something is wrong with the view that they are creating. Instead CKAN creates the view for the resource and gives a Server Error for the view.
If the user wants to delete the view, that is not possible since deleting is only available in the view edit form, which can't be opened since there is a Server Error.
The only option that the users have is to go directly to the database and delete the view. 

Create a dataset
Add a resource to the dataset
Add a view to the resource (something that you know is not correct, and it would cause a Server Error)
Try deleting the view

Gauravp-NEC:
Sometimes users create wrong views for the resource, and  CKAN creates the view for the resource and gives a Server Error for the view. If the user wants to delete the view, that is not possible since deleting is only available in the view edit form, which can't be opened since there is a Server Error. This PR will allow a user to delete a resource from the `Explore` drop-down button.

Fixes #5159 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/package/snippets/resource_item.html
+++ b/ckan/templates/package/snippets/resource_item.html
@@ -71,6 +71,16 @@
         </a>
       </li>
       {% endif %}
+       {% if res.id %}
+        {% if h.check_access('resource_delete', {'id': res.id})  %}
+        <li>
+	       <a href="{{ h.url_for(pkg.type ~ '_resource.delete', id=pkg.name, resource_id=res.id) }}" data-module="confirm-action" data-module-content="{{ _('Are you sure you want to delete this resource?') }}" >
+	         <i class="fa fa-trash-o"></i>
+            {{ _('Delete') }}
+	       </a>
+        </li>  
+       {% endif %}
+      {% endif %}
       {% endblock %}
     </ul>
   </div>


smotornyuk:
Can you explain why the initial issue and your first comment addressed errors with the resource **preview**, but in the end, you've implemented a delete button for the resource itself? :laughing: 

I don't think that we need this button. Datasets/organizations/groups/anything has a delete button on the edit form. What is the value of simplifying resource removal?

Gauravp-NEC:
@smotornyuk , User can directly delete a resource from the Explore drop-down button instead of going to the edit form page. Edit does not reflect to delete, the main purpose of edit form is to update a resource. I will suggest that we should remove delete button from the edit form page. I will also suggest that we should keep a cancel/back option button in place of delete button in edit form page so that user can go back to previous page if no update done by user. If dev team agrees on my suggestion I will raise a separate PR to implement it.

amercader:
@Gauravp-NEC , @smotornyuk is Tech Team member and his input on reviews is final, no need to ping other developers about it.

He clearly mentioned that the original issue was about Resource Views, not Resources, and that the new delete option on the dropdown resource does not make sense because it doesn't align with the rest of the forms behaviour. This makes total sense so not sure what else is there to add. (You can of course implement it in your own site)

jj-crypto (approved):

Gauravp-NEC:
@JGulic, What's your opinion on this PR?

smotornyuk:
jj-crypto has nothing to do with CKAN core, my opinion is the same as it was before

amercader:
@Gauravp-NEC see my previous comment https://github.com/ckan/ckan/pull/6544#issuecomment-972907757

PR #6547, opened 2021-11-17
Clean old deprecated License methods

pdelboca:
This attributes and methods were deprecated in CKAN 2.4.0 so it should be safe to clean the code now.


jj-crypto (approved):

Gauravp-NEC:
@pdelboca , with reference to issue https://github.com/ckan/ckan/issues/7447.

I think we need to re-adding the function or is there a better solution? Please let me know your opinion.

PR #6556, opened 2021-11-24
Add RootPathMiddleware to flask stack to support non-root installs

### LINKED ISSUE #6555: Non-root installs broken with python3
**CKAN version** 
2.9.3

**Describe the bug**
Root path is added twice to any generated URLs on the site. For example `/data/data/dataset`

**Steps to reproduce**
Install CKAN 2.9.3 into root_path `/data`, see `data-site-root` body tag attribute when loading `/data/dataset`

**Expected behavior**
The URLs are generated correctly

**Additional details**
This is because the Python 3 app stack does not contain any step to remove the `SCRIPT_NAME` environment variable set by the WSGI host. For Pylons this is done in RootPathMiddleware and in Python 2 installs `flask_app` does it in `can_handle_request`.

The issue can be fixed by adding a RootPathMiddleware inclusion step to `make_flask_stack`. I'll make a PR later today.

bzar:
Fixes #6555 

### Proposed fixes:
- Add RootPathMiddleware to flask application stack to clear `SCRIPT_NAME`

### Features:

- [ ] includes API changes

The Service is provided by the Digital and Population Data Services Agency (https://dvv.fi/en/).

### DIFF
--- a/ckan/config/middleware/common_middleware.py
+++ b/ckan/config/middleware/common_middleware.py
@@ -11,6 +11,24 @@ import sqlalchemy as sa
 from ckan.common import config
 
 
+class RootPathMiddleware(object):
+    '''
+    Prevents the SCRIPT_NAME server variable conflicting with the ckan.root_url
+    config. The routes package uses the SCRIPT_NAME variable and appends to the
+    path and ckan addes the root url causing a duplication of the root path.
+    This is a middleware to ensure that even redirects use this logic.
+    '''
+    def __init__(self, app, config):
+        self.app = app
+
+    def __call__(self, environ, start_response):
+        # Prevents the variable interfering with the root_path logic
+        if 'SCRIPT_NAME' in environ:
+            environ['SCRIPT_NAME'] = ''
+
+        return self.app(environ, start_response)
+
+
 class TrackingMiddleware(object):
 
     def __init__(self, app, config):
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -32,7 +32,8 @@ from ckan.lib import uploader
 from ckan.lib import i18n
 from ckan.common import config, g, request, ungettext
 from ckan.config.middleware.common_middleware import (TrackingMiddleware,
-                                                      HostHeaderMiddleware)
+                                                      HostHeaderMiddleware,
+                                                      RootPathMiddleware)
 import ckan.lib.app_globals as app_globals
 import ckan.lib.plugins as lib_plugins
 from ckan.lib.webassets_tools import get_webassets_path
@@ -205,6 +206,7 @@ def make_flask_stack(conf):
         session_opts['session.data_dir'] = '{data_dir}/sessions'.format(
             data_dir=cache_dir)
 
+    app.wsgi_app = RootPathMiddleware(app.wsgi_app, session_opts)
     app.wsgi_app = SessionMiddleware(app.wsgi_app, session_opts)
     app.session_interface = BeakerSessionInterface()
 


bzar:
Seems RootPathMiddleware has been removed in https://github.com/ckan/ckan/commit/7be274d5faab4d07e9c3f908aa8446df0c1ab7a0#diff-74422062ee7347e621dc8693c28a9e661e81f043ffb6599bcd04a0591b6be70b

Should I create a new middleware for this?

smotornyuk:
* [there is a line](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L431-L432) in code that uses `SCRIPT_NAME` and with your PR it becomes senseless. 
* [there is a code that does it](https://github.com/ckan/ckan/blob/master/ckan/config/middleware/flask_app.py#L450-L453) but apparently it's not called anymore
* When using Nginx+uWSGI(recommended by docs) I cannot reproduce the problem. Can you share more details about your environment? If you are using non-standard setup, can it be fixed via host configuration?

bzar:
The problem can't be fixed there as far as I can tell, because it needs to be done before the `werkzeug.routing.MapAdapter` used for generating the URL is created. That object stores the `SCRIPT_NAME` currently in the environment and in the case of non-root installs then produces URLs where the root path is added twice.

That code is not called when Python 3 is used. [It was called by AskAppDispatcherMiddleware](https://github.com/ckan/ckan/blob/ckan-2.9.3/ckan/config/middleware/__init__.py#L59-L69) when pylons was around to figure out whether a call should handled by Pylons or Flask and had a side effect of emptying `SCRIPT_NAME`. In the case of Python 3 that middleware is not used.

bzar:
For the environment you need a CKAN installed in a non-root path with uWSGI, configured like in [this PR](https://github.com/ckan/ckan/pull/5718) to make it work at all.

quinoescobar:
Applying this patch my CKAN node acts weird when an extension has a redirect implemented with the BluePrint Interface.
e.g.:
This URL https:www.my-ckan-app/opendata/ must be redirect to https:www.my-ckan-app/opendata/dataset, but instead it's being redirect to the server name https:my-server-name.net/dataset

quinoescobar:
But the way, this patch also solves the problem i've been struggling too.
If you make a consecutive API calls to upgrade/update a resource data the download paths missed the root_path configured, but this patch solves this problem

This issue was solve using this patch, CKAN 2.9.3 and XLoader version 0.9.0

smotornyuk:
@bzar, perfect, thanks. 

@quinoescobar, any additional details? I've applied this patch to my CKAN-master installation, configured uWSGI/Nginx, and was not able to reproduce an error. Here is my configuration on Nginx 

```
# ...
location /data/ {
     proxy_pass http://127.0.0.1:8080/;
}
# everything else is standard, so I've committed the rest of config
```

uWSGI:

```
[uwsgi]

http                =  127.0.0.1:8080
uid                 =  sergey
guid                =  sergey
wsgi-file           =  /home/sergey/Projects/core/wsgi.py
virtualenv          =  /home/sergey/.virtualenvs/core
chdir               =  /home/sergey/Projects/core
mount               =  /data=wsgi:application
manage-script-name  =  true
master              =  true
pidfile             =  /tmp/%n.pid
harakiri            =  50
max-requests        =  5000
vacuum              =  true
callable            =  application
buffer-size         =  32768
```

CKAN URL related section:

```
ckan.root_path = /data/{{LANG}}
ckan.site_url = http://localhost/
```

and code that creates a redirect on `/data/xxx` URL:

```
stats = Blueprint(u'stats', __name__)

@stats.route("/xxx")
def redirect():
    return h.redirect_to("dataset.search")
```

Combination of config and code above just works as expected, without any weird duplicated root paths.

PR #6560, opened 2021-11-25
Add Flask-Login into CKAN

TomeCirun:
## Fixes: 
#6714 

### Proposed fixes:

I am integrating Flask-login into CKAN, and I have to mention that this is not the final code, so I will put PR on draft so we can discuss further changes, if you have an idea how to improve the code, do not hesitate to say.

### Features:

- [ ] includes API changes

“This PR is Financed by Salsa Digital [www.salsadigital.com.au](http://www.salsadigital.com.au/). Find all Services at [www.salsadigital.com.au/services](http://www.salsadigital.com.au/services)”


amercader:
This looks promising! Repoze.who is certainly a bit of a black box so anything that simplifies it is worth exploring. 
We should definitely take backwards compatibility into account though. I wonder if the order in which Flask-login is called would allow us to set `REMOTE_USER` in the wsgi environment as repoze.who does. That would perhaps help ckanext-saml2auth, ckanext-oauth, etc transition. Or maybe not, but perhaps worth exploring

TomeCirun:
hey, @amercader thanks for your reply...
I was thinking... can we just set the `REMOTE_USER` from `session["_user_id"]`? If there is `user_id` in the session, that means the user is logged in... so we can get the `user_id` from the `session`.. convert it to `user.username` and set the `REMOTE_USER` in the `environ`? Is this possible?

TomeCirun:
Hey @amercader can you check my last commit: `add remote_user()`... I m not sure if I m doing the right things... I m just experimenting ... this way I successfully set the `REMOTE_USER` and if you approve those changes I m going to revert back the `tests` as they were before with `remote_user`... and I will also make a PR in `saml2auth` with a few changes

TomeCirun:
@smotornyuk can you help with the tests ? 

Thanks!

TomeCirun:
Hey @pdelboca 
Can we start with reviewing this PR and see what else can be done? 
I thought that you might help since you are more referred to `Flask` framework and since you want to get `CKAN` closer to `Flask`.

Thank you

TomeCirun:
Hey @pdelboca 
Can we start with reviewing this PR and see what else can be done? 
I thought that you might help since you are more referred to `Flask` framework and since you want to get `CKAN` closer to `Flask`.

Thank you

smotornyuk:
I see that you are [using UsernamePasswordAuthenticator](https://github.com/ckan/ckan/pull/6560/files#diff-c75d87cc168a832b07afb50d27c4af6dbb7cacc4508040af3f06a30ecf5e739fR512) that was used by **repoze.who**. 
Previously, using `who.ini` it was possible to set up different authenticators using the `who.ini` config. Now the original class is hardcoded. So, what will be my strategy if I need to log in user via email and password? How I can achieve it in my extension?

TomeCirun:
> I see that you are [using UsernamePasswordAuthenticator](https://github.com/ckan/ckan/pull/6560/files#diff-c75d87cc168a832b07afb50d27c4af6dbb7cacc4508040af3f06a30ecf5e739fR512) that was used by **repoze.who**. Previously, it was possible to set up different authenticators using the `who.ini` config. Now the original class is hardcoded. So, what will be my strategy if I need to log in user via email and password? How I can achieve it in my extension?
> 
> Just remove unused code, don't comment it. It will be available in the git history anyway

Thanks for your note i will fix it.

TomeCirun:
@smotornyuk  , is this what you meant? see (`enable custom authentication`) commit

smotornyuk:
Yes, something like this

TomeCirun:
Hey @amercader @pdelboca , just to note, extensions like `saml2auth` will brake with these changes as they use `set_repoze_user` to log the user in.

https://github.com/keitaroinc/ckanext-saml2auth/blob/7412ff7aba3d215f95a08f99216410e72e60c5bc/ckanext/saml2auth/views/saml2auth.py#L267

So i was thinking to bring back that function and inside to do some logic for backward compatibility, though, I m not quite sure what logic should i implement inside `set_repoze_user`.
I believe we are receiving the `user.name` together with the `resp (redirect URL)`.

One of options is to get the `user_obj` from that name and use the flask's [login_user](https://flask-login.readthedocs.io/en/latest/_modules/flask_login/utils/#login_user)  but then the code will be redundant as they do the same logic before logging the user in.
Another option that comes through my mind is to set the g.user = user.name that we receiving, we need to test this as i m not sure if it will work.

What's your opinion on this?

Thanks

ckan/config/middleware/flask_app.py
```diff
@@ -340,6 +304,16 @@ def get_locale() -> str:
...
 
 
+def remote_user() -> None:
+    header = "REMOTE_USER"
+
+    if "_user_id" in request.environ.get("beaker.session", ""):
+        user_id = request.environ["beaker.session"]["_user_id"]
+        userobj = model.Session.query(model.User).get(user_id)
+        if userobj:
+            request.environ[header] = userobj.name
```
pdelboca, line 314: This function is for backwards compatibility. Can we add some docstring to document:
- Why it is needed here
- If we can remove in the future and conditions for it.

ckan/lib/authenticator.py
```diff
@@ -37,3 +32,13 @@ def authenticate(
...
+
+
+class CkanAuthenticator(object):
+
+    @classmethod
+    def __call__(cls):
+        for item in plugins.PluginImplementations(plugins.IAuthenticator):
+            return item
+
+        return UsernamePasswordAuthenticator()
```
pdelboca, line 44: Can we add some documentation of this class and why it is needed?

Also, this implementation will only accept one plugin in our stack implementing `plugins.IAuthenticator`. Is that the expected behavior?

ckan/views/__init__.py
```diff
@@ -156,35 +133,29 @@ def identify_user() -> Optional[Response]:
...
-
-            ev = request.environ
-            if u'repoze.who.plugins' in ev:
-                pth = getattr(ev[u'repoze.who.plugins'][u'friendlyform'],
-                              u'logout_handler_path')
-                redirect(pth)
+            from flask_login import logout_user
+            logout_user()
+            g.userobj = None
+            g.user = None
```
pdelboca, line 158: Is this piece of code still necessary? It looks like we are trying to fit a `repoze` logic to `flask-login`.

It says: 
```
#...But flask-login doesn't check the user (still) exists
# in our database - we need to do that here.
```
But we have the `@login_manager.user_loader` that specifically looks for the user in the Database.

ckan/views/user.py
```diff
@@ -493,58 +470,67 @@ def get(self,
...
-def logged_in() -> Union[Response, str]:
-    # redirect if needed
-    came_from = request.args.get(u'came_from', u'')
-    if h.url_is_local(came_from):
-        return h.redirect_to(str(came_from))
+        auth = authenticator.CkanAuthenticator()
+        if auth().authenticate(identity):
+            if _remember:
+                import datetime
+                dur_time = datetime.timedelta(days=5)
```
pdelboca, line 501: Why `days=5`? `login_form.html` passes a value of `63072000`.

pdelboca:
@TomeCirun ! Great work you're doing here! Thanks!

I left some comments, let me know if they make sense :)

ckan/views/dataset.py
```diff
@@ -944,7 +944,7 @@ def follow(package_type: str, id: str) -> Response:
     return h.redirect_to(u'{}.read'.format(package_type), id=id)
 
 
-def unfollow(package_type: str, id: str) -> Response:
+def unfollow(package_type: str, id: str) -> Union[Response, str]:
```
pdelboca, line 947: In which cases we are returning `str`?

ckan/config/middleware/flask_app.py
TomeCirun: Thanks for your notes @pdelboca  i really appreciate it!

First i will try to explain why at first i thought we needed function like this.

We needed it to set the `g.user` to the current logged user which we are getting from our `beaker_session['_user_id']`  not because we only need it for backward compatibility for the extensions like `saml2auth`, but everywhere in `CKAN` we are dependant from `g.user/g.userobj`. 

Even if we log in the user using `login_user` method from flask, our `CKAN` is not recognizing it if i comment out this line of code: 
https://github.com/ckan/ckan/blob/71a82c4b0bb499fd3a6d1ccfd038b2231f50f92a/ckan/views/__init__.py#L170
because as i said we are dependant from `g.user` and `g.userobj` everywhere in our app, thats why i made this function in the first place.

But  now as I write this, i think i come up with a better solution: 

To make this to be more in a `Flask` way, we should use the `current_user` proxy, and instead of `REMOTE_USER`, `current_user` will set the g.user and g.userobj and with this, we will remove our dependancy from `REMOTE_USER`

We could remove our dependency even from g.user/g.userobj (we can access the logged-in user with the [current_user](https://flask-login.readthedocs.io/en/latest/#flask_login.current_user) proxy, which is available in every template:)

e.g
```python
{% if current_user.is_authenticated %}
  {{ current_user.name }}!
{% endif %}
```


Thanks

ckan/views/user.py
TomeCirun: thanks for you suggestion, it's already changed

ckan/views/dataset.py
TomeCirun: I m not sure, i think it was a fix for `Check types / typecheck (pull_request) ` tests, can you help with this one?

ckan/views/__init__.py
TomeCirun: @pdelboca 
I think we don't need this code anymore, we don't need to invent a way to pass the current logged user, as we do here with  `g.user/g.userobj` as we already have it for free in `current_user` proxy from `flask_login`, and I think we should start using it. 

I started to make these changes step by step.

Please, if you have more suggestions, let me know, or feel free to make a commit :)

ckan/lib/authenticator.py
TomeCirun: I will document it for sure, first, I needed to be sure if I m doing the right thing here.


> Also, this implementation will only accept one plugin in our stack implementing `plugins.IAuthenticator`. Is that the expected behavior?

The expected behavior as @smotornyuk suggested, is that you can set up a custom authenticator from your extensions.

ckan/views/__init__.py
pdelboca: @TomeCirun we still need `g.user/g.userobj` as is widely used in extensions too. But I think this logic can be further clean. 

Some suggestions:
 - What is the goal of `_identify_user_default()`? Looks like is not identifying the user but rather setting the `g.user/g.userobj` attributes. In this case, I would rename to be more explicit on the function's goal.
 - Since this is CKAN core, I would try to set `g.user/g.userobj` using Flask functions. (Do not rely on `REMOTE_USER` anymore.)

Note:
- REMOTE_USER is also widely used in test, so we may need to mantain it here. However, If needed for tests I would try to either add a specific documented line of code or move the logic to the tests fixtures.

I would try to add an example commit later this week. :)

TomeCirun:
@amercader @pdelboca @tino097 @smotornyuk 

I removed `g.user/g.userobj` from CKAN core, i left them to live only on one place (inside `identify_user()`) for backward compatibility for the extensions.

I also removed `_identify_user_default()` as the function main goal was to set the g.user/g.userobj (i implemented this logic on `identify_user()`), and  for `_get_user_for_apitoken()` I made a `flask_login` callback function `load_user_from_request` that will set the `current_user` behind the scene in the same way `load_user` does. 

image

So we are no more dependant of `REMOTE_USER` and `g.user/g.userojb` in CKAN core. 

Note: from now on, when writing new code and we want to check for the user, we are no longer doing: 
```python 
if current_user
``` 
because this will always be true, when there is no user logged: `current_user == Anonymous`, so instead: 
```python
if not current_user.is_anonymous
``` 
remember, `model.User` has a property `is_anonymous`, which always gonna return False,  opposite of that, not logged user would be an instance of `AnonymousUserMixin` which has a property `is_anonymous` set to True.
image

TomeCirun:
I also update the tests according to the new changes in the code.

Note: when you come to code in the tests that have a decorator: `@mock.patch("flask_login.utils._get_user")` know that I m mocking the current_user because some tests like this: https://github.com/ckan/ckan/blob/0e81cd0e068b711b39ff62d40f1cfe2f143c88c5/ckan/tests/controllers/test_package.py#L183 are calling blueprints which contains redirects inside it and when such redirect happens current_user is resetting to Anonymous and does not pass the authentication.

Thanks

ckan/tests/controllers/test_group.py
```diff
@@ -60,63 +59,77 @@ def test_invalid_sort_param_does_not_crash(self, app):
...
+    return data
+
+
+@pytest.fixture
+def user():
+    user = factories.User(password="correct123")
+    user_token = factories.APIToken(user=user["name"])
+    env = {"Authorization": user_token["token"]}
+    data = {"env": env, "user_dict": user}
+    return data
```
pdelboca, line 77: I like the idea of the fixture but I'm not quite convinced on returning the `env` and the `user_dict` since it's a little bit confusing: Why a user dict has an `env` attribute?

I would refactor it to be:
```python
def user():
    user = factories.User(password="correct123")
    user["token"] = factories.APIToken(user=user["name"])
    return user
```

And later in the tests:
```python
    def test_name_required(self, app, user):
        env = {"Authorization": user["token"]}
        url = url_for("group.new")
        response = app.post(url=url, extra_environ=env, data={"save": ""})

        assert "Name: Missing value" in response
```


I think it's:
- more legible, 
- the fixture is isolated to be a user attribute
- We have the same amount of lines

pdelboca (commented):
Thanks @TomeCirun ! This is a massive work, so I'm checking it with detail. I left some comments.

Overall I like the new semantic, I found `if current_user.is_anonymous` way more explicit and understandable than ` if not g.userobj`.

I'm not quite convinced about the `get_user_name()` function though. Can we achieve the same with the `current_user` object?

Can we (for example) override the `name` property of the `User` object to return `''` if it's an anonymous user? That way we can call directly to `current_user.name` instead of importing and calling `get_user_name()`.

ckan/tests/controllers/test_group.py
pdelboca: In fact, the `User` model have an attribute of API Key. Can we return a `user_dict` with the attribute `apikey` directly from the factory?

That way we would reduce the overhead of calling `user["token"] = factories.APIToken(user=user["name"])`.

pdelboca:
@TomeCirun can we start adding some more information about the technical decisions in the PR description? That way it's clear what the changes in the architecture are.

Could you write down some paragraph to explain:
 - `g.user` and `g.userobj` and how it is replaced with `current_user`?
 - What happened with the `REMOTE_USER` and how we changed it. What implications will have this for other extensions?
 - I see that on `datapusher` you changed the way we call `app.get()` in the tests. This will require that all extensions will need to adapt their tests suit? Can we mantain `REMOTE_USER` logic with a deprecation warning? (This are just open questions for debate)

TomeCirun:
Hey @pdelboca thanks for your suggestions, can you look at my last commit [(`add custom AnonymousUser class / update docstrings`)](https://github.com/ckan/ckan/pull/6560/commits/8fa7d43a56233ee4887f755d100a52ee3da7b1f1) and let me know if this is better?

TomeCirun:
@amercader @pdelboca you can try now `saml2auth` extension, I hope it won't break, I've tested it for 3h :)

ckan/tests/controllers/test_group.py
TomeCirun: We cannot use apikey in 2.10 as we removed this line of code: 
https://github.com/ckan/ckan/blob/ca537ad25e13320b6fbf91f41031eadc30500f12/ckan/views/__init__.py#L231
on #6352

TomeCirun:
@smotornyk i need help from you with this:

```python
if current_user.is_anonymous: # type: ignore
```
that `# type: ignore` is everywhere in the code :) can you think of some workaround ?

Thanks

pdelboca:
Hey @TomeCirun, while testing this I found a weird issue. When I first try to login using a new window (i.e: private browsing) I got a `Not authorized to see this page` after clicking in login. My second attempt works correctly. Does it happen to you also?

TomeCirun:
> Hey @TomeCirun, while testing this I found a weird issue. When I first try to login using a new window (i.e: private browsing) I got a `Not authorized to see this page` after clicking in login. My second attempt works correctly. Does it happen to you also?

Once, It might be a caching problem. I m using Chrome on Mac: CMD + SHIFT + R did the job for me :D
or try testing on a clean db

pdelboca:
@TomeCirun this is getting closer to get merged. We need to ensure, however, that we provide some kind of backwards compatibility for the `REMOTE_USER` feature since most extensions relies heavily on it for testing. (As you might have noticed in the core extensions).

This PR will make all tests relying on `REMOTE_USER` to fail, forcing it's maintainers to change all their tests. I saw some experiments in your commits to try to maintain it, How that experience went?

If the goal is to keep it for tests, probably we can implement some hook into the TestClient that could be easily removed in a future. That way, the compatibility is not set at a Core level (where we don't need it) but as a "Middleware" in our Test Application.

Anyhow, we will bring this PR to the Dev Meeting to discuss it broadly since we want it into `CKAN 2.10`. I will also try to explore some options on how to fix this :)

TomeCirun:
> I saw some experiments in your commits to try to maintain it, How that experience went?

Yes, at first I was experimenting with it, it was successfully setting up the `REMOTE_USER` header, but then I realized that we don't need to set up the `REMOTE_USER`, as we already have `current_user`, in fact, the process of how `current_user` is settled, is very similar to that of how  the g.userobj together with the REMOTE_USER were set.

The process is following: when the user first logs in, one of the `repoze.who` middlewares identify the user and is setting  the username into the environ["REMOTE_USER"], then we are assigning that value to the g.user -> checking if g.user has a value, if yes, query the DB for the `userobj` and if there is an `userobj` assign its value to g.userobj.... this is almost the same how flask_login  loads the user and setting up the current_user for us behind the scene with just this lines of code

image

the only difference is that, instead of setting the username to the `REMOTE_USER`, flask_login is setting up the `user_id` to the flask session.

TomeCirun:
> If the goal is to keep it for tests, probably we can implement some hook into the TestClient that could be easily removed in a future. That way, the compatibility is not set at a Core level (where we don't need it) but as a "Middleware" in our Test Application.

@amercader @tino097 @pdelboca 
I believe that this is the easiest way of doing it and we shouldn't have any problems removing it in the future.
I have tried with the TestClient as you said, but it's too much complicated.

pdelboca:
@TomeCirun looking good! I'm now able to run tests in other extensions. 
Some comments:
 - Previously, the `RepozeAdapterMiddleware` turned a `401` into a `302` since it redirected the user to the login form. Now, it directly throws a `401` error.
 
Steps to reproduce:
  - Add ckanext-showcase extension
  - Navigate to `/showcase/new` with an Anonymous user
  - CKAN `master` redirects to login, this branch throws `401` without a redirect.
 
 What do we want the behaviour to be like in CKAN 2.10?
 
 So far, to mimic this behaviour we will need to:
  - Add `login_manager.login_view = "users.login"` AND,
  - Decorate views with `@login_required`

Which is the canonical way of `flask-login` to ask for authentication when accessing views.
  
We provide a `site_read` auth function that kinda acts as `@login_required` (I think). Maybe it's time to remove it since the docstrings are already pointing to it being an obsolete behavior of the system.

What are your thoughts @amercader and @tino097 ?

TomeCirun:
@pdelboca 

Can you test it now? (with my latest commit)
I will try to explain what's happening there:

`@login_required` would execute ONLY if the `error_handler` receives `401/403` with `anonymous_user`, otherwise, our core code will proceed.

Thanks to the error_handler that exists, we don't have to decorate all the functions separately.

Let me know if you agree with the latest changes, so I can fix the tests. 

ps. while writing this I have noticed `RepozeAdapterMiddleware` on master, successfully responds to `401` and fails to respond to `403` while `flask_login` responds successfully on both.

Thanks

amercader:
I still haven't tested this in detail but here are some initial thoughts:

1. We really need to sort that typing stuff to remove all the `# type: ignore`  bits added here. @smotornyuk can you help here?
2. I think the `set_remote_user_as_current_user_for_tests()` approach on before request is a sensible one to maintain compatibility with the `environ["REMOTE_USER"]`, while tests migrate to the "new way" (more on this later), but I'd add two changes.
   * For some reasone repoze wanted `bytes` as the value of that key so most extension tests use `env = {'REMOTE_USER': user['name'].encode('ascii')}` (tbh it's a mess, with some other extensions doing `env = {"REMOTE_USER": six.ensure_str(user["name"])}`). In any case it's good to add a check to decode the user name in case is provided as bytes otherwise you get an SQLAlchemy exception
   * `request.environ['beaker.session']` is too implementation specific and won't work if users use different backends for the session, so let's use the `ckan.common.session` object (ie Flask's session object):
    ```diff
    diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
    index 9a614505e..d8bbfe2dd 100644
    --- a/ckan/config/middleware/flask_app.py
    +++ b/ckan/config/middleware/flask_app.py
    @@ -25,7 +25,7 @@ from flask_babel import Babel
     
     from beaker.middleware import SessionMiddleware
     from flask_login import LoginManager
    -from ckan.common import CKANConfig, asbool
    +from ckan.common import CKANConfig, asbool, session
     
     import ckan.model as model
     from ckan.lib import base
    @@ -325,9 +325,12 @@ def set_remote_user_as_current_user_for_tests():
         '''
         if "REMOTE_USER" in request.environ:
             username = request.environ["REMOTE_USER"]
    +        if isinstance(username, bytes):
    +            username = username.decode()
    +
             userobj = model.User.get(username)
             if userobj:
    -            request.environ['beaker.session']['_user_id'] = userobj.id
    +            session['_user_id'] = userobj.id
     
     
     def ckan_before_request() -> Optional[Response]:
```
3. I'm not conviced by the "new way" of testing ... TODO

**Comment by TomeCirun** [comment]
> 3\. I'm not conviced by the "new way" of testing ...

Well, I needed something to replace the `REMOTE_USER`, and, sending the user_token in the Authorization header seemed sensible to me. I like to hear from you any other suggestions, what other options do we have? I have tried to log the user in like on the UI, but I think it's not a good practice implementing this for the tests...

**Comment by TomeCirun** [comment]
@smotornyuk 
image

**Comment by smotornyuk** [comment]
In this case, use string:

```
...
current_user = cast(Union["Model.User", "Model.AnonymousUser"], _cu)
```

**Comment by amercader** [comment]
On the topic of tests, here are my thoughts.

The way we have traditionally authenticated a user on a functional test (ie one where we perform an actual HTTP request to the test app client) is the following:

```python

def test_dataset_form_renders(app):

    url = url_for("dataset.new")

    user = factories.User()
    env = {"REMOTE_USER": user["name"]}

    app.get(url, environ_overrides=env)

```

(Note that `environ_overrides` is the new Flask-supported version of `extra_environ`, the Pylons version which is still very prevalent cc @TomeCirun ).

Thanks to compatibility handlers included in this PR, `REMOTE_USER` will still be supported but we need to decide what is our suggested approach for logging users in tests going forward.

In this PR @TomeCirun suggests using tokens and the `Authorization` header:


```python
def test_dataset_form_renders(app):

    url = url_for("dataset.new")

    user = factories.User()
    token = factories.APIToken(user=user["name"])
    env = {"Authorization": token]}

    app.get(url, environ_overrides=env)
```

Which I think can be simplified to this if you don't care that much about the actual user (ie whether it's a sysadmin, belongs to an org, etc):

```python
def test_dataset_form_renders(app):

    url = url_for("dataset.new")

    token = factories.APIToken()
    env = {"Authorization": token]}

    app.get(url, environ_overrides=env)
```

The problem with that is that, even though the lines and syntax are pretty much the same, conceptually we are using a completely separate login mechanism, we are authenticating the request like if it was an API one (with the `Authorization` header instead of the one used in web request). That might be fine in most cases if we are not explicitly testing the login process but I find it a bit counterintuitive.

Maybe I'm giving this too much thought but maybe we could explore other APIs that were more user-friendly (I haven't checked how feasible those actually are):

```python
def test_dataset_form_renders(app):

    url = url_for("dataset.new")

    user = factories.User()

    app.get(url, user=user)
```

or 

```python
def test_dataset_form_renders(app):

    url = url_for("dataset.new")

    user = factories.User()

    app.authenticate(user)

    app.get(url, user=user)
```

Worth checking what [Flask-login](https://flask-login.readthedocs.io/en/latest/#automated-testing) offers in terms of logging in  users, but I'm not sure we can adopt that given our custom wrapper around the test client.

I'll raise this for discussion in the dev meeting

TomeCirun:
@adria thanks for your notes, the whole idea behind using the headers was to avoid another post request by calling the view login, which would increase the execution time of the tests. 
Also a good practice I have seen other developers doing, would be by setting the `current_user` explicitly by mocking the flask-login `_get_user` method as I've done with some of the tests in this PR.

TomeCirun:
@pdelboca I want to introduce you to the latest changes I've made, I created a custom `login_required` that would intercept `401/403` if the user is anonymous and will force him to the login page. I created this function so we can use it only in one place: `error_handler`.
As we know `error_handler` is where all the exceptions in `CKAN` happen. Instead of using the `flask-login` decorator and separately decorating all the views that we want to force the users to log in, I created this custom decorator that would decorate only one function.

A recent feature I've added to `CKAN` is the `NEXT` parameter so that the login view can redirect back to that URL. For example, if you go to route `dataset/new` as `anonymous` you will get redirected to the login page, after you log in, CKAN will have use of that NEXT parameter and will redirect you back to that URL where you was before, if you have access you would see the FORM for creating the dataset, otherwise you would get `401 Unauthorized`.

Since we use a custom decorator we needed something to append that `NEXT` parameter (this is handled by `flask-login` if we use its `login_required`) so I copied and modified two functions that would do the same as `flask-login` do for us, behind the scene, (because I was not sure if we have something similar in `CKAN` core).

Thanks

TomeCirun:
@amercader
```python
def test_dataset_form_renders(app):

    url = url_for("dataset.new")

    user = factories.User()

    app.get(url, user=user)
```
This could be possible if we add this function to CKAN core:
```python
def login_users_for_ckan_tests(user_id):
       if user_id is None:
          return
       session["_user_id"] = user_id
```

and we will need to call it in `identify_user()`

```python
if config.get("testing") and request.environ["id"]]:
    login_user_for_ckan_tests(request.environ["id"])
```

This way, we can get much cleaner tests, let me know your opinion on this.

Note: this is not possible to implement on CKAN TestClient atm, it will require refactoring in order to use the session in the tests...

amercader:
@TomeCirun @pdelboca 

> A recent feature I've added to CKAN is the NEXT parameter so that the login view can redirect back to that URL.

This param was named `came_from` in repoze so would be good to keep support for this param name.

I'm not keen on [copying code](https://github.com/ckan/ckan/pull/6560/files#diff-e37df3cabf7b9e56825183a8bd3dabf6ae2f52c7ca0155860151c69f8cd94f68R2873) over from Flask-login. I haven't been following this thread of the discussion closely, but if the reason is to not having to define `@login_required` on all view functions that require an authorized user, I don't think it's worth it. That is I think is fine to add the decorator everywhere where is needed. Or is there another reason for copying over that code?

TomeCirun:
> 1. Can you add a `token` property to the User factory (`token = factory.LazyFunction(lambda: APIToken()["token"])`  [here](https://github.com/ckan/ckan/blob/1bb3387e5b8f95266435d68b69e0d48ad2c64178/ckan/tests/factories.py#L235))

As now implemented i m getting `RecursionError: maximum recursion depth exceeded in comparison`...

If i refactor it, as it was [before](https://github.com/ckan/ckan/pull/6352/files#diff-224ece8505835c5080faa6c82a82223d228b60e47257b16ccf0b37dd7b1c57e5), i m sucesfully creating the token, however, it  didn't show in user_dict, i m not sure if we can add an extra attribute to the factory.User... 

if you know how, please let me know.

TomeCirun:
> This param was named `came_from` in repoze so would be good to keep support for this param name.

'next' is widely used in a flask world, plus it comes as 'next' in request.args

don't you think it's a better  to use like:
```python
next = request.args.get('next')
```
instead of :
```python
came_from = request.args.get('next')
```

@amercader

amercader:
@TomeCirun on the User factory + token I think I prefer this implementation:

```python
class UserWithToken(User):
    """A factory class for creating CKAN users with an associated API token."""

    @factory.post_generation
    def token(obj, create, extracted, **kwargs):
        if not create:
            return
        api_token = APIToken(user=obj["id"])
        obj["token"] = api_token["token"]
```

and then you call `user = factories.UserWithToken()` in the tests. Same for Sysadmin (new `SysadminWithToken` factory)

amercader:
@TomeCirun on the `next` parameter I was meaning that we should support an incoming parameter named `came_from`, internally it can of course be named `next`:

```python
next = request.args.get('next')
if not next:
    next = request.args.get('came_from')
```

or 

```python
next = request.args.get('next', request.args.get('came_from'))
```

TomeCirun:
> 'm not keen on [copying code](https://github.com/ckan/ckan/pull/6560/files#diff-e37df3cabf7b9e56825183a8bd3dabf6ae2f52c7ca0155860151c69f8cd94f68R2873) over from Flask-login. I haven't been following this thread of the discussion closely, but if the reason is to not having to define `@login_required` on all view functions that require an authorized user, I don't think it's worth it. That is I think is fine to add the decorator everywhere where is needed. Or is there another reason for copying over that code?

@amercader 
My main goal was to add support for the extensions, if you decide to go with flask-login's login_required, I can decorate all the views in CKAN core that need it, but then, what about the extensions?

Our custom login_required is very similar to that repoze.middleware, it has the support for the extensions without having to decorate anything.

And regarding the code that was copied from flask-login, I m sure we can find a way to make our own functions that would do the same and delete the copied ones. (they are only creating our NEXT parameter, nothing more)

Thanks

TomeCirun:
@smotornyuk i modified [me()](https://github.com/ckan/ckan/pull/6560/commits/cbada45b5e16b0e31899da4610afd83f145040c7) so it won't break the app with the latest changes you've made, i would like to hear your thoughts on this 

Thanks

ckan/logic/__init__.py
```diff
@@ -329,7 +329,7 @@ def check_access(action: str,
     if not context.get('ignore_auth'):
         if not context.get('__auth_user_obj_checked'):
             if context["user"] and not context["auth_user_obj"]:
-                context['auth_user_obj'] = model.User.by_name(context['user'])
+                context['auth_user_obj'] = model.User.get(context['user'])
```
amercader, line 332: This was discussed in https://github.com/ckan/ckan/pull/6762 and there were some concerns raised about it. Was this change required by Flask login?

ckan/model/user.py
```diff
@@ -87,7 +88,7 @@ class User(core.StatefulObjectMixin,
     DOUBLE_SLASH = re.compile(r':\/([^/])')
 
     @classmethod
-    def by_email(cls: Type[TUser], email: str) -> Optional["User"]:
+    def by_email(cls: Type[TUser], email: Optional[str]) -> Optional["User"]:
```
amercader, line 91: This seems dangerous, is there a reason why we need it optional? Perhaps an `if email` check is needed further up the call stack

amercader (commented):
@TomeCirun some comments below

ckan/model/user.py
```diff
@@ -333,7 +334,21 @@ def user_ids_for_name_or_id(cls, user_list: Iterable[str]=()) -> list[str]:
                                  cls.id.in_(user_list)))  # type: ignore
         return [user.id for user in query.all()]
 
-    def set_user_last_active(self):
+    def get_id(self) -> str:
+        '''Needed by flask-login'''
+        return str(self.id)
```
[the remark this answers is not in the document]
TomeCirun: @amercader 
It was required by Flask-login: 
image

but now I can see that `user_id`  is already a string...
Thanks for this

ckan/templates/header.html
```diff
@@ -20,7 +20,7 @@
             <span class="username">{{ c.userobj.display_name }}</span>
           </a>
         </li>
-        {% set new_activities = h.new_activities() %}
+        {% set new_activities = h.new_activities() or 0 %}
```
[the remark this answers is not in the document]
TomeCirun: indeed

ckan/config/middleware/flask_app.py
```diff
@@ -291,22 +265,22 @@ def save_session(self, app: Any, session: Any, response: Any):
...
+    login_manager = LoginManager()
+    login_manager.init_app(app)
+    # make anonymous_user an instance of CKAN custom class
+    login_manager.anonymous_user = model.AnonymousUser
+    # The name of the view to redirect to when the user needs to log in.
+    login_manager.login_view = "user.login"  # type: ignore
+
+    @login_manager.user_loader
+    def load_user(user_id: str) -> Optional["model.User"]:  # type: ignore
+        return model.User.get(user_id)
```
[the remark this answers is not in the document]
TomeCirun: we are firing it on `login()` 
image

the purpose of this callback function is:
That user_id will come from the user session. If they're logged in, they will have it set as their identity. When a request comes in, user loader attempts to load a user object using this identity and check whether the user exists and is active (not blocked). If all goes well the user is considered authenticated and the object user loader returns can be accessed across the app (within request context) as current_user.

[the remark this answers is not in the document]
TomeCirun: i also removed the extra query here (in def login() ):
image

as we have the same logic in `authenticator.ckan_authenticator` and instead of returning `username`, I modifed it to return `userboj` as we need it in order to log the user in via `login_user`

ckan/lib/authenticator.py
```diff
@@ -37,3 +32,18 @@ def authenticate(
             return user.name
         signals.failed_login.send(login)
         return None
+
+
+def ckan_authenticator(identity: 'Mapping[str, Any]') -> Optional[str]:
```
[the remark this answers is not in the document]
TomeCirun: > Where is this function called from?

we are calling it on login()
image

ckan/lib/authenticator.py
```diff
@@ -1,23 +1,18 @@
...
-from repoze.who.interfaces import IAuthenticator
-
 from ckan.model import User
 from . import signals
 
 log = logging.getLogger(__name__)
 
 
-@implementer(IAuthenticator)
 class UsernamePasswordAuthenticator(object):
```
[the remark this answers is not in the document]
TomeCirun: Ok

ckan/logic/__init__.py
TomeCirun: I believe I encountered an error while testing it because some context ['user'] contains user_id, like this [one](https://github.com/ckan/ckan/blob/96cb766218f723a398cd2a03059bd32b978701f1/ckan/views/user.py#L680-L688)

should I revert it and change that id to `current_user.name` instead?

ckan/model/user.py
TomeCirun: my mistake sry

ckan/lib/helpers.py
```diff
@@ -2751,3 +2758,78 @@ def check_ckan_version(min_version: Optional[str] = None,
...
+        return urlunparse(parsed_base)
+    return base
+
+
+def _ckan_login_required(  # type: ignore
+    func: Any,
+) -> Callable[..., Union[Response, tuple[str, Optional[int]]]]:
+    """
+    This `CKAN` decorator exists only to decorate
+    the `error_handler`.
```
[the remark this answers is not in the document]
TomeCirun: Don't you think that would be too much code for one function?  I thought it's better to make a separate function and use it as a decorator that would be configurable via our `.ini` file, so we can turn it off/on depending on our needs.

ckan/lib/authenticator.py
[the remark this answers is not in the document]
TomeCirun: @amercader `@login_manager.user_loader` is fired after the user logs in, it has only one purpose, to set the current_user for us, pretty same as we were setting the g.userobj with repoze [here](https://github.com/ckan/ckan/blob/a18603ec182e279488c0e91b78cd957a3ed49d30/ckan/views/__init__.py#L171-L172)

ckan/lib/helpers.py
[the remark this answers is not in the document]
amercader: Maybe I'm missing something but I think it's better to have all related logic in the same place:

```diff
diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
index 874bef9c4..f64d41e6f 100644
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -17,7 +17,12 @@ from flask.ctx import _AppCtxGlobals
 from flask.sessions import SessionInterface
 from flask_multistatic import MultiStaticFlask
 
-from werkzeug.exceptions import default_exceptions, HTTPException
+from werkzeug.exceptions import (
+    default_exceptions,
+    HTTPException,
+    Unauthorized,
+    Forbidden
+)
 from werkzeug.routing import Rule
 from werkzeug.local import LocalProxy
 
@@ -33,7 +38,7 @@ from ckan.lib import helpers
 from ckan.lib import jinja_extensions
 from ckan.lib import uploader
 from ckan.lib import i18n
-from ckan.common import config, g, request, ungettext
+from ckan.common import config, g, request, ungettext, current_user
 from ckan.config.middleware.common_middleware import (TrackingMiddleware,
                                                       HostHeaderMiddleware,
                                                       RootPathMiddleware)
@@ -521,6 +526,25 @@ def _register_error_handler(app: CKANApp):
 
     @helpers._ckan_login_required
     def error_handler(e: Exception) -> tuple[str, Optional[int]]:
+
+        if not config.get_value("ckan.login_required.disabled"):
+            # if the url is not local we dont want to redirect the user
+            # instead, we want to show the actual 403(Forbidden)...
+            # same for user.perform_reset if the current_user.is_deleted()
+            # the view returns 403 hence we want to show the actual exception.
+            endpoints = tuple(["util.internal_redirect", "user.perform_reset"])
+            if request.endpoint not in endpoints:
+                if isinstance(e, HTTPException):
+                    if current_user.is_anonymous and type(e) in (Unauthorized, Forbidden):
+                        # make next_url
+                        # TODO: Any reason not to use url_for("user.login") here?
+                        login_url = helpers.redirect_to("user.login").headers["location"]
+                        next_url = request.url
+                        redirect_url = helpers.make_login_url(login_url, next_url=next_url)
+
+                        helpers.flash("Please log in to access this page.")
+                        return helpers.redirect_to(redirect_url)
+
         debug = config.get_value('debug')
         if isinstance(e, HTTPException):
             if debug:

```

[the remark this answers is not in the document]
TomeCirun: Yes, this is much better, i've made only one minor change to avoid double `if isinstance(e, HTTPException):` in the code, please see my latest commit and let me know if you agree to this.

Thanks!

amercader (commented):
@TomeCirun some more comments coming your way, thanks for your patience

amercader:
@TomeCirun sorry this is taking a while, I'm really trying to understand how this works and its implications but I think the time it's worth it as we are getting a really solid implementation.

For anyone interested, I wrote down a detailed summary of the steps followed by repoze.who / Flask-login when identifying/authenticating the users, just to help me get a better understanding:

https://github.com/ckan/ckan/wiki/Login---Authentication-internals

For transparency, here's a list in no particular order of things I want to go through / review / discuss before marking this as done:

- [ ] New login server state in flask-login and its implications, potentially inactivating sessions
- [ ] CKAN wrapper for `flask_login.current_user`
- [ ] Check extensions (ckanext-saml2auth, ckanext-oauth, ???)
- [ ] Flask session cookie configuration (HTTPOnly, Secure, etc)
- [ ] Remember me configuration
- [ ] Builtin Flask-login signals
- [ ] `ckan.login_required.disabled` really needed?

amercader:
Quick update on of the main things I wanted to get the tech team views on: in terms of [cookies and session management](https://github.com/ckan/ckan/wiki/Login-and-Authentication-internals#cookies-and-sessions):

* We won't enforce any specific behaviour regarding the session backend under the hood. If users don't change anything sessions will still be stored in the server, which might means changes in the login behaviour (ie if the session is cleared in the server, users will be logged out). We will clearly document this.
* For *new* instances, we will default to client-side sessions to be closer to Flask's default behaviour (and CKAN < 2.10 with regards to auth cookies). This will mean changing config declarations to set the relevant Beaker config options and maybe changes in the `generate config` command. 
* Changes in Session and Cookies will be documented in the CHANGELOG
* We should also document beaker settings to further secure the session cookie (`beaker.session.secure`, `beaker.session.samesite` etc), and set the appropriate defaults ourselves if necessary

All these changes can be done in a separate PR

amercader:
Quick question @TomeCirun, any reason not to use the latest Flask-login version (0.6.1)?

ckan/config/middleware/flask_app.py
```diff
@@ -517,9 +524,36 @@ def is_blueprint(mm: Any):
...
+                if request.endpoint not in endpoints:
+                    if current_user.is_anonymous and type(e) in (
+                        Unauthorized, Forbidden
+                    ):
+                        login_url = h.url_for('user.login', qualified=True)
+                        next_url = request.url
+                        redirect_url = h.make_login_url(
+                            login_url, next_url=next_url
+                        )
+                        h.flash('Please log in to access this page.')
```
amercader, line 554: I'm not seeing any flash messages when being redirected. Also this message needs to be translated:
```suggestion
                        h.flash_error(_('Please log in to access this page'))
```

amercader:
Ok, some more comments:

> Remember me configuration

I think we should document some or all of the [Remember Me cookie settings](https://flask-login.readthedocs.io/en/latest/#cookie-settings) in our config declaration. (We can do that right @smotornyuk? eg declaring a third-party config option like Flask-login's `REMEMBER_COOKIE_NAME`). We can do that as part of the separate doc RP mentioned in https://github.com/ckan/ckan/pull/6560#issuecomment-1161693905

> CKAN wrapper for flask_login.current_user / flask_login.login_user()

`flask_login.current_user` and `flask_login.login_user()` are so important that I don't feel comfortable importing them directly from core and extensions. I think we should create wrappers for them in `ckan.common` and use those in core / expose them in the toolkit. Otherwise if we ever need to switch from Flask-login to something else it's going to be a nightmare. But maybe I'm being too paranoid. @ckan/core any thoughts on this?


> Builtin Flask-login signals

In addition to our builtin signals for successful and failed logins, Flask-login fires its [own ones](https://flask-login.readthedocs.io/en/latest/#signals). I think it's fine to keep both but maybe it's unnecessary to keep our custom ones. @smotornyuk any views on this?

> ckan.login_required.disabled really needed?

I now understand this setting and it makes sense but the naming is not great can you rename it to `ckan.redirect_to_login_if_not_authorized`, with a default value of `True`. For the config setting description you can use:

```
By default, if a user is not authorized to visit a page, they will get redirected to the login page before trying to access it again. If you want to raise a 401 or 403 error instead, set this setting to `False`
```
BTW I'm not getting any flash message when being redirected to the login page, see separate code comment


I'll now test auth extensions

smotornyuk:
#### Config declaration
Yes, we can(and should) declare FlaskLogin options. Because from the end-user perspective, they are CKAN options, and users don't know that we are just bypassing them to flask-login.
@TomeCirun , don't get confused by existing declarations of the Flask config options, such as [APPLICATION_ROOT](https://github.com/ckan/ckan/blob/master/ckan/config/config_declaration.yaml#L47-L48). As you see, they have `internal: true` flag. I've added it intentionally, in order to hide these options from the users. 

For the Flask-login options, you can use [common declarations](https://github.com/ckan/ckan/blob/master/ckan/config/config_declaration.yaml#L1148-L1152), without the `internal` flag(because we don't want to hide them).
Basically you need to add something like this to the `config_declarations.yaml`:
```yaml

  - annotation: Authentication Settings
    options:
      - key: REMEMBER_COOKIE_NAME
        default: remember_token
        example: custom_remember_cookie_name
        description: The name of the cookie...(rest of description)
      ...(rest of options)
```


If you want to make this feature more resilient, you can declare a CKAN-specific option, like `ckan.auth.remember_cookie_name` and manually pass it to the flask-login during the app initialization. But I don't really think that it is worth such efforts. Even if we migrate to a different authenticator in the future, the chances that it will use the same configuration layout are too low. So the only reason to do such a thing is to be consistent in naming.

#### CKAN wrapper

Even though it creates an extra layer of abstraction and(arguably) makes it harder to learn CKAN, I think that we *should* use a wrapper here. 
Generally, whenever we are talking about processes in CKAN, I prefer developers not to import anything from the 3rd party libraries and use CKAN tools. For example, using `ckan.lib.helpers.literal`, instead of `markupsafe.Markup`.
In this way, we can build backward-compatible adapters / provide CKAN-specific patches /  reliably mock objects during tests. And we can guarantee, that our wrappers work in the same with newer versions of the underlying library. If the extension developer uses flask-login directly and in the next release CKAN will upgrade `flask-login`, the developer has to handle multi-version compatibility issues himself


#### Signals

I don't see value in keeping duplicated signals. They are not included in any public release yet, so it's safe to drop them. 

It looks that [user_logged_in](https://flask-login.readthedocs.io/en/latest/#flask_login.user_logged_in) is identical to our [successful_login](https://github.com/ckan/ckan/blob/master/ckan/lib/signals.py#L54), so let's drop it.

As for [failed_login](https://github.com/ckan/ckan/blob/master/ckan/lib/signals.py#L58), I'm not so sure. If [user_unauthorized](https://flask-login.readthedocs.io/en/latest/#flask_login.user_unauthorized) has the same meaning, we can drop `failed_login` as well

ckan/config/middleware/flask_app.py
TomeCirun: image

I m actually, seeing two error messages,  but you are right, they are not constantly showing, sometimes they pop up sometimes they don't, and I m not sure about the real reason behind this.

@amercader

ckan/model/user.py
```diff
@@ -332,7 +333,21 @@ def user_ids_for_name_or_id(cls, user_list: Iterable[str]=()) -> list[str]:
...
+    @property
+    def is_authenticated(self) -> bool:
+        '''Needed by flask-login'''
+        return True
+
+    @property
+    def is_anonymous(self):
+        '''Needed by flask-login'''
+        return False
+
```
amercader, line 349: Our `User` class inherits from `StatefulObjectMixin`, which has a `is_active()` function. But Flask-login requires it to be a [property](https://flask-login.readthedocs.io/en/latest/#your-user-class), so we need to override it or otherwise it will always return `True` when it is evaluated:

```diff
diff --git a/ckan/model/user.py b/ckan/model/user.py
index 61d65f997..9d3c5728a 100644
--- a/ckan/model/user.py
+++ b/ckan/model/user.py
@@ -347,6 +347,12 @@ class User(core.StatefulObjectMixin,
         '''Needed by flask-login'''
         return False
 
+    @property
+    def is_active(self):
+        '''Needed by flask-login'''
+        return super().is_active()
+
+
     def set_user_last_active(self) -> None:
         if self.last_active:
             session["last_active"] = self.last_active
```

We should probably document this change in the CHANGELOG

amercader:
Thanks @smotornyuk I agree with you on all points, @TomeCirun quick summary:

* Bump the requirement to 0.6.1
* We'll document the remember me (and beaker.session.*) config options in a separate PR
* Create a ckan.common wrapper for `flask_login.login_user()`. Expose `ckan.common.current_user` and `ckan.common.login_user()` on the toolkit
* Let's remove our `successful_login` signal. I think that our `failed_login` is different from Flask-login's `unauthorized` so let's keep that one

ckan/model/user.py
TomeCirun: flask-login uses it only in one place, [here](https://github.com/maxcountryman/flask-login/blob/bb254077ac98ad4bd0ca15f5b6818a3d8f8b5dd7/src/flask_login/utils.py#L193-L194), and it useses it for one purpose only, if the user is not active it won't let the user log in unless you pass force=True. 
But we have our custom check [here](https://github.com/ckan/ckan/pull/6560/files#diff-bced5766e0c0f50671e5e0f57217ef7e79f2496ed3b2c711f90daeef25dccdf1) in our custom authenticator that checks for user inactivity and will return None instead of user_obj, that means the login_user() in this situation is never reached as it will fail to the else block
image


@amercader please let me know how do you want me to proceed 

Thanks

TomeCirun: I can make these changes, remove that check in our custom authenticator and let the flask-login to handle this, then, we could end up with something like this:
```python
if login_user(user_obj):
    return next_page_or_default(next)
else:
     err = _('Login as {} failed - user isn\'t active'.format(username_or_email))
     h.flash_error(err)
     return base.render("user/login.html", extra_vars)
```

amercader:
ckanext-saml2auth is not working for me with the latest changes. The whole Saml2 cycle works fine but the user is not logged in in the CKAN side. I think it's related to the session or rather two sessions being created. Passing `remember=True` to `login_user()` in `set_repoze_user()` seems to work but I don't understand the implications yet

@TomeCirun I'll investigate more but if you can have a look as well that would be great

TomeCirun:
@amercader it works for me, i just noticed a bug [here](https://github.com/keitaroinc/ckanext-saml2auth/blob/7412ff7aba3d215f95a08f99216410e72e60c5bc/ckanext/saml2auth/views/saml2auth.py#L79-L81) our `User.by_email` method no longer returns a list, instead, it returns a single user.

https://user-images.githubusercontent.com/72216462/176220235-560a6b16-3301-4485-8abb-bcb25b6164fe.mov

TomeCirun:
@amercader can someone have a look at this [code?](https://github.com/ckan/ckan/pull/6560/files#diff-e35b23a8adc712f7e307e16a41789f9ca8a5fc6151758ff998263699958fd561L23) I m not sure I m doing the right thing here

Thanks

TomeCirun:
any news on this, @amercader ?

amercader:
@TomeCirun 

Remove the whole `check_session_cookie()` function and any calls to it. Flask should be handling this (and should be doing it correctly after https://github.com/ckan/ckan/pull/6954)

The only thing that I'm still not sure about before merging this is the extensions support. ckanext-saml2auth is definitely not working for me using Azure Active Directory as backend. What SAML2 backend are you using? Looks like it's a different one.
I guess we can fix that separately.

So once you do the small change above I'll merge, with the commitment of working on the following two items before the 2.10 release:

* Document this properly https://github.com/ckan/ckan/issues/6962
* Make sure at least ckanext-saml2auth works fine with different providers. This is good start: https://github.com/keitaroinc/ckanext-saml2auth/pull/71 let me know if you need help with that

amercader:
Let's goooo! :rocket: :rocket:  Thanks very much for bearing with me on the now officially most commented PR in CKAN history @TomeCirun 
There are things to iron out but I do think that CKAN it's more secure and maintainable because of this.

pdelboca:
Yeaaahhh!!! 🚀 

Thanks so much @TomeCirun for this contribution and @amercader for the hard work on reviewing and feedback!

TomeCirun:
@amercader just to know, I've tested saml2auth with  [auth0](https://auth0.com/) provider + Kantega SSO for FishEye/Crucible on Azure Active Directory and it works like a charm

PR #6569, opened 2021-12-02
Fix #6568: add "restart: unless-stopped" to docker-compose

### LINKED ISSUE #6568: Let Docker services restart "always" to survive system reboot
**CKAN version**
Current master [ace6850dd](https://github.com/ckan/ckan/commit/ace6850ddabb1bc533c6942cc434e806c62236e5)

**Describe the bug**
When installed via docker-compose, the services do not restart after a system reboot.

**Steps to reproduce**
Install CKAN following the [CKAN docker-compose docs](http://docs.ckan.org/en/2.9/maintaining/installing/install-from-docker-compose.html), run via `docker-compose up -d`, reboot system.

**Expected behavior**

CKAN services should reboot after a system reboot.

**Additional details**
If possible, please provide the full stack trace of the error raised,  or add screenshots to help explain your problem.

A solution that works on my end:

* Noop: Configure Docker to restart on system boot, ref. [Docker docs on restart](https://docs.docker.com/engine/install/linux-postinstall/#configure-docker-to-start-on-boot) - note: on Ubuntu this is already the default. I have tested this only on a Ubuntu 20.04 VM.
* New: Add `restart: always` policies to each service in docker-compose, ref. [docker-compose docs on restart policy](https://docs.docker.com/compose/compose-file/compose-file-v3/#restart)
* Noop: Run CKAN as per docs as a daemonized process with `docker-compose up -d`.
* CKAN now reboots after the system reboots.

florianm:
Fixes #6568

### Proposed fixes:
In a default Ubuntu VM, this will cause services run with `docker-compose up -d` to restart after system reboot.
This PR uses a more cautious "unless stopped" restart policy to preserve the state of running services across a host reboot. 
The more aggressive "restart: always" discussed in the issue changes the system state after a reboot, which may not be intended.

This particular fix will not require changes to documentation.
Related, the docs chapter on docker-compose could be updated with a few other details. I'll send a separate PR once I've worked freshly through a docker-compose setup.

### Features:

- [ ] includes API changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ defaults:
     name: ckan-redis
 
   solr_image: &solr_image
-    image: ckan/ckan-solr-dev:master
+    image: ckan/ckan-solr-dev:2.9
     name: ckan-solr
 
 jobs:
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -34,6 +34,7 @@ services:
       - CKAN_MAX_UPLOAD_SIZE_MB=${CKAN_MAX_UPLOAD_SIZE_MB}
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
       - DS_RO_PASS=${DATASTORE_READONLY_PASSWORD}
+    restart: unless-stopped
 
     volumes:
       - ckan_config:/etc/ckan
@@ -45,6 +46,7 @@ services:
     image: clementmouchet/datapusher
     ports:
       - "8800:8800"
+    restart: unless-stopped
 
   db:
     container_name: db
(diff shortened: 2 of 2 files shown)


florianm:
I'm seeing a CI breakage on a SOLR schema mismatch outside of the code I've changed:
```
 File "/root/project/ckan/lib/search/__init__.py", line 327, in check_solr_schema_version
    % (version, ', '.join(SUPPORTED_SCHEMA_VERSIONS)))
ckan.lib.search.common.SearchError: SOLR schema version not supported: 1.6. Supported versions are [2.8, 2.9]
ERROR:  syntax error at or near "2021"
LINE 1: 2021-12-02 03:18:10,904 INFO  [ckan.cli] Using configuration...
```

kowh-ai:
@amercader - this can be merged - thanks

PR #6591, opened 2021-12-10
Allow indexing deleted packages

avdata99:
Related to [UNHCR#287](https://github.com/okfn/ckanext-unhcr/issues/287)

### Proposed fixes:

CKAN remove all deleted packages from the search index.
This PR propose to make this behavior, optional (by adding the `ckan.search.remove_deleted_packages` new config value)

Deleted data sets will not be visible until an extension adds the `include_deleted` parameter to the `before_search` function.

The purge dataset form is also updated to list deleted from search index instead DB

Also proposed fix: Start to take into account the value of `verbose` param for the `search-index rebuild` command.
### Features:

- [x] includes API changes

### DIFF
--- a/ckan/cli/search_index.py
+++ b/ckan/cli/search_index.py
@@ -43,7 +43,7 @@ def rebuild(
                 only_missing=only_missing,
                 force=force,
                 defer_commit=(not commit_each),
-                quiet=quiet,
+                quiet=quiet and not verbose,
                 clear=clear)
     except Exception as e:
         error_shout(e)
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1761,6 +1761,10 @@ def package_search(context: Context, data_dict: DataDict) -> ActionResult.Packag
         sysadmin will be returned all draft datasets. Optional, the default is
         ``False``.
     :type include_drafts: bool
+    :param include_deleted: if ``True``, deleted datasets will be included in the
+        results (site configuration "ckan.search.remove_deleted_packages" must
+        be set to False). Optional, the default is ``False``.
+    :type include_deleted: bool
     :param include_private: if ``True``, private datasets will be included in
         the results. Only private datasets from the user's organizations will
         be returned and sysadmins will be returned all private datasets.
@@ -1890,14 +1894,23 @@ def package_search(context: Context, data_dict: DataDict) -> ActionResult.Packag
         else:
             data_dict['fl'] = ' '.join(result_fl)
 
+        data_dict.setdefault('fq', '')
+
         # Remove before these hit solr FIXME: whitelist instead
         include_private = asbool(data_dict.pop('include_private', False))
         include_drafts = asbool(data_dict.pop('include_drafts', False))
-        data_dict.setdefault('fq', '')
+        include_deleted = asbool(data_dict.pop('include_deleted', False))
+        
         if not include_private:
             data_dict['fq'] = '+capacity:public ' + data_dict['fq']
-        if include_drafts:
-            data_dict['fq'] += ' +state:(active OR draft)'
+
+        if '+state' not in data_dict['fq']:
+            states = ['active']
+            if include_drafts:
+                states.append('draft')
+            if include_deleted:
+                states.append('deleted')
+            data_dict['fq'] += ' +state:({})'.format(' OR '.join(states))
 
         # Pop these ones as Solr does not need them
         extras = data_dict.pop('extras', None)
--- a/ckan/views/admin.py
+++ b/ckan/views/admin.py
@@ -12,6 +12,7 @@ import ckan.lib.app_globals as app_globals
 import ckan.lib.base as base
 import ckan.lib.helpers as h
 import ckan.lib.navl.dictization_functions as dict_fns
+import ckan.lib.search as search
 import ckan.logic as logic
 import ckan.model as model
 import ckan.logic.schema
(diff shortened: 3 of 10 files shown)


ckan/cli/search_index.py
```diff
@@ -41,7 +41,7 @@ def rebuild(
                 only_missing=only_missing,
                 force=force,
                 defer_commit=(not commit_each),
-                quiet=quiet,
+                quiet=quiet and not verbose,
```
avdata99, line 44: The `verbose` parameter is not being used. This looks like an error

amercader:
@avdata99 latest master should take care of this failing test

ckan/views/admin.py
```diff
@@ -156,6 +157,34 @@ def __init__(self):
...
+        ).filter_by(
+            state=model.State.DELETED
+        )
+
+    def _get_deleted_datasets_from_search_index(self):
+        query = search.query_for(model.Package)
+        search_params = {
+            'fq': '+state:deleted',
+            'df': 'text',
+            'fl': 'id name title',
```
amercader, line 178: You can add `dataset_type` to `fl` to get the `result['type']`. It's probably better anyway than hardcoding it to support purging other datasets types

doc/maintaining/configuration.rst
```diff
@@ -1199,6 +1199,19 @@ Default value: |config:solr_timeout|
...
+
+Example::
+
+ ckan.search.remove_deleted_packages = true
+
+Default value: |config:ckan.search.remove_deleted_packages|
+
+Removes deleted datasets from search index (default behaviour).  
+If `False`, you can use `include_deleted` in `package_search` 
+and deleted packages will be visible in the packages list.
```
amercader, line 1213:
```suggestion
By default, deleted datasets are removed from the search index so are no longer available in searches. To keep them in the search index, set this setting to ``False``. This will enable the  ``include_deleted`` parameter in the  :py:func:`ckan.logic.action.get.package_search` API action.
```

amercader (commented):
@avdata99 this looks good but I wonder about privacy implications. Can you add some tests to ensure that deleted datasets can only be queried by the relevant users? eg if you delete a private dataset, only members of the org or sysadmins can see it.

ckan/views/admin.py
avdata99: Fixed [here](https://github.com/avdata99/ckan/commit/20d35a04bfa1074308928f2fec5e23bb4ef92c37) @amercader

avdata99:
> @avdata99 this looks good but I wonder about privacy implications. Can you add some tests to ensure that deleted datasets can only be queried by the relevant users? eg if you delete a private dataset, only members of the org or sysadmins can see it.

More tests added @amercader [here](https://github.com/avdata99/ckan/commit/5a6195b78ecadf94d2f0e08aa3329a5f8eb0b8a0).

Deleted datasets will only be visible if we explicitly (e.g. a plugin) uses `include_deleted` in `package_search`.

avdata99:
@Zharktas this PR was rebased and is ready for review. 
I can add more test cases if required

ckan/logic/action/get.py
```diff
@@ -1761,6 +1761,10 @@ def package_search(context: Context, data_dict: DataDict) -> ActionResult.Packag
         sysadmin will be returned all draft datasets. Optional, the default is
         ``False``.
     :type include_drafts: bool
+    :param include_deleted: if ``True``, deleted datasets will be included in the
+        results (site configuration "ckan.search.remove_deleted_packages" must
+        be set to False). Optional, the default is ``False``.
+    :type include_deleted: bool
```
amercader, line 1767:
```suggestion
    :param include_deleted: if ``True``, deleted datasets will be included in the
        results (the configuration option :ref:`ckan.search.remove_deleted_packages` must
        be set to ``False``, otherwise this parameter is ignored). Optional, the default is ``False``.
    :type include_deleted: bool
```

amercader (commented):
Looking good @avdata99, some minor comments

ckan/views/admin.py
```diff
@@ -169,6 +168,36 @@ def __init__(self):
...
+        )
+
+    def _get_deleted_datasets_from_search_index(self) -> List[Any]:
+        query = search.query_for(model.Package)
+        search_params = {
+            'fq': '+state:deleted',
+            'df': 'text',
+            'fl': 'id name title dataset_type',
+        }
+        query.run(search_params)
```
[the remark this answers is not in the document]
amercader: What you did is what I meant @avdata99 

Incidentally, ` package_search({"include_deleted": True, "fq": "+state:deleted"})` works when called directly from python, but it doesn't when coming from an HTTP request to the API, because of how params and parsed and the `fq` checks performed. But that's something for another PR

amercader:
Looks good @avdata99, quick question before merging. Do you have plans for enabling displaying deleted datasets in the UI (ie the `/dataset` page)? AFAICT this is not possible now

avdata99:
> Looks good @avdata99, quick question before merging. Do you have plans for enabling displaying deleted datasets in the UI (ie the `/dataset` page)? AFAICT this is not possible now. To be clear, I don't think it should be enabled by default but it should be possible to implement it from an extension.

@amercader, the way to implement this in an extension is to:
 1- Set `ckan.search.remove_deleted_packages` to `False`
 2- Override the `before_search` function ([like this](https://github.com/okfn/ckanext-unhcr/pull/757/files#diff-1d504120269e385744160b61bb9d6e775ee818ec2e031a8c194af1e18a0819d0R364) to add a `search_params['include_deleted'] = True` it's enough.

PR #6594, opened 2021-12-14
Clean old IRoutes interface

pdelboca:
This Interface is no longer used since it was part of the Pylons architecture.


amercader:
@pdelboca can you add a changes entry (removal) mentioning that this interface is no longer available? Then we can merge

pdelboca:
Done @amercader !

PR #6612, opened 2021-12-20
Accept empty string in one of validator

### LINKED ISSUE #6605: OneOf validator does not accept empty value
**CKAN version**
2.9, master

**Describe the bug**
OneOf validator was added in https://github.com/ckan/ckan/issues/4800, it is used in for example scheming where there are select fields where selecting one is not required. 2.8 accepts it as is, 2.9 fails in validation error. 

Simple test added to scheming here:

https://github.com/vrk-kpa/ckanext-scheming/blob/a4618a7c6dea180a125eaa52441ec9d1d528586a/ckanext/scheming/tests/test_validation.py#L71-L77

In 2.7 and 2.8 the test passes, on 2.9 it fails on validation error 

https://github.com/vrk-kpa/ckanext-scheming/runs/4548504508?check_suite_focus=true#step:7:431

image

This results in optional fields becoming required fields when upgrading ckan to 2.9.

Zharktas:
Fixes #6605 

### Proposed fixes:

Adds a simple check for empty string in OneOf validator. Previous implementation also accepted empty lists and everything except 0 value as empty: [link](https://github.com/bdoms/formencode/blob/3ee909d2dd0e304f195f7625dfb9cf642e8573da/validators.py#L408-L438), should this be also replicated or should this be handled in some other way ?

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -951,9 +951,9 @@ def email_is_unique(key, data, errors, context):
 
 
 def one_of(list_of_value):
-    ''' Checks if the provided value is present in a list '''
+    ''' Checks if the provided value is present in a list or is an empty string'''
     def callable(value):
-        if value not in list_of_value:
+        if value == "" or value not in list_of_value:
             raise Invalid(_('Value must be one of {}'.format(list_of_value)))
         return value
     return callable
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -953,7 +953,7 @@ def email_is_unique(key, data, errors, context):
 def one_of(list_of_value):
     ''' Checks if the provided value is present in a list or is an empty string'''
     def callable(value):
-        if value == "" or value not in list_of_value:
+        if value != "" and value not in list_of_value:
             raise Invalid(_('Value must be one of {}'.format(list_of_value)))
         return value
     return callable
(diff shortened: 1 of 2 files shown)


smotornyuk:
Isn't it better to supply an empty string as an allowed value for `one_of`? Or, if a field should accept an empty option, combine `ignore_empty one_of(...)`. I prefer explicit and small validators, and `one_of` at the moment does exactly what its name says. while your validator is rather `one_of_or_empty_string` :)

Zharktas:
I was thinking that it would probably be enough to add the empty string to accepted values, but since this is an actual regression between 2.8 and 2.9, I thought to restore old functionality. For example scheming adds empty value to option lists automatically if the field is defined optional: https://github.com/ckan/ckanext-scheming/blob/1307aedec8ae8b1c4c86d33ede654a3d9afe7fa3/ckanext/scheming/templates/scheming/form_snippets/select.html#L5-L8

Zharktas:
I'm not happy with my hacky solution though, it could be improved :D

TomeCirun:
Hey @wardi I've noticed when using scheming... this line of code:
https://github.com/ckan/ckanext-scheming/blob/1307aedec8ae8b1c4c86d33ede654a3d9afe7fa3/ckanext/scheming/validation.py#L143-L144
forcing us with `oneOf` validator even if the field is not set to required:True ..
Do you want me to make a PR and make an extra check: `if 'choices' in field and field.get(required)`  only then to enter in `oneOf` validator?

PR #6618, opened 2021-12-23
Exclude site_user from user_list

smotornyuk:
Do not show site_user in the output of `user_list` action by default.

If somebody needs it(why?), add an extra parameter `include_site_user=True` to the request


tino097:
@ckan/core we shoud backport this one to 2.9

PR #6633, opened 2022-01-06
Remove unnecessary block

avdata99:
Fixes #

### Proposed fixes:

The `read.html` file for packages already includes a block called `package_additional_info`

```
  {% block package_additional_info %}
    {% snippet "package/snippets/additional_info.html", pkg_dict=pkg %}
  {% endblock %}
```
The proposed block for deletion exists inside a block with the same name.
This probably it's an error and fail to use this block from extensions

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/package/snippets/additional_info.html
+++ b/ckan/templates/package/snippets/additional_info.html
@@ -8,7 +8,6 @@
       </tr>
     </thead>
     <tbody>
-      {% block package_additional_info %}
         {% if pkg_dict.url %}
           <tr>
             <th scope="row" class="dataset-label">{{ _('Source') }}</th>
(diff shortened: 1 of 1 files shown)


smotornyuk:
It's not an error. Block from `read.html` can be extended when you are rewriting `read.html` in your extension. Block from `additional_info` snippet can be overridden when you extend `additional_info.html`

avdata99:
@smotornyuk I see a block called `package_additional_info` to show the `url` field at the `additional_info.html` file.
Then I see `author_email`, `author`, and `maintainer` fields without any block.
It makes no sense to me.

I think the `package_additional_info` block for the `read.html` file it's ok but I didn't get what we are trying to achieve with the other one.

smotornyuk:
The ability to redefine this block. If someone decides to hide `url` or to add a few extra fields alongside `url`, he can just extend `additional_info.html` and redefine this block. I cannot say that it's a really valuable feature, but somebody may rely on it, so there is no sense in deleting this block.

`author_email`, `author`, and `maintainer` are outside of this block just because someone forgot to put them into it:) Now it's too late to move these fields into `package_additional_info` because it can break existing templates(and doesn't have much value, to be honest, so I doubt that it worth creating a new block for them)

amercader:
Closing for the reasons stated above

PR #5929, opened 2021-03-04
CKAN 2.8 Dockerfile: pin pip version to lower than 21 where python2 is still supported

OriHoch:
Fixes #5925

pip 21 dropped support for Python2 which breaks the docker build, this pins it to a version lower than 21 where Python2 is still supported

PR #6634, opened 2022-01-09
Backport Docker Python 3 support to CKAN 2.9

frafra:
Backport #5929 to CKAN 2.9.


amercader:
Given that 2.9 is a realeased version perhaps is better to have a separate Dockerfile (and image tag) for the Python 3 based images. So you can create a Dokcerfile.py3 or something file instead.

PR #6645, opened 2022-01-12
Remove SynchronousSearch plugin (and related code)

pdelboca:
This is a Draft PR to remove the SynchronousSearch plugin.

### Issue

This plugin triggers an indexing on Solr when the underlying SQLAlchemy object changes. Reasons for removing:
 - It's way over engineered for what's needed
 - Having such a low level trigger generates undesired side-effects that are difficult con control
 - It relies on ISessionExtension which has been deprecated in SQLAlchemy 1.4.
 - It's flawed by design as I documented (a little bit) [in this comment](https://github.com/ckan/ckan/issues/6243#issuecomment-900702852).

### Main Changes
 - Adds indexing in the Action Layer
 - Removes `SynchronousSearch` plugin

### Proposed fixes:

This PR moves the index logic to the Action Layer adding explicit calls to Solr indexing whenever is needed. When it is needed? Whenever some metadata of the package changes and we need to refresh the index:
 - package_create
 - package_update
 - package_delete
 - organization changes (TODO)
 - Tags changes (TODO)
 - others? (TODO)

Moving this change from the Model Layer to the Action Layer means that some current behaviors of the system will change. The main one is the following: **Doing changes at the SQLAlchemy level will not be reflected in the result of `package_show`**.

Example:
```python
pkg = model.Package.by_name('my-package')
pkg.name = 'new-name'
pkg.save()

result = toolkit.get_action('package_show')(context, {'id': pkg.id})

assert result['name'] == 'new-name' # This will fail since Solr hasn't been updated
```
The new approach should be:

```python
pkg = toolkit.get_action('package_show')(context, {'id': 'my-id'})
pkg['name'] = 'new-name'
result = toolkit.get_action('package_update')(context, pkg) # or package_patch

assert result['name'] == 'new-name'
```

This is a more correct way to write our code given our [3-Tier Architecture](https://docs.ckan.org/en/latest/contributing/architecture.html?highlight=Architecture#ckan-code-architecture).

### TODO

- [ ] Define strategy for `org_owner` edit
- [ ] Define strategy for `tags` delete
- [ ] Fix pending tests (8)


pdelboca:
One of the tests failings is:
```
FAILED ckan/tests/lib/test_changes.py::TestChanges::test_remove_org - IndexError: list index out of range
```

It has the following structure:
```python
    @pytest.mark.ckan_config(u"ckan.auth.create_unowned_dataset", True)
    @pytest.mark.usefixtures(u"with_request_context")
    def test_remove_org(self):
        changes = []
        old_org = Organization()
        original = Dataset(owner_org=old_org["id"])

        import ckan.model as model
        pkg = model.Package.get(original["id"])
        pkg.owner_org = None
        pkg.save()

        new = helpers.call_action(u"package_show", id=original["id"])
        check_metadata_changes(changes, original, new)
        assert changes[0]["type"] == u"org"
        assert changes[0]["method"] == u"remove"
```

However it doesn't make any sense since it bypass all the logic to control ownership of the package. I have tried the following approach but is still failing:

```python
@pytest.mark.ckan_config(u"ckan.auth.create_unowned_dataset", True)
    @pytest.mark.usefixtures(u"with_request_context")
    def test_remove_org(self):
        changes = []
        old_org = Organization()
        original = Dataset(owner_org=old_org["id"])

        helpers.call_action(
            "package_patch",
            id=original["id"],
            owner_org=None,
            organization=None
            )

        new = helpers.call_action(u"package_show", id=original["id"])
        check_metadata_changes(changes, original, new)
        assert changes[0]["type"] == u"org"
        assert changes[0]["method"] == u"remove"

```

Seems like the `package_update` logic does not handle well the organization changes. Any suggestions on how to deal with changes in the organization of a package?

pdelboca:
I'm closing this issue since I'm going to work on a different approach in a new one.

PR #6648, opened 2022-01-13
Clean PluginMapperExtension (deprecated dependency)

pdelboca:
This is part of the cleanup needed to update SQLAlchemy to it's latest version since MapperExtension has been removed.

This is part of the work being done as part of: #6243 

### Proposed fixes:

The PluginMapperExtension is not being used in our code, and for a quick search in the Organization repository it looks like it is not used in other extensions:

https://github.com/search?q=org%3Ackan+PluginMapperExtension&type=code

### DIFF
--- a/changes/6648.removal
+++ b/changes/6648.removal
@@ -1,0 +1,2 @@
+The `PluginMapperExtension` has been removed since it was no longer used in core
+and it had a deprecated dependency.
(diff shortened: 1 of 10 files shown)


amercader:
If you add the changes file with the changelog entry for the removal this is good to merge

pdelboca:
@amercader done! :)

PR #6653, opened 2022-01-17
Clean global variable

pdelboca:
This PR cleans the `_connection_pool` global variable from the Redis module.

### DIFF
--- a/ckan/lib/redis.py
+++ b/ckan/lib/redis.py
@@ -16,10 +16,6 @@ from ckan.common import config
 
 log = logging.getLogger(__name__)
 
-# Redis connection pool. Do not use this directly, use ``connect_to_redis``
-# instead.
-_connection_pool = None
-
 
 def connect_to_redis():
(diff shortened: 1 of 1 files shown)


EricSoroos:
https://pypi.org/project/redis/#connection-pools

> Connection Pools
> Behind the scenes, redis-py uses a connection pool to manage connections to a Redis server. By default, each Redis instance you create will in turn create its own connection pool. You can override this behavior and use an existing connection pool by passing an already created connection pool instance to the connection_pool argument of the Redis class. You may choose to do this in order to implement client side sharding or have fine-grain control of how connections are managed.

>>> pool = redis.ConnectionPool(host='localhost', port=6379, db=0)
>>> r = redis.Redis(connection_pool=pool)

Pretty sure this PR basically takes the common connection pool and makes it per-instance, which then becomes per-request to this function. The functionality that was here created one common connection pool and reused it.

PR #6658, opened 2022-01-18
[#6657] Fix not empty validator

### LINKED ISSUE #6657: not_empty validator does not support `False` or 0
**CKAN version**
All

**Describe the bug**
The `not_empty` validator raises an error when the value is any "[falsy](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)" one , including `False` or `0`, which might be legitimately correct in a boolean or int field.

This is generally not an issue because the validators get data sent from an HTTP request so the incoming data_dict that is validated has all values defined as strings.

```python
{
    "name": "my-dataset",
    "visible_to_external_users": "False",
}
```

But if calling the actions directly from Python (eg in the tests with `call_action`) the value can be passed as a boolean and then validation fails:

```python
{
    "name": "my-dataset",
    "visible_to_external_users": False,
}

# raises ckan.logic.ValidationError: None - {'visible_to_external_users': ['Missing value']}
```

`not_empty` should allow `False` and `0` for numeric types.

amercader:
Fixes #6657 

### Proposed fixes:

Modify the `not_empty` validator to allow `False` and 0 values. Add missing tests

### Features:


ckan/lib/navl/validators.py
```diff
@@ -32,7 +32,8 @@ def not_missing(key, data, errors, context):
 def not_empty(key, data, errors, context):
 
     value = data.get(key)
-    if not value or value is missing:
+    valid_values = [False, 0, 0.0]
```
wardi, line 35: 0 and 0.0 are the same for `in` test below

```suggestion
    valid_values = [False, 0]
```

amercader:
@wardi anything else needed here?

pdelboca:
@amercader this one looks good!

As a matter of preference, I rather avoid using long if sentences and go for a more explicit conditions that are easy to read and follow. (I found it difficult to read the code without the context of the bug we are trying to fix.)

Example (`test_validators.py` passing):

```python
def not_empty(key: FlattenKey, data: FlattenDataDict,
              errors: FlattenErrorDict, context: Context) -> None:
    value = data.get(key)
    valid_values = [False, 0]

    if value in valid_values:
        return

    if value is missing:
        errors[key].append(_('Missing value'))
        raise StopOnError

    if not value:
        errors[key].append(_('Missing value'))
        raise StopOnError
```

Or we can shrink it a bit:

```python
def not_empty(key: FlattenKey, data: FlattenDataDict,
              errors: FlattenErrorDict, context: Context) -> None:
    value = data.get(key)
    valid_values = [False, 0]

    if value in valid_values:
        return

    if value is missing or not value:
        errors[key].append(_('Missing value'))
        raise StopOnError
```

amercader:
Fair enough @pdelboca , I've refactored the function

pdelboca (approved):

pdelboca:
Great @amercader ! Thanks :D

PR #6660, opened 2022-01-18
[#6649] gettext not for metadata fields

### LINKED ISSUE #6649: The title of a dataset is translated in some cases
**CKAN version**
2.8.4

**Describe the bug**
The title of a dataset is translated in some cases if you run ckan localized.

**Steps to reproduce**

- run ckan localized, choose the german language (de)
- create a new dataset with the title "private"
- review the created dataset in the detail page: It shows up as "Privat" in title and Breadcrumb

Same happens for example if you name the created dataset "Draft" - it will lead to the title "Entwurf".

**Expected behavior**
The dataset keeps it's given name (stays untranslated).

**Additional details**
grafik
grafik

wardi:
Fixes #6649

### Proposed fixes:

Don't call gettext from `get_translated` helper. This helper is for pulling translations from fields like `title_translated`, `notes_translated`: custom metadata fields that contain different language versions of core metadata fields. It doesn't make sense to fall back to gettext for normal `title`, `notes` fields when a translation is not found. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2748,8 +2748,7 @@ def get_translated(data_dict, field):
     try:
         return data_dict[field + u'_translated'][language]
     except KeyError:
-        val = data_dict.get(field, '')
-        return _(val) if val and isinstance(val, str) else val
+        return data_dict.get(field, '')
 
 
 @core_helper


pdelboca:
@wardi looks good!

PR #6661, opened 2022-01-18
Clean non necessary fixture calls in action tests

pdelboca:
The fixture `with_request_context` should be use only if the test needs access to the session or the request. They are not necessary while testing the action layer since it doesn't access those objects (except specific cases that need some refactoring).

`with_request_context` internally calls flask `create_app` method, so it adds a lot of overhead and it should be use when really needed. Just by removing this fixture I reduced the execution time of the action tests locally from ~9 to ~5 minutes.

From:
```
===== 621 passed, 1 skipped in 515.43s (0:08:35) =====
```

To:
```
===== 621 passed, 1 skipped in 290.00s (0:04:49) =====
```

It also reduces our CircleCI run from ~13m to ~10m.

### DIFF
--- a/ckan/tests/logic/test_logic.py
+++ b/ckan/tests/logic/test_logic.py
@@ -24,6 +24,7 @@ def test_check_access_auth_user_obj_is_not_set():
         assert context["__auth_user_obj_checked"]
         assert context["auth_user_obj"] is None
 
+
 @pytest.mark.usefixtures("clean_db")
 def test_check_access_auth_user_obj_is_set():
     user = factories.User()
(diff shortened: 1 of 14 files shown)


ckan/tests/logic/test_logic.py
```diff
@@ -25,7 +25,6 @@ def test_check_access_auth_user_obj_is_not_set():
         assert context["auth_user_obj"] is None
 
 
-@pytest.mark.usefixtures("clean_db", "with_request_context")
 def test_check_access_auth_user_obj_is_set():
     user = factories.User()
```
wardi, line 29: we're creating a user here, should we not clean_db in case some other test relies on listing users?

ckan/tests/logic/test_logic.py
```diff
@@ -37,7 +36,6 @@ def test_check_access_auth_user_obj_is_set():
     assert context["auth_user_obj"].name == user["name"]
 
 
-@pytest.mark.usefixtures("clean_db", "with_request_context")
```
wardi, line 40: same question as above

pdelboca:
Good catch @wardi ! I fixed both :)

PR #6664, opened 2022-01-20
[6662]add strict_slashes=False

### LINKED ISSUE #6662: Broken request when URL has trailing slash
**CKAN version**
2.9.3

**Describe the bug**
If you browse to a page with a trailing slash, ckan can't find it and shows 404 error

**Steps to reproduce**
Go to                    https://demo.ckan.org/es/dataset/sample-dataset-1                         ................. [200]
then try going to  https://demo.ckan.org/es/dataset/sample-dataset-1/                       .................. [404]
https://demo.ckan.org/es/organization/sample-organization/ .................. [404]
https://demo.ckan.org/es/group/test-group/  .................. [404]

**Expected behavior**
CKAN finds the info requested and show it, or redirect to the correct page without the trailing slash

**Additional details**
None

TomeCirun:
Fixes #6662 

### Proposed fixes:
Added `strict_slashes=False` to prevent 404 when URL has trailing slash at the end 
e.g /organization/my-org/ and /dataset/my-dataset/ 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -1370,7 +1370,7 @@ def history(package_type, id):
 def register_dataset_plugin_rules(blueprint):
     blueprint.add_url_rule(u'/', view_func=search, strict_slashes=False)
     blueprint.add_url_rule(u'/new', view_func=CreateView.as_view(str(u'new')))
-    blueprint.add_url_rule(u'/<id>', view_func=read)
+    blueprint.add_url_rule(u'/<id>', view_func=read, strict_slashes=False)
     blueprint.add_url_rule(u'/resources/<id>', view_func=resources)
     blueprint.add_url_rule(
         u'/edit/<id>', view_func=EditView.as_view(str(u'edit'))
--- a/ckan/views/group.py
+++ b/ckan/views/group.py
@@ -1308,7 +1308,7 @@ def register_group_plugin_rules(blueprint):
         u'/new',
         methods=[u'GET', u'POST'],
         view_func=CreateGroupView.as_view(str(u'new')))
-    blueprint.add_url_rule(u'/<id>', methods=[u'GET'], view_func=read)
+    blueprint.add_url_rule(u'/<id>', methods=[u'GET'], view_func=read, strict_slashes=False)
     blueprint.add_url_rule(
         u'/edit/<id>', view_func=EditGroupView.as_view(str(u'edit')))
     blueprint.add_url_rule(


TomeCirun:
I m not sure if those failures are related to my changes

wardi:
It used to be true that having a single canonical url for pages was better for search rankings. Assuming that hasn't changed, can we redirect instead of accepting both versions?

TomeCirun:
Hey Ian, do these changes seem sensible to you? Btw, the same test is failing and I don't think it's related to my changes... seems  `master` is broken again.

Thanks

ckan/config/middleware/flask_app.py
```diff
@@ -502,6 +502,9 @@ def _register_error_handler(app):
 
     def error_handler(e):
         debug = config.get_value('debug')
+        if request.path.endswith('/'):
+            site_url = config.get_value('ckan.site_url')
+            return helpers.redirect_to(site_url + request.path[:-1])
```
wardi, line 507: I suspect this won't work on sites with a [root_path](https://docs.ckan.org/en/2.9/maintaining/configuration.html#ckan-root-path) setting. Also you could trigger lots of redirects with silly urls like https://mysite.example.com///////////////////////...

TomeCirun: @wardi thanks again for the help I totally forgot about that condition with ////// 
I can fix that problem with strip/regex, and regarding `ckan.rooth_path`, to be honest, I've never used it, but I will test it and will figure out the best option.

wardi:
This change goes against the way flask `strict_slashes` creates a redirect to add a missing trailing slash to urls, and changes the default for ckan sites that have always had a trailing slash, right?

I don't think this is a good idea. We should follow the flask and historical conventions. If some pages aren't being redirected we can add `strict_slashes=True` to the flask app itself instead of redirecting URLs in our own custom middleware.

TomeCirun:
@wardi you are right I will revert it as it was before, I will go only with the changes in `error_handler ()`  CKAN will keep working as before, the only difference will be that now we can catch the errors described in the issue #6662 and we will redirect users to proper URL.

The only thing I m missing that needs to be done is to test CKAN with `ckan.rooth_path` settings.

Thanks

TomeCirun:
@smotornyuk can you help with the test? 
Thanks

smotornyuk:
Sure. Here's your patch:

```diff
diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
index 7ae5ad77f..6dab83181 100644
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -516,7 +516,7 @@ def _register_core_blueprints(app: CKANApp):
 def _register_error_handler(app: CKANApp):
     u'''Register error handler'''
 
-    def error_handler(e: Exception) -> tuple[str, Optional[int]]:
+    def error_handler(e: Exception) -> Union[Response, tuple[str, Optional[int]]]:
         debug = config.get_value('debug')
 
         if request.path.endswith('/'):
```

Previously this handler rendered the template as a string and provided status code(if available), that's why the original type was `tuple[str, Optional[int]]`.
With your change, this handler can return a redirect, which is represented by `flask.Response` object. That requires us to update the signature:
```python
... Union[  # i.e, either of following
    Response,  # redirect
    tuple[str, Optional[int]] # rendered error page
]
```

TomeCirun:
> Sure. Here's your patch:
> 
> ```diff
> diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
> index 7ae5ad77f..6dab83181 100644
> --- a/ckan/config/middleware/flask_app.py
> +++ b/ckan/config/middleware/flask_app.py
> @@ -516,7 +516,7 @@ def _register_core_blueprints(app: CKANApp):
>  def _register_error_handler(app: CKANApp):
>      u'''Register error handler'''
>  
> -    def error_handler(e: Exception) -> tuple[str, Optional[int]]:
> +    def error_handler(e: Exception) -> Union[Response, tuple[str, Optional[int]]]:
>          debug = config.get_value('debug')
>  
>          if request.path.endswith('/'):
> ```
> 
> Previously this handler rendered the template as a string and provided status code(if available), that's why the original type was `tuple[str, Optional[int]]`. With your change, this handler can return a redirect, which is represented by `flask.Response` object. That requires us to update the signature:
> 
> ```python
> ... Union[  # i.e, either of following
>     Response,  # redirect
>     tuple[str, Optional[int]] # rendered error page
> ]
> ```

Thanks

TomeCirun:
Hey @wardi, I added a test for this behavior, if you can look and say is it okay?

Thank you

TomeCirun:
> It used to be true that having a single canonical url for pages was better for search rankings. Assuming that hasn't changed, can we redirect instead of accepting both versions?

Heya, @wardi, it's been a long since this PR, what you've decided? should we redirect or accept both versions by using strict_slashes=True?

wardi:
This isn't how `root_path` works. Basically we should almost never create a url with string operations. Instead we should rely on `url_for` to properly format `site_url` and `root_path`.

TomeCirun:
> root_path

I know, but `request.endpoint` gives me `static` instead of `dataset.read` so I cannot build the URL properly with url_for,

I also tried to catch the error inside the dataset module via @dataset.errorhandler, it would catch the error if you've misspelled the name but not the trailing slash, same goes with @dataset.before_request

@wardi

PR #6679, opened 2022-01-31
Use latest Ubuntu Focal Docker image

### LINKED ISSUE #6635: Dockerfile uses an old version of Ubuntu Focal
https://github.com/ckan/ckan/blob/86474733977828686e8fd1d1d0d3476b8cca2731/Dockerfile#L2

Shouldn't `:focal` be used instead of `:focal-20210119`? Otherwise, someone should always update this Dockerfile and run some tests to assure that compatibility is not broken, which does not seem to be the case.

frafra:
Fixes #6635

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # See CKAN docs on installation from Docker Compose on usage
-FROM ubuntu:focal-20210119
+FROM ubuntu:focal
 MAINTAINER Open Knowledge
 
 # Set timezone


frafra:
Tests are failing due to unrelated changes. I would suggest not to merge changes if test are not passed to simplify maintenance.

kowh-ai (approved):
OK - tested this change - looks good. 

@frafra, this changed was discussed in the Dev meeting this afternoon and there was unanimous agreement this is the better option than pinning a specific focal Docker image

kowh-ai:
This can be merged. The failing test is completely unrelated

PR #6685, opened 2022-02-02
Fixes incorrectly encoded url current url

### LINKED ISSUE #6684: Helpers.current_url returns inconsistently quoted url, causing 500 errors.
**CKAN version**
Definitely 2.8, Probably 2.9, Not sure about master. Bug appears to have been added 5 years ago.  Definitely on python2, maybe on python3.

**Describe the bug**

https://site/%EF%AC%81?foo=bar&bz=%AC%81

`ckan.config.middleware.__init__.py:handle_i18n` urlencodes the querystring, but this function is explictly not for querystrings, as it encodes the special characters. (it encodes & and =, so It's for things that go into urls and querystrings, not the whole querystring).

This urlencoded querystring gets added back into the current url, at which point we have a current "url" which is potentially not a valid url for anything in our system.  As evidence for this, datastore/backend/postgres.py has added a patch to parse the returned url, unquote, and then reassemble.

Because of this, helpers.current_url returns a urldecoded version of the url, which is also not really something that we want to do, because that decodes the path component as well.

Now, once we're url decoding, we're taking it out of ascii and putting essentially arbitrary bytes through a unicode encode/decode process and potentially getting invalid UTF-8, or trying to put UTF-8 bytes into an ascii string, causing an error.

**Steps to reproduce**

Visit https://site/%EF%AC%81 and it returns a 500 error, not a 404.  The path is the fi ligature. 

**Expected behavior**
Return a 404, not a server error.

EricSoroos:
Fixes #6684

### Proposed fixes:

* url in question /%EF%AC%81?foo=bar&bz=%AC%81
* This is a unicode character, which can't be decoded from
ascii. Jinja templates will handle this if it's unicode, or if it's
hex encoded ascii, but can't take a non-unicode string in python 2 and
put this in a template.
* The querystring was being quoted, which is incorrect, as:
  1) the special characters in the query string mean something
  2) The rest of the querystring is already quoted. This makes it
  double quoted, as seen in the datastore file
* We don't want to unquote urls before putting them in the template
anyway.
* There was s further error passing this unicode path to the template
resolution, where in posix path, it fails:
```
File '/usr/lib/ckan/default/lib/python2.7/posixpath.py', line 73 in join
  path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position 1: ordinal not in range(128)
```
The solution here is to make sure it's unicode passed into the
function.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/__init__.py
+++ b/ckan/config/middleware/__init__.py
@@ -151,8 +151,6 @@ class AskAppDispatcherMiddleware(object):
             qs = environ.get('QUERY_STRING')
 
             if qs:
-                # sort out weird encodings
-                qs = urllib.quote(qs, '')
                 environ['CKAN_CURRENT_URL'] = '%s?%s' % (path_info, qs)
             else:
                 environ['CKAN_CURRENT_URL'] = path_info
--- a/ckan/controllers/template.py
+++ b/ckan/controllers/template.py
@@ -29,6 +29,10 @@ class TemplateController(base.BaseController):
         By default this controller aborts the request with a 404 (Not
         Found)
         """
+        try:
+            url = unicode(url.encode('utf-8'))
+        except UnicodeDecodeError, UnicodeEncodeError:
+            return base.abort(404)
         if url.endswith(u'.txt'):
             response.headers[b'Content-Type'] = b'text/plain; charset=utf-8'
         # Default content-type is text/html
--- a/ckan/controllers/template.py
+++ b/ckan/controllers/template.py
@@ -30,7 +30,7 @@ class TemplateController(base.BaseController):
         Found)
         """
         try:
-            url = unicode(url.encode('utf-8'))
+            url = unicode(url.encode(u'utf-8'))
         except UnicodeDecodeError, UnicodeEncodeError:
             return base.abort(404)
         if url.endswith(u'.txt'):
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -601,7 +601,7 @@ def full_current_url():
 @core_helper
 def current_url():
     ''' Returns current url unquoted'''
-    return urllib.unquote(request.environ['CKAN_CURRENT_URL'])
+    return request.environ['CKAN_CURRENT_URL']
 
 
 @core_helper
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -656,7 +656,7 @@ def _insert_links(data_dict, limit, offset):
 
     # change the offset in the url
     parsed = list(urlparse.urlparse(urlstring))
-    query = urllib2.unquote(parsed[4])
+    query = parsed[4]
 
     arguments = dict(urlparse.parse_qsl(query))
     arguments_start = dict(arguments)


amercader:
Let's search usage of CKAN_CURRENT_URL on popular extensions and mention this on the release notes. 
And patch 2.9 and master as well

ckan/controllers/template.py
```diff
@@ -29,6 +29,10 @@ def view(self, url):
         By default this controller aborts the request with a 404 (Not
         Found)
         """
+        try:
+            url = unicode(url.encode('utf-8'))
```
amercader, line 33:
```suggestion
            url = unicode(url.encode(u'utf-8'))
```

amercader (commented):
This sounds good @EricSoroos but can you add the tiny suggestion to make the tests pass, and also submit the same patch for master and 2.9? Thanks

ckan/config/middleware/__init__.py
```diff
@@ -151,8 +151,6 @@ def handle_i18n(self, environ):
             qs = environ.get('QUERY_STRING')
 
             if qs:
-                # sort out weird encodings
-                qs = urllib.quote(qs, '')
```
wardi, line 155: this is done above in line 149 as well. does it make sense to remove only this instance?

amercader:
@EricSoroos do you want to have a look at this?

ckan/controllers/template.py
```diff
@@ -29,6 +29,10 @@ def view(self, url):
         By default this controller aborts the request with a 404 (Not
         Found)
         """
+        try:
+            url = unicode(url.encode(u'utf-8'))
+        except UnicodeDecodeError, UnicodeEncodeError:
+            return base.abort(404)
```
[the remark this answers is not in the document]
EricSoroos: No, it's not. My recollection here is that I specifically need a unicode object in python2 because it eventually gets used in a template context.

ckan/config/middleware/__init__.py
EricSoroos: I don't know. The path info part didn't trigger my error, but my test cases may not have hit it. It doesn't seem to be required from a quick attempt at poking at it. 

This particular issue with this code is that `urllib.quote(qs)` encodes the `&` and `=` in the querystring, and then created an invalid url that required further processing. I don't think that's actually the case with the pathinfo, as it's splitting on the separators already.

wardi:
@amercader this looks good to me but is against the 2.8 branch. If this code doesn't exist on master anymore would we still want something similar for 2.9?

amercader:
@wardi yes, that's https://github.com/ckan/ckan/pull/6926 , right?

amercader:
@EricSoroos This needs to be fixed in master as well:

* https://github.com/ckan/ckan/blob/830a65d3623b1be8c26f96637d03f1878923a510/ckan/views/__init__.py#L194
* https://github.com/ckan/ckan/blob/830a65d3623b1be8c26f96637d03f1878923a510/ckan/lib/helpers.py#L658

Can you send a PR against master with these fixes? Thanks

EricSoroos:
@amercader Will do.

PR #6565, opened 2021-11-30
update the python 3.6 to 3.7 in doc.

Sunny-NEC:
Fixes #6549 

### Proposed fixes:
 3.6 end of life date  in a month , update python version 3.6 to 3.7 in  documents.
and  update the docs to reflect that master requires Python 3.7
### Features:

- [ ] includes tests covering changes
- [X] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #6692, opened 2022-02-08
Fix py3.8

### LINKED ISSUE #6691: ModuleNotFoundError: No module named 'typing_extensions'
In addition to updating the documentation, a fix could be used for newer versions of Python, such as:

```
try:
    # Python 3.8+
    from typing import TypedDict
except ModuleNotFoundError:
    # Python <3.8
    from typing_extensions import TypedDict
```

It could be a good idea to split this issue in two.

_Originally posted by @frafra in https://github.com/ckan/ckan/issues/6549#issuecomment-1032662197_

frafra:
Fixes #6691

### Proposed fixes:
Use built-in `typing` provided with Python 3.5+ even for `TypedDict` (Python 3.8+). This fixes compatibility with newer versions of Python.
We need to use try/except, as CKAN master targets Python 3.7+.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/declaration/load.py
+++ b/ckan/config/declaration/load.py
@@ -3,7 +3,10 @@
 import logging
 import pathlib
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Union
-from typing_extensions import TypedDict
+try:  # Python 3.8+
+    from typing import TypedDict
+except ImportError:  # Python <3.8
+    from typing_extensions import TypedDict
 import yaml
 
 from .key import Key


frafra:
`test_building_the_docs` fails, but it is not related to this PR.
I would suggest enforcing working tests to allow PRs to be merged, to avoid such kind of regressions.

I am not sure if this PR should be backported.

frafra:
It looks like it is not needed: https://github.com/ckan/ckan/pull/6565#issuecomment-1033507541

PR #6705, opened 2022-02-15
[6694]Fix organization package count incorrect

### LINKED ISSUE #6694: /Organization package count incorrect, Collaborator private packages aren't included.
**CKAN 2.9**

Packages that you are a collaborator of are not included in the dataset count on the /organization view page but then are included on the organizations/<organistion_name> page.

I'm not sure whether this is the intended functionality? But it seems like packages you are a collaborator of should be included.

Thanks in advance.

TomeCirun:
Fixes #6694 

I added logic that was missing in `model_dictize` for `dataset-collaborators`. 

In `group.py` module, `def index()` calls for `group_list` which internally calls `organization_show` ... inside `organization_show`, `group_dictize` is handling the authorization for the user that is allowed to see the private dataset. 
However, `model_dictize` was only checking if the user is a member of that organization or is `sysadmin`, the dataset-collaborators were excluded.

`dataset-search` and `organization-read` are good, as they both calls for  `package_search` which has the logic to handle dataset-collaborators.

- [ ] includes API changes


TomeCirun:
Hey Patricio, what's your opinion about this PR?

Thanks

pdelboca:
@TomeCirun what's the goal behind getting and checking the labels?

Every user will have by default 2 labels like for example `['public', 'creator-74db8eef-6458-4e49-9584-7232b740a4e4']` that will trigger the logic even if it is not a collaborator of a private dataset. Also, a few lines later the function calls `package_search` that internally calls again for the labels.

I think that only checking for `config.get('ckan.auth.allow_dataset_collaborators')` should be enough and then let `package_search` handle which private datasets that collaborator can view.

TomeCirun:
@pdelboca  yes, you are right, my mistake...
Thanks

TomeCirun:
I think that failed check-types test is not related to my changes

TomeCirun:
@pdelboca is it ok now ?

Thanks

PR #6710, opened 2022-02-18
Clean isolated Session.close() code.

pdelboca:
All code that opens a connection is responsible for closing it. Having isolated `just-in-case` sessions close is not recommended since it can hide errors in the logic.

This PR also cleans an isolated `model.Session.commit()` that we don't even control what's going to commit.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -68,11 +68,6 @@ def load_environment(conf):
 
     app_globals.reset()
 
-    # issue #3260: remove idle transaction
-    # Session that was used for getting all config params nor committed,
-    # neither removed and we have idle connection as result
-    model.Session.commit()
-
     # Build JavaScript translations. Must be done after plugins have
     # been loaded.
     build_js_translations()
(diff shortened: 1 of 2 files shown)


smotornyuk:
[Removing the code that refers an issue](https://github.com/ckan/ckan/pull/6710/files#diff-fa05bfb271e2e93a67efbbf10f3a8dc019edd3d6d509a1fc78225225a8a705f6L71-L75), without checking for regressions, is also not recommended:)

pdelboca:
I have been testing it before submitting the PR and I couldn't replicate the original issue. I would gladly suggest a proper fix but it seems that the behavior no longer exist in the system.

Do you have a reproducible example of what's currently causing it? :thinking: 

The only thing I can think of is plugins leaving open connections when loading, but in that case I think it would be safer to `close()` the Session rather than committing. (what are we committing to the database in that line?)

smotornyuk:
I see.. the way of reproducing the issue is not clear from the original issue

So, here is the state of all active connections to the PostgreSQL server, when CKAN starts on the master branch: 
image

And here is the same stats with your PR applied:

image


---

I like your suggestion to `close` the session. 

---
> what are we committing to the database in that line?

The incomplete transaction, that you see on the latter screenshot. Don't ask me, why this query runs in transaction, I don't remember it

pdelboca:
Closing this one since I'm gonna be working on each case individually.

PR #6720, opened 2022-02-25
Link to requirements for installation in docs

loleg:
Adds a link to the hardware requirements for CKAN installation. It took me a while to find this information, yes I [searched](https://docs.ckan.org/en/latest/search.html?q=hardware&check_keywords=yes&area=default#) first. Maybe it should just directly be part of the guide. 

Note that the [wiki page](https://github.com/ckan/ckan/wiki/Hardware-Requirements) in question has not been updated since 2014 - still seems appropriate to me, but could use a quick check from the admins.

### Features:


kowh-ai:
Thanks @loleg - I'll update the wiki page to include a bit more information on resources needed to run small-to-medium CKAN implementations

PR #6722, opened 2022-02-27
[6721] Create and return api token for site-user

### LINKED ISSUE #6721: Api Token and `get_site_user`
After introducing the Api Token and merging  #6352 we are moving towards removing apikey from the [user](https://github.com/ckan/ckan/blob/master/ckan/model/user.py#L53) model. 

That will cause [`get_site_user`](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L2402) action not to work anymore so we will need to create api token and return its value back in the response

tino097:
Fixes #6721 

Create and return api token for site_user

- [x] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2399,19 +2399,26 @@ def get_site_user(context: Context, data_dict: DataDict) -> ActionResult.GetSite
     site_id = config.get_value('ckan.site_id')
     user = model.User.get(site_id)
     if not user:
+        #To be removed in next release
         apikey = str(uuid.uuid4())
         user = model.User(name=site_id,
                           password=apikey,
                           apikey=apikey)
+       
         # make sysadmin
         user.sysadmin = True
         model.Session.add(user)
         model.Session.flush()
         if not context.get('defer_commit'):
             model.repo.commit()
+            
+    # make token 
+    api_token = logic.get_action("api_token_create")(context, 
+                                {'user': user.id, 'name': 'API access'})
 
     return {'name': user.name,
-            'apikey': user.apikey}
+            'apikey': user.apikey,
+            'api_token': api_token['token']}
 
 
 def status_show(context: Context, data_dict: DataDict) -> ActionResult.StatusShow:


wardi:
`get_site_user` is called a lot. It doesn't make sense to create a new api token and store it in the database every time.

Which callers of `get_site_user` need an api key?

tino097:
@wardi here is one [example](https://github.com/ckan/ckanext-xloader/blob/710e1de80db0350998b3bced2b581a39d37453fb/ckanext/xloader/action.py#L149) 
Also we are using it in some projects for the CLI commands

wardi:
@tino097 closing this for now. datapusher and other extensions that pass a sysadmin apikey to a microservice or other application will need to be modified to generate and store an api token on initial setup.

tino097:
OK i will check that, but the part of cleanup of the apikey should be done as we will remove that field from the model

wardi:
Yes, that part we need to keep. Feel free to reopen if you prefer to use the same PR (the description would be pretty misleading though)

tino097:
No i will go with separate PR

TomeCirun:
Hey @wardi, if you enable the xloader extension against CKAN master, when you try to `Upload to DataStore`, its status never gets `Complete`, instead, it shows `Pending`, and it's a problem, especially for extensions that depend on status 'complete' which triggers `after_upload()`.
https://github.com/ckan/ckanext-xloader/blob/4450bbb9f7085b951e5fb1a6680549a8c8c90e8e/ckanext/xloader/action.py#L266


This happens if we are using apikey as it is set by default :
https://github.com/ckan/ckanext-xloader/blob/4450bbb9f7085b951e5fb1a6680549a8c8c90e8e/ckanext/xloader/action.py#L154

but if we create API token, and instead of apikey, we send the token, then its status will show as `Complete`.
I m not quite sure but this function might be the reason: https://github.com/ckan/ckanext-xloader/blob/4450bbb9f7085b951e5fb1a6680549a8c8c90e8e/ckanext/xloader/jobs.py#L373

PR #6723, opened 2022-03-01
GitHub Actions: Use bandit to check security

cclauss:
https://pypi.org/project/bandit

Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -38,10 +38,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v3
         with:
           python-version: ${{ env.PYTHON_VERSION }}
-      - uses: actions/setup-node@v2-beta
+      - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_VERSION }}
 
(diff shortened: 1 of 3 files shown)


Zharktas:
@cclauss what does bandit actually do ? it would be nice have a description about it without having to read the documentation itself. How does it differ from CodeQL which we already are running on commits ?

And if we would start using this, it should be in its on job and not part of the linting. Linting should not fail due to security issues.

cclauss:
> Bandit is a tool designed to find common security issues in Python code. To do this Bandit processes each file, builds an AST from it, and runs appropriate plugins against the AST nodes. Once Bandit has finished scanning all the files it generates a report. Bandit was originally developed within the OpenStack Security Project and later rehomed to PyCQA.

When was the last time that CodeQL told you something useful?  If your experience is like mine, the answer is never.

Building and running an image takes time and resources so It is not really a good idea to do it separately for each test runner.  I usually stack up tests: https://github.com/cclauss/itinerant-tester/blob/master/.github/workflows/lint_python.yml to avoid repetitive steps.

Zharktas:
Does bandit find anything useful, I find it hard to believe that 15 year old code base would not have any issues but the Action run says there isn't any ?

The CKAN project has a convention of creating an issue and discussing it first especially if the change affects work processes. Now this just seems to be imposed as "you should run this" but it doesn't seem to anything useful for us.

cclauss:
I have landed about 30 PRs on this repo so please hold the lecture.

PR #6725, opened 2022-03-01
Fix for Norwegian language

### LINKED ISSUE #4933: Norwegian language is not working with Flask blueprints
### CKAN Version
2.8, master

### Expected behaviour
When changing language to Norwegian, it should translate the pages. (Implementation from views, not controllers)

### Actual behaviour
When changing language to Norwegian nothing happens

### Steps to reproduce
1. Setup pure CKAN 2.8 or master.
2. Change language to Norwegian or homepage or about page.

frafra:
Fixes #4933

### Proposed fixes:
This is mostly a workaround which should allow Norwegian Bokmål users to see CKAN in Norwegian (Bokmål is the most common way to write Norwegian).


frafra:
Meeting notes 01/03/2022: instead of fixing it with a symlink, we will move Transifex "no" to "nb_NO", and update CKAN accordingly.

amercader:
Closed in favour of https://github.com/ckan/ckan/pull/6746

PR #6735, opened 2022-03-02
[6611]Groups should be selectable as Dataset collaborators

### LINKED ISSUE #6611: Groups should be selectable as Dataset collaborators
The Dataset collaborators feature is a great addition in 2.9. However, it only allows to select individual users as collaborators, not groups.
CKAN describes groups as means for _catalogue datasets for a particular project or team_. Having to add all individual users of a team to a dataset as collaborators individually can be error prone for larger teams. Being able to make groups collaborators on Datasets also means a user could just be removed from a group versus having to remove the user from all datasets individually.
Is this maybe planned for the _Dataset collaborators_ feature or is there a solution already?
Thank you!

TomeCirun:
Fixes #6611 

I made a few changes that will allow us to push all the users in a group, on one go.
If you have any suggestions on how to make the code better, please do not hesitate to propose.

Here is the screenshot from the UI:
image

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2381,7 +2381,12 @@ def format_resource_items(
         # size is treated specially as we want to show in MiB etc
         if key == 'size':
             try:
-                value = formatters.localised_filesize(int(value))
+                # type_ignore_reason: ``value`` variable got overrided and it's
+                # type skrewed as result. It's better to save localized value
+                # into a different variable, but for now let's just ignore the
+                # following line as it wrapped into try/except block.
+                value = formatters.localised_filesize(
+                    int(value))  # type: ignore
             except ValueError:
                 # Sometimes values that can't be converted to ints can sneak
                 # into the db. In this case, just leave them as they are.
@@ -2398,7 +2403,7 @@ def format_resource_items(
         elif isinstance(value, bool):
             value = str(value)
         elif isinstance(value, numbers.Number):
-            value = formatters.localised_number(value)
+            value = formatters.localised_number(cast(float, value))
         key = key.replace('_', ' ')
         output.append((key, value))
     return sorted(output, key=lambda x: x[0])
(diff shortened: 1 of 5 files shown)


TomeCirun:
@smotornyuk  can you help with the test ?
Thanks

jeverling:
Hi, thanks for working on this!
This MR iterates through all users of a group, and individually adds them as collaborators on a given dataset, right?
This has the drawback, that it's not going to be updated when users are added or removed from a group later.

Ideally, group membership should just be used to determine whether a user is allowed to access a dataset or not, and it should not be necessary to add all users as collaborators individually.

Check out the extension that was mentioned in the issue: https://github.com/USEPA/ckanext-collaborator_orgs/blob/master/ckanext/collaborator_orgs/logic/auth.py#L62

I think hooking into the auth check to include group memberships when trying to determine whether a given user has access to a dataset or not is more flexible and more future-proof than statically adding all users as collaborators once.

TomeCirun:
> I think hooking into the auth check to include group memberships when trying to determine whether a given user has access to a dataset or not is more flexible and more future-proof than statically adding all users as collaborators once.

Hey @jeverling , yep I agree with that, but then we will have to add another column to  `package_member` table https://github.com/ckan/ckan/blob/9dc138543580e8002b0f34b735a96079307496b6/ckan/model/package.py#L78-L85 that would contain `group_or_org_id` as we need to know which group has been added as a collaborator so we can implement the logic according that...
I could use some help from @pdelboca  to tell me how to refactor `package_member` (if it's even possible to implement it like this)

Thanks

pdekraker-epa:
I would like to see this feature integrated into CKAN, and it looks like this PR has stalled. The last question was how to add it to the database. Being an extension ckanext-collaborator_orgs used a separate table, is that a better approach than adding a column to the existing table? There may be other nuances about capacity that differ as well between user collaborators and organization collaborators, so a separate table may be more flexible in that regard.

EricSoroos:
Closing this, as #6905 a successor to this.

PR #6737, opened 2022-03-04
Downloaded file have the same resource name.

### LINKED ISSUE #6564: enh: Resource download buttons should have the HTML5 option `<a download="{{ res.name }}">`
**CKAN version**
2.9.4

**Describe the bug**
Some uploaded resources may have upper-case chacaters, such as "MyUploadIng.txt". The [Download](https://github.com/ckan/ckan/blob/ckan-2.9.4/ckan/templates/package/resource_read.html#L36) in the resource view links to an all-lower case file name. It would be nice if the downloaded file would have the correct name.

**Steps to reproduce**
1. Upload a resource with upper-case characters.
2. Download it (the downloaded file will not have the same name (lower-case)

**Expected behavior**
Downloaded file should have the same resource name.

**Fix**
This can be fixed by setting the HTML5 `download` attribute to the resource name. Modern browsers support that.
Replace [this line](https://github.com/ckan/ckan/blob/ckan-2.9.4/ckan/templates/package/resource_read.html#L36):

```
<a class="btn btn-primary resource-url-analytics resource-type-{{ res.resource_type }}" href="{{ res.url }}">
```

with this line:

```
<a class="btn btn-primary resource-url-analytics resource-type-{{ res.resource_type }}" href="{{ res.url }} download="{{ res.name }}"">
```

robin-NEC:
Fixes #6564 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/package/resource_read.html
+++ b/ckan/templates/package/resource_read.html
@@ -33,7 +33,7 @@
             {% if res.url and h.is_url(res.url) %}
               <li>
                 <div class="btn-group">
-                <a class="btn btn-primary resource-url-analytics" href="{{ res.url }}">
+                <a class="btn btn-primary resource-url-analytics" download = "{{ res.name }}" href="{{ res.url }}">
                   {% if res.resource_type in ('listing', 'service') %}
                     <i class="fa fa-eye"></i> {{ _('View') }}
                   {% elif  res.resource_type == 'api' %}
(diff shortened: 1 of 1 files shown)


wardi:
The "name" field of a resource isn't necessarily the file name.

The resource form presents it as a text input with a placeholder that says "eg. January 2011 Gold Prices" and the name field isn't required so it's likely to be left blank.

robin-NEC:
Hi @wardi , can you brief me on this, where the file names are being handled when we're uploading them?
I've tried with multiple file names and it's working the way it should.

robin-NEC:
Hi @amercader . Please review this PR and let me know if there are changes needed to be done.

PR #6746, opened 2022-03-11
[#4933] Fix Norwegian translation

### LINKED ISSUE #4933: Norwegian language is not working with Flask blueprints

amercader:
Fixes #4933

@frafra This PR is a follow up from our discussion in #6725 

In this approach, I've copied the existing `no` translation on Transifex and copied it to a new [`nb_NO` language](https://www.transifex.com/okfn/ckan/language/nb_NO/) and renamed the i18n folder in the CKAN source.
I've also added a redirect for `/no/` URLs to `/nb_NO/`

Do you think this is the most sensible approach?
This will probably break sites that are currently using `no` in the `ckan.locale_default` config option. We can mention that in the CHANGELOG, or is it worth exploring keeping a symlink no -> nb_NO for compatibility?

TODO on Transifex:
* @frafra I've made you a manager of this language, would you mind inviting the existing users from `no` over to `nb_NO`?
* Shall I delete the existing `no` language on Transifex?


frafra:
@amercader Invitations sent.
I think we can remove the `no` language.

amercader:
Thanks @frafra

PR #6747, opened 2022-03-11
Added template helpers functions to encode and decode special charact…

JVickery-TBS:
…ers in the filters for datatables view.

Fixes #

### Proposed fixes:

Encoding and decoding special characters for the datatables view filters. Special characters in fanstatic throw ansii encoding errors.

### Features:

- [ ] includes API changes


ckanext/datatablesview/plugin.py
```diff
@@ -22,6 +22,14 @@ class DataTablesView(p.SingletonPlugin):
     p.implements(p.IConfigurer, inherit=True)
     p.implements(p.IResourceView, inherit=True)
     p.implements(p.IBlueprint)
+    p.implements(p.ITemplateHelpers, inherit=True)
+
+    # ITemplateHelpers
+
+    def get_helpers(self):
+        return{
+            'encode_datatables_request_filters': helpers.encode_datatables_request_filters
```
amercader, line 31: @JVickery-TBS This sounds like a valuable helper for other view plugins as well. Shall we rename them to `encode/decode_view_request_filters` move it to `ckan/lib/helpers.py` so it's available to everybody?

amercader (commented):
Thanks @JVickery-TBS, see comment below.

Also we should a small test for this. Something in `ckan/tests/lib/test_helpers.py` that checks that the output of the functions is the expected one.
Cheers

ckanext/datatablesview/helpers.py
```diff
@@ -0,0 +1,14 @@
+from urllib.parse import quote, unquote
+from cgi import escape
+
+from ckan.plugins.toolkit import request
+
+def encode_datatables_request_filters():
+    if 'filters' not in request.params:
+        return ''
+    return escape(quote(str(request.params['filters'].encode('utf-8')).encode('ascii')))
```
amercader, line 9: Use `request.args` instead, as that's the Flask supported name. Same in the function below

ckan/public/base/javascript/modules/resource-view-filters.js
```diff
@@ -60,6 +60,18 @@ this.ckan.module('resource-view-filters', function (jQuery) {
   }
 
   function _appendDropdowns(dropdowns, resourceId, template, fields, filters) {
+    //TODO: handle decoding the filters object to be able to properly available in the current filter dropdowns
+    console.log(filters);
```
amercader, line 64: Maybe remove this to avoid noise
```suggestion
    console.log(filters);
```

amercader (approved):
This looks good @JVickery-TBS, see a small comment and can you fix the conflicts with master before we merge? cheers

JVickery-TBS:
@amercader Hey!

Okay I believe I have accomplished the desired affect here with the view filters. There were cases in which users may have colons (:) and/or pipes (|) in their data. This would break the filters because the filters are flattened. The filter names and values are separated by colons and multiple filters are separated by pipes.

So what will be done now is that the filter names and values will be URI encoded in JS before they are flattened into the colon and pipe string. The new helper now does the separation of the colons and pipes, THEN decodes the filter names and values. The helper now returns a dictionary, the properties being the filter names, and the values being arrays containing the filter values.

Ultimately this will allows for special accented characters in filter names and values. It will also allow for colons and pipes in the filter names and values.

amercader:
Thanks @JVickery-TBS, can you add a `changes/6747.misc` file with a short summary for the changelog? Thanks

JVickery-TBS:
@amercader `changes/6747.misc` should be there now. And I just fixed some syntax stuff and added/modified some annotations as to pass all of the tests.

amercader:
Thanks @JVickery-TBS

PR #6748, opened 2022-03-13
[6165]Users can login with username or email

### LINKED ISSUE #6165: Users should be able to login with email address
**CKAN version**
master

Email addresses have been unique since 2.9 and password resets can be done with email addresses. Currently it's confusing for users who reset their passwords using email address but need to know their user names when logging in. There should not be any issues enabling login with email address.

TomeCirun:
Fixes #6165

### Proposed fixes:

Allow users to login using their registered email.

### Features:

- [ ] includes API changes


pdelboca:
@TomeCirun this one looks good!

Just one extra comment to @amercader or @tino097 . Why `User.by_email(...)` is returning a list? I would suggest to refactor it to:
```python
    @classmethod
    def by_email(cls: Type[TUser], email: str) -> Optional["User"]:
        return meta.Session.query(cls).filter_by(email=email).first()
```
Just as we return `.first()` when using `User.by_name(...)`

amercader:
> Why User.by_email(...) is returning a list?

@pdelboca because in the olden times CKAN didn't enforce a unique email per user. It makes total sense to return a single user now (or None) but it's worth doing a quick search in the code to make sure this won't break anything, and add a quick note in the changelog

pdelboca:
Thanks for the fixes @TomeCirun !

Can you add an new change file `changes/6748.changes` to document the change in `User.by_email(...)`? With that one we're ready to go!

pdelboca:
Thanks for your contribution @TomeCirun ! :)

PR #6759, opened 2022-03-21
Minor fix: update install-from-docker-compose

FrancescoMalluzzo:
Inside the DB container you should be sure to be in right folder before launch command, minor fixes

Fixes #

### Proposed fixes:

Modified the docs to ensure be in the actual folder, as follow:

```bash
/docker-entrypoint-initdb.d/20_postgis_permissions.sql
```

Feel free to give your opinion on this documentation update :) 

### Features:

- [X ] includes updated documentation
- [ ] includes API changes

### DIFF
--- a/doc/maintaining/installing/install-from-docker-compose.rst
+++ b/doc/maintaining/installing/install-from-docker-compose.rst
@@ -363,7 +363,10 @@ E.g., `ckanext-spatial <https://github.com/ckan/ckanext-spatial.git>`_::
     exit
 
     # On the host
-    docker exec -it db psql -U ckan -f 20_postgis_permissions.sql
+    docker exec -it db bash
+    #Inside the db container 
+    cd docker-entrypoint-initdb.d/
+    psql -U ckan -f 20_postgis_permissions.sql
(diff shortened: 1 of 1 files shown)


kowh-ai:
@FrancescoMalluzzo  - You are correct, the documentation needs to be updated, however as the command that needs changing is in the  `# On the host` section,  maybe you could just give the absolute path to the .sql file ie: `docker exec -it db psql -U ckan -f /docker-entrypoint-initdb.d/20_postgis_permissions.sql`...this works from the host, no need to login to the container to run the command. What do you think?

FrancescoMalluzzo:
> @FrancescoMalluzzo - You are correct, the documentation needs to be updated, however as the command that needs changing is in the `# On the host` section, maybe you could just give the absolute path to the .sql file ie: `docker exec -it db psql -U ckan -f /docker-entrypoint-initdb.d/20_postgis_permissions.sql`...this works from the host, no need to login to the container to run the command. What do you think?

This could be very useful and IMHO launch command from the host is more safely, I've made another commit with your suggestion. 

Thank you :)

kowh-ai (approved):
Looks good

PR #6762, opened 2022-03-21
Allow context['user'] to be the users id

avdata99:
Is the `name` of the user mandatory for the context?
Can we use the `id`?

**This is just a question PR**

### Proposed fixes:
Allow to use the user id in context and not only the name

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -328,7 +328,7 @@ def check_access(action: str,
     if not context.get('ignore_auth'):
         if not context.get('__auth_user_obj_checked'):
             if context["user"] and not context["auth_user_obj"]:
-                context['auth_user_obj'] = model.User.by_name(context['user'])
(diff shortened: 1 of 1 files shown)


amercader:
Some thoughts from the dev call:

* We use the user name in `context['user']` probably because that's the default value in `g.user`, which gets set on each request based on the user authentication.
* While it will be good to work towards using user ids across the board, we need to be careful with unintended consequences, eg code that relies on user names, etc
* @wardi I checked and `user_name_validator` prevents you from creating a new user that uses an existing user id as their user name
* If we wanted to support passing user ids in the context for auth purposes we could use a separate key to be on the safe side (eg `context['user_id']`). That would mean reviewing all current code that relies on `context['user']` to support the new key, but maybe this is an opportunity to centralize this handling

avdata99:
Cleaning, too old

PR #6765, opened 2022-03-23
Clean old c attribute from helpers.py

pdelboca:
The `c` object is a legacy attribute that adds noise and complexity to our code base. While working with Pylons, the `c` object was used a lot to store variables needed when rendering templates. Currently we have the `extra_vars` dict so some of the changes is to explicitly pass the variable to the snippets and do not rely on it being magically stored in the `c` object.

This PR cleans its usage in the `helpers.py` module.

## Changes
- There are some direct replacements `c` -> `g` since currently we proxy the value
- Add some explicit `search_facets` parameters to remove the dependency on `c.search_facets`

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1168,7 +1168,7 @@ def unselected_facet_items(
 
     '''
     return get_facet_items_dict(
-        facet, c.search_facets, limit=limit, exclude_active=True)
+        facet, g.search_facets, limit=limit, exclude_active=True)
 
 
 @core_helper
@@ -1774,11 +1774,11 @@ def dump_json(obj: Any, **kw: Any) -> str:
 
 @core_helper
 def auto_log_message() -> str:
-    if (c.action == 'new'):
+    if (g.action == 'new'):
         return _('Created new dataset.')
-    elif (c.action == 'editresources'):
+    elif (g.action == 'editresources'):
         return _('Edited resources.')
-    elif (c.action == 'edit'):
+    elif (g.action == 'edit'):
         return _('Edited settings.')
     return ''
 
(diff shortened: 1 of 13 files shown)


ckan/lib/helpers.py
```diff
@@ -1140,37 +1136,13 @@ def has_more_facets(facet: str,
...
-        limit = c.search_facets_limits.get(facet)
+    if getattr(g, 'search_facets_limits', None) and limit is None:
+        limit = g.search_facets_limits.get(facet)
     if limit is not None and len(facets) > limit:
         return True
     return False
 
 
-@core_helper
-def unselected_facet_items(
```
tino097, line 1151: this helper can be used in extensions, maybe we should mark first as deprecated and remove it in next release ?

pdelboca: As discussed in the Tech Team meeting, I've added a change file for this change.

ckan/templates/home/snippets/search.html
```diff
@@ -1,4 +1,4 @@
-{% set tags = h.get_facet_items_dict('tags', c.search_facets, limit=3) %}
+{% set tags = h.get_facet_items_dict('tags', search_facets, limit=3) %}
```
MarkCalvert, line 1: FYI This always returns no tags because `search_facets` does not exist because it is not being set in the `extra_vars` object passed to this template from the blueprint `home.index`.

pdelboca: Thanks @MarkCalvert ! Would you like to create a PR fixing it? :)

MarkCalvert: Sure. We will add it to our list to contribute back a PR.

tino097: This was fixed in 2.10 dev branch but i didnt added the labels for backports 
https://github.com/ckan/ckan/issues/7519

tino097:
@amercader this PR shoudl be included in 2.10.2

amercader:
@tino097 I guess you mean https://github.com/ckan/ckan/pull/7520 should be included right? Both #6765 and #7520 are backported to the dev-v2.10

PR #6766, opened 2022-03-23
Clean c object reference from plugins module (and group templates)

pdelboca:
This PR started as updating the reference `c` -> `g` until I realize that the implementation of `setup_template_variables` is no longer needed since group templates no longer uses `g.auth_for_change_state` nor `g.group`.

I ended up cleaning this method.

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -553,18 +553,9 @@ class DefaultGroupForm(object):
 
     def setup_template_variables(self, context: Context,
                                  data_dict: dict[str, Any]) -> None:
-        ## This is messy as auths take domain object not data_dict
         context_group = context.get('group', None)
         group = context_group or getattr(g, 'group', None)
-        if group:
-            try:
-                if not context_group:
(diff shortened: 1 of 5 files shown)


ckan/lib/plugins.py
```diff
@@ -553,22 +553,7 @@ def check_data_dict(self, data_dict: dict[str, Any]) -> None:
...
-        c.is_sysadmin = ckan.authz.is_sysadmin(c.user)
-
-        ## This is messy as auths take domain object not data_dict
-        context_group = context.get('group', None)
-        group = context_group or getattr(c, 'group', None)
-        if group:
-            try:
-                if not context_group:
-                    context['group'] = group
-                logic.check_access('group_change_state', context)
```
pdelboca, line 565: As a reference, the logic for `group_change_state` is now being executed in a validator: https://github.com/ckan/ckan/blob/37f86db98b8e2e8b67155b8ba1e591873b9ba2de/ckan/logic/validators.py#L547-L568

PR #6772, opened 2022-03-24
API serializers

smotornyuk:


amercader:
> PS due to war I still cannot attend meetings and don't check notifications every day, but I'll definitely answer your questions (if there are any) at some point :)

Setting aside the API serializers (which look really cool) just want to say that you are an example and an inspiration to us all. I'm glad you have the chance of doing such great work in the midst of this cursed war. Thanks for your work, now more than ever!

ckan/tests/controllers/test_api.py
```diff
@@ -346,3 +350,32 @@ def test_i18n_only_known_locales_are_accepted(app):
     url = url_for("api.i18n_js_translations", ver=2, lang="unknown_lang")
     r = app.get(url, status=400)
     assert "Bad request - Unknown locale" in r.get_data(as_text=True)
+
+
+def yaml_serializer(data, headers, status):
+    return yaml.dump(data)
```
wardi, line 356: this won't be able to handle objects with `for_json` methods so customizable serializers like this will "leak" into all our action calls by preventing their use.

wardi:
What's the typical use case for this feature? A client requirement for a particular API format?

ckan/tests/controllers/test_api.py
smotornyuk: I think that it's the responsibility of the serializer's author to handle `for_json` usage.

CKAN core and any extension still can/should use `for_json` because it's a feature supported by the native serializer(not documented, but supported:)). Whenever an action returns a non-standard object(neither dictionary nor list, nor the other primitive), the native serializer "somehow" handles it. So whenever a developer registers a custom serializer, it's his responsibility to check the type of returned object and handle it properly.

As I understand, `for_json` is used because `json` library can handle only around 6 object types and everything else requires extra processing. If somebody uses `bson` or `toml` serializer, it gives him the support of additional types, such as DateTime, so the chances that `for_json` is needed are reduced. Some serializers, probably, allow you to use absolutely any type without preprocessing.

Other formats, like CSV, have almost no data types at all, so even `for_json` is not enough and the developer has to handle every returned value in a special way.

So I think that ability to set a custom serializer should not change anything in the existing code. It's just a possibility, with no guarantees.

BTW, it looks like `ckanapi` doesn't use `for_json` when serializing results during CLI invocation. Probably I should create an issue inside `ckanapi`, but I have no chance to verify whether this problem is real or I just missed something.

smotornyuk:
There is no real use case or reason to add this feature. I've just changed this code because the current logic feels a bit unfair :smile: 
One can add a requrst parser so that API actions can be called using XML or MessagePack payload. But there is no simple to change the output format(other than rewriting the existing `api.action` endpoint or registering an extra route).

It's rather an idea without any additional background. I'm just wondering if it can be useful. And I don't mind closing this PR if you don't see any value in it.

ckan/tests/controllers/test_api.py
wardi: The only user of `for_json` right now is the `datastore_search` api: https://github.com/ckan/ckan/blob/a18603ec182e279488c0e91b78cd957a3ed49d30/ckanext/datastore/backend/postgres.py#L1406

Using it improves performance when the final output of the call is going to be json. Code that treats this object like a list/dict should still work, the only problem I can see is with isinstance checks in some other serializer

wardi:
I worry about adding dead code to the project because extra functionality makes it harder change core when we want to, so if you don't have a need for it I'd prefer to park this feature.

PR #6778, opened 2022-03-30
[QOL-7905] ignore auth on user API when performing a password reset, #6777

### LINKED ISSUE #6777: Password reset fails if user API is not public
Requesting a password reset uses the `ignore_auth` flag to access a user profile. However, _performing_ the reset does not, causing a failure if `ckan.auth.public_user_details` is set to False. See https://github.com/ckan/ckan/blob/fc91b296da2e6d7d6cfbb949d6616fcabe072640/ckan/views/user.py#L720

This affects both CKAN 2.8 and 2.9.

Steps to reproduce:

- Configure a CKAN instance with `ckan.auth.public_user_details = False`
- Perform a password reset for any account.

ThrawnCA:
We verify the user via token without them logging in. This is not relevant in the default configuration, but becomes necessary if 'ckan.auth.public_user_details' is set.

Fixes #6777

### Proposed fixes:

Set the `ignore_auth` flag before retrieving user dict during a password reset.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/user.py
+++ b/ckan/views/user.py
@@ -716,10 +716,12 @@ class PerformResetView(MethodView):
         except logic.NotAuthorized:
             base.abort(403, _(u'Unauthorized to reset password.'))
 
+        context[u'ignore_auth'] = True
         try:
             user_dict = logic.get_action(u'user_show')(context, {u'id': id})
         except logic.NotFound:
             base.abort(404, _(u'User not found'))
+        del context[u'ignore_auth']
         user_obj = context[u'user_obj']
         g.reset_key = request.args.get(u'key')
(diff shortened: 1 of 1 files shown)


Zharktas:
We discussed this on the tech meeting and we concluded that we do not want to it like this, it might have unforeseen consequences via side effects of `user_show`. Instead this could be improved by new API which could check existence of user and not use context['user'] which now just happens during `user_show`. If you want to secure your APIs, they should handle all the cases which the core handles as is, instead of modifying core to handle custom auth functions which aren't covered by tests.

ThrawnCA:
This isn't needed if custom auth functions are annotated with `auth_allow_anonymous_access`.

PR #6780, opened 2022-03-30
Rename a deprecated helper left out on #6764

amercader:
This is causing an exception on Recline views now:

```
ckan.exceptions.HelperError: Helper 'url' has not been defined
```


amercader:
@pdelboca an easy one for you

PR #6796, opened 2022-04-06
Fix isoformat argument for Python 2

### LINKED ISSUE #6789: Listing API tokens with CKAN CLI runs into error if token has already been used
**CKAN version**
2.9.5 (Python 2.7.13)
**Describe the bug**
When using the CKAN CLI to run `ckan user token list default` and there is an existing token that has been accessed before, then the CKAN CLI throws an exception: `TypeError: isoformat() argument 1 must be char, not unicode`.

**Steps to reproduce**
Run CKAN, create API token for user `default`, use this token at least once, run `ckan user token list default`.

**Expected behavior**
All API tokens for the `default` user should be listed.

**Additional details**
The full exception looks like this:

```
ckan@f858a3d49658:/$ ckan --config "${CKAN_CONFIG}/production.ini" user token list default
2022-04-01 15:01:44,626 INFO  [ckan.cli] Using configuration file /etc/ckan/production.ini
2022-04-01 15:01:44,626 INFO  [ckan.config.environment] Loading static files from public
2022-04-01 15:01:44,675 INFO  [ckan.config.environment] Loading templates from /usr/lib/ckan/venv/src/ckan/ckan/templates
2022-04-01 15:01:45,019 INFO  [ckan.config.environment] Loading templates from /usr/lib/ckan/venv/src/ckan/ckan/templates
Tokens([id] name - lastAccess):
        [DiJpgRiMDnPVrFU8PN9X6lO0D46n_qTzRerLvfNThDSoFZXlb3N_tCqYDMrLeuhkJ9Xz3R458TuvS54m] default_token_1648824584 - Never
  File "/usr/local/bin/ckan", line 11, in <module>
    load_entry_point('ckan', 'console_scripts', 'ckan')()
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/ckan/venv/local/lib/python2.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/ckan/venv/src/ckan/ckan/cli/user.py", line 244, in list_tokens
    accessed.second).isoformat(u" ")
TypeError: isoformat() argument 1 must be char, not unicode
```

The issue seems to be in `ckan/cli/user.py`:

```python
@token.command(u"list")
@click.argument(u"username")
def list_tokens(username):
    u"""List all API Tokens for the given user"""
    try:
        tokens = plugin.toolkit.get_action(u"api_token_list")(
            {u"ignore_auth": True}, {u"user": username}
        )
    except plugin.toolkit.ObjectNotFound as e:
        error_shout(e)
        raise click.Abort()
    if not tokens:
        click.secho(u"No tokens have been created for user yet", fg=u"red")
        return
    click.echo(u"Tokens([id] name - lastAccess):")

    for token in tokens:
        last_access = token[u"last_access"]
        if last_access:
            accessed = plugin.toolkit.h.date_str_to_datetime(last_access)
            if six.PY2:
                """
                Strip out microseconds to force formatting as isoformat doesnt
                have a timespec param on Python 2.
                """
                accessed = datetime(
                    accessed.year,
                    accessed.month,
                    accessed.day,
                    accessed.hour,
                    accessed.minute,
                    accessed.second).isoformat(u" ")
            else:
                accessed = accessed.isoformat(u" ", u"seconds")

        else:
            accessed = u"Never"
        click.echo(
            u"\t[{id}] {name} - {accessed}".format(
                name=token[u"name"], id=token[u"id"], accessed=accessed
            )
        )
```

According to [the documentation](https://docs.python.org/2.7/library/datetime.html#datetime.datetime.isoformat) `isoformat()` expects this:

> The optional argument sep (default 'T') is a one-character separator, placed between the date and time portions of the result.

Which `u" "` isn't.

frafra:
Fixes #6789

### Proposed fixes:
Remove the `u` prefix from the first argument of `isoformat`.
This is a Python 2 only patch, which should be applied to CKAN 2.9 only.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/user.py
+++ b/ckan/cli/user.py
@@ -235,13 +235,13 @@ def list_tokens(username):
                 Strip out microseconds to force formatting as isoformat doesnt
                 have a timespec param on Python 2.
                 """
-                accessed = datetime(
+                accessed = str(datetime(
                     accessed.year,
                     accessed.month,
                     accessed.day,
                     accessed.hour,
                     accessed.minute,
-                    accessed.second).isoformat(u" ")
+                    accessed.second))
             else:
                 accessed = accessed.isoformat(u" ", u"seconds")
 


ThrawnCA:
The lint checker objects to the non-unicode literal string...is there a way to mark that line as an exception?

ThrawnCA:
https://github.com/qld-gov-au/ckan/pull/47/commits/c2f5fa4e5934e89ac1469fb2516896e53ec19ecf fixes it on our fork.

ckan/cli/user.py
```diff
@@ -241,9 +241,9 @@ def list_tokens(username):
                     accessed.day,
                     accessed.hour,
                     accessed.minute,
-                    accessed.second).isoformat(u" ")
+                    accessed.second).isoformat(" ")
```
Zharktas, line 244:
```suggestion
                    accessed.second).isoformat(b" ")
```

ckan/cli/user.py
```diff
@@ -241,9 +241,9 @@ def list_tokens(username):
...
             else:
-                accessed = accessed.isoformat(u" ", u"seconds")
+                accessed = accessed.isoformat(" ", u"seconds")
```
Zharktas, line 246:
```suggestion
                accessed = accessed.isoformat(b" ", u"seconds")
```

Zharktas (commented):
Adding b' ' literal instead probably fixes linting issue, bytes literals in python2 are converted to unicode in python3.

Zharktas:
My suggestion actually didn't work on python3 :|

ckan/cli/user.py
Zharktas, line 244:
```suggestion
                    accessed.second).isoformat(r" ")
```

ckan/cli/user.py
Zharktas, line 246:
```suggestion
                accessed = accessed.isoformat(r" ", u"seconds")
```

Zharktas (commented):
Prefixing the with r should work in both python2 and 3

frafra:
The linter will always fail: `Test that string literals are prefixed by ``u``, ``b`` or ``ur``.`. None of them is suitable for this case.
Better use `str(...)`.

Zharktas:
that ur'' check is actually quite useless, as ur'' was removed from python in 3.4 :D

frafra:
Python 2:

````python
>>> from datetime import datetime
>>> accessed = datetime.today()
>>> accessed = str(datetime(
...     accessed.year,
...     accessed.month,
...     accessed.day,
...     accessed.hour,
...     accessed.minute,
...     accessed.second))
>>> accessed
'2022-04-07 10:37:50'
```

Check passed.
Ready to be merged :)

Zharktas:
We had a quick chat about this in the tech meeting and thought that we backpedal a bit and just prefix it with b'' since it already is  in ```if six.PY2``` block, we don't have to have py3 compatible code.

wolfm89:
@Zharktas will this be fixed with this PR or did you open another one because you changed the approach of the solution?

PR #6798, opened 2022-04-06
Paginate the activity list

### LINKED ISSUE #6108: Fix activity pagination in web UI
**CKAN version**: 2.9

**Describe the bug**

"Activity Stream" pages only show the first `activity_list_limit` activities.
In some previous version of CKAN there would be a "load more" button linking to the next page of activities, but the list just ends once there are more than `activity_list_limit` activities:

with less than  `activity_list_limit` activities:
Screenshot at 2021-05-20 11-20-29

with more than  `activity_list_limit` activities:
Screenshot at 2021-05-20 11-21-46

There is some code in https://github.com/ckan/ckan/blob/ckan-2.9.2/ckan/templates/activity_streams/activity_stream_items.html which does look like it should implement the 'Load more' pagination functionality, but I don't think it is currently used/working.

**Steps to reproduce**

Create more than `activity_list_limit` activities on a dataset/org/etc
Look at the "Activity Stream" page

**Expected behavior**

There should be a "Load More" button linking to the next page of activities.
The pagination should respect the value of `activity_list_limit`.

avdata99:
Fixes #6108 

pkg_activity_list

### Proposed fixes:

This PR:
 - Completely change the JS module `activity-stream.js`. The `*_activity_list_html` actions were removed at #4627 and this file should be removed
 - Remove template with the _Load more_ button (no longer in use)
 - Allow pagination over activity lists adding the _next_ and _previous_ page buttons
 - Allow filter activities by type

If this is correct we can implement same changes for other activity streams

### Features:

- [x] includes API changes


wardi:
for performance it would be much better to use timestamps to paginate activities instead of offsets. The activity table will be very large on busy sites and a web crawler could inadvertently create a DoS when using offset.

ckan/views/dataset.py
```diff
@@ -1108,21 +1117,42 @@ def activity(package_type: str, id: str) -> Union[Response, str]:  # noqa
...
         dataset_type = pkg_dict[u'type'] or u'dataset'
     except NotFound:
         return base.abort(404, _(u'Dataset not found'))
     except NotAuthorized:
         return base.abort(403, _(u'Unauthorized to read dataset %s') % id)
 
-    # TODO: remove
-    g.pkg_dict = pkg_dict
-    g.pkg = pkg
+    all_activities = activity_model.package_activity_list(
```
amercader, line 1146: This looks like a really expensive way of calculating the total number of activities. Maybe we can create a new `package_activity_count` function  that accepts the same params but just returns the count needed, computed in an efficient way

amercader:
@wardi This is how activity lists are paginated now:

https://github.com/ckan/ckan/blob/636135fadb79f107fd875082d06bc6a1e3b1333e/ckan/model/activity.py#L136-L140

Essentially we pass `?offset=10&limit=20` 

What you are suggesting is paginating with  something like `?since=<timestamp>&limit=10`?
Would we need some extra index in the activity table?

wardi:
Offset by timestamp also has the benefit of having persistent urls and not having duplicate items show up while moving through pages when new activities are added at the same time

ckan/views/dataset.py
avdata99: I added the `package_activity_count` [here](https://github.com/avdata99/ckan/commit/cc146984c2b4762ea20bcfc03a3e37c244330194) @amercader

avdata99:
> @avdata99 @amercader This is good to go, but we need to check if we shoudl merge this first or #6790
> 
> My suggestion is to merge this one first

@tino097 @amercader, I agree to merge this one first.
But we need to consider other activity streams (e.g. user, groups and others).
Do we want to apply this to all of them before #6790?

amercader:
So we could approach this in two ways:

**Option A**

1. Implement timestamp based pagination on this PR (for datasets)
2. Replicate the changes on this PR to all activity streams (user, group/org, dashboard) on a separate PR
3. Port the changes to the new activities plugin PR (#6790)

or **Option B**

1. Merge the new activities plugin PR (#6790)
2. Port the changes of this PR to the activities plugin
3. Implement timestamp based pagination on this PR (for datasets) on the activities plugin
4. Replicate the changes on this PR to all activity streams (user, group/org, dashboard) on the activities plugin

I don't have any strong opinions either way, @smotornyuk you probably have a better idea of the effort involved and are best placed to decide

smotornyuk:
I prefer **option A**. For me, it will be as simple as picking a few extra changes into a separate plugin(i have plenty of macros, so it will be a trivial task). And I don't want @avdata99  to repeat his work:)

After merging this PR, if we go with #6790 next, it will be even easier for @avdata99 to replicate the changes on this PR to all activity streams.

BTW, starting the next week I'll stop my volunteer activities(as the situation is more or less stabilized and I'm not really useful anymore). This means, I'll have more time for CKAN and will get back to the TechTeam meetings:)

avdata99:
Thanks @smotornyuk, I agree to go for this PR first, then move this feature to this PR and finally create independent PRs for other activity streams.

avdata99:
@smotornyuk @amercader 
I finished a first version of pagination by timestamp (including some basic tests).
Some notes:
 - If we expect duplicate timestamps (I think we don't), we'll need to change this approach to one that includes [timestamp+id approach](https://phauer.com/2018/web-api-pagination-timestamp-id-continuation-token/)
 - For the _previous page_ I'm [still using `offset`](https://github.com/avdata99/ckan/blob/6108_improve_pkg_actvity_list/ckan/model/activity.py#L157-L158), is there a better way to do it?

ckan/views/dataset.py
```diff
@@ -1103,12 +1103,17 @@ def get(self, package_type: str, id: str) -> str:
 def activity(
     package_type: str,
     id: str,
-    page: Union[int, str] = 0,
-    activity_type: Optional[str] = None,
 ) -> Union[Response, str]:  # noqa
 
     """Render this package's public activity stream page.
     """
+    after = h.get_request_param(u'after')
```
amercader, line 1108: Don't introduce new `u` prefixes please :pray:

ckan/views/dataset.py
```diff
@@ -1097,9 +1098,22 @@ def get(self, package_type: str, id: str) -> str:
...
+    before = h.get_request_param(u'before')
+    activity_type = h.get_request_param(u'activity_type')
+
+    if after and before:
+        raise ValidationError(
+            {'after': ['Cannot be used together with `before']}
+        )
+
```
amercader, line 1116: These checks should be done via schema validators, not directly in the blueprints (otherwise they are not applied when using the API)

ckan/views/dataset.py
```diff
@@ -1108,21 +1122,78 @@ def activity(package_type: str, id: str) -> Union[Response, str]:  # noqa
...
     data_dict = {u'id': id}
+    base_limit = int(config.get(u'ckan.activity_list_limit', 30))
+    max_limit = int(config.get(u'ckan.activity_list_limit_max', 100))
+    limit = min(base_limit, max_limit)
+    activity_types = [activity_type] if activity_type else None
+    if after is not None:
+        after = datetime.fromtimestamp(float(after))
+    if before is not None:
+        before = datetime.fromtimestamp(float(before))
+    is_page_1 = after is None and before is None
```
amercader, line 1133:
```suggestion
    is_first_page = after is None and before is None
```

ckan/logic/action/get.py
```diff
@@ -2533,14 +2533,17 @@ def package_activity_list(
 
     :param id: the id or name of the package
     :type id: string
-    :param offset: where to start getting activity items from
```
amercader, line 2536: @wardi was your idea to remove support for `offset` pagination entirely or keep it at the API level and just not use it from the UI?

ckan/views/dataset.py
```diff
@@ -1108,21 +1122,78 @@ def activity(package_type: str, id: str) -> Union[Response, str]:  # noqa
         u'auth_user_obj': g.userobj
     })
     data_dict = {u'id': id}
+    base_limit = int(config.get(u'ckan.activity_list_limit', 30))
+    max_limit = int(config.get(u'ckan.activity_list_limit_max', 100))
```
amercader, line 1126: Since https://github.com/ckan/ckan/pull/6467, if you have declared a config option it will automatically get validated and parsed (and the defaults used) if you use `config.get_value()`

```suggestion
    base_limit = config.get_value('ckan.activity_list_limit')
    max_limit = config.get_value('ckan.activity_list_limit_max')
```

ckan/views/dataset.py
```diff
@@ -1108,21 +1122,78 @@ def activity(package_type: str, id: str) -> Union[Response, str]:  # noqa
...
+    limit = min(base_limit, max_limit)
+    activity_types = [activity_type] if activity_type else None
+    if after is not None:
+        after = datetime.fromtimestamp(float(after))
```
amercader, line 1130: You should handle users (or rather scrapers) passing a non-float here

ckan/logic/schema.py
```diff
@@ -656,6 +656,8 @@ def default_activity_list_schema(
         ignore_missing, ignore_not_sysadmin, boolean_validator]
     schema['activity_types'] = [ignore_missing, list_of_strings]
     schema['exclude_activity_types'] = [ignore_missing, list_of_strings]
+    schema['before'] = [ignore_missing]
+    schema['after'] = [ignore_missing]
```
amercader, line 660: These should be validated as timestamps, and as mentioned before if you want to check that both are not present do it here.

ckan/tests/controllers/test_package.py
```diff
@@ -2051,6 +2052,77 @@ def test_create_dataset(self, app):
         assert dataset["id"] in href.select_one("a")["href"].split("/", 2)[-1]
         assert dataset["title"] in href.text.strip()
 
+    @pytest.mark.ckan_config("ckan.activity_list_limit", "3")
```
amercader, line 2055: These are great but can you add a couple more tests to ensure that:
* Navigating to the activity page using `before` and `after` actually works
* Passing a non-timestamp value returns an error (or ignores the params)
* Previous button is not shown in the first page
* Next button is not shown in the last page

ckan/templates/package/activity.html
```diff
@@ -8,5 +8,51 @@ <h1 class="hide-heading">
...
+  
+  <div id="activity_page_buttons" style="margin-top: 25px;">
+
+  {% if prev_page %}
+    <a href="{{ prev_page }}" class="btn btn-default btn-rounded" >{{ _('Previous page') }}</a>
+  {% endif %}
+
+  {% if next_page %}
+      <a href="{{ next_page }}" class="btn btn-default btn-rounded">{{ _('Next page') }}</a>
+  {% endif %}
```
amercader, line 55: I found the "Previous page" and "Next page" buttons confusing, as when you click on previous you are actually requesting Newer (more recent) activities. Maybe is a bit more intuitive to use "Newer" and "Older" explicitly (and let's not use `btn-rounded` as it is not used elsewhere):

```suggestion
  {% if prev_page %}
    <a href="{{ prev_page }}" class="btn btn-default" >{{ _('Newer activities') }}</a>
  {% endif %}

  {% if next_page %}
      <a href="{{ next_page }}" class="btn btn-default">{{ _('Older activities') }}</a>
  {% endif %}
```

ckan/templates/package/activity.html
```diff
@@ -8,5 +8,51 @@ <h1 class="hide-heading">
...
+
+  {# allow extensions to continue using this template without new vars #}
+  {% if activity_types is defined %}
+  <div 
+    id="activity_types_filter"
+    style="margin-bottom: 15px;"
+    data-module="activity-stream"
+    >
+    <label for="activity_types_filter_select">Activity type</label>
+    <select id="activity_types_filter_select">
```
amercader, line 20: We really should not show "package" to the end user, specially if other activities have dataset in it. And they are not translatable.
So maybe instead of using the keys with `.title()` we need to somehow use the new [humanize](https://github.com/ckan/ckan/blob/c00891b0a35c7c334701bbcb56cfad61dbfedabe/ckan/lib/helpers.py#L962) functions to get a proper label

Screenshot from 2022-05-03 12-49-02

ckan/templates/package/activity.html
```diff
@@ -8,5 +8,51 @@ <h1 class="hide-heading">
...
+  </div>
+  {% endif %}
+
+  {% if activity_stream|length > 0 %}
+    {% snippet 'snippets/activity_stream.html', activity_stream=activity_stream, id=id, object_type='package' %}
+  {% else %}
+    <p>
+      {{ _('No activity found') }}
+      {% if activity_type %}
+        {{ _('for this type') }}.
```
amercader, line 42: Don't split sentences, otherwise it makes translating into other languages difficult

```suggestion
      {% if activity_type %}
        {{ _('No activity found for this type'') }}
      {% else %}        
        {{ _('No activity found') }}.
      {% else %}        
```

amercader (commented):
@avdata99 looking good! I added a few more comments to polish the feature a bit

ckan/views/dataset.py
avdata99: Changed [here](https://github.com/ckan/ckan/commit/da28b9a4bf17f2e91802938444eb9c466f55bc66).

ckan/logic/action/get.py
wardi: Do we know how many users of the activity API we have?

My guess is we would want to keep it for compatibility, at least for a while.

ckan/model/activity.py
```diff
@@ -136,7 +141,21 @@ def _activities_limit(q: QActivity, limit: int,
...
+
     if limit:
+        # if we want activities after some timestamp, we need to limit
+        # last N rows from this query
+        if after:
+            total = q.count()
+            offset = total - limit
+            # going back until the first page, we don't need to offset
+            if offset > 0:
+                q = q.offset(offset)
```
wardi, line 158: instead of getting all results when using then applying an offset, use an ascending `order_by` and then reverse the results returned.

ckan/model/activity.py
```diff
@@ -158,7 +177,27 @@ def _activities_at_offset(q: QActivity,
...
+
+
+def _activities_before(q: QActivity,
+                      limit: int,
+                      before: datetime.datetime
+    ) -> list[Activity]:
+    '''
+    Return a list of all activities before a timestamp.
+    '''
+    return _activities_limit(q, limit, before=before).all()
```
wardi, line 200: what do we need these 1-line functions for?

wardi:
Some thoughts about this API design:

Technically we could allow `before` + `after` at the same time if we want to allow users to request a range of activities. We could decide that when both are provided the `limit` parameter limits from the `before` side to eliminate ambiguity.

If we re-enable the `offset` parameter in this PR do we prevent its use when `before` or `after` are given with a validation error or do we do something else? Should we add a configuration option to set a maximum value for `offset` to allow sites to phase out its use?

ckan/views/dataset.py
avdata99: Updated [here](https://github.com/avdata99/ckan/commit/b1a86bfea67d869b55022bec0936c64665bb63e4)

ckan/templates/package/activity.html
avdata99: This is a simple proposal to humanize all activities: https://github.com/avdata99/ckan/commit/7a55c0e9a564da110dec2092f24a05bc4b77a2c1
The ´humanize_entity_type´ function doesn't fit exactly in what we need but we can use it @amercader

ckan/logic/action/get.py
avdata99: The `offset` para was restored. Works well in conjunction with `after` and `before` (and not used for pagination).
Also, some test were added

ckan/views/dataset.py
avdata99: A new validator was created [here](https://github.com/avdata99/ckan/commit/42975b649e9370a5827c7aa148223991fd588281). 
Now we allow after and before together.

ckan/views/dataset.py
avdata99: Done

ckan/logic/schema.py
avdata99: Done [here](https://github.com/avdata99/ckan/commit/a3173276cbccd31bfe4c29d9490c40d5dee9cdbe)

ckan/model/activity.py
avdata99: We now [reverse the order](https://github.com/avdata99/ckan/blob/6108_improve_pkg_actvity_list/ckan/model/activity.py#L143-L147) when we [look for `before` without `after`](https://github.com/avdata99/ckan/blob/6108_improve_pkg_actvity_list/ckan/model/activity.py#L266-L272)

ckan/model/activity.py
avdata99: They are useful to reuse between packages, users, group for activities. I delete the 2 I create.

avdata99:
Last changes:
 - Validator added
 - Test added
 - after and before works together
 - offset can be used again (also combined with after and before)

ckan/model/activity.py
```diff
@@ -517,6 +545,24 @@ def _filter_activitites_from_users(q: QActivity) -> QActivity:
...
+    '''Adds a filter to an existing query object to include or exclude
+    (include=False) activities based on a list of types.
+
+    '''
+    import ckan.model as model
+    if after:
+        q = q.filter(model.Activity.timestamp > timestamp)
+    else:
+        q = q.filter(model.Activity.timestamp < timestamp)
+    return q
```
wardi, line 563: This kind of 1-line function shouldn't exist because it makes calling code more difficult to understand. Instead please use the line of code itself.

ckan/model/activity.py
```diff
@@ -240,7 +256,19 @@ def package_activity_list(
...
 
-    return _activities_at_offset(q, limit, offset)
+    if after:
+        q = _filter_activitites_from_timestamp(q, timestamp=after, after=True)
+    if before:
+        q = _filter_activitites_from_timestamp(q, timestamp=before, after=False)
+
+    # revert sort queries for "only after" queries
+    revese_order = before and not after
+    results = _activities_limit(q, limit, offset, revese_order).all()
```
wardi, line 266: instead of calling `_activities_limit` please use `order_by`, `offset` and `limit` here. We should aim to remove functions like `_activities_limit` because they make the code harder to understand and optimize.

wardi:
@avdata99 thank you for your work on this. I suggested a few other small changes above to reduce the number of functions that hide really simple operations and to make the code easier to understand and improve in the future.

ckan/model/activity.py
avdata99: updated [here](https://github.com/avdata99/ckan/commit/64443b263186615ffda09c0cc5c155690d278442)

avdata99:
@wardi thanks for your review
I updated the activity queries [here](https://github.com/avdata99/ckan/commit/64443b263186615ffda09c0cc5c155690d278442).

wardi:
@avdata99 thank you!

amercader:
This is good to go, but let's wait on https://github.com/ckan/ckan/pull/6790#issuecomment-1128707195

amercader:
@avdata99 thanks for your work on this. @smotornyuk will port this changes to #6790 and once that is merged you can replicate the changes to the user and group streams

PR #6791, opened 2022-04-04
initialize bootstrap.Popover

TomeCirun:
Fixes #6784

### Proposed fixes:
bootstrap5 popovers no longer work as they were initialized previously in `bs3`. 
In bs5, one way of doing this is: 
`var popover = new bootstrap.Popover(document.querySelector('.example-popover'), {
  options
})`

image
image

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #6800, opened 2022-04-07
[6793]Fix datastore_upsert update not working with index _id

### LINKED ISSUE #6793: Datastore upsert update not working with index _id
**CKAN version** 2.9.5

**Describe the bug**
datastore/upsert method=update does not work with _id 

**Steps to reproduce**
Create resource in datastore
insert values
read values => resulting records have unique field _id, as expected
update command with _id fails on "undefined primary_key"
{
    "error": {
        "__type": "Validation Error",
        "table": [
            "table does not have a unique key defined"
        ]
    },

**Expected behavior**
Record number _id can be used as primary key, as in regular SQL database

**Additional details**
If possible, please provide the full stack trace of the error raised,  or add screenshots to help explain your problem.

Use case: update resource with timeseries data which by design do not come with an inherent primary key. So primary key must be database id. However, this cannot be enforced on datastore_create()

Maybe you consider this not a bug - but we have an id in the data, so not allowing to use it is quite close ...

TomeCirun:
Fixes #6793 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -261,7 +261,6 @@ def _get_unique_key(context: Context, data_dict: dict[str, Any]) -> list[str]:
         AND a.attnum = ANY(idx.indkey)
         AND t.relkind = 'r'
         AND idx.indisunique = true
-        AND idx.indisprimary = false
         AND t.relname = %s
     '''
     key_parts = context['connection'].execute(sql_get_unique_key,
@@ -1166,8 +1165,10 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
             raise ValidationError({
                 'table': [u'table does not have a unique key defined']
             })
-
         for num, record in enumerate(records):
+            # remove the _id if not provided when calling update method
+            if '_id' in unique_keys and '_id' not in record:
+                unique_keys.remove('_id')
             # all key columns have to be defined
             missing_fields = [field for field in unique_keys
                               if field not in record]
@@ -1184,6 +1185,9 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
                     # a tuple with an empty second value
                     record[field['id']] = (json.dumps(value), '')
 
+            if "_id" in record:
+                # _id is not a field_name so we need to append
+                field_names.append('_id')
             non_existing_filed_names = [field for field in record
                                         if field not in field_names]
             if non_existing_filed_names:


TomeCirun:
Hey, @wardi 
I'm not sure what's happening with the tests, the one from `Cypress / cypress (pull_request)`, we fixed in #6791, also, I've tested the one that failing in  `ci/circleci: test-python-3` it passes locally, however, I'm not sure that it's related to my changes, and regarding `@check_types` I think @smotornyuk can help as they behave weird in my latest PR's.

Thanks

smotornyuk:
It looks like we have `pyright` updated recently and it now can detect type incompatibilities much better:) I'll take a look at it and provide a fix in a separate PR. 
For anyone who is going to review this PR, I think it's safe to ignore type check as long as we manage to fix it before release 2.10. But, if I'll finish the fix in the next couple of days, I'll mention this PR, so that you can rebase it.

As for `test-python-3`, it looks like one of our tricky stateful tests. Basically, some tests produce nondeterministic results depending on the test execution order. That's why you cannot reproduce this error locally: either you run the single tests, or run all the tests but they are executed in a different order. I'll also try to find out what exactly causes the problem. But at the moment I don't have much time, so if you find any workaround for the problem, I'll help a lot.

ckanext/datastore/backend/postgres.py
```diff
@@ -261,7 +261,6 @@ def _get_unique_key(context: Context, data_dict: dict[str, Any]) -> list[str]:
         AND a.attnum = ANY(idx.indkey)
         AND t.relkind = 'r'
         AND idx.indisunique = true
-        AND idx.indisprimary = false
```
wardi, line 264: this function is also used in `create_indexes` so this change would break that functionality. Better to only modify `upsert_data` instead

ckanext/datastore/backend/postgres.py
```diff
@@ -1166,8 +1165,10 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
             raise ValidationError({
                 'table': [u'table does not have a unique key defined']
             })
-
         for num, record in enumerate(records):
+            # remove the _id if not provided when calling update method
+            if '_id' in unique_keys and '_id' not in record:
+                unique_keys.remove('_id')
```
wardi, line 1171: If `_id` is the only key in `unique_keys` then this would allow "upserts" without an `_id` specified in `record`, right?

wardi:
Thanks for your work on this @TomeCirun . We should decide on the behaviour that we want with and without primary keys given. How about:

- when table has no primary key
  - `_id` must be provided for update/upsert method of `datastore_upsert`
- when table has a single or composite primary key
  - primary keys must always be provided for update/upsert
  - `_id` may be provided for update/upsert to allow replacing primary key values in-place

datastore_upsert should have documentation that describes the expected behaviour when using `_id`

each of the above cases also needs a test and we should have tests covering invalid usage like some or all primary keys missing with or without `_id` for tables with primary keys.

wardi:
I've submitted a different PR to address this issue #6933

PR #6811, opened 2022-04-16
Set absolute URL for using by who-friend behind SSL proxy

### LINKED ISSUE #4129: Login over HTTPS redirects to HTTP
CKAN 2.7.2 (but present on master) running with Docker

When running CKAN over HTTPS with a correctly specified CKAN_SITE_URL, users can login/logout without issue.

When running over HTTPS and attempting to login/logout, the service hangs and then bails out. After investigation, it seems the current usage of repoze.who used for the user login does not support https. Closest I got to understanding the problem is this discussion https://groups.google.com/forum/#!topic/pylons-discuss/u_YCE-6rgkw 

I suspect most people who run CKAN over HTTPS don't run it through Docker and as a result have an nginx configuration that masks this problem. We are running it with Docker on an AWS stack and had not set up nginx. 

As a workaround we added a small nginx service that listens on a port and redirects all requests to the https equivalent. Then we made the Load Balancer redirect HTTP port 80 to this service. 

Run CKAN with docker-compose over HTTPS

Sheile:
Fixes #4129

### Proposed fixes:
If anyone build docker container behind SSL proxy such as AWS ALB, redirect to HTTP insted of HTTPS regardless of `ckan.site_url` after login.

who-friendly use an exception in WebOb to redirect to `post_login_url`.
But [_HTTPMove exception](https://github.com/Pylons/webob/blob/master/src/webob/exc.py#L591) does not support `X-Forwarded-Proto` header.

`req.path_url` has HTTP protocol because already terminated SSL in AWS ALB or reverse proxy.
Therefore, joined URL has HTTP protocol but that URL is not valid to connect with SSL.

### Features:

- [ ] includes API changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,6 +65,8 @@ RUN ckan-pip3 install -U pip && \
     ckan-pip3 install --upgrade --no-cache-dir -r $CKAN_VENV/src/ckan/requirement-setuptools.txt && \
     ckan-pip3 install --upgrade --no-cache-dir -r $CKAN_VENV/src/ckan/requirements.txt && \
     ckan-pip3 install -e $CKAN_VENV/src/ckan/ && \
+    sed -i "s|post_login_url = /user/logged_in|post_login_url = ${CKAN_SITE_URL}/user/logged_in|" $CKAN_VENV/src/ckan/ckan/config/who.ini && \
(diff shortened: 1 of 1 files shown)


pdelboca:
Closing as our docker file has been migrated to it's own repository. Thanks!

PR #6812, opened 2022-04-17
Add healthchecks

frafra:
Fixes #6489

### Proposed fixes:
Add health checks is beneficial for monitoring as well as dependency management between containers.
Additional CKAN commands (like ckanext-harvest tasks) can rely on the status of the main CKAN container, waiting for it to be ready, instead of being launched at the same time (thus triggering databases issues).
I also removed the `links` section, as this does nothing, since Docker compose creates a new network by default.

### Features:

- [ ] includes API changes

### DIFF
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -14,12 +14,13 @@ services:
     container_name: ckan
     build:
       context: ../../
-    links:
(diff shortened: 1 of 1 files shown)


kowh-ai (approved):
Looks great @frafra - thanks for this

PR #6821, opened 2022-04-26
Small documentation fixes

smotornyuk:
* Add [missing docstrings inside toolkit](http://docs.ckan.org/en/latest/extensions/plugins-toolkit.html#ckan.plugins.toolkit.ckan.plugins.toolkit.h)
* [Fix the order of versions](https://github.com/ckan/ckan/compare/2.10-version-in-docs?expand=1#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624L136). Previously versions were sorted alphabetically and `2.10.0` was reported as "no longer supported", because it's smaller that `2.8.0`:)
* Some tiny changes in the documentation
* [Correct reference to the current minor version](https://github.com/ckan/ckan/compare/2.10-version-in-docs?expand=1#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624L264). At the moment documentation takes first three characters of the full version(`2.9.5` -> `2.9`) in order to produce a reference to debian package/latest documentation/etc. But in case of `2.10` it will not work: `"2.10.0"[:3]` -> `2.1`

### DIFF
--- a/ckan/cli/__init__.py
+++ b/ckan/cli/__init__.py
@@ -106,6 +106,8 @@ class CKANConfigLoader(object):
 
 
 def error_shout(exception: Any) -> None:
+    """Report CLI error with a styled message.
+    """
     click.secho(str(exception), fg=u'red', err=True)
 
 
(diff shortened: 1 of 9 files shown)


amercader:
Great stuff @smotornyuk

PR #6823, opened 2022-04-26
Feature/license validator

JVickery-TBS:
Fixes #

### Proposed fixes:

Added a core validator for licenses. Validates if the passed license ID exists in the registered licenses from the licenses_group_url setting.

Added a test for v1 and v2 of the licenses group json structure.

### Features:

- [ ] includes API changes
- [X] none of the above


EricSoroos:
Looks reasonable.

What is the policy for changing default schemas WRT validation, so that existing datasets don't validate anymore?  This is something that has to be done very carefully or automated processes (e.g. harvesters/datapusher) fail.  May want to put this behind a feature flag that defaults to on for new installs but not for existing ones.

PR #6830, opened 2022-04-28
skip updating package 'metadata_modified' when only updating resources

### LINKED ISSUE #6829: Package modified date should not change when resources are reordered
**CKAN version**

2.9

**Describe the bug**

Package "metadata_modified" timestamp should not change when resources are reordered.

**Steps to reproduce**

* Create a dataset with multiple resources
* Swap the order of the two resources
* Check the "metadata modified" timestamp for the dataset and verify that it was updated when the order changed.

**Expected behavior**

"Metadata modified" should not change on resource reorder.

**Additional details**

Will open a pull request soon.

ThrawnCA:
This includes the resource_update API and the package_resource_reorder API.

Fixes #6829

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -110,6 +110,7 @@ def resource_update(context: Context, data_dict: DataDict) -> ActionResult.Resou
 
     try:
         context['use_cache'] = False
+        context['resources_only'] = True
         updated_pkg_dict = _get_action('package_update')(context, pkg_dict)
     except ValidationError as e:
         try:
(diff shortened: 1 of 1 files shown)


duttonw:
👍

ThrawnCA:
I'm unclear on why this change would break the test that's failing. The only behaviour that's altered is conditionally preventing a Package model object from being updated, and I can't see why that would prevent `datapusher_submit` from being triggered. The Datapusher plugin isn't even listening for package events.

PR #6832, opened 2022-04-28
Improve error when downloading resource

pdelboca:
This PR handles `NotAuthorized` exception separately when trying to download a resource.

Currently, the error message will be `Not Found 404` even when the resource exist but the user doesn't have access to it.

### DIFF
--- a/ckan/views/resource.py
+++ b/ckan/views/resource.py
@@ -176,8 +176,10 @@ def download(package_type: str,
     try:
         rsc = get_action(u'resource_show')(context, {u'id': resource_id})
         get_action(u'package_show')(context, {u'id': id})
-    except (NotFound, NotAuthorized):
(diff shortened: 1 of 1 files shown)


ckan/views/resource.py
```diff
@@ -176,8 +176,10 @@ def download(package_type: str,
     try:
         rsc = get_action(u'resource_show')(context, {u'id': resource_id})
         get_action(u'package_show')(context, {u'id': id})
-    except (NotFound, NotAuthorized):
+    except NotFound:
         return base.abort(404, _(u'Resource not found'))
+    except NotAuthorized:
+        return base.abort(401, _(u'Not authorized to download resource'))
```
amercader, line 182: We use 403 for NotAuthorized errors

PR #6836, opened 2022-04-29
6833 auth function for status show

### LINKED ISSUE #6833: Action status_show has no corresponding auth function
**CKAN version**

2.9.5

**Describe the bug**
The Action-API function `status_show` does not call a corresponding auth function (there is none). This means that, if a plugin _does_ register an auth function for `status_show`, there will be an error when the action is called.

**Steps to reproduce**
* Implement IAuthFunctions in a plugin and add `status_show` to `get_auth_functions()`.
* Call `status_show` as a non-sysadmin user.
* This leads to the exception above.

**Expected behavior**
If a plugin _does_ registers an auth function for `status_show`, I expect it to be called when the action is called.

**Additional details**
Stacktrace:

```
2022-04-28 08:06:19,368 ERROR [ckan.views.api] Action function status_show did not call its auth function
  File "/usr/lib/ckan/default/src/ckan/ckan/config/middleware/../../views/api.py", line 292, in action
  File "/usr/lib/ckan/default/src/ckan/ckan/logic/__init__.py", line 511, in wrapped
    raise Exception(
Exception: Action function status_show did not call its auth function
```

Action function:
https://github.com/ckan/ckan/blob/ca537ad25e13320b6fbf91f41031eadc30500f12/ckan/logic/action/get.py#L2406

knudmoeller:
Fixes #6833 

### Proposed fixes:

- add an auth function for `status_show`
- call new auth function from `status_show` action function

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/logic/auth/test_get.py
+++ b/ckan/tests/logic/auth/test_get.py
@@ -621,3 +621,11 @@ class TestFollowee:
         sysadmin = factories.Sysadmin()
         context = {"user": sysadmin["name"], "model": model}
         assert helpers.call_auth(func, context=context)
+
+
+@pytest.mark.usefixtures("non_clean_db")
+class TestStatusShow:
+    
+    def test_status_show_is_visible_to_anonymous(self):
+        context = {"user": "", "model": model}
+        assert helpers.call_auth("status_show", context)
--- a/ckan/tests/logic/auth/test_get.py
+++ b/ckan/tests/logic/auth/test_get.py
@@ -625,7 +625,7 @@ class TestFollowee:
 
 @pytest.mark.usefixtures("non_clean_db")
 class TestStatusShow:
-    
+
     def test_status_show_is_visible_to_anonymous(self):
         context = {"user": "", "model": model}
         assert helpers.call_auth("status_show", context)
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2433,6 +2433,7 @@ def status_show(context: Context, data_dict: DataDict) -> ActionResult.StatusSho
     :rtype: dictionary
 
     '''
+    _check_access('status_show', context, data_dict)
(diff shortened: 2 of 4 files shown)


duttonw:
👍

ckan/tests/logic/auth/test_get.py
```diff
@@ -621,3 +621,11 @@ def test_sysadmin_can_list_followees(self, func):
         sysadmin = factories.Sysadmin()
         context = {"user": sysadmin["name"], "model": model}
         assert helpers.call_auth(func, context=context)
+
+
+@pytest.mark.usefixtures("non_clean_db")
+class TestStatusShow:
+    
+    def test_status_show_is_visible_to_anonymous(self):
+        context = {"user": "", "model": model}
```
ThrawnCA, line 630: Can there also be a test with no `user` object in the context at all?

knudmoeller: I can add it if necessary. However, if I'm not mistaken, there isn't a single test in `auth/test_get.py` with no `user` attribute in the context object.

ThrawnCA: It sounds like it would be a good thing to have at least once. I'm pretty sure it won't be a problem, since anything in `get.py` automatically gets the effect of the `auth_allow_anonymous_access` annotation, but that should be tested.

knudmoeller: It turns out that `call_auth()` has this: 

https://github.com/ckan/ckan/blob/802ec31ccc0d7122fab73a6d84f5d2739fc6d0f8/ckan/tests/helpers.py#L159-L161

So a test with no `user` in `context` won't work.

Zharktas: Most tests either use `"user": None` or `"user": ""`, it really doesn't matter which is used as there is no clear convention.

PR #6837, opened 2022-04-29
Check for metadata_modified in update.py

roc:
* allow passing of `metadata_modified` value to package update/patch
* if no `metadata_modified` value passed, writes `datetime.datetime.utcnow()` as before

Currently `"metadata_modified"` is always set to `datetime.datetime.utcnow()`
even when a `"metadata_modified"` value is passed in the `data_dict`.

This means updates from the API will always overwrite the value with a
timestamp from the time of the call.

In most cases this makes sense, but if a user wants to update some small
piece of information retroactively, it inaccurately represents the time
that the information in the package was last set.

See https://github.com/ckan/ckan/issues/6829 for more

Fixes #

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -320,9 +320,16 @@ def package_update(
         model.Session.rollback()
         raise ValidationError(errors)
 
-    #avoid revisioning by updating directly
+    # check for metadata_modified
+    if data_dict.get("metadata_modified"):
+        mmod = datetime.datetime.strptime(data_dict.get(
+            "metadata_modified"),  "%Y-%m-%dT%H:%M:%S.%f")
+    else:
+        mmod = datetime.datetime.utcnow()
+
+    # avoid revisioning by updating directly
     model.Session.query(model.Package).filter_by(id=pkg.id).update(
-        {"metadata_modified": datetime.datetime.utcnow()})
+        {"metadata_modified": mmod})
     model.Session.refresh(pkg)
 
     pkg = model_save.package_dict_save(data, context)


Zharktas:
ckanext-harvest relies on metadata_modified not being modified externally https://github.com/ckan/ckanext-harvest/blob/e2e8acbb6d87d44663e92c5ff5cbd67401588c2d/ckanext/harvest/harvesters/ckanharvester.py#L233

wardi:
@Zharktas updating metadata_modified is intended for sites that *are* harvesting data, not sites *being* harvested. It's good enough to include a warning about the possible side effect of making ckanext-harvest not work against your site if you use this feature.

Another option for ckanext-harvest is to get changes to packages by paginating through the package activity stream. This is what we've done for years to publish datasets from one internal ckan instance to an external one. With some simple de-duplication of package ids it works very well (we also added a timestamp-index-paginated version of `changed_packages_activity` for this purpose, would be nice to have in core)

ckan/logic/action/update.py
```diff
@@ -320,9 +320,16 @@ def package_update(
         model.Session.rollback()
         raise ValidationError(errors)
 
-    #avoid revisioning by updating directly
+    # check for metadata_modified
+    if data_dict.get("metadata_modified"):
+        mmod = datetime.datetime.strptime(data_dict.get(
+            "metadata_modified"),  "%Y-%m-%dT%H:%M:%S.%f")
```
wardi, line 326: instead of strptime here, put it in a validator on the field instead so that a validation error can be raised cleanly in the API.

Also `metadata_modified` will now need to be stripped any time data_dict is copied from existing data, e.g. in `package_patch`, `resource_create`, `resource_update` etc. Check for all existing callers to `package_update` to find other instances.

Zharktas:
Yeah, I'm just pointing out that modifying `metadata_modified` might have side-effects. Harvesting breaking down was one of them when we did update the timestamps for one of our clients with validators.

PR #6844, opened 2022-05-06
2022 may types update

smotornyuk:
Update `pyright` to the latest version. 

* Fix all newly discovered problems
* `Self` reference from PEP 673 that allows simplifying the following:
  ```python
  TUser = TypeVar("TUser", bound="User")
  class User:
      @classmethod
      def x(cls: Type[TUser]) -> list[TUser]:
         ...
  ```
  
  turns into 
  
  ```python
  class User
      @classmethod
      def x(cls) -> list[Self]:
          ...
  ```


tino097:
@smotornyuk hey can you please fix the conflict

smotornyuk:
Sure, done

PR #6863, opened 2022-05-16
[#6158] Add CSRF Protection

### LINKED ISSUE #6158: Builtin CSRF protection
While you can add CSRF protection to your site using [ckanext-security](https://github.com/data-govt-nz/ckanext-security), we would benefit a much larger number of users if this was built-in and enabled by default. The last time this was discussed the consensus was to aim for a new module on CKAN based on the excellent [Django's CSRF protection](https://docs.djangoproject.com/en/dev/ref/csrf/). To avoid having to maintain a port of Django's code I think we should look at leverage another library. Here are the best contenders IMO (please propose more options if you know some!):

1. [Flask-Seasurf](https://flask-seasurf.readthedocs.io/en/latest/) is exactly what we were thinking of, a port Django's CSRF module to Flask. It has no dependencies other than Flask and it seems [fairly well maintained](https://github.com/maxcountryman/flask-seasurf).
2. The [Flask-WTForms CSRF module](https://flask-wtf.readthedocs.io/en/0.15.x/csrf/) can be used in any view without using WTForms (if I understood correctly). This is a much more widely used package, but of course adds the WTForms dependency. This might not be a big deal (and even in the future we might want to think about replacing our venerable navl module)

Both libraries offer what I think are our requirements:

* Configure the CSRF via middleware:
```python
from flask_seasurf import SeaSurf

csrf = SeaSurf(app)
```

or 

```python
from flask_wtf.csrf import CSRFProtect

csrf = CSRFProtect(app)
```

* A Jinja2 helper function to generate the token:
```html
<form method="post">
    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
</form>
```
* Support for AJAX requests ([Flask-Seasurf](https://flask-seasurf.readthedocs.io/en/latest/#ajax-usage) / [Flask-WTForms](https://flask-wtf.readthedocs.io/en/0.15.x/csrf/#javascript-requests))

The implementation plan here would be:

0. Choose a library
1. Add the middleware (TODO do we need to pass any config options?)
2. Update forms in core templates to include the hidden fields with the csrf token, probably with a macro (eg `macro.csrf_token()`)
3. Update the core AJAX calls as needed to pass the token
4. Document steps needed for extensions to apply these changes to their custom forms
5. Update ckanext-scheming (? only if it's overriding the whole core form)

Comments and thoughts welcome

TomeCirun:
Fixes #6158

### Proposed fixes:
* Added CSRFProtection to all forms in CKAN core
* Added CSRFProtection Header to AJAX calls
* All CKAN extensions are exempted till their implementation to forms is done. (this is configurable through a `ckan.csrf_protection.enabled = false` in the config file) when extensions are ready simply put True to `ckan.csrf_protection.enabled` to enable CSRF protection

For now, I keep it simple, and instead of making and using the hidden input as macro I am simply using:
`<input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>`

Regarding: `add_csrf_token_to_session`. 
After 3days of research, I failed to find why csrf_token is not added in the session when login/register and it fails with 400: Session token is missing. (I opened an [issue](https://github.com/wtforms/flask-wtf/issues/519) in flask_wtf repo)
So I made this helper function that would add the csrf token in the session in the same way as flask_wtf does, till we find the main source of the problem.

On `ckan_before_request` I can see the token coming from the form via request.form, but not in the session, so a response is returned with `400: Session token is missing`, probably from this function: https://github.com/wtforms/flask-wtf/blob/0596de04c1a93a338a96a3e5e2d742d65f77bf2f/src/flask_wtf/csrf.py#L207

On `ckan_after_request` finally I can see the token in the session but it's too late, we are already 400.

- [ ] includes API changes


TomeCirun:
@amercader @wardi  you can check my commits `add csrf protection` and the latest [one](https://github.com/ckan/ckan/pull/6863/commits/98a388b0bb8bf28f3780d81a6578123fe9154aaa) 
and let me know your opinion on this.
With this on, we are entering the final phase. 
Next is to find what causes the problem with the session on login/register and that would be all. Will be ready for merging :) 

Thanks

amercader:
@TomeCirun , @pdelboca will be leading the review of this one. Let us know when things are ready for review

TomeCirun:
> @TomeCirun , @pdelboca will be leading the review of this one. Let us know when things are ready for review

Thanks, @amercader, I believe it's ready for review, though, we need to solve this one...
image

TomeCirun:
I found what causes the bug, `WTF_CSRF_FIELD_NAME` was stored in the `session` as `csrf_token` by default,  so it was passing the check on `_is_valid_session_cookie_data()` and it sets the `allow_cache`  to `False`

TomeCirun:
@pdelboca how do you prefer to include the csrf hidden input in our templates? As included atm `<input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>` or should i change it to `{{ form.hidden(g.csrf_field_name, csrf_token()) }}` ?

test-core.ini
```diff
@@ -90,6 +90,8 @@ ckan.tracking_enabled = true
 beaker.session.key = ckan
 beaker.session.secret = This_is_a_secret_or_is_it
 
+ckan.csrf_protection.field_name = _csrf_token
+
```
pdelboca, line 94: Is this needed? Since we are also doing:
```
    # exempt the CKAN TESTS from csrf-validation.
    config["WTF_CSRF_METHODS"] = []
```

If it is needed, can we document why?

ckan/logic/__init__.py
```diff
@@ -177,6 +179,15 @@ def parse_params(
...
         parsed[key] = value
+    # checks and delete, if the csrf_token is in "parsed".
+    # we don't want the csrf_token to be a part of a data_dict
+    # as it will expose the token to the resource-metadata.
+    # this way we are deleting the token from every data_dict that fills
+    # from request.form instead of deleting it separately in every view.
+    # WTF_CSRF_FIELD_NAME is added by flask_wtf
+    csrf_token = config.get_value("WTF_CSRF_FIELD_NAME")
+    if csrf_token in parsed:
+        parsed.pop(csrf_token)
```
pdelboca, line 190: I usually don't like big comments since it's an indicator that the logic is too complex to understand. In this cases is better to extract the logic to a small method, add a descriptive name to it and use the docstring to document it. (instead of multiple comment lines.)

Can we refactor this into a cleaner function?

ckan/config/middleware/flask_app.py
```diff
@@ -340,6 +356,28 @@ def set_remote_user_as_current_user_for_tests():
...
+    register_blueprint = "user.register"
+    is_login_form = request.endpoint == login_blueprint
+    is_register_form = request.endpoint == register_blueprint
+
+    # WTF_CSRF_FIELD_NAME//TIME_LIMIT are set by flask_wtf
+    field_name = config.get_value("WTF_CSRF_FIELD_NAME")
+    time_limit = config.get_value("WTF_CSRF_TIME_LIMIT")
+
+    if not session.get(field_name) and (is_login_form or is_register_form):
+        secret_key = config.get_value("SECRET_KEY")
```
pdelboca, line 372: Shouldn't we try to use `WTF_CSRF_SECRET_KEY` first?

pdelboca (commented):
Thanks for this @TomeCirun and sorry for the delay on reviewing! I added some comments :)

Also, I think it would be nice to add a `macro` and use is the the forms that are already using `macros`. Others (like small Accept/Decilne forms) can keep the HTML tag since it's not clear the criteria on when to use macros and when to use HTML.

ckan/config/middleware/flask_app.py
TomeCirun: I commented out this function, I think we might not need this anymore, I tested it for 2 hours and it works as it is supposed to.
Previously the csrf_token was not added in the session when login/register and it was failing with 400: Session token is missing. 
That's why at first I made this helper function
It works now without it, can you test it and confirm, please?

@pdelboca

TomeCirun: we can delete it later if everything's okay

test-core.ini
TomeCirun: Not anymore, I deleted it

ckan/logic/__init__.py
TomeCirun: sure, thanks for your note

ckan/config/config_declaration.yaml
```diff
@@ -543,6 +543,61 @@ groups:
...
+      - key: WTF_I18N_ENABLED
+        type: bool
+        default: True
+        description: |
+          Set to False to disable Flask-Babel I18N support.
+          Also set to False if you want to use WTForms’s built-in messages directly, see more info here. Default is True.
+
+      - key: ckan.csrf_protection.field_name
+        default: csrf_token
+        description: Name of the form field and session key that holds the CSRF token.
```
pdelboca, line 600: Why do we need to configure this? 🤔

ckan/config/middleware/flask_app.py
```diff
@@ -246,12 +249,25 @@ def save_session(self, app: Any, session: Any, response: Any):
...
-            plugin_blueprints = [plugin_blueprints]
-        for blueprint in plugin_blueprints:
+        # plugin_blueprints must not send as list in csrf.exempt
+        if isinstance(plugin_blueprints, list):
+            plugin_blueprints = plugin_blueprints[0]
+        # we need to exempt CKAN extensions till they are ready
+        # to implement the csrf_token to their forms, otherwise
+        # they will get 400 Bad Request: The CSRF token is missing.
+        if not asbool(config.get("ckan.csrf_protection.enabled", False)):
+            _csrf.exempt(plugin_blueprints)
```
pdelboca, line 268: I find the `ckan.csrf_protection.enabled` confusing since it is enabled by default, but this is just an exempt for extensions. I think we should make this config name more explicit to its use. 

Something like: `ckan.csrf_protection.exempt_extensions` or `ckan.csrf_protection.ignore_extensions`

Also, I would add a `log.warn()` saying that we are allowing extensions to run without CSRF protection but that It will be forced in the future.

ckan/config/middleware/flask_app.py
```diff
@@ -59,6 +60,8 @@
 from ckan.types import CKANApp, Config, Response
 
 log = logging.getLogger(__name__)
+# do not use _csrf unless you know what you're doing
+_csrf = CSRFProtect()
```
pdelboca, line 64: Let's call it `csfr` since by design is not a private variable (we are accesing it from other modules).

Also, let's remove or change the comment. Is there a specific scenario in mind that can break everything? If so, it would be better to document it. If not I don't find the comment being useful.

pdelboca (commented):
Thanks for the changes @TomeCirun ! I have added some extra comments as I keep testing this feature.

So far I have tested with `add_csrf_token_to_session` function commented and it stills works (I receive a 400 if I try to hack it). But keep it for now until I finish the review. (I'm still trying to understand what's going on behind the scenes)

ckan/config/middleware/flask_app.py
pdelboca: If fact, a good log.warn(...) can also be usefult to explain what are we explaining in the comment. (And therefore, remove it)

TomeCirun:
@pdelboca thanks for your time and for your review, please, let me know if I can do anything else.

pdelboca:
@TomeCirun some comments:
 First, move the warning outside the for loop (next to the csrf init code) so the warning is displayed once and not each time we are setting up Blueprints. (Since it is a server config and not a Blueprint one)
 
Second, for warning messages, it is better to use the following format to define the string since it will print all the of it in one line (best practice for logging):
 
 ```python
csrf_warn_extensions = (
        "Extensions are excluded from CSRF protection! "
        "We allow extensions to run without CSRF protection "
        "but will be forced in the future. "
        "To be able to use CSRF protection you would need to "
        "set ckan.csrf_protection.ignore_extensions=False in your .ini file "
        "and to set the csrf_token in your forms. "
        "e.g {% import 'macros/form.html' as form %} "
        "then you can use it like: {{ form.csrf_protection() }}"
        )
```

Third, this is a change that will require some work for extensions maintainers so it would be nice to add some paragraphs into the docs with instructions. I'm thinking on something like this in Best Practices section:

```
## Implementing CSRF protection

CKAN 2.10 introduces CSRF protection for all the forms. This behavior will be enforced for all extensions 
in the upcoming release.

To add CSRF protection to your extensions use the CSRF macro in all your 
forms: `{{ form.csrf_protection() }}`
```

This documentation will give some info to developers and allow the warning message to be less verbose. We could replace it for something shorter like: `CSRF protection in extensions is disabled but will be enforced in upcoming releases. See docs for more info.`

What do you think @wardi @smotornyuk @Zharktas ? How did we manage this communication with users/developers in the past?

**Comment by TomeCirun** [comment]
@pdelboca, the failed test is not related to my changes

File: ckan/config/middleware/flask_app.py [hunk 7]
```diff
@@ -60,6 +61,19 @@
...
+csrf_warn_extensions = (
+        "Extensions are excluded from CSRF protection! "
+        "We allow extensions to run without CSRF protection "
+        "but will be forced in the future. "
+        "To be able to use CSRF protection you would need to "
+        "set ckan.csrf_protection.ignore_extensions=False in your .ini file "
+        "and to set the csrf_token in your forms. "
+        "e.g {% import 'macros/form.html' as form %} "
+        "then you can use it like: {{ form.csrf_protection() }}"
+    )
```
pdelboca, line 75:
```suggestion
csrf_warn_extensions = (
        "Extensions are excluded from CSRF protection! "
        "We allow extensions to run without CSRF protection "
        "but it will be forced future releases. "
        "Read the documentation for more information on how to add "
        "CSRF protection to your extension."        
    )
```

ckan/config/middleware/flask_app.py
```diff
@@ -340,6 +366,30 @@ def set_remote_user_as_current_user_for_tests():
...
+#     if not session.get(field_name) and (is_login_form or is_register_form):
+#         secret_key = secret_key
+#         token = request.form.get(field_name)
+#         salt = "wtf-csrf-token"
+#         s = URLSafeTimedSerializer(secret_key, salt)
+#         if token:
+#             token_key = s.loads(token, max_age=time_limit)
+#             session[field_name] = token_key
+
+
```
pdelboca, line 392: We are safe to remove this. It is working properly without it.

ckan/config/middleware/flask_app.py
```diff
@@ -353,6 +403,8 @@ def ckan_before_request() -> Optional[Response]:
 
     g.__timer = time.time()
 
+    # add_csrf_token_to_session()
+
```
pdelboca, line 407: And this also can be removed.

doc/extensions/best-practices.rst
```diff
@@ -189,3 +189,18 @@ and other problems`_.
...
+
+CKAN 2.10 introduces CSRF protection for all the forms. This behavior will be enforced for all extensions
+in the upcoming release.
+
+To add CSRF protection to your extensions use the CSRF macro in all your
+forms
+
+::
+
+    {{ form.csrf_protection() }}
```
pdelboca, line 206:
```suggestion
----------------------------
Implementing CSRF protection
----------------------------

CKAN 2.10 introduces CSRF protection for all the forms. This behavior will be enforced for all extensions
in the upcoming release.

To add CSRF protection to your extensions use the CSRF macro in all your
forms

::
    {% import 'macros/form.html' as form %}
    ...
    {{ form.csrf_protection() }}
```

pdelboca (commented):
This looks good @TomeCirun! Just some clean up follow up and we are ready to go!

TomeCirun:
> This looks good @TomeCirun! Just some clean up follow up and we are ready to go!

Done! :) 

@pdelboca

pdelboca:
Thanks @TomeCirun ! My apologies but forgot to request if you could add a change file for this new feature. It should appear in the changelog :)

TomeCirun:
@pdelboca please  let me know if there is anything else I can do

pdelboca:
Thanks for this @TomeCirun ! I was taking a look at the failing test but it's not related to this.

amercader:
Great work @TomeCirun and @pdelboca .
FYI there are a couple of followup issues:

* https://github.com/ckan/ckan/pull/7014
* https://github.com/ckan/ckan/issues/7016

ckan/public/base/javascript/modules/follow.js
```diff
@@ -30,6 +30,19 @@ this.ckan.module('follow', function($) {
 			this.el.on('click', this._onClick);
 		},
 
+		/* Adds the X-CSRFToken header
+		*/
+		addCsrfTokenHeader: function() {
```
amercader, line 35: It would be nice to move this function to the base `module.js` class so it's available to all CKAN modules

amercader: Actually, having tested it again now, this doesn't work at all :)

First of all because the name is wrong, should be:

```
csrftoken = $('meta[name=_csrf_token]').attr('content')
```

Otherwise csrftoken is `undefined`. 
But that doesn't matter anyway because we are exempting all API calls from CSRF protection here:
https://github.com/ckan/ckan/pull/6863/files#diff-97f365b947bcdf4f698327bcdfcb6409e574e9ef13ac899a18e81be73b61d2d4R212

See other comment for discussion about this

ckan/views/api.py
```diff
@@ -207,7 +208,8 @@ def mixed(multi_dict: "MultiDict[str, Any]") -> dict[str, Any]:
 
 # View functions
 
-
+# exempt the CKAN API actions from csrf-validation.
+@csrf.exempt
```
amercader, line 212: @wardi @pdelboca It seems like I spoke too soon and got confused by this

If we exempt API calls from CSRF checks we are essentially still allowing cookie based authentication on POST requests to the API

Removing this exemption would prevent POST API calls unless they send the CSRF header (if CSRF protection is on). Would that be enough or do we need to ensure it at a lower level (flask-login maybe)?

wardi: We need to check the CSRF token on any call (POST or GET) that because e.g. GET can be used to return private user details

amercader: OK, then we need to drop the first part of this check in JS code (or remove GET from it):

```js
if (!/^(GET|HEAD|OPTIONS|TRACE)$/i.test(settings.type) && !this.crossDomain) {
```

wardi: maybe this isn't such a big deal. The same information can be retrieved with a GET request to a normal web page without requiring a CSRF token so adding the requirement for our API might be unnecessary.

TomeCirun: I remember, in one of the dev meetings, we discussed CSRF protection, @wardi  told me that we don't need to include CSRF Protection for the CKAN API actions as they have their own protection from CKAN and that should be enough.

That's why I exempt the CKAN API actions from csrf-validation

ckan/public/base/javascript/modules/follow.js
TomeCirun: Yes, I made this function before we decided to exempt the CKAN API actions, and after a while, I changed the csrf_name... and forgot to change it here.
I think there is no harm in having this function, maybe we will have use of it later :) @amercader

ckan/views/api.py
TomeCirun: CSRF attacks rely on cookies being implicitly sent with all requests to a particular domain. Since we do not allow cookie-based authentication for our CKAN API Actions and we are using an access token which we are sending it using Authorization: <token> Header,  we should be secure.
@amercader @wardi

wardi: but, we *do* allow cookie-based auth for the CKAN API, unless something has changes.

ckan/templates/base.html
```diff
@@ -24,6 +24,7 @@
     #}
     {%- block meta -%}
       <meta charset="utf-8" />
+      <meta name="{{ g.csrf_field_name }}" content="{{ csrf_token() }}" />
```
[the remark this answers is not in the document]
TomeCirun: JavaScript can include it in any form submissions the page makes.
for e.g `csrftoken = $('meta[name=_csrf_token]').attr('content')`

ckan/views/api.py
amercader: Follow-up discussion here: https://github.com/ckan/ckan/issues/7028

PR #6865, opened 2022-05-17
form input with append

### LINKED ISSUE #6768: form input has prepend but not append
It would be nice to extend `form.html` so it supports `append` too, which would show a label after the input, instead of before like `prepend` does.

https://github.com/ckan/ckan/blob/37f86db98b8e2e8b67155b8ba1e591873b9ba2de/ckan/templates/macros/form.html#L177

https://github.com/ckan/ckan/blob/37f86db98b8e2e8b67155b8ba1e591873b9ba2de/ckan/templates/macros/form.html#L197-L201

Gauravp-NEC:
Fixes #6768

### Proposed fixes:

This PR fixes the issue https://github.com/ckan/ckan/issues/6768.
`form.html` is extended and it supports `append`

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/macros/form.html
+++ b/ckan/templates/macros/form.html
@@ -11,6 +11,7 @@ and here, we are importing them all to maintain backward compatibility.
 {% from "macros/form/markdown.html" import markdown %}
 {% from "macros/form/textarea.html" import textarea %}s
 {% from "macros/form/prepend.html" import prepend %}
+{% from "macros/form/append.html" import append %}
 {% from "macros/form/custom.html" import custom %}
 {% from "macros/form/errors.html" import errors %}
 {% from "macros/form/info.html" import info %}
@@ -27,6 +28,7 @@ and here, we are importing them all to maintain backward compatibility.
 {% set markdown = markdown %}
 {% set textarea = textarea %}
 {% set prepend = prepend %}
+{% set append = append %}
 {% set custom = custom %}
 {% set errors = errors %}
 {% set info = info %}


ckan/templates/macros/form/append.html
```diff
@@ -0,0 +1,38 @@
+{% from "macros/form/input_block.html" import input_block %}
+{% from "macros/form/attributes.html" import attributes %}
+
+{#
+Creates all the markup required for an input element with a prefixed segment.
+These are useful for showing url slugs and other fields where the input
+information forms only part of the saved data.
```
amercader, line 7:
```suggestion
Creates all the markup required for an input element with a segment attached at the end.
These are useful for fields where the input information forms only part of the saved data.
```

ckan/templates/macros/form/append.html
```diff
@@ -0,0 +1,38 @@
...
+{# We manually append the error here as it needs to be inside the .input-group block #}
+{% set classes = (classes|list) %}
+{% do classes.append('error') if error %}
+{%- set extra_html = caller() if caller -%}
+{% call input_block(id or name, label or name, error='', classes=classes, extra_html=extra_html, is_required=is_required) %}
+<div class="input-group">
+    {% if append %}<label class="input-group-text">{{ append }}</label>{%- endif -%}
+    <input id="{{ id or name }}" type="{{ type }}" name="{{ name }}" value="{{ value | empty_and_escape }}" placeholder="{{ placeholder }}" {{ attributes(attrs) }} />
+    {% if error and error is iterable %}<span class="error-block">{{ error|join(', ') }}</span>{% endif %}
+</div>
```
amercader, line 36: This HTML is identical to the `prepend` macro one so obviously the rendered field will look the same. Please test the changes locally before submitting a PR

amercader (changes requested):
@Gauravp-NEC It looks like you haven't tested the changes locally as the field is rendered exactly in the same way as the prepend macro. Please resubmit when it is working as expected or focus on another issue.

PR #6868, opened 2022-05-19
ckan_config test mark works with request context

### LINKED ISSUE #6841: `ckan_config` pytest mark doesn't work when used with the `with_request_context` fixture
**CKAN version**
2.9

**Describe the bug**
If you use the `with_request_context` fixture, changes in config introduced by `@pytest.mark.ckan_config`  are ignored.

**Steps to reproduce**
Consider the following new tests, first one passes, second fails:

```diff
diff --git a/ckan/tests/pytest_ckan/test_fixtures.py b/ckan/tests/pytest_ckan/test_fixtures.py
index 1cca6ae28..9cca8a044 100644
--- a/ckan/tests/pytest_ckan/test_fixtures.py
+++ b/ckan/tests/pytest_ckan/test_fixtures.py
@@ -26,6 +26,17 @@ def test_ckan_config_mark(ckan_config):
 
 # END-CONFIG-OVERRIDE
 
+@pytest.mark.ckan_config("some.new.config", "exists")
+@pytest.mark.usefixtures("with_request_context")
+def test_new_ckan_config_mark_with_test_request(ckan_config):
+    assert ckan_config["some.new.config"] == "exists"
+
+
+@pytest.mark.ckan_config("ckan.site_url", "https://example.org")
+@pytest.mark.usefixtures("with_request_context")
+def test_existing_ckan_config_mark_with_test_request(ckan_config):
+    assert ckan_config["ckan.site_url"] == "https://example.org"
+
 
 @pytest.mark.ckan_config(u"some.new.config", u"exists")
 @pytest.mark.usefixtures(u"ckan_config")
```

smotornyuk:
Fixes #6841

`app_globals` module caches a set of the config options with their values when test execution started(just like it does every time web-server is launched). After that, every time the application object is created or the list of plugins is reloaded, the mentioned config options restore their original value, even if they were modified(via the `ckan_config` mark, for example)

It can be fixed by the clearing `app_globals` cache whenever a config object is required inside the test. But it creates another issue. Without the "initial" cached state, every change to the config object that is done directly(not via `ckan_config` mark or `monkeypatch` fixture), will affect all consequent tests. For example, code that tests overriding configuration via admin interface can set site title and all the other tests will reuse the updated title, not the original value defined by the configuration file.

That's why I'm creating a copy of the config object that was used before any test execution and resetting `ckan.common.config` state before each test using this snapshot.

### DIFF
--- a/ckan/tests/pytest_ckan/ckan_setup.py
+++ b/ckan/tests/pytest_ckan/ckan_setup.py
@@ -3,10 +3,16 @@
 import ckan.plugins as plugins
 from ckan.config.middleware import make_app
 from ckan.cli import load_config
+from ckan.common import config
 
 # This is a test Flask request context to be used internally.
 # Do not use it!
 _tests_test_request_context = None
+
+# Initial config snapshot that is used to restore config object before each
+# test. This allows us to keep tests independent while we are using global
+# config object.
+_config = config.copy()
 
 
 def pytest_addoption(parser):
@@ -39,6 +45,10 @@ def pytest_sessionstart(session):
         flask_app = app._wsgi_app
     _tests_test_request_context = flask_app.test_request_context()
 
+    # Create the snapshot of the initial configuration
+    global _config
+    _config = config.copy()
+
 
 def pytest_runtestloop(session):
     """When all the tests collected, extra plugin may be enabled because python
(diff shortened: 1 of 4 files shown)


amercader:
@smotornyuk this makes sense, but it would be nice to get rid of all that app_globals weird logic eventually

PR #6879, opened 2022-05-27
[#6878] Fix custom dataset validation

### LINKED ISSUE #6878: Validation doesn't work on plugins inheriting the core `IDatasetForm` methods
**CKAN version**
Current master (2.10)

**Describe the bug**
Consider the following test:

```python
import ckan.plugins.toolkit as tk

class CustomDatasetType(plugins.SingletonPlugin, tk.DefaultDatasetForm):

    plugins.implements(plugins.IDatasetForm, inherit=True)

    def package_types(self):

        return ["custom_dataset"]


@pytest.mark.ckan_config("ckan.auth.create_unowned_dataset", True)
def test_search_datasets_returns_only_datasts(self, app):
 
    factories.Dataset(type="custom_dataset")

```

This will fail with an SQLAlchemy error

```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "name" violates not-null constraint

```

Upon further investigation, when calling the NAVL validation function from `package_create`:

https://github.com/ckan/ckan/blob/a18603ec182e279488c0e91b78cd957a3ed49d30/ckan/logic/action/create.py#L178-L179

`data` is an empty dict!

I'll send a proposed fix as PR

amercader:
Fixes #6878 

The output of the default return value for `IDatasetForm.validate()` was changed during the typing work from `None` to `({}, {})` (https://github.com/ckan/ckan/commit/b2a03aa66b50fde7a68e0b2921104475117e7093), but the check performed in `plugin_validate()` to see if validation had been performed was not updated.

This meant that when validating a custom dataset on `package_create` the validation returned an empty dict instead of the data, causing SQLAlchemy to fail.

@smotornyuk I assume we need to check the `IDatasetForm.validate()` output for typing purposes so I've updated the check to account for this. Let me know if there is a better approach.

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -320,7 +320,7 @@ def plugin_validate(
     """
     if hasattr(plugin, 'validate'):
         result = plugin.validate(context, data_dict, schema, action)
-        if result is not None:
+        if result is not None and result != ({}, {}):
             return result
 
     return validate(data_dict, schema, context)
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -320,7 +320,7 @@ def plugin_validate(
     """
     if hasattr(plugin, 'validate'):
         result = plugin.validate(context, data_dict, schema, action)
-        if result is not None and result != ({}, {}):
+        if result is not None:
             return result
 
     return validate(data_dict, schema, context)
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -1268,7 +1268,7 @@ class IDatasetForm(Interface):
         return ''
 
     def validate(self, context: Context, data_dict: DataDict, schema: Schema,
-                 action: str) -> tuple[dict[str, Any], dict[str, Any]]:
+                 action: str) -> Optional[tuple[dict[str, Any], dict[str, Any]]]:
         u'''Customize validation of datasets.
 
         When this method is implemented it is used to perform all validation
@@ -1298,7 +1298,7 @@ class IDatasetForm(Interface):
           and lists-of-string-error-messages as values
         :rtype: (dictionary, dictionary)
         '''
-        return {}, {}
+        return
 
     def prepare_dataset_blueprint(self, package_type: str,
                                   blueprint: Blueprint) -> Blueprint:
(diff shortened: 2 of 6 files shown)


smotornyuk:
I can't tell the exact reason, why I made this change. Most likely, I tried to narrow down the diversity of return types from functions. It really helps in the initial stages of typing. And it simplifies code because with mandatory tuple returned from validation function, you can do such thing:
```python
data, errors = plugin.validate()
if errors: 
    ...
```

While original implementation requires you to write 
```python
result = plugin.validate()
if result:
    data, errors = result()
    if errors:
        ...
```

I've just checked, you can revert to original behavior without any consequences. Here's the change required for it:
```diff
diff --git a/ckan/plugins/interfaces.py b/ckan/plugins/interfaces.py
index 593bcd3b2..bbabdb9e1 100644
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -1268,7 +1268,7 @@ class IDatasetForm(Interface):
         return ''
 
     def validate(self, context: Context, data_dict: DataDict, schema: Schema,
-                 action: str) -> tuple[dict[str, Any], dict[str, Any]]:
+                 action: str) -> Optional[tuple[dict[str, Any], dict[str, Any]]]:
         u'''Customize validation of datasets.
 
         When this method is implemented it is used to perform all validation
@@ -1298,7 +1298,7 @@ class IDatasetForm(Interface):
           and lists-of-string-error-messages as values
         :rtype: (dictionary, dictionary)
         '''
-        return {}, {}
+        return
 
     def prepare_dataset_blueprint(self, package_type: str,
                                   blueprint: Blueprint) -> Blueprint:
@@ -1484,7 +1484,7 @@ class IGroupForm(Interface):
         '''
 
     def validate(self, context: Context, data_dict: DataDict, schema: Schema,
-                 action: str) -> tuple[dict[str, Any], dict[str, Any]]:
+                 action: str) -> Optional[tuple[dict[str, Any], dict[str, Any]]]:
         u'''Customize validation of groups.
 
         When this method is implemented it is used to perform all validation
@@ -1515,7 +1515,7 @@ class IGroupForm(Interface):
           and lists-of-string-error-messages as values
         :rtype: (dictionary, dictionary)
         '''
-        return {}, {}
+        return
 
     def prepare_group_blueprint(self, group_type: str,
                                 blueprint: Blueprint) -> Blueprint:
```

amercader:
Thanks @smotornyuk , I reverted the change

PR #6883, opened 2022-05-31
Make actions' data_dict param optional again

amercader:
Because of how the wrapped functions returned by `get_action()` params are defined, you must pass a data_dict value when calling them, otherwise you get an error:

    E       TypeError: wrapped() missing 1 required positional argument:
'data_dict'

But not all actions require a `data_dict`, eg `status_show`, `package_list`, etc

This was not an issue on the HTTP API endpoint because the blueprint always passes a value for `data_dict`, even if it's None, but calling actions directly from Python will fail:

    toolkit.get_action("package_list")({"ignore_auth": True})

    toolkit.get_action("package_list")()

@smotornyuk let me know if the type hints make sense

### DIFF
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -502,8 +502,8 @@ def get_action(action: str) -> Action:
     # wrap the functions
     for action_name, _action in _actions.items():
         def make_wrapped(_action: Action, action_name: str):
-            def wrapped(context: Optional[Context],
-                        data_dict: DataDict, **kw: Any):
+            def wrapped(context: Optional[Context] = None,
+                        data_dict: Optional[DataDict] = None, **kw: Any):
                 if kw:
                     log.critical('%s was passed extra keywords %r'
                                  % (_action.__name__, kw))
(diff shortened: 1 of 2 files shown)


smotornyuk:
If we allow optional `data_dict`, it means that absolutely any action can accept either a real dictionary or `None`, because `tk.get_action` returns an abstract action and there is no way to figure out, whether this abstract action uses `data_dict`. Only real human knows whether `data_dict` is actually used inside an action. In addition, `chained_action`(let's imagine, you chained `status_show` in order to provide extra details depending on `data_dict`) and auth/`chained_auth` can start using `data_dict`, which means that one cannot even be sure, that `status_show` never uses `data_dict`.

I think, from the code-quality point, it's better to pass the `data_dict` all the time. The same as context, it may not be used everywhere, but it guarantees consistent and predictable API.

But if you insist on keeping old code compatible, we can use a compromise here. In the same way `context` is replaced with dictionary whenever it's `None`, we can set `data_dict` to a `{}` when it's missing:
```diff
diff --git a/ckan/logic/__init__.py b/ckan/logic/__init__.py
index b8c4bb921..a682fb930 100644
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -509,6 +509,8 @@ def get_action(action: str) -> Action:
                                  % (_action.__name__, kw))
 
                 context = _prepopulate_context(context)
+                if data_dict is None:
+                    data_dict = {}
 
                 # Auth Auditing - checks that the action function did call
                 # check_access (unless there is no accompanying auth function).
```

amercader:
All green @smotornyuk

PR #6887, opened 2022-05-31
Split DomainObject into smaller mixins

smotornyuk:
Extract independent parts of the `ckan.model.domain_object.DomainObject` into smaller mixins inside `ckan.model.base`. It simplifies reusing these chunks

### DIFF
--- a/ckan/model/base.py
+++ b/ckan/model/base.py
@@ -1,6 +1,231 @@
 # -*- coding: utf-8 -*-
+from __future__ import annotations
 
+import datetime
+from collections import OrderedDict
+from typing import Any, Callable, Optional
+
+import sqlalchemy as sa
+from sqlalchemy import orm
 from sqlalchemy.ext.declarative import declarative_base
-from .meta import metadata
+from typing_extensions import Self
+
+from .meta import metadata, Session
 
 Base = declarative_base(metadata=metadata)
+"""Base class for SQLAlchemy declaration-style models.
+"""
+
+
+class SessionMixin:
+    """Attach session to the model class.
+
+    Example:
+        class MyModel(SessionMixin, Base):
+            ...
+            @classmethod
+            def count(cls):
+                return cls.Session.query(cls).count()
+
+    """
+    Session = Session
+
+
+class DebugMixin:
+    """Defines __repr__ method that shows all the model's columns.
+
+    Example:
+        class MyModel(DebugMixin, Base):
+            name: str
+            ...
+
+        >>> MyModel(name="hello world")
+        <MyModel name=hello world>
+    """
+
+    def __repr__(self):
+        """Show model name and all the columns with their values.
+        """
+        output = u'<%s' % self.__class__.__name__
+        table: Any = orm.class_mapper(self.__class__).persist_selectable
+        for col in table.c:
+            try:
+                output += u' %s=%s' % (col.name, getattr(self, col.name))
+            except Exception as inst:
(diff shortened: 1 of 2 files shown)


ckan/model/base.py
```diff
@@ -1,6 +1,231 @@
...
+
+
+class TextSearchMixin:
+    """Provide base text-search functionality via LIKE SQL operator.
+
+    Text-search uses columns specified by `text_search_fields` property.
+
+    Example:
+        class MyModel(TextSearchMixin, Base):
+            text_search_fields = ["name]
```
pdelboca, line 106:
```suggestion
            text_search_fields = ["name"]
```

ckan/model/base.py
```diff
@@ -1,6 +1,231 @@
...
+    text_search_fields: list[str] = []
+
+    @classmethod
+    def text_search(cls, query: Any, term: str) -> Any:
+        """Add ILIKE filters to the query.
+        """
+        register = cls
+        make_like: Callable[
+            [Any, str], str] = lambda x, y: x.ilike('%' + y + '%')
+        q: Any = sa.null() | sa.null()
```
pdelboca, line 124: `sa.null() | san.null()`?

Is this intended? 🤔

ckan/model/base.py
```diff
@@ -1,6 +1,231 @@
...
+        """Mark an instance as deleted.
+        """
+        # stateful objects have this method overridden - see
+        # core.StatefulObjectMixin
+        self.Session.delete(self)
+
+    def purge(self) -> None:
+        """Mark an instance as deleted.
+        """
+        self.Session().autoflush = False
```
pdelboca, line 171: Why are we setting `autoflush` to False?

pdelboca (approved):
Overall looks good @smotornyuk . I left some minor comments.

One question I would like to address is: What's the goal of manually managing the autoflush of sessions? It is not clear though the codebase why are we setting it and when/why we remove it.

Example:
 - We create our Session object with `autoflush=False` but then by default we are setting it to `True` (by default) when calling `by_name(...)`. So all the `by_name` calls behaves implicitly different.
 - Why setting `self.Session().autoflush() = False` when purging? The session has already been created with that attribute.
 - When editing the `autoflush` value, it is recommended to use a context manager. https://docs.sqlalchemy.org/en/13/orm/session_basics.html?highlight=autoflush#flushing

I always found difficult to understand this. Do you have any context on its use?

ckan/model/base.py
smotornyuk: Looks like a way to ensure that the query exists all the time because the next line begins a loop, that is going to append new clauses to the query.

ckan/model/base.py
smotornyuk: I assume because the author of these lines didn't want to purge data at that very moment. 11 years passed since this line was added and we have transactions so now it has doesn't look great. 
Still, it won't be very wise to drop such lines inside a PR which only moves code and focuses on preserving existing behavior.

smotornyuk:
No, these lines were added even before I learned how to write python code XD
I just extracted the code without changes, so that all the existing features/tricks/bugs are available for future exploits and fixes:)

pdelboca:
Sounds good @smotornyuk ! I'll merge this. Are you working on a documentation to update how we work with models in CKAN right?

amercader:
Woo yeah, the CKAN book! :blue_book: Sounds great @smotornyuk

PR #6888, opened 2022-05-31
Use filename when downloading resource

smotornyuk:
Fixes #6866 

The filename passed to `resource.view` route is ignored and the downloaded file uses the ID of the resource as a name. Can be fixed by passing the filename into the underlying Flask function.

### DIFF
--- a/ckan/views/resource.py
+++ b/ckan/views/resource.py
@@ -6,6 +6,7 @@ import json
 import logging
 from typing import Any, cast, Optional, Union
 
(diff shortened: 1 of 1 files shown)


kowh-ai (approved):
Iv'e just tested this PR on a brand new CKAN install @smotornyuk - works like a charm! This can be merged.

amercader:
Go ahead @kowh-ai, you can do it yourself :)

PR #6897, opened 2022-06-07
[4986]user_invite

### LINKED ISSUE #4986: user_invite crashes if generated password does not validate
The `user_invite` function generates a hidden password that will pass the built-in validators, but if custom validators are supplied and the generated password fails, the function will crash with a KeyError ('password' does not exist in the error dict).

Furthermore, code review suggests that if the crash were fixed, it would be in danger of an infinite loop, constantly generating passwords that are not accepted.

CKAN 2.8.3

Inviting a new member, identified by email address, to join an organisation, should result in an email to that address.

Internal server error. Stack trace shows that this is a `KeyError: ('password',)` in `user_password_validator`.

- Add a password validator that requires non-alphanumeric symbols.
- Go to an organisation, select Manage-Members.
- Select an email address and Add Member.

TomeCirun:
Fixes #4986 

### Proposed fixes:
Hey @wardi, at first I tried with your proposal number 2, what confuses me is that we need to send that flag `password_disabled ` into the password validators in order to turn them off, this way if you have your own custom validator you will have to implement this logic on your own validator too.

I come up with this solution that from my point of view seems cleaner ( maybe I m wrong I don't see the bigger pic as you :) ) we are disabling the password field in the schema so the extensions developers would not care what implementation they have in their custom validators.

Please, let me know your opinion on this.

Thanks

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1076,14 +1076,10 @@ def user_invite(context: Context,
         'role': data['role']
     }
 
-    if group.is_organization:
-        _get_action('organization_member_create')(context, member_dict)
-        group_dict = _get_action('organization_show')(context,
-                                                      {'id': data['group_id']})
-    else:
-        _get_action('group_member_create')(context, member_dict)
-        group_dict = _get_action('group_show')(context,
-                                               {'id': data['group_id']})
+    _get_action(f'{group.type}_member_create')(context, member_dict)
+    group_dict = _get_action(f'{group.type}_show')(context,
+                                                    {'id': data['group_id']})
+    breakpoint()
     try:
         mailer.send_invite(user, group_dict, data['role'])
     except (socket_error, mailer.MailerException) as error:
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1077,8 +1077,9 @@ def user_invite(context: Context,
     }
 
     _get_action(f'{group.type}_member_create')(context, member_dict)
-    group_dict = _get_action(f'{group.type}_show')(context,
-                                                    {'id': data['group_id']})
+    group_dict = _get_action(f'{group.type}_show')(
+        context, {'id': data['group_id']})
+
     try:
         mailer.send_invite(user, group_dict, data['role'])
     except (socket_error, mailer.MailerException) as error:
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1079,7 +1079,7 @@ def user_invite(context: Context,
     _get_action(f'{group.type}_member_create')(context, member_dict)
     group_dict = _get_action(f'{group.type}_show')(context,
                                                     {'id': data['group_id']})
-    breakpoint()
+
     try:
         mailer.send_invite(user, group_dict, data['role'])
     except (socket_error, mailer.MailerException) as error:
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1062,12 +1062,12 @@ def user_invite(context: Context,
     data['name'] = name
     # send the proper schema when creating a user from here
     # so the password field would be ignored.
-    context['schema'] = ckan.logic.schema.create_user_for_user_invite_schema()
+    invite_schema = ckan.logic.schema.create_user_for_user_invite_schema()
+    
     data['state'] = model.State.PENDING
-    user_dict = _get_action('user_create')(context, data)
-    # we already created the user so we don't need the schema anymore
-    # as we are calling for other actions bellow
-    context.pop('schema')
+    user_dict = _get_action('user_create')(
+        cast(Context, dict(context, schema=invite_schema)),
+        data)
     user = model.User.get(user_dict['id'])
     assert user
     member_dict = {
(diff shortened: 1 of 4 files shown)


ckan/logic/action/create.py
```diff
@@ -1059,24 +1058,16 @@ def user_invite(context: Context,
...
     data['name'] = name
-    data['password'] = password
+    # send the proper schema when creating a user from here
+    # so the password field would be ignored.
+    context['schema'] = ckan.logic.schema.create_user_for_user_invite_schema()
     data['state'] = model.State.PENDING
     user_dict = _get_action('user_create')(context, data)
+    # we already created the user so we don't need the schema anymore
+    # as we are calling for other actions bellow
+    context.pop('schema')
```
wardi, line 1070:
```suggestion
    invite_schema = ckan.logic.schema.create_user_for_user_invite_schema()
    data['state'] = model.State.PENDING
    user_dict = _get_action('user_create')(dict(context, schema=invite_schema), data)
    # we already created the user so we don't need the schema anymore
    # as we are calling for other actions bellow
```
passing in new dicts to context is safer and means we don't have to remember to `pop` anything

ckan/logic/action/create.py
```diff
@@ -1085,14 +1076,10 @@ def user_invite(context: Context,
...
 
-    if group.is_organization:
-        _get_action('organization_member_create')(context, member_dict)
-        group_dict = _get_action('organization_show')(context,
-                                                      {'id': data['group_id']})
-    else:
-        _get_action('group_member_create')(context, member_dict)
-        group_dict = _get_action('group_show')(context,
-                                               {'id': data['group_id']})
+    _get_action(f'{group.type}_member_create')(context, member_dict)
```
wardi, line 1079: `group.type` and `group.is_organization` aren't the same thing. You can have multiple group types and organization types.

wardi:
I like this solution.

Just a couple small comments above, and we need a test to verify that a user can be invited and that no password is assigned.

ckan/logic/action/create.py
TomeCirun: I wasn't aware of this, thanks

ckan/logic/action/create.py
TomeCirun: Yep, forgot about that :D , thanks

ckan/logic/action/create.py
```diff
@@ -1062,12 +1062,12 @@ def user_invite(context: Context,
...
-    context['schema'] = ckan.logic.schema.create_user_for_user_invite_schema()
+    invite_schema = ckan.logic.schema.create_user_for_user_invite_schema()
+    
     data['state'] = model.State.PENDING
-    user_dict = _get_action('user_create')(context, data)
-    # we already created the user so we don't need the schema anymore
-    # as we are calling for other actions bellow
-    context.pop('schema')
+    user_dict = _get_action('user_create')(
+        cast(Context, dict(context, schema=invite_schema)),
```
wardi, line 1069: @smotornyuk I'm still learning about best practices with python typing, would writing this as `Context(context, schema=invite_schema)` be better than using `cast` here?

smotornyuk: Yes, exactly. Your option is preferred all the time. `cast` can be used only when we have no other options. But due to the nature of the `Context`(`TypedDict`), the following form must be used(with no positional args):
```
Context(**context, **{"schema": invite_schema})
```

For example `Context(model=ckan.model)` will likely fail, because at the moment typing libraries do not work well with python modules passed around. And we have to use cast in such a case(eventually, this also will be allowed and `cast` will no longer be needed). 
Almost everywhere else, `Context(**context)` will do the thing.

TomeCirun:
@wardi I've added a test that would create the user without a password, what I m missing is  an invitation as I have no idea right now how to mock `mailer.send_invite`

ckan/logic/action/create.py
smotornyuk: It's a fun part.

`TypedDict` is substituted by the `dict` in runtime. In this way, all the operations from the `dict` are allowed, including positional arguments to the constructor.

But technically, `TypedDict` is a `Mapping`, which does not define a constructor with the positional args. So, while your example does not produce an error as a python code, it'll likely report an invalid operation if you check it with `pyright`

wardi: I think you're saying it works but pyright doesn't understand it so we have to make our code uglier to pass the typing tests we added. That makes me sad.

smotornyuk: Yes, if we put it straight, that's the reason.

smotornyuk: But I'll create a feature request in the pyright source. Maybe, even if the authors won't change their mind, they'll suggest more coinse way to deal with our case

smotornyuk: But, there is a [PEP-589](https://peps.python.org/pep-0589/) which clearly says:

> It can be used as a callable object with keyword arguments corresponding to the TypedDict items. Non-keyword arguments are not allowed.

So this ugly code is a part of python's design :)

TomeCirun:
Hey @wardi, is there anything else that can be done in this PR?

Thanks

wardi:
@TomeCirun how about a changelog entry?

TomeCirun:
@wardi polite reminder :)

PR #6898, opened 2022-06-08
Add E2E tests using Cypress

pdelboca:
This PR adds some end to end tests using cypress. The goal is not to provide an extensive coverage of the system but rather cover some critical workflows like login, upload, data views, etc.

The strategy I wanna follow is simple:
 - Do not rely on a complex set of fixtures/commands to deal with the Database, but use the UI to create the data you need (during your tests). Is it a valid strategy and it will require less maintenance. https://learn.cypress.io/advanced-cypress-concepts/database-initialization-and-seeding
 - Create 4 or 5 specs to cover the main workflows.

Let me know your comments!


Zharktas:
I've been pondering this for years, just haven't had the time to implement anything :D

I was thinking that we could implement seeding extension which would populate database with predefined content like users, datasets, organizations and groups. It takes forever to create them from the UI for every test, especially those that are needed for every test, like user.

My thinking behind this is that the database and solr needs to be cleared before most tests so there wouldn't be dependencies between the tests.

Zharktas:
I looked at the recording to find whats going on and i've seen development server just not responding immediately locally, i've assumed its just something on my end but clearly not since its also happens during tests. Sometimes when you post a form, the respond waits a while and if you post it again it comes back immediatly. So that needs to be fixed and its not an issue with your cypress tests.

Zharktas:
It might make a difference if the server is run in threaded mode by adding `-t` to here https://github.com/ckan/ckan/blob/c2db99b9b10e18e41c4cf0f40dc3629152e67c3b/.github/workflows/cypress.yml#L58

pdelboca:
@Zharktas I can use some help here. The workflow seems to be okay, but for some reason it's not finding the user:

```
2022-06-14 08:54:47,450 INFO  [werkzeug] 127.0.0.1 - - [14/Jun/2022 08:54:47] "POST /login_generic?came_from=/user/logged_in HTTP/1.1" 302 -
2022-06-14 08:54:47,487 INFO  [ckan.config.middleware.flask_app]  302 /user/logged_in render time 0.015 seconds
2022-06-14 08:54:47,488 INFO  [werkzeug] 127.0.0.1 - - [14/Jun/2022 08:54:47] "GET /user/logged_in?came_from=%2Fuser%2Flogged_in&__logins=0 HTTP/1.1" 302 -
2022-06-14 08:54:47,613 INFO  [ckan.config.middleware.flask_app]  200 /user/logged_in render time 0.070 seconds
2022-06-14 08:54:47,614 INFO  [werkzeug] 127.0.0.1 - - [14/Jun/2022 08:54:47] "GET /user/logged_in HTTP/1.1" 200 -
2022-06-14 08:54:47,730 INFO  [ckan.config.middleware.flask_app] 404 Not Found: User not found
```

I tried playing around with `cy.session(...)` but I didn't get any result. This workflow is working locally (and it's working on our project). Do you know any reason why this could be happening?  I'm suspecting that at some point the `g.user` object is not being set properly.

Another hint is that somehow the cookie is not being set and that's causing the error:
```
 2022-06-14 08:54:53,898 INFO  [werkzeug] 127.0.0.1 - - [14/Jun/2022 08:54:53] "POST /login_generic?came_from=/user/logged_in HTTP/1.1" 302 -
2022-06-14 08:54:53,926 INFO  [ckan.config.middleware.flask_app]  302 /user/logged_in render time 0.004 seconds
2022-06-14 08:54:53,928 INFO  [werkzeug] 127.0.0.1 - - [14/Jun/2022 08:54:53] "GET /user/logged_in?came_from=%2Fuser%2Flogged_in&__logins=0 HTTP/1.1" 302 -
2022-06-14 08:54:54,030 INFO  [ckan.config.middleware.flask_app]  200 /user/logged_in render time 0.084 seconds
2022-06-14 08:54:54,033 INFO  [werkzeug] 127.0.0.1 - - [14/Jun/2022 08:54:54] "GET /user/logged_in HTTP/1.1" 200 -
  3) Loging using a POST request

...
...


3) Login form
       Loging using a POST request:
     AssertionError: expected null to exist
      at Context.eval (http://localhost:5000/__cypress/tests?p=cypress/integration/e2e/login.spec.js:158:26)

```

Maybe the application doesn't have permissions to write on the containers disk? 🤔

PR #6905, opened 2022-06-10
[6611]Groups should be selectable as Dataset collaborators 

### LINKED ISSUE #6611: Groups should be selectable as Dataset collaborators

TomeCirun:
Fixes #6611 

I will add more info soon

### Features:

- [ ] includes API changes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,6 +48,7 @@ ckan.click_command =
 
 ckan.plugins =
     activity = ckanext.activity.plugin:ActivityPlugin
+    collaborator_group = ckanext.collaborator_group.plugin:CollaboratorGroupPlugin
     stats = ckanext.stats.plugin:StatsPlugin
     multilingual_dataset = ckanext.multilingual.plugin:MultilingualDataset
     multilingual_group = ckanext.multilingual.plugin:MultilingualGroup


amercader:
@TomeCirun If you are working on this please pause as we need to think on the implications. Apologies as I should have responded to the original issue before.

TomeCirun:
Hey, @amercader, this would work as follow: if the `collaborator_group` plugin is provided,  you will have to manually create the `package_group_member_table` via `CLI` command, next, you will need to have the necessary permissions in order to add a group as a collaborator (same as before) and every member in that group will automatically be a collaborator.

If you decide to disable the `collaborator_group` plugin,
the app will continue to work and the tables from the plugin will still exist in the DB, but we wouldn't have access to them as the plugin is not active (no queries, etc... ) so the group collaborators would be disabled,  the app will fall back on the core code (`collaborators`) so no error will happen.

I can't think of any other scenarios that might occur as a problem.
(this doesn't mean that there wouldn't be any)

ps. please provide more scenarios (if you can think of any) so I can continue to test the plugin.

TomeCirun:
@amercader any news on this ?

EricSoroos:
Related/Reimplementation of #6735?

TomeCirun:
> #6735

Yes, it's a reimplementation, as an extension of CKAN @EricSoroos

EricSoroos:
Group discussion:
* Not entirely sure that the groups have the correct permission limits or facilities to work as expected in all of the edge cases.

PR #6912, opened 2022-06-15
Package plugin_extras

### LINKED ISSUE #4147: Plugin Extras

TomeCirun:
Fixes #4147

### Features:

- [ ] includes API changes


TomeCirun:
@ward, I would need some help here, as you know in package_show, use_cache by default is True, therefore I'm struggling to get the plugin_extras from solar, I would need help with solar,  please let me know if I'm on the right track and what action should I take next.

p.s Should I delete package_extra?

Thanks

tino097:
Ian's commen to continue with the [discsussion](https://github.com/ckan/ckan/issues/4147#issuecomment-1158014489)

TomeCirun:
Hey @wardi, is there anything else that can be done in this PR?

wardi:
We need to look at how plugin_data interacts with the search index. It should be possible to copy fields from plugin_data into solr fields from `before_index` (e.g. for a plugin that adds search terms from file contents or a plugin that uses ML to improve search results) but plugin_data itself shouldn't be stored in the `validated_data_dict` or `data_dict` values by default.

ckan/logic/action/get.py
```diff
@@ -1009,6 +1012,7 @@ def package_show(context: Context, data_dict: DataDict) -> ActionResult.PackageS
     package_dict = None
     use_cache = (context.get('use_cache', True))
     package_dict_validated = False
+    include_plugin_data = asbool(data_dict.get('include_plugin_data', False))
 
     if use_cache:
```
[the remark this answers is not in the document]
TomeCirun: @wardi Yes, plugin_data won't be in the cache, it's only available if the requester is sysadmin and he must pass include_plugin_data=True to package_show alongside use_cache=False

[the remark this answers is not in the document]
TomeCirun: Yep, you are right, that makes sense, thanks for the help

TomeCirun:
> We need to look at how plugin_data interacts with the search index. It should be possible to copy fields from plugin_data into solr fields from `before_index` (e.g. for a plugin that adds search terms from file contents or a plugin that uses ML to improve search results) but plugin_data itself shouldn't be stored in the `validated_data_dict` or `data_dict` values by default.


How should I achieve that? through the tests maybe? I have no idea how before_dataset_index works, maybe you can share some examples of it so I can get the idea? 
Thanks

wardi:
> How should I achieve that? through the tests maybe? I have no idea how before_dataset_index works, maybe you can share some examples of it so I can get the idea? @wardi Thanks

This could be future work separate from this PR, but would you check to make sure that `plugin_data` *doesn't* appear in the data stored in `validated_data_dict` https://github.com/ckan/ckan/blob/0a596b8394dbf9582902853ad91450d2c0d7959b/ckan/lib/search/index.py#L127 ?

TomeCirun:
I sent the data from the tests, plugin_data doesn't appear in the data stored in  validated_data_dict, but i m getting this error:

pysolr.SolrError: Solr responded with an error (HTTP 400): [Reason: Error:[doc=fe29f7f546eee02a6fed907051abe1e1]  Unknown operation for the an atomic update: plugin1]

from this [line](https://github.com/ckan/ckan/blob/0a596b8394dbf9582902853ad91450d2c0d7959b/ckan/lib/search/index.py#L308)

TomeCirun:
@wardi it seems the error comes from the nested dict, I also encountered the same error while using `ckanext-scheming` on CKAN 2.10

image

TomeCirun:
I opened an issue in their repo django-haystack/pysolr#385

TomeCirun:
@wardi polite reminder

ckan/tests/logic/action/test_create.py
```diff
@@ -1946,3 +1946,60 @@ def test_member_create_raises_if_object_type_is_invalid(self):
...
+            }
+        }
+        context = {
+            "user": user["name"],
+            "ignore_auth": False,
+        }
+        created_pkg = helpers.call_action(
+            'package_create', context=context, **pkg_dict
+        )
+        assert "plugin_data" not in created_pkg
```
wardi, line 2003: better to use the same sql as above to check for the existence of plugin_data

wardi:
`plugin_extras` appears in the migration filename, a comment and a couple variables. Would you update those names to `plugin_data` and make the small change to the test mentioned above?

TomeCirun:
> `plugin_extras` appears in the migration filename, a comment and a couple variables. Would you update those names to `plugin_data` and make the small change to the test mentioned above?


@wardi done :)
Thanks

ckan/tests/logic/action/test_create.py
```diff
@@ -2001,5 +2001,9 @@ def test_ignored_on_create_if_non_sysadmin(self):
             'package_create', context=context, **pkg_dict
         )
         assert "plugin_data" not in created_pkg
-        pkg = model.Package.get(created_pkg["id"])
-        assert pkg.plugin_data is None
+
+        plugin_daya_from_db = model.Session.execute(
```
wardi, line 2005: daya?

TomeCirun: It's a typo, tnx, good catch btw the failed test is not related to my changes

PR #6916, opened 2022-06-17
[6893]Delete User Web UI

### LINKED ISSUE #6893: Delete User Web UI
**CKAN version**
2.9.3

**Describe the bug**
I delete a user with the web-ui. Now I try to add a user with the same e-mail and I got an error the email already exists, but
the user is removed.

**Steps to reproduce**
Create a user an try to delete the user in the web ui. Then create a user with the same email address.

**Expected behavior**
Removing a user removes all data.

**Additional details**
Where is the e-mail saved? In the database I found nothing.

TomeCirun:
Fixes #6893 

### Proposed fixes:
I made the following changes:
* add extra filter when checking for email uniqueness
* add checkbox to reactivate the user (available only for sysadmin)
* remove delete button if the user is already deleted
* Update Profile button is changed to Reactivate Profile if the user is deleted

image

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/__init__.py
+++ b/ckan/logic/__init__.py
@@ -826,7 +826,7 @@ def _import_module_functions(
 
 
 @contextmanager
-def guard_agains_duplicated_email(email: str):
+def guard_against_duplicated_email(email: str):
     try:
         yield
     except exc.IntegrityError as e:
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1000,7 +1000,7 @@ def user_create(context: Context,
     upload.upload(uploader.get_max_image_size())
 
     if not context.get('defer_commit'):
-        with logic.guard_agains_duplicated_email(data_dict['email']):
+        with logic.guard_against_duplicated_email(data_dict['email']):
             model.repo.commit()
 
     # A new context is required for dictizing the newly constructed user in
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -823,7 +823,7 @@ def user_update(context: Context, data_dict: DataDict) -> ActionResult.UserUpdat
     upload.upload(uploader.get_max_image_size())
 
     if not context.get('defer_commit'):
-        with logic.guard_agains_duplicated_email(data_dict['email']):
+        with logic.guard_against_duplicated_email(data_dict['email']):
             model.repo.commit()
 
     author_obj = model.User.get(context.get('user'))


ckan/logic/validators.py
```diff
@@ -1001,8 +1001,9 @@ def email_is_unique(key: FlattenKey, data: FlattenDataDict,
     session = context['session']
 
     users = session.query(model.User) \
-        .filter(model.User.email == data[key]).all()
-    # is there is no users with this email it's free
+        .filter(model.User.email == data[key]) \
+        .filter(model.User.state == 'active').all()
+    # if there are no active users with this email, it's free
     if not users:
```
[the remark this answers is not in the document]
TomeCirun: image
is this what we need?
@smotornyuk

ckan/logic/action/create.py
```diff
@@ -1000,7 +1001,17 @@ def user_create(context: Context,
...
 
     if not context.get('defer_commit'):
-        model.repo.commit()
+        try:
+            model.repo.commit()
+        except exc.IntegrityError as e:
+            if e.orig.pgcode == _PG_ERR_CODE['unique_violation']:
+                session.rollback()
+                raise ValidationError(cast(ErrorDict, {
+                    'constraints': ['The email address \'{email}\' belongs to '
```
smotornyuk, line 1011: I think it's better to add an error to the `email` key, not to the `constraints`. In the WebUI it will attach the error message to the email field and, after all, the error is raised because of unacceptable email, so it will be fair

ckan/logic/action/create.py
```diff
@@ -1000,7 +1001,17 @@ def user_create(context: Context,
     upload.upload(uploader.get_max_image_size())
 
     if not context.get('defer_commit'):
-        model.repo.commit()
+        try:
+            model.repo.commit()
+        except exc.IntegrityError as e:
```
smotornyuk, line 1007: Can you extract this check into some small function, so that instead of these scary blocks `user_create` and `user_update` will do something like:

```
with logic.guard_agains_duplicated_email(data_dict["email"]):
    model.repo.commit()
```

smotornyuk:
> @smotornyuk I wasn't sure if we need the `email_is_unique` validator, so I left it there, if required I will remove it.
> 
> Thanks

Let's keep it. Every public function can be potentially used by the 3rd party extensions, so we have to announce the deprecation of any unused function and keep it for a couple of versions before removing it. I see no harm in this validator, so let's just leave it as is.

ckan/logic/__init__.py
```diff
@@ -137,6 +139,7 @@ def prettify(field_name: str):
                     assert isinstance(error, list)
                     summary[_('Tags')] = error[0]
                 else:
+                    breakpoint()
```
smotornyuk: Ooops..

ckan/migration/versions/102_ff13667243ed_create_conditinal_index_on_user.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_index("idx_only_one_active_email", "user", ["email", "state"], unique=True, postgresql_where=sa.text('"user".state=\'active\''))
+
+
+def downgrade():
+    op.drop_index("only_one_active_email")
```
smotornyuk: Shouldn't it be 
```
op.drop_index("idx_only_one_active_email")
```
, with the `idx_` prefix?

ckan/logic/action/create.py
TomeCirun: Yep, my bad

ckan/migration/versions/102_ff13667243ed_create_conditinal_index_on_user.py
TomeCirun: already fixed in https://github.com/ckan/ckan/pull/6916/commits/1d447fddb738eaccc058fbb13a0762b5292c525b

ckan/logic/__init__.py
TomeCirun: hehe i know, but you are looking at the older commit, already fixed in https://github.com/ckan/ckan/pull/6916/commits/1d447fddb738eaccc058fbb13a0762b5292c525b

TomeCirun:
hey @smotornyuk is there anything else that can be done in this PR ? 

Thanks

PR #6924, opened 2022-06-19
translate to spanish all installing folder docs

johnfelipe:
translate to spanish all installing folder docs

### DIFF
--- a/doc/maintaining/installing/index-es.rst
+++ b/doc/maintaining/installing/index-es.rst
@@ -1,0 +1,62 @@
+***
+
+## Instalación de CKAN
+
+Antes de que pueda usar CKAN en su propia computadora, debe instalarlo.
+Hay tres formas de instalar CKAN:
+
+\#. Instalar desde un paquete de sistema operativo
+\#. Instalar desde el origen
+\#. Instalar desde Docker Compose
+
+CKAN 2.9 soporta Python 3.7 y Python 3.8 más 2.7. La próxima versión de CKAN
+solo admitirá Python 3.
+
+Se pueden encontrar consejos de implementación adicionales en nuestro wiki, como los recomendados
+`Hardware Requirements <https://github.com/ckan/ckan/wiki/Hardware-Requirements>`\_.
+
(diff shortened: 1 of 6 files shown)


doc/maintaining/installing/index-es.rst
```diff
@@ -0,0 +1,62 @@
...
+
+La instalación desde el paquete es la forma más rápida y fácil de instalar CKAN, pero requiere
+Ubuntu 18.04 de 64 bits o Ubuntu 20.04 de 64 bits.
+
+**Debe instalar CKAN desde el paquete si**:
+
+*   Desea instalar CKAN en un servidor Ubuntu 18.04 o 20.04, de 64 bits, *y*
+*   Solo desea ejecutar un sitio web CKAN por servidor
+
+Véase :d oc:`install-from-package`.
```
amercader, line 26: All `:doc:` directives seem to be broken

doc/maintaining/installing/install-from-package-es.rst
```diff
@@ -0,0 +1,206 @@
+.. include:: /\_substitutions.rst
```
amercader, line 1: This should not be escaped:
```suggestion
.. include:: /_substitutions.rst
```

In general, none of the Sphynx directives should be touched at all

amercader:
Thanks a lot for this @johnfelipe , it would be very valuable to have these docs available in different languages. But we need to find a way for properly managing translations, including how to setup [Read the Docs](https://docs.readthedocs.io/en/stable/localization.html), keeping translations updated, reviews etc

Did you use an automated translated service to get the Spanish version? There seem to be a lot of broken Sphynx directives and extra characters.

@johnfelipe if you need to update some po files please do so in [Transifex](https://www.transifex.com/okfn/ckan) instead.

If anyone has experience managing documentation translations in an open source project we'd be very keen on hearing from you.

doc/maintaining/installing/index-es.rst
johnfelipe: pls mark all suggestions

PR #6926, opened 2022-06-20
Fixes two errors when dealing with an encoded url.

EricSoroos:
The 2.9 version of #6685

* url in question /%EF%AC%81?foo=bar&bz=%AC%81
* This is a unicode character, which can't be decoded from
ascii. Jinja templates will handle this if it's unicode, or if it's
hex encoded ascii, but can't take a non-unicode string in python 2 and
put this in a template.
* The querystring was being quoted, which is incorrect, as:
  1) the special characters in the query string mean something
  2) The rest of the querystring is already quoted. This makes it
  double quoted, as seen in the datastore file
* We don't want to unquote urls before putting them in the template
anyway.
The solution here is to make sure it's unicode passed into the
function.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -654,7 +654,7 @@ def full_current_url():
 @core_helper
 def current_url():
     ''' Returns current url unquoted'''
-    return unquote(request.environ['CKAN_CURRENT_URL'])
+    return request.environ['CKAN_CURRENT_URL']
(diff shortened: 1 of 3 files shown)


EricSoroos:
@amercader There's an additional tweak here that fixes a site that was still having issues filtering on multibyte characters.

PR #6936, opened 2022-06-24
f-strings

TomeCirun:
As we all working on improving CKAN 2.10 with the latest python tools/technologies to make it more readable, less code, more user friendly, etc I thought we can replace all format() with the f-strings to make it even better. Please If anybody doesn't like the idea, feel free to close this PR.

f-Strings: A New and Improved Way to Format Strings in Python.
* They joined the party in Python 3.6. You can read all about it in [PEP 498](https://www.python.org/dev/peps/pep-0498/), which was written by Eric V. Smith in August of 2015.
* Also called “formatted string literals,” f-strings are string literals that have an f at the beginning and curly braces containing expressions that will be replaced with their values. The expressions are evaluated at runtime and then formatted using the __format__ protocol. As always, the [Python docs](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) are your friend when you want to learn more.
* f-strings have a simple syntax than their predecessors and are easy readable
* Because f-strings are evaluated at runtime, you can put any and all valid Python expressions in them. This allows you to do some nifty things.
* You can have multiline f-strings
* f-strings are faster than both %-formatting and str.format(), f-strings are expressions evaluated at runtime rather than constant values. Here’s an excerpt from the docs:
“F-strings provide a way to embed expressions inside string literals, using a minimal syntax. It should be noted that an f-string is really an expression evaluated at run time, not a constant value. In Python source code, an f-string is a literal string, prefixed with f, which contains expressions inside braces. The expressions are replaced with their values.” ([Source](https://www.python.org/dev/peps/pep-0498/#abstract))

Here’s a speed comparison:
image

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/authz.py
+++ b/ckan/authz.py
@@ -238,7 +238,7 @@ def is_authorized(action: str, context: Context,
                 name = auth_function.__name__
             return {
                 'success': False,
-                'msg': 'Action {0} requires an authenticated user'.format(name)
+                'msg': f'Action {name} requires an authenticated user'
             }
 
         return auth_function(context, data_dict or {})
--- a/ckan/cli/__init__.py
+++ b/ckan/cli/__init__.py
@@ -132,12 +132,11 @@ def load_config(ini_path: Optional[str] = None) -> Config:
                 break
         if not filename:
             # give really clear error message for this common situation
-            msg = u'''
+            msg = f'''
 ERROR: You need to specify the CKAN config (.ini) file path.
 
 Use the --config parameter or set environment variable CKAN_INI
-or have one of {} in the current directory.'''
-            msg = msg.format(u', '.join(default_filenames))
+or have one of {', '.join(default_filenames)} in the current directory.'''
             raise CkanConfigurationException(msg)
 
     if not filename or not os.path.exists(filename):
@@ -147,6 +146,6 @@ or have one of {} in the current directory.'''
 
     config_loader = CKANConfigLoader(filename)
     loggingFileConfig(filename)
-    log.info(u'Using configuration file {}'.format(filename))
+    log.info(f'Using configuration file {filename}')
 
     return config_loader.get_config()
--- a/ckan/cli/cli.py
+++ b/ckan/cli/cli.py
@@ -183,16 +183,14 @@ def _get_commands_from_entry_point(entry_point: str = 'ckan.click_command'):
     registered_entries = {}
     for entry in iter_entry_points(entry_point):
         if entry.name in registered_entries:
+            first=registered_entries[entry.name].dist
             error_shout((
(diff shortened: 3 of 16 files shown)


wardi:
If we're writing new code or making changes to some existing code then I'm all for using f-strings.

But I'm -1 on replacing existing `format()` calls because of the conflicts that are created when backporting fixes. Also it's a manual task that could quietly introduce bugs or incompatibilities, especially when combined with gettext.

TomeCirun:
I agree with you on the latter,  PEP 498 states each formatting method “have their advantages, but in addition have disadvantages” so my intention was not to get into such problems and the format method in such cases remains the only one useable.

Regarding the backports, I also thought that we gonna drop py2 support on ckan 2.9 version in the near future, but I was wrong though.
I will continue exploring and see if I can find something similar to 
https://github.com/asottile-archive/future-fstrings
so we can use it on CKAN 2.9 and when it comes to backporting there won't  be any issues.

@wardi

smotornyuk:
I'll join Ian here. 
I'd like to have the codebase modernized and I'm constantly trying to sneak new features into it. But literally, every time I have added some fancy formatting when it has no functional meaning, I cursed myself afterward, trying to backport one of the PRs into an older release. It creates conflicts that slow down the process of preparing a new release, and sometimes it requires you to adapt a fix, write more tests, and so on.

So here are recommendations that were discovered through my suffering: 

* change as few lines of code as possible(here you have to change ~500 lines through ~90 files)
* do it only if it improves the project in a visible way(apart from brevity, f-strings give you a speed boost, which won't be noticed because we have much slower HTML-rendering/io delays/SQL-queries/etc)
* If you are moving code / rewriting a feature, change whatever you want(recently I've moved activity-feature into a separate plugin and I was really satisfied, applying black to its code)


PS for logging, aka `log.info(f"hello {variable}")`, it's better to use `log.info("hello %s", variable)` form. As far as I remember, it guarantees that values are evaluated only if it necessary(depending on log level). This means this form should be even faster than f-string, as long as you are not setting log-level=debug.

ThrawnCA:
> Regarding the backports, I also thought that we gonna drop py2 support on ckan 2.9 version in the near future, but I was wrong though.

I think there is a difference between dropping support in the sense of "new versions will no longer run on Python 2", vs dropping support in the sense of "there will be no backported fixes for any version that might run on Python 2". They can easily be several years apart.

Zharktas:
Unless policy for supported versions is changed, CKAN 2.8 and 2.9 would be supported for a long time, at least a year for each after 2.10 is released. Future 2.8 and 2.9 releases should, at least in theory, support python 2.7. Of course we can't control if installing python 2.7 environments and packages becomes impossible over the years.

PR #6939, opened 2022-06-25
Added validation message when invalid url on dataset search query e.g with comma is given

### LINKED ISSUE #6910: Appending a comma to dataset sort parameter causes itnernal server error
**CKAN version**
master

**Describe the bug**
Works: https://demo.ckan.org/dataset/?q=&sort=title_string+asc
Breaks: https://demo.ckan.org/dataset/?q=&sort=title_string+asc,

**Steps to reproduce**
See above

**Expected behavior**
Either a sensible error message or no sorting

**Additional details**
Breaks here: https://github.com/ckan/ckan/blob/master/ckan/views/dataset.py#L245
`sort_by_fields = [field.split()[0] for field in sort_by.split(u',') if field]` would probably "fix" it by ignoring empty fields.

hussnainwithss:
Fixes #6910 

### Proposed fixes:
- Fixes 500 internal error when a comma is appended at end of dataset query e.g `http://localhost:8080/dataset/?q=&sort=title_string+asc,`
- Now shows proper validation error with a 400 bad request response.

### Features:

- [ ] includes API changes


smotornyuk:
Can you report this problem to the user in a more gentle way? For example, if [an invalid Solr query is used](https://demo.ckan.org/dataset?q=a%3A%22b&sort=score+desc%2C+metadata_modified+desc), the user still sees the search page with a message, that error happened. I think that the wrong sorting order should produce either the same or even more clear error message.

PR #6950, opened 2022-07-01
Document the interactions of multiple plugins overriding the same template

ThrawnCA:
- ckan_extends extends the next plugin instead of the default
- New nested blocks can be defined for use by child templates in other plugins
- Plugin ordering is very important

### Features:

- [ ] includes API changes


ThrawnCA:
This is a bit of an info dump, but the interactions between multiple plugins really do need to be documented...I'm open to adjusting it or moving it to a better location.

ThrawnCA:
https://stackoverflow.com/questions/52936903/ckan-extensions-overwrite-each-other-what-is-the-correct-order describes some of this.

ThrawnCA:
Bump. This behaviour is complex enough that it needs to be documented in some fashion.

PR #6954, opened 2022-07-04
[#6948] Avoid storing sessions on each request

### LINKED ISSUE #6948: Unsustainable Beaker cache growth on CKAN 2.9
**CKAN version**

2.9 (tested on 2.9.0 and 2.9.5)

Does not occur on 2.8.8.

**Describe the bug**

Beaker session cache files are created on every unauthenticated request to non-API URLs. This did not occur under CKAN 2.8.

**Steps to reproduce**

* Create a CKAN 2.9 instance on a Linux system.
* Check total inode usage with `df -i`
* Send a request to /dataset, eg `curl http://my-site/dataset/`
* Check total inode usage again and confirm that it has risen by 2.
* Check the session files in `/tmp/*/sessions` to confirm that two new files have been created, `find /tmp/*/sessions -type f -cmin -5`

**Expected behavior**

Unauthenticated requests should not create Beaker session files, as per CKAN 2.8 behaviour.

**Additional details**

This can easily lead to hundreds of thousands of files being created per day, rapidly reaching the inode limit of the hard disk. It could also be deliberately exploited as a Denial of Service attack.

amercader:
Fixes #6948 

See https://github.com/ckan/ckan/issues/6948#issuecomment-1172152804 for details. We were storing a session on each request because Beaker always adds some default keys.

This is 2 files per request, and considering we serve assets via flask by default that mean dozens of files per request

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -10,7 +10,7 @@ import pkgutil
 import logging
 
 from logging.handlers import SMTPHandler
-from typing import Any, Iterable, Optional, Union, cast
+from typing import Any, Iterable, Optional, Union, cast, Dict
 
 from flask import Blueprint, send_from_directory
 from flask.ctx import _AppCtxGlobals
@@ -134,7 +134,7 @@ class BeakerSessionInterface(SessionInterface):
     def save_session(self, app: Any, session: Any, response: Any):
         session.save()
 
-    def is_null_session(self, obj: object) -> bool:
+    def is_null_session(self, obj: Dict[str, Any]) -> bool:
 
         is_null = super(BeakerSessionInterface, self).is_null_session(obj)
 
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -126,6 +126,29 @@ def _ungettext_alias():
     return dict(ungettext=ungettext)
 
 
+class BeakerSessionInterface(SessionInterface):
+    def open_session(self, app: Any, request: Any):
+        if 'beaker.session' in request.environ:
+            return request.environ['beaker.session']
+
+    def save_session(self, app: Any, session: Any, response: Any):
+        session.save()
+
+    def is_null_session(self, obj: object) -> bool:
+
+        is_null = super(BeakerSessionInterface, self).is_null_session(obj)
+
+        if not is_null:
+            # Beaker always adds these keys on each request, so if these are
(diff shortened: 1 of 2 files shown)


ckan/config/middleware/flask_app.py
```diff
@@ -126,6 +126,29 @@ def _ungettext_alias():
...
+
+    def is_null_session(self, obj: object) -> bool:
+
+        is_null = super(BeakerSessionInterface, self).is_null_session(obj)
+
+        if not is_null:
+            # Beaker always adds these keys on each request, so if these are
+            # the only keys present we assume it's an empty session
+            is_null = (
+                sorted(obj.keys()) == [
```
amercader, line 145: @smotornyuk can you help me out with the typing of this? I added the `object` type definition to the method because that what the parent Flask class uses but this gives the following [typing error](https://github.com/ckan/ckan/runs/7178711102?check_suite_focus=true#step:7:18):

```
   /home/runner/work/ckan/ckan/ckan/config/middleware/flask_app.py:145:28 - error: Cannot access member "keys" for type "object"
    Member "keys" is unknown (reportGeneralTypeIssues)
```

I tried various combinations of defining `obj` as a dict but failed, as my typing knowledge is limited

amercader: Thanks a lot for the super-detailed explanation, I learnt a lot!

ckan/config/middleware/flask_app.py
```diff
@@ -126,6 +126,29 @@ def _ungettext_alias():
...
+    def is_null_session(self, obj: Dict[str, Any]) -> bool:
+
+        is_null = super(BeakerSessionInterface, self).is_null_session(obj)
+
+        if not is_null:
+            # Beaker always adds these keys on each request, so if these are
+            # the only keys present we assume it's an empty session
+            is_null = (
+                sorted(obj.keys()) == [
+                    "_accessed_time", "_creation_time", "_domain", "_path"]
```
ThrawnCA, line 146: This should also handle the possibility of setting `save_accessed_time` to false (not storing the accessed time).

amercader: @ThrawnCA the keys are always added to the session object, regardless of the `save_accessed_time` setting: 

https://github.com/bbangert/beaker/blob/d4ab7d37387feddd0d544792d9e32533e39ff868/beaker/session.py#L222-L224

I double checked manually, so I think this works as intended

PR #6956, opened 2022-07-05
Prevent non-sysadmin users to change their own state

amercader:
Only sysadmins can update a user state

### Features:

### DIFF
--- a/ckan/views/user.py
+++ b/ckan/views/user.py
@@ -720,10 +720,10 @@ class PerformResetView(MethodView):
             updated_user = logic.get_action("user_update")(context, user_dict)
             # Users can not change their own state, so we need another edit
             if (updated_user["state"] == model.State.PENDING):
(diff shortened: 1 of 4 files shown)


ckan/views/user.py
```diff
@@ -717,8 +717,17 @@ def post(self, id: str) -> Union[Response, str]:
             if (username is not None and username != u''):
                 user_dict[u'name'] = username
             user_dict[u'reset_key'] = g.reset_key
-            user_dict[u'state'] = model.State.ACTIVE
-            logic.get_action(u'user_update')(context, user_dict)
+            updated_user = logic.get_action("user_update")(context, user_dict)
+            # Users can not change their own state, so we need another edit
+            if (updated_user["state"] == model.State.PENDING):
```
tino097, line 722: some strings are left unprefixed, is that on purpose?

PR #6961, opened 2022-07-06
remove `rank` from datastore_search()

### LINKED ISSUE #6172: datastore_search full_text and fields parameters not processed properly
**CKAN version**
master

**Describe the bug**
When the new `full_text` parameter is used along with the `fields` parameter, `datastore_search` always returns the entire record, along with the "hidden" rank column (used for relevance in fts search)

**Steps to reproduce**
Use `datastore_search` `full_text` and `fields` in the same call.

**Expected behavior**
`datastore_search` will process `fields` properly when `full_text` is specified, and the hidden rank value should not be returned.

robin-NEC:
Fixes #6172 

### Proposed fixes:
- Remove the hidden `rank` field from response, when `full_text` and `fields` parameters are passed in the same API call.

### Features:

- [X] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1901,7 +1901,7 @@ class DatastorePostgresqlBackend(DatastoreBackend):
         elif fields and fts_q:
             field_ids = datastore_helpers.get_list(fields)
             all_field_ids = list(fields_types.keys())
-            all_field_ids.remove('rank')
+            all_field_ids.remove("rank")
             field_intersect = [x for x in field_ids
                                if x not in all_field_ids]
             field_ids = all_field_ids + field_intersect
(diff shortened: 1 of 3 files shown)


ckanext/datastore/backend/postgres.py
```diff
@@ -1901,6 +1901,7 @@ def datastore_search(
         elif fields and fts_q:
             field_ids = datastore_helpers.get_list(fields)
             all_field_ids = list(fields_types.keys())
+            all_field_ids.remove("rank")
```
amercader, line 1904: Users might rely on `rank` being returned, after all it determines the order of the returned records so there should be a way to retrieve it. 
Let's keep `rank` in the returned fields if it is defined in the incoming `fields` parameter, eg something like


```suggestion
            if "rank" not in fields:
                all_field_ids.remove("rank")
```

And please add a `changes/6961.misc` file explaining the change in behaviour (eg The "rank" field is no longer returned in `datastore_search` results unless explicitly defined in the `fields` parameter)

robin-NEC: @amercader , thanks for your response, will do the changes as you suggested.

PR #6980, opened 2022-07-21
Update release policy documentation

amercader:
I made some changes to the documentation to hopefully make clear what versions are supported and which aren't and reflect recent changes in number of old versions supported.

On this PR:

* Move release information to its own dedicated page.
* Highlight the actual latest patch release on the top of the releases page (and potentially the install / upgrade pages). I think it's important that we are clear in that "we support 2.9.5" instead of "we support 2.9"
* Limit supported versions to just current minor plus a previous one.(so currently 2.9 and 2.8)
* **Update**: Added the supported versions box to the install and upgrade pages as well

Here's how the changes look in RTD: http://docs.ckan.org/en/update-release-docs/maintaining/releases.html
(**note**: Looks like RTD is struggling today so I added screenshots of the most relevant parts below.)

Apart from the documentation changes, I'll like to write a blog post explaining the policy and describing specifically what happens to each version. I'll share the draft before publishing but here are my proposed points:

* Going forward we will support only one previous minor version (so 2.9 and 2.8 now, 2.10 and 2.9 once 2.10 is out)
* Use the latest patch release!
* We no longer support 2.7
* For 2.8 we can decide, either we stick to the policy and stop supporting it when 2.10 is out (realistically end of summer / fall) or we extend 2.8 support a bit until the end of the year (1, maybe 2 more patch releases)
* Python 2.7 support for CKAN 2.9 dropped at the end of the year (TODO discuss)

----

Screenshot 2022-07-26 at 12-22-33 CKAN releases — CKAN 2 10 0a0 documentation

Screenshot 2022-07-21 at 13-29-37 CKAN releases — CKAN 2 10 0a0 documentation

### DIFF
--- a/doc/_supported_versions.rst
+++ b/doc/_supported_versions.rst
@@ -1,0 +1,6 @@
+.. note::
+    The currently supported CKAN version is |latest_release_version_format| 
+
+    Security and performance fixes are also provided for |previous_release_version_format|.
+
+    Read more about :ref:`officially supported versions <supported_versions>`
--- a/doc/_supported_versions.rst
+++ b/doc/_supported_versions.rst
@@ -1,3 +1,6 @@
+
+.. include:: /_substitutions.rst
+
 .. note::
     The currently supported CKAN version is |latest_release_version_format| 
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -124,7 +124,7 @@ release = ckan.__version__
 version_re = None
 point_releases_ = None
 
-SUPPORTED_CKAN_VERSIONS = 3
+SUPPORTED_CKAN_VERSIONS = 2
 
 
 def get_release_tags():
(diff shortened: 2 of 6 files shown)


amercader:
Tests fixed now

amercader:
@smotornyuk can you have a quick look? I'd like to publish the blog post and need to link to these docs. Thanks!

smotornyuk:
Yes, sorry for the delay.

Looks perfect, I have no objections.
There is one potential problem in the following line: 
```python
previous_tag_prefix = f"ckan-{current_version[0]}.{int(current_version[1]) - 1}"
```

In release 3.0.0 this logic won't point to the real version. But I see no reason to fix it now because:
* release 3.0.0 means quite a big change, so we may even review the support policy once again
* this problem will be noticed straight away because it'll produce IndexError. So it will be fixed anyway.

PR #6982, opened 2022-07-22
fix wrong zh_Hans_CN character in ckan.po

FuhuXia:


amercader:
Thanks @FuhuXia, can you make the change in Transfix, that where we manage the translations, and you can patch more than one version:

https://www.transifex.com/okfn/ckan/

FuhuXia:
Change did on transifex. This can be closed.

PR #6991, opened 2022-07-28
[6987]Fix Inconsistent behaviuor in non-authorized errors

### LINKED ISSUE #6987: Inconsistent behaviuor in non-authorized errors (2.10)
**CKAN version**
2.10 (master)

**Describe the bug**

Following up from https://github.com/ckan/ckan/pull/6560, here's what happens when a user is not authorized to perform an action on the UI:

| CKAN Version | Logged in user | HTTP status, result |
| --- | --- | --- |
| <= 2.9 (tested 2.9 and 2.8) | No | 403 - Not authorized message
| <= 2.9 (tested 2.9 and 2.8) | Yes | 403 - Not authorized message
| 2.10, default behaviour | No | 302 + 200, Redirect to login page
| 2.10, default behaviour | Yes |  403 - Not authorized message
| 2.10, if `ckan.redirect_to_login_if_not_authorized` is False | No | 403 - Not authorized message
| 2.10, if `ckan.redirect_to_login_if_not_authorized` is False | Yes | 403 - Not authorized message

It seems that at the very least, we should 

1. fix the inconsistency between logged in users and non-logged in users in 2.10. Looks like we don't need to check `if current_user.is_anonymous` here:

https://github.com/ckan/ckan/blob/ed155dca25265562e1a9125ee6917b832d514b10/ckan/config/middleware/flask_app.py#L542

2. Decide if we want to change the default behaviour to redirect to the login page or keep the 403 error.

TomeCirun:
Fixes #6987 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -20,8 +20,6 @@ from flask_multistatic import MultiStaticFlask
 from werkzeug.exceptions import (
     default_exceptions,
     HTTPException,
-    Unauthorized,
-    Forbidden
 )
 from werkzeug.routing import Rule
 from werkzeug.local import LocalProxy
@@ -38,7 +36,7 @@ from ckan.lib import helpers as h
 from ckan.lib import jinja_extensions
 from ckan.lib import uploader
 from ckan.lib import i18n
-from ckan.common import config, g, request, ungettext, current_user, _
+from ckan.common import config, g, request, ungettext, _
 from ckan.config.middleware.common_middleware import (TrackingMiddleware,
                                                       HostHeaderMiddleware,
                                                       RootPathMiddleware)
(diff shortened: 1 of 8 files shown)


TomeCirun:
@amercader ,  I deleted all logic that was related to the previous behaviour, and I added a login link to the error_document_template.html but it first checks if the user is anonymous, let me know your opinion over this.

Thanks

ckan/templates/error_document_template.html
```diff
@@ -10,6 +10,13 @@ <h1>{{ code }} {{ name }}</h1>
       {% endif %}
       {{ content}}
     </div>
+    {% block login_redirect %}
+    {% if current_user.is_anonymous %}
+    <div class="module-content">
```
amercader, line 15: This message should only be shown in the error page if the raised error was a 403 or 401, not when the user is anonymous. So keep the following check in the `error_handler()` and if it matches you can pass an extra `show_login_redirect_link` var to show this block:

```python
                    if current_user.is_anonymous and type(e) in (
                        Unauthorized, Forbidden
                    ):
```

Also the text needs to be translatable so please use something along these lines:
```
{{ _("You might need to login to access this page.") }} <a href="{{ h.url_for('user.login') }}">{{ _("Click here to login") }}</a> 
```

TomeCirun: thanks for this, it looks better, you can check my changes :) @amercader

ckan/config/middleware/flask_app.py
```diff
@@ -565,41 +565,23 @@ def error_handler(e: Exception) -> Union[
...
-                u'name': e.name
+                u'name': e.name,
+                u'show_login_redirect_link': False
             }
 
+            if current_user.is_anonymous and type(e) in (
+                        Unauthorized, Forbidden
+                    ):
+                extra_vars.update({"show_login_redirect_link": True})
+
```
amercader, line 584: This is a bit more concise:

```python
show_login_redirect_link = current_user.is_anonymous and type(e) in (Unauthorized, Forbidden)
extra_vars = {
    'code': e.code,
    'content': e.description,
    'name': e.name
    'show_login_redirect_link': show_login_redirect_link,
}
```

amercader (commented):
Great work @TomeCirun, just some minor suggestions

TomeCirun:
> Great work @TomeCirun, just some minor suggestions

Done :)

TomeCirun:
@amercader  this is ready to go

PR #7012, opened 2022-08-15
[7008]Fix ckan generate config debug issue

### LINKED ISSUE #7008: ckan generate config debug issue
**CKAN version**
master

**Describe the bug**
`ckan generate config` exports a configuration including `debug = false` in the `[app:main]` section of the ini file. This setting overrides any debug setting in `[DEFAULT]` which is where it normally occurs, and where okfn/docker-ckan adds the `debug = true` setting for development ckan environments

**Steps to reproduce**
Steps to reproduce the behavior: run `ckan generate config`

**Expected behavior**
`debug = false` should be in the `[DEFAULT]` section not `[app:main]`

**Additional details**

TomeCirun:
Fixes #7008 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/server.py
+++ b/ckan/cli/server.py
@@ -9,7 +9,7 @@ from typing import Optional
 import click
 from werkzeug.serving import run_simple
 
-from ckan.common import config
+from ckan.common import config, asbool
 from . import error_shout
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


TomeCirun:
@smotornyuk can you tell if this is the right way to do it?

Thanks

smotornyuk:
I prefer to avoid using this approach because debug option slips out standard config option workflow: it's not validated anymore and not auto-documented.

I'll take a look at the issue and check the complexity of handling configuration sections via config declarations. If it's too complex, let's go with your solution(but we need to update the documentation in this case).
So, for now, I'll put this PR on hold and ping you after this small research.

TomeCirun:
any news on this one ? @smotornyuk

smotornyuk:
I've created PR #7105, but it's not discussed/reviewed yet, so it may be not the final version yet)

EricSoroos:
This has been superseded by #7105

PR #7015, opened 2022-08-17
Including Private Dataset Count

### LINKED ISSUE #5786: Private datasets are not counted in Group dataset count even if the user has access to the datasets
**CKAN version**
2.9.1

**Describe the bug**
Private datasets do not show in the Groups count in the Group list page, even if I have access to the private dataset.  If I click through the Group, the private datasets are listed as expected.

**Steps to reproduce**
1. Create a private dataset and add it to a group.
2. Go to the Groups page and take note of the dataset count for that group.  It's not counting the just added private dataset.
3. Click through the group, the private dataset is listed.

**Expected behavior**
Similar to other list pages (dataset, organization), the group list page should include in the count, private datasets the current user has access to.

Deep-NEC:
Fixes #5786

### Proposed fixes:
This PR provides private dataset count in group list dataset count.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1895,7 +1895,7 @@ def package_search(context: Context, data_dict: DataDict) -> ActionResult.Packag
         data_dict.setdefault('fq', '')
 
         # Remove before these hit solr FIXME: whitelist instead
-        include_private = asbool(data_dict.pop('include_private', False))
+        include_private = asbool(data_dict.pop('include_private', True))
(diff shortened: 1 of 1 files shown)


amercader:
This is not the right place for this fix as you are modifying the behaviour of `include_private` on all dataset searches. Instead see how the `views/group.py` blueprint (`index` method) requests the list of groups to be rendered (probably calling the `group_list` action and see what's going on there.

PR #7018, opened 2022-08-18
Added `ckan.record_private_activity` check for private package activi…

JVickery-TBS:
…ty stream item creations.

Fixes #

Add config check to allow users to record package activity streams for private packages. Still does not create activity stream items by default.

### Proposed fixes:

This feature could be patched back into version 2.9, into the `logic.actions.(update | create | delete)`

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7018.feature
+++ b/changes/7018.feature
@@ -1,0 +1,1 @@
+Added `ckan.record_private_activity` config to allow for private package activity streams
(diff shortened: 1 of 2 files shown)


amercader:
Thanks @JVickery-TBS, but while this would enable activities on private datasets there wider implications in terms of access to unauthorized activities that the current system wouldn't really address, like eg creating a dataset private and then making it public, allowing certain activities to be public but not others etc. We probably need a system similar to the Permission labels for datasets
We feel that there needs to be a bigger discussion and proper specification of this feature than just enabling them

PR #7020, opened 2022-08-19
Add spatial field to Solr schema

avdata99:
### Features:
The current Solr schema do not allow indexing geometries.
Our [CKAN Solr image](https://github.com/ckan/ckan-solr) relies on this repo for the schema. To allow any extension to use spatial features, update this Solr schema is required.

- [ ] includes API changes

### DIFF
--- a/ckan/config/solr/schema.xml
+++ b/ckan/config/solr/schema.xml
@@ -96,6 +96,13 @@ attribute with the form `ckan-X.Y` -->
       </analyzer>
     </fieldType>
 
+    <fieldType 
+        name="location_rpt" class="solr.SpatialRecursivePrefixTreeFieldType"
+        spatialContextFactory="com.spatial4j.core.context.jts.JtsSpatialContextFactory"
+        autoIndex="true" distErrPct="0.025" maxDistErr="0.000009"
+        distanceUnits="degrees"
+    />
+
 </types>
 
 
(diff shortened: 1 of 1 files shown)


amercader:
@avdata99 let's finish all the planned work on ckanext-spatial and then we can propose changes in core if necessary

ckan/config/solr/schema.xml
```diff
@@ -96,6 +96,13 @@ attribute with the form `ckan-X.Y` -->
       </analyzer>
     </fieldType>
 
+    <fieldType 
+        name="location_rpt" class="solr.SpatialRecursivePrefixTreeFieldType"
```
EricSoroos, line 100: I'm curious why you chose this instead of the `RptWithGeometrySpatialField` model.

avdata99: Is the oe from ckanext-spatial [docs](https://github.com/ckan/ckanext-spatial/blob/master/doc/spatial-search.rst#choosing-a-backend-for-the-spatial-search) (_solr-spatial-field_) recommends @EricSoroos.

amercader: @EricSoroos we are early in our investigation of the different Solr fields and options so we just went along with the first option suggested in the [Solr docs](https://solr.apache.org/guide/8_11/spatial-search.html#rpt). Do you have experience working with these fields? Seems like `RptWithGeometrySpatialField` provides greater accuracy at the expense of a bigger index size (because it stores geometries)?

avdata99:
Cleaning, too old

PR #7021, opened 2022-08-21
[7016]Provide a mechanism to support CSRF on extensions that support older CKAN version

### LINKED ISSUE #7016: Provide a mechanism to support CSRF on extensions that support older CKAN version
@TomeCirun @pdelboca 

After #6863, [best practices](https://github.com/ckan/ckan/blob/master/doc/extensions/best-practices.rst) say:

> To add CSRF protection to your extensions use the CSRF macro in all your forms

> {% import 'macros/form.html' as form %}
> ...
> {{ form.csrf_protection() }}

This won't work on CKAN<=2.9 because that macro doesn't exist. Similarly, trying to add the input directly like in [here](https://github.com/ckan/ckan/pull/6863/files#diff-09d2291ff01f465bfadf71b865872b83bd0301fc32f72e3fc96964a74a1619b0R7) will fail because the `csrf_token()` function is not defined.

My suggestion would be:

1. Consolidate all CSRF inputs into a single `{{ h.csrf_input() }}` call (both the ` {{ form.csrf_protection() }}` ones and the direct ` <input type="hidden" ...` ones). Using a helper will allow extensions to override it if necessary. Then we can use the same template markup everywhere.
2. Provide `h.csrf_input()` stubs on CKAN<2.9 that don't actually return anything. This could be done either:
   2.1. Patching CKAN core so 2.9.6 and 2.8.11 register `csrf_input()` template helpers, but they return an empty string
   2.2 Something similar but using ckantoolkit. I think this would be messier and even impossible.

At the end of this, an extension wanting to support CKAN<=2.9 and CKAN 2.10 with CSRF protection can just add the same code as core, and it will work across CKAN versions without messy checks on each extension:

```
  <form class="dataset-form add-member-form" method='post'>
    {{ h.csrf_input() }}
```

Does this make sense?

**edit**: updated to use a template helper instead of a Jinja tag

TomeCirun:
Fixes #7016 

### Features:

- [ ] includes API changes


TomeCirun:
@amercader please check this [PR](https://github.com/ckan/ckanext-scheming/pull/342) too

ckan/plugins/toolkit.py
```diff
@@ -99,7 +100,7 @@
     "mail_recipient", "mail_user",
     "render_snippet", "add_template_directory", "add_public_directory",
     "add_resource", "add_ckan_admin_tab",
-    "check_ckan_version", "requires_ckan_version", "get_endpoint",
+    "check_ckan_version", "requires_ckan_version", "get_endpoint", "csrf_input"
```
amercader, line 103: No need to add this to the toolkit, it's already accessible to templates (and to python code) via the helper.

ckan/lib/helpers.py
```diff
@@ -2781,3 +2781,8 @@ def make_login_url(
         parsed_base = parsed_base._replace(netloc=netloc, query=urlencode(md))
         return urlunparse(parsed_base)
     return base
+
+
+@core_helper
+def csrf_input():
+    return snippet('snippets/csrf_token.html')
```
amercader, line 2788: Can you rename this snippet to `csrf_input.html` to be consistent?

amercader (commented):
Looks great @TomeCirun 
Can you submit two PRs to `dev-v2.9` and `dev-v2.8`  adding a `csrf_input` helper that just returns an empty string?

This way extensions won't have to implement checks like this one:

https://github.com/ckan/ckanext-scheming/pull/342/files#diff-6972c64a4aad16aea70733c8636ce62ed4a8546409d7fe8210e3ada10cc7f5f9R438

TomeCirun:
everything's done @amercader

amercader:
Thanks @TomeCirun !

PR #7024, opened 2022-08-22
Added `organization_show` action call into loop of orgs for the retur…

JVickery-TBS:
…n of `organization_list_for_user` action.

Fixes #

`organization_list_for_user` action does not return processed org schemas.

### Proposed fixes:

Implement reusable code to process org/group schemas (such as shown in `_group_or_org_show`)?

Basically we do not wish to have the `organization_list_for_user` action do DB executions on the group table, and then loop through the orgs to just redo very similar DB executions on the same table.

Doing this on the dictization level does not really allow for schema integration. Nor would it easily be able to implement the `IOrganizationController.read` and `IGroupController.read`.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -743,7 +743,12 @@ def organization_list_for_user(context: Context,
     context['with_capacity'] = True
     orgs_list = model_dictize.group_list_dictize(orgs_and_capacities, context,
         with_package_counts=asbool(data_dict.get('include_dataset_count')))
-    return orgs_list
+    return_list = []
+    for org in orgs_list:
+        return_list.append(logic.get_action('organization_show')(
+                                context,
+                                {'id': org['id']}))
+    return return_list
 
 
 def license_list(context: Context, data_dict: DataDict) -> ActionResult.LicenseList:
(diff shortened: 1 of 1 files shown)


JVickery-TBS:
@amercader , @wardi and I have been trying to figure out the best way to do this. I have some execution times here to try to give you a sense of the affect of this code. These tests were done with our list of organizations which is a list of around 350 orgs.

- Original code with just the `group_list_dictize`: ~ 0.03324 seconds
- Code with the `group_list_dictize` + the loop into `organization_show:` ~ 13.4711 seconds
- Code without `group_list_dictize` and just looping `orgs_and_capacities`: ~ 11.1222360134 seconds
- Code without `group_list_dictize` and just looping `orgs_and_capacities` (with `include_dataset_count=False, include_datasets=False, include_users=False, include_groups=False, include_tags=False, include_followers=False`): ~ 1.8 seconds

I personally don't feel like these are great numbers. And I have now noticed that we also need to fix the `model.package.get_groups` and `model.user.get_groups` not running the schemas and the `iGroupController.read`/`iOrganizationController.read`.

My original idea was to include `_translated` group fields at a dictization level. However, I am not too sure if that is a good approach, and that would not easily allow for group schemas to be brought in (as far as my knowledge goes).

In theory, I suppose that the schemas and plugin implementations should be executed in all of the list get actions. I.e. `package_list`, `group_list`, and `organization_list`.

Wondering if you have any thoughts and ideas on this? And if this is even something that should actually happen?

PR #7036, opened 2022-08-24
6641 make private dataset behaviour more configurable

ThrawnCA:
Fixes #6641 

### Proposed fixes:

Add a config flag to control whether authorisation errors for reading private datasets and resources should result in a 404 Not Found (False, default, existing behaviour) or whether they should acknowledge the object's existence and return either a 403 Forbidden or a redirect to the login page (True).

### Features:

- [ ] includes API changes


ckan/tests/controllers/test_group.py
```diff
@@ -424,21 +424,20 @@ def test_membership_edit_page(self, current_user, app):
         assert role_option and role_option.get('value') == 'admin'
         assert page.select_one('#username').get('value') == member['name']
 
-    @pytest.mark.usefixtures("clean_db")
```
smotornyuk, line 427: Just in case, copy of the comment from #7040:

This clean_db cannot be removed:

test creates a user, so it will fail if db is not initialized
user created with static name, so it will fail if you run this test multiple times

smotornyuk:
Just restore the `clean_db` fixture mentioned above and I'll merge it

PR #7039, opened 2022-08-24
Modified the default `sort_key` callable for `group_list_dictize` to …

JVickery-TBS:
…not be ASCIIbetical, matching the sorting coming from the db queries elsewhere.

Fixes #

Used `lower()` call in the `sort_key` default callable in `group_list_dictize`. Python sort/sorted is ASCIIbetical, whereas the db queries are purely alphabetical. This change makes sure that the `group_list_dictize` sorting matches the db query sorting.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7039.bugfix
+++ b/changes/7039.bugfix
@@ -1,0 +1,1 @@
+Changed default sort key for group and user lists from ASCII Alphebitized to new `strxfrm` helper, resulting in human-readable alphebitization.
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -532,7 +532,7 @@ def tag_dictize(tag: model.Tag, context: Context,
 def user_list_dictize(
         obj_list: Union[list[model.User], list[tuple[model.User, str]]],
         context: Context,
-        sort_key: Callable[[Any], Any] = lambda x:x['name'],
+        sort_key: Callable[[Any], Any] = lambda x: x['name'].lower(),
         reverse: bool=False) -> list[dict[str, Any]]:
 
     result_list = []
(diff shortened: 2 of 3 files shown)


wardi:
bonus points if we can use `locale.strcoll` for these

JVickery-TBS:
Added the `.lower()` call to the sorting for the user dictization as well.

I attempted to make use of `locale.strxfrm` to prepare the strings for locale comparison:
```
from locale import strxfrm
......
sort_key: Callable[..., Any]=lambda x: strxfrm(x['display_name'].lower())
```

Sadly this does not work as the flask app uses Babel's Locale object (https://github.com/python-babel/babel/blob/master/babel/core.py#L107) which does not implement the python core Locale, nor does it have `strxfrm` or `strcoll` methods.

Making use of `locale.setlocale` is not thread safe. We could make use of a new dependency that would help accomplish this. However, as postgres can already do locale sorting, it would probably make more sense to just do that instead of bringing in a new library when postgres can already do it.

I think instead of calling `sorted` in the dictizations, we can pass the flask app locale and sorting key to the db query?

JVickery-TBS:
@wardi sadly, it actually would not be possible to do the `COLLATE locale` in the dictization, as the dictization looks something like this:
```
group_list = [
            group_dictize(group, group_dictize_context,
                          **group_dictize_options)
            for group in cast("list[model.Group]", obj_list)]
```
So at this point, the sql query has already happened, and the list of objects is already passed into the dictization.

So perhaps we do need to just find a way to instantiate the python Locale into the flask app to store a locale object state?

wardi:
I think the best way to do this would be to add a helper called `strxfrm` that is "close enough" to properly sorted for many languages e.g.

```python
def strxfrm(s):
    '''
    Transform a string to one that can be used in locale-aware comparisons.
    Override this helper if you have different text sorting needs.
    '''
    return unicodedata.normalize('NFD', s).lower()
```

JVickery-TBS:
@wardi I added the helper and made use of it in the dictization methods and it seems to work quite well.

wardi:
@JVickery-TBS all this needs is a changelog entry now

JVickery-TBS:
@wardi I have added a changelog for this now.

PR #7040, opened 2022-08-25
change test usernames so they don't conflict

ThrawnCA:
Fixes #7036

### Proposed fixes:

Use distinct usernames in different group membership tests, so we don't get validation errors and don't need to reset the whole db before each test.

### DIFF
--- a/ckan/tests/controllers/test_group.py
+++ b/ckan/tests/controllers/test_group.py
@@ -367,13 +367,13 @@ class TestGroupMembership(object):
         user_obj = model.User.get(user["name"])
         # mock current_user
         current_user.return_value = user_obj
-        factories.User(fullname="My Fullname", name="my-user")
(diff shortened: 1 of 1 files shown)


ckan/tests/controllers/test_group.py
```diff
@@ -434,21 +434,20 @@ def test_membership_edit_page(self, current_user, app):
         assert role_option and role_option.get('value') == 'admin'
         assert page.select_one('#username').get('value') == member['name']
 
-    @pytest.mark.usefixtures("clean_db")
```
smotornyuk, line 437: This `clean_db` cannot be removed:
* test creates a user, so it will fail if db is not initialized
* user created with static name, so it will fail if you run this test multiple times

smotornyuk (changes requested):

PR #7044, opened 2022-08-25
Added `list-orphans` and `clear-orphans` sub-commands to the `search-…

JVickery-TBS:
…index` command.

Fixes #

Adds sub-commands to `search-index` to list and clear any orphaned package indexes.

### Background info:

We ran into an issue in which there were some left-over packages in the solr index, despite that the packages were purged from the database. So we had a report of a few packages not existing but still coming up in the searches due to the solr index. We had to just run `ckan search-index clear DATASET_ID` on each one. So this sparked adding these simple cli sub-commands to deal with this issue if it happens for others.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/search_index.py
+++ b/ckan/cli/search_index.py
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
         clear_all()
 
 
+def list_orphans(return_list: bool=False) -> list|None:
+    import ckan.logic as logic
+    search = logic.get_action('package_search')({},{
+                'q': '*:*',
+                'fl': 'id'})
+    indexed_package_ids = {r['id'] for r in search['results']}
+
+    import ckan.model as model
+    from sqlalchemy.sql import select
+    package_ids = {r[0] for r in select([model.package_table.c['id']]).execute()}
+
+    orphaned_package_ids = []
+    for indexed_package_id in indexed_package_ids:
+        if indexed_package_id not in package_ids:
+            orphaned_package_ids.append(indexed_package_id)
+
+    if return_list:
+        return orphaned_package_ids
+    from pprint import pprint
+    pprint(orphaned_package_ids)
+
+
+@search_index.command(name=u'list-orphans', short_help=u'Lists any non-existant packages in the search index')
+def list_orphans_command():
+    return list_orphans()
+
+
+@search_index.command(name=u'clear-orphans', short_help=u'Clear any non-existant packages in the search index')
+@click.option(u'-v', u'--verbose', is_flag=True)
+def clear_orphans(verbose: bool=False):
+    from ckan.lib.search import clear
+    for orphaned_package_id in list_orphans(return_list=True):
+        if verbose:
+            print("Clearing search index for dataset %s..." % orphaned_package_id)
+        clear(orphaned_package_id)
+
+
 @search_index.command(name=u'rebuild-fast',
                       short_help=u'Reindex with multiprocessing')
 def rebuild_fast():
--- a/ckan/cli/search_index.py
+++ b/ckan/cli/search_index.py
@@ -83,7 +83,7 @@ def get_orphans() -> list[str]:
     search = None
     indexed_package_ids = []
     while search is None or len(indexed_package_ids) < search['count']:
-        search = logic.get_action('package_search')({},{
+        search = logic.get_action('package_search')({}, {
                 'q': '*:*',
                 'fl': 'id',
                 'start': len(indexed_package_ids),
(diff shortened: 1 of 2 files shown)


ckan/cli/search_index.py
```diff
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
         clear_all()
 
 
+def list_orphans(return_list: bool=False) -> list|None:
```
smotornyuk, line 80: Prefer using `typing.Uinion`, because CKAN supports python >= v3.7

ckan/cli/search_index.py
```diff
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
...
+    import ckan.logic as logic
```
smotornyuk, line 81: Can you import it on the top-level of module?

ckan/cli/search_index.py
```diff
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
...
+    search = logic.get_action('package_search')({},{
+                'q': '*:*',
+                'fl': 'id'})
+    indexed_package_ids = {r['id'] for r in search['results']}
+
+    import ckan.model as model
```
smotornyuk, line 87: Do all the import in the one place

ckan/cli/search_index.py
```diff
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
...
+    from sqlalchemy.sql import select
+    package_ids = {r[0] for r in select([model.package_table.c['id']]).execute()}
```
smotornyuk, line 89: * `model.Session.query(model.Package.id)`, just to be consistent with the other SQL queries
* Do you really need to convert the DB cursor into the set? I don't think that you can get duplicated IDs here

ckan/cli/search_index.py
```diff
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
         clear_all()
 
 
+def list_orphans(return_list: bool=False) -> list|None:
+    import ckan.logic as logic
+    search = logic.get_action('package_search')({},{
```
smotornyuk, line 82: I believe that `package_search` returns 10-20 records by default. What if there  are more of them?

ckan/cli/search_index.py
```diff
@@ -77,6 +77,43 @@ def clear(dataset_name: str):
...
+
+    orphaned_package_ids = []
+    for indexed_package_id in indexed_package_ids:
+        if indexed_package_id not in package_ids:
+            orphaned_package_ids.append(indexed_package_id)
+
+    if return_list:
+        return orphaned_package_ids
+    from pprint import pprint
+    pprint(orphaned_package_ids)
```
smotornyuk, line 99: use `click.echo/secho`, like other CLI commands do

smotornyuk:
In addition to review comments:
* Run `flake8` and fix all style issues
* Run `npx pyright` and fix all type issues
* add `changes/7044.feature` with a short description of the feature for the changelog

ckan/cli/search_index.py
wardi: package_ids is a set here so that line 93 below will be O(1)

wardi: indexed_package_ids could be a list since we're only iterating over it though :-)

ckan/cli/search_index.py
JVickery-TBS: Changed function to only return a list.

ckan/cli/search_index.py
JVickery-TBS: Import is now at module top level.

ckan/cli/search_index.py
JVickery-TBS: There is now a loop for `package_search` at an iteration of 1000 which was described as the "upper limit" in `ckan/logic/action/get/package_search`. Though, this particular `package_search` is only getting the `id` fields so should be fine by the 1000.

ckan/cli/search_index.py
JVickery-TBS: Using `click.echo` now instead of `pprint`.

ckan/cli/search_index.py
JVickery-TBS: Imports at module top level now.

ckan/cli/search_index.py
JVickery-TBS: `package_ids` remaining as a set, but indexed_package_ids is now just staying as a list and iterating through that.

JVickery-TBS: `select([model.package_table.c['id']]).execute()` has been changed to `model.Session.query(model.Package.id)`

JVickery-TBS:
@smotornyuk I have done all of your suggested changes. I also ran `flake8` for the file but no issues came up locally for me. I ran `npx pyright` and fixed any issues with those as well.

The changelog file has also been added.

JVickery-TBS:
@smotornyuk nevermind regarding the `flake8`, turns out I just don't know how to use it. Figured it out and made those changes.

Because of the line length, I did end up putting the `@search_index.command` definitions across multiple lines. Hope that is alright?

Issue #6708, opened 2022-02-18
Flask-Babel 2.0.0 breaks ITranslation


amercader:
Upgrading Flask-Babel from 1.0.0 to [2.0.0](https://github.com/python-babel/flask-babel/releases/tag/v2.0.0) makes the `ITranslation` tests fail:

https://app.circleci.com/pipelines/github/ckan/ckan/1902/workflows/8ef66c69-6def-4d1b-8bbf-1678e93f56aa/jobs/9630

Basically our custom override to support multiple translation domains is never called:

https://github.com/ckan/ckan/blob/d47eb0972a6f99460926e2869bd858caee1f74bf/ckan/config/middleware/flask_app.py#L97-L106

@TkTech I *think* that the way this works in Flask-babel changed as part of this change: https://github.com/python-babel/flask-babel/pull/163 but I couldn't find any docs on how the new approach works.

Do you have any pointers on how to refactor this on our end?

TkTech:
@amercader I'll look into this, however it'll have to wait until Saturday.

TkTech:
I'm not entirely sure how this works as-is. It seems to depend on the order it's called (?) which is prone to breaking. Under v2.0.0, the approach I think will work here is to create a:
 ```domain = flask_babel.Domain(translation_directories=plugin.i18n_directory(), domain=plugin.i18n_domain())```
 for each plugin that defines `ITranslation`, then pass in `domain.gettext` to the templates instead of `flask_babel.gettext`.

amercader:
@TkTech I wasn't able to figure out your approach, so I proposed this on the Flask-Babel side:

PR #7064, opened 2022-09-08
Upgrade requirements September 2022

amercader:
* Flask-login 0.6.1 -> 0.6.2 causes multiple [auth related failures](https://app.circleci.com/pipelines/github/ckan/ckan/4182/workflows/8b780647-ee05-4808-8d2a-0187e4551db0/jobs/12073). The [changelog](https://github.com/maxcountryman/flask-login/blob/main/CHANGES.md#version-062) sounds pretty inocuous so it would be good to investigate if this is a fragile implementation on our end or a bug in Flask-login (cc @TomeCirun @pdelboca )
* **Update:** fixed in 785a0d308. There is one [package model failure](https://app.circleci.com/pipelines/github/ckan/ckan/4188/workflows/e9da5c74-cc28-455a-bb23-2a56d462f6d4/jobs/12081/tests#failed-test-0) introduced by SQLAlchemy==1.4.2 (current version in master is 1.4.0, last published one 1.4.41)
* We still can't update Flask-Babel to 2.0.0 as it breaks `ITranslation` (see https://github.com/ckan/ckan/issues/6708 and https://github.com/python-babel/flask-babel/pull/194 cc @TkTech )
* Updated manually lxml (req from feedgen), and mako (req from alembic, which has it unpinned)
* Dev requirements: docutils pinnned at 0.17 because sphinx-rtd-theme doesn't support a bigger version
* Dev requirements: pytest-factoryboy pinnned at 2.4.0 because on 2.5.0 we get:

```
      File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/tests/pytest_ckan/fixtures.py", line 98, in <module>
        register(factories.Sysadmin, "sysadmin")
      File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/pytest_factoryboy/fixture.py", line 124, in register
        assert model_name != factory_name, (
    AssertionError: Naming collision for <Sysadmin for <class 'ckan.model.user.User'>>:
       * factory fixture name: sysadmin
       * model fixture name: sysadmin
      Please provide different name for model fixture.
    assert 'sysadmin' != 'sysadmin'
```


pdelboca:
Great @amercader ! Thanks for this, looks great and it's good to have CKAN running on latest versions! 🚀

TomeCirun:
hey @amercader  @pdelboca  I've tested a bit the new `flask-login` version, at line [70](https://github.com/ckan/ckan/blob/51f8566a2478a7f6f6b366d8ca89a986b4903785/ckan/tests/controllers/test_admin.py#L70) in app.post we are sending the header with the token, but it seems that `before_request` in [admin.py](https://github.com/ckan/ckan/blob/51f8566a2478a7f6f6b366d8ca89a986b4903785/ckan/views/admin.py#L63) module executes first even before our [request_loader](https://github.com/ckan/ckan/blob/51f8566a2478a7f6f6b366d8ca89a986b4903785/ckan/config/middleware/flask_app.py#L308), so when it checks for access it would fail. 
I've tried to comment out `before_request`, just to see if `request_loader` is going to execute, and it would, we have the user, but it's too late.

I m not sure why we have this behavior, but I will keep digging, and see if I can fix something.

Thanks

PR #7077, opened 2022-09-13
[#7076] Fix duplicate calls to index code

### LINKED ISSUE #7076: Search indexing logic called twice after update or create dataset
**CKAN version**
2.10

**Describe the bug**

When updating or creating a dataset, we are indexing the dataset twice in a row, ie the [`index_package()`](https://github.com/ckan/ckan/blob/9f1b5cfaff8c135b589e2ea0275f1286c2e02711/ckan/lib/search/index.py#L108) function gets called twice during the same operation (and of course any `IPackageController.before_index()` hook gets called twice as well.

The root cause is the the obscure code run in the [`DomainObjectModificationExtension`](https://github.com/ckan/ckan/blob/9f1b5cfaff8c135b589e2ea0275f1286c2e02711/ckan/model/modification.py#L27), which considers the same Package object both new and changed at the same time, and fires two separate notification events.

amercader:
Fixes #7076

See issue for details. [`DomainObjectModificationExtension`](https://github.com/ckan/ckan/blob/9f1b5cfaff8c135b589e2ea0275f1286c2e02711/ckan/model/modification.py#L27) was sending duplicate notifications to hooks, including the one used by the indexing logic so the core indexing (and extensions using `before_dataset_index()`) was called twice.

### DIFF
--- a/ckan/model/modification.py
+++ b/ckan/model/modification.py
@@ -47,8 +47,12 @@ class DomainObjectModificationExtension(plugins.SingletonPlugin):
                 for item in plugins.PluginImplementations(plugins.IResourceUrlChange):
                     item.notify(obj)
 
-        changed_pkgs = set(obj for obj in changed
-                           if isinstance(obj, model.Package))
+
+        changed_pkgs = set()
+        new_ids = [obj.id for obj in new]
+        for obj in changed:
+            if isinstance(obj, model.Package) and obj.id not in new_ids:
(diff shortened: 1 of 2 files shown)


wardi:
IIRC some time ago we talked about removing the use of DomainObjectModification and instead index directly in the package_update/package_create actions

tino097 (approved):

PR #7078, opened 2022-09-13
Deprecate Recline views

amercader:
It has served CKAN very well but in 2.10 is time to start saying goodbye to the venerable Recline-based views. Recline hasn't been maintained in many years, and it doesn't even have its [own repository now](https://github.com/datopian/portal.js#what-happened-to-recline). 

The DataTables view is a much more solid replacement for tabular data, and for the functionality that will be lost (generate (crappy) graphs or maps from tabular data) there are more modern alternatives maintained elsewhere (eg from [Keitaro](https://github.com/keitaroinc/ckanext-visualize) or [Datopian](https://github.com/datopian/ckanext-dataexplorer-react)). The other thing that Recline offered that has no immediate replacement is the ability to preview tabular data *not* present in the DataStore via the DataProxy service but this has always proven very unreliable and the source of confusion for users, so again I think it's worth deprecating it.

This PR does a bit of cleanup and it mainly includes:
* Remove recline views from config defaults, tests and examples in docs
* Adds a warning deprecation notice when loading a recline plugin
* Adds `datatables_view` as a default view to be created (if loaded). I didn't add it as default to `ckan.plugins` as without the `datastore` one (and `xloader`) it won't work, but happy to discuss
* Refreshes the data viewer documentation, marking the recline views as deprecated and mentioning community alternatives like the ones mentioned before.

It doesn't remove the actual plugins, which we can do later on.

### DIFF
--- a/changes/7078.removal
+++ b/changes/7078.removal
@@ -1,0 +1,1 @@
+The Recline-based view plugins (`recline_view`, `recline_grid_view`, `recline_graph_view` and `recline_map_view`) are deprecated and will be removed in future versions. Check :ref:`data-viewer` for alternatives.
--- a/ckan/cli/views.py
+++ b/ckan/cli/views.py
@@ -40,10 +40,10 @@ def views():
 def create(ctx: click.Context, types: list[str], dataset: list[str],
            no_default_filters: bool, search: str, yes: bool):
     """Create views on relevant resources. You can optionally provide
-    specific view types (eg `recline_view`, `image_view`). If no types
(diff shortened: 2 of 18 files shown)


ThrawnCA:
The Datopian plugin is interesting, but has no tagged release versions. Are we meant to just run from the master branch? That's not acceptable for a production system.

The Keitaro plugin is also interesting, but doesn't seem to offer maps. So it isn't really a replacement for Recline.

PR #7081, opened 2022-09-13
[#7066] gzip .test_durations

### LINKED ISSUE #7066: .test_durations string matches
**CKAN version**
master

**Describe the bug**
Not a bug, but `.test_durations` is machine generated and contains lots of common strings, so `git grep` will often return its single giant line of text at the top of every result

**Steps to reproduce**
`git grep <almost anything>`

**Expected behavior**
find source code, not fill terminal with machine generated text with extremely long line lengths

**Additional details**
Could we, say, gzip this file in the repo or something to avoid matching common strings?

wardi:
Fixes #7066

### Proposed fixes:
store test durations gzipped in repo to avoid text matches in a large auto-generated json file

### Features:

- [ ] includes API changes


amercader:
Add the `.test_durations` generation step to the release process

PR #7082, opened 2022-09-14
Revert Flask requirement back to 2.0.3

pdelboca:
This PR fixes an mistake made when merging #7064 

### Proposed Fixes

Even when the PR with updated requirements works (#7064) the `requirements.in` and `requirements.txt` file are not sync. (executing `pip-compile` returns a different `requirements.txt` file).

This is causing `master` to work with `requirements.txt` but it will fail when compiling a new one. Looks like https://github.com/ckan/ckan/pull/7064/commits/b685d64501cce3dfba0eb9f5c6aa4eda1609a2e8 added the requirement, but never compiled.

I'm creating this PR to revert Flask requirements to `2.0.3` since running CKAN with `2.2.2` does not work.


amercader:
This is annoying, sorry about the `pip-compile` confusion.

Looks like we need to upgrade Flask-login to 0.6.2 to support Flask 2.2.x, but apart from the auth failures mentioned in #7064 looks like also there are also failures in Flask-multistatic, which doesn't look like it's maintained. We better revisit this after 2.10 is released. Can you create an issue with the details @pdelboca ?

Also no need to add a changelog file for changes within the same version

PR #7086, opened 2022-09-14
restore original plugin template directory order

### LINKED ISSUE #7085: plugin template order reversed (since #7033)
**CKAN version**
master

**Describe the bug**
After we swapped the order of calling plugins' update_config hooks the resulting order of templates and public directories is now reversed

**Steps to reproduce**
have more than one plugin that adds template directories

**Expected behavior**
first plugin templates override later plugin templates (historic behaviour)

**Additional details**

wardi:
Fixes #7085

### Proposed fixes:

restore original plugin template directory order after update_config order change #7033

also cleans up some code that used string concatenation for adding to a list of paths and clearly separates "extra_{template,public}_paths" settings from an ini file from template/public paths added by plugins at startup.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7085.bugfix
+++ b/changes/7085.bugfix
@@ -1,0 +1,1 @@
+restore original plugin template directory order after update_config order change
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -11,6 +11,8 @@ groups:
         internal: true
       - key: plugin_template_paths
         ignored: true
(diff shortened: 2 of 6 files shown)


wardi:
@amercader was just a silly mistake on my end. Was thrown off because I expected more than two test failures if I got this order wrong

amercader:
Looks good, thanks @wardi

PR #7091, opened 2022-09-21
Fix default values for REMEMBER_COOKIE_* options

### LINKED ISSUE #7065: Exception occurs when logging in, while setting a cookie
**CKAN version**: 2.9.5

**Describe the bug**

Exception occurs when logging in - seems like setting the "remember me" cookie fails.

> ValueError: Setting 'domain' for a cookie on a server running locally (ex: localhost) is not supported by complying browsers. You should have something like: '127.0.0.1 localhost dev.localhost' on your hosts file and then point your server to run on 'dev.localhost' and also set 'domain' for 'dev.localhost'

**Steps to reproduce**

1. Install from source: https://docs.ckan.org/en/2.9/maintaining/installing/install-from-source.html.
2. Add a sysadmin user, using the CLI command.
3. Start CKAN using built-in webserver (run CLI command).
4. Try to login.

**Expected behavior**

Be able to login.

**Additional details**

It seems like a string `"None"` is always passed as the domain to the [_make_cookie_domain](https://github.com/pallets/werkzeug/blob/b1911cd0a054f92fa83302cdb520d19449c0b87b/src/werkzeug/_internal.py#L446) function rather than the value `None` causing an exception.

I tried patching that function so it returns `None` if `domain == "None"` (a string) and then I'm able to login.

Changing the cookie_domain setting in ckan.ini has no effect at all, `"None"` (a string) is always passed to: [_make_cookie_domain](https://github.com/pallets/werkzeug/blob/b1911cd0a054f92fa83302cdb520d19449c0b87b/src/werkzeug/_internal.py#L446)

smotornyuk:
Modern browsers won't allow the addition of a cookie for single-word domains(like `localhost`, `com`, `xxx`). There must be at least one `.` in the domain name.
The declared default value of `REMEMBER_COOKIE_DOMAIN` is set to `None`, which is interpreted by YAML parser as a string. As result, flask-login tries to set a remember-cookie for the `None` domain, which is not allowed, as mentioned above.

### Proposed fixes:

Change declaration of `REMEMBER_COOKIE_DOMAIN` to `default: null`, which is parsed to python's `None` and work as expected.
In addition, change the declaration `REMEMBER_COOKIE_DURATION`. flask-login expects it to be an integer, not the `365 days` string.

@amercader, this change fixes the problem in the master. There is a similar issue for v2.9.5 - #7065 - but I just realized that it's probably unrelated because there is no `flask-login` in v2.9.5. And, there were no incorrect config declarations that actually caused the problem.

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -679,12 +679,13 @@ groups:
         description: The name of the cookie to store the “remember me” information in.
 
       - key: REMEMBER_COOKIE_DURATION
-        default: 365 days (1 non-leap Gregorian year)
+        default: 31_536_000
+        type: int
         description: |
           The amount of time before the cookie expires, as a datetime.timedelta object or integer seconds.
 
       - key: REMEMBER_COOKIE_DOMAIN
-        default: None
+        default: null
         description: |
           If the “Remember Me” cookie should cross domains, set the domain value here
           (i.e. .example.com would allow the cookie to be used on all subdomains of example.com).


amercader:
Good catch @smotornyuk , hopefully that gives pointers to fix #7065

PR #7094, opened 2022-09-22
Fix to display site description when logo defined

### LINKED ISSUE #6111: `ckan.site_description` isn't used
**CKAN version**
2.9

**Describe the bug**
The docs (https://docs.ckan.org/en/2.9/maintaining/configuration.html#ckan-site-description) say:

> This is for a description, or tag line for the site, as displayed in the header of the CKAN web interface.

But when I set this on a brand new install of CKAN, the only place it's updated is in the "Site Tag Line" field in the config page (/ckan-admin/config) - I also note that this field doesn't have any help info in the left hand panel

**Expected behavior**
Either this option should be implemented somewhere, or the docs should explain that maintainers need to implement it themselves

Deep-NEC:
Fixes #6111 

### Proposed fixes:
This fix will allow to use ckan site description even when site logo is defined

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/header.html
+++ b/ckan/templates/header.html
@@ -75,6 +75,11 @@
         <img src="{{ h.url_for_static_or_external(g.site_logo) }}" alt="{{ g.site_title }}"
           title="{{ g.site_title }}" />
       </a>
+        {% if g.site_description %}
+       <div>
+      {{ g.site_description }}
+       </div>
+       {% endif %}
       {% else %}
       <h1>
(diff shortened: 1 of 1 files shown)


Deep-NEC:
Hi @kowh-ai I have investigated ``ci/circleci: test-python-3`` failure, it is failed due to ``def test_tag_line``. In this function it is defined to test the description/tag-line only when no logo and in this PR the fix is to display description/tag-line even when logo is defined. Hence we need to modify the testcase ``def test_tag_line`` according to this PR. Please confirm my understanding.

Thanks

Issue #7028, opened 2022-08-23
Lock down cookie based authentication on API endpoints


amercader:
Discussed on 2022-08-23 dev meeting:

1. Default behaviour: discard cookie based authentication in the API if it doesn't include the CSRF token
    - Is this equivalent to remove the `@csrf.exempt` decorator [here](https://github.com/ckan/ckan/blob/34401f000846aa59559c77e755e56656cf1355f9/ckan/views/api.py#L212)?
2. Add new configuration option: turn off entirely cookie based authentication on the API endpoint
    - Not sure how to implement this, but probably we need to fiddle at the Flask-login level. Maybe [this](https://flask-login.readthedocs.io/en/latest/#disabling-session-cookie-for-apis) is helpful?
3. (Optional): Allow cookie based authentication without valid CSRF tokens on development mode

@wardi @EricSoroos did I capture that correctly?

amercader:
@TomeCirun you were kindly volunteered to work on this by @tino097 :)

Let's start with the first point. I think what we need to do is:

1. Remove the `@csrf.exempt` check from the API `action` endpoint. This will cause all client side JS API calls to fail, so
2. We need to update the JS calls to include the CSRF header, like you did for the `follow` module on #6863 . Remember to move the `addCsrfTokenHeader()` function to the [base module class](https://github.com/ckan/ckan/blob/master/ckan/public/base/javascript/module.js). Or maybe we even don't have to change any modules and just modify the [client class](https://github.com/ckan/ckan/blob/master/ckan/public/base/javascript/client.js) they use to add the header
3. Document the behaviour change (No cookie based authentication in the API without CSRF token) and the changes needed in JS if they are not using the ckan client

Does this make sense?

PR #7096, opened 2022-09-22
[#7028] CSRF and API calls, tests

### LINKED ISSUE #7095: POST requests to API don't work with CSRF, tests bypass CSRF protection
**CKAN version**
2.10

**Describe the bug**
The API is completely unusable since https://github.com/ckan/ckan/pull/7058 was merged.

Any POST API call (and presumably PUT, DELETE etc) returns a `400 Bad Request: The CSRF token is missing` error, which in hindsight makes sense because we removed the `@csrf.exempt` decorator from the `api.action` endpoint. We'll need to see how we can implement:

* API call + Cookie authentication = CSRF enabled (to support current frontend API usage)
* API call + Header authentication = CSRF disabled (to support external API calls)

But more worryingly,I found out that the tests are happily ignoring all these issues,  by bypassing CSRF protection entirely.

https://github.com/ckan/ckan/blob/367fed2fa572e896c5eb703db9c663e097133e41/ckan/tests/helpers.py#L271-L272

This might made sense while we were implementing the feature, but I think we need to discuss what we want going forward.

amercader:
Fixes #7095 

* Disable CSRF protection if the user was authenticated via `Authorization` header
* Re-enable CSRF protection in tests, still WIP

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -12,7 +12,7 @@ import logging
 from logging.handlers import SMTPHandler
 from typing import Any, Iterable, Optional, Union, cast
 
-from flask import Blueprint, send_from_directory
+from flask import Blueprint, send_from_directory, current_app
 from flask.ctx import _AppCtxGlobals
 from flask.sessions import SessionInterface
 from flask_multistatic import MultiStaticFlask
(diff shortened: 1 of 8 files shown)


ckan/config/middleware/flask_app.py
```diff
@@ -327,7 +327,10 @@ def load_user_from_request(request):  # type: ignore
         This callback function is called whenever a user could not be
         authenticated via the session cookie, so we fall back to the API token.
         """
+        g.login_via_request = True
```
wardi, line 330: A better name might reduce the need for comments below
```suggestion
        g.login_via_auth_header = True
```

amercader:
@wardi all checks except the changelog (which doesn't apply) are green, so this is good to review, thanks!

PR #7106, opened 2022-09-27
[7095]POST requests to API don't work with CSRF, tests bypass CSRF protection

### LINKED ISSUE #7095: POST requests to API don't work with CSRF, tests bypass CSRF protection

TomeCirun:
Fixes #7095

Add `CKAN` wrapper for `CSRFProtect`, re-enable the `CSRF` protection for the tests.

### Features:

- [x] includes API changes

### DIFF
--- a/changes/6992.changes
+++ b/changes/6992.changes
@@ -1,0 +1,1 @@
+Add CKAN wrapper for CSRFProtect, re-enable the CSRF protection for the tests.
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -34,6 +34,7 @@ from flask_wtf.csrf import CSRFProtect
 from ckan.common import CKANConfig, asbool, session, current_user
 
 import ckan.model as model
+import ckan.plugins.toolkit as toolkit
 from ckan.lib import base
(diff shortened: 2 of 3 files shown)


TomeCirun:
@amercader let me know if you agree to these changes and if something should be removed/added.

Thanks

PR #7108, opened 2022-09-28
Fix CKAN registering only plugin's first blueprint

pdelboca:
### Proposed fixes:
This PR fixes an error in the logic that makes CKAN to register only the first Blueprint defined in an extension.

It also refactors some of the CSRF logic to internal method for better readability and maintenance.


pdelboca:
@smotornyuk what do you think is the best way to test that all blueprints returned by `get_blueprint()` gets registered?

I'm thinking on adding a test to `test_middleware.py` by refactoring a little bit the `MockRoutingPlugin` class to register two blueprints. What are your thoughts on how to test this scenarios?

smotornyuk:
How about testing inside the existing [`iblueprint` plugin](https://github.com/ckan/ckan/tree/master/ckanext/example_flask_iblueprint)? If you update this example, people who are checking these examples will also know about the ability to register multiple blueprints at once.

As for testing, stubbing [register_extension_blueprint](https://github.com/ckan/ckan/blob/2d34d873ca0fbaa0448e735cfb9009473e345329/ckan/config/middleware/flask_app.py#L526) or Flask's `Blueprint.register_blueprint` would be the most reliable option, but honestly, I don't like such low-level tests. What do you think about registering a blueprint with the route `/x` and another one with the route `/y` and just making a couple of requests to these endpoints? In this way, no matter how the code changes(even if we migrate from flask to something else in the future), this test will remain the same.

pdelboca:
@smotornyuk I have implemented the tests and updated the ExampleIBlueprintPlugin.

let me know your thoughs!

PR #7110, opened 2022-09-28
Enable DateTime to be returned through Actions

EricSoroos:
Version: 2.9, master

Fixes #

### Proposed fixes:

Some actions, at the very least, `datapusher_status` return a datetime object as part of a nested dictionary of data. The native JSON formatter won't process this, and triggers a 500 error. This prevents getting `datapusher_status` through the API. 

This patch provides an object formatter that converts the Datetime to an ISO-formatted string, thus allowing `datapusher_status` to be accessed remotely. 

### Features:

- [ ] includes API changes


wardi:
@EricSoroos would you add a changes file for this bugfix?

Note to self: a similar change will be required in ckanapi when `-c` is used on the cli

PR #7112, opened 2022-09-29
Don't lose the user!

avdata99:
Fixes problems in extensions that requires always a user 
The `group_follower_count` and other actions now have an auth function and this is a problem.
With this change we preserve context['user'] for places is which this is required.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -13,7 +13,7 @@ import logging
 from collections.abc import MutableMapping
 
 from typing import (
-    Any, Iterable, Optional, TYPE_CHECKING,
+    Any, Iterable, List, Optional, TYPE_CHECKING,
     TypeVar, cast, overload, Container, Union)
 from typing_extensions import Literal
 
@@ -29,7 +29,7 @@ from flask_babel import (gettext as flask_ugettext,
 import simplejson as json  # type: ignore # noqa: re-export
 import ckan.lib.maintain as maintain
 from ckan.config.declaration import Declaration, Key
-from ckan.types import Model
+from ckan.types import Model, Context
 
 
 if TYPE_CHECKING:
@@ -265,6 +265,28 @@ def aslist(obj: Any, sep: Optional[str] = None, strip: bool = True) -> Any:
         return []
     else:
         return [obj]
+
+
+def clean_context(
+    context: Context,
+    fields: Optional[List[str]] = None
+) -> Context:
+    """ Copy just the minimum fields into a new context
+        for cases in which we reuse the context and
+        we want a clean version with minimum fields """
+    new_context = {}
+    if fields is None:
+        fields = [
+            'model', 'session', 'user', 'auth_user_obj',
+            'ignore_auth'
+        ]
+
+    for field in fields:
+        if field in context:
+            new_context[field] = context[field]
+
+    new_context = cast(Context, new_context)
+    return new_context
 
 
 local = Local()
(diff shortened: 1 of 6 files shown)


ckan/logic/action/get.py
```diff
@@ -1254,8 +1254,9 @@ def _group_or_org_show(
         schema = group_plugin.db_to_form_schema()
 
     if include_followers:
+        context['session'] = model.Session
         group_dict['num_followers'] = logic.get_action('group_follower_count')(
-            {'model': model, 'session': model.Session},
+            context,
```
wardi, line 1259: If we pass None (or an empty dict) in for the context a new context will be generated with the default user determined from the request. This might be better than passing along `context` because often context contains internal parameters that only make sense for the original action it was passed to.
```suggestion
            None,
```

avdata99: This change don't work for my project.
I think this PR is clearer than just sending `None`, but I'm not sure if any specific item in "context" could be a problem.

wardi: @avdata99 which fields do you actually need from context on your project? Was thinking of writing a `clean_context()` function that could copy just the minimum fields into a new context dict for cases like this (we have them everywhere in the code, and no consistent way of handling them)

wardi: @avdata99 which fields do you actually need from context on your project? Was thinking of writing a `clean_context()` function that could copy just the minimum fields into a new context dict for cases like this (we have them everywhere in the code, and no consistent way of handling them)

avdata99: @wardi I updated this PR to include a `clean_context` function
Do you think is what we need?

ckan/common.py
```diff
@@ -267,6 +267,28 @@ def aslist(obj: Any, sep: Optional[str] = None, strip: bool = True) -> Any:
...
+            'model', 'session', 'user', 'auth_user_obj',
+            'ignore_auth'
+        ]
+
+    for field in fields:
+        if field in context:
+            new_context[field] = context[field]
+
+    new_context = cast(Context, new_context)
+    return new_context
```
wardi, line 289: I like it, but let's keep it simple to make it easier for us to change in the future. If callers want to pass along extra variables they can do so after calling `clean_context`.

```suggestion
def clean_context(
    context: Context,
) -> Context:
    """ Copy just the minimum fields into a new context
        for cases in which we reuse the context and
        we want a clean version with minimum fields """
    new_context = {
        k:context[k] for k in (
            'model', 'session', 'user', 'auth_user_obj', 'ignore_auth'
        ) if k in context
    }
    new_context = cast(Context, new_context)
    return new_context
```

ckan/common.py
```diff
@@ -267,6 +267,21 @@ def aslist(obj: Any, sep: Optional[str] = None, strip: bool = True) -> Any:
         return [obj]
 
 
+def clean_context(
```
amercader, line 270: Whatever this function ends up being called, it should be defined in `ckan/logic/__init__.py` with the rest of the actions/context logic

amercader: And please add it to the toolkit as well, as in [here](https://github.com/ckan/ckan/blob/dc49cece33e447858d0008bc71b9a4ab724602ee/ckan/plugins/toolkit.py#L20)

ckan/logic/action/get.py
amercader: I think this is a great idea, but perhaps `clean_context()` implies emptiness (like our `clean_db`, `clean_index` test fixtures) and we actually mean "clean context but keeping the user and model". Maybe something like `fresh_context()`? Maybe I'm reading too much into it :)

wardi: /polls clean_context fresh_context trim_context new_context minimal_context

PR #7122, opened 2022-10-04
package_update performance purge resources approach

### LINKED ISSUE #7119: package_update performance of repeated calls

wardi:
Fixes #7119 

### Proposed fixes:

This solves the issue by purging resources that are removed in a package_update. This seems like a better solution because I can't think of any value in keeping these deleted resources in the db

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7119.bugfix
+++ b/changes/7119.bugfix
@@ -1,0 +1,1 @@
+remove old deleted resources on package_update so that performance is consistent over time (no longer degrading)
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -69,9 +69,14 @@ def package_resource_list_save(
     session = context['session']
     model = context['model']
     resource_list = package.resources_all
+    # existing resources not marked as deleted - when removed these
+    # need to be kept in the db marked as deleted so that extensions like
+    # datastore have a chance to remove tables created for those resources
     old_list = session.query(model.Resource) \
         .filter(model.Resource.package_id == package.id) \
         .filter(model.Resource.state != 'deleted')[:]
+    # resources previously deleted can be removed permanently as part
+    # of this update
     deleted_list = session.query(model.Resource) \
         .filter(model.Resource.package_id == package.id) \
         .filter(model.Resource.state == 'deleted')[:]
--- a/ckan/lib/dictization/model_save.py
+++ b/ckan/lib/dictization/model_save.py
@@ -72,6 +72,9 @@ def package_resource_list_save(
     old_list = session.query(model.Resource) \
         .filter(model.Resource.package_id == package.id) \
         .filter(model.Resource.state != 'deleted')[:]
+    deleted_list = session.query(model.Resource) \
+        .filter(model.Resource.package_id == package.id) \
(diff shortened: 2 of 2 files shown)


smotornyuk:
@wardi, have you forgotten about this PR? It only needs these 3 errors fixed before merge

wardi:
haven't forgotten, still need to find a good fix for these issues

wardi:
@smotornyuk I combined the two different approaches so now resources are still "soft deleted" then the next time they aren't sent in a package_update they are actually removed. This should maintain compatibility with things that expect them to be saved with state=deleted like the datastore code and still prevents the performance from getting worse over time.

wardi:
@amercader how does this look for an explanation of the resource saving changes [27a64e1](https://github.com/ckan/ckan/pull/7122/commits/27a64e1537af9d29e16d47081bbc616f27c550d3) ?

PR #7123, opened 2022-10-06
Remove dataset count from group view defaults

Zharktas:
While fixing ckanext-hierarchy, I was puzzled why there is multiple package_search calls while calling group views. 

`group_dictize` calls package_search if `packages_field` is not none, `group_show` only sets it to None if no packages or count is requested:

https://github.com/ckan/ckan/blob/07669305decebaec827142cb26ecbd739473d2ab/ckan/logic/action/get.py#L1197-L1202

### Proposed fixes:

Removes dataset count from group view as it is later fetched with package_search itself.

### Features:

- [ ] includes API changes


tino097 (approved):
LGTM

Issue #6840, opened 2022-05-03
Theme issues on 2.10 branch (master)


amercader:
**CKAN version**
master

- [x] Pagination buttons are not styled

![Screenshot 2022-05-03 at 12-39-24 Dataset - CKAN](https://user-images.githubusercontent.com/200230/166456932-f9cff6e6-3bcb-451c-b522-ac02331625b7.png)

- [x] Activity filter on Dashboard is floating on the left

![Screenshot 2022-05-03 at 12-22-15 Manage - Adrià 333332 - Users - CKAN](https://user-images.githubusercontent.com/200230/166456997-c3950760-717f-4216-b121-1d018353a762.png)

- [x]  Activity icons not showing up (not sure if this is a font issue in my machine)

![Screenshot 2022-05-03 at 14-59-04 Test datasets 222222 - CKAN](https://user-images.githubusercontent.com/200230/166457385-02751527-ea40-48d1-abe2-7e2e1905eb01.png)

- [ ] DataTables sort buttons
![Screenshot 2022-09-13 at 13-05-44 REREDS - 25_best_tacos_austin csv - CKAN](https://user-images.githubusercontent.com/200230/191740513-53bf298c-0330-41af-8ab5-69a3601f0f4d.png)

- [x] Clear upload button
![Screenshot 2022-09-13 at 13-02-50 Edit - 25_best_tacos_austin csv - REREDS - CKAN](https://user-images.githubusercontent.com/200230/191740534-dc13eae7-0576-4985-bab2-d59eda5f1bd0.png)

- [x] Spacing in custom extras fields
![Screenshot 2022-09-13 at 11-38-30 Create Dataset - CKAN](https://user-images.githubusercontent.com/200230/191740568-4baf8418-a17e-4ad4-95a7-ee22b016b36e.png)

- [x] License field margins and spacing
![Screenshot 2022-09-13 at 11-38-22 Create Dataset - CKAN](https://user-images.githubusercontent.com/200230/191740594-43ccbe93-ba18-4ade-bc17-d8dd4bf88b9c.png)

PR #7126, opened 2022-10-07
6840 theme issue fixes

robin-NEC:
Fixes part of #6840 

### Proposed fixes:

- This is a partial fix for #6840 containing  `custom` and `license` fields fixes.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/6840.misc
+++ b/changes/6840.misc
@@ -1,0 +1,1 @@
+Added license and custom fields theme fixes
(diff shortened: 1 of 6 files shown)


pdelboca:
Hello @robin-NEC !

I don't feel comfortable overriding Bootstrap's core `row` and `col-md6` classes to fix an issue in one element since it can backfire. (This are core classes for displaying and aligning elements).

Can you simplify the solution to only use one new class? Also, you are introducing a new CSS class called `custom-component` what's the difference with the already `control-custom` class that we are assigning to those components?

Zharktas:
Lets close this as nothing has happened in couple of months, new PR can be opened to fix the original issue.

PR #7139, opened 2022-10-18
Use API Tokens in Datapusher

amercader:
The DataPusher extension relied on an API key being sent to the DataPusher service (the site user one). From CKAN 2.10 onward this no longer works, so we need to use API tokens instead. This follows the pattern already used in [ckanext-xloader](https://github.com/ckan/ckanext-xloader/pull/164), where a  new config option is added so users can explicitly define a token in their configuration.

In the future we might want to automate this partially using something like service accounts, but this seems like an easy first step

### DIFF
--- a/changes/7139.migration
+++ b/changes/7139.migration
@@ -1,0 +1,1 @@
+Users of the Xloader or DataPusher need to provide a valid API Token in their configurations using the ``ckanext.xloader.api_token`` or ``ckan.datapusher.api_token`` keys respectively.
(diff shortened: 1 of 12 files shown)


amercader:
@pdelboca Config options are built from the [config declaration](https://github.com/ckan/ckan/pull/7139/files#diff-2dd950c53970a6073d2d51ccb03d9326efa4968d4b9dcb42dcb60febdf38eea1) so that should come up once merged, and I have also documented it on the [DataPusher repo](https://github.com/ckan/datapusher/commit/5dbd4e7cc3f17f4efc4e5fd3b7bf84a53e0b8cd0).

PR #7133, opened 2022-10-11
Update Beaker Parameters in config_declaration.yaml

kowh-ai:
Fixes https://github.com/ckan/ckan/issues/7128 https://github.com/ckan/ckan/issues/7092 https://github.com/ckan/ckan/issues/7120

### Proposed fixes:
The `config_declaration.yaml` file contains 3 Beaker Session settings that need to be initialised in a newly generated CKAN config file (ckan.ini _or_ production.ini for a Docker install)

These are: 

`beaker.session.cookie_expires`
`beaker.session.cookie_domain`
`beaker.session.timeout`

It seems when the PR https://github.com/ckan/ckan/pull/7068 was merged it caused this error:

… `File "/usr/lib/ckan/default/lib/python3.8/site-packages/beaker/util.py", line 281, in verify_options raise Exception(error) 
Exception: Cookie expires was not a boolean, datetime, int, or timedelta instance.` ...

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

Issue #7128, opened 2022-10-10
Beaker Session parameters need to be initialised in the CKAN config file


kowh-ai:
**Beaker Error in CKAN  when starting**

It seems when the PR https://github.com/ckan/ckan/pull/7068 was merged it caused an error with CKAN starting up

The [config_declaration.yaml](https://github.com/ckan/ckan/blob/master/ckan/config/config_declaration.yaml) file contains 3 Beaker Session settings that need to be initialised in a newly generated CKAN config file (ckan.ini _or_ production.ini)

`beaker.session.cookie_expires`
`beaker.session.cookie_domain`
`beaker.session.timeout`

Current generated **ckan.ini** file
![Screenshot 2022-10-10 at 4 01 25 pm](https://user-images.githubusercontent.com/54408245/194884026-eb5cb9c3-1c7c-4fac-9edd-ce85169a840d.png)

When CKAN tries to start it error's as follows:
 `…
File "/usr/lib/ckan/default/lib/python3.8/site-packages/beaker/util.py", line 281, in verify_options
    raise Exception(error)
Exception: Cookie expires was not a boolean, datetime, int, or timedelta instance.
...`

This particular error is from the `beaker.session.cookie_expires` parameter not being initialised

**Steps to reproduce**
Install CKAN (current **master** branch) from source _or_ Docker Compose will result in an error 

**Expected behavior**
CKAN should start!

**Additional Details**
Issues: https://github.com/ckan/ckan/issues/7092 https://github.com/ckan/ckan/issues/7120

Issue #7120, opened 2022-10-04
CKAN w/ docker-compose won't start


lbam:
**CKAN version**

d199d39d5e225ecbc4ee9c75418009226c7c8571

**Describe the bug**
CKAN fails to start when following the instructions from [Installing CKAN with Docker Compose](https://docs.ckan.org/en/2.9/maintaining/installing/install-from-docker-compose.html).

**Steps to reproduce**

From the installation instructions + the additional instructions listed in `contrib/docker/.env.template`:

```
cd contrib/docker
cp .env.template .env
docker-compose down
docker rmi -f docker_ckan docker_db
docker rmi $(docker images -f dangling=true -q)
docker-compose build
docker-compose up -d
docker-compose restart ckan
```

All containers start, but the `ckan` container runs into an exception and gets stuck in a restart loop.

**Expected behavior**
CKAN runs.

**Additional details**
This is what docker-compose spews out at regular intervals:

```
ckan          | Traceback (most recent call last):
ckan          |   File "/usr/lib/ckan/venv/bin/ckan", line 8, in <module>
ckan          |     sys.exit(ckan())
ckan          |   File "/usr/lib/ckan/venv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
ckan          |     return self.main(*args, **kwargs)
ckan          |   File "/usr/lib/ckan/venv/lib/python3.8/site-packages/click/core.py", line 1054, in main
ckan          |     with self.make_context(prog_name, args, **extra) as ctx:
ckan          |   File "/usr/lib/ckan/venv/lib/python3.8/site-packages/click/core.py", line 920, in make_context
ckan          |     self.parse_args(ctx, args)
ckan          |   File "/usr/lib/ckan/venv/src/ckan/ckan/cli/cli.py", line 116, in parse_args
ckan          |     result = super().parse_args(ctx, args)
ckan          |   File "/usr/lib/ckan/venv/lib/python3.8/site-packages/click/core.py", line 1613, in parse_args
ckan          |     rest = super().parse_args(ctx, args)
ckan          |   File "/usr/lib/ckan/venv/lib/python3.8/site-packages/click/core.py", line 1378, in parse_args
ckan          |     value, args = param.handle_parse_result(ctx, opts, args)
ckan          |   File "/usr/lib/ckan/venv/lib/python3.8/site-packages/click/core.py", line 2360, in handle_parse_result
ckan          |     value = self.process_value(ctx, value)
[...]

Issue #7092, opened 2022-09-21
Beaker - Cookie expires installation error on Mac


MoSHad91:
**CKAN version**
Latest version

**Describe the bug**
I tried to install the CKAN according to the [documentation with docker compose](https://docs.ckan.org/en/latest/maintaining/installing/install-from-docker-compose.html) on Mac and got the following error. 
Mac M1
Docker version 4.12.0

> ckan          | db:5432 - accepting connections
ckan          | Option ckan.requests.timeout is not declared
ckan          | Option ckan.requests.timeout is not declared
ckan          | 2022-09-21 14:56:44,289 INFO  [ckan.cli] Using configuration file /etc/ckan/production.ini
ckan          | 2022-09-21 14:56:44,289 INFO  [ckan.config.environment] Loading static files from public
ckan          | 2022-09-21 14:56:44,374 WARNI [ckan.common] Option ckan.plugins is not declared
solr          | 2022-09-21 14:56:44.461 INFO  (qtp736709391-21) [   x:ckan] o.a.s.c.S.Request [ckan]  webapp=/solr path=/select params={q=*:*&rows=1&wt=json} hits=0 status=0 QTime=2
solr          | 2022-09-21 14:56:44.473 INFO  (qtp736709391-12) [   x:ckan] o.a.s.c.S.Request [ckan]  webapp=/solr path=/schema params={wt=schema.xml} status=0 QTime=2
ckan          | 2022-09-21 14:56:44,479 INFO  [ckan.config.environment] Loading templates from /usr/lib/ckan/venv/src/ckan/ckan/templates
db            | 2022-09-21 14:56:44.645 UTC [211] ERROR:  relation "user" does not exist at character 555
db            | 2022-09-21 14:56:44.645 UTC [211] STATEMENT:  SELECT "user".password AS user_password, "user".id AS user_id, "user".name AS user_name, "user".fullname AS user_fullname, "user".email AS user_email, "user".apikey AS user_apikey, "user".created AS user_created, "user".reset_key AS user_reset_key, "user".about AS user_about, "user".last_active AS user_last_active, "user".activity_streams_email_notifications AS user_activity_streams_email_notifications, "user".sysadmin AS user_sysadmin, "user".state AS user_state, "user".image_url AS user_image_url, "user".plugin_extras AS user_plugin_extras 
db            | 	FROM "user" 
db            | 	WHERE "user".name = 'default' OR "user".id = 'default' 
db            | 	 LIMIT 1
[...]

PR #7141, opened 2022-10-18
Declaration flag for options commented by default

smotornyuk:
Additional fix for the series of cookie-issues: #7128 #7092 #7120

@kowh-ai has already fixed the original issue in #7133. But I just realized, that now we have a default value for the cookie domain, that will be added to every new config file. This, in turn, will force every user to update this value, because the default value will be different from the real app's domain(likely). If someone won't update the cookie domain, cookies will be set for the **wrong** domain and ignored by the browser. I.e, authentication will not work at all:)

It won't take much time to update the documentation, mentioning this additional mandatory setup step. But, personally, I don't want to introduce new steps. 

We can add `default: null` to the cookie_domain declaration(which I did before, but now I see that it doesn't solve the problem completely). Beaker/flask_login uses `None` as a special value, which falls back to the current application's domain. It nice feature, but it will work only in `strict` config mode or via `config.get_value`, which validates values and resolve defaults(explanation in the following paragraph).

Empty(i.e, `default: null`) options are added to the config file as 
```ini
xxx = 
```
And, while `config.get_value` manages to get the default `None` in such case, `config.get` returns the empty string for such an option. It wouldn't be a problem if we're using config object only inside CKAN code, because we can easily convert all the `config.get` to `config.get_value`. But the thing is, `flask_login` also puts hands on our config object. And it uses the problematic `config.get`.

Using `mode = strict` would solve the problem. It validates/converts all the config options during the app's initialization so that all the further `config.get` work with correct values. But config-modes is a completely new feature and we don't want to make them mandatory just yet.

The only other option I see is to add a cookie domain as a commented option to the config file generated via `ckan config generate`. In this way, it will be set to `None` in normal mode(because it's basically a missing option) and in strict mode(because of validators/defaults/etc.)

This PR adds a new flag `commented: true` to the config option declaration. With this flag option added to the config template as a commented line.

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -294,7 +294,8 @@ groups:
           Defaults to never expiring.
 
       - key: beaker.session.cookie_domain
-        default: .example.com
+        commented: true
+        placeholder: .example.com
         description: |
           What domain the cookie should be set to. When using sub-domains,
           this should be set to the main domain the cookie should be valid for. For example,
@@ -696,7 +697,8 @@ groups:
           The amount of time before the cookie expires, as a datetime.timedelta object or integer seconds.
 
       - key: REMEMBER_COOKIE_DOMAIN
-        default: null
+        placeholder: .example.com
+        commented: true
         description: |
           If the “Remember Me” cookie should cross domains, set the domain value here
           (i.e. .example.com would allow the cookie to be used on all subdomains of example.com).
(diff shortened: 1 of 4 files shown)


kowh-ai (approved):
Tested this change on a new Ubuntu 20.04 environment and a local docker environment. Both work fine. Here is the session section from a newly generated ckan.ini file:
Screenshot 2022-10-19 at 10 57 55 am

amercader:
This makes sense, thanks @smotornyuk and @kowh-ai

PR #7146, opened 2022-10-19
[7104]Fix User ID does not have to be unique?

### LINKED ISSUE #7104: User ID does not have to be unique?
**CKAN version**
CKAN 2.9.5

**Describe the bug**
The `id` field can be set by the user when calling the user_create API action. It appears that no check for ID uniqueness is made when doing so. This surprised me!  I was able to send multiple requests with the same `id` field, so long as they had different `name` fields, without any error being thrown.

**Steps to reproduce**
Perhaps the simplest way to reproduce is to run the following test:
```
import pytest
import ckan.tests.factories as factories

def test_user_id_uniqueness(self):
    factories.User(id='test-id')
    with pytest.raises(Exception):
        factories.User(id='test-id')
```

**Expected behavior**
I'd have expected the test above to pass, i.e. an exception to be raised (presumably a ValidationError) when creating a second user with the same ID. But no error is thrown, so the test fails. Furthermore I have confirmed the same behaviour appears when making similar requests as a sysadmin using the user_create action. 

Since the id is actually the primary key of the user table in the model, I'm assuming something odd is happening under the hood. 

**Is this a bug or intended behaviour?**

TomeCirun:
Fixes #7104 

I can't think of a better solution at the moment, in this case, we don't have to be concerned about duplicate users with the same id because `primary_key=True implies null=False and unique=True.`

So I added a flag when `user_create` is triggered, which will go down to the `table_dict_save` and check if the user has an `id` and `came_from` is `'user_create'`... if so, then we want to show the `ValidationError`.

Any suggestions are welcome.

### Features:

- [x] includes API changes

### DIFF
--- a/ckan/lib/dictization/__init__.py
+++ b/ckan/lib/dictization/__init__.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, Iterable
 import sqlalchemy
 from sqlalchemy import Table
 # type_ignore_reason: incomplete SQLAlchemy types
-from sqlalchemy.engine import Row # type: ignore
+from sqlalchemy.engine import Row
 from sqlalchemy.orm import class_mapper
 
 from ckan.model.core import State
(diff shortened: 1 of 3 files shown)


TomeCirun:
I also thought about adding a validator to the `default_user_schema`, but that will affect the `default_update_user_schema` and all schemas that inherit from the default one

@ckan/core

TomeCirun:
another solution might be to remove the ID from the payload in the user_create, why do we need to send a custom id for the user?

PR #7148, opened 2022-10-20
Automatically rename deprecated IPackageController methods(fix)

smotornyuk:
Last year we've deprecated `before_*` and `after_*` methods of `IPackageController` and `IResourceController` in favor of `before_dataset_*`/`before_resource_*` forms. Methods with old names are copied as the methods with new names during interface initialization.

But I noticed that this auto-copying doesn't work on master. So I had to fix it a bit. 
Instead of copying methods during corresponding *interface* initialization (which happens somewhere inside `pyutilib`), these methods are copied during the *plugin* initialization(`SingletonePlugin.__init__`) now.
And I modified a couple of tests so that they'll report if it stops working in the future.

### DIFF
--- a/ckan/plugins/core.py
+++ b/ckan/plugins/core.py
@@ -6,6 +6,7 @@ Provides plugin services to the CKAN
 from __future__ import annotations
 
 import logging
+import warnings
 from contextlib import contextmanager
 from typing import (Any, Generic, Iterator, Optional,
                     Type, TypeVar, Union)
@@ -21,6 +22,7 @@ import ckan.plugins.interfaces as interfaces
 
 from ckan.common import config
 from ckan.types import SignalMapping
+from ckan.exceptions import CkanDeprecationWarning
 
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


pdelboca:
@smotornyuk I have created a simple extension implementing this interfaces:

```python
    # IPackageController
    def after_search(self, search_results, search_params):
        print("### CALLING after_search!! ###")
        return search_results

```

when trying to test this I found two issues:

1) I cannot see any warning neither at init time, neither when using it:

```
(.venv)  ~/Repos/ckan (fix-support-of-deprecated-ipackagecontroller-methods)$ ckan run
Option ckan.requests.timeout is not declared
Option ckan.requests.timeout is not declared
2022-10-21 10:02:24,977 WARNI [ckan.common] Option ckan.plugins is not declared
2022-10-21 10:02:25,463 WARNI [ckan.config.middleware.flask_app] Extensions are excluded from CSRF protection! We allow extensions to run without CSRF protection but it will be forced future releases. Read the documentation for more information on how to add CSRF protection to your extension.
Option ckan.requests.timeout is not declared
Option ckan.requests.timeout is not declared
2022-10-21 10:02:26,044 WARNI [ckan.common] Option ckan.plugins is not declared
2022-10-21 10:02:26,475 WARNI [werkzeug]  * Debugger is active!
2022-10-21 10:02:26,529 WARNI [ckan.config.middleware.flask_app] Extensions are excluded from CSRF protection! We allow extensions to run without CSRF protection but it will be forced future releases. Read the documentation for more information on how to add CSRF protection to your extension.
Option ckan.requests.timeout is not declared
Option ckan.requests.timeout is not declared
2022-10-21 10:02:41,390 WARNI [ckan.common] Option ckan.plugins is not declared
2022-10-21 10:02:41,855 WARNI [werkzeug]  * Debugger is active!
2022-10-21 10:02:41,909 WARNI [ckan.config.middleware.flask_app] Extensions are excluded from CSRF protection! We allow extensions to run without CSRF protection but it will be forced future releases. Read the documentation for more information on how to add CSRF protection to your extension.
### CALLING after_search!! ###


```

2) Extensions like `ckanext-showcase` get into a recursion error:

```
2022-10-21 09:58:15,878 ERROR [ckan.views.api] maximum recursion depth exceeded
  File "/home/pdelboca/Repos/ckan/ckan/config/middleware/../../views/api.py", line 275, in action
  File "/home/pdelboca/Repos/ckan/ckan/logic/__init__.py", line 551, in wrapped
  File "/home/pdelboca/Repos/ckan/ckan/logic/action/get.py", line 1885, in package_search
    data_dict = item.before_dataset_search(data_dict)
  File "/home/pdelboca/Repos/ckanext-showcase/ckanext/showcase/plugin/__init__.py", line 224, in before_search
    return self.before_dataset_search(search_params)
  File "/home/pdelboca/Repos/ckanext-showcase/ckanext/showcase/plugin/__init__.py", line 224, in before_search
    return self.before_dataset_search(search_params)
  File "/home/pdelboca/Repos/ckanext-showcase/ckanext/showcase/plugin/__init__.py", line 224, in before_search
    return self.before_dataset_search(search_params)
  [Previous line repeated 974 more times]
RecursionError: maximum recursion depth exceeded
```

Probably by the way they're implementing the backwards compatibility: https://github.com/ckan/ckanext-showcase/blob/07ee01c5210e0c6fa736e0f86272be79e7393d02/ckanext/showcase/plugin/__init__.py#L205-L224

smotornyuk:
1. Yes, that's how python warnings work - you won't see them unless you run the application with a custom warnings-level set via the `-W` interpreter's flag or via the `PYTHONWARNINGS` environment variable. But you are right, not everyone knows how to properly check warnings, so it has the sense to duplicate these messages as log records. Anyway, [we are already doing something similar](https://github.com/ckan/ckan/blob/master/ckan/lib/maintain.py#L52-L53) 
2. Yep, it's completely ckanext-showcase's fault. As I understand, it tries to handle the bug that is fixed by the current PR in its own way. But, obviously, it was a bad decision. It would be much better to do something like this:
```python
    # IPackageController
    def _after_search_common(self, search_results, search_params):
        ...

    after_search = _after_search_common
    after_package_search = _after_search_common
```
Anyway, this hack from ckanext-showcase can be just removed after this PR merged into core

PR #7165, opened 2022-10-30
Fix duplicated template call

### LINKED ISSUE #7152: secondary_content template block is being executed twice.
**CKAN version**
master
2.9.6

## Describe the bug
Template's block `secondary_content` is being executed twice on each page load

## Steps to reproduce

Creating a testing helper and adding it to any `secondary_content` block it's enough. In my case I have the following helper:

```python
    # ITemplateHelpers

    def get_helpers(self):

        def testing_helper():
            print("Testing Helper!")

        return {
            "testing_helper": testing_helper
        }
```

And I added into whatever `secondary_content` block like in `ckan/templates/package/search.html` or `ckan/templates/organization/read_base.html` like for example:
```jinja
{% block secondary_content %}
  {{ h.testing_helper() }}
  {% snippet 'snippets/organization.html', organization=group_dict, show_nums=true, group_type=group_type %}
  {% block organization_facets %}{% endblock %}
{% endblock %}
```

It will be executed twice:
```
2022-10-24 13:27:21,469 WARNI [werkzeug]  * Debugger is active!
2022-10-24 13:27:21,523 WARNI [ckan.config.middleware.flask_app] Extensions are excluded from CSRF protection! We allow extensions to run without CSRF protection but it will be forced future releases. Read the documentation for more information on how to add CSRF protection to your extension.
Testing Helper!
Testing Helper!

```

**Expected behavior**

`secondary_content` block is executed only once.

pdelboca:
Fixes #7152 

### Proposed fixes:

This PR directly removes the code since it has no effect under the current implementation:
 - `c.action == "resource_read"` will never evaluate to True since `resource_read` is part of the old Pylons implementation of ruotes.
 - `self.secondary()|trim == ''` will never evaluate to True in core since the secondary block itself already renders an `aside` element.
 - Extensions may override the whole `secondary` block to not render anything but in that case what's the use of having a `no-nav` class if nothing is displayed?

### DIFF
--- a/ckan/templates-bs3/page.html
+++ b/ckan/templates-bs3/page.html
@@ -45,7 +45,7 @@
             </div>
           {% endblock %}
 
-          <div class="row wrapper{% block wrapper_class %}{% endblock %}{% if self.secondary()|trim == '' or c.action=='resource_read' %} no-nav{% endif %}">
+          <div class="row wrapper{% block wrapper_class %}{% endblock %}">
             {#
             The pre_primary block can be used to add content to before the
             rendering of the main content columns of the page.
--- a/ckan/templates/page.html
+++ b/ckan/templates/page.html
@@ -45,7 +45,7 @@
             </div>
           {% endblock %}
 
-          <div class="row wrapper{% block wrapper_class %}{% endblock %}{% if self.secondary()|trim == '' or c.action=='resource_read' %} no-nav{% endif %}">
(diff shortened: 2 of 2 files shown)


pdelboca:
@Zharktas I think that it is time to give this PR a new chance now that CKAN 2.10 is out.

This is a performance related and it will be nice to merge it soon so we can deal with whatever CSS issue it can cause in extensions with time.

What do you think?

PR #7175, opened 2022-11-03
Added a `search-index` subcommand `list-unindexed`

JVickery-TBS:
Adds a subcommand to `ckan search-index` called `list-unindexed` which lists unindexed packages that are in the database.

Includes a command option `-p` `--include-private` to also query private packages along with public ones (in the case that users need to index both public and private packages). Default is to only report on public packages, i.e. `--include-private=False`

### Features:

- [ ] includes API changes
- [X] None of the above

### DIFF
--- a/changes/7175.feature
+++ b/changes/7175.feature
@@ -1,0 +1,3 @@
+Added `list-unindexed` sub-commands to the `search-index` command.
+
+Accepts `-p` or `--include_private` flag to list all unindexed packages. Otherwise, it will only list unindexed public packages.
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
Not too sure how to fix the one lint error as `filter` needs a conditional statement.

wardi:
I'd remove the private flag and default it to True - that should be more in line with all the other `search-index` commands, right?
Also is there a way to share more code with the "refresh" option because it has to collect the same information, right?

wardi:
rather, try using the same method as here https://github.com/ckan/ckan/blob/4a438423a4f5752509e812ddb1ec349adf456781/ckan/lib/search/__init__.py#L228-L238 (get_all_entity_ids)

JVickery-TBS:
@wardi I changed my method to just re-use a bunch of code from the `search-index rebuild --only-missing`. Is cleaner and makes a lot more sense now I think.

I did test this a bunch locally and it works.

PR #7178, opened 2022-11-06
Finish updating from control-label to form-label

pdelboca:
When migrating to Bootstrap 5 we changed our custom `control-label` class for the native `form-label`. This PR finishes the migration and cleans the attribute from our less files.

### Proposed fixes:
Use Bootstrap native `form-label` class in all forms as we do in our `input_block` template.

https://github.com/ckan/ckan/blob/4a438423a4f5752509e812ddb1ec349adf456781/ckan/templates/macros/form/input_block.html#L23-L32

### DIFF
--- a/ckan/public/base/css/main-rtl.css
+++ b/ckan/public/base/css/main-rtl.css
@@ -11888,10 +11888,6 @@ textarea {
   width: 100%;
 }
 
-.control-large .control-label {
-  font-size: 1.09375rem;
-}
-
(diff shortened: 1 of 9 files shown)


pdelboca:
@amercader BTW, this PR will fix also the alignment of the `Clear Upload` button listed in #6840 since the `control-label` class in the input of `resource_upload_field` is misplacing the button.

image

PR #7191, opened 2022-11-09
Fix 404 when selecting the same date in the changes view

amercader:
In the `/dataset/<id>/changes_multiple` page, if the user selects the same date in both select controls a 404 is returned. The case when the second date was more recent than the first one is already handled (selecting the most recent version), we just need to account for the case when both dates are the same.

https://user-images.githubusercontent.com/200230/200816966-83fc7fbe-00d4-4d8f-9501-197fc782a6b6.mp4

This could probably be handled better with more clever select controls that update with the relevant options but this is a good first patch

### DIFF
--- a/changes/7191.bugfix
+++ b/changes/7191.bugfix
@@ -1,0 +1,1 @@
+Fix 404 error when selecting the same date in the changes view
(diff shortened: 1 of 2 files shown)


kowh-ai (approved):
Tested this PR on a new CKAN Master installation - works great!

kowh-ai:
@amercader - another scenario I tested was where there was just 1 update to a dataset, therefore 1 activity associated with this dataset - clicking through the "Changes" link in the Activity Stream tab also caused a 404

PR #7205, opened 2022-11-12
Clean old ckan-icons

pdelboca:
Since we are using FontAwesome 6 there's no longer a need for maintaining our own `ckan-icon` CSS structure. This PR removes it since already most of them are not being used.

### Proposed fixes:
 - Removes old `subscribe.html` template that it is not being used anymore.
 - Migrate the popular snippet to use FontAwesome (and to actually fixes it since it is no longer working)
 - Changes the remove icon for an X  icon.
 - Removes `_icons.scss` since it is no longer needed
 - Cleans some old `whithin_tertiary` CSS class and logic

### Demo
[Screencast from 11-15-22 14:53:20.webm](https://user-images.githubusercontent.com/6672339/201936863-93e51736-974d-45e6-8ea5-50f2ca1c6853.webm)

### Before
Screenshot from 2022-11-12 13-01-58

### After
Screenshot from 2022-11-15 14-59-13


amercader:
The cross icon in the selected facet does actually provide a function, remove that particular filter from the search. How would you do that in the "After" UI?

pdelboca:
@amercader I have updated the PR to add an `x` icon at the end of each facet to indicate the user that it can be removed by clicking again. 

Let me know your thoughts!

Screenshot from 2022-11-15 14-59-13


cc @EricSoroos

amercader:
Thanks @pdelboca I polished the facet close button a bit in [707c705](https://github.com/ckan/ckan/commit/707c705afba6e5efbb1cf53d36476502fe845afe)

Screenshot 2022-11-16 at 11-11-30 Dataset - CKAN

pdelboca:
Thanks @amercader ! It looks way nicer 😄

PR #7208, opened 2022-11-15
Change user creation defaults

amercader:
As agreed, this tightens up the default values with regards to user creation to avoid inadvertently open up sites to spam.

1. Restrict user and group/org image upload formats by default. Effectively just PNG, GIF and JPG. No SVGs. Additional formats can be added via `ckan.upload.user.mimetypes` and `ckan.upload.group.mimetypes`
2. Disable public registration of users by default (`ckan.auth.create_user_via_web = false`) as this is the most common
deployment scenario for CKAN. With this default value, new users can be created by:
  * Being invited by an organization admin,
  * Being created directly by a sysadmin in the ``/user/register`` endpoint, or
  * Being created in the CLI using ``ckan user add``

Adapted the tests to the new defaults

---

Aside: `ckan.auth.create_user_via_web` really means "create users, period": 

https://github.com/ckan/ckan/blob/addd3fbe2e9a8ea8726f46c7ca3b6ddd1d257d48/ckan/logic/auth/create.py#L165-L167

So if it's set to False, direct calls to `user_create` from plugins or tests that are not made by a sysadmin or include `ignore_auth:True` will fail with a NotAuthorized error. I think that's fine even if the naming is unfortunate, and decided not to try to be clever and check if it was an actual web UI request in the auth function


kowh-ai (approved):
Changes look good - ran through a series of tests, all good

PR #7209, opened 2022-11-15
Separate requirements install instruction

pdelboca:
A common issue for some installs are the requirements for Solr which are not mandatory if the user will run it using Docker.

This PR proposes a small change into the introduction of the instructions to install it from source to:

 - Explicitly name all the service required (I think it provide a good general view of the upcoming install process)
 - Separate each requirement in a different command so specifically link it to the service.
 - Cleans the old `who.ini` file section since it's no longer needed in CKAN 2.10.

### DIFF
--- a/doc/maintaining/installing/install-from-source.rst
+++ b/doc/maintaining/installing/install-from-source.rst
@@ -26,8 +26,8 @@ work on CKAN.
 1. Install the required packages
 --------------------------------
 
-If you're using a Debian-based operating system (such as Ubuntu) install the required
-packages with this command::
+If you're using a Debian-based operating system (such as Ubuntu) install the
+required packages with this command::
 
     sudo apt-get install python3-dev libpq-dev python3-pip python3-venv git-core redis-server
 
--- a/doc/maintaining/installing/install-from-source.rst
+++ b/doc/maintaining/installing/install-from-source.rst
@@ -24,10 +24,26 @@ work on CKAN.
 1. Install the required packages
 --------------------------------
 
-If you're using a Debian-based operating system (such as Ubuntu) install the
-required packages with this command::
+CKAN is a Python application that requires three main services: PostgreSQL, Solr and Redis.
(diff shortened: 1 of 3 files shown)


doc/maintaining/installing/install-from-source.rst
```diff
@@ -24,10 +24,26 @@ work on CKAN.
...
+The 3 main services (PostgreSQL, Solr and Redis) can be installed locally or dockerized. If
+you want to run them locally you will need to install the following dependencies.
+
+PostgreSQL requirements::
+
+    sudo apt-get install postgresql
+
+Solr requirements::
+
+    sudo apt-get install solr-tomcat openjdk-8-jdk
```
amercader, line 42: I think we should encourage people to use the Dockerized Solr version whenever possible, so let's remove these instructions for postgres and solr, as people will find them in their dedicated pages (actually check the postgres one). redis-server can be added back to the packages above

pdelboca: Done @amercader !

pdelboca:
I have implemented your suggestions @amercader ! Let me know your thoughts!

doc/maintaining/installing/install-from-source.rst
```diff
@@ -4,6 +4,8 @@
 Installing CKAN from source
 ===========================
 
+CKAN is a Python application that requires three main services: PostgreSQL, Solr and Redis.
+
 This section describes how to install CKAN from source. Although
 :doc:`install-from-package` is simpler, it requires Ubuntu 18.04 64-bit or
 Ubuntu 16.04 64-bit. Installing CKAN from source works with other
```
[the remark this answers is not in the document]
EricSoroos: There's also a place in the troubleshooting that mentions 16.04 and 18.04.

EricSoroos:
We should probably merge this prior to 2.10 & backport

PR #7216, opened 2022-11-17
Remove log warnings about ckan.requests.timeout not being declared

amercader:
When running any `ckan` command there were these two lines coming up:

```
Option ckan.requests.timeout is not declared
Option ckan.requests.timeout is not declared
```

That's because we were trying to access this config option at import time with `get_value()`, before the config has been properly parsed. This moves the assignments to the functions, where config is properly initialized.


pdelboca:
Looks good @amercader ! Thanks for fixing this!

PR #7217, opened 2022-11-17
Add option to output token value without extra messages

amercader:
When running `ckan user token add` you get this output by default:

```
API Token created:
        eyJ0eXAiOiJKV1QiL...
```

This is user-friendly, but is not great for integrating in automated scripts as you need to clean up the output a bit, eg:

```
export CKAN_API_TOKEN=$(ckan user token add ckan_user my_token | tail -n 1 | tr -d '\t')
```

### Proposed fixes:

The changes suggested add a `--quiet` / `-q` flag that make the command just output the token directly so you can do:

```
export CKAN_API_TOKEN=$(ckan user token add ckan_user my_token -q)
```

### Features:

### DIFF
--- a/changes/7217.misc
+++ b/changes/7217.misc
@@ -1,0 +1,1 @@
+Add ``--quiet`` option to ``ckan user token add`` command to mak easier to integrate with automated scripts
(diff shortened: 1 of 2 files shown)


kowh-ai (commented):
Looks good @amercader!!
Screenshot 2022-11-23 at 11 09 24 am

PR #7223, opened 2022-11-18
[#7222] Use parsed config values in Flask config

### LINKED ISSUE #7222: Exception on a fresh install / no session when trying to login
**CKAN version**
master/2.10

**Describe the bug**
On a fresh install, if `REMEMBER_COOKIE_DURATION`  is set in the ini file I get an exception trying to login

```
  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask_login/login_manager.py", line 476, in _set_cookie
    expires = datetime.utcnow() + duration
TypeError: unsupported operand type(s) for +: 'datetime.datetime' and 'str'

The above exception was the direct cause of the following exception:

  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask/app.py", line 1540, in finalize_request
    response = self.process_response(response)
  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask/app.py", line 1885, in process_response
    response = self.ensure_sync(func)(response)
  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask_login/login_manager.py", line 447, in _update_remember_cookie
    self._set_cookie(response)
  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask_login/login_manager.py", line 478, in _set_cookie
    raise Exception(
Exception: REMEMBER_COOKIE_DURATION must be a datetime.timedelta, instead got: 31536000
```

**Steps to reproduce**
* Add `REMEMBER_COOKIE_DURATION = 31536000` to your ini file
* Delete the sessions folder in `/tmp/<ckan.site_id>`
* Try to login

**Additional details**
Based on the stack trace complaining about `duration` being a str, I guess that the problem is that we are passing raw config values to Flask's `app.config` object instead of parsed ones or something along those lines

amercader:
Fixes #7222 

Ensure that we apply the validators to config values before assigning them to the Flask config object.
In strict mode this has been done before in `load_environment` but in default mode we need to call `get_value()` on all keys to make sure they are parsed.

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -163,7 +163,13 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     # Update Flask config with the CKAN values. We use the common config
     # object as values might have been modified on `load_environment`
     if config:
-        app.config.update(config)
+        if config.get_value("config.mode") == "strict":
+            # Config values have been already parsed and validated
+            app.config.update(config)
+        else:
+            # Parse all values to ensure Flask gets the validated values
+            for key in config.keys():
+                app.config[key] = config.get_value(key)
     else:
         app.config.update(conf)
 
--- a/ckan/tests/config/test_middleware.py
+++ b/ckan/tests/config/test_middleware.py
@@ -93,3 +93,18 @@ def test_all_plugin_blueprints_are_registered(app):
     assert url == "/another_simple_url"
     res = app.get(url, status=200)
     assert "Hello World, this is another view served from an extension" in res.body
+
+
+@pytest.mark.ckan_config("REMEMBER_COOKIE_DURATION", "12345")
+def test_flask_config_values_are_parsed(app):
+    assert (
+        app.flask_app.config["REMEMBER_COOKIE_DURATION"] == 12345
+    )
+
+
+@pytest.mark.ckan_config("config.mode", "strict")
+@pytest.mark.ckan_config("REMEMBER_COOKIE_DURATION", "12345")
+def test_flask_config_values_are_parsed_in_strict_mode(app):
+    assert (
+        app.flask_app.config["REMEMBER_COOKIE_DURATION"] == 12345
+    )


amercader:
@smotornyuk this seems like the simpler approach but let me know otherwise

smotornyuk:
I like the idea of passing validated options to the Flask app, but there are a few concerns:

* it means that `toolkit.config` and `app.config` can contain different values in the `default` configuration mode. It's a pretty low-level thing, but aren't you expecting these two objects to have the same values for matching keys?
* `strict` mode validates only declared options. But here, every key from `toolkit.config` goes through `config.get_value`. It means that every non-declared config option(configuration for extensions, for example) will produce a warning, like this: 
image. I foresee tons of complaints in GitHub issues/gitter channel from people who see it for the first time.


Probably, the first concern can't be helped, and only switching to the strict mode solves this problem.
As for second one, how about adding optional argument for `config.get_value` that suppresses the warning? Or, introducing `config.is_declared` that can be used before attempting `config.get_value` in this solution:
```diff
diff --git a/ckan/common.py b/ckan/common.py
index 8eb693926..aad957039 100644
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -128,6 +128,9 @@ class CKANConfig(MutableMapping):
         except RuntimeError:
             pass

+    def is_declared(self, key: str) -> bool:
+        return key in config_declaration
+
     def get_value(self, key: str) -> Any:
         if self.get("config.mode") == "strict":
             return self[key]
diff --git a/ckan/config/declaration/__init__.py b/ckan/config/declaration/__init__.py
index cd8953b9b..d8beb0e53 100644
--- a/ckan/config/declaration/__init__.py
+++ b/ckan/config/declaration/__init__.py
@@ -64,7 +64,7 @@ class Declaration:
     def __bool__(self):
         return bool(self._members)

-    def __contains__(self, key: Key):
+    def __contains__(self, key: Union[Key, str]):
         return key in self._options

     def __getitem__(self, key: Key) -> Option[Any]:
diff --git a/ckan/config/middleware/flask_app.py b/ckan/config/middleware/flask_app.py
index b9904f51d..59069d7be 100644
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -169,7 +169,9 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
         else:
             # Parse all values to ensure Flask gets the validated values
             for key in config.keys():
-                app.config[key] = config.get_value(key)
+                if config.is_declared(key):
+                    app.config[key] = config.get_value(key)
+
     else:
         app.config.update(conf)

```

amercader:
Thanks for the input @smotornyuk, to your points:

> it means that toolkit.config and app.config can contain different values in the default configuration mode. It's a pretty low-level thing, but aren't you expecting these two objects to have the same values for matching keys?

Well, yes ideally they would be the same value, but if we keep both values, then Flask login crashes because `REMEMBER_COOKIE_DURATION` is not an int, and we decided to not parse and validate things in `default` mode to avoid breaking things at the CKAN level so I think validating the ones passed to Flask (which we or extension writers can't control) is a good compromise.

I like adding an `is_declared()` method and I've updated this PR but I think your suggestion was missing an extra check to also apply undeclared options to `app.config` ([ce2453a](https://github.com/ckan/ckan/pull/7223/commits/ce2453a4891ec0b3f805f5c3363582eee647efb9)):

``
index b9904f51d..59069d7be 100644
@@ -169,7 +169,9 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
         else:
             # Parse all values to ensure Flask gets the validated values
             for key in config.keys():
-                app.config[key] = config.get_value(key)
+                if config.is_declared(key):
+                    app.config[key] = config.get_value(key)
+                else:
+                    app.config[key] = config.get(key)
+
     else:
         app.config.update(conf)
```

PR #7231, opened 2022-11-22
Updated readme

shashigharti:
Fixes #

Added note in README for `solr_url` when installed using docker.

### Proposed fixes:

Update README

### Features:

- [ ] includes API changes


amercader:
Hi @shashigharti , thanks for this PR!
The Solr docs were completely rewritten for the latest version (master/2.10):

https://docs.ckan.org/en/latest/maintaining/installing/install-from-source.html#setup-solr

Rather than change the 2.9 docs perhaps is better to port the whole solr install docs currently in master to the `dev-v2.9` branch (we don't merge stuff to the `2.9` one, that's used for docs) so they are both consistent.

What do you think?

shashigharti:
Thank you @amercader!. I agree and will submit a new PR and close this one.

PR #7241, opened 2022-11-26
Add ckan user clean command

### LINKED ISSUE #7225: User clean up command
Some auth configurations can lead to the creation of spam user account. Default changes in #7208 will hopefully make more difficult for maintainers to end up in this situation but we can also provide a small command to cleanup spam user accounts. 

The logic should be:

* Filter any user accounts that
   1. Has a value for `image_url`
   2. The file referenced in `image_url` does not match one of the ones allowed by `ckan.uploads.user.mimetypes`
* Ask user to confirm (unless the `-f / --force` flag is passed) showing a list of the accounts that will be deleted (if admins want to keep one they can update the image_url value)
* Delete the user object and the linked uploaded file

Where should this command be available? 
1. Keep it on `ckan user clean` or something similar
2. Add a new `ckan clean ....` or `ckan cleanup ...` command that we can later expand to eg purge datasets / orgs, delete old activities etc

Any thoughts?

pdelboca:
Fixes #7225  part of #7208 

### Proposed fixes:

This PR adds a new command `ckan clean users` that will scan the system for users with invalid images and delete both the image and the user from the database.

Invalid images are all images that doesn't contain allowed mimetypes set in `ckan.upload.user.mimetypes` config option.

### Features:

- [ ] includes API changes


pdelboca:
For now I'm trying to keep a logic that do not delete the user if the image cannot be deleted. (Permission issues, ..., ?) This is to avoid leaving unreferenced images in the store.

pdelboca:
The second issue I'm getting is that this logic is aggressive and considers a user account as spam if they have a wrong file in the image. 

What about false positives? What if the administrator forgets one mimetype?

amercader (commented):
Looks good @pdelboca only one minor tweak

pdelboca:
Done @amercader !

Failing Cypress test seems unrelated.

PR #7243, opened 2022-11-28
Fix failing test after updating Werkzeug to 2.1.1

pdelboca:
When updating Werkzeug to 2.1.1 a test started failing.

This is because in this new version, Werkzeug does not autocorrect the `Location` header by default:

https://github.com/pallets/werkzeug/blob/3115aa6a6276939f5fd6efa46282e0256ff21f1a/src/werkzeug/wrappers/response.py#L145-L153
```python
    #: If a redirect ``Location`` header is a relative URL, make it an
    #: absolute URL, including scheme and domain.
    #:
    #: .. versionchanged:: 2.1
    #:     This is disabled by default, so responses will send relative
    #:     redirects.
    #:
    #: .. versionadded:: 0.8
    autocorrect_location_header = False
```

### Proposed fixed

Since we are redirecting to a relative path, the expected value of the Location header should be a relative path.

https://github.com/ckan/ckan/blob/0aa468c0421088bc9f3bb65e8333c3471de697c2/ckan/views/home.py#L98-L105

### DIFF
--- a/ckan/tests/controllers/test_home.py
+++ b/ckan/tests/controllers/test_home.py
@@ -135,7 +135,7 @@ class TestI18nURLs(object):
             assert response.status_code == 308
             assert (
                 response.headers['Location'] ==
-                f'/{new_locale}'
+                f'http://test.ckan.net/{new_locale}'
             )
 
             response = app.get(f'/{legacy_locale}/dataset?some=param', follow_redirects=False)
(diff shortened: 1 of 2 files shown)


pdelboca:
@amercader as we discussed I have modified the view to redirect to a full URL so it matches what the test expect.

However this raises some questions:
 - Why in this particular case we need to redirect using a full URL?
 - What's the difference between using `flask` redirect (as we are using here) and `h.redirect_to(...)` method?

amercader:
@pdelboca it's just safer to ensure we redirect to an internal site page to avoid open redirect issues. The `h.redirect_to` function is a wrapper around `flask.redirect` to handle the locale part of the URL so maybe using it interfered with this particular endpoint as it deals with locales and URLs

wardi:
Fixes #8182

Issue #7083, opened 2022-09-14
Update CKAN to Flask 2.2.2 (or latest)


pdelboca:
CKAN 2.9.5 and next 2.10.0 will be running under Flask 2.0.3.

Migration to Flask 2.2.2 will have some work that we would address once 2.10.0 is released. Some of the key points to be able to run the latest version:
- [x] We will need to update from Flask-Login `0.6.1` to `0.6.2` (See #7064 description)
- [X] We will need to fix/patch/replace `flask-multistatic` since it will not work with latest Werkzeug. It looks like the package it's not being maintained in a while, since it only and last release was 6 years ago: https://pypi.org/project/flask-multistatic/#history 
- [x] Fixing all tests ~122 failing tests when quickly fixing `flask-multistatic` locally.
- [x] We still can't update Flask-Babel to 2.0.0 as it breaks ITranslation (see https://github.com/ckan/ckan/issues/6708 and
https://github.com/python-babel/flask-babel/pull/194 cc @TkTech ) -> PR https://github.com/ckan/ckan/pull/7436
- [ ] `dev-requirements.txt`: docutils pinnned at 0.17 because sphinx-rtd-theme doesn't support a bigger version
- [ ] `dev-requirements.txt`: pytest-factoryboy pinnned at 2.4.0 because on 2.5.0 we get:

amercader:
On the Flask-login failures, I'm pretty sure that they are caused by this change: https://github.com/maxcountryman/flask-login/pull/691

Somehow the combination of storing the logged in user in `g` instead of the stacks and the way CKAN tests with the `app` fixture are run breaks tings. We need to keep digging.

PR #7244, opened 2022-11-28
Integrate flask-multistatic into our own library

pdelboca:
Fixes one of the points detailed in https://github.com/ckan/ckan/issues/7083. Necessary to update to Flask 2.2.2.

The plugin `flask-multistatic` is no longer supported and it is a key component of CKAN's architecture since it allows CKAN extensions to register static folders.

Flask currently support a way for each blueprint to define new static folders but it doesn't match our current requirements since we provide the option to add static folders without creating new Blueprints.

While reviewing `flask-multistatic` code I noticed that it is actually a pretty simple plugin so I came up with the idea of integrating it to our lib.

### Proposed fixes:

- Adds a new module called `flask-multistatic`
- Fixes the deprecated paramenter `cache_timeout`

### Features:

- [ ] includes API changes


tino097 (approved):
LGTM

PR #7246, opened 2022-11-29
Refactor _internal_test_request_context into fixture

pdelboca:
### Proposed fixes:

This PR refactors the global `_internal_test_request_context` variable into a fixture that pushes a request context whenever we are using the `app` fixture

### Notes:

This solution is based in https://github.com/pytest-dev/pytest-flask fixtures. Note that using `url_for` in tests is not considered a good practice, however we have adopted it widely.

Also, some internal methods in `model_dictize.py` depends on calling `h.url_for(...)` making them dependent on a Request Context even when they are a low level function.


pdelboca:
Close in favor of #7251

PR #7254, opened 2022-12-02
Fix config option call

pdelboca:
This PR fixes the error happening in master:
```python
    def natural_number_validator(value: Any, context: Context) -> Any:
        """Ensures that the value is non-negative integer.
        """
        value = int_validator(value, context)
>       if value < 0:
E       TypeError: '<' not supported between instances of 'NoneType' and 'int'

ckan/logic/validators.py:138: TypeError
```

### Proposed fixes:

This PR fixes a call to get values from the config object. The current logic doesn't cast the value to `int` and therefore causes an error later in the validation process.

```python
ipdb> config.get_value('ckan.datasets_per_page')
20
ipdb> config.get('ckan.datasets_per_page', 10)
'20'
```

The fix is simple however this error raises two concerns:
 - Having `config.get(...)` and `config.get_value(...)` is confusing. We should document this better and edit the `CKANConfig` object to avoid this common mistake.

### DIFF
--- a/changes/7254.bugfix
+++ b/changes/7254.bugfix
@@ -1,0 +1,2 @@
+Fixes a bug causing `ckan.datasets_per_page` config not being used.
+`limit` parameter in group/organization view has been removed in favor of the config.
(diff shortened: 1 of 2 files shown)


amercader:
Thanks @pdelboca. BTW, no need to add changelog entries for things that didn't change between 2.9 and 2.10 (even with the annoying check failure)

Issue #7200, opened 2022-11-09
Translation "group" to german "Gruppe"?


ondics:
In the CKAN user interface the english word `group"`is translated as `Gruppe`. 

We have many CKAN installations where customers like to see `Kategorie` instead of `Gruppe`.

This means we had to consistently change the german translations in 
https://www.transifex.com/okfn/ckan/translate/#de/2-9/294238824?q=text%3Agroup

This requires some work, but should we start to change this? What are your opinions?

EricSoroos:
The feedback I'm getting is that `Kategorie` wouldn't apply to people (and presumably users).

ondics:
Sorry, top be more precise: I'm talking about CKAN-groups for datasets. And most of our customers had us to change the translation from "Gruppe" to "Kategorie".

You are right, that this is no general translation but in in the context of CKAN dataset grouping "Kategorie" applies better than "Gruppe".

The "group" translation with relation to users or people should remain "Gruppe" of course.

EricSoroos:
The trouble here is that users can be added to groups as well, so both senses of the word are used. It's not just datasets.

![image](https://user-images.githubusercontent.com/36156267/201145712-f7a3cff9-8b9e-4e4f-83a1-334ee02fca71.png)

EricSoroos:
I think that the best response here would be to have the naming for this concept be more flexible -- as we tend to rename organizations to other concepts in our sites, I think groups are a building block that are used for different purposes on different sites and deserve different names.

abedkhooli:
I can see Groups is not translated when I switch to Arabic despite being in the po file under i18n/ar/LC_MESSAGES (Ckan 2.10). Is there an easy way I can enable it?

smotornyuk:
@abedkhooli , it will be fixed by #7500 7500

abedkhooli:
Thanks @smotornyuk 
Changed in header.html and worked.

ondics:
@smotornyuk @abedkhooli your problem is now fixed but that is different from my quesions raised.

In most cases groups are used to classify dataset in different categories. When looking at DCAT groups are used for [classifying datasets thematically](https://www.w3.org/TR/vocab-dcat-2/#classifying-datasets). And several bodies as the EU have defined vocabularies for this ["data theme classification"](https://op.europa.eu/en/web/eu-vocabularies/concept-scheme/-/resource?uri=http://publications.europa.eu/resource/authority/data-theme).

[...]

PR #7257, opened 2022-12-05
config.get -> config.get_value / extra config declarations

smotornyuk:
❗  contains a breaking change

* Change a couple of `config.get` to `config.get_value` 
* Add a few missing config declarations

### Breaking change

Before, `toolkit.aslist` wrapped any value other than str/list/tuple into a list. I.e `aslist({1,2}) -> [{1,2}]`. Now iterables will be converted to the list, i.e `aslist({1,2}) -> list({1,2}) -> [1,2]`. Non-iterables behave in the same way: `aslist(1) -> [1]` 

Not sure if such a change is acceptable. But documentation of this method has no mention of types other than `str`, so there were no guarantees, right? :)
In this particular case, we need this change because of the `WTF_CSRF_METHODS` config options. It's declared as a list, but internally flask-wtf converts it into the set. *If* validators are applied to this converted value again(it generally happens only in tests, but it's a possible case for the real world as well and I even mentioned in in [the docs](https://docs.ckan.org/en/latest/maintaining/configuration.html#declaring-config-options) - check first note about idempotence of validators), we'll get `aslist({"POST", "PUT"}) -> [{"POST", "PUT"}]`. And then we'll get a type error when `flask-wtf` try to convert this value into a set once again.

*We can* create a separate validator or even declaration type(i.e, allow using something like `type: set` in the declaration file), but:
* It's really a rare case. Even though I would choose it, there is a second reason:
* We are releasing a new CKAN version. It's the best time to introduce breaking changes. And this particular change makes code more intuitive, isn't it?

ℹ️ Tests are fixed in #7254

### DIFF
--- a/changes/7257.feature
+++ b/changes/7257.feature
@@ -1,2 +1,2 @@
-`toolkit.asbool` now converts any iterable other than `list` and `tuple` into a `list`: `list(value)`.
+`toolkit.aslist` now converts any iterable other than `list` and `tuple` into a `list`: `list(value)`.
 Before, such values were just wrapped into a list, i.e: `[value]`
--- a/changes/7257.feature
+++ b/changes/7257.feature
@@ -1,2 +1,22 @@
-`toolkit.aslist` now converts any iterable other than `list` and `tuple` into a `list`: `list(value)`.
-Before, such values were just wrapped into a list, i.e: `[value]`
+`toolkit.aslist` now converts any iterable other than ``list`` and `tuple` into a ``list``: ``list(value)``.
+Before, such values were just wrapped into a list, i.e: ``[value]``.
+
+.. list-table:: Short overview of changes
+   :widths: 40 30 30
+   :header-rows: 1
+
+   * - Expresion
+     - Before
+     - After
+   * - ``aslist([1,2])``
+     - ``[1, 2]``
+     - ``[1, 2]``
+   * - ``aslist({1,2})``
+     - ``[{1, 2}]``
+     - ``[1, 2]``
+   * - ``aslist({1: "one", 2: "two"})``
+     - ``[{1: "one", 2: "two"}]``
+     - ``[1, 2]``
+   * - ``aslist(range(1,3))``
+     - ``[range(1, 3)]``
+     - ``[1, 2]``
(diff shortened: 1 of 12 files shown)


changes/7257.feature
```diff
@@ -0,0 +1,2 @@
+`toolkit.asbool` now converts any iterable other than `list` and `tuple` into a `list`: `list(value)`.
```
amercader, line 1:
```suggestion
`toolkit.aslist` now converts any iterable other than `list` and `tuple` into a `list`: `list(value)`.
```

Also maybe it's easier to understand the change with an example

ckan/config/config_declaration.yaml
```diff
@@ -1241,16 +1254,29 @@ groups:
...
-        example: templates
+        example: templates-bs3
         description: |
           This config option is used to configure the base folder for templates used
           by CKAN core. It is currently unused and it only accepts one value: ``templates``
           (Bootstrap 3, the default value from CKAN 2.8 onwards).
 
+      - key: ckan.default.package_type
+        default: dataset
+        description: Default type of dataset that used in UI links(eg. **New Dataset**)
```
amercader, line 1265:
```suggestion
        description: Default type of dataset that will be used in the UI links (eg. "New Dataset")
```
Maybe an example would make this clear. Does it mean that if `ckan.default.package_type=record`, the UI will show "New Record"? And the same for groups and orgs

amercader (commented):
@smotornyuk I think it's a reasonable breaking change to introduce if we document it properly in the changelog. I just added some comments to make clearer the changelog entry and the config docstrings

ckan/config/config_declaration.yaml
smotornyuk: Yes, exactly. If you have a `record` package type defined, of course. I and Ian added this functionality 2 years ago. #5337 contains details for the group/organization. The package entity behaves in the same manner. 
This feature potentially provides an easy way to solve issues like #7200 or #5331, where people insist, that the "group" term does not reflect the idea of a group entity. Now they can just create a custom group type and set it as a default one. For non-standard types, that cannot be easily transformed into a human-readable label, chaining [`humanize_entity_type`](https://github.com/ckan/ckan/blob/master/ckan/lib/helpers.py#L1002) helper can solve a problem.

smotornyuk:
I accepted your suggestions and added another couple of lines about the breaking change and `default.package_type` option

Issue #5185, opened 2020-02-04
Password reset fails in internal server error when ckan.auth.public_user_details is configured to false


Zharktas:
### CKAN Version if known (or site URL)
2.8.3

### Please describe the expected behaviour
When [ckan.auth.public_user_details](https://docs.ckan.org/en/2.8/maintaining/configuration.html#ckan-auth-public-user-details) is configured to false, password reset requests should work.

### Please describe the actual behaviour
Password reset request fails in internal server error as https://github.com/ckan/ckan/blob/dev-v2.8/ckan/views/user.py#L511 does not include ignore_auth.

### What steps can be taken to reproduce the issue? 
Configure ckan.auth.public_user_details to false.
Try requesting password reset.

https://github.com/ckan/ckan/pull/4636 probably should be backported which includes ignore_auth to password reset.

Zharktas:
https://github.com/ckan/ckan/pull/4636/commits/8a5beb9b53d23173c8d072926a6b84f76131fcd8#diff-989e559988ddb2baccfd41061a98e135R512, ignore_auth was added in part on larger commit, so cherry-picking would add all sorts of things, so it might be that wiser just to add it with another commit instead.

ThrawnCA:
Testing of CKAN 2.8.4 shows that this is incomplete. Password reset emails can now be successfully sent, but the link to actually perform the reset does not work. 

Compare 0fbe25e281055d53279f20e17a7587a891fd8534, in the qld-gov-au fork, which fixes this.

Zharktas:
I'll look into it, probably today at some point.

Zharktas:
I tested reset links in 2.8.4 and the passwords were reset like they should. And they are quite well tested in tests too. You probably should file a new issue about it.

ThrawnCA:
@Zharktas To be clear, the failure happens when `ckan.auth.public_user_details` is false. I don't think that the tests cover that.

Zharktas:
@ThrawnCA well, technically you are correct, tests don't cover that. 

[...]

PR #7266, opened 2022-12-06
Fix user reset password

### LINKED ISSUE #7624: Activity Workflow in CKAN 2.10
## CKAN version
2.10.x

## Describe the change
When migrating activities to it's own extension, we changed the trigger for when the activity is created. Starting CKAN 2.10, activities are created using signals after a successful execution of actions (including chaining actions).

This means that a logic in CKAN 2.9 doing the following will no longer work under CKAN 2.10:
```python
@toolkit.chained_action
def package_create(up_func, context, data_dict):
    result = up_func(context, data_dict)
    _custom_method_doing_something_with_the_activity(result["id"]) # This will raise Activity not found
    return result
```

## Issues

Let me add an example. The extension https://github.com/datopian/ckanext-versions/ creates a `version` object that allows a user to add a name and a description to our activity object for a better UX. Basically each `version` is just a reference to an activity (like a `tag`):
image

Currently a workflow to create the first version for a resource automatically for the users is something like:
```python
@toolkit.chained_action
def package_create(up_func, context, data_dict):
    result = up_func(context, data_dict)
    for resource un result["resources"]: 
        create_initial_version(resource, "1.0", "Notes for initial version") # This internally gets the current activity and relates it to the activity
    return result
```

In CKAN 2.10 that workflow does not work anymore since  the first activity is created after all the chained actions of `package_create` are executed.

## Possible Fix (and another issue...)

So far the current I found that I could create a new signal to create the activity and I did something like this:

```python
# ckan.plugins = ... activity example_extension ...

def receiver_for_creating_version(sender, **kwargs):
    data_dict = kwargs.get("result")
    for resource in data_dict["resources"]: 
        create_initial_version(resource, "1.0", "Notes for initial version") 

class ExampleExtension(plugins.SingletonPlugin):
    plugins.implements(plugins.ISignal)
    
    def get_signal_subscriptions(self):
        return {
            plugins.toolkit.signals.action_succeeded: [
                {
                    "receiver": custom_method_handling_package_activity,
                    "sender": "package_create",
                }
            ]
        }
```
Initially this was working properly but then I started having random fails when executing tests. Some of them failing in our CD/CI pipeline but working locally, failing when executing in bulk but passing when executing only one, etc. Debugging I found out that sometimes my custom signal was being executed before `activity` signals and causing my logic to fail due to missing Activity.

Reading CKAN's documentation I found out that: _all signal handlers are executed in undefined_ therefore I understand that there is no way to ensure dependency on the activities.

I can do something like the following but of course doesn't look promising:

```python
def receiver_for_creating_version(sender, **kwargs):
    data_dict = kwargs.get("result")
    for resource in data_dict["resources"]: 
        try:
            create_initial_version(resource, "1.0", "Notes for initial version") 
        except toolkit.NotFound as e:
            # try again
            time.sleep(2)
            create_initial_version(resource, "1.0", "Notes for initial version") 
```

## Notes

I could separate the `package_create` and `resource_create` and in that way I will always have an activity when creating the resource but it doesn't solve the scenario when a full package (including resources) on the `package_create` call.

EricSoroos:
Fix issue where if ckan.auth.public_user_details is false, a user can't reset their password

Fixes #7624

### Proposed fixes:

Ignores auth when getting the user via user_show.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/user.py
+++ b/ckan/views/user.py
@@ -693,10 +693,18 @@ class PerformResetView(MethodView):
         except logic.NotAuthorized:
             base.abort(403, _(u'Unauthorized to reset password.'))
 
+        # Ignore auth for the user_show action only.
+        # user_show will fail for anon users if
+        # `ckan.auth.public_user_details` is False
+        context[u'ignore_auth'] = True
         try:
             user_dict = logic.get_action(u'user_show')(context, {u'id': id})
         except logic.NotFound:
             base.abort(404, _(u'User not found'))
+
+        # remove ignore auth for any future use of this context.
+        del(context[u'ignore_auth'])
+
         user_obj = context[u'user_obj']
         g.reset_key = request.params.get(u'key')
         if not mailer.verify_reset_link(user_obj, g.reset_key):
--- a/ckan/tests/controllers/test_user.py
+++ b/ckan/tests/controllers/test_user.py
@@ -885,6 +885,7 @@ class TestUser(object):
 
         assert "Error sending the email" in response
 
+
     def test_sysadmin_not_authorized(self, app):
         user = factories.User()
         env = {"REMOTE_USER": user["name"].encode("ascii")}
@@ -1037,6 +1038,21 @@ class TestUser(object):
 
         userobj = model.User.get(userobj.id)
         assert userobj.is_deleted(), userobj
+
+    @pytest.mark.ckan_config(u"ckan.auth.public_user_details", u"false")
+    def test_request_reset_public_user_details(self, app):
+        user = factories.User()
+        user_obj = helpers.model.User.by_name(user["name"])
+        create_reset_key(user_obj)
+        key = user_obj.reset_key
+        offset = url_for(
+            controller="user",
+            action="perform_reset",
+            id=user_obj.id,
+            key=key,
+        )
+        response = app.get(offset)
+        assert "Reset Your Password" in response
 
 
 @pytest.mark.usefixtures("clean_db", "with_request_context")


Zharktas:
There's an old issues https://github.com/ckan/ckan/issues/5185 and https://github.com/ckan/ckan/issues/6777 in which the same fix was suggested. The problem only occurred if `user_show` was modified additionally to setting the config option and was hesitant just to ignore authentication in this case as it might cause other unforeseen issues.  More or less the real fix is to let `user_show` to do its checks itself.

ckan/views/user.py
```diff
@@ -748,10 +748,18 @@ def _prepare(self, id: str) -> tuple[Context, dict[str, Any]]:
         except logic.NotAuthorized:
             base.abort(403, _(u'Unauthorized to reset password.'))
 
+        # Ignore auth for the user_show action only.
+        # user_show will fail for anon users if
+        # `ckan.auth.public_user_details` is False
+        context[u'ignore_auth'] = True
```
Zharktas, line 754:
```suggestion
        # Ignore auth for the user_show action only.
        # user_show might fail for anon users if
        # `ckan.auth.public_user_details` is False
        # and user_show has non core auth functions.
        context[u'ignore_auth'] = True
```

Zharktas (changes requested):
@EricSoroos I suggested slight modification to a comment, since password reset should work again after #7871 is merged. There are cases where this might fail and in the end ignore_auth should not cause problems here anyway. I would be willing to merge this if the comment would not be so absolute. If you agree with the suggestion that is.

ckan/views/user.py
```diff
@@ -748,10 +748,18 @@ def _prepare(self, id: str) -> tuple[Context, dict[str, Any]]:
...
         try:
             user_dict = logic.get_action(u'user_show')(context, {u'id': id})
         except logic.NotFound:
             base.abort(404, _(u'User not found'))
+
+        # remove ignore auth for any future use of this context.
+        del context[u'ignore_auth']
```
wardi, line 761: I tend to create a new context e.g.
```python
show_context = dict(context, ignore_auth=True)
```
instead of inserting and removing keys. We have issues where actions and auth functions modify the context that gets returned and creating a new context dict makes sure that doesn't lead to difficult to trace bugs

EricSoroos: Not sure I understand this, because we're explicitly getting the user_obj from the context for further use, and then we're returning that context to the caller. So we're clearly modifying the context in `user_show`, and those side effects might be used downstream. 

```
        context[u'ignore_auth'] = True
        try:
            user_dict = logic.get_action(u'user_show')(context, {u'id': id})
        except logic.NotFound:
            base.abort(404, _(u'User not found'))

        # remove ignore auth for any future use of this context.
        del context[u'ignore_auth']

        user_obj = context[u'user_obj']
        g.reset_key = request.args.get(u'key')
        if not mailer.verify_reset_link(user_obj, g.reset_key):
            msg = _(u'Invalid reset key. Please try again.')
            h.flash_error(msg)
            base.abort(403, msg)
        return context, user_dict
```

ThrawnCA:
Could this be handled instead by making `authz.is_authorized` recognise the context `user` object as an alternative to `auth_user_object`?

PR #7271, opened 2022-12-08
Update setuptools requirement

EricSoroos:
### Proposed fixes:
Unpin the old version of setuptools

The previous pin was likely for zope-interface < 5.0 or rdflib-jsonld, which was required for python 3.6 compatibility.

### Features:

- [ ] includes API changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,6 @@ jobs:
       - <<: *install_deps
       - run: |
           # Python Dependencies
-          pip install -r requirement-setuptools.txt
           pip install -r requirements.txt
           pip install -r dev-requirements.txt
           python setup.py develop
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -49,7 +49,7 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
 
       - name: Install python deps
-        run: pip install -r requirement-setuptools.txt -r requirements.txt -r dev-requirements.txt -e.
+        run: pip install -r requirements.txt -r dev-requirements.txt -e.
 
       - name: Init environment
         run: |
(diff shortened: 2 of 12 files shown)


EricSoroos:
This is core python infra -- I don't think that there's a valid reason to not let it float, unless we have evidence that specific versions don't work with a dependency that we're using. We should be compatible with anything after a minimum setuptools release.  

Now, granted, setuptools definitely has had some backwards incompatible changes, including why it was previously pinned at 44.1.0 and another one (locally) at 50something for zope.interface for py 3.6.

What I don't like seeing is having an up to date virtualenv start downgrading packages because there's a hard requirement that isn't really a hard requirement.

EricSoroos:
Are we ok with this approach, or should we completely drop the setuptools requirement?

pdelboca:
I have been testing using Python 3.10 when dropping the `setuptools` requirement and it's looking good.

image

image

Also, the brand new https://github.com/ckan/ckan-python-monitor shows all tests passing on Python 3.10 while installing it with newer versions of setuptools: https://github.com/ckan/ckan-python-monitor/actions/runs/3898314050/jobs/6656918822

@EricSoroos are you good to updated this PR?

amercader:
As discussed on dev call let's just drop this file entirely on master / 2.10 to support Python 3.10 and make our life easier going forward.

We'll also need to remove the step mentioning it from the source install docs:

https://docs.ckan.org/en/2.9/maintaining/installing/install-from-source.html?highlight=setuptools#install-ckan-into-a-python-virtual-environment

EricSoroos:
Ok, updated. It was a little bigger of an impact than I was expecting.

amercader:
@kowh-ai I've backported this to 2.10 so you'll probably have to update the 2.10 docker images to remove the requirements-setuptools.txt step in the Dockerfile

kowh-ai:
@amercader - yep, had to remove the `pip3 install -r requirement-setuptools.txt && \` line in the **master**  image I built the other day. Will update 2.10 this afternoon. Thanks

PR #6709, opened 2022-02-18
Update Python requirements

amercader:
All fairly uncontroversial. Things to note:

* This unpins pytz on master as discussed, so it fixes #6665 
* The pysolr==3.9.0 actually makes Solr 6 no longer work on CKAN master / 2.10 (as discussed and announced)

The only requirements still outdated are:
* SQLAlchemy: https://github.com/ckan/ckan/issues/6243
* Flask-Babel: https://github.com/ckan/ckan/issues/6708

We are getting there!

PR #7279, opened 2022-12-15
Fixed icons and alignment in activity stream

### LINKED ISSUE #7247: UI broken in Resource DataStore tab
**CKAN version**
2.10/master

**Describe the bug**
Missing icons and misaligned

Screenshot 2022-11-28 at 11-38-11 dadada - 25_best_tacos_austin csv - CKAN

shashigharti:
Fixes #7247

### Proposed fixes:
Fix broken timeline UI in resource view page.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datapusher/templates/datapusher/resource_data.html
+++ b/ckanext/datapusher/templates/datapusher/resource_data.html
@@ -63,25 +63,29 @@
     <h3>{{ _('Upload Log') }}</h3>
     <ul class="activity">
       {% for item in status.task_info.logs|sort(attribute='timestamp') %}
-        {% set icon = 'ok' if item.level == 'INFO' else 'exclamation' %}
+        {% set icon = 'check' if item.level == 'INFO' else 'exclamation' %}
         {% set class = ' failure' if icon == 'exclamation' else ' success' %}
         {% set popover_content = 'test' %}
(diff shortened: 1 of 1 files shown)


pdelboca:
Hello @shashigharti !

Thanks for the submission! Overall looks good, however we need to do some extra changes. 

Currently this template depends on the `activity` CSS class which is no longer in CKAN core since we removed the Activity plugin to its own extension: #6709 . So what we need to do is to remove the dependency on that CSS class.

The CSS class is used in the `ul` element of the template and if we keep it this way the display will change when the activity plugin is enabled/disabled (since it registers the `activity` CSS class).

Concrete steps (not exhaustive though):
 - Create a new css file in `datapusher` extension called `datapusher.css`
 - Get the relevant elements from the [activity css](https://github.com/ckan/ckan/blob/master/ckanext/activity/assets/activity.css) file and add them to `datapusher.css`
 - Register the new CSS file so it is loaded by CKAN.

Note: avoid calling this CSS class `activity` since it will clash with the existing one. `datapusher-activity` can be a good name for it.

Let me know if this makes sense!

shashigharti:
Hi @pdelboca, thank you! I have made the required changes as per your suggestion.

pdelboca:
Thanks @shashigharti !!

PR #7288, opened 2022-12-20
Fix for numeric return of data through datastore_search_sql

### LINKED ISSUE #5753: datastore_search_sql returns numeric as string
**CKAN version**
2.8.4

**Describe the bug**
Datastore api function datastore_search_sql returns fields with type "numeric" as string.

**Steps to reproduce**
Create a resource and upload the following CSV file, making sure it was uploaded to the datastore:

```csv
foo,bar
1,2
3,4
```

Then access it via api using datastore_search_sql, for example like this:

    GET https://beta.ckan.org/api/3/action/datastore_search_sql?sql=SELECT * from "66ca1e80-e697-40ce-9c50-2579b59e48f2"

Which returns:

```json
{
    "help": "https://beta.ckan.org/api/3/action/help_show?name=datastore_search_sql",
    "success": true,
    "result": {
        "sql": "SELECT * from \"66ca1e80-e697-40ce-9c50-2579b59e48f2\"",
        "records": [
            {
                "_id": 1,
                "_full_text": "'1':1 '2':2",
                "foo": "1",
                "bar": "2"
            },
            {
                "_id": 2,
                "_full_text": "'3':1 '4':2",
                "foo": "3",
                "bar": "4"
            }
        ],
        "fields": [
            {
                "id": "_id",
                "type": "int4"
            },
            {
                "id": "_full_text",
                "type": "tsvector"
            },
            {
                "id": "foo",
                "type": "numeric"
            },
            {
                "id": "bar",
                "type": "numeric"
            }
        ]
    }
}
```

In the response json, result.records.foo is a string.

**Expected behavior**
result.records.foo should be a number in the json response

**Additional details**
If the data is instead queried with

    GET https://beta.ckan.org/api/3/action/datastore_search?resource_id=66ca1e80-e697-40ce-9c50-2579b59e48f2

The response is:

```json
{
    "help": "https://beta.ckan.org/api/3/action/help_show?name=datastore_search",
    "success": true,
    "result": {
        "include_total": true,
        "limit": 100,
        "records_format": "objects",
        "resource_id": "66ca1e80-e697-40ce-9c50-2579b59e48f2",
        "total_estimation_threshold": null,
        "records": [
            {
                "_id": 1,
                "foo": 1,
                "bar": 2
            },
            {
                "_id": 2,
                "foo": 3,
                "bar": 4
            }
        ],
        "fields": [
            {
                "id": "_id",
                "type": "int"
            },
            {
                "id": "foo",
                "type": "numeric"
            },
            {
                "id": "bar",
                "type": "numeric"
            }
        ],
        "_links": {
            "start": "/api/3/action/datastore_search?resource_id=66ca1e80-e697-40ce-9c50-2579b59e48f2",
            "next": "/api/3/action/datastore_search?resource_id=66ca1e80-e697-40ce-9c50-2579b59e48f2&offset=100"
        },
        "total": 2,
        "total_was_estimated": false
    }
}
```

So this seems like an inconsistency to me.

Deep-NEC:
Fixes #5753

### Proposed fixes:
This fix handles correct return of numeric data as numeric data instead of by default string return.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -824,6 +824,8 @@ def convert(data: Any, type_name: str) -> Any:
     if type_name == 'nested':
         return json.loads(data[0])
     # array type
+    if type_name == 'numeric':
+        return data
     if type_name.startswith('_'):
         sub_type = type_name[1:]
         return [convert(item, sub_type) for item in data]


ckanext/datastore/backend/postgres.py
```diff
@@ -824,6 +824,8 @@ def convert(data: Any, type_name: str) -> Any:
     if type_name == 'nested':
         return json.loads(data[0])
     # array type
+    if type_name == 'numeric':
```
wardi, line 827: there are a number of integer and floating point types in postgres that we could include here

wardi:
@Deep-NEC would you mind adding a test to show that this change works as expected?

ckanext/datastore/backend/postgres.py
amercader: Another fix (suggested by @EricSoroos) is to expand the Python types check we do a few lines below to include `decimal`, which is how these are interpreted:

```python
   import decimal

   #...

    if isinstance(data, (int, float, decimal.Decimal)):
        return data
```

Just teste

Gauravp-NEC:
@wardi @amercader , This PR is overpassed by https://github.com/ckan/ckan/pull/7533.

PR #7309, opened 2023-01-05
Fixed contrast of delete button in user edit page

### LINKED ISSUE #7199: White on Yellow?? Color Contrast
**CKAN version**
https://demo.ckan.org/en/user/edit/mgifford

**Describe the bug**
Title: WCAG 1.4.3: Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds (.btn-warning)
Tags: Accessibility, WCAG 1.4.3, color-contrast

Issue: Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds (color-contrast - https://accessibilityinsights.io/info-examples/web/color-contrast)

Target application: Manage - mgifford - Users - CKAN Demo - https://demo.ckan.org/en/user/edit/mgifford

Element path: .btn-warning

Snippet: <a class="btn btn-warning" href="/en/user/generate_key/b8037a86-a216-4c9b-8211-e197fa09143a" data-module="confirm-action" data-module-content="Are you sure you want to regenerate the API key?">Regenerate API Key</a>

How to fix: 
Fix any of the following:
  Element has insufficient color contrast of 1.94 (foreground color: #ffffff, background color: #f0ad4e, font size: 10.5pt (14px), font weight: bold). Expected contrast ratio of 4.5:1

Environment: Microsoft Edge version 107.0.1418.35

====

This accessibility issue was found using Accessibility Insights for Web 2.35.0 (axe-core 4.4.1), a tool that helps find and fix accessibility issues. Get more information & download this tool at http://aka.ms/AccessibilityInsights.

shashigharti:
Fix for Bootstrap v5 and v3

Fixes #7199

### Proposed fixes:

Fix color contrast of delete button for web accessibility

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public-bs3/base/css/main-rtl.css
+++ b/ckan/public-bs3/base/css/main-rtl.css
@@ -3864,8 +3864,8 @@ a.btn.disabled, fieldset[disabled] a.btn {
 
 .btn-danger, .control-custom.disabled .checkbox.btn {
   color: #fff;
-  background-color: #d9534f;
-  border-color: #d43f3a;
+  background-color: #d43f3a;
+  border-color: #c9302c;
 }
 .btn-danger:focus, .control-custom.disabled .checkbox.btn:focus, .btn-danger.focus, .control-custom.disabled .focus.checkbox.btn {
   color: #fff;
--- a/ckan/public-bs3/base/css/main.css
+++ b/ckan/public-bs3/base/css/main.css
@@ -3864,8 +3864,8 @@ a.btn.disabled, fieldset[disabled] a.btn {
 
 .btn-danger, .control-custom.disabled .checkbox.btn {
   color: #fff;
-  background-color: #d9534f;
-  border-color: #d43f3a;
+  background-color: #d43f3a;
+  border-color: #c9302c;
 }
 .btn-danger:focus, .control-custom.disabled .checkbox.btn:focus, .btn-danger.focus, .control-custom.disabled .focus.checkbox.btn {
   color: #fff;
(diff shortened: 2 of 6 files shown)


pdelboca:
@shashigharti the end result looks good, however our main CSS file ends up with two setup for `.btn-danger, .control-custom.disabled .checkbox.btn` (lines 11131 and 3094) which seems confusing and clear indication that something is off.

I suspect that this was happening before this change, but can you take a look to see why is this happening and clean it? 

### Quick tip (in case you don't know it)

If you do:
```
$ export DEBUG=true
$ ckan sass
```
It will add the sourcemap to the CSS file allowing you to trace better wich `scss` file is generating the `css` values:
image

Looks like `_custom.scss` and `_buttons.scss` are both messing around with the same CSS classes.

pdelboca:
Also, it will be nice to have a `ckan sass --debug` flag. But this is out of scope 😛

shashigharti:
Thank you! @pdelboca . I tried fixing it [here](https://github.com/ckan/ckan/pull/7309/commits/b70afb03e04a56e481c50ffed0d147d9cd38a1d0) using the default contrast [setting](https://github.com/ckan/ckan/blob/master/ckan/public/base/scss/_bootstrap-variables.scss#L94), it fixes the issue but it has effect in many places. what do you suggest?

pdelboca:
Hello @shashigharti ! Yes, this will impact in all the buttons using the `btn-danger` class, but since the new color is slighlty different but compliant with accessibility standards I'm not worried about it.

Some minor changes and we are good to go:
- You are commiting the source map of the CSS files. Can you re-compile the files without them?
- Thanks for the debug flag! Will be super useful. 
  - Can you add a line to the `.change` file describing it?
  - Can you add a small `short_help` in the command definition? Something like: `Compile css with sourcemaps.`
  - Lastly, the `u` prefix in strings is an old Python 2 compatibility so we are trying to not add it anymore (and cleaning it when possible).

shashigharti:
@pdelboca thank you! I have made changes as per your feedback.

I also removed all u prefix from the file.

pdelboca:
Thanks @shashigharti !!

PR #7314, opened 2023-01-06
Improve docstrings of webassets_tools library.

pdelboca:
This PR improves the docstrings of our `webassets_tools.py` library.

It also adds some extra logging to help detect silent corner cases and help debugging.

### DIFF
--- a/ckan/lib/webassets_tools.py
+++ b/ckan/lib/webassets_tools.py
@@ -9,6 +9,7 @@ from typing import Any, Optional
 import yaml
 from markupsafe import Markup
 from webassets import Environment
+from webassets.env import RegisterError
 from webassets.loaders import YAMLLoader
 
 from ckan.common import config, g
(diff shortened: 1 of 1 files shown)


ckan/lib/webassets_tools.py
```diff
@@ -40,13 +49,23 @@ def create_library(name: str, path: str) -> None:
     # TODO: make PR into webassets with preferable solution
     # Issue: https://github.com/miracle2k/webassets/issues/519
     for name, bundle in bundles.items():
-        env._named_bundles.pop(name, None)
-        env.register(name, bundle)
+        try:
```
pdelboca, line 52: Note: This is silently overriding webassets if they are duplicated. Instead of silently doing it I added a warning. Testing it with 10+ extensions the only one going into this issue is `ckanext-geoview` wich is overriding due to (what i think) it is an issue in a duplicated call in subclasses: https://github.com/ckan/ckanext-geoview/blob/97036fabd85d2a6f80a2d6b3bf48414a03dcb825/ckanext/geoview/plugin/__init__.py#L176

ckan/lib/webassets_tools.py
```diff
@@ -135,12 +161,20 @@ def _to_tag(url: str, type_: str):
...
+    include the assets in the html response.
+    """
     try:
         assets = g.webassets
     except AttributeError:
+        logger.warn("Webassets are not initialized.")
         return Markup(u'')
 
     if not assets:
+        logger.warn("Webassets are initialized but none was found.")
```
pdelboca, line 177: I added a warning message since CKAN will not render properly if webassets are not initialized when this function gets call.

PR #7317, opened 2023-01-10
Clean group extra revisions before purging a group

avdata99:
Fixes https://github.com/okfn/ckanext-unhcr/issues/728

### Proposed fixes:

It's not clear to me if Group extra revisions continues working in CKAN.
It's also not clear if they are some kind of cascade deletion policy running.
In case this don't work anymore but I have old records for revision, a severe error could happen.
This PR ensures that we remove group extra revision while we purge a group.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7317.bugfix
+++ b/changes/7317.bugfix
@@ -1,0 +1,1 @@
+Ensure that we remove group extra revision while we purge a group.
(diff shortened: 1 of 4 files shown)


avdata99:
Cleaning, too old

PR #7326, opened 2023-01-12
Activity Stream for Private Datasets

### LINKED ISSUE #5772: Private datasets don't have an Activity Stream
**CKAN version**
2.9.1

**Describe the bug**
With the new detailed Activity Stream introduced in 2.9, the team may want to consider adding Activity Streams for private datasets as well.

As mentioned in #3160 this was by design, but with the new detailed Activity Stream functionality, the team may want to revisit this decision.

As Data Coordinators groom and curate datasets before publishing, they often set the dataset private, perhaps, even curating the dataset with their colleagues using the new [Dataset Collaborators](https://docs.ckan.org/en/2.9/maintaining/authorization.html#dataset-collaborators) feature.  

For traceability purposes, it'd be very useful if the Activity Stream captures this metadata curation activity before the package is made public.

**Steps to reproduce**
Set dataset to private, modify it, then check Activity Stream.

**Expected behavior**
Private datasets will have Activity Stream info, at the very least, the diff of the metadata.

pdekraker-epa:
Fixes #5772

### Proposed fixes:

- Enables recording of dataset activities when private
- Adds include_private_activity option to activity API calls to toggle visibility 
  - Access to private datasets is limited to those with update access

Privacy of each activity is determined via activity.data['package']['private'] and activity entries without that key should have same accessibility as dataset. (activity entries created via activity_create (while the dataset is private) that lack this key would be publicly visible when the dataset becomes public. This matches current behavior. An alternate implementation would track the privacy of the dataset and block out activity while private (if desired).  If all this is too complicated a private or visibility column could be added to the activity table itself.

To Do / Needs Input:
- UI updates?
  - Edit Access = private visible others only see public?
  - Should there be a UI cue that the dataset was private?
- Add / Update Tests
- Option to make private activity public? on a per dataset basis?

### Features:

- [X] includes API changes


pdekraker-epa:
Implemented via #7885

PR #7329, opened 2023-01-13
Add details about JWT secrets to docs

smotornyuk:
Describe a bit more `api_token` config options. For example, mention the fact, that `encode.secret` and `decode.secret` must be identical in the default setup

PR #7330, opened 2023-01-13
Add details about JWT secrets to docs [backport v2.10]

smotornyuk:
Backport or #7329 

Describe a bit more `api_token` config options. For example, mention the fact, that `encode.secret` and `decode.secret` must be identical in the default setup


smotornyuk:
Yes, forgot about 2.9. Here's another PR that I assigned to you: #7337

PR #7338, opened 2023-01-17
Revisit the config.get / config.get_value

### LINKED ISSUE #7282: Revisit the config.get / config.get_value approach
Config declaration introduced in #6467 is undoubtedly useful and one of the highlights of CKAN 2.10, in terms of cataloguing the configuration options available from core and extensions, and in terms of being able to rely on config options being the expected format.

But the current implementation with two config modes (`default` / `strict`) and different access methods (`config.get()` / `config.get_value()` ) is confusing and error-prone (see eg #7141,#7223, #7254, #7257) so before formalizing it in major CKAN release maybe we can re-think the approach and simplify it.

## Current behaviour (pre-CKAN 2.10)
Let's assume we have these configuration options in our ini file:

```ini
ckan.myconf1 = 12
ckan.myconf2 = one two three
ckan.myconf3 = yes
# This config option is undeclared
ckan.myconf4 = some_text 
```
And we have the following config declaration:
```yaml
- key: ckan.myconf1
  type: int
- key: ckan.myconf2
  type: list
- key: ckan.myconf3
  type: bool
```

| conf option | CKAN 2.9 config.get(x) | CKAN 2.10 config.get(x) - config.mode = default | CKAN 2.10 config.get_value(x) - config.mode = default | CKAN 2.10 config.get(x) - config.mode = strict | CKAN 2.10 config.get_value(x) - config.mode = strict |
| --- | --- | --- | --- | --- | --- |
| ckan.myconf1 | `"12"` | `"12"` | `12` | `12` | `12` |
| ckan.myconf2 | `"one two three"` | `"one two three"` | `["one", "two", "three"]` | `["one", "two", "three"]` | `["one", "two", "three"]` |
| ckan.myconf3 | `"yes"` | `"yes"` | `True` | `True` | `True` |
| ckan.myconf4 | `"some_text"` | `"some_text"` | `"some_text"` + Warning: "Option ckan.myconf4 is not declared" | `"some_text"` | `"some_text"`
| ckan.not_defined | `None` | `None` | `None` + Warning: "Option ckan.not_defined is not declared" | `KeyError` raised | `None` |

I think that `KeyError` is a bug, we shouldn't override default `.get()` behaviour like this. Aside from this the differences are clear, we chose to be conservative when using the `default` config mode and the standard `.get()` method and return the unparsed value. If you want the parsed value you need to call `.get_value()` or use `config.mode = strict`.

But `config.mode = strict` has other implications, mainly that failing to parse a declared config option will raise an exception and prevent CKAN from start:

```ini
ckan.myconf1 = not_an_int
```
| conf option | CKAN 2.9 config.get(x) | CKAN 2.10 config.get(x) - config.mode = default | CKAN 2.10 config.get_value(x) - config.mode = default | CKAN 2.10 config.get(x) - config.mode = strict | CKAN 2.10 config.get_value(x) - config.mode = strict |
| --- | --- | --- | --- | --- | --- |
| ckan.myconf1 | `"not_an_int"` | `"not_an_int"` | `"not_an_int"` | CKAN won't start with `AssertionError` | CKAN won't start with `AssertionError` |

```
AssertionError: Invalid value for ckan.myconf1: not_an_int
```
## Proposal

1. I'd remove `.get_value()`, make `.get()` return the parsed configuration value. I know this has potential to break things but assuming we clearly communicate this in the CHANGELOG it would be limited in scope:

  * This will only potentially affect configuration options that 1) are declared (ie only core ones) and 2) have a custom validation/formatting/default values applied 
  * If some extension was already parsing the configuration option with `int()` or `toolkit.aslist()` that will keep working as expected
  * There's still potential to break things if users were doing stuff like `print("Count: " + conf.get("ckan.some_int_value")`, this will work on CKAN 2.9 and fail on 2.10, but we need to assess if this risk outweighs the complexity and confusion that having both `.get()` and `.get_value()` will mean for users.

2. I'd keep `config.mode = strict` but just enforcing the config declaration, ie raising an exception if the configuration is not properly defined. I can definitely see how this is useful for site maintainers, specially as more extensions adopt config declarations
3. I'd remove the `Warning: "Option ckan.not_defined is not declared"` messages in `default` mode and only show them in `strict` mode. But I'd also add warnings if a declared option was not properly parsed in `default mode.

@smotornyuk you are the one who wrote this and the one with more experience in how the config is managed. What do you think? I probably missed something, but is what I'm proposing easy to implement (apart from changing a lot of `get_value()` calls :upside_down_face: )?

@ckan/core it would be good to get your take on this, I know we discussed this in the past but I feel we can avoid future confusion before releasing CKAN 2.10

smotornyuk:
Fixes #7282

- [x] revert the usage of config.get_value(). Config options will be parsed at startup time by default, so when calling config.get() you will get the parsed value.
- [x] There will be clear documentation in the changelog about changes
- [x] When using config.mode = strict an exception will be raised if the options don't pass validations
- [x] We'll limit the warning to:
    * If config.mode= default : config options that don't pass validation
    * If config.mode = strict : config options that are not declared (ie the current Option ckan.not_defined is not declared).

### DIFF
--- a/changes/6467.feature
+++ b/changes/6467.feature
@@ -1,1 +1,58 @@
-:ref:`declare-config-options`: declare configuration options to ensure validation and default values
+:ref:`declare-config-options`: declare configuration options to ensure validation and default values.
+
+All the native config options from CKAN are validated and converted to the
+expected type during application startup. That's how change the behavior::
+
+    debug = config.get("debug")
+
+    # CKAN <= v2.9
+    assert type(debug) is str
+    assert debug == "false" # or any value that is specified in the config file
+
+    # CKAN >= v2.10
+    assert type(debug) is bool
+    assert debug is False # or ``True``
+
+If you are using ``aslist``, ``asbool``, ``asint`` convertes from
+``ckan.plugins.toolkit``, nothing will change, because they are idempotent::
+
+    # produces the same result in v2.9 and v2.10
+    assesrt tk.asbool(config.get("debug")) is False
+    assesrt tk.asint(config.get("devserver.port")) == 5000
+    assesrt tk.aslist(config.get("ckan.plugins")) == ["stats"]
+
+If you are using custom logic, the code requires a review. For example, the
+following will produce an ``AttributeError``, because ``ckan.plugins`` is
+converted into a list during application's startup::
+
+    # AttributeError
+    plugins = config.get("ckan.plugins").split()
+
+Depending on desired backward compatibility, one of the following expressions
+can be used instead::
+
+    # if both v2.9 and v2.10 are supported
+    plugins = tk.aslist(config.get("ckan.plugins"))
+
+    # if only v2.10 is supported
+    plugins = config.get("ckan.plugins")
+
+The second major change affects default values. Starting from CKAN v2.10,
+majority of the config options has a declared default value. It means that
+whenever you invoke ``config.get`` method, the **declared default** value is
+returned instead of ``None``. Example::
+
+    # CKAN v2.9
+    assert config.get("search.facets.limit") is None
+
+    # CKAN v2.10
+    assert config.get("search.facets.limit") == 10
+
+The second argument to ``config.get`` is used only when you are trying to get
+the value of the missing **undeclared** option::
+
+    assert config.get("not.declared.and.missing.from.config", 1) == 1
+
+
+The above is the same for any extension that **declares** its config options
+using ``IConfigDeclaration`` interface or ``config_declarations`` blanket.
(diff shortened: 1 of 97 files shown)


amercader (approved):
Thanks @smotornyuk for the quick PR!
I've just added some minor corrections to make the docs and messages clearer.

(In the future it would help to split the changes in behaviour, docs and function rewrites into three separate commits)

smotornyuk:
Thank you for your suggestions. I squashed them into the last commit, just to avoid a mess during backport to 2.10

PR #7341, opened 2023-01-18
`datastore_records_delete` action

JVickery-TBS:
Added `datastore_records_delete` action, which will never delete the datastore table.

### Features:

- [ ] includes API changes
- [X] includes feature for possible backport 


ckanext/datastore/logic/action.py
```diff
@@ -431,6 +431,36 @@ def datastore_delete(context: Context, data_dict: dict[str, Any]):
     return result
 
 
+def datastore_records_delete(context: Context, data_dict: dict[str, Any]):
+    '''Deletes all records or a set of records from the DataStore.
```
wardi, line 435:
```suggestion
    '''Deletes records from a DataStore table but will never remove the table itself.
```

JVickery-TBS: Updated.

ckanext/datastore/logic/action.py
```diff
@@ -431,6 +431,36 @@ def datastore_delete(context: Context, data_dict: dict[str, Any]):
...
+
+    :param resource_id: resource id that the data will be deleted from.
+                        (optional)
+    :type resource_id: string
+    :param force: set to True to edit a read-only resource
+    :type force: bool (optional, default: False)
+    :param filters: filters to apply before deleting (eg {"name": "fred"}).
+                   If missing delete whole table and all dependent views.
```
wardi, line 443:
```suggestion
                   If missing, null or {} delete all records.
```

ckanext/datastore/logic/action.py
```diff
@@ -431,6 +431,36 @@ def datastore_delete(context: Context, data_dict: dict[str, Any]):
...
+    :type calculate_record_count: bool (optional, default: False)
+
+    **Results:**
+
+    :returns: Original filters sent.
+    :rtype: dictionary
+
+    '''
+    filters = data_dict.pop('filters', {})
+    data_dict['filters'] = filters
```
wardi, line 460:
```suggestion
    filters = data_dict.get('filters', None)
    # passing filters=None will cause table to be deleted, use {} to remove records instead
    data_dict['filters'] = {} if filters is None else filters
```

ckanext/datastore/logic/action.py
JVickery-TBS: Updated.

ckanext/datastore/logic/action.py
```diff
@@ -456,9 +456,9 @@ def datastore_records_delete(context: Context, data_dict: dict[str, Any]):
     :rtype: dictionary
 
     '''
+    filters = data_dict.pop('filters', None)
```
wardi, line 459: This is super minor, but I used `get` in my suggested change to save an unnecessary modification to the dict here.

ckanext/datastore/logic/action.py
JVickery-TBS: Updated!

This is in the case that a user calls the action and passes `filters: None`? In such a case, the `data_dict.pop` would not use the default value of an empty dict, because it would exists as `None` already?

ckanext/datastore/logic/action.py
JVickery-TBS: A right, sorry, makes sense as we are really not doing anything else with the data_dict in this method. Okay should be updated now!

ckanext/datastore/logic/action.py
wardi: that's right

```
>>> help(dict.get)
Help on method_descriptor:

get(...)
    D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.
```

JVickery-TBS:
The new test is dependent on the bugfix here: https://github.com/ckan/ckan/pull/7345

JVickery-TBS:
Also recommend merging along with https://github.com/ckan/ckan/pull/7346 as that solves issues with the datastore `json_validator` 500 error on null.

ckanext/datastore/logic/schema.py
```diff
@@ -169,6 +169,20 @@ def datastore_delete_schema() -> Schema:
     return schema
 
 
+def datastore_records_delete_schema() -> Schema:
+    schema = {
+        'resource_id': [not_missing, not_empty, unicode_safe],
+        'force': [ignore_missing, boolean_validator],
+        'filters': [not_missing, json_validator],
```
wardi, line 176: filters must be `{str:str or list of str, ...}` dictionaries, not JSON. We could use a new validator for this.

The filters parameter isn't properly validated or covered in the datastore documentation either. Would be really nice to have a section on filters for the endpoints that support them, kind of like we have for fields: https://docs.ckan.org/en/2.9/maintaining/datastore.html?highlight=datastore#fields

wardi:
We could also mention this endpoint in the docs for `datastore_delete`

ckanext/datastore/logic/schema.py
JVickery-TBS: @wardi Okay we are now using `dict_only`; I also added `dict_only` and `ignore_missing` to the `datastore_delete` schema.

wardi: `datastore_delete` also accepts `null` to its filters value, so you might not want to change its validation (or at least we should discuss that change on a separate issue)

JVickery-TBS:
Because we are no longer using the `json_validator` method, there is no recommendation to merge along with https://github.com/ckan/ckan/pull/7346 anymore.

JVickery-TBS:
@wardi I have now added some new documentation.

There is now a reference to `datastore_records_delete` in the `datastore_delete` docs.

And there is a new `Filters` section in the datastore docs. And `datastore_search`, `datastore_delete`, and `datastore_records_delete` reference the new section.

Let me know if you want to change the wording in the new `Filters` section at all.

doc/maintaining/datastore.rst
```diff
@@ -411,6 +411,36 @@ You can find more information about the formatting of dates in the `date/time ty
...
+-------
+
+Filters define the matching conditions to select from the DataStore. A filter is defined as follows::
+
+    {
+        "resource_id":  # the resource ID (required)
+        "filters": {
+            # column name: # field value
+            ...:  # other user-defined filters
+	}
```
wardi, line 426: is this not aligned because of mixing tabs and spaces?

doc/maintaining/datastore.rst
```diff
@@ -411,6 +411,36 @@ You can find more information about the formatting of dates in the `date/time ty
...
+Example::
+
+    {
+        "resource_id":  "5f38da22-7d55-4312-81ce-17f1a9e84788",
+        "filters": {
+            "name": "Fred",
+            "dob":  "1994-7-07"
+	}
+    }
+
```
wardi, line 443: Thank you for adding this section. Filters also support passing a list of strings as values that are treated as "column must match one of these values", could you add a simple example for that as well?

JVickery-TBS: @wardi new wording and example for Lists has been added.

wardi:
thanks. @JVickery-TBS would you investigate the test failure

JVickery-TBS:
@wardi the test is failing on `assert resource.extras.get('datastore_active') is True` because this is dependant on this PR: https://github.com/ckan/ckan/pull/7345 (which should get backported)

wardi:
#7345 has been merged, if you merge from master to this PR we should see the test fixed here too

JVickery-TBS:
@wardi Should I merge master into this branch, or just cherry-pick the one commit?

wardi:
I would merge, pulling in unrelated changes on master shouldn't be a concern.

JVickery-TBS:
@wardi okay tests are passing now.

PR #7344, opened 2023-01-19
Fix api_token_list parameter

amercader:
The documentation for `api_token_list` (which shows all API tokens for a particular user) didn't show the input parameter name.

This turned out to be `user`, which is not great as it can mean many things in CKAN. I updated it to be `user_id`, in line with other similar actions, and documented this one, but kept support for `user` for backwards compatibility.

### Features:

- [x] includes API changes


kowh-ai (approved):
Tested change - looks good
user_id_error

PR #7353, opened 2023-01-24
[#7352] Provide a fallback value for WTF_CSRF_SECRET_KEY

### LINKED ISSUE #7352: No default for WTF_CSRF_SECRET_KEY config option
## CKAN version
2.10 

## Describe the bug
The `WTF_CSRF_SECRET_KEY` config option, used to generate the CSRF tokens is defined as such in the declaration:

https://github.com/ckan/ckan/blob/9e60350f27c9266b5afbd741484265c5e06a2d38/ckan/config/config_declaration.yaml#L645-L647

This means that when a new ini file is generated, a secret value is added to the ini file by default, but when that config option is not defined in an ini file (ie you are using an ini file generated with a previous version of CKAN when upgrading your instance) the default value is `None`.

This causes the app to fail on all requests with:

```
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/templates/base.html", line 27, in block 'meta'
    <meta name="{{ g.csrf_field_name }}" content="{{ csrf_token() }}" />
  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask_wtf/csrf.py", line 36, in generate_csrf
    secret_key = _get_config(
  File "/home/adria/dev/pyenvs/ckan-py3/lib/python3.8/site-packages/flask_wtf/csrf.py", line 136, in _get_config
    raise RuntimeError(message)
RuntimeError: A secret key is required to use CSRF.
```
That's because wtforms falls back to Flask's `SECRET_KEY` only if `WTF_CSRF_SECRET_KEY` is not present in the config, but we add it to the config, just empty.

This setting should have either be checked at startup to see if it's set or fallback to a default like we do with the rest of the secret keys (I prefer the latter)

amercader:
Fixes #7352

If `config["WTF_CSRF_SECRET_KEY"]` is not set (or present and empty), we set it to `SECRET_KEY` (which falls back to `beaker.session.secret`)

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -262,7 +262,10 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     _register_error_handler(app)
 
     # CSRF
-    app.config['WTF_CSRF_FIELD_NAME'] = "_csrf_token"
+    wtf_key = "WTF_CSRF_SECRET_KEY"
+    if not app.config.get(wtf_key):
+        config[wtf_key] = app.config[wtf_key] = app.config["SECRET_KEY"]
+    app.config["WTF_CSRF_FIELD_NAME"] = "_csrf_token"
     csrf.init_app(app)
 
     # Set up each IBlueprint extension as a Flask Blueprint
--- a/ckan/tests/config/test_middleware.py
+++ b/ckan/tests/config/test_middleware.py
@@ -100,3 +100,13 @@ def test_flask_config_values_are_parsed(app):
     assert (
         app.flask_app.config["REMEMBER_COOKIE_DURATION"] == 12345
     )
+
+
+@pytest.mark.ckan_config("WTF_CSRF_SECRET_KEY", None)
+def test_no_wtf_secret_falls_back_to_secret_key(make_app):
+    assert (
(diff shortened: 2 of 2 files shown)


pdelboca:
@amercader I'm okay with the approach. However I would add a warning notifying that we are fallbacking to the Secret Key. This silent assignments tend to be difficult to debug.

pdelboca:
@amercader it looks like the same `flask-wtforms` fallbacks silently to SECRET_KEY, so why do we need to replicate the code? 🤔 

https://github.com/wtforms/flask-wtf/blob/1cf8b19bcbc53e5357a4aa0c5ccb5cf7d3ea0d22/src/flask_wtf/csrf.py#L23-L41

amercader:
@pdelboca flask-wtf will fallback to `SECRET_KEY` if `WTF_CSRF_SECRET_KEY` is *not present* in the `app.config`, but not if it's there but empty:

https://github.com/wtforms/flask-wtf/blob/1cf8b19bcbc53e5357a4aa0c5ccb5cf7d3ea0d22/src/flask_wtf/csrf.py#L132-L133

That's why we need to ensure that it has a value. I don't think is necessary to add a warning, but we can revisit this when we consolidate all secrets management

PR #7354, opened 2023-01-24
Set WTF_CSRF_FIELD_NAME to _csrf_token

pdelboca:
Fixes 400 Bad Request

Version: `master`, `dev-v2.10`.

`WTF_CSRF_FIELD_NAME` is set as an option in the config file with a default of `csrf_token`.  However, all around the code it has been called `_csrf_token` so this may cause errors if the user tries to set a different name.

### Proposed fixes:

For purposes of simplicity we will enforce a name everywhere (`_csrf_token`) and ignore this WTForms config option.

### Note

While developing this fix I tried to rename this field to `csrf_token` since it is a more common convention. However it will mess around with CKAN's cache if a session data do not start with `_`. (See added docstring in `_is_valid_session_cookie_data` method)


ckan/config/middleware/flask_app.py
```diff
@@ -262,7 +262,7 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     _register_error_handler(app)
 
     # CSRF
-    app.config['WTF_CSRF_FIELD_NAME'] = "_csrf_token"
```
amercader, line 265: This config declaration has a different default:

https://github.com/ckan/ckan/blob/978875e41643b53fc2c714b2264f32e502db3f8d/ckan/config/config_declaration.yaml#L664-L666

That is probably fine.

But the old name is hardcoded in some JS:

The [meta tag](https://github.com/ckan/ckan/blob/978875e41643b53fc2c714b2264f32e502db3f8d/ckan/templates/base.html#L27) with the token uses the config name, but the code that reads the value [in the js client](https://github.com/ckan/ckan/blob/978875e41643b53fc2c714b2264f32e502db3f8d/ckan/public/base/javascript/client.js#L54) has the name hardcoded.

There's also [this module](https://github.com/ckan/ckan/blob/978875e41643b53fc2c714b2264f32e502db3f8d/ckan/public/base/javascript/modules/confirm-action.js#L106) that uses the old name.

Maybe we can simplify things and enforce a name everywhere (ie ignore this WTForms config option) or alternatively the JS needs to be aware of the field name (proabably adding a separate `csrf_field_name` meta tag that they can read first before getting the actual token value)

pdelboca:
I'm following this in #7365

PR #7365, opened 2023-01-27
Support configurable CSRF field name in the frontend

pdelboca:
Fixes 400 Bad Request when users add custom `WTF_CSRF_FIELD_NAME` config. (See also https://github.com/ckan/ckan-docker-base/issues/7)

### Version: 
master, dev-v2.10.

### Notes

Currently we allow to change the field name for the CSRF token through a config option. However there are also places in the code where we are hardcoding it to `_csrf_token`. This PR fixes the frontend logic to get the `CSRF_FIELD_NAME` from a new `meta` tag.

### Note
This is a follow up of #7354 since for a cleaner git history. While developing this I found several issues:
 - Allowing the user to change the field name requires extra logic in the javascript modules to detect first the name of the field and then get the value.
 - As `csrf_token` is added to the session cookie, users will be force to introduce a field name starting with `_` to not break our cache logic. I forced to skip this config in `_is_valid_session_cookie_data()` but I'm not convinced by this solution.
 - Removing it from the `config_declaration.yaml` will cause an undeclared warning and a test to fail. For the current logic of config declarations, we need to add all internal configs of our Flask extensions to be able to run in `strict` mode without warnings: 
 ```python
     def test_ignore_declared(self, command):
        result = command("undeclared", "-idatapusher", "-idatastore")
>       assert not result.output
E       AssertionError: assert not 'WTF_CSRF_FIELD_NAME\n'
E        +  where 'WTF_CSRF_FIELD_NAME\n' = <Result okay>.output

 ```

### DIFF
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -138,6 +138,8 @@ def _allow_caching(cache_force: Optional[bool] = None):
 def _is_valid_session_cookie_data() -> bool:
     is_valid_cookie_data = False
     for key, value in session.items():
+        if key == "csrf_token":
+            continue
         if not key.startswith(u'_') and value:
             is_valid_cookie_data = True
             break
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -138,7 +138,7 @@ def _allow_caching(cache_force: Optional[bool] = None):
 def _is_valid_session_cookie_data() -> bool:
     is_valid_cookie_data = False
     for key, value in session.items():
-        if key == "csrf_token":
+        if key == config.get("WTF_CSRF_FIELD_NAME"):
             continue
         if not key.startswith(u'_') and value:
             is_valid_cookie_data = True
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -661,6 +661,10 @@ groups:
 
         description: HTTP methods to protect from CSRF.
 
+      - key: WTF_CSRF_FIELD_NAME
+        default: _csrf_token
+        description: Name of the form field and session key that holds the CSRF token.
+
       - key: WTF_CSRF_HEADERS
         type: list
         default:
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -661,10 +661,6 @@ groups:
 
         description: HTTP methods to protect from CSRF.
 
-      - key: WTF_CSRF_FIELD_NAME
-        default: csrf_token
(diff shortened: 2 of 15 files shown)


ckan/lib/base.py
```diff
@@ -138,6 +138,8 @@ def _allow_caching(cache_force: Optional[bool] = None):
 def _is_valid_session_cookie_data() -> bool:
```
pdelboca, line 138: I had a hard time understanding why is this function here and why are we by-passing data elements starting with `_` it would be nice to document.

@Zharktas do you remember why is this working like this? I would like to add a docstring to the function.

wardi: IIRC this logic was to prevent our local directory from filling up with cookie data that is discarded by end users (if all cookie fields are automatically added, don't bother storing them)

Does this change mean that an unauthenticated curl request for a ckan page creates a new file with cookie data on the server? Maybe that's unavoidable for pages with forms now? Is the fix to make sure that old cookie data is cleaned up regularly?

amercader: The issue with storing too many session files (https://github.com/ckan/ckan/issues/6948) was not directly related to this code but maybe the original reasoning was similar. The bottom line of the session files saga (#6954 first and #7014 later) was that yes, with the CSRF protection enabled there is always a session file (or redis/db record depending on your config) created to store the `_csrf_token`  token (or whatever is called) but is only one per-user / browser session (not per-request as the original issue reported)

The `_is_valid_session_cookie_data()` check was introduced in https://github.com/ckan/ckan/pull/4781 ([here](https://github.com/ckan/ckan/pull/4781/files#diff-68d59b70c2143d203a1aec158f5eecc93c2f6912357ff1f5598baba1067d3595R264)). I guess the reasoning was, "if there is session data added by the user, don't allow caching", and "by added by the user" what was meant was "not starting with underscore", probably because Beaker always adds stuff like `_last_accessed` to the session.

So I don't think this change will affect the current logic to store session files (they will be stored) but we need to think about how it affects caching. Maybe I'm missing something but does having the csrf token in the session (and added as a meta tag) mean that we can't enable caching at all?

There are some [Django docs](https://docs.djangoproject.com/en/4.1/howto/csrf/#using-csrf-protection-with-caching) that might help

wardi: the issue I mention wasn't created by this PR, so let's investigate it separately

kowh-ai:
> @wardi I have just backported this PR to `dev-v2.10`. I'm removing the tag.

@pdelboca - thanks! works a charm in the new ckan-2.10.0 docker image

Issue #7372, opened 2023-02-01
`build_nav_icon` Helper on Resource Edit and Views


JVickery-TBS:
## CKAN version
master branch (latest)
AND 2.9.7

## Describe the bug
The `build_nav_icon` helper does not set the active class for the urls of `<package_type>/<id>/resource/<resource_id>/edit` and `<package_type>/<id>/resource/<resource_id>/views`

### Steps to reproduce
* Navigate to a url such as `/dataset/<id>/resource/<resource_id>/edit`

### Expected behavior
The "Edit Resource" and "Views" tab/nav_icon links should be visually active with the active class.

### Additional details
![Edit Resource Tab](https://user-images.githubusercontent.com/97247789/216089875-0df2ab6a-fcc6-4ed4-b644-e0aa419af6f6.png)

JVickery-TBS:
Looks like the helper calls `blueprint, endpoint = p.toolkit.get_endpoint()` which returns `resource, edit`. But then the matching controller and action is `dataset_resource, edit`.

So the `get_endpoint` function (which just uses the `ckan.common.request,endpoint`), is resolving the request endpoint to the non-prefixed resource routes, instead of the prefixed resource routes. E.g. it is resolving to `resource.edit` and not `<package_type>_resource.edit`.

Same thing with the views.

JVickery-TBS:
Created PR with suggested changes: https://github.com/ckan/ckan/pull/7373

JVickery-TBS:
@smotornyuk if that other PR suggestion is too big, and the `resource` blueprint needs to stay in: I have made this PR with a small addition to the helper functions https://github.com/ckan/ckan/pull/7379

It just adds `highlight_blueprint_endpoints` option much like the `highlight_controllers` option.

smotornyuk:
I like #7373, but removing blueprint may break extensions, so let's change it a bit. I'll add a couple of comments there

PR #7373, opened 2023-02-01
Removed `resource` Blueprint; Using Only `<pkg_type>_resource` Blueprint;

JVickery-TBS:
# Fixes
Removed the `resource` blueprint, referred to as `prefixed_resource` in the code.
See https://github.com/ckan/ckan/issues/7372 for more details.

### Proposed fixes:
The helper function `build_nav_icon` fails to resolve active routes for `<package_type>_resource` blueprints. Unless if there is a reason to keep the `resource` blueprint, I propose to just drop it, and have only the `<package_type>_resource` blueprints

### Features:

- [ ] includes API changes


smotornyuk:
Everything is ok, but it's better not to remove the `resource` blueprint. It will exist just in case someone still uses `resource.xxx` routes. But it would be nice to update the changelog record and mention that `dataset_resource.` blueprint is a preferred name and plain `resource.` will be eventually removed

JVickery-TBS:
@smotornyuk Okay the other `resource` blueprint has been brought back in. And the change log has been updated regarding the future removal of it, and the preferred blueprint to use.

smotornyuk:
@JVickery-TBS, if you solve the `_get_package_type` issue that I mentioned earlier(do not import it from `views` into `lib`), I can merge this PR

JVickery-TBS:
@smotornyuk okay I have removed the import from the views module in lib now. I just copied some of the code from the `_get_package_type` so that the dictization has the correct url for the package type. Or is it fine to just force `dataset_resource.read` here?

I merged in master to the branch to fix some merge conflicts as well.

smotornyuk:
@JVickery-TBS, your implementation is better. Even though it won't make much difference if we use `dataset_resource.read` here, some extensions may rely on the actual package type. So it's better to use the real type instead of the `dataset` whenever it's possible

PR #7374, opened 2023-02-01
Do not auto-register prefixed resource blueprint;

JVickery-TBS:
# Fixes
The `prefixed_resource` blueprint by the name of `resource` needs auto register `False` to allow for proper blueprint overrides.

### Proposed fixes:
Set `prefixed_resource.auto_register = False`

### Features:

- [ ] includes API changes


JVickery-TBS:
@wardi As discussed, `prefixed_resource` will not be auto-loaded with the core blueprints. It will NOW get loaded at the very end of the `register_package_blueprints` method.

This should allow for the `<package_type>_resource` blueprints to take priority over the `resource` blueprints. This should solve issues with overriding the `<package_type>_resource` blueprints.

ckan/tests/lib/test_helpers.py
```diff
@@ -635,7 +635,7 @@ def test_active_in_resource_controller(self, test_request_context):
         with test_request_context(u'/dataset/' + resource['package_id'] + '/resource/' + resource['id']):
             menu = (
                 ("home.index", "Home"),
-                ("dataset.search", "Datasets", ['dataset', 'resource']),
+                ("dataset.search", "Datasets", ['dataset', 'resource', 'dataset_resource']),
```
wardi, line 638: what's this change for?

JVickery-TBS: @wardi because it has `with test_request_context(u'/dataset/' + resource['package_id'] + '/resource/' + resource['id'])`, with the latest changes, it is going to load the `dataset_resource` blueprint as a priority. So the `dataset_resource` addition is to just highlight that blueprint if it is there, giving it the active class to match in the test.

PR #7391, opened 2023-02-15
Fix package names in docs

amercader:
This has already been pushed to the `2.9` and `2.10` branches (i.e. the live docs) and the `dev-*`  ones, so no need to backport


kowh-ai (approved):
All good. Built HTML files using Sphinx, no errors

PR #7401, opened 2023-02-20
Update Transifex config and documentation

amercader:
Transifex has a new CLI tool and the configuration and commands to setup are slightly different.


kowh-ai (approved):
Built fine, looks good

PR #7417, opened 2023-02-23
Add CSRF token to pkg groups form

pdelboca:
### Proposed fixes:

This PR adds CSRF token to the Remove form of package groups.

### Current Behaviour:
- Clicking on Remove button will return a 400 Bad Request 
image


tino097 (approved):

tino097:
Thanks @pdelboca

PR #7425, opened 2023-03-04
Add suffix to dataset id when creating another dataset with the same name.

### LINKED ISSUE #4637: Adding suffix to package name if exists another with same name

Gauravp-NEC:
Fixes #4637 

### Proposed fixes:
If you want to create a dataset with a name that already exists an error is always raised.
This PR will allow CKAN user to create dataset with the same dataset name in the exiting organisation or in another organisation.

**Example:**
1st attempt to create dataset name "car" => dataset created with name "car" => dataset id created with name "car"
2nd attempt to create dataset name "car" =>dataset created with name "car" => dataset id created with name "car1"
3rd attempt to create dataset name "car" =>dataset created with name "car" => dataset id created with name "car2"
and so on.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/4637.bugfix
+++ b/changes/4637.bugfix
@@ -1,0 +1,1 @@
+Add suffix to dataset id when creating another dataset with the same name.
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -366,7 +366,12 @@ def package_name_validator(key: FlattenKey, data: FlattenDataDict,
         query = query.filter(model.Package.id != package_id)
 
     if session.query(query.exists()).scalar():
-        errors[key].append(_('That URL is already in use.'))
+        if any(chr.isdigit() for chr in data[key]):
+            data[key] = re.sub(r'[0-9]+$',
+             lambda x: f"{str(int(x.group())+1).zfill(len(x.group()))}",
+             data[key])        
+        else:        
+            data[key] = data[key] + str(1)
 
     value = data[key]
     if len(value) < PACKAGE_NAME_MIN_LENGTH:


smotornyuk:
It's a bad idea. Remember python's zen:
> Explicit is better than implicit.

Imagine that you are making the request to the `package_create` with the name `xxx`, but CKAN just silently creates `xxx1` dataset instead. You are not aware of this and just continue using `xxx` in your script, working with a completely different dataset.

If your project needs such behavior, just override validators of the package schema in your extension. Setting this behavior as a default for everyone is a breaking change, so you need more than just "I don't like errors".

But outside of this implementation, the idea of having a "safe" way to generate the package name is actually nice. If you create a helpers\action\tooktit member that accepts the name of the entity and returns you a free version of this name, it may be useful. I.e, something like this:

```python
name = "xxx"
# returns "xxx" if this name is not yet occupied
# returns "xxx-1" if this name is already used by the package 
safe_name = tk.safe_name_for("package", name) 
```

Another option, you can modify JS which generates the name from the title. If you pay attention to the original issue, you'll see the following 
> This is a common issue if several datasets has the same title (probably from different orgs with same information) because **the name is calculated using title**.

The only place, where "the name is calculated using title" is the dataset creation form. You can try updating the corresponding JS so that it checks whether the name is already taken and provides other versions of the name. In this way, the user has a chance to check the modified name before it's written to the database.

smotornyuk:
@Gauravp-NEC , I'm closing this PR because the solution is not appropriate and too disruptive. But I created another issue, #7481, with some ideas. You can work on a different solution inside a separate extension. If it proves itself useful, eventually we can get it merged into CKAN core.

PR #7449, opened 2023-03-15
Fix missing resource URL on update resource with uploaded file

mutantsan:
When you have a resource with an uploaded file, try to **update** it (e.g update the description) and save it. It must break the `url`, because we are not sending it to the server if it's not an upload-type resource. So the URL will be still uploadable, but you are going to miss the filename and the preview won't work. 

### Features:

- [ ] includes API changes


tino097 (approved):
LGTM

tino097:
Thanks @mutantsan !

PR #7450, opened 2023-03-15
[7409]Secrets consolidation

### LINKED ISSUE #7409: Secrets consolidation
These are to my knowledge the secrets related config options needed to run CKAN 2.10 (please mention any I missed below):
| config name | `ckan generate config` default value | fallback value |
| --- | --- | --- |
| `beaker.session.secret` | Random token | None |
| Flask's `SECRET_KEY` | Not added | `beaker.session.secret` |
| `WTF_FORMS_CSRF_SECRET` | Random token | `SECRET_KEY` |
| `api_token.jwt.encode.secret` | `string:%(beaker.session.secret)s` | `beaker.session.secret` |
| `api_token.jwt.decode.secret` | `string:%(beaker.session.secret)s` | `beaker.session.secret` |

So for historic reasons, `beaker.session.secret` ends up being the default fallback for all of these:

```mermaid
graph TD;
    api_token.jwt.encode.secret-->beaker.session.secret;
    api_token.jwt.decode.secret-->beaker.session.secret;
    SECRET_KEY-->beaker.session.secret;
    WTF_FORMS_CSRF_SECRET-->SECRET_KEY;
```
More things to note:

* `beaker.session.secret` is the one added to the ini file and the one all options fall back to, but if you don't set it on the ini file, the error you get is:
```
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/config/middleware/flask_app.py", line 176, in make_flask_stack                      
    raise RuntimeError(u'You must provide a value for the secret key'                                                                 
RuntimeError: You must provide a value for the secret key with the SECRET_KEY config option
```
* You can use `SECRET_KEY` in your ini file instead, but if you don't define `beaker.session.secret`, API calls will fail:
```
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/lib/api_token.py", line 84, in decode
    _get_secret(encode=False),                                       
  File "/home/adria/dev/pyenvs/ckan-py3/src/ckan/ckan/lib/api_token.py", line 39, in _get_secret
    secret = u"string:" + config.get(_config_secret_fallback)

```

Looks like there is space to consolidate how these settings are managed and improve these inconsistencies.
My feeling is that we should try to fall back to either `SECRET_KEY` or a new `ckan.secret_key` and consolidate how all keys are set in the ini file and how their defaults are handled. And think of ways of clearly documenting them.

TomeCirun:
Fixes #7409 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -86,8 +86,6 @@ groups:
         internal: true
       - key: PROPAGATE_EXCEPTIONS
         internal: true
-      - key: SECRET_KEY
-        internal: true
       - key: SEND_FILE_MAX_AGE_DEFAULT
         internal: true
       - key: SERVER_NAME
@@ -161,6 +159,13 @@ groups:
         placeholder: egg:ckan
         validators: not_empty
         required: true
+
+      - key: SECRET_KEY
+        validators: not_empty
+        required: true
+        placeholder_callable: secrets:token_urlsafe
+        callable_args:
+          nbytes: 20
 
       - key: ckan.legacy_route_mappings
         default: {}
@@ -258,11 +263,7 @@ groups:
         default: ckan
         description: Name of the cookie key used to save the session under.
       - key: beaker.session.secret
-        validators: not_empty
-        required: true
-        placeholder_callable: secrets:token_urlsafe
-        callable_args:
-          nbytes: 20
+        placeholder: string:%(SECRET_KEY)s
         description: |
           This is the secret token that the beaker library uses to hash the
           cookie sent to the client. `ckan generate config` generates a unique
@@ -649,7 +650,7 @@ groups:
 
       - key: WTF_CSRF_SECRET_KEY
         description: Random data for generating secure tokens.
-        placeholder_callable: secrets:token_urlsafe
+        placeholder: string:%(SECRET_KEY)s
 
       - key: WTF_CSRF_METHODS
         type: list
@@ -755,7 +756,7 @@ groups:
         description: Number of bytes used to generate unique id for API Token.
 
       - key: api_token.jwt.encode.secret
-        placeholder: string:%(beaker.session.secret)s
+        placeholder: string:%(SECRET_KEY)s
         example: file:/path/to/private/key
         description: |
           A key suitable for the chosen algorithm(``api_token.jwt.algorithm``):
@@ -773,7 +774,7 @@ groups:
           * ``file:`` - Path to file. Content of the file will be used as key.
 
       - key: api_token.jwt.decode.secret
-        placeholder: string:%(beaker.session.secret)s
+        placeholder: string:%(SECRET_KEY)s
         example: file:/path/to/public/key.pub
         description: |
           A key suitable for the chosen algorithm(``api_token.jwt.algorithm``):
--- a/ckan/lib/api_token.py
+++ b/ckan/lib/api_token.py
@@ -19,7 +19,7 @@ log = logging.getLogger(__name__)
 
 _config_encode_secret = u"api_token.jwt.encode.secret"
 _config_decode_secret = u"api_token.jwt.decode.secret"
-_config_secret_fallback = u"beaker.session.secret"
+_config_secret_fallback = u"SECRET_KEY"
 
 _config_algorithm = u"api_token.jwt.algorithm"
 
--- a/test-core.ini
+++ b/test-core.ini
@@ -20,6 +20,7 @@ cache_dir = %(here)s/data
 debug = false
 testing = true
 
+SECRET_KEY = This_is_a_secret_or_is_it
 
 # Specify the Postgres database for SQLAlchemy to use
 sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_test
@@ -88,8 +89,8 @@ ckan.activity_streams_email_notifications = True
 ckan.tracking_enabled = true
 
 beaker.session.key = ckan
-beaker.session.secret = This_is_a_secret_or_is_it
-WTF_CSRF_SECRET_KEY = %(beaker.session.secret)s
+beaker.session.secret = %(SECRET_KEY)s
+WTF_CSRF_SECRET_KEY = %(SECRET_KEY)s
 
 ## background jobs
 ckan.jobs.timeout = 180


smotornyuk:
How about adding using [configured_default](https://github.com/ckan/ckan/blob/master/ckan/lib/navl/validators.py#L200) in addition to the placeholder?

`placeholder` is good for newly generated config files, but there are two cases when it's not convenient:

1. You already have an existing config file without options with a placeholder. In this case, you must add missing options to your config file. Looks like we have no new secrets at the moment, so it's not a problem right now. But someone who is upgrading from v2.9 to v2.11 have to add `WTF_CSRF_SECRET_KEY`. It's a trivial task - just run `ckan config declaration --core | grep WTF_CSRF_SECRET` and copy the line into config, but it can be avoided
2. You want to keep your config minimal, specifying as few options as possible and relying on default values. Probably, you'll use `ckan config declaration --core --minimal` in order to get only these options, which have a `required` flag. But it means that all the options that are using placeholder won't have any value, because they are missing from the minimal config example.

If, in addition to the placeholder, you'll add:
```yaml
validators: configured_default("SECRET_KEY",None) not_empty 
```

all existing config files, without these options will get the value of the `SECRET_KEY` because of validators. All new config files will use `SECRET_KEY` because of explicit placeholder

amercader:
@smotornyuk Maybe I'm not understanding `configured_default` correctly but can we not end up with a situation where `config["SECRET_KEY"]` is None? isn't this we want to avoid? 

My understanding from the discussion in #7409 was that:

1. When creating a new ini file with `ckan generate config` we add a default value (with `placeholder_callable: secrets:token_urlsafe`. 
2. If there is no `SECRET_KEY` defined in the ini file, ckan should fail to start (ie where are not setting a default value under the hood with `default_callable: secrets:token_urlsafe`)

I'm personally leaning towards enforcing having a `SECRET_KEY` value defined in the ini file going forward (that all other secrets can fallback to) but happy to discuss further.

There's one thing we should improve in point 2. Right now the presence of `SECRET_KEY` is checked when building the Flask stack [here](https://github.com/ckan/ckan/blob/4588c88e1b5c0547435ada4a282eaaa3b5dd7b71/ckan/config/middleware/flask_app.py#L143), but there is stuff that happens earlier that might require a valid `SECRET_KEY` in place much earlier, (like here where I'm trying to create a [persitent tmp folder](https://github.com/ckan/ckan/pull/7585/files#diff-7110119dd36330c03fc2a664739c2098a0421c1a1d98efc204cdac87df3a54b5R71), which [fails](https://app.circleci.com/pipelines/github/ckan/ckan/5502/workflows/ff744a90-818c-461f-94ae-bdda8d36a84b/jobs/13508) if `SECRET_KEY` is not set). Probably just after normalizing the config [here](https://github.com/ckan/ckan/blob/4588c88e1b5c0547435ada4a282eaaa3b5dd7b71/ckan/config/environment.py#L122).

Additionally I'm not sure why even if there is a `required: true` and `not_empty` validator the config normalization / validation will not fail and the code will keep running until the `flask_app.py` check.

Happy to go over this in the dev meeting.

smotornyuk:
I want to use `configured_default` on values other than `SECRET_KEY`. For example, if `WTF_CSRF_SECRET_KEY` has `configured_default` set to the main `SECRET_KEY`, we can remove `WTF_CSRF..` from the config file and it will use the value of `SECRET_KEY`. 
Of course, `SECRET_KEY` should not use any default value, and using only `non_empty` for it is perfect.

As for running silently till the `flask_app` without failing - are you using `config.mode = default`? When we discussed config declarations, we decided to register `config.mode` option. If it's set to `default` (this is a default value), CKAN will run even with the wrong configuration, until it fails later in the code. But you'll get warnings in logs, like `Invalid value for SECRET_KEY (None): Missing value`.

If you want to prevent an application from startup if the configuration is not valid(for example, `SECRET_KEY` is missing), you have to switch to the strict mode: `config.mode = strict`. It has the sense to use `strict` as a default, but for v2.10 we decided to stay in less aggressive `default` mode.

amercader:
I think it's time to bite the bullet and enforce the current `strict` behaviour so we can handle mandatory config options consistently. 
As of current master the handling of required options is a mess. These are the current options marked as `required` in the config declaration:

| Missing option | Error | 
| ------------------- | ------- |
| `ckan.site_url` | `RuntimeError: ckan.site_url is not configured and it must have a value. Please amend your .ini file.` |
| `SECRET_KEY` / `beaker.session.secret` | `RuntimeError: You must provide a value for the secret key with the SECRET_KEY config option` |
| `sqlalchemy.url`| `AttributeError: 'NoneType' object has no attribute '_instantiate_plugins'` |
| `beaker.session.validate_key` | CKAN will start, fail on first request with `beaker.exceptions.BeakerException: No validate_key specified for Cookie only Session.` (#7602 ) |
| `solr_url` | CKAN will start, but lots of Solr errors are shown and will crash on first request  |
| `ckan.plugins` | CKAN will start, I don't think this option is actually required. |

If we manually check for a config option (first two options), it's done in different parts of the stack, leading to problems like the one I described in the comment above.

The expected behaviour should be that if an option is flagged as required, CKAN should fail to start when the configuration is validated (and hopefully nothing is using required options before that). I think given that 2.10 has the `config.mode= default/strict` option we can set the default to strict for 2.11. We can decide to keep the current behaviour as `config.mode = lax` or remove it entirely.

what do you think?

smotornyuk:
My huge +1 for strict mode as default. We are running our v2.10 application on strict mode and it does not create any problems, as long as the config file is valid. And in case of invalid config, it helps to identify an error, so it's rather an improvement, than a hindrance. But I'd like to review all the validators for built-in config options and double-check that there are no edge cases, where invalid validators are added to an option.

And you mentioned before the `required` flag. It's one of the features I planned but didn't yet finalize. At the moment, the `required` flag does not change the validation of the config option and it's possible to start the application without the `required` option even in strict mode, as long as it has no validators.
This problem is addressed in #7658. It adds a `not_empty` validator to every `required` option. This means, you can write the following:
```yaml
  key: SECRET_KEY
  required: true
```
and CKAN will not start in the strict mode without `SECRET_KEY`. And it will show the appropriate error message.

amercader:
Thanks for the great work @TomeCirun , I picked it up and follow up with more changes in #7781

PR #7440, opened 2023-03-13
Upgrade Flask-login, refactor tests

amercader:
All details in #7439 

This PR:

* Removes most of the `with_request_context` fixture, which should be used sparingly
* Removes usage of `REMOTE_USER` as a way to authenticate, as it does not skip CSRF checks (current tests were only incidentally passing)
* Consolidates all tests to use `{"Authorization": user["token"]}` (or `{"HTTP_AUTHORIZATION": user["token"]}` when testing redirect responses)
* Upgrades Flask-login from 0.6.1 to 0.6.2

PR #7453, opened 2023-03-16
[7408]Authentication header name

### LINKED ISSUE #7408: Authentication header name
As of CKAN 2.10, you can provide auth information in an API call:

> To provide your API token in an HTTP request, include it in either an `Authorization` or `X-CKAN-API-Key` header. (The name of the HTTP header can be configured with the `apikey_header_name` option in your CKAN configuration file.)

[Source](https://docs.ckan.org/en/2.10/api/index.html#authentication-and-api-tokens)

This was a bit of a surprise to me, as I've always assumed that we only checked for the `Authorization` header, which you could change to `X-CKAN-API-Key`  or whatever using the `apikey_header_name` conf. Except for one exception, the docs always show `Authorization` being used.

So my first question is, do we want to keep this default behaviour (ie support both headers)?

Regardless, I think that the [code](https://github.com/ckan/ckan/blob/e8b57b94aa083e0de922714091632e14ce29ccc0/ckan/views/__init__.py#L122-L135) that reads the headers needs a bit of cleanup (`##` Comments are mine):

```python
 def _get_user_for_apitoken() -> Optional[model.User]:  # type: ignore 
     
     ## This defaults to `X-CKAN-API-Key`, not `Authorization` (see [1])
     apitoken_header_name = config.get("apikey_header_name") 

     ## First we check if there's a `X-CKAN-API-Key` header in the request, 99.9% of the time it ain't there
     apitoken: str = request.headers.get(apitoken_header_name, u'') 

     if not apitoken: 
         ## We will never get here, headers are not added directly to the WSGI environ
         apitoken = request.environ.get(apitoken_header_name, u'') 
     if not apitoken:
         ## Cryptic ancient comment below, but actually that's the condition that is met 99.9% of the time
         ## when a request with the `Authorization` header is sent (see [2])
         # For misunderstanding old documentation (now fixed). 
         apitoken = request.environ.get(u'HTTP_AUTHORIZATION', u'') 
     if not apitoken: 
         ## We will never get here, headers are not added directly to the WSGI environ
         apitoken = request.environ.get(u'Authorization', u'') 
         # Forget HTTP Auth credentials (they have spaces). 
         if u' ' in apitoken:
             apitoken = u''
```
[1] [config declaration](https://github.com/ckan/ckan/blob/e8b57b94aa083e0de922714091632e14ce29ccc0/ckan/config/config_declaration.yaml#L371) for  `apikey_header_name`
[2] HTTP Request Headers are added to the [WSGI environ](https://wsgi.readthedocs.io/en/latest/definitions.html#standard-environ-keys) as `HTTP_<header_name_in_all_caps>`

The `request.environ.get(u'HTTP_AUTHORIZATION', u'')` seems too brittle to rely on considering it is the one that will catch most of the requests. I would suggest refactor ad cleanup the function above to specifically check the request headers for `Authorization` and fall back to `X-CKAN-API-Key` / custom key if not found.

In addition on a newly created ini file, you get the default value of `X-CKAN-API-Key`, which I think is confusing considering that we are using `Authorization` everywhere else:

```ini 
## Site Settings ###############################################################                                                                                                                                                                                                     
ckan.site_url = http://localhost:5000                                                                                                                                                                                                                                                
apikey_header_name = X-CKAN-API-Key                                                                                                                                                                                                                                                  
ckan.cache_expires = 0                                                                                                                                                                                                                                                               
ckan.cache_enabled = false   
#...
```

So we could

- Change the default header name to `Authorization`
- Support `X-CKAN-API-Key`  as special case to support current behaviour (or not, and just drop it on 2.11)
- Rename `apikey_header_name` to `apitoken_header_name` using the legacy option to support the old key

TomeCirun:
Fixes #7408 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -367,13 +367,13 @@ groups:
 
           .. warning:: This setting should not have a trailing / on the end.
 
-      - key: apikey_header_name
-        default: X-CKAN-API-Key
-        example: API-KEY
+      - key: apitoken_header_name
+        default: Authorization
+        legacy_key: apikey_header_name  # since v2.10.0
+        example: API-TOKEN
         description: |
-          This allows another http header to be used to provide the CKAN API
-          key. This is useful if network infrastructure blocks the
-          Authorization header and ``X-CKAN-API-Key`` is not suitable.
+          This allows Authorization http header to be used to provide the CKAN API
+          token.
 
       - key: ckan.cache_expires
         default: 0
--- a/ckan/tests/cli/test_config.py
+++ b/ckan/tests/cli/test_config.py
@@ -199,7 +199,6 @@ class TestSearch(object):
 class TestUndeclared(object):
     def test_no_undeclared_options_by_default(self, command):
         result = command("undeclared", "-idatapusher", "-idatastore")
-        breakpoint()
         assert not result.output
         assert not result.exit_code, result.output
 
--- a/ckan/tests/cli/test_config.py
+++ b/ckan/tests/cli/test_config.py
@@ -199,6 +199,7 @@ class TestSearch(object):
 class TestUndeclared(object):
     def test_no_undeclared_options_by_default(self, command):
         result = command("undeclared", "-idatapusher", "-idatastore")
+        breakpoint()
         assert not result.output
         assert not result.exit_code, result.output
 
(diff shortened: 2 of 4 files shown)


TomeCirun:
@amercader the tests fails because of `extra_environ=sysadmin_env` <-- this will add the Authorization header into `request.environ` and not in the `request.headers`, after #7440 gets merged this wouldn't be a problem. So the question is, do you want me to keep this extra line I've added here [34c87738efd291048474d615c2b770be324ab728](`https://github.com/ckan/ckan/pull/7453/commits/34c87738efd291048474d615c2b770be324ab728`) ?

amercader:
Closed in favour of #7841

PR #7464, opened 2023-03-17
[#7406] Update config to default to client-based sessions for new sites

### LINKED ISSUE #7406: Default to client-based sessions for new sites
The Flask-Login refactor introduced changes regarding logged in users and the session. To re-cap:
Flask-login stores the logged-in user identifier in the Flask session (as opposed to the `auth_tkt` in CKAN<2.10). CKAN uses [Beaker](https://beaker.readthedocs.io/en/latest/sessions.html) to manage the session, and the default session backend stores this session information as files on the server (on `/tmp`). This means that **if the session data is deleted in the server, all users will be logged out of the site**.
This can happen for instance:
* if the CKAN container is redeployed in a Docker / cloud setup and the session directory is not persisted
* if the sessions are periodically cleaned by an external script

| Action | CKAN < 2.10 | CKAN >= 2.10 |
| ------ | ----------- | ------------ |
| Clear cookies | User logged out | User logged out (If `remember_me` cookie is deleted) |
| Clear server sessions | User still logged in | User logged out |

The way to keep the old behaviour with the Beaker backend is to store the session data in the [cookie itself](https://beaker.readthedocs.io/en/latest/sessions.html#cookie-based) (but this stores *all* session data, not just the user identifier):

```
# ckan.ini
beaker.session.type = cookie
beaker.session.validate_key = tT0ka0aOAWEPbvSOSog7c4ZFu

# These should probably be defaults anyway
beaker.session.httponly = True
beaker.session.secure = True
beaker.session.samesite = Lax # or Strict
```

Going through the Flask-login discussion, it looks like we decided to default to storing the session in the cookie on new instances ( https://github.com/ckan/ckan/pull/6560#issuecomment-1161693905). We didn't end up implementing this in CKAN 2.10, should we aim for it in CKAN 2.11?

If we go ahead this will probably just involve configuration and documentation changes.

tino097:
Fixes #7406

   - beaker.session.samesite
   - beaker.session.type
   - beaker.session.httponly
   - beaker.session.validate_key
   - beaker.session.secure

Values should be updated in the .ini file

- [ ] includes API changes


ckan/config/declaration/option.py
```diff
@@ -363,12 +364,13 @@ def _validators_from_string(s: str) -> list[Validator]:
     from ckan.logic import get_validator
 
     out = []
-    parts = s.split()
+    parts = re.split(r"\s+", s) if s else s.split()
```
smotornyuk, line 367: Can you explain it? The only difference I see it that string that consists of multiple space characters - `      ` - will be turned into `['', '']` instead of `[]`.

ckan/config/declaration/option.py
```diff
@@ -363,12 +364,13 @@ def _validators_from_string(s: str) -> list[Validator]:
...
+
     for p in parts:
         if "(" in p and p[-1] == ")":
             name, args = p.split("(", 1)
             args: Any = args[:-1].split(",")  # trim trailing ')', break up
-            v = get_validator(name)(*args)
+            v = get_validator(name)(args)  # type: ignore
```
smotornyuk, line 373: I really don't like it. Yes, you get the ability to use `OneOf`, but you won't be able to use custom validator factories anymore, like:
```python
def create_min_max_validator(min, max):
    def validator(value):
        ...
    return validator 

# declaration
validators: create_min_max_validator(1, 2)
```


And don't forget, that in the python code `OneOf(x, y)` won't work either. it must be `OneOf([x, y])`. You have to pass there a single collection.

So here we have a breaking change + a new behavior. I think that it has more sense to:
 
* parse literals(i.e. allow something like `validators: OneOf(["Lax", "Strict"])`). It's probably too much for now and requires careful planning
* create a separate validator. maybe 
  ```python
  def samesite_value(value):
      return one_of(["Lax", "Strict", "None"])(value)
  ```

ckan/config/declaration/option.py
tino097: i've used the regex becase we can improve how we can split the incoming string but also `s.split` didnt returned the values i was expecting.  And because  the `[' ']` im addint the `if` statement. I was aware that this is not the greatest solution but only workaround to get current implementation to work

ckan/config/declaration/option.py
tino097: I thought having specific validator only for `samesite` field is not making much sense but we will need more generic solution to check options with `OneOf` 

I was unaware about the custom factorie and i will try to improve the solution, although spliting the string is the toughest challenge as we could hit more validators e.g `validator1(a,b) validator2 validator3(a)`

wardi: parsing literals with `ast.literal_eval` should be fine. We can fall back to treating the parameter as a simple string when that fails with a ValueError for backwards compatibility. I'd like to do the same in scheming if you make that change here

tino097:
@smotornyuk @wardi 
This is not going well with the more generic apporach so i will leave it with the `seamsite_validator` for now and the PR is ready for review

smotornyuk:
After the upgrade, all sites will use cookie-based session, because it's a default value of `beaker.session.type`. We need a changelog entry that describes this change and mentions that `beaker.session.validate_key` is required for cookie-based sessions. In addition, it would be nice to explain, how to keep server-side session if somebody likes it.

And I think that we don't need `required: True` for any of the new options. `beaker.session.type` has a default value, so it's not a required option. `beaker.session.validate_key` is required only for cookie-based session and it's already has a `not_empty` validator. `beaker.session.samesite` has a default value.


Oh, and you need to update test-core.ini, because a lot of tests fail without `beaker.session.validate_key

smotornyuk:
Can you remove `samesite_validator` as well, because you no longer using it

PR #7465, opened 2023-03-17
Group and Organization Title Translations

JVickery-TBS:
### Features:

- New and modified helpers to get group translated titles over `display_name`
- Updated templates to use new helper
- Added condition to facets to be able to display group translated titles

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1286,7 +1286,7 @@ def linked_user(user: Union[str, model.User],
 def group_name_to_title(name: str) -> str:
     group = model.Group.by_name(name)
     if group is not None:
-        return group.display_name
+        return p.toolkit.h.get_translated(group, 'title')
     return name
 
 
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1285,8 +1285,10 @@ def linked_user(user: Union[str, model.User],
 @core_helper
 def group_name_to_title(name: str) -> str:
     group = model.Group.by_name(name)
-    if group is not None:
-        return p.toolkit.h.get_translated(group, 'title')
+    group_dict = logic.get_action(u'%s_show' % group.type)(
+                    {}, {u'id': group.id})
+    if group_dict is not None:
+        return p.toolkit.h.get_translated(group_dict, 'title') or group_dict.name
     return name
 
 
@@ -1687,17 +1689,29 @@ def time_ago_from_timestamp(timestamp: int) -> str:
     return formatters.localised_nice_date(datetime_, show_date=False)
 
 
+def _object_display_name(obj_dict: dict[str, Any]) -> str:
+    return get_translated(obj_dict, 'title') or obj_dict['name']
+
+
 @core_helper
 def dataset_display_name(
         package_or_package_dict: Union[dict[str, Any], model.Package]) -> str:
+    if not isinstance(package_or_package_dict, dict):
+        package_or_package_dict = logic.get_action(u'package_show')(
+            {}, {u'id': package_or_package_dict.id})
+    return _object_display_name(package_or_package_dict)
 
-    if isinstance(package_or_package_dict, dict):
-        return get_translated(package_or_package_dict, 'title') or \
-            package_or_package_dict['name']
-    else:
-        # FIXME: we probably shouldn't use the same functions for
-        # package dicts and real package objects
-        return package_or_package_dict.title or package_or_package_dict.name
+
+@core_helper
+def group_display_name(
+        group_or_group_dict: Union[dict[str, Any], model.Group]) -> str:
+    if not isinstance(group_or_group_dict, dict) \
+       or 'display_name' not in group_or_group_dict:
+        group_or_group_dict = logic.get_action(u'%s_show' %
+                                                group_or_group_dict[u'type'])(
+                                                {}, 
+                                                {u'id': group_or_group_dict[u'id']})
+    return _object_display_name(group_or_group_dict)
 
 
 @core_helper
@@ -2378,7 +2392,6 @@ def list_dict_filter(list_: list[dict[str, Any]],
 
     :param value: the value to search for
     '''
-
     for item in list_:
         if item.get(search_field) == value:
             return item.get(output_field, value)
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -2897,7 +2897,7 @@ def followee_list(
         '''Return a display name for the given user, group or dataset dict.'''
         display_name = followee.get('display_name')
         fullname = followee.get('fullname')
-        title = followee.get('title')
+        title = plugins.toolkit.h.get_translated(followee, 'title')
         name = followee.get('name')
         return display_name or fullname or title or name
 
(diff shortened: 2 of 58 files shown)


ckan/lib/helpers.py
```diff
@@ -1285,9 +1285,14 @@ def linked_user(user: Union[str, model.User],
...
-        return group.display_name
-    return name
+    if group is None:
+        return name
+    group_dict = logic.get_action(u'%s_show' % group.type)(
+                    {}, {u'id': group.id})
+    if group_dict is None:
+        return name
+    return p.toolkit.h.get_translated(group_dict, 'title') \
+        or group_dict.name
```
wardi, line 1295: If this is called in a loop like in the facet titles below it's going to be really slow. I don't see what other part of core is using this helper, maybe we should plan to remove it instead?

ckan/lib/helpers.py
```diff
@@ -1687,17 +1692,36 @@ def time_ago_from_timestamp(timestamp: int) -> str:
...
+        package_or_package_dict = logic.get_action(u'package_show')(
+            {}, {u'id': package_or_package_dict.id})
+    return _object_display_name(package_or_package_dict)  # type: ignore
 
-    if isinstance(package_or_package_dict, dict):
-        return get_translated(package_or_package_dict, 'title') or \
-            package_or_package_dict['name']
-    else:
-        # FIXME: we probably shouldn't use the same functions for
-        # package dicts and real package objects
```
wardi, line 1699: These comments make sense, we should keep them

ckan/lib/helpers.py
```diff
@@ -1687,17 +1692,36 @@ def time_ago_from_timestamp(timestamp: int) -> str:
...
-        return package_or_package_dict.title or package_or_package_dict.name
+
+@core_helper
+def group_display_name(
```
wardi, line 1709: If we're making a new helper can we not follow the one above that works on models or dicts and only accept dicts?

ckan/lib/helpers.py
```diff
@@ -1687,17 +1692,36 @@ def time_ago_from_timestamp(timestamp: int) -> str:
...
+        group_or_group_dict: Union[dict[str, Any], model.Group]) -> str:
+    if not isinstance(group_or_group_dict, dict):
+        group_type = group_or_group_dict.type
+        group_id = group_or_group_dict.id
+    elif 'display_name' not in group_or_group_dict:
+        group_type = group_or_group_dict[u'type']
+        group_id = group_or_group_dict[u'id']
```
wardi, line 1716: why do we need to call *_show if display_name is missing? We're not using it now, right?

ckan/logic/action/get.py
```diff
@@ -2897,7 +2897,7 @@ def display_name(followee: dict[str, Any]) -> Optional[str]:
         '''Return a display name for the given user, group or dataset dict.'''
         display_name = followee.get('display_name')
         fullname = followee.get('fullname')
-        title = followee.get('title')
+        title = plugins.toolkit.h.get_translated(followee, 'title')
```
wardi, line 2900: I've never used this endpoint. Does a change this simple really work here?

wardi:
The big problem with this change is making group_show/organization_show calls in a loop for generating pages like search results and organization lists. We could find some way of getting the translated names back from solr to avoid all that extra work. How are the group/org titles populated in facets right now?

ckan/lib/helpers.py
JVickery-TBS: @wardi I did not fully remove the display_name from the groups, as it was used in a lot of places still, and seems like packages also has some display name thing going on with them.

If I am remembering correctly, these conditions were for the facets, in which the passed dict of the facet item does not contain the organization object or anything.

Should I do a bunch more stuff to just fully remove the display_name from groups, and then figure out the facet stuff later?

JVickery-TBS:
@wardi Yeah, the best way would for sure just be getting translated fields back from solr or the db in the first query. How would I do that?

And group titles in facets use the display_name definition from the model.Group. So it populated as `title` or `name`. But of course, if you are using fluent, then there is no title and you would only get the `name` always.

ckan/lib/helpers.py
JVickery-TBS: I recall now that I think the condition is for the breadcrumb, not the facet.

ckan/lib/helpers.py
JVickery-TBS: Okay, nothing is going to be using this now. I have added in a deprecation notice for it's removal.

ckan/lib/helpers.py
JVickery-TBS: Comments are added back in.

ckan/lib/helpers.py
JVickery-TBS: New helper is only doing dicts now. Simplified a lot, which is nice.

ckan/lib/helpers.py
JVickery-TBS: Removed all of these conditions, as we should really already have the dicts and `_translated` fields at this point. So we should not be doing that in the helpers here. This helper is only really returning the `get_translated` title OR name.

ckan/logic/action/get.py
JVickery-TBS: I will test it out some more to make sure that it works for groups and orgs, and this does not break following users.

ckan/lib/helpers.py
```diff
@@ -1687,17 +1692,36 @@ def time_ago_from_timestamp(timestamp: int) -> str:
...
+    if not isinstance(group_or_group_dict, dict) \
+       or 'display_name' not in group_or_group_dict:
+        group_or_group_dict = logic.get_action(
+                                u'%s_show' %
+                                group_type)(  # type: ignore
+                                {},
+                                {u'id': group_id})  # type: ignore
```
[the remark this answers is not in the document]
JVickery-TBS: N/A now, as new helper has been simplified.

ckan/lib/helpers.py
```diff
@@ -1687,17 +1692,36 @@ def time_ago_from_timestamp(timestamp: int) -> str:
...
+        group_type = group_or_group_dict.type
+        group_id = group_or_group_dict.id
+    elif 'display_name' not in group_or_group_dict:
+        group_type = group_or_group_dict[u'type']
+        group_id = group_or_group_dict[u'id']
+    if not isinstance(group_or_group_dict, dict) \
+       or 'display_name' not in group_or_group_dict:
+        group_or_group_dict = logic.get_action(
+                                u'%s_show' %
+                                group_type)(  # type: ignore
```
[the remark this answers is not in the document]
JVickery-TBS: N/A now, as new helper has been simplified.

ckan/logic/action/get.py
JVickery-TBS: @wardi okay this does not work, but for the same reason that the Dashboard group and organization route do not work.

These things just do some queries and puts the results directly into the dictization methods, and returns that. So these things do not go through any `plugin_validate` stuff, for any further processing??

JVickery-TBS: As discussed, we will grab the relative `group_show`, `organization_show`, or `package_show` in the followee stuffs.

JVickery-TBS:
@wardi Everything is passing now.

Are you able to re-review whenever you have time?

Notable things are that the package_dictize will no longer contain organization and group objects, those get set in package_show.

PR #7466, opened 2023-03-19
[7415]Resource does not refresh the format

### LINKED ISSUE #7415: Replacing a file in an existing resource does not refresh the format
## Describe the bug

When replacing an existing uploaded or linked file in a resource with one with a different format, the format of the previous file is kept. Apart from the inconsistency this is a problem because it affects the views created for the resource, as most view plugins rely on the format to be available or not.

The resource name is also kept, and as this defaults to the file name it is confusing.

### Steps to reproduce
* Create a resource, upload a `test.csv` CSV file. The format of the created resource is CSV and the name "test.csv"
* Edit this resource, clear the file and upload a PNG. The format and the name are still CSV and test.csv

### Expected behavior
I'd expect the metadata to be updated to reflect the actual file being uploaded, ie correct format and name

### Additional details

* The format file has the [`if_empty_guess_format`](https://github.com/ckan/ckan/blob/f5f74b7da9bc42fe0b4a0d4690de616e7a46fd2d/ckan/logic/schema.py#L43) validator, which makes sense for new resources (if the user provided their own format, don't overwrite it). But with existing files is more complicated as we need to know if the user has changed the value of format.
* I don't think anything on the schema about defaulting to the file name, so I assume this is done in javascript

TomeCirun:
Fixes #7415 

### Proposed fixes:
more info soon..

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/navl/validators.py
+++ b/ckan/lib/navl/validators.py
@@ -1,6 +1,6 @@
 # encoding: utf-8
 
-from typing import Any, Callable, NoReturn
+from typing import Any, Callable, NoReturn, Optional
 import six
 
 
@@ -8,7 +8,8 @@ import ckan.lib.navl.dictization_functions as df
 
 from ckan.common import _, json, config
 from ckan.types import (
-    Context, FlattenDataDict, FlattenErrorDict, FlattenKey, Validator
+    Context, FlattenDataDict, FlattenErrorDict, FlattenKey, Validator,
+    Model
 )
 
 missing = df.missing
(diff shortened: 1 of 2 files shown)


TomeCirun:
Hello @amercader , I've added and tested the same logic as a schema validator, attached to `default_update_package_schema`, the problem with this approach is, it will also trigger the validator when we create the resource, plus with this approach, if we had let's say 10 resources, it will call the validator precisely 10 times.

That's why I found this solution to be more suitable, as it's going to call only once when we update the resource. 

Let me know your opinion on this, I will also add tests and will do the same for the resource name if you agree.

wardi:
There are multiple ways a resource can be uploaded or a url changed without going through `resource_update` including:
- package_update
- package_patch
- package_revise

So an approach using a validator as part of the schema makes more sense (sites can also override the logic easily this way)

Regarding update performance, this is a known issue for datasets with many resources. We can fix this by making `package_update` smarter, this way it will be solved for all resource fields and all API calls not just this one field and endpoint.

PR #7468, opened 2023-03-20
Fix primary-secondary blocks direction for mobile view

mutantsan:
The primary block now goes before the secondary for mobile view, but before it was the opposite which makes sense. 
Restoring previous order.

image

### Features:

- [ ] includes API changes


ThrawnCA:
I disagree with the reasoning behind this. If there is content so important that it needs to be visible first, like the warning to be careful with sysadmin privileges, then it should not be marked as _secondary_ content; it has been misplaced.

The primary content should be the first thing seen on mobile; that is what "primary" means.

PR #7475, opened 2023-03-20
Fix mobile styles for activity, few updates

mutantsan:
Few minor changes to the activity. I don't think that it's a good idea to split it.

image
image

1. Fix mobile styles
2. Use default CKAN styles for a label/select
3. Do not show pagination if we don't have the next or previous page.
4. Show a message if we don't have any activities
5. Fix untranslated string
6. Title select items to match the `All activity types` item.

### Features:

- [ ] includes API changes


pdelboca:
Thanks @mutantsan !

PR #7477, opened 2023-03-20
Show flash error on promote page

mutantsan:
Show flash error messages on the promote page, instead of abort response.
1. Show "user not found" if the user doesn't exist
2. Show ValidationError's if the user schema has been altered and `user_patch` throws an exception.

### Features:

- [ ] includes API changes


mutantsan:
@wardi, fixed the test

mutantsan:
still a problem, updated @wardi

pdelboca:
I like this change @mutantsan ! Thanks!

For future reference: We are not using the `u` prefix anymore. So feel free to remove them for the strings if you see one.

PR #7478, opened 2023-03-20
Munge site user name and add default email for it

mutantsan:
**I'm not sure about this one.** The initial problem - we have a mandatory field email, and probably site user should have a default one. I've pressed the Unpromote button on the `admin.index` page and got a 5xx error. As a sysadmin, I don't want to see any errors while I'm navigating the site.

The whole idea of demoting `site_user` from being a sysadmin is weird. Maybe we don't have to allow it?

The other side is that there's no validation for `ckan.site_id` config option. And we are creating a user via database, so no schema applies here. So we could use something like that `test 001 002 $` that ends up creating a user with an improper name. I assume we have to munge.

Adding the `munge` function here will trigger the recreation of `site_user`, could it be a problem?

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -27,6 +27,7 @@ import ckan.model as model
 import ckan.model.misc as misc
 import ckan.plugins as plugins
 import ckan.lib.search as search
+from ckan.lib.munge import munge_name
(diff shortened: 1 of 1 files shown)


mutantsan:
still a problem, requires discussion or smth @wardi @amercader

PR #7487, opened 2023-03-22
[#7486] Fix error when calling user_create with defer_commit

### LINKED ISSUE #7486: `defer_commit` not working properly on most create actions
## CKAN version
2.10

## Describe the bug

When calling `user_create` with `defer_commit=True` in the context it fails with:

```
2023-03-21T21:33:15.0753914Z     def do_execute(self, cursor, statement, parameters, context=None):
2023-03-21T21:33:15.0754056Z >       cursor.execute(statement, parameters)
2023-03-21T21:33:15.0754502Z E       sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "user_id" violates not-null constraint
2023-03-21T21:33:15.0754816Z E       DETAIL:  Failing row contains (null, 2023-03-21 21:32:13.042803, 2023-03-21 21:32:13.042813).
2023-03-21T21:33:15.0754904Z E       
2023-03-21T21:33:15.0755181Z E       [SQL: INSERT INTO dashboard (activity_stream_last_viewed, email_last_sent) VALUES (%(activity_stream_last_viewed)s, %(email_last_sent)s)]
2023-03-21T21:33:15.0755651Z E       [parameters: {'activity_stream_last_viewed': datetime.datetime(2023, 3, 21, 21, 32, 13, 42803), 'email_last_sent': datetime.datetime(2023, 3, 21, 21, 32, 13, 42813)}]
2023-03-21T21:33:15.0755871Z E       (Background on this error at: https://sqlalche.me/e/14/gkpj)
2023-03-21T21:33:15.0755879Z 
2023-03-21T21:33:15.0756186Z /usr/lib/python3.10/site-packages/sqlalchemy/engine/default.py:736: IntegrityError

```

### Additional details
 This is are regression in CKAN 2.10, introduced by #6790 

When activities were created as part of the `user_create` call, we did a `session.flush()` call to populate the the `id` field in the user model created, so the activities could store it as well:

https://github.com/ckan/ckan/blob/ad1b08c66a12538f1ca3017fb3080dc99c017bd6/ckan/logic/action/create.py#L1063-L1065

When the activities code was removed, this was removed as well. In most cases this is fine, because we are committing a bit below:

https://github.com/ckan/ckan/blob/ed5367765fb22686b3aa5446d83aa8a9e57207ba/ckan/logic/action/create.py#L1004-L1006

But if `defer_commit` is True, the commit doesn't happen, and when we try to create a `Dashboard` object `user.id` is still None:

https://github.com/ckan/ckan/blob/ed5367765fb22686b3aa5446d83aa8a9e57207ba/ckan/logic/action/create.py#L1028

We should flush the session if we are not committing.

**Update**:

In addition to the issue above, `defer_commit` is not being honored because:

* `user_create` calls `Dashboard.get()`, which creates and commits an object under the hood (booo)
* The activity plugins always commits when creating activities

amercader:
Fixes #7486

We need to flush the session if we are not committing in order to populate user.id

### Proposed fixes:

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1004,6 +1004,10 @@ def user_create(context: Context,
     if not context.get('defer_commit'):
         with logic.guard_against_duplicated_email(data_dict['email']):
             model.repo.commit()
+    else:
+        # The Dashboard object below needs the user id, and if we didn't
+        # commit we need to flush the session in order to populate it
+        session.flush()
 
     # A new context is required for dictizing the newly constructed user in
     # order that all the new user's data is returned, in particular, the
--- a/changes/7487.bugfix
+++ b/changes/7487.bugfix
@@ -1,0 +1,1 @@
+Fix exception when calling the `user_create` action with `defer_commit` in the context.
--- a/ckan/tests/logic/action/test_create.py
+++ b/ckan/tests/logic/action/test_create.py
@@ -1218,6 +1218,17 @@ class TestUserCreate(object):
             "password": ["Your password must be 8 characters or longer"]
         }
 
+    def test_user_create_defer_commit(self):
+        stub = factories.User.stub()
+        user_dict = {
+            "name": stub.name,
+            "email": stub.email,
+            "password": "test1234",
+        }
+        context = {"defer_commit": True}
+
+        helpers.call_action("user_create", context=context, **user_dict)
+
 
 @pytest.mark.usefixtures("clean_db")
 @pytest.mark.ckan_config("ckan.auth.create_user_via_web", True)


ckan/logic/action/create.py
```diff
@@ -774,18 +775,17 @@ def _group_or_org_create(context: Context,
         'model': model,
         'user': user,
         'ignore_auth': True,  # we are not a member of the group at this point
-        'session': session
+        'session': session,
+        'defer_commit': context.get("defer_commit") or False,
```
wardi, line 779: As discussed at the dev call if we're going to more widely support `defer_commit` across actions let's add it to the `fresh_context` function and use that one instead of the code here

amercader:
@wardi added `defer_commit` to `fresh_context()` and used that instead of manually building context objects in the places touched by this PR

wardi:
Thank you @amercader !

PR #7491, opened 2023-03-23
Restyle activites, rework layout

mutantsan:
The activities layout is pretty bad and doesn't allow you to restyle it properly if you want to apply some custom changes. I've reworked the templates and updated styles to give it a modern look.

Old:
image
image

New:
image
image

But the main changes aren't in styles but in layout.
Old:
image
New:
image

Also, I've dropped some templates that definitely now working. (new/updated tag, resources, because it's covered by Package update). I didn't drop follow templates, while they aren't working (as well I can tell), but maybe I'll fix them in the future because it might be an activity we want to have.

### Features:

- [ ] includes API changes


mutantsan:
Requires review @Zharktas
IMHO, it's much better than current activities and supports easier restyle.

PR #7493, opened 2023-03-24
Fix member list to exclude deleted users

### LINKED ISSUE #7170: member_list action returns deleted users
**CKAN version**
2.9.7

**Describe the bug**
The `member_list` action with  `'object_type': 'user'` returns all member users, including deleted ones
This was fixed with #3078 but relies on the `user_delete` action.
If I manually delete a user from the DB, this will fail.

**Steps to reproduce**
Trying to iterate all members, you will get `toolkit.ObjectNotFound` for deleted users

```py
 users = toolkit.get_action('member_list')(context, {
            'id': container['id'],
            'object_type': 'user',
        })
       
for user in users:
     duser = toolkit.get_action('user_show')(context, {'id': user[0]})
```

**Expected behavior**
`member_list` action must return only active users

**Possible fix**
Add an extra DB query to filter only active users

shashigharti:
Fixes #7170

### Proposed fixes:

Fix member_list action to exclude deleted user when state(deleted) is not same as in user table. 

### Features:

- [ ] includes API changes


shashigharti:
@pdelboca  Thanks! . I will make the changes.

shashigharti:
@pdelboca I have made the changes to filter deleted user. Thanks!

pdelboca:
@shashigharti I'm having some issues following the proposed logic. Can you elaborate a little bit your approach and why are you proposing those two filters?

I was thinking on something more straight forward like:
```python
existing_users = model.Member.table_id.in_(model.Session.query(model.User.id))

q = model.Session.query(model.Member).\
    filter(model.Member.group_id == group.id).\
    filter(model.Member.state == "active").\
    filter(existing_users)
```

This will use the classic subquery approach and add: `AND member.table_id IN (SELECT "user".id FROM "user")` statement to the final SQL query.

avdata99:
My two cents here:
 - As @amercader mentioned in the issue, we need to implement a foreign key with the proper `CASCADE` behaviour between the `user` and `member` tables. The issue was raised when I manually deleted a user in the DB directly. Considering the `table_id` field in the `member` table is not always referring to `users`, maybe it's not so easy.
 - @pdelboca, this function retrieves member of different types, no just `users`, it's not easy to write a proper filter for all cases.
 - @shashigharti, I think we can simplify the query and extend this fix to `packages` with something like:

```py
outer_mdl = None
if obj_type == 'user':
    outer_mdl = model.User
elif obj_type == 'package':
    outer_mdl = model.Package

q = model.Session.query(model.Member)
if outer_mdl:
    q = q.join(outer_mdl, outer_mdl.id == model.Member.table_id, isouter=True).\
        filter(outer_mdl.state=='active')

 q = q.filter(model.Member.group_id == group.id).\
     filter(model.Member.state == "active").\
```

shashigharti:
Thank you very much @pdelboca and @avdata99. 

As @avdata99 mentioned, members table have different types(screenshot attached) . And because of that I used the above method to keep all the records from members table and then only remove records that were missing in the users table. I did not know if there were other types except user and package so I used the above query. 

If there are only two types of records in member table, then the code suggested by @avdata99 works. I tested and only 1 test was failing. So I will use the solution shared by Andrez.

What do you think?

avdata99:
@shashigharti I think also groups can be members of groups ([hierarchy extension do that](https://github.com/ckan/ckanext-hierarchy/blob/master/ckanext/hierarchy/tests/fixtures.py#L11-L14))

If all tests are green, I think that query will be ok

shashigharti:
Thanks @avdata99! . It also has "tag" type.

avdata99 (approved):

shashigharti:
Thank you @pdelboca!. I have revised the code and it is ready for review. I tried both solutions, another solution is [here](https://github.com/shashigharti/ckan/commit/cf98138148e1d925a3c2a953d98d24334eeb2701)

pdelboca:
@avdata99 this one looks good to go for me. Let me know if you have any other questions/suggestions.

PR #7494, opened 2023-03-24
Refactor get object logic

### LINKED ISSUE #7364: Refactor old _get_object logic
The authorization layer provides a _get_object method that can generate side effects if context objects are used in sequential action calls. (See #7348)

https://github.com/ckan/ckan/blob/e21141098f33c54e1e3fdf0a6d19922f5d54a3f7/ckan/logic/auth/__init__.py#L46-L66

This method is part of an old implementation of a "cache like" layer that could be modernized. Newer versions of sqlalchemy and databases provide their own cache logic when doing selects queries, therefore there would be no need to implement custom logic.

### Some ideas of implementation:

- `_get_object` should directly call to the database using the object id.
- There would be no need to store and pass the object in the context for "later reuse".

shashigharti:
Fixes #7364

### Proposed fixes:

Change `_get_object` logic to remove cache like logic.

### Features:

- [ ] includes API changes


test-core.ini
pdelboca, line 1: This might be undesired changes @shashigharti

shashigharti: Thanks! I will restore it.

shashigharti:
@pdelboca Thanks! 

> @shashigharti this looks like the desired solution. As I mentioned in #7364 the side effect of this "cache" logic was the missuse of the context object to retrieve sqlalchemy objects.
> 
> ```python
>         # Save in case we need this again during the request
>         context[name] = obj
> ```
> 
I did use the same logic and it was still failing. I will retry.

> I'm expecting tests to fail because some parts of our logic probably depends on an object existing in the context object (and shouldn't be the case). Can you dig into the failing tests and look for those scenarios?

Sure! I will also look into the causes of failing tests.

shashigharti:
@Patricio del Boca Regarding this change https://github.com/ckan/ckan/pull/7494 ,  I traced the flow of this test just to understand the error. While creating new resource, it makes multiple calls to `get_object` function  in `auth/_init__.py` and fails in one of the calls as it doesn't pass `id`. Line 311 in test_package.py causes that:
```
response = app.post(location, headers=headers, data={
            "id": "",
            "url": "http://example.com/resource",
            "save": "go-metadata"
        })
```
It works fine in first pass and also returns a valid object, but due to multiple calls the error occurs. 


2. It seems that error doesn't display exact reason in this case, it returns 409 error only. And I found 
that Error summary fails in ckan/logi/__init__.py in line 144 because it is a dict object and not a list.

Just wanted some feedback if I am understanding flow correctly or not? I will keep on looking into it thanks!

ckan/views/dataset.py
```diff
@@ -804,8 +804,9 @@ def get(self,
         pkg = context.get(u"package")
         resources_json = h.json.dumps(data.get(u'resources', []))
         user = current_user.name
+        pkg_id = pkg.id if pkg else None
         try:
-            check_access(u'package_update', context)
+            check_access(u'package_update', context, {"id": pkg_id})
```
pdelboca, line 809: @shashigharti we already have `pkg_dict` object so there should not be any reason to depend on a `package` that comes from the `context` object. (Part of the logic that we are trying to remove.)

Could you clean up this part to call directly to `pkg_dict["id"]` instead of adding a logic to get it from `pkg` object?

pdelboca:
Thanks @shashigharti for all the work here! We are getting there!

Regarding the tests, feel free to remove them since I would not like to keep code testing old behaviour that no longer exists.

Can you also add a changes file to describe that this function will no longer return the object in the `context`?  This will be useful for people in the future if for some reason they are depending on this for their workflow.

ckan/views/dataset.py
shashigharti: Thanks! I have removed it.

shashigharti:
Thank you very much @pdelboca! for the guidance. I have made the changes.

pdelboca:
@wardi this one is ready to review! Feel free to drop any comment!

pdelboca:
Thanks @shashigharti for this work!

I tested and it seems to be working just fine @wardi , let me know if you have any follow up on this.

PR #7509, opened 2023-03-30
[#7508] Escape names in emails

### LINKED ISSUE #7508: Mailer doesn't quote names in To/From headers
## CKAN version
2.9.8, though it doesn't appear to have changed on 2.10
## Describe the bug
The username and site title are included in the To/From headers respectively, but they aren't quoted. When one contains, for example, commas, it's displayed as several addresses.
(When site_title is `Dept of A, B, and C`, the emails appear to be from `Dept@{host}, of@{host}, A@{host}, B@{host}, and C,`)
### Steps to reproduce
Set a site_title to a string containing commas and trigger an email somehow. I've been using `ckanext-harvester` notifications.
### Expected behavior
That the name would be quoted and therefore show as `Dept of A, B, and C`.

### Additional details
#4343 appeared to address the same / a similar issue

bluepython508:
Fix #7508 by quoting names in To/From headers

### Proposed fixes:
Quote names in To/From headers
### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7508.bugfix
+++ b/changes/7508.bugfix
@@ -1,0 +1,1 @@
+Names are now quoted in From and To addresses in emails, meaning that site titles with commas no longer break email clients.
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -64,8 +64,8 @@ def _mail_recipient(
         else:
             msg.add_header(k, v)
     msg['Subject'] = subject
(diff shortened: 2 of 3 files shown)


amercader:
Maybe this helps quote the fields properly? https://docs.python.org/3/library/email.utils.html#email.utils.quote

wardi:
@bluepython508 please add a changes file that describes the bug fix

PR #7512, opened 2023-03-31
Allow use of X-Accel-Redirect for downloaded resources

bluepython508:
Allows the use of range requests when the handling server supports it, without needing to go through flask.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7512.feature
+++ b/changes/7512.feature
@@ -1,0 +1,1 @@
+Add `ckan.downloads_use_x_accel_redirect` to offload resource downloads to a reverse proxy, enabling range requests.
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1322,6 +1322,23 @@ groups:
         example: 10
         description: The maximum in megabytes an image upload can be.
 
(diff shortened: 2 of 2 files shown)


pdelboca:
@EricSoroos what's the status of this?

In terms of code what is missing is:
 - Proper entry in our `config_declarations.yaml`
 - A changelog file since it is a new addition.

pdelboca:
@EricSoroos `X-Accel` comes with a set of header to customize it's behaviour: https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/#special-headers

Should we add them to the config declaration as well?

EricSoroos:
I don't think we need to add the other headers -- we're pretty explicitly returning a path to a file, none of the others seem like they're really be necessary or appropriate for our use case. 

I think some in addition to what we're missing up there, some docs about what support is necessary on the NGINX side, since you do need the internal location block to properly serve the file.

EricSoroos:
I think this is mergable and potentially backportable to 2.9. (at least -- it's running in our 2.9 instances and this is effectively a forward port.)

PR #7523, opened 2023-04-06
Private Dataset Activities

### LINKED ISSUE #5772: Private datasets don't have an Activity Stream

vikig02:
Fixes #5772 

### Proposed fixes:

Enabled saving of activities on private datasets.
Added `permission_labels` column to `Activity` table.
Added filtering of dataset activities based on user permission labels.

### Features:

- [x] includes API changes


wardi:
@vgeorgieva-cstbg thank you for this, but instead of having an option that will make private datasets public through the activity feed we should use the IPermissionLabels interface to cover activities on datasets.

We need a new column to store permission labels in the activity table and a filter applied to activity feeds based on the user's permissions, like we have for dataset search right now. This would work for the default permissions in ckan as well as any custom permission scheme a site might have.

vgeorgieva-cstbg:
@wardi

When should the new db column be written to - when the activity is being created or at another time? Also should it be an array or string like it is for the datasets? And do I use the same labels that are used in datasets (creator, collaborator, etc.) or something else?

wardi:
@vgeorgieva-cstbg the permission labels are an array of strings https://github.com/ckan/ckan/blob/9ee6b25089026f84bdc0960f91ac4676311ff373/ckan/lib/search/index.py#L294 so in postgres that would be a `text[]` type column for storing with the activity.

When retrieving activities we need to filter results on the user's permission labels https://github.com/ckan/ckan/blob/8aaf1c65b1a83d53296b71e16cfff4464694aad1/ckan/logic/action/get.py#L1549 (as long as any user label is present on the activity then they can see it)

vgeorgieva-cstbg:
I made some changes to this. Can you check it out again?

vgeorgieva-cstbg:
@wardi I made some changes to this. Can you check it out again?

ckan/migration/versions/105_21d03ed1e1ca_add_permission_labels_in_activity_table.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '21d03ed1e1ca'
+down_revision = '9f33a0280c51'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('activity', sa.Column('permission_labels', sa.Text))
```
wardi, line 20: this should be an array like `text[]`, right?

ckanext/activity/logic/action.py
```diff
@@ -257,7 +258,22 @@ def package_activity_list(
...
+    user = context.get('user')
+    if not authz.is_sysadmin(user):
+        labels = get_permission_labels()
+        user_permission_labels = labels.get_user_activity_labels(context['auth_user_obj'])
+
+        for activity in activity_objects:
+            activity.permission_labels = labels.get_activity_labels(activity)
+
+            if activity.permission_labels and any(permission in activity.permission_labels for permission in user_permission_labels):
+                filtered_activity_objects.append(activity)
```
wardi, line 272: this should be done by the database not in python, or it's going to be very very slow

vgeorgieva-cstbg:
@wardi Thanks for the review. I made the changes you requested.

vgeorgieva-cstbg:
@wardi Thanks for the review. I made the changes you requested.

vgeorgieva-cstbg:
@wardi 

Is there anything else I should do for this?

vgeorgieva-cstbg:
@wardi

Is there anything else I should do for this?

wardi:
Thanks @vgeorgieva-cstbg I'll try to find some time to review your work soon

ckanext/activity/logic/action.py
```diff
@@ -246,6 +247,19 @@ def package_activity_list(
     after = data_dict.get("after")
     before = data_dict.get("before")
 
+    # Get activity permission labels
+    labels = get_permission_labels()
+    activities = model.Session.query(
+        model_activity.Activity).filter_by(object_id=package.id)
+    for activity in activities:
+        activity.permission_labels = labels.get_activity_labels(activity)
+    model.Session.commit()
```
wardi, line 256: updating the activity labels from a "side_effect_free" action doesn't make sense. Shouldn't these have already been set when the activities were created?

ckan/migration/versions/105_21d03ed1e1ca_add_permission_labels_in_activity_table.py
```diff
@@ -0,0 +1,27 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.add_column('activity', sa.Column(
+        'permission_labels', sa.ARRAY(sa.Text)))
+
+
+def downgrade():
+    op.drop_column('activity', 'permission_labels')
```
wardi, line 27: @smotornyuk where should activity model migrations live now that the model is in the activity extension?

ckan/lib/plugins.py
```diff
@@ -663,3 +663,41 @@ def get_user_dataset_labels(self, user_obj: model.User) -> list[str]:
...
+            {u'user': user_obj.id}, {u'permission': u'read'})
+        labels.extend(u'member-%s' % o[u'id'] for o in orgs)
+
+        if ckan.authz.check_config_permission('allow_dataset_collaborators'):
+            datasets = logic.get_action('package_collaborator_list_for_user')(
+                {'ignore_auth': True}, {'id': user_obj.id})
+
+            labels.extend('collaborator-%s' % d['package_id'] for d in datasets)
+
+        return labels
```
wardi, line 703: Importing `ckanext.activity.model` .. from core doesn't sit well with me.  @smotornyuk WDYT? Since the only labels generated and tested here are a generated with a copy of the same code used in the dataset labels, would we be better off without the new methods and just assuming that the only permission checks required will be based on dataset-activities, or should we create a new interface for activity permissions in the activity extension? We could start with dataset-only and think about adding a new interface when there is a specific need for one.

ckanext/activity/logic/action.py
vgeorgieva-cstbg: @wardi I've moved the initial set of the activity labels to when the activity is created. But aren't there things (ex. changing dataset's organization) that would require a change in the activity labels? How should that happen?

wardi: If we leave the labels on the activities as they are then members of the old org can see changes that occurred while the dataset was part of that org and members of the new org can only see changes that occurred since. I think that's a sensible default.

ckan/lib/plugins.py
wardi: Just confirmed with the team: It would be better to remove these methods and call the existing dataset ones for dataset-related activities from the activity extension instead.

ckanext/activity/model/activity.py
```diff
@@ -784,3 +807,31 @@ def _activity_stream_get_filtered_users() -> list[str]:
...
+    """Adds a filter to an existing query object to 
+    exclude package activities based on user permissions.
+    """
+    print(user_permission_labels)
+    print()
+
+    # `user_permission_labels` is None when user is sysadmin
+    if user_permission_labels:
+        print(q.all())
+        print()
```
wardi, line 825: clean up print statements?

ckanext/activity/subscriptions.py
```diff
@@ -94,16 +95,19 @@ def package_changed(sender: str, **kwargs: Any):
...
     else:
         user_id = "not logged in"
 
     activity = Activity.activity_stream_item(pkg, type_, user_id)
+
+    # Set activity permission labels
+    if activity:
+        activity.permission_labels = get_permission_labels(
+        ).get_dataset_labels(pkg)
+
```
wardi, line 110: move this to activity_stream_item so it covers all cases?

vgeorgieva-cstbg:
@wardi Can you please resolve the comment threads that are fixed now? Anything else I should change?

ckanext/activity/model/activity.py
```diff
@@ -784,3 +810,26 @@ def _activity_stream_get_filtered_users() -> list[str]:
...
+def _filter_activities_by_permission_labels(
+    q: QActivity,
+    user_permission_labels: Optional[list[str]] = None,
+):
+    """Adds a filter to an existing query object to
+    exclude package activities based on user permissions.
+    """
+
+    # `user_permission_labels` is None when user is sysadmin
+    if user_permission_labels:
```
wardi, line 824:
```suggestion
    if user_permission_labels is not None:
```
Don't want to confuse an empty list (possible normal user with no permission labels) with None

ckanext/activity/tests/logic/test_action.py
```diff
@@ -840,6 +948,78 @@ def test_limit_hits_max(self):
...
+            "package_update", context={"user": another_user["name"]}, **dataset
+        )
+        activities = helpers.call_action(
+            "user_activity_list",
+            id=another_user["id"],
+            context={"user": user["name"], "auth_user_obj": auth_user_obj}
+        )
+
+        assert len(activities) == 1
+        assert activities[0]["activity_type"] == "changed package"
```
wardi, line 970: These sorts of asserts tell us more when they fail than the `len()` asserts
```suggestion
        assert [activity['activity_type'] for activity in activities] == ['changed package']
```

wardi:
Test changes have a lot of repeated `auth_user_obj = model.User.get(user["id"])` calls. I see that `logic.check_access` only sets the `auth_user_obj` when `ignore_auth` is False, and `helpers.call_action` sets `ignore_auth` to true, so the actions that expect `auth_user_obj` set in the context would fail without it being passed.

I think the way `check_access` is written doesn't make sense. Like the docstring says, it should populate `auth_user_obj` in the context automatically regardless of `ignore_auth` setting. Would you make that change and try removing all the extra `auth_user_obj` code from your test changes?

ckan/migration/versions/105_21d03ed1e1ca_add_permission_labels_in_activity_table.py
smotornyuk: Let's keep activities separately, because it doesn't add any complexity here.

 @vgeorgieva-cstbg, when creating migrations, use `ckan generate migration -p activity -m "Add permission labels in activity table"` command. It will create `ckanext/activity/migration` folder and put all the activity-related migrations there.

In order to apply such migrations you can use either `ckan db upgrade -p activity` to apply migrations from the specific plugin, or `ckan db pending-migrations --apply` if you want to apply all the migrations(that are not applied yet) from all the enabled plugins.

PS you can revert migration as well. use `ckan db downgrade -p activity` if you want to cancel all the activities from the given plugin. Or `ckan db downgrade -p activity -v -1`  if you want to revert only the latest activity(use `-v -2` for two latest activities, etc)

vgeorgieva-cstbg: @wardi @smotornyuk I moved the migration, but now it doesn't get executed automatically when the tests are ran. What's the best way to fix this?

vgeorgieva-cstbg: @wardi @smotornyuk I moved the migration, but now it doesn't get executed automatically when the tests are ran. What's the best way to fix this?

wardi: @vgeorgieva-cstbg Can you have the migration run as part of the extension's test setup? I don't know if there's a module level test setup or if you would have to group all tests into a class and use a class setup method.

vgeorgieva-cstbg: @wardi Fixed the tests. Anything else left to do?

ckanext/activity/tests/conftest.py
```diff
@@ -13,3 +20,24 @@ class ActivityFactory(CKANFactory):
...
+def apply_activity_migrations():
+    plugin = "activity"
+
+    factories.fake.unique.clear()
+    helpers.reset_db()
+
+    if not ckan.plugins.plugin_loaded(plugin):
+        ckan.plugins.load(plugin)
+
+    _run_migrations(plugin, version="head", forward=True)
```
wardi, line 35: running migrations as a side effect of importing a module, even if only for tests, is very surprising behavior. Can't you have the migrations run when a test that needs them is run instead?

vgeorgieva-cstbg: @wardi I changed it to only running the migrations in test classes that need them. Is it okay now?

smotornyuk: The most pytest-like approach here is to define [session-scoped reset_db_once and function scoped clean_db](https://github.com/ckan/ckanext-editable-config/blob/master/ckanext/editable_config/tests/conftest.py#L23-L32) fixtures instead of creating a new autouse-fixture that does such low-level things.

In pytest philosophy, the fixture should always do whatever makes sense for the current situation. for example, in CKAN core tests `clean_db` ensures that all the tables that CKAN uses are created and empty.

If we use `clean_db` during activity tests, it should as well guarantee that **all tables are created and empty**. But because we are inside a plugin that defines additional tables(or migrations), the definition of "all tables" is a bit different. Not it's all the tables that are affected by CKAN core migrations + all the tables that are affected by `activity` migrations. That's why I suggest registering activity-specific `clean_db`:
```python
@pytest.fixture()
def clean_db(reset_db, migrate_db_for):
    reset_db()
    migrate_db_for("activity")
```

Do not load `activity` plugin if it's missing. Fixtures should not be too smart. It's up to the test to include the required plugin alongside with `with_plugin` fixture. I.e, a test that interacts with DB inside the activity plugin will look like this:
```python
@pytest.mark.ckan_config("ckan.plugins", "activity")
@pytest.mark.usefixtures("with_plugins", "clean_db")
class TestActivity: ...
```
You may think that it's the wrong approach. What if the author of the test includes `clean_db` and forgets to enable the `activity` plugin? As you may have guessed, it will cause an error. But it's an expected and obvious outcome. If you interact with DB but forgot to include `clean_db`, you also can get an error. If you call datastore action but forgot to load the corresponding plugin - you'll get an error. Sometimes test authors **have to** understand the structure and do explicitly extra work. Fixtures should simplify our life, but they should never do anything that has no relation to their names.

Of course, you can load the plugin, as you do at the moment. But it creates more problems:
* Should this plugin be unloaded after the test? Even if it was enabled before? Wouldn't it be better to leave environment in the exactly the same state it was before?
* How can we test behavior when someone runs migrations without the plugin if fixture that runs migrations loads the plugin unconditionally?


`reset_db_once` can be defined in order to modify `non_clean_db`. You can redefine `non_clean_db` directly, but it's more complex, so overriding `reset_db_once` is just a simple workaround

vgeorgieva-cstbg: @smotornyuk  Thanks for the explanation.

The `clean_db` tests work well now, but the `non_clean_db` ones fail with `Plugin 'activity' cannot be loaded.` Any idea why? The same error was showing with `clean_db`, but was fixed after I changed the order of `clean_db` and `with_plugins` in the `usefixtures` mark, it's not working for `non_clean_db` though.

Also some tests from other modules (ex. `ckan/tests/lib/dictization/test_dictization.py`) fail because of the missing activity migration, but they don't activate the activity plugin. I was wondering what to do with those?

vgeorgieva-cstbg: @smotornyuk Thanks for the explanation.

The clean_db tests work well now, but the non_clean_db ones fail with Plugin 'activity' cannot be loaded. Any idea why? The same error was showing with clean_db, but was fixed after I changed the order of clean_db and with_plugins in the usefixtures mark, it's not working for non_clean_db though.

Also, some tests from other modules (ex. ckan/tests/lib/dictization/test_dictization.py) fail because of the missing activity migration, but they don't activate the activity plugin. I was wondering what to do with those?

vgeorgieva-cstbg: @smotornyuk Thanks for the explanation.

The `clean_db` tests work well now, but the `non_clean_db` ones fail with `Plugin 'activity' cannot be loaded.` Any idea why? The same error was showing with `clean_db`, but was fixed after I changed the order of `clean_db` and `with_plugins` in the `usefixtures` mark, it's not working for `non_clean_db` though.

Also, some tests from other modules (ex. `ckan/tests/lib/dictization/test_dictization.py`) fail because of the missing activity migration, but they don't activate the activity plugin. I was wondering what to do with those?

vgeorgieva-cstbg: @smotornyuk Thanks for the explanation.

The `clean_db` tests work well now, but the `non_clean_db` ones fail with `Plugin 'activity' cannot be loaded.` Any idea why? The same error was showing with `clean_db`, but was fixed after I changed the order of `clean_db` and `with_plugins` in the `usefixtures` mark, it's not working for `non_clean_db` though.

smotornyuk: My bad, I forgot about the order of fixture loading. In this scenario we cannot use `non_clean_db`, so switch to `clean_db` everywhere inside the `activity` plugin.

Explanation, if you are interested:

Every pytest fixture has a scope - it's the frequency of calling the fixture. By default, every fixture is called once for every separate test(`function` scope), but `reset_db_once` has a `session` scope instead, so it's called only once during the whole test session and, as a result, cleans DB only once, no matter how many times it's used.

In the initialization stage, pytest sorts fixtures by scope first and then by the order they are listed in the test. For example, in the following test:
```python
def test_smth(FUNCTION_x, SESSION_y, FUNCTION_z):
   ...
```
assuming the prefix reflects the scope of a fixture, pytest will call them in the following order:
* SESSION_y - because it has the highest scope
* FUNCTION_x - because among all fixtures with function scope, it's listed first in test arguments
* FUNCTION_z

This means that `reset_db_once` is always initialized first, before the function-scoped `with_plugins`. You cannot change the scope of `reset_db_once` to the `function`, because it will be called for every test then and essentially will turn into simple `clean_db`. Neither you can include `with_plugins` as a dependency of `reset_db_once`, because you are not allowed to include narrow-scoped fixtures into wide-scoped ones. 

So the only sensible option is to use `clean_db` instead of `non_clean_db` here. There are only a few tests that are using this fixture, so it won't create any visible downgrade in the execution time of tests

wardi:
@smotornyuk I'd really like to get this feature in. If you have time can you take a look at how we can have the activity migration run to fix the tests?

smotornyuk:
<details> 
<summary>Here's the patch with test fixes</summary>

```diff
diff --git a/ckan/lib/create_test_data.py b/ckan/lib/create_test_data.py
index d0174b0e7..0ce2927bf 100644
--- a/ckan/lib/create_test_data.py
+++ b/ckan/lib/create_test_data.py
@@ -471,12 +471,6 @@ left arrow <
             ])
         cls.user_refs.extend([u'tester', u'joeadmin', u'annafan', u'russianfan', u'testsysadmin'])
 
-        # Create activities for packages
-        for item in [pkg1, pkg2]:
-            from ckanext.activity.model import Activity
-            activity = Activity.activity_stream_item(item, 'new', 'not logged in')
-            model.Session.add(activity)
-
         model.repo.commit_and_remove()
 
     # method used in DGU and all good tests elsewhere
diff --git a/ckan/tests/controllers/test_user.py b/ckan/tests/controllers/test_user.py
index dc4714e03..c051aa211 100644
--- a/ckan/tests/controllers/test_user.py
+++ b/ckan/tests/controllers/test_user.py
@@ -171,8 +171,6 @@ class TestUser(object):
         )
         # assert "/user/activity" in response.headers["location"]
 
-    @pytest.mark.ckan_config("ckan.plugins", "activity")
-    @pytest.mark.usefixtures("with_plugins")
     def test_registered_user_login(self, app):
         """
         Registered user can submit valid login details at /user/login and
@@ -191,7 +189,7 @@ class TestUser(object):
             },
         )
         # the response is the user dashboard, right?
-        assert '<a href="/dashboard/">Dashboard</a>' in response
+        assert '<a href="/dashboard/datasets">Dashboard</a>' in response
         assert (
             '<span class="username">{0}</span>'.format(user["fullname"])
             in response
diff --git a/ckanext/activity/tests/conftest.py b/ckanext/activity/tests/conftest.py
index f7004f50b..480f78970 100644
--- a/ckanext/activity/tests/conftest.py
+++ b/ckanext/activity/tests/conftest.py
@@ -20,9 +20,3 @@ class ActivityFactory(CKANFactory):
 def clean_db(reset_db, migrate_db_for):
     reset_db()
     migrate_db_for("activity")
-
-
-@pytest.fixture(scope="session")
-def reset_db_once(reset_db, migrate_db_for):
-    reset_db()
-    migrate_db_for("activity")
diff --git a/ckanext/activity/tests/logic/test_action.py b/ckanext/activity/tests/logic/test_action.py
index a5de3fa71..ae52f0ba2 100644
--- a/ckanext/activity/tests/logic/test_action.py
+++ b/ckanext/activity/tests/logic/test_action.py
@@ -56,7 +56,7 @@ class TestLimits:
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestActivityShow:
     def test_simple_with_data(self, package, user, activity_factory):
         activity = activity_factory(
@@ -2485,7 +2485,7 @@ class TestSendEmailNotifications(object):
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestDashboardMarkActivitiesOld(object):
     def test_mark_as_old_some_activities_by_a_followed_user(self):
         # do some activity that will show up on user's dashboard
@@ -2546,7 +2546,7 @@ class TestDashboardMarkActivitiesOld(object):
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestFollow:
     @pytest.mark.usefixtures("app")
     def test_follow_dataset_no_activity(self):
diff --git a/ckanext/activity/tests/model/test_activity.py b/ckanext/activity/tests/model/test_activity.py
index ffafdf66a..c3efab5d5 100644
--- a/ckanext/activity/tests/model/test_activity.py
+++ b/ckanext/activity/tests/model/test_activity.py
@@ -13,7 +13,7 @@ from ckanext.activity.model import (
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestActivity(object):
     def test_include_data(self, package, user, activity_factory):
         activity = activity_factory(
diff --git a/ckanext/activity/tests/test_changes.py b/ckanext/activity/tests/test_changes.py
index df149dd8f..5ec8b8dbb 100644
--- a/ckanext/activity/tests/test_changes.py
+++ b/ckanext/activity/tests/test_changes.py
@@ -11,7 +11,8 @@ from ckan.tests import helpers
 from ckan.tests.factories import Dataset, Organization
 
 
-@pytest.mark.usefixtures("non_clean_db")
+@pytest.mark.ckan_config("ckan.plugins", "activity")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestChanges(object):
     def test_title(self):
         changes = []
diff --git a/ckanext/activity/tests/test_views.py b/ckanext/activity/tests/test_views.py
index 467c53bb2..c3a058dd9 100644
--- a/ckanext/activity/tests/test_views.py
+++ b/ckanext/activity/tests/test_views.py
@@ -21,7 +21,7 @@ def _clear_activities():
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestOrganization(object):
     def test_simple(self, app):
         """Checking the template shows the activity stream."""
@@ -174,7 +174,7 @@ class TestOrganization(object):
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestUser:
     def test_simple(self, app):
         """Checking the template shows the activity stream."""
@@ -909,7 +909,7 @@ class TestPackage:
 
 
 @pytest.mark.ckan_config("ckan.plugins", "activity")
-@pytest.mark.usefixtures("with_plugins", "non_clean_db")
+@pytest.mark.usefixtures("with_plugins", "clean_db")
 class TestGroup:
     def test_simple(self, app):
         """Checking the template shows the activity stream."""
```

</details>


There were 3 problems:

1. One of the login-tests expected the `activity` plugin to be enabled. Instead of enabling the plugin and running its migrations, I changed the test and now it does not rely on activities.
2. The `CreateTestData` class that we have historically used for test data creation was creating fake activities. As we are testing activities separately, inside the `activity` plugin, I removed fake activities from `CreateTestData`. Ideally, we should remove the whole `CreateTestData` class and use factories for data creation. But it's not urgent and requires a certain amount of effort, so I believe it's going to live for a while
3. `non_clean_db`. This fixture reuses existing DB in order to make tests faster. But inside the `activity` plugin this fixture depends on the test execution order. If one of the activity tests was executed first - `activity` migration was applied and all columns were present in the DB. But if the non-activity test runs first, the `permission_labels` column is missing from the existing DB and this breaks all consequent activity tests with `non_clean_db`. For now, I switched all activity tests to the `clean_db` fixture. It will force them to remove and recreate DB before every test(which makes `activity` DB tests ~30% slower), but it guarantees that all migrations are applied. I'll check if we can overcome this problem using different implementations of `non_clean_db` and add more details to the documentation.

wardi:
@smotornyuk thank you so much! @vikig02 would you be able to update this PR with these changes so we can get your feature merged?

darashev-daiteap:
Hello everyone, I'm sorry to inform you that the author of this pull request, @vikig02 , has left our company and is no longer available to work on this feature. I will take over the responsibility of implementing the changes requested by the reviewers and addressing any issues that may arise. I have created a new pull request [https://github.com/ckan/ckan/pull/7885](https://github.com/ckan/ckan/pull/7885). I will include the suggested changes from @smotornyuk in the PR.

darashevcstbg:
Hello everyone, I'm sorry to inform you that the author of this pull request, @vikig02 , has left our company and is no longer available to work on this feature. I will take over the responsibility of implementing the changes requested by the reviewers and addressing any issues that may arise. I have created a new pull request https://github.com/ckan/ckan/pull/7885. I will include the suggested changes from @smotornyuk in the PR.

PR #7535, opened 2023-04-11
Upgrade requirements April 2023

amercader:


amercader:
Flask was upgraded to 2.3.x, but that drops support for python 3.7 so wanted to check first what we want out policy to be for CKAN 2.11 before upgrading

requirements.txt
```diff
@@ -37,16 +37,20 @@ flask==2.2.3
...
     #   flask-wtf
-flask-babel==3.0.1
+flask-babel==3.1.0
     # via -r requirements.in
 flask-login==0.6.2
     # via -r requirements.in
-flask-wtf==1.0.1
+flask-wtf==1.1.1
     # via -r requirements.in
 greenlet==1.1.2
```
smotornyuk, line 46: If we update greenlet to v2.0.2, we can add Python v3.11 support. Do you want to give it a try? 
[SQLAlchemy should work with it](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_47/setup.cfg#L45)

PR #7537, opened 2023-04-11
Update extension template adding configuration for popular tools and git-hooks

smotornyuk:
An update that encourages developers to write standard code and publish packages.

Changes:

* Update test workflow(GitHub action). Now it aims CKAN v2.9 and v2.10
* **Suggest** using git-hooks for code formatting and lining. I.e, they are not enabled by default, but README explains how to enable it 
* Add `pyproject.toml` with simple configuration for popular tools: pyright, pytest, black, isort
* Add configuration for basic hooks: isort, black, ruff, secrets detection
* **Suggest** using [Conventional Commit messages](https://www.conventionalcommits.org/en/v1.0.0/) and compiling CHANGELOG when new version released.
* add `yaml` and `toml` extensions to the package's manifest, so that config declarations or similar config files are published to PyPI.
* Recommend using `build` + `twine` instead of `setuptools` + `wheel` + `twine` for publishing(recommended approach for setuptools)
* Small README fixes(padding for code snippets, badge for test workflow` 
* Fix `tests/test_plugin.py` example - previously it had missing imports
* Fresh extensions are `black`-compatible out of the box.

### DIFF
--- a/contrib/alembic/generic/alembic.ini.mako
+++ b/contrib/alembic/generic/alembic.ini.mako
@@ -35,6 +35,7 @@ script_location = %(here)s
 # are written from script.py.mako
 # output_encoding = utf-8
 
+# pragma: allowlist nextline secret
 sqlalchemy.url = driver://user:pass@localhost/dbname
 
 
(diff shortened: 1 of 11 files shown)


amercader:
@smotornyuk I refreshed this and add a small change in the workflow file to not support multiple CKAN versions by default as I think in most cases users just want to target one. The multiple versions setup is commented and easily enabled if needed.
If you are happy with this I'll merge it

PR #7540, opened 2023-04-13
Allow scheduling jobs

avdata99:
### Proposed feature:

The `RQ` library we use for CKAN jobs allows to schedule tasks: https://python-rq.org/docs/scheduling/
We are not implementing this feature in CKAN
It will be a good idea?
I didn't test already but this is something I need for an extension and maybe it could be added to CKAN instead using external tools

### Features:

- [ ] includes API changes


avdata99:
Cleaning, too old

PR #7541, opened 2023-04-13
Typing updates: April 2023

smotornyuk:
Latest typing updates. Apart from obvious fixes:

* `ckan/lib/cli.py`: ignore types here completely. It's an activity migration script written for py2. It's kept just for reference and it can be even removed if somebody bothers to create a PR
* Removed a couple of senseless typechecks, like `if not isinstance(request_data, dict):` in `ckan/views/api.py`(impossible due to [this check](https://github.com/ckan/ckan/blob/master/ckan/views/api.py#L202))

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -797,10 +797,10 @@ def _preprocess_dom_attrs(attrs: dict[str, Any]) -> dict[str, Any]:
 
 
 @core_helper
-def link_to(label: str, url: str, **attrs: Any) -> Markup:
+def link_to(label: Optional[str], url: str, **attrs: Any) -> Markup:
     attrs = _preprocess_dom_attrs(attrs)
     attrs['href'] = url
-    if label == '':
+    if label == '' or label is None:
         label = url
     return literal(str(dom_tags.a(label, **attrs)))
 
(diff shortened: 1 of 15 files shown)


ckan/lib/helpers.py
```diff
@@ -800,7 +800,7 @@ def _preprocess_dom_attrs(attrs: dict[str, Any]) -> dict[str, Any]:
 def link_to(label: str, url: str, **attrs: Any) -> Markup:
     attrs = _preprocess_dom_attrs(attrs)
     attrs['href'] = url
-    if label == '' or label is None:
+    if label == '':
```
wardi, line 803: helpers could be passed anything from jinja2 including 'none', can't they?

wardi:
everything else LGTM

ckan/lib/helpers.py
smotornyuk: Yes, types do not guarantee the actual type. But the same rules apply to `url`. If we are not checking that `url` is a string, why would we do it for `label`?

Taking into account the fact that `label` is a mandatory argument and comes first, there are only 3 ways to get `None` here:

* `None` is passed intentionally: I'd say it's incorrect usage of the function, but since it has no documentation, passing `None` is completely legit.
* This helper is used in the following manner - `h.link_to(PRODUCE_LABEL(), URL)` and `PRODUCE_LABEL` function returns `None`. Same as above
* It called as `h.link_to(LABEL_VAR, url)` and `LABEL_VAR` is undefined. It probably means that there is a problem with application logic.


More strict behavior sounds like an improvement for all 3 cases above. But I don't mind reverting this line, just to be more compatible with the existing ecosystem. It would be much better to create some kind of linter that checks templates and reports misuse of helpers, before making such changes.

PR #7544, opened 2023-04-13
Deprecate old site_read authz function.

pdelboca:
`site_read` function always returns `True` and has a deprecation note. So I'm deprecating it.


avdata99:
image

Issue #5865, opened 2021-01-29
Custom activities improvements on CKAN 2.9


amercader:
The changes around activities introduced in CKAN 2.9 were a massive improvement from previous versions. Up until that point you had to do a fair amount of monkey-patching core functionality and piggyback on "changed package" activities in order to get custom ones.

In CKAN 2.9 it should be as easy as creating your own activities and providing a matching template snippet for the activity type. And I say it should because we are not quite there.

## TL;DR

If you want to create custom activities right now on CKAN 2.9 (2.9.1) you need to 
1. Monkey patch some validators:

          core_validators.object_id_validators['lock dataset'] = core_validators.package_id_exists

2. Create your activities with a specific structure in the `data` object:

        activity_dict = {
            "activity_type": "lock dataset",
            "user_id": current_user.id,
            "object_id": data_dict["id"],
            "data": {
                "package": {
                    "title": dataset_title,
                }
            },
        }
3. On custom dataset types the generated links in the activity snippets are broken (use `/dataset`)

## Background

Let's assume we want to create custom activities for editors "locking" and "unlocking" a dataset. The actual logic around this is not relevant so let's simplify and just try to create a "lock dataset" activity after an update:

```python
from ckan import model
import ckan.plugins as plugins
import ckan.plugins.toolkit as toolkit


class TestextPlugin(plugins.SingletonPlugin):

    plugins.implements(plugins.IPackageController, inherit=True)

    # IPackageController

    def after_update(self, context, data_dict):

        current_user = model.User.get(context["user"])

        activity_dict = {
            "activity_type": "lock dataset",
            "user_id": current_user.id,
            "object_id": data_dict["id"],
            "data": {
                "name": data_dict["name"],
                "id": data_dict["id"],
            },
        }

        toolkit.get_action("activity_create")(context, activity_dict)

```

And we also have a snippet ready in our extension, in `templates/snippets/activities/lock_dataset.html`. This is a simplified version of core's `change_package.html` one. To simplify let's remove all the "View this version" / "Changes" links:

```
<li class="item {{ activity.activity_type|replace(' ', '-')|lower }}">
  <i class="fa icon fa-lock"></i>
  <p>
  {{ _('{actor} has locked the dataset {dataset}').format(
      actor=ah.actor(activity),
      dataset=ah.dataset(activity)
    )|safe }}
    <br />
    <span class="date">
      {{ h.time_ago_from_timestamp(activity.timestamp) }}
    </span>
  </p>
</li>
```

[...]

PR #7547, opened 2023-04-14
Automatically close inactive issues/PRs

smotornyuk:
A lot of issues have no sense anymore. I noticed that people sometimes close them manually, leaving comments like "This issue is closed due to inactivity". Let's do it automatically.

Time limits for PRs and issues can be configured separately. I set the same values:
* If an issue has had no comments for 2 years, it's marked as inactive.
* If an issue has no comments for 30 days after it's marked as inactive, it's closed

### DIFF
--- a/.github/workflows/close-inactive-issues.yml
+++ b/.github/workflows/close-inactive-issues.yml
@@ -1,0 +1,24 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "1 1 1 * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 730
+          days-before-issue-close: 30
+          stale-issue-label: "Inactive?"
+          stale-issue-message: "This issue is stale because it has been open for 2 years with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 30 days since being marked as stale."
+          days-before-pr-stale: 730
+          days-before-pr-close: 30
+          stale-pr-message: "This PR is stale because it has been open for 2 years with no activity."
+          close-pr-message: "This PR was closed because it has been inactive for 30 days since being marked as stale."
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


tino097:
I think 2 yesars for an issue to get inactive is to long. Maybe we can set it on few months?

smotornyuk:
I have no preference here, just checked the stats and picked a time period that leaves most of the relevant issues open. Some details to help with the decision:

Issue distribution by date of last activity(comment, new commit, issue assigned to someone, label added, etc.):
image

Cumulative number of issues by date:
image

pdelboca:
I also think that 2 years is a lot of time. 

We should be honest with our time and priorities. Sadly (and also not!) we receive more contributions than we can handle and it's okay to politely let some things go. If a ticket has 2/3 months without any activity is clearly not a priority or a "nice to have if we had the time for it".

smotornyuk:
What is the purpose of closing PRs if we don't have time to handle it? Just to show nice small numbers against Issue/PRs counters? 

2 years is a bound after which issues\PRs are about CKAN 2.8, which is not supported. That's what should control the lifetime of issues. If we are going to close issues and PRs after 3 months, we have to make enough releases during these 3 months, so that closed PRs are no longer relevant.

Otherwise, we'll be lying to the public, closing these PRs. It will cause resentment among users and contributors. 

If I have a lot issue\PR assigned to me and I don't have enough time, there are multiple options:
* let it be. If I have time, eventually I'll check this issue. If it will be closed because it becomes irrelevant, it's ok 
* If I'm bothered by a long list of issues I always can ask other team members to help me and take some issues.
* If this issue is just too boring and really simple, I can assign a "Good for contribution" label and give someone else an opportunity to become a CKAN contributor. We don't have to solve issues assigned to us, we can curate them and play the role of mentor.


So I can't see the value of closing potentially useful issues\PRs just because we don't have time to work on them. As long as they don't do any harm they have the right to exist

pdelboca:
You have a good point, it is true that it doesn't hurt. 

For me is just a matter of keeping tidy the workspace for better focus. I don't see a point on leaving something open for 2 years just because sometime someone can suddenly have time to review it. For me it is okay to just let things go given that the context has changed or because we don't have time. 

One think I consider negative and that hurts is that the more the information, and the more the open PR are the more difficult is to navigate through the priorities and to review them.

I guess at this point is just a matter of personal preference on how we like to tidy our working spaces.

I have no hard opinion on this topic and if everyone is okay with having hundreds of PRs open then we can just leave them in peace and harmony :)

smotornyuk:
I neither have a strong opinion as to how long we must keep the issue in an active state, but it doesn't feel right to close something without a direct reason. 

If we are going to close inactive issues\PRs after 3 months, it may have sense to add a new section to our meetings. 
I mentioned in PR that issues are not closed immediately. The receive `Inactive?` label first. And after another 30 days, they are closed.
When an issue receives a label, it's "updated" so it will appear at the top of the issue list during tech team meetings (because we are using the "most recently updated" order). We can review this labeled issue and make a decision:

* if the assignee doesn't have time to work on it and somebody else volunteers to take the issue(not like we are doing on tech meetings when someone **has to** volunteer), the label is removed and the issue is reassigned 
* If nobody volunteers, the issue is unassigned, and the "Good for contribution" label is added(inactive label kept) with an optional comment describing the reasons, why this issue is ignored. If nobody is interested in this issue, it will be closed automatically in a month. This point is a game changer. If the author of the PR\issue is interested in the fix, he can add a comment or update the PR. If the author has already forgotten about the issue, he won't be frustrated when it's closed.
* if the issue is no longer relevant(describes a bug from a previous CKAN version) it's closed immediately(or left as is and closed in a month automatically)


It solves two problems:

* contributors don't feel that all their efforts go to `/dev/null` because we transparently doing as much as we can to process their contributions.
* When issues are assigned during tech team meetings, obviously, there are cases, when no one has time to handle the issue, but someone has to. If the issue got assigned to the "wrong" person, who won't have time next 3 months, it still has a chance to be reassigned to a person with the amount of work.


It doesn't solve the problem you mentioned about the complexity of managing priorities. But here you are probably fighting with consequences, not with the reason. If it's hard to go through a long list of similar issues, closing them is not the appropriate solution.
Consider adding an additional label when an issue is assigned to you, like "High/Medium/Low priority", or "Docs/Tests/API/Frontend". In this way, you will be able to reduce the list to just a few relevant issues. We have 129 labels already, I'm sure you'll find something suitable. Or you can always create a new label.

wardi:
@smotornyuk can we add the rockstar tickets label to the ignore list too?

smotornyuk:
Filter by the project is not supported at the moment. But it looks like almost all issues from the rockstar board have "Good for Contribution" so including this label in the exempt list should solve the problem. And I added a "Long-standing" label to [the only issue](https://github.com/ckan/ckan/issues/5865) from the rockstar board that does not have "Good for Contribution"

PR #7554, opened 2023-04-18
[PYDANTIC]

TomeCirun:
Fixes #

A modern replacement to our schema/navl validation model

I will provide more details soon...

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/plugins.py
+++ b/ckan/lib/plugins.py
@@ -9,10 +9,10 @@ from typing import Any, Optional, TYPE_CHECKING, TypeVar, cast
 from flask import Blueprint
 
 import ckan.logic.schema as schema
-from ckan.common import g
+from ckan.common import g, config
 from ckan import logic, model, plugins
 import ckan.authz
-from ckan.types import Context, DataDict, Schema
+from ckan.types import Context, DataDict, Schema, PydanticModel
 from . import signals
 from .navl.dictization_functions import validate
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


smotornyuk:
Previously, when I was working with pydantic, it didn't have one important feature that is required by CKAN(or I just didn't find it). Can you confirm if the scenario below is possible with pydantic.

At the moment, schemas are dynamic. Let's imagine an `x` action, that has the following schema:
```python
{
  "a": [not_missing],
  "b": [default(0)]
}
```

Is it possible for the plugin PLUGIN_ONE to chain this action, changing its schema:
```python
@chained_action
def x(...):
   context["schema"]["a"].append(dict_only)
   ...
```
At the same time,  PLUGIN_TWO can chain this action, changing its schema in a different way:
```python
@chained_action
def x(...):
   context["schema"]["custom_field"] = [convert_to_extras]
   ...
```

In the end, we have a schema that contains all the modifications:
```python
{
  "a": [not_missing, dict_only],
  "b": [default(0)],
  "custom_field": [convert_to_extras],
}
```


--- 

As far as I remember, pydantic doesn't allow to patch schema. We can use new schema that replaces an old one. But for plugins, it means that either `class One(Default): ...` schema or `class Two(Default): ...` schema is used. Because plugins cannot patch schema and the only option is to replace the schema, our scenario turns into:
```
default schema created
plugin one replaces the default schema with its schema
plugin two replaces one's schema with two's schema
```

In this case, we are losing all the customizations from plugin *one*. Does pydantic have a solution for this case?

TomeCirun:
Hey @smotornyuk , tnx for this, I didn't think about such a problem before, though, I think I found a solution but didn't have a chance to test it with multiple plugins that have chained actions, could you help with the tests?

FYI `__init_subclass__` serves so we can add validators directly in the fields as I did [here](https://github.com/ckan/ckan/pull/7554/commits/216806dc8a46b64e9500e2883ac835664f003a1b#diff-5dab9bc0468c48c4e7e38b148a0674b693eab7017e590dd6da73feeb8f4070d0R117)

Thanks

TomeCirun:
Also, we can achieve this in a slightly different way, maybe better:

```python
class UserCreateSchema(BaseModel):
    name: str
    age: int
    email: str
    address: str

    _validators = {
        'name': [not_empty_validator, length_validator(min_length=2)],
        'age': [lambda v: v >= 18],
        'email': [not_empty_validator],
        'address': [length_validator(min_length=10)]
    }

    @validator('*', pre=True, always=True)
    def validate_fields(cls, v, field):
        extra_validators = cls._validators.get(field.name)
        if extra_validators:
            for validator_func in extra_validators:
                v = validator_func(v)
        return v

     @classmethod
    def add_validator(cls, field_name, validator_func):
        if field_name in cls.__fields__:
            cls.__fields__[field_name].validators.insert(0, validator_func)
        else:
            raise ValueError(f"{field_name} not found in UserCreateSchema")


# add a new validator to an existing field
UserCreateSchema.add_validator('name', length_validator(min_length=3))

# add a new field with validators
UserCreateSchema.__fields__['phone'] = {'type': str, 'validators': [not_empty_validator]}
```

@smotornyuk

PR #7574, opened 2023-05-02
Datatables View Ext: Null / NoneType Front-End Label Option & Helper

JVickery-TBS:
feat(frontend): added null label;

- Added new config option `ckan.null_label`.
- Added new helper `get_null_label`.
- Used `get_null_label` in datatables view.

Fixes #

Added a new config option and helper to set labels for NoneType (or null from sql) string labels. This is to be used to display NoneType values on the front-end.

The new config option also supports a dict for i18n labels.

Included is its use in the datatableview plugin to display null datastore values with the label instead of it casting NoneType to the string of `"None"`.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7574.feature
+++ b/changes/7574.feature
@@ -1,0 +1,3 @@
+Added `ckan.null_label` config option.
+
+Added `get_null_label` helper.
--- a/changes/7574.feature
+++ b/changes/7574.feature
@@ -1,3 +1,6 @@
-Added `ckan.null_label` config option.
+Added `ckan.datatables.null_label` config option.
 
-Added `get_null_label` helper.
+Added `datatablesview_null_label` helper.
+
+Datatables Views will now show blank cells
+for NoneType field values by default.
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1715,15 +1715,3 @@ groups:
         type: int
         default: 180
         description: The option defines the timeout in seconds until giving up on a job
-
-
-  - annotation: NoneType Display Settings
-    options:
-      - key: ckan.null_label
-        description: |
-          The option defines the label used to display NoneType values for the front-end.
-          This can be a string or a dictionary for locale strings.
-          example::
-
-            ckan.null_label = "N/A"
-            ckan.null_label = {"en": "N/A", "fr": "Aucun"}
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1715,3 +1715,15 @@ groups:
         type: int
         default: 180
         description: The option defines the timeout in seconds until giving up on a job
+
+
+  - annotation: NoneType Display Settings
+    options:
+      - key: ckan.null_label
+        description: |
+          The option defines the label used to display NoneType values for the front-end.
+          This can be a string or a dictionary for locale strings.
+          example::
+
+            ckan.null_label = "N/A"
+            ckan.null_label = {"en": "N/A", "fr": "Aucun"}
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2832,3 +2832,24 @@ def make_login_url(
 @core_helper
 def csrf_input():
     return snippet('snippets/csrf_input.html')
+
+
+@core_helper
+def get_null_label() -> Union[str, None]:
+    _auto_flask_context = _get_auto_flask_context()
+
+    if _auto_flask_context:
+        _auto_flask_context.push()
+
+    null_label = config.get('ckan.null_label', None)
+    
+    if null_label is not None:
+        try:
+            null_label = json.loads(null_label)
+            if lang() in null_label:
+                return null_label[lang()]
+        except ValueError as e:
+            if '{' not in null_label:
+                return null_label
+            raise ValueError(e)
+    return None
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2832,23 +2832,3 @@ def make_login_url(
 @core_helper
 def csrf_input():
     return snippet('snippets/csrf_input.html')
-
-
-@core_helper
-def get_null_label() -> str:
-    _auto_flask_context = _get_auto_flask_context()
-
-    if _auto_flask_context:
-        _auto_flask_context.push()
-
-    null_label = config.get('ckan.null_label', None)
-    if null_label is not None:
-        try:
-            null_label = json.loads(null_label)
-            if lang() in null_label:
-                return null_label[lang()]
-        except ValueError as e:
-            if '{' not in null_label:
-                return null_label
-            raise ValueError(e)
-    return ''
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2835,14 +2835,13 @@ def csrf_input():
 
 
 @core_helper
-def get_null_label() -> Union[str, None]:
+def get_null_label() -> str:
     _auto_flask_context = _get_auto_flask_context()
 
     if _auto_flask_context:
         _auto_flask_context.push()
 
     null_label = config.get('ckan.null_label', None)
-    
     if null_label is not None:
         try:
             null_label = json.loads(null_label)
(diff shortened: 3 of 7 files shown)


ckan/config/config_declaration.yaml
```diff
@@ -1715,3 +1715,15 @@ groups:
         type: int
         default: 180
         description: The option defines the timeout in seconds until giving up on a job
+
+
+  - annotation: NoneType Display Settings
+    options:
+      - key: ckan.null_label
```
wardi, line 1722: this is a datatables configuration option, would you move it to ckanext/datatables/config_declaration.yaml and update the name?

ckanext/datatablesview/blueprint.py
```diff
@@ -122,8 +122,9 @@ def ajax(resource_view_id: str):
         dtdata = {u'error': query_error}
     else:
         data = []
+        null_label = h.get_null_label()
         for row in response[u'records']:
-            record = {colname: str(row.get(colname, u''))
+            record = {colname: str(row.get(colname, u'') or null_label)
```
wardi, line 127: this will convert any falsey value to the null_label (e.g. a false boolean column, an empty list and the number 0) use an `is None` check instead.

ckan/lib/helpers.py
```diff
@@ -2832,3 +2832,23 @@ def make_login_url(
...
+    null_label = config.get('ckan.null_label', None)
+    if null_label is not None:
+        try:
+            null_label = json.loads(null_label)
+            if lang() in null_label:
+                return null_label[lang()]
+        except ValueError as e:
+            if '{' not in null_label:
+                return null_label
+            raise ValueError(e)
```
wardi, line 2853: we can define validation on configuration options to catch errors at startup. Template helpers shouldn't raise any exceptions because there's no try/except in jinja2.

ckan/lib/helpers.py
```diff
@@ -2832,3 +2832,23 @@ def make_login_url(
...
+
+    if _auto_flask_context:
+        _auto_flask_context.push()
+
+    null_label = config.get('ckan.null_label', None)
+    if null_label is not None:
+        try:
+            null_label = json.loads(null_label)
+            if lang() in null_label:
+                return null_label[lang()]
```
wardi, line 2849: this can fail with a KeyError. Maybe supporting translated text in the default helper is more complexity than we need for this feature.

ckan/config/config_declaration.yaml
JVickery-TBS: moved to `datatablesview` plugin.

ckanext/datatablesview/blueprint.py
JVickery-TBS: Updated with `None` condition.

ckan/lib/helpers.py
JVickery-TBS: N/A, using simple string now.

ckan/lib/helpers.py
JVickery-TBS: No longer supporting json string for the option. Users can translate in po/mo files if they want. (put the option value through `toolkit._` now).

JVickery-TBS:
@wardi okay the code has been moved to the datatablesview extension exclusively. 

The config option name has been updated, and the declaration moved.

The helper has also been moved into datatablesview ext.

The config option has been simplified, so it will just be a string and the helper will now just return the value of the option, put through gettext.

If the option is not set, the helper will just return a blank string.

changes/7574.feature
```diff
@@ -0,0 +1,3 @@
+Added `ckan.null_label` config option.
```
wardi, line 1: update config parameter name?

ckanext/datatablesview/blueprint.py
```diff
@@ -122,8 +122,17 @@ def ajax(resource_view_id: str):
...
     else:
         data = []
+        null_label = h.get_null_label()
+        from pprint import pprint
+        from logging import getLogger
+        log = getLogger(__name__)
+        log.info("    ")
+        log.info("DEBUGGING::")
+        log.info(pprint(null_label))
+        log.info("    ")
```
wardi, line 132: remove debugging?

ckanext/datatablesview/plugin.py
```diff
@@ -96,3 +98,10 @@ def info(self) -> dict[str, Any]:
                 u'filterable': [default(True), boolean_validator],
             }
         }
+
+    # ITemplateHelpers
+
+    def get_helpers(self) -> dict[str, Callable[..., object]]:
+        return {'get_null_label': lambda: toolkit._(self.null_label)
```
wardi, line 105: how about `datatablesview_null_label` or similar? template helpers share a common namespace.

Also using a lambda here is no shorter than defining a normal function, so I'd use a normal function instead.

JVickery-TBS: renamed to `datatablesview_null_label` and added helpers module instead of lambda function.

ckanext/datatablesview/config_declaration.yaml
```diff
@@ -93,3 +93,9 @@ groups:
       the user to view, copy and print the details of a specific row. This
       value can be overridden at the resource level when configuring a
       DataTables resource view.
+
+  - key: ckan.datatables.null_label
+    example: N/A
+    description: |
+      The option defines the label used to display NoneType values for the front-end.
+      This should be a string and can be translated via po files.
```
wardi, line 101: can we add the new default value here too?

JVickery-TBS: @wardi default value now added. I set it as `''` but should it just be blank?

Are we able to add `type: string` or `type: str` here too? I have not seen any other declaration file use type string?

smotornyuk: Yes, exactly. `type` defines some basic conversion rules. There is no such `type` as `string`, but if it were there, it would just add a `unicode_safe` or `unicode_only` validator. Such type(and these validators used by themselves, without any additional validators) have no practical sense inside config declaration. Validators that are applied in the following cases:

* When value read from config file. All the values initially are strings because of the internal logic of `configparser` library, which handles INI files. It means that `unicode_only` won't do much here
* When the value is missing from the config file, but declared `default` is available. If you expect the value to be a string, you should set the default to string. So here we don't need the validator as well.
* When the value is missing from the config file and the declaration has no `default`. In this case, we are using `None` as a default `default`(sounds awesome). I don't think that applying `unicode_*` validators to `None` is a great idea. The only case, when it may be useful - required options. You set no default and add a `unicode_only` validator. Then you'll see a validation error because `None` won't pass this check. But the much better way to handle it is using a `not_empty` validator, so `unicode_*` validators are not needed here as well


I thought about adding `str` for completeness. But if a type that doesn't do much exists, it has sense to add more useful types, like `dict`, `enum`, `float`, `set`, `timedelta`, `Path`, etc. So, for the first version, I decided to go with the minimal required set of types. And, when we get some feedback from the real world, we can define hundreds of types for every possible use case. But remember, that `type` is just a shortcut for `validators: ...` and don't overestimate its value

ℹ️ [here is the list of all available types](https://github.com/ckan/ckan/blob/master/ckan/config/declaration/load.py#L40-L46)

ckanext/datatablesview/plugin.py
```diff
@@ -22,6 +22,7 @@ class DataTablesView(p.SingletonPlugin):
     p.implements(p.IConfigurer, inherit=True)
     p.implements(p.IResourceView, inherit=True)
     p.implements(p.IBlueprint)
+    p.implements(p.ITemplateHelpers)
```
smotornyuk, line 25: BTW, instead of `implements(ITemplateHelpers)` and defining `get_helpers` you can use [`@tk.blanket.helpers`](https://docs.ckan.org/en/2.10/extensions/plugins-toolkit.html#ckan.plugins.toolkit.ckan.plugins.toolkit.blanket) class decorator. It will take all public functions defined in the `helpers` module and register them as plugin's helpers.

It's not a mandatory way, but it a nice solution for people, like you, who creates all the functions in the right place:)

ckanext/datatablesview/config_declaration.yaml
JVickery-TBS: Okay, so we do not need a type here in our declaration. Sounds good!

PR #7575, opened 2023-05-03
Fix Sphinx extension for Python 3.11 support

amercader:
We use an extension to generate the documentation for the toolkit methods, and that module used the `inspect.formatargspec()` method, which has been deprecated since py 3.5, and was dropped in py 3.11


kowh-ai (approved):
Recreated issue, applied fix, worked like a charm
Screenshot 2023-05-09 at 10 12 18 am

PR #7577, opened 2023-05-08
Update license file

amercader:
Remove section about SQLAlchemy license (AFAICT we are not vendoring any parts of it), clean up a bit


Zharktas:
Should webhelpers bit just refer here https://github.com/ckan/ckan/blob/b2afa91a8e57bdeac874c58b144256732577d9a2/ckan/lib/pagination.py#L4-L34 ?

Zharktas (approved):

PR #7583, opened 2023-05-11
SQLAlchemy 2.0

smotornyuk:
Prepare code for SQLAlchemy v2.0. 

We are using v1.4 right now, and the best part of it is the fact, that this version contains all the features that will be removed in v2.0(that's why we were able to start using it without big code changes) as well as **almost** all features that are available in v2.0. It means we can migrate the code staying on v1.4 and at some point in the future just change SQLAlchemy's version in the requirements.txt and everything will work like a charm(ideally).

Plan:
* [x] [before CKAN v2.11] Replace all deprecated features with the recommended for 2.0
* [ ] [release CKAN v2.11] Now the last two CKAN versions are using SQLAlchemy v1.4, which means that extension maintainers, who support only the two latest CKAN versions can update their extensions to be completely compatible with SQLAlchemy v1.4 and v2.0 at the same time.
* [ ] [after CKAN v2.11] Turn on exceptions for all deprecation warnings from SQLAlchemy. It will help us to identify anything we've missed during the first step
* [ ] [after CKAN v2.11] Write a guide for the extension maintainers
* [ ] [halfway between CKAN v2.11 and v2.12, ~3months after v2.11 release] Upgrade to SQLAlchemy v2.0

I expect that code changes won't be visible outside of CKAN and almost everything will be completely compatible with the existing extensions. At the moment I identified only one place, that requires breaking changes: the `where` key for a dictionary returned from the `IDatastore.datastore_search` method. SQLAlchemy v2.0 does not allow %-placeholders(`WHERE x = %s`). Instead it requires :-placeholders(`WHERE x = :x`). [Here you can find details from changelog entry](https://github.com/ckan/ckan/blob/2a802a4a50a731c966ee0c1f7836255d1127562c/changes/7583.misc)


ckan/migration/versions/093_7f70d7d15445_remove_activity_revision_id.py
```diff
@@ -51,9 +51,6 @@ def downgrade():
     op.add_column(u'system_info',
                   sa.Column(u'revision_id', sa.TEXT(), autoincrement=False,
                             nullable=True))
-    op.create_foreign_key(u'resource_view_resource_id_fkey', u'resource_view',
-                          u'resource', ['resource_id'], ['id'],
-                          onupdate=u'CASCADE', ondelete=u'CASCADE')
```
wardi, line 56: this just isn't needed?

ckan/model/types.py
```diff
@@ -43,6 +43,8 @@ class JsonType(types.TypeDecorator):  # type: ignore
     '''
     impl = types.UnicodeText
 
+    cache_ok = False
```
wardi, line 46: this seems like a separate bug-fix. Are we working around this issue in other parts of our code, like in the handling of plugin_extras fields?

ckan/templates/user/list.html
```diff
@@ -16,7 +16,7 @@ <h1 class="page-heading">
         <ul class="user-list">
           {% block users_list_inner %}
             {% for user in page.items %}
-              <li>{{ h.linked_user(user['name'], maxlength=20) }}</li>
+              <li>{{ h.linked_user(user[0].name, maxlength=20) }}</li>
```
wardi, line 19: what's going on here?

ckan/migration/versions/093_7f70d7d15445_remove_activity_revision_id.py
smotornyuk: Yes, when I tried to downgrade, I got a message, that this key already exist. So there is no sense to create it.

ckan/model/types.py
smotornyuk: It's not exactly a bug fix, I'd say. That's the default behavior of SQLAlchemy now. But if you don't specify `cache_ok` explicitly, SQLAlchemy will produce warnings(and errors in the future) and set this attribute to False.

This JsonType/JsonDictType is used only by activity.data, resource.extras, and resource_view.config columns. 
For other cases, like `plugin_extras` of user\package, we are using native JSON/JSONB fields. And it would be nice to remove this class completely and switch to the native type here as well, but it's a bigger change, that's why I only explicitly set `cache_ok = False` to suppress errors.

ckan/templates/user/list.html
smotornyuk: `Session.query(User, User.name, User.id)` returns an iterator of items, that behaves in a special way.

if you write a loop like this:
```python
for item in query:
   assert item[0] is user 
   assert item[1] is user_name
   assert item[2] is user_id
```
it behaves in the expected way, allowing you to extract separate parts of the query by position. This is not changed

But if you write the following:
```
for item in query:
   assert item.email is user_email
```
where this `email` property will be taken from? Not quite expected(for me), but sqlalchemy just transforms it like the following: `item.email -> item[0].email`. And you'll get an email from the first item that was queried, from the `User` object.

In 2.0 this behavior was changed. Now you have to explicitly specify the source of an attribute. i.e, `user.email` for the given query no longer allowed.


In this case, we are passing to templates not a  simple`query(User)` but something [much more monstrous](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L847-L861). That's why we have to specify the index of item, from which we are extracting the name.

ckanext/datastore/tests/test_db.py
```diff
@@ -120,7 +120,7 @@ def _assert_created_index_on(
             )
 
         calls = connection.execute.call_args_list
-        was_called = [call for call in calls if call[0][0].find(sql_str) != -1]
+        was_called = [call for call in calls if str(call[0][0]).find(sql_str) != -1]
```
wardi, line 123:
```suggestion
        was_called = any(sql_str in call[0][0] call in calls)
```

ckanext/example_idatastorebackend/example_sqlite.py
```diff
@@ -23,15 +28,14 @@ def _get_engine(self):
...
         if len(records):
             for record in records:
-                self._get_engine().execute(
-                    u'INSERT INTO "{0}"({1}) VALUES({2})'.format(
+                self.execute(
+                    'INSERT INTO "{0}"({1}) VALUES({2})'.format(
                         table,
                         u', '.join(record.keys()),
-                        u', '.join(['?'] * len(record.keys()))
+                        u', '.join([f":{key}" for key in record])
```
wardi, line 35: sql injection from user-provided column names?

ckanext/example_idatastorebackend/example_sqlite.py
```diff
@@ -43,28 +47,26 @@ def configure(self, config):
     def create(self, context, data_dict):
         columns = str(u', '.join(
             [e['id'] + u' text' for e in data_dict['fields']]))
-        engine = self._get_engine()
-        engine.execute(
-            u' CREATE TABLE IF NOT EXISTS "{name}"({columns});'.format(
+        self.execute(
+            'CREATE TABLE IF NOT EXISTS "{name}"({columns});'.format(
```
wardi, line 51: here too

ckan/migration/versions/093_7f70d7d15445_remove_activity_revision_id.py
wardi, line 56: no longer needed?

smotornyuk: I mentioned it above, this key exists when I'm trying to downgrade the DB version, so this line just prevents me from downgrading. And reverse function(upgrade) does not drop this index, so this line probably was added by mistake

ckanext/example_idatastorebackend/example_sqlite.py
smotornyuk: haven't paid much attention to this backend because it's just an example, but you are right. I'll update these two injection opportunities

ckanext/datastore/backend/postgres.py
```diff
@@ -370,24 +384,31 @@ def _validate_record(record: Any, num: int, field_names: Iterable[str]):
 
 def _where_clauses(
         data_dict: dict[str, Any], fields_types: dict[str, Any]
-) -> list[Any]:
+) -> WhereClauses:
     filters = data_dict.get('filters', {})
-    clauses: list[Any] = []
+    clauses: WhereClauses = []
+
+    idx_gen = iter(range(999_999))
```
wardi, line 391: itertools.count?

ckanext/example_idatastorebackend/example_sqlite.py
smotornyuk: PS I just love SQL constructor from SQLALchemy:
```python
sql = sa.insert(
  sa.table(table, *map(sa.column, record))
).values(record)
self.execute(sql)
```

ckanext/datastore/backend/postgres.py
smotornyuk: Yes, definitely nicer choice

ckanext/datastore/backend/postgres.py
```diff
@@ -1968,27 +2053,19 @@ def datastore_search(
...
+        with engine.begin() as conn:
+            context["connection"] = conn
             # check if table exists
             if 'filters' not in data_dict:
-                context['connection'].execute(
-                    u'DROP TABLE "{0}" CASCADE'.format(
-                        data_dict['resource_id'])
-                )
+                conn.execute(sa.text('DROP TABLE "{0}" CASCADE'.format(
+                    data_dict['resource_id']
```
wardi, line 2066:
```suggestion
                conn.execute(sa.text('DROP TABLE {0} CASCADE'.format(
                    identifier(data_dict['resource_id'])
```

ckanext/datastore/backend/postgres.py
```diff
@@ -2121,54 +2201,59 @@ def resource_fields(self, id: str) -> dict[str, Any]:
...
+            with engine.connect() as conn:
+                meta_results = conn.execute(meta_sql)
+            info['meta']['count'] = meta_results.one()[0]
 
             # table_type - BASE TABLE, VIEW, FOREIGN TABLE, MATVIEW
-            tabletype_sql = sqlalchemy.text(u'''
+            tabletype_sql = sa.text(u'''
                 SELECT table_type FROM INFORMATION_SCHEMA.TABLES
                 WHERE table_name = '{0}'
                 '''.format(id))
```
wardi, line 2217: use literal_string or normal parameter here instead?

ckanext/datastore/backend/postgres.py
```diff
@@ -1808,26 +1889,30 @@ def _read_connection_has_correct_privileges(self):
...
+                        )
+                    ))
+                    if have_privilege:
+                        return False
         finally:
-            write_connection.execute(drop_foo_sql)
-            write_connection.close()
+            with engine.begin() as conn:
+                conn.execute(drop_foo_sql)
+
```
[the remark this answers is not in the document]
smotornyuk: Don't think so. And now it's not possible to create a resource with such an id. But let me check it

ckan/templates/user/list.html
wardi: instead of passing the monstrous query object to the templates should we pass a simple list of dicts so that the template can remain unchanged and avoid the mysterious `[0].name`-style notation?

ckanext/datastore/backend/postgres.py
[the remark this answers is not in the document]
smotornyuk: Nope. Aliases created as views and `DROP TABLE` will cause error if you try to apply it to the SQL View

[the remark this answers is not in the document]
smotornyuk: PS and yes, it's possible to have `_foo` view and `_foo` temp table. Apparently, they are using different namespaces

ckan/templates/user/list.html
smotornyuk: Nope, it's not that simple.

On the `user.index` endpoint, we ask user_list for [a query object](https://github.com/ckan/ckan/blob/master/ckan/views/user.py#L90) instead of a list of dictionaries. It's already a bad decision, but I have an idea, why it's done in this way. [user_list does not support pagination](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L917-L925), and to avoid pulling and dictizing **all** users, the `user.index` takes the query and passes it to `Page` object. Which in turn can compute the number of pages and fetch only the required range of users`

Adding `limit`/`offset` to the `user_list` action won't help much. This action [returns list](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L925) and we won't know how many users are there, so the pagination widget will not work. It would be cool to have a result of `user_list` similar to `package_search`, with a `count` attribute. But it's a breaking change, so it's not an option for now. It's possible to solve this by making two requests: one returns a list of dictionaries restricted by new `limit`/`offset` parameters and one returns a query that we'll use to get a total number of items and build a pagination widget. But, as a result, we'll just switch from non-readable to non-efficient solution. 
So changing the API is a matter of further discussion and a separate PR.

Here, in order to preserve simplicity in the template, I can suggest using [`all_fields=False`](https://github.com/ckan/ckan/blob/master/ckan/logic/action/get.py#L917)(I already updated the PR). In this way, instead of `Query[tuple[User, str, str, str ...]]` we'll get the list of named tuples with `name` attribute in the template. I.e `user.name` will work just as it did before. As we are using only the name in the template in order to build a link to the user profile, it definitely solves the problem at the core.
The only problem that may appear: if any extension is actually relying on this monstrous object, it stops working. But I'm sure it's an acceptable risk because this structure is quite unusual for CKAN and relying on it means relying on CKAN's internal implementation details, which is never a good idea. And we still have username, so extensions should be able to get enough user details [just as we are doing in core template](https://github.com/ckan/ckan/blob/master/ckan/templates/user/list.html#L19)

ckan/model/types.py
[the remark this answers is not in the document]
smotornyuk: Sorry, just haven't noticed this comment among others initially. But now I had to fix conflicts after your tracking-middleware-extraction PR, and that's how I came back here:)

These cache keys are quite a complex matter, but fortunately, I can suggest a simple rule, how to decide if `cache_ok`. If you are implementing a new type over value, that is hashable(can be used as dictionary key), it's `cache_ok`. For example, UuidType is derived/contains just a single property - UUID of the record. UUIDs are hashable, so this type is `cache_ok = True`.
 
JsonType, on the contrary, contains a dictionary as a value. We cannot hash a dictionary, so it's `hash_of = False`


Alternatively, you can use string interpolation in order to decide if the type is `cache_ok`(but this option is much less reliable, so better not to even read this option). If there is a chance that `str(value)` produces a different result for values `a` and `b` that are equal(`a == b`), the value is not `cache_ok`. We can easily prove that JsonType that contains a dictionary is not `cache_ok` using the following code snippet:
```python
a = {1: 1, 2: 2}
b = {2: 2, 1: 1}

assert a == b
assert str(a) != str(b)
```

But keep in mind, that this doesn't work all the time. So "type's value can be a dictionary key" is more reliable strategy to determine the value of `cache_ok`

ckan/model/__init__.py
```diff
@@ -207,12 +208,18 @@ def init_db(self) -> None:
 
     def clean_db(self) -> None:
         self.commit_and_remove()
-        meta.metadata = MetaData(self.metadata.bind)
+        meta.metadata = MetaData()
+
+        if not meta.engine:
+            raise CkanConfigurationException("Model is not initialized")
```
pdelboca, line 214: Can we add a small tip on how to fix this error in the message? I would not know what to do or where to check if I get that message.

ckan/model/__init__.py
```diff
@@ -221,7 +228,10 @@ def create_db(self) -> None:
         i.e. the same as init_db APART from when running tests, when init_db
         has shortcuts.
         '''
-        self.metadata.create_all(bind=self.metadata.bind)
+        if not meta.engine:
+            raise CkanConfigurationException("Model is not initialized")
```
pdelboca, line 232: Same here. What should we do at this point if the Model is not initialized? What's the cause of it? Some pointers to the user would be great.

ckan/model/meta.py
```diff
@@ -79,9 +78,9 @@ def ckan_after_rollback(session: Any):
         del session._object_cache
 
 
-#mapper = Session.mapper
 mapper = orm.mapper
 
 # Global metadata. If you have multiple databases with overlapping table
 # names, you'll need a metadata for each database
```
pdelboca, line 84:
```suggestion
```


I think it is safe to remove this comment right? Are we handling multiple databases?

pdelboca:
Executing this branch with `harvest` plugin will cause an error:
```
  File "/home/pdelboca/Repos/ckan/.venv/lib/python3.10/site-packages/ckanext/harvest/plugin.py", line 279, in configure
    model_setup()
  File "/home/pdelboca/Repos/ckan/.venv/lib/python3.10/site-packages/ckanext/harvest/model/__init__.py", line 50, in setup
    if not model.package_table.exists():
  File "<string>", line 2, in exists
  File "/home/pdelboca/Repos/ckan/.venv/lib/python3.10/site-packages/sqlalchemy/util/deprecations.py", line 468, in warned
    return fn(*args, **kwargs)
  File "/home/pdelboca/Repos/ckan/.venv/lib/python3.10/site-packages/sqlalchemy/sql/schema.py", line 941, in exists
    bind = _bind_or_error(self)
  File "/home/pdelboca/Repos/ckan/.venv/lib/python3.10/site-packages/sqlalchemy/sql/base.py", line 1659, in _bind_or_error
    raise exc.UnboundExecutionError(msg)
sqlalchemy.exc.UnboundExecutionError: Table object 'package' is not bound to an Engine or Connection.  Execution can not proceed without a database to execute against.

```

Are we expecting everything to work? For your comment I suspect yes:

> I expect that code changes won't be visible outside of CKAN and almost everything will be completely compatible with the existing extensions.

So it would be nice to have some guidelines if we are gonna need to migrate or change extensions.

smotornyuk:
Some guidelines are already there - I added a changelog entry with all incompatible changes that I identified.
As for ckanext-harvest, it's a special case. It still defines and initializes tables everytime during the application startup instead of using [database migrations](https://docs.ckan.org/en/2.10/extensions/best-practices.html?highlight=migrations#use-migrations-when-introducing-new-models). I'll see what can be done, but instead of introducing an exception for plugins that relying on undocumented logic, I'll switch ckanext-harvest to alembic migrations.

smotornyuk:
BTW, @pdelboca , It's good that you identified this problem. Can you trigger your github workflow that tests popular extensions against this branch somehow?

ckan/templates/user/list.html
```diff
@@ -16,7 +16,8 @@ <h1 class="page-heading">
         <ul class="user-list">
           {% block users_list_inner %}
             {% for user in page.items %}
-              <li>{{ h.linked_user(user['name'], maxlength=20) }}</li>
+              {# `user` is a tuple with a single item - username #}
+              <li>{{ h.linked_user(user[0], maxlength=20) }}</li>
```
wardi, line 20: Would be better if switching sqlalchemy versions didn't affect templates this way, can we update the view to generate the old dict format instead of the template?

ckanext/datastore/backend/postgres.py
```diff
@@ -369,24 +386,31 @@ def _validate_record(record: Any, num: int, field_names: Iterable[str]):
...
-            clause_str = (u'"{0}" in ({1})'.format(field,
-                          ','.join(['%s'] * len(value))))
-            clause = (clause_str,) + tuple(value)
+            placeholders = [
+                f"value_{next(idx_gen)}" for _ in value
+            ]
+            clause_str = ('"{0}" in ({1})'.format(
+                field,
+                ','.join(f":{p}" for p in placeholders)
+            ))
```
wardi, line 407: while making changes here let's make this sql generation from fieldnames properly safe. I would use my old `identifier` function but maybe there's a `sa.column` or something that is better now?

ckanext/datastore/backend/postgres.py
```diff
@@ -369,24 +386,31 @@ def _validate_record(record: Any, num: int, field_names: Iterable[str]):
...
+            clause = (clause_str, dict(zip(placeholders, value)))
         else:
-            clause: tuple[Any, ...] = (u'"{0}" = %s'.format(field), value)
+            placeholder = f"value_{next(idx_gen)}"
+            clause: tuple[Any, ...] = (f'"{field}" = :{placeholder}', {
```
wardi, line 411: same issue here

ckanext/datastore/backend/postgres.py
```diff
@@ -697,17 +728,20 @@ def _drop_indexes(context: Context, data_dict: dict[str, Any],
...
     indexes_to_drop = context['connection'].execute(
-        sql_get_index_string, data_dict['resource_id']).fetchall()
+        sa.text(sql_get_index_string),
+        {"relname": data_dict['resource_id']}
+    ).fetchall()
     for index in indexes_to_drop:
-        context['connection'].execute(
-            sql_drop_index.format(index[0]).replace('%', '%%'))
+        context['connection'].execute(sa.text(
+            sql_drop_index.format(index[0])
```
wardi, line 739: potentially unsafe index name formatting here, but only if a bad name was created in the datastore db by another application/db user

ckan/templates/user/list.html
smotornyuk: Sqlalchemy v2 disabled some magic that we are using here, so using old syntax is not an option. Just to be clear, `page.items` does not contain dictionaries. Even in `master`. The type is the same on `master` and `sqlalchemy-2.0` - `list[Row]`, where `Row` is some sort of tuple. In SQLAlchemy v1 this `Row` allowed inconsistent access to its content by named attributes. In SQLAlchemy v2 this implicit logic was removed.

In order to "generate the old dict format", the view has to be rewritten and we have to manually dictize every user that is passed to the template. While it allows us using `user.name` once again in templates, such change:
* implies dictization inside view. We are usually doing it inside action, so it sounds like a wrong thing 
* breaks plugins, that expected `list[Row]` in template.

wardi: As you said in your first bullet our usual pattern is to have view call actions that return lists/dicts to pass to templates. Passing sqlalchemy rows to templates (that now behave differently in sqlalchemy 2) is the oddity here that I think is worth fixing.

I don't imagine there are any plugin templates that depend on user being `list[Row]` because jinja2 treats everything as a dict-like object even when accessing attributes.

ckanext/datastore/backend/postgres.py
smotornyuk: Yes, `sa.column` solves the issue. Is it worth backporting? every field itself is checked in the beginning of the method against a list of existing column names: `field not in fields_types:`.

The only edge case I can imagine - datastore field contains `"` in its name. I doubt that there is a possible vector for SQL ingestion, because all statements that rely on where-clauses are executed as a single query. So basically user who uploads a "dangerous" resource can only affect this exact resource's datastore table.

ckanext/datastore/backend/postgres.py
smotornyuk: +

ckan/templates/user/list.html
smotornyuk: We cannot call an action that returns dictionaries in this case:)

A long time ago we had a list of dictionaries. But the `user_list` action does not have limit/ offset support, so it returned **all** the users, which caused timeouts on some portals.

Now `user.index` view gets a query object from the action, which is itself is pretty strange solution, but that's our reality. In order to fix it we have to add `limit`/`offset` to the `user_list` action. But it's not enough. This action returns a list of users, and we don't know how many users there are in order to build a pager.
 we cannot use a total number of users in DB because `user_list` accepts `q` parameter that filters users by name. So we have to make two requests: one that returns a list of users inside limit/offset window and another one that returns a query, which we'll use to get the number of users matching the `q` for pagination widget.

And, again, the only alternative is to violate our rules and perform dictization over query results inside view.

So neither solution is perfect. That's why I am so reluctant it terms of making this change

ckanext/datastore/backend/postgres.py
smotornyuk: Fixed - `sa.column` should work here as well

ckan/templates/user/list.html
smotornyuk: UPD: I added dictization to `user.index` endpoint. Even though it's not perfect, at least now we have comments that explain, why we are doing things in this way. And when somebody decides to update this view/action, it will save him a bit of time

pdelboca:
Let's go 🚀

PR #7585, opened 2023-05-12
Store JS translation files in the CKAN storage folder

### LINKED ISSUE #6278: i18n js files are built in source directory
**CKAN version**
all

**Describe the bug**
On startup the ckan process tries to rebuild all JS translations by overwriting a file in the ckan source directory. This means the web process needs to have permission to write to its own source directory just to be able to start. This encourages a bad security practice so we should generate this file in a location that the web process would normally have write access, such as the upload directory.

We should also stop regenerating these files on startup when `debug=false` (require use of the cli or deployment for updating translations) for production sites to cut down on startup time.

**Steps to reproduce**
Make a change to a .po file covering text that appears in js on a site supporting multiple languages, then start ckan with a user that doesn't have permission to modify files in its own source directory

**Expected behavior**
ckan should start even when it doesn't have permission to modify its own source directory

**Additional details**
```
...
  File "/project/ckan/ckan/config/middleware/__init__.py", line 55, in make_app
    load_environment(conf, app_conf)
  File "/project/ckan/ckan/config/environment.py", line 131, in load_environment
    build_js_translations()
  File "/project/ckan/ckan/lib/i18n.py", line 408, in build_js_translations
    _build_js_translation(lang, po_files, js_entries, dest_file)
  File "/project/ckan/ckan/lib/i18n.py", line 348, in _build_js_translation
    with open(dest_filename, u'w') as f:
IOError: [Errno 30] Read-only file system: u'/project/ckan/ckan/public/base/i18n/fr.js'
```

amercader:
Instead than in a folder inside the CKAN source code, which causes a lot of permission headaches when deploying. The API endpoint use remains unchanged, this is just an internal refactoring

See note about `ckan.storage_path`

fixes #6278


ckan/lib/i18n.py
```diff
@@ -77,6 +75,16 @@ def get_ckan_i18n_dir() -> str:
     return path
 
 
+def get_js_translations_dir() -> str:
+    storage_path = config.get(
+        "ckan.storage_path"
+    ) or tempfile.gettempdir()
```
amercader, line 81: I copied this logic from the [webassets](https://github.com/ckan/ckan/blob/2718a137ecaf2956e471ba1fd4286566800c2861/ckan/lib/webassets_tools.py#L151) handling module (minus the dedicated config options, which seemed overkill). I think that works fine for something semi-ephemeral like these small json files, but maybe at some point we would need to think about requiring a valid and existing value for `ckan.storage_path` to start up, like we require `ckan.site_url` or others.

wardi: we should avoid predictable temp files by using [mkdtemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) or similar instead of gettempdir (this goes for webassets too)

ckan/lib/io.py
```diff
@@ -53,3 +58,22 @@ def decode_path(p: Union[bytes, Any]) -> str:
...
+
+def get_ckan_temp_directory() -> str:
+    """
+    Returns the path to a securely created temporary directory that can be used
+    to store internal generated files like webassets, i18n files, etc
+
+    It is used as fallback when `ckan.storage_path` is not defined.
+    """
+
+    unique_suffix = hashlib.sha256(config["SECRET_KEY"].encode()).hexdigest()[:10]
```
wardi, line 71: @smotornyuk looks like "SECRET_KEY" is our one secret to rule them all, re #7651

amercader: Yes, I think we should work towards that: https://github.com/ckan/ckan/issues/7409

ckan/lib/io.py
```diff
@@ -53,3 +58,22 @@ def decode_path(p: Union[bytes, Any]) -> str:
...
+    directory_name = f"ckan_{unique_suffix}"
+
+    path = os.path.join(tempfile.gettempdir(), directory_name)
+
+    if not os.path.exists(path):
+        os.mkdir(path, mode=0o700)
```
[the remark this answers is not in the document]
amercader: I just copied it from the official Python docs :shushing_face: (with the right perms)

amercader:
There is a problem when `SECRET_KEY` is not defined, #7450 needs to be updated and merged first. I'll add the details there andark this as draft

amercader:
@wardi this is ready to be re-reviewed. The only potentially controversial thing is [e640af6](https://github.com/ckan/ckan/pull/7585/commits/e640af638f692f1928e3f34b4412033c874004ce), where I add an explicit check for SECRET_KEY because the config hasn't been validated yet

amercader:
boo still one test failing, checking...

ckan/tests/cli/test_asset.py
```diff
@@ -12,9 +12,7 @@ def test_build_and_clean(cli, ckan_config, tmpdir, monkeypatch):
     """
     monkeypatch.setitem(ckan_config, u'ckan.storage_path', str(tmpdir))
     cli.invoke(ckan, [u'asset', u'build'])
-    assert len(tmpdir.listdir()) == 1
-    webassets_folder = tmpdir.listdir()[0]
-    assert webassets_folder.basename == u'webassets'
+    webassets_folder = [d for d in tmpdir.listdir() if d.basename == "webassets"][0]
```
wardi, line 15: removed some asserts here, we could at least assert that this list isn't empty?

ckan/views/api.py
```diff
@@ -473,12 +473,12 @@ def i18n_js_translations(
...
-    source = os.path.abspath(os.path.join(ckan_path, u'public',
-                             u'base', u'i18n', u'{0}.js'.format(lang)))
+    js_translations_folder = get_js_translations_dir()
+
+    source = os.path.join(js_translations_folder, f"{lang}.js")
     if not os.path.exists(source):
-        return u'{}'
-    translations = json.load(io.open(source, u'r', encoding='utf-8'))
+        return "{}"
+    translations = json.load(io.open(source, "r", encoding="utf-8"))
```
wardi, line 481: I see this existed before this PR, but why are we parsing the json only to immediately send it in the response? We could use LazyJSONObject here or better yet have them served as normal files with nginx or like webassets with x-sendfile.

wardi: Doesn't need to be part of this PR, I only noticed because of the diligent `u'string'` cleanup here.

PR #7590, opened 2023-05-16
faster navigation between dataset and resource edit pages

### LINKED ISSUE #7586: Edit navigation in fewer clicks
Example: I'm at a dataset page and I want to edit the data dictionary for resource number 2.

Option 1:
- click "manage" (page refresh)
- click resources tab (page refresh)
- click on resource number 2 (page refresh)
- click data dictionary tab (page refresh)

Option 2:
- click "Explore" drop-down next to resource 2
- click "more information" (page refresh)
- click "manage" (page refresh)
- click data dictionary tab (page refresh)

There are some shortcuts, e.g. we can click the text of the resource name on the dataset page to save one click.

I'd like to be able to navigate to resource editing features in 1/2 clicks instead of 3/4 clicks and page refreshes.

### Proposed solution:

Dataset Page:
- [X] add "Manage", "Data Dictionary" and "Views" buttons to the "Explore" drop-down.

Resource Page:
- [X] add "Data Dictionary" and "Views" buttons next to "Manage"
- [X] add "Manage", "Data Dictionary" and "Views" drop-down for the other resources on the sidebar

Dataset Edit Page:
- [x] in sidebar add resource list + edit drop-downs like on resource page

Dataset Edit Resources Tab:
- [x] rename to "Reorder resources"
- [x] in sidebar add resource list + edit drop-downs like on resource page

Resource Edit/Data Dictionary/Views Pages:
- [x] in sidebar add resource list + edit buttons like on resource page

wardi:
Fixes #7586

### Proposed fixes:
Use otherwise empty sidebar, add more direct links to edit pages

### Features:

- [ ] includes API changes


smotornyuk:
Re: style changes. Every diff must be applied after the previous one.

### Current state
| Width | Screenshot | 
|-|-|
|320px | image|
| 360px |image |
| 600px and more |image |

Below I won't include >=600px, because it's ok and doesn't require changes

---
### minimal change

1. Push buttons to the right side. It will make them more similar to the desktop version
2. Add a gap between buttons. When they stack vertically, there is no space between them.

| Width | Screenshot |
|-|-|
| 320px | image|
| 360px | image |

Changes explanation:
* d-flex - turns block into flex element, simplifying positioning
* flex-wrap - force buttons to split into few lines, when there is not enough space. Otherwise, they will always be rendered in one row. On a small screen, it means a tall narrow button, which doesn't look good.
* gap-1 - add ~4px horizontal and vertical space between buttons 
* justify-content-end - push buttons to the right side of the screen(or left for Hebrew, Arabic, etc.)

```diff
diff --git a/ckan/templates/package/resource_read.html b/ckan/templates/package/resource_read.html
index 9bf379288..2c0c3d79a 100644
--- a/ckan/templates/package/resource_read.html
+++ b/ckan/templates/package/resource_read.html
@@ -25,7 +25,7 @@
       <div class="module-content">
         <div class="actions">
           {% block resource_actions %}
-            <ul>
+            <ul class="d-flex flex-wrap gap-1 justify-content-end">
               {% block resource_actions_inner %}
                 {% block action_manage %}
                   {% if h.check_access('package_update', {'id':pkg.id }) %}

```

---
### Better alignment

* on 360px blue button(last in the row, actually, no matter what color it has) placed further to the right than other buttons. That happens because all buttons but the last one have a 5px margin on the right.
* in addition, all the buttons are not aligned with the gray `show more` row of the table above them. It happens because the whole actions block moved 15px to the right from its normal position.

| Width | Screenshot |
|-|-|
| 320px |image|
|360px | image |

Update SASS file and recompile CSS running `npm run build` (you have to install JS dependencies by running `npm ci` first)
```diff
diff --git a/ckan/public/base/scss/_module.scss b/ckan/public/base/scss/_module.scss
index c2274a9ac..e60e86f30 100644
--- a/ckan/public/base/scss/_module.scss
+++ b/ckan/public/base/scss/_module.scss
@@ -155,7 +155,10 @@
         position: relative;
         float: right;
         top: -10px;
-        right: -15px;
+        right: 0;
+        li {
+          margin-right: 0px;
+        }
     }
 }
```

---

### Slightly improved UX 

* Using separate full-width buttons for each row on small screens should be more convenient. But it is quite a big change of original styles, so we can leave things as they are now. 

| Width | Screenshot |
|-|-|
| 320px | image |
| **756px** | image |

Changes explained:

* remove `float: right` in order to use full-width
* set `width: 100%` for buttons. Starting from 576px this width will be ignored and buttons will switch no standard width
* Add `flex-column flex-sm-row` in order to show one button per row on a small screen and put all buttons into a single row starting from 576px. Use `flex-md-row` if you want to switch into row view on 768px.
* ps-0 - remove left padding
* float-md-end - add `float: right` starting from 768px, so that buttons and resource names are shown on the same line. On smaller screens, it's better to render them on different rows because they won't fit on a single line.
* *Optionally*. Replace `gap-1` with `gap-3 gap-sm-1`. It will add extra space between buttons on the mobile screen and you won't accidentally tap the wrong button. Starting from `sm`-breakpoint(576px) this extra space will be removed.

Run `npm run build` in order to compile CSS.

```diff
diff --git a/ckan/public/base/scss/_module.scss b/ckan/public/base/scss/_module.scss
index e60e86f30..9bfbf0a56 100644
--- a/ckan/public/base/scss/_module.scss
+++ b/ckan/public/base/scss/_module.scss
@@ -153,11 +153,13 @@
     @include border-radius(3px 3px 0 0);
     .actions {
         position: relative;
-        float: right;
         top: -10px;
         right: 0;
         li {
           margin-right: 0px;
+          > .btn-group, a {
+            width: 100%;
+          }
         }
     }
 }
diff --git a/ckan/templates/package/resource_read.html b/ckan/templates/package/resource_read.html
index 2c0c3d79a..fdc6a1674 100644
--- a/ckan/templates/package/resource_read.html
+++ b/ckan/templates/package/resource_read.html
@@ -25,7 +25,7 @@
       <div class="module-content">
         <div class="actions">
           {% block resource_actions %}
-            <ul class="d-flex flex-wrap gap-1 justify-content-end">
+            <ul class="d-flex flex-wrap flex-column flex-sm-row gap-1 justify-content-end ps-0 float-md-end">
               {% block resource_actions_inner %}
                 {% block action_manage %}
                   {% if h.check_access('package_update', {'id':pkg.id }) %}
```

wardi:
@smotornyuk do you have any idea how to fix the drop-down menu z-order issue on the resource lists in the sidebar?

smotornyuk:
You can use flex(I feel, that 80% of position issues can be solved via flex 😆 ) instead of precise positioning:
```diff
diff --git a/ckan/templates/package/snippets/resources.html b/ckan/templates/package/snippets/resources.html
index 28e3b446e..530ff88db 100644
--- a/ckan/templates/package/snippets/resources.html
+++ b/ckan/templates/package/snippets/resources.html
@@ -25,13 +25,15 @@ Example:
             {% for resource in resources %}
               {% set url = h.url_for('%s_resource.read' % pkg.type, id=pkg.name, resource_id=resource.id) %}
               {% if active == resource.id %}
+
                 <li class="nav-item active">
                   <a href="#">{{ h.resource_display_name(resource)|truncate(25) }}</a>
                 </li>
               {% elif can_edit %}
-                <li class="nav-item position-relative">
+
+                <li class="nav-item d-flex justify-content-between">
                   <a href="{{ url }}">{{ h.resource_display_name(resource)|truncate(25) }}</a>
-                  <div class="dropdown position-absolute top-50 end-0 translate-middle">
+                  <div class="dropdown">
                     <button class="btn btn-light btn-sm dropdown-toggle" type="button" id="dropdownRes{{ loop.index }}" data-bs-toggle="dropdown" aria-expanded="false"><i class="fa fa-wrench"></i></button>
                     <ul class="dropdown-menu" aria-labelledby="dropdownRes{{ loop.index }}">
                       <li>{% link_for _('Edit resource'), named_route=pkg.type ~ '_resource.edit', id=pkg.name, resource_id=resource.id, class_='dropdown-item', icon='pencil' %}</li>

```

PS `activity` has [its own resources.html](https://github.com/ckan/ckan/blob/master/ckanext/activity/templates/package/snippets/resources.html) which removes these dropdowns. Probably you can add a block and inherit in in activity template

wardi:
Screenshot from 2023-05-29 11-52-57
@smotornyuk with the flex change the z-order for the drop-down menus are fixed, but the the `d-flex` class on the `<li>` makes the resource `<a>` only clickable on the text of the link (see the hover-over gray area) not the full column width. Ideas?

smotornyuk:
Sorry, didn't notice this half-width link.

Here we have to get back to absolute positions but in a bit different way. The problem with overlapping items happens when you apply the `translate-middle` class because it applies the `transform` CSS property to the element, and this property slightly changes the way `z-index` works. In our case, the best option is to avoid `transform` completely.

You were using `translate-middle` in order to align elements on both X and Y axis. If we add `position-absolute` to the item, it will move to the beginning of the row, but its vertical alignment will be ok, thanks to `d-flex`. So we only need to put it at the end of the row. Add familiar class `end-0`, which moves dropdown to the right, and then add `me-2` which adds  .5rem space after the item.

Finally, add `flex-fill` to `a` tag, so that it fills all available space.

```diff
diff --git a/ckan/templates/package/snippets/resources.html b/ckan/templates/package/snippets/resources.html
index 0942f05b8..32c95e5ee 100644
--- a/ckan/templates/package/snippets/resources.html
+++ b/ckan/templates/package/snippets/resources.html
@@ -29,9 +29,9 @@ Example:
                   <a href="#">{{ h.resource_display_name(resource)|truncate(25) }}</a>
                 </li>
               {% elif can_edit %}
-                <li class="nav-item d-flex justify-content-between">
-                  <a href="{{ url }}">{{ h.resource_display_name(resource)|truncate(25) }}</a>
-                  <div class="dropdown">
+                <li class="nav-item d-flex justify-content-between position-relative">
+                  <a class="flex-fill" href="{{ url }}">{{ h.resource_display_name(resource)|truncate(25) }}</a>
+                  <div class="dropdown position-absolute end-0 me-2">
                     <button class="btn btn-light btn-sm dropdown-toggle" type="button" id="dropdownRes{{ loop.index }}" data-bs-toggle="dropdown" aria-expanded="false"><i class="fa fa-wrench"></i></button>
                     <ul class="dropdown-menu" aria-labelledby="dropdownRes{{ loop.index }}">
                       <li>{% link_for _('Edit resource'), named_route=pkg.type ~ '_resource.edit', id=pkg.name, resource_id=resource.id, class_='dropdown-item', icon='pencil' %}</li>

```

wardi:
@smotornyuk fix for activity override, CSS, merge conflict are in

PR #7591, opened 2023-05-16
[7581]Fix Flask-login break IAuthenticator identify interface

### LINKED ISSUE #7581: Flask-login break IAuthenticator identify interface
## CKAN version

2.10.0

## Describe the bug

IAuthenticator identify interface says that extensions should set `g.user` and `g.userobj` https://docs.ckan.org/en/2.10/extensions/plugin-interfaces.html#ckan.plugins.interfaces.IAuthenticator.identify

This works fine on 2.9 and setting those objects lets developer to authenticate users via some other method.

On 2.10, `g.user` is over written with `current_user` from flask-login immediately after the interface calls and the authentication is broken.
https://github.com/ckan/ckan/blob/b2afa91a8e57bdeac874c58b144256732577d9a2/ckan/views/__init__.py#L80-L95

### Steps to reproduce
Implement extension with IAuthenticator identify method and set `g.user` and `g.userobj` in there.
Try to authenticate with it.

### Expected behavior
User should be authenticated with given user object.

TomeCirun:
Fixes #7581 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/header.html
+++ b/ckan/templates/header.html
@@ -3,10 +3,10 @@
 {% block header_wrapper %} {% block header_account %}
 <div class="account-masthead">
   <div class="container">
-    {% block header_account_container_content %} {% if c.userobj %}
-    <div class="account avatar authed" data-module="me" data-me="{{ c.userobj.id }}">
+    {% block header_account_container_content %} {% if current_user.is_authenticated %}
+    <div class="account avatar authed" data-module="me" data-me="{{ current_user.id }}">
       <ul class="list-unstyled">
-        {% block header_account_logged %} {% if c.userobj.sysadmin %}
+        {% block header_account_logged %} {% if current_user.sysadmin %}
         <li>
           <a href="{{ h.url_for('admin.index') }}" title="{{ _('Sysadmin settings') }}">
             <i class="fa fa-gavel" aria-hidden="true"></i>
@@ -16,9 +16,9 @@
         {% endif %}
         {% block header_account_profile %}
         <li>
-          <a href="{{ h.url_for('user.read', id=c.userobj.name) }}" class="image" title="{{ _('View profile') }}">
-            {{ h.user_image((c.user if c and c.user else ''), size=22) }}
-            <span class="username">{{ c.userobj.display_name }}</span>
+          <a href="{{ h.url_for('user.read', id=current_user.name) }}" class="image" title="{{ _('View profile') }}">
+            {{ h.user_image((current_user.name), size=22) }}
+            <span class="username">{{ current_user.display_name }}</span>
           </a>
         </li>
         {% endblock %}
@@ -32,7 +32,7 @@
         {% endblock %}
         {% block header_account_settings_link %}
         <li>
-          <a href="{{ h.url_for('user.edit', id=c.userobj.name) }}" title="{{ _('Profile settings') }}">
+          <a href="{{ h.url_for('user.edit', id=current_user.name) }}" title="{{ _('Profile settings') }}">
             <i class="fa fa-cog" aria-hidden="true"></i>
             <span class="text">{{ _('Profile settings') }}</span>
           </a>
--- a/ckan/views/__init__.py
+++ b/ckan/views/__init__.py
@@ -90,9 +90,12 @@ def identify_user() -> Optional[Response]:
                     break
             except AttributeError:
                 continue
-    # sets the g.user/g.userobj for extensions
-    g.user = current_user.name
-    g.userobj = '' if current_user.is_anonymous else current_user
+
+    # do not overwrite g.user if the extensions provide it
+    if not getattr(g, u'user', None):
+        # sets the g.user/g.userobj for extensions
+        g.user = current_user.name
+        g.userobj = '' if current_user.is_anonymous else current_user
 
     # logout, if a user that was still logged in is deleted.
     if current_user.is_authenticated:
@@ -103,15 +106,11 @@ def identify_user() -> Optional[Response]:
     # that IAuthenticator extensions do not need to access the user model
     # directly.
     if g.user:
+        g.author = g.user
         if not getattr(g, u'userobj', None):
             g.userobj = model.User.by_name(g.user)
 
-    # general settings
-    if g.user:
-        if g.userobj:
-            userobj = model.User.by_name(g.user)
-            userobj.set_user_last_active()  # type: ignore
-        g.author = g.user
+        g.userobj.set_user_last_active()  # type: ignore
     else:
         g.author = g.remote_addr
     g.author = str(g.author)
--- a/ckan/views/user.py
+++ b/ckan/views/user.py
@@ -604,17 +604,15 @@ def logout() -> Response:
         response = item.logout()
         if response:
             return response
+
     user = current_user.name
     if not user:
         return h.redirect_to('user.login')
 
-    came_from = request.args.get('came_from', '')
     logout_user()
+    session.clear()
 
-    field_name = config.get("WTF_CSRF_FIELD_NAME")
-    if session.get(field_name):
-        session.pop(field_name)
-
+    came_from = request.args.get('came_from', '')
     if h.url_is_local(came_from):
         return h.redirect_to(str(came_from))
 


TomeCirun:
@smotornyuk  I'm not entirely sure if I understand @Zharktas correctly, If you can provide a precise explanation of the desired behavior, I will do my best to offer a solution.

Let's review the situation:

Previously, with repoze.who the behavior was as follows: when a user logged in, repoze.who would set up the username in request.environ, which we then assigned to the g object. 

The current behavior with flask-login is different: when a user logs in, flask-login sets up the current_user for us immediately after the login_user() function is executed.

So, I'm unsure about how we can currently log in a user by only providing the g.user (as we did with repoze.who), even without validation. Could you please provide more information so that I can understand the logic better?

Thank you

TomeCirun:
One solution I can think of is by placing `current_user` in the `ckan.common` module. We can then import it within the extension, similar to how we imported `g` (e.g., `from ckan.common import current_user`). After all the custom validations succeed, we can set `current_user` to the `userobj` (e.g., `current_user = userobj`). Previously, the extension would call `user.model` to retrieve the `userobj`.

@smotornyuk

smotornyuk:
No, it won't work without a lot of changes, while we only need to keep a small feature from the past.

Before jumping into the use case, let's recall CKAN API. Imagine, that you call an action:

```
curl http://ckan/action/create_package  -H Authorization:token_of_sergey
```
In such a request, you'll be authenticated as `sergey`. Whatever you'll do will check access for `sergey` and will store `sergey` as an author. Look at the following example of possible code for `create_package`:

```python
def create_package(context, data_dict):
   tk.get_action("package_create")(context, data_dict)
   return tk.get_action("package_show")(context, data_dict)
```
We create a package using the context of the current user(`sergey`) and then show the package using the same context(`sergey`).

Now imagine, that we have a requirement: every package must be owned by user `admin`. We can change the logic of `create_package` in the following way:
```python
def create_package(context, data_dict):
   tk.get_action("package_create")({"user": "admin"}, data_dict)
   return tk.get_action("package_show")(context, data_dict)
```
You see, we are using a completely different context when the package is created(`admin`). In this way, the package will be owned by the admin and all authorization checks will run against `admin` inside `package_create`. But in order to show the package, we are still using the original context(`sergey`)

Basically, inside CKAN API you can complete different parts of action as a different user. Initial context is not a constant, you can always patch it. But another important moment: you can patch it only internally, using different contexts for part of your action. But whenever your action is finished, the original caller of your action will remain the same. And if he is going to call another action, he'll use the original context, not your custom version.

---

Now let's look at the possibilities that existed in CKAN v2.9 and what we have\want instead in v2.10

Keep in mind, that even though `tk.current_user` is a source of truth, there are still a lot of places, where `g.user` is used.

### persistent session

Inside a code, you were able to call the `set_repose_user` function, which logged in the user. From now on, all the following requests will be done on behalf of the given user. But the current request(where you called `set_repose_user`) didn't know anything about the user(`g.user` is not updated). You had to return a redirect and only inside the next request user would be available as `g.user`

Now, you can do `tk.login_user`. It does not update `g.user` either, so you have to return a redirect and only the next request will set `g.user` to the expected value. It's the same as it was before. In addition, `tk.login_user` updates `tk.current_user` instantly. It's not bad by itself, but as we are using `g.user` and `tk.current_user` simultaneously, it creates a space for a huge amount of problems, because `g.user` and `tk.current_user` are out of sync starting from the moment you called `tk.login_user` and till the beginning of the next request.

#### conclusion
So, we need to either postpone the effect of `login_user`, so that `current_user` is not updated immediately(🙅) or wrap login_user into a logic, that will synchronize `g.user` and `tk.current_user`(👍)

### temporal, context-like session

In CKAN  v2.9 you can change `g.user` at any moment. For example, if you edit `views/home.py` and add the following lines at the beginning of `def index()` view:
```python
     g.user = "sergey"
     g.userobj = model.User.get("sergey")
```
(use existing username), the page will be rendered as if you were logged in as `sergey`. If you visit any other page, you'll see that you are not logged in. So, overriding `g.user/g.userobj` does not login user. It rather behaves like `context['user']` inside actions - it switches the current user for a while. 
It allows you to implement something like `do action as a different user` feature. For example, as an admin, you can edit the profile of user `john`(even if it's not your profile). But you cannot follow a user/organization/dataset as `john`. You cannot pretend that you are `john`. The only way to do something on behalf of another user is to login into the application using his credentials.
Now imagine, that you, as an admin, see a dropdown near the `Follow` button. This dropdown lists all the users and when you choose one and click `Follow`, this chosen user will become a follower of the entity. Such a thing can be achieved by overriding `g.user` inside a request. It's not a good solution, but it's a viable solution.

Let's go back to v2.10. You want to do something similar to the example above. You call `tk.login_user`. The user is replaced immediately and it works. But `login_user` overrides the session. The next request will be done on behalf of another user as well, and that's not what you need. You want to do something as a different user only this time and switch back to your original account on the next request.
As an option you can do the following:
```python
original_user = tk.current_user()
tk.login_user(different_user)
# do stuff 
tk.login_user(original_user)
```
But it works only if you start and finish working as a different user inside the same scope. What if you want to switch to a different account side one function, and switch back inside a different function? For example, you want to start using a different account inside `IAuthenticator.identify` and switch back right before returning the response to the user. It could be done like this(it won't actually work, but I'm only showing the idea, not the working code 😄):
```python
#IAuthenticator.identify
if want_to_switch:
    g.__original_user = tk.current_user
    tk.login_user(different_user)


# IMiddleware.make_middleware(app.after_request)
if hasattr(g, "__original_user"):
    tk.login_user(g.__original_user)

```

In the original issue you can see the working version of this idea:
```python
from flask import current_app
current_app.login_manager._update_request_context_with_user(desired_ckan_user_obj)
```

This code overrides flask's request context, but does not actually call `login_user`, that's why you don't have to login the original user back and it's more like the effect of overriding `g.user` in CKAN v2.9

#### conclusion
We need a mechanism that overrides `current_user/g.user`, but does not changes the session. Or, if the session is changed, in the end, it must be restored to its original state. We don't have to implement it - suggesting a nice, short, and elegant approach would do the thing. But `current_app.login_manager._update_request_context_with_user(desired_ckan_user_obj)` is definitely not that approach. We cannot recommend relying on internal implementation details of flask-login.

TomeCirun:
I tested the new feature with the `example_iauthenticator` plugin and I set the following logic:
```python
if request.endpoint == 'user.index':
            g.user = 'cirun'
            g._temp_login = True
if request.endpoint == 'dataset.search':
            g.user = 'default'
            g._temp_login = True
```

let me know your opinion on this @smotornyuk

TomeCirun:
btw thanks for the excellent explanation above

smotornyuk:
Not bad, but too fragile, especially this line, where you are cleaning the whole session. 

What if instead all the changes, you introduce just a single new function:

```python
# toolkit
def set_user_for_current_request(user: User | AnonymousUser):
    current_app.login_manager._update_request_context_with_user(user)
```

With such a function it would be possible to do `set_user_for_current_request(user)` from the `IAuthenticatior.identify` and everything else will just work, without increased complexity

I don't like the fact that we are going to use private methods. But given the fact that you have a comment in your changes `hack for flask-login`, we have no other choice:)

TomeCirun:
I tried on `2.9` with `repoze.who`, and it seems that the session doesn't store anything, nor is the `REMOTE_USER` set in the environ. It behaves exactly as you mentioned before, similar to the `context['user']` inside actions.

I apologize for the confusion. I initially thought we could have a session set up in the same way as when a user is properly logged in with credentials. However, if that is unnecessary, I will remove that part of the code. Additionally, do we still need to keep track of the user's last_active status in such situations?

FYI, in `2.10`, whenever a user is authenticated, `flask-login` sets a key named `_login_user` in the g object. The internal function `_update_request_context_with_user(user)` assigns the user to `g._login_user`, and if no user is provided, it sets it to anonymous.

That's the 'hack' I discovered. Inside `flask-login`, they utilize `g._login_user` in the same way we use `g.user/g.userobj`. So, do we still need to support `g.userobj` when we already have the same user-object inside `g._login_user`?

With this information, please let me know what steps I should take next.

Thank you.
@smotornyuk

smotornyuk:
>  I initially thought we could have a session set up in the same way

Yes, it's actually a good idea, but it will work only in pure CKAN because CKAN does not use sessions(almost). As for extensions, I'm sure that there are a lot of them, who are utilizing sessions. And they won't be happy if you just clean up the session during logout:)

--- 

Yes, `_update_request_context_with_user` basically just sets `_login_user`. But, I believe, private methods are a bit more stable, than private variables. It's always safer to call a method instead of doing whatever is written in the method manually. In addition, if flask-login removes this method in the future, we'll notice it immediately, because of the error. If flask-login will stop using `_login_user`, we won't notice it until someone tries to temporarily log in a user. That's why, when it comes to hacks, I prefer methods over variables.


As for using `g._login_user` instead of `g.userobj`. You see, we don't need `g.user`/`g.userobj` at all inside CKAN, because we have `tk.current_user` (and the `current_user` variable in templates). We are keeping `g.user`/`g.userobj` just for backward compatibility. That's why we cannot tell everyone just to user `g._login_user` instead of `g.userobj`. First of all, it requires a lot of code updates. The second reason - why would we recommend `g._login_user` when we have `tk.current_user`?

TomeCirun:
this is why the tests fail: `flask_wtf.csrf.CSRFError: 400 Bad Request: The CSRF token is missing`.

I also made a fix for the `login_via_auth_header`. Since `@login_manager.request_loader` executes on every request, I added an additional check to ensure it's an actual login via auth header..

PR #7599, opened 2023-05-24
Documentation user sessions

Filip3mac:
Adds documentation for user sessions in Ckan 2.9.


tino097 (commented):
You will need to update the docs with other options, after the #7406 (if) get merged

PR #7607, opened 2023-05-26
fix: `generate fake-data` with custom factory accepts params

smotornyuk:
`ckan generate fake-data` accepts the factory name in one of the forms:
* `generate fake-data alias`
* `generate fake-data --factory-class module.name:Class`

Both commands are supposed to accept an arbitrary number of additional arguments that specify values for files:
```
ckan generate fake-data dataset --title="My Dataset" --notes="Description"
```

But due to changes in `click` now the form with `--factory-class` doesn't accept such extras. 

This PR extends the syntax of this command and allows passing `factory-flag` as an alias. i.e:

```sh
# instead of the old form
ckan generate fake-data --factory-class x.y.z:Dataset

# use the short form
ckan generate fake-data x.y.z:Dataset
```

This new form works with extra params and does not conflict with alias style (because it's easy to tell the difference between aliases `dataset`/`organization`/`user` and import path `x.y.z:Class`).

The old form, with `--factory-class` is still available for compatibility. Bit its help-text clearly states that this option is deprecated and it does not accept extra parameters.

### DIFF
--- a/changes/7607.bugfix
+++ b/changes/7607.bugfix
@@ -1,0 +1,1 @@
+`ckan generate fake-data --factory-class x.y.z:Factory` does not accept field values.
--- a/ckan/cli/generate.py
+++ b/ckan/cli/generate.py
@@ -230,31 +230,37 @@ _factories = {
     "allow_extra_args": True, "ignore_unknown_options": True
 })
 @click.argument(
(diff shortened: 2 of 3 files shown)


kowh-ai (approved):
tested and works as specified

PR #7609, opened 2023-05-27
Controlled iteration order for interfaces

smotornyuk:
Add `ckan_reverse_iteration_order` flag to the Interface class. Whenever it is set to `True`, `PluginImplementations` will return plugins implementing this interface in reverse order.

It reduces the risk of writing code where iteration goes not in the expected order. In addition, it makes more explicit, which interfaces are iterated in reverse order. 
Finally(but I'd rather not mention it in the docs), it allows you to write `IActions.ckan_reverse_iteration_order = True` in order to revert the order of plugins during action registration(just an example, of course, you can do it with any interface). It may save someone who wants to patch things locally if we decide to change the order for a particular interface in the future or discover, that order was non-intentionally changed

Interfaces that are traversed in reverse order:

* IConfigDeclaration
* IConfigurer
* IValidators
* ITranslation

### DIFF
--- a/ckan/lib/api_token.py
+++ b/ckan/lib/api_token.py
@@ -124,7 +124,7 @@ def get_user_from_token(token: str,
     # do preprocessing in reverse order, allowing onion-like
     # "unwrapping" of the data, added during postprocessing, when
     # token was created
-    for plugin in reversed(list(_get_plugins())):
+    for plugin in _get_plugins():
         data = plugin.preprocess_api_token(data)
     if not data or u"jti" not in data:
         return None
--- a/ckan/lib/api_token.py
+++ b/ckan/lib/api_token.py
@@ -121,9 +121,11 @@ def get_user_from_token(token: str,
     data = decode(token)
     if not data:
         return None
-    # do preprocessing in reverse order, allowing onion-like
-    # "unwrapping" of the data, added during postprocessing, when
-    # token was created
+    # do preprocessing in reverse order, allowing onion-like "unwrapping" of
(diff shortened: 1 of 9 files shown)


smotornyuk:
@wardi, it's ready for review

ckan/lib/api_token.py
```diff
@@ -124,7 +124,7 @@ def get_user_from_token(token: str,
     # do preprocessing in reverse order, allowing onion-like
     # "unwrapping" of the data, added during postprocessing, when
     # token was created
-    for plugin in _get_plugins():
+    for plugin in reversed(list(_get_plugins())):
```
wardi: Why is this one back to reversing the old way?

smotornyuk: There are two sides to processing API tokens. When it's created and encoded into a string, all plugins can add additional properties to the token or modify existing properties. When the application receives the token, it decodes the token from a string and extracts all the properties. It sounds like a middleware for an API token, that's why data extraction should happen in reverse order compared to data injection. 

This line used `reversed` originally, but I misinterpret it and set reverse_order to the whole interface. After additional checks(in order to add a comment with the reason behind the reverse order for the interface), I realized, that this interface has 4 methods and 3 of them are executed in normal order and only data extraction from the token received from user happens in reverse order. That's why I restored APIToken to its original form.

wardi: Great, can you add the reason that the reversed plugin order can't be used here to the comment above?

smotornyuk: Sure, done

PR #7610, opened 2023-05-28
Webasset fixes

### LINKED ISSUE #7318: Webassets is not rendering with the same plugin order.
## CKAN version
master, 2.10

## Describe the bug
The order of templates is **first one prevails** but order of webassets creation is **last one prevails**. This happens when an extension tries to override other asset.

### Steps to reproduce
Given `ckanext-resourcefilter` and `ckanext-testing`  (example) if the second one overrides `resources_list.html` template and `modules/resources-list`, when rendering the template of `ckanext-example` will be render with the `webasset` of `ckanext-resourcefilter`.

So:
 - Add `resourcesfilter testing` to `ckan.plugins`
 - Implement `resource_list.html` and `webassets.yaml` in `ckanext-testing` extension. (see below)
 - Load the Dataset read page.
 - `ckanext-example` template will be rendered, but with `ckanext-resourcesfilter` webasset.

### Expected behavior
The webasset created is the one corresponding to the extension.

### Additional details

`webasset.yaml` for `ckanext-testing` (similar to `ckanext-resourcesfilter` to override behaviour):

```yaml
modules/resources-list:
    contents: modules/resources-list.js
    output: testing/modules/%(version)s_resources-list.js
    filters: rjsmin
    extra:
        preload:
           - base/main
```

### Screenshot 1:

With `ckan.plugins= ... testing resourcesfilter...`

This screenshot shows that the template is from `ckanext-testing` but the webasset is from `ckanext-resourcesfilter` even when the `asset` tag is explicitly calling for the testing asset:

```jinja2
...
  <h1 style="color:darkgreen">FROM CKANEXT-TESTING</h1>
  {% asset 'testing/modules/resources-list' %}
...
```

image

### Screenshot 2

With `ckan.plugins = ... resourcesfilter testing ...`

This screenshot shows that the template is from resourcesfilter extension. The `moustache` and `pagination` webassets are from `resourcefilter` but the `resource-list.js` is from `ckanext-testing` (I expect it also to be from `ckanext-resourcesfilter`).

image

smotornyuk:
Fixes #7318

If extension A registers asset `xxx/styles` and extension B registers an asset with the same name, the last loaded plugin wins, and its version of the asset will be used in the end. It is different from a template loading order, where the first plugin wins.

This PR updates the logic and now the first registered version of an asset will be used and all other versions are ignored(just like Jinja2 templates without `extends` tag).

In addition, a couple of other issues were fixed:
* improve types and logging
* prevent recursion when asset A requires asset B and vice versa.\
* add `ckan.webassets.url` config option for those who don't want to serve assets from `/webassets` path.
* Add brief comments for logic, that may be not obvious. If you don't want to read through the webassets code but want to change something inside `ckan.lib.webassets_tools`, you can do it now.

### DIFF
--- a/ckan/lib/webassets_tools.py
+++ b/ckan/lib/webassets_tools.py
@@ -1,12 +1,11 @@
-# encoding: utf-8
 from __future__ import annotations
 
 import logging
 import os
 import tempfile
 from typing import Any, Optional
+from typing_extensions import Literal, TypedDict, assert_never
 
-import yaml
 from markupsafe import Markup
 from webassets import Environment
 from webassets.loaders import YAMLLoader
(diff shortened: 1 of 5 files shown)


ckan/lib/webassets_tools.py
```diff
@@ -1,172 +1,244 @@
...
+    style: list[str]
+    included: set[str]
 
 
 def create_library(name: str, path: str) -> None:
     """Create WebAssets library(set of Bundles).
     """
-    config_path = os.path.join(path, u'webassets.yml')
+    if not env:
+        log.debug("Webassets environment is not initialized yet")
```
pdelboca, line 32: Shouldn't we increase the logging level to CRITICAL? If Webassets environment is not initialized then this means that something important is not gonna we working.

ckan/lib/webassets_tools.py
```diff
@@ -1,172 +1,244 @@
...
+
+        # use absolute path to files. Otherwise they'll behave like templates
+        # and if plugin A and plugin B has their own `x.css`(i.e
+        # `ckanext/A/assets/x.css` and `ckanext/B/assets/x.css`), both plugins
+        # will load the same `x.css` from the plugin that was first registered
+        bundle.contents = [
+            os.path.join(path, item)
+            for item in bundle.contents
+        ]
+        log.debug("Register asset %s", name)
```
pdelboca, line 73: I'm finding this comment a little bit confusing. On the previous if we write: `It gives us templates-like behaviour` but in this new comment it says: `Otherwise the'll behave like templates` so I'm a little bit confusing on what's going on here with the absolute paths.

ckan/lib/webassets_tools.py
```diff
@@ -1,172 +1,244 @@
...
-    public_folder = os.path.abspath(os.path.join(
-        os.path.dirname(__file__), u'..', public))
-
-    base_path = os.path.join(public_folder, u'base')
-
     env = Environment()
     env.directory = static_path
-    env.debug = config.get(u'debug')
-    env.url = u'/webassets/'
+    env.debug = config["debug"] and False
```
pdelboca, line 86: Why do we have `and False`?

ckan/lib/webassets_tools.py
```diff
@@ -1,172 +1,244 @@
...
-        base_path, u'vendor'))
+def register_core_assets():
+    """Register CKAN core assets.
 
-    create_library(u'base', os.path.join(base_path, u'javascript'))
+    Call this function after registration of plugin assets. Asset overrides are
+    not alowed, so if plugin tries to replace CKAN core asset, it has to
+    register an asset with the same name before core asset is added. In this
+    case, asset from plugin will have higher precedence and core asset will be
+    ignored.
```
pdelboca, line 97: Here it says: `Asset overrides are not allowed` but then explains that if plugins tries to replace CKAN core asset, they can.

ckan/lib/webassets_tools.py
```diff
@@ -1,172 +1,244 @@
...
     from ckan.lib.helpers import url_for_static_or_external
-    try:
-        if not g.webassets:
-            raise AttributeError(u'WebAssets not initialized yet')
-    except AttributeError:
-        g.webassets = _make_asset_collection()
-    if name in g.webassets[u'included']:
+
+    if not env:
+        log.debug("Webassets environment is not initialized yet")
```
pdelboca, line 122: Shouldn't this be a CRITICAL as well?

ckan/lib/webassets_tools.py
```diff
@@ -1,172 +1,244 @@
...
         return
 
-    assert env
-    try:
-        bundle: Any = env[name]
-    except KeyError:
-        logger.error(u'Trying to include unknown asset: <{}>'.format(name))
+    if not hasattr(g, "webassets"):
+        g.webassets = _make_asset_collection()
```
pdelboca, line 126: I would add a log.debug() statement since we are expecting something to have happened but it didn't so we are doing it again. This silent scenarios tends to be difficult to debug.

pdelboca (commented):
Hello @smotornyuk! I left some comments about comments in the codebase.

I'm still having a weird behaviour. My expectation is that if `ckanext-testing` and `ckanext-resourcesfilter` implements the same webasset whoever is first in the plugins order wins.

However, while testing this PR `ckanext-resourcesfilter` seems to always win:


image

ckan/lib/webassets_tools.py
smotornyuk: These are different parts of webassets.

 The previous comment describes named assets. One can register `xxx/style` and override existing `xxx/style` asset as we do in templates. That's exactly what we want.

Here we are talking about actual files. We don't want to override a single css/js file just by creating the file with the same name, as we do int templates. If some extension has `style.css`, I don't want this extension to use a different `style.css` that I created inside my extension.

ckan/lib/webassets_tools.py
smotornyuk: Ooops, my debuggin

ckan/lib/webassets_tools.py
smotornyuk: Yes. Webassets library does not allow overrides. If we want to use our custom asset instead of the existing one, we have to register it earlier. In this way, our asset is registered, and when core asset is added to registry, we ignore it, because overrides are not possible

ckan/lib/webassets_tools.py
smotornyuk: Not exactly. We are initializing collection the first time we visit this function. We are not expecting anything, just making sure that something exists. It's like `dict.setdefault`.

smotornyuk:
I suspect that previously you haven't overriden the asset, but override the source file(behavior, that I disabled in this PR).

In order to override webasset, you have to register an asset with **exactly the same name**.

I.e, plugin X has the following asset:
```
# plugin.py/update_config
tk.add_resource("assets", "x")

# webassets.yml
styles:
   ...
```
In order to enable this asset, you use `{% asset "x/styles" %}`

If plugin Y wants to override this asset, it should do the following:
```
# plugin.py/update_config
tk.add_resource("assets", "x")

# webassets.yml
styles:
   ...
```

See? Exactly the same name as an asset. I suspect, that you are doing instead:
```
# plugin.py/update_config
tk.add_resource("assets", "y")

# webassets.yml
x/styles:
   ...
```
This code gives us an asset with a different name - `y/x/styles`. Can you verify that you are using second codeblock, not the last one?

PR #7611, opened 2023-05-28
Remove cast-Context / remove model and standard session from context

smotornyuk:
Whenever possible:
* replace `context = cast(Context, ...)` with `context: Context = ...`
* remove `cast(...)`
* remove `model` from context because it's added by `get_action`
* remove `session = model.Session` from context because it's added by `get_action`

The first two steps are transparent - they make our types more reliable. `model` and `session` are not required(as long as it's `ckan.model` and `ckan.model.Session`), because `tk.get_action` has `context.setdefault(...)` for these two properties. 

ℹ️ For now I updated the type definition of `context['auth_user_obj']`. It has `model.AnonymousUser` option added because the way we are passing `auth_user_obj` to context allows this to happen.

What does it mean:
`if context['auth_user_obj'] and context['auth_user_obj'].is_sysadmin` [now is reported as an error](https://github.com/ckan/ckan/pull/7611/files#diff-de994d21743eecf63b0983d3dd5c1254224ae84ac8d2b99a00304d0bc5d1650dR1002-R1005) by typechecker. Because `auth_user_obj` can be represented by `model.AnonymousUser` which doesn't has `is_sysadmin` property.

How it can be improved:
If all `context['auth_user_obj'] = tk.current_user` lines are removed, we can remove `model.AnonymousUser` from the type definition of `auth_user_obj` and everything will be safe once again.

### DIFF
--- a/changes/7611.bugfix
+++ b/changes/7611.bugfix
@@ -1,0 +1,1 @@
+Context requires type-casting when `model` passed explicitly.
--- a/ckan/cli/notify.py
+++ b/ckan/cli/notify.py
@@ -34,10 +34,9 @@ def send_emails():
     import ckan.logic as logic
     import ckan.lib.mailer as mailer
     from ckan.types import Context
-    from typing import cast
 
     site_user = logic.get_action("get_site_user")({"ignore_auth": True}, {})
-    context = cast(Context, {"user": site_user["name"]})
+    context: Context = {"user": site_user["name"]}
     try:
         logic.get_action("send_email_notifications")(context, {})
     except (NotAuthorized, ValidationError, mailer.MailerException) as e:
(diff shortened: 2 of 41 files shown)


pdelboca:
I love this cleanup @smotornyuk ! Both in terms of the new syntax for the type and the amount of code.

As a follow up question: I see a lot of `context: Context = {'user': current_user.name}` even when we are setting the `user` key in `_prepopulate_context()`. 

Why did you cleanup only `model` and `session`?

smotornyuk:
I really wanted to remove `user: current_user.name` as well. But when I started to do so, I thought that 
```
context: Context = {}
get_action(...)(context, ...)
```
has not much sense either and it's better to remove the whole context variable and use literal in `get_action` instead. When it did it as well, the number of changes massively increased. 

That's why if decided not to do all the changes in one go. Let's do instead a couple of smaller iterations and contributors will have a chance to solve smaller conflicts in their pending PRs.

PR #7615, opened 2023-05-29
Parse validators string using ast

smotornyuk:
When loading config declarations, try parsing arguments to the validator as python literal.

Port of [a new ckanext-scheming](https://github.com/ckan/ckanext-scheming/pull/372#event-9369278556) behavior for validators string.


tino097 (approved):
LGTM

PR #7616, opened 2023-05-29
App returns 500 error instead of 405

smotornyuk:
Requesting a page that expects a GET-request using POST method causes 500 error.

This happens because during 405 request flask doesn't set `view`/`request.endpoint`. I added simple conditions for two cases that  trigger errors


tino097 (approved):
Good one @smotornyuk !

PR #7618, opened 2023-05-29
Customize R/W datastore resource url_types

### LINKED ISSUE #7617: Customize R/W datastore resource `url_type`s
The datastore API returns an error when `datastore_create`, `datastore_upsert`, `datastore_delete` are used on resources that have a resource_type value different than `"datastore"`, unless `force=True` is passed.

This is a *good* thing for the common case of data in the datastore created by xloader/datapusher because any updated data will be lost on the next reload.

Other cases (like table designer resources) should be able to allow use without `force=True`, so I'd like to make it possible for plugins to define new resource types that allow datastore modifications.

wardi:
Fixes #7617

### Proposed fixes:

Add a datastore_rw_resource_url_types helper that lists url_type values that may be edited without passing force=True

Improve error message when force=True is required but not passed.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7617.feature
+++ b/changes/7617.feature
@@ -1,0 +1,2 @@
+datastore_rw_resource_url_types helper can be overridden to define additional
+resource url_type values that can be modified without force=True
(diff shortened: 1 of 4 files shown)


wardi:
feedback from call: use configuration option that can be updated from plugins instead. Although we still need a helper for templates to check the R/W state, and I can't think of a case of a user wanting to configure this setting in their environment/ini file. :thinking:

smotornyuk:
If you are going to use it in templates, it's justified. Initially, I got the impression, that this list of resource types is going to be used solely by this validation check

PR #7627, opened 2023-06-04
Create all activities using API, not model

smotornyuk:
Background:
All activities are created after a signal is emitted from the API action. Signal's subscriptions are executed in random order. 

Requirements:
You want to modify an activity that was created after the `package_create` action.

Problem:
You cannot just add a subscription to `package_create` and expect that this new subscription will be called after the package activity is created. And even if it did, you have to somehow get the latest activity from the package in order to do something with it.

Possible solution:
Register a chained action for `activity_create`. In this way, you will get a newly created activity right after it's get created.

Another problem:
Activities for the package are created directly via the model, without the `activity_create` action.

---

This PR changes the code of activity creation. Now all the activities are created using API action, thus you can always create a chained `activity_create` and work with any activity created by the `activity` plugin. 

⚠️ `package_create`, `package_update`, and every other action that creates an activity will raise a ValidationError if `context["user"]` is empty. Previously behavior was different:
* `package_*` and `bulk_update_*` actions used `not logged in` instead of the user reference.
* `group_*` and `organization_*` actions produced 500 error 

Alternative options:
* use site user, when if `context['user']` is empty. Don't like it because activities from the site user are hidden by default 
* allow using arbitrary string or even `None` instead of `user_id`(or only `not logged in`). It may be cool because it allows the creation of anonymous activities. For example: `Organization XXX celebrates its 10 years anniversary`, `Resource YYY was not downloaded for 3 months and automatically marked as obsolete`. But it's a brand new direction so I don't want to use it without planning. It is always simpler to add new features when you start with strict rules.
* Automatically use `not logged in` when the user is missing. It's similar to the old behavior, but it doesn't feel right. First of all, `not logged in updated the dataset XXX` - it's just silly. The second reason - activities are created inside other actions that are performed by a specific user. When it's done via WebUI, the user is always available. When it's done via CLI or tests - you can add an actor explicitly using one additional line of code. It isn't much and it makes your intentions more clear.


smotornyuk:
@pdelboca, can you check this PR with the extension you've mentioned in gitter?

You can chain `activity_create` now and do something with newly created activities like this:

```python
@tk.chained_action
def activity_create(next_, context, data_dict):
    activity = next_(context, data_dict)

    # your custom code
    if activity["activity_type"] == "changed package":
        ...

    return activity
```

I hope it will make things better because now you don't have to fetch the latest activity from DB - you'll get it right from the original action.

ℹ️ IAction iterated over in normal order. It means that you have to put the plugin that chains `activity_create` **after** activity plugin.

smotornyuk:
I added label v2.11 because I may forget to raise this issue at the meeting. As it's breaking change, we should discuss and decide what to do with this inconsistency before 2.11 is released. Otherwise, the problem will persist till v2.12

wardi (approved):

amercader:
Manually backported to dev-v2.11

PR #7630, opened 2023-06-06
reset_redis/clean_redis fixtures

smotornyuk:
Add two fixtures for removing records from Redis.

`reset_redis`: returns callable that removes records from redis on call. Accepts a redis pattern for keys to remove(i.e `*` - all keys, `ckan:default:*` - all keys prefixed by `ckan:default:`). By default removes all keys

`clean_redis`: fixture that cleans redis via side-effect(like `clean_db` or `clean_index`)

### DIFF
--- a/changes/7630.feature
+++ b/changes/7630.feature
@@ -1,0 +1,1 @@
+New `reset_redis` and `clean_redis` test fixtures for removing data from Redis.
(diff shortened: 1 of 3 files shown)


pdelboca (approved):

pdelboca:
Looks good @smotornyuk .

Could you add some entry in the Testing docs about how it works and possible scenarios to use it?

smotornyuk:
Sure, I updated docs for these two fixtures

PR #7633, opened 2023-06-06
Automatically register all factories as fixtures

smotornyuk:
At the moment, most of the factory classes are registered as fixtures and available inside tests:
```python
def test_something(user, group_factory):
    ...
```

But when a new factory class is created, it's easy to forget about the corresponding fixture and create discrepancy between different test styles.

Let's automatically register every child class of CKANFactory as a fixture, using predictable naming:
```
class SomeThing(CKANFactory): ...
```
now will automatically register two fixtures: `some_thing`(entity created with the factory) and `some_thing_factory`(factory, that can be used for creating multiple entities).


ckan/tests/factories.py
```diff
@@ -310,8 +310,8 @@ class Meta:
     type = "organization"
 
 
-class Dataset(CKANFactory):
-    """A factory class for creating CKAN datasets."""
+class Package(CKANFactory):
```
wardi, line 313: do we need two names for the same thing in tests?

ckan/tests/pytest_ckan/fixtures.py
```diff
@@ -51,61 +66,115 @@
...
+    cls = type(name, (factory, ), {})
+
+    # this function creates both fixtures
+    register(
+        # name of the factory-fixture will be obtained from `cls.__name__`
+        cls,
+        # name of instance-fixture must be passed explicitly. Strip `Factory`
+        # suffix from instance name and convert it to underscore-form.
+        inflection.underscore(name[:-7])
+    )
```
wardi, line 177: I don't love the ruby-style automatic camel case to underscore name generation. Also the comments say that all CKANFactory subclasses will be registered this way but the implementation only applies to classes in this one module.

Can we lose the magic and use a class decorator with the names we want registered instead? This could be used equally by extensions and is easily grep-able in the source code, e.g.

```python
@register_fixtures('dataset', 'dataset_factory')
class Dataset(CKANFactory):
    """A factory class for creating CKAN datasets."""
```

ckan/tests/factories.py
wardi: So you think we should switch to the 'package' naming internally in ckan. I'm not against that but it's something we should bring up in a dev meeting and get a consensus.

smotornyuk:
Refreshed in memory this change and realized, that it would not work in the way we want. Pytest and pytest-factoryboy(the library that I'm using for fixture registration) are all about magic. It's just normal for these libraries and it's the cornerstone of their philosophy.

We cannot register fixtures automatically or via *simple* decorator. Both, pytest and pytest-factoryboy require fixtures to be defined inside `conftest`/`fixtures`. And that's the reason why I wrote this code that automatically registers fixtures: to register them inside `fixtures.py`. 

Just applying the decorator inside `factories.py` will not work, because it must happen inside `fixtures.py` - that's how pytest's magic work:) And moving definitions of factories to a different file isn't worth the effort. 

So I removed this way-too-smart code and just manually registered two missing fixtures - user-with-token and sysadmin-with-token. Doing things manually is the most CKANish thing we can do in this situation. And we already have this workflow [documented](https://docs.ckan.org/en/2.10/contributing/testing.html?highlight=%40register#module-ckan.tests.factories). The whole idea of this PR was wrong, because I basically made an attempt to rewrite part of CKAN as if it was part of `pytest`.

PR #7638, opened 2023-06-10
Stricter with_plugins

smotornyuk:
Fix two controversial moments in the plugins' behavior during tests.

### Current behavior
All plugins are enabled by default. Let's assume that we have `ckan.plugins = custom` and this `custom` plugin registers `custom_helper`. You can create the following test and it will work even without the `with_plugins` fixture:

```python
def test_helper()
    tk.h.custom_helper()
```

It may seem convenient, but if the test relies on plugins, it's better to explicitly say about it by including the `with_plubins` fixture. 

And it may become much worse, if some tests use `with_plugins`, while others don't. Look at the following test sequence:
```python
def test_helper_1()
    tk.h.custom_helper()

def test_helper_2(with_plugins)
    tk.h.custom_helper()

def test_helper_3()
    tk.h.custom_helper()
```

Even though the tests are almost identical, the first and second will pass, while the third will fail. 
The first test passed because the `custom` plugin is enabled in the beginning. The second plugin will pass, because the `custom` plugin is still enabled. But the `with_plugins` fixture disables every plugin from `ckan.plugins` after the test. It means, when the third test starts, the `custom` plugin is no longer enabled and there is no `custom_helper`.

### Fix

Disable all plugins, even one specified by `ckan.plugins` config option in the beginning. It means that whenever you are going to test something that requires a certain plugin enabled, you have to use the `with_plugin` fixture. Extension maintainers have to add `with_plugins` to the tests if it's missing. 

### Current behavior.

Imagine, that you have `ckan.plugins = first second`. Plugin `first` must be loaded first because it overrides templates of the plugin `second`. Now we write some tests:
```python

@pytest.mark.ckan_config("ckan.plugins", ["first"])
@pytest.mark.usefixtures("with_plugins")
def test_first_functionality():
    pass 

@pytest.mark.usefixtures("with_plugins")
def test_combined_functionality():
    pass 
```

I told you above that all `ckan.plugins` are enabled at the beginning of tests. Keep that in mind.
The first test overrides `ckan.plugins`. It's going to test only the `first` plugin, so we don't want to spend time loading the `second` plugin here. But remember, `with_plugins` unloads plugins after the test. It means that `first` is disabled now. But the `second` is still enabled, because we've overridden `ckan.plugins` for the first test and `with_plugins` doesn't bother to disable anything outside of the current value of `ckan.plugins`.

When the second test starts, `with_plugins` rely on default value of `ckan.plugins` specified in the config file. It loads `first`, but as we still have `second` enabled, the order of plugins now changes: `second first`. And now all the overridden templates and chained actions may behave in unexpected way.

### Fix 

`with_plugins` always does the following:

* Unload **all** enabled plugins. Don't look at `ckan.plugins`. Disable every plugin which is enabled right now.
* Load only plugins from `ckan.plugins`
* After the test, disable **all** enabled plugins. Don't look at `ckan.plugins`. It will guarantee that there will be no side-effects from using `plugins.load(...)` from inside the test.

### DIFF
--- a/ckan/tests/pytest_ckan/ckan_setup.py
+++ b/ckan/tests/pytest_ckan/ckan_setup.py
@@ -54,9 +54,11 @@ def pytest_runtestloop(session):
     """When all the tests collected, extra plugin may be enabled because python
     interpreter visits their files.
 
-    Make sure only configured plugins are active when test loop starts.
+    Make sure all plugins are disabled. If test requires a plugin, it must rely
+    on `with_plugins` fixture.
+
     """
-    plugins.load_all()
+    plugins.unload_all()
 
 
 def pytest_runtest_setup(item):
--- a/ckan/tests/pytest_ckan/fixtures.py
+++ b/ckan/tests/pytest_ckan/fixtures.py
@@ -268,23 +268,57 @@ def clean_index(reset_index):
 
 @pytest.fixture
 def with_plugins(ckan_config):
-    """Load all plugins specified by the ``ckan.plugins`` config option
-    at the beginning of the test. When the test ends (even it fails), it will
-    unload all the plugins in the reverse order.
+    """Load all plugins specified by the ``ckan.plugins`` config option at the
+    beginning of the test(and disable any plugin which is not listed inside
+    ``ckan.plugins``). When the test ends (including fail), it will unload all
+    the plugins.
 
     .. literalinclude:: /../ckan/tests/test_factories.py
        :start-after: # START-CONFIG-OVERRIDE
        :end-before: # END-CONFIG-OVERRIDE
 
+    Use this fixture if test relies on CKAN plugin infrastructure. For example,
+    if test calls an action or helper registered by plugin XXX::
+
+        @pytest.mark.ckan_config("ckan.plugins", "XXX")
+        @pytest.mark.usefixtures("with_plugin")
+        def test_action_and_helper():
+            assert call_action("xxx_action")
+            assert tk.h.xxx_helper()
+
+    It will not work without ``with_plugins``. If ``XXX`` plugin is not loaded,
+    ``xxx_action`` and ``xxx_helper`` do not exist in CKAN registries.
+
+    But if the test above use direct imports instead, ``with_plugins`` is
+    optional::
+
+        def test_action_and_helper():
+            from ckanext.xxx.logic.action import xxx_action
+            from ckanext.xxx.helpers import xxx_helper
+
+            assert xxx_action()
+            assert xxx_helper()
+
+    Keep in mind, that generally it's a bad idea to import helpers and actions
+    directly. If **every** test of extension requires standard set of plugins,
+    specify these plugins inside test config file(``test.ini``)::
+
+        ckan.plugins = essential_plugin another_plugin_required_by_every_test
+
+    And create an autouse-fixture that depends on ``with_plugins`` inside
+    the main ``conftest.py`` (``ckanext/ext/tests/conftest.py``)::
+
+        @pytest.fixture(autouse=True)
+        def load_standard_plugins(with_plugins):
+            ...
+
+    This will automatically enable ``with_plugins`` for every test, even if
+    it's not required explicitely.
+
     """
-    plugins = aslist(ckan_config["ckan.plugins"])
-    for plugin in plugins:
-        if not ckan.plugins.plugin_loaded(plugin):
-            ckan.plugins.load(plugin)
+    ckan.plugins.load_all()
     yield
-    for plugin in reversed(plugins):
-        if ckan.plugins.plugin_loaded(plugin):
-            ckan.plugins.unload(plugin)
+    ckan.plugins.unload_all()
 
 
 @pytest.fixture
--- a/ckan/tests/pytest_ckan/test_fixtures.py
+++ b/ckan/tests/pytest_ckan/test_fixtures.py
@@ -33,10 +33,21 @@ def test_ckan_config_mark_without_explicit_config_fixture():
     assert config[u"some.new.config"] == u"exists"
 
 
-@pytest.mark.ckan_config(u"ckan.plugins", u"stats")
-@pytest.mark.usefixtures(u"with_plugins")
-def test_with_plugins_is_able_to_run_with_stats():
-    assert plugins.plugin_loaded(u"stats")
+class TestWithPlugins:
+    @pytest.fixture()
+    def load_example_helpers(self):
+        plugins.unload_all()
+        plugins.load("example_itemplatehelpers")
+
+    @pytest.mark.ckan_config(u"ckan.plugins", u"stats")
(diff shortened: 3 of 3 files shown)


pdelboca:
+1 on making with_plugins (and plugin load overall) more intuitive. However this change raises some questions.

Currently, our extension's template and most of the core extensions have `ckan.plugins = <extension>` in their `test.ini` file.

According to:
> Disable all plugins, even one specified by ckan.plugins config option in the beginning. It means that whenever you are going to test something that requires a certain plugin enabled, you have to use the with_plugin fixture. Extension maintainers have to add with_plugins to the tests if it's missing.

- Does that mean that all of the core extensions will need to be updated?
- What would be the use of the config `ckan.plugins` in our `test.ini` files after this change?
- While testing an extension, it's kinda obvious that the extension needs to be enabled in `ckan.plugins`, therefore, we add it in the `test.ini` file to cover the basic environment I expect to have in my tests.

Also, sometimes is inevitable to have some hard-dependency between plugins when working on integrations For example, my `customer-plugin` integrates very well with activities. Since we moved out activities to a separate plugin I had to add it to my `ckan.plugins` on the `test.ini` file. Same with scheming datasets, I needed it loaded in all my tests since my `DatasetFactory` relies on the custom schema of my `customer-plugin`.

How is it gonna work for such scenarios?

smotornyuk:
> Does that mean that all of the core extensions will need to be updated?

No, they already were using `with_plugins` explicitly. At the moment all core tests are green and I haven't changed a single line in core extensions

> What would be the use of the config ckan.plugins in our test.ini files after this change?

The same as before - we can set the default plugins via this option. If you have `ckan.plugins = a b` and include `with_plugins`, both `a` and `b` will be loaded:

```python
def test_plugins_from_config(with_plugins):
    assert plugin_loaded("a")
    assert plugin_loaded("b")
```

The only difference is that you now **have** to use `with_plugins`. Without it, all plugins will be disabled. 


> While testing an extension, it's kinda obvious ...

You are doing the right thing. Just don't forget to include `with_plugins` in integration tests so that the CKAN testing framework knows that you want to initialize the environment with plugins before the test. If you want all plugins from `ckan.plugins` to be loaded implicitly without `with_plugins` fixture, you can rely on autoloading:
```
# conftest.py
@pytest.fixture(autouse=True)
def autouse_plugins(with_plugins):
    pass
```

> ... How is it gonna work for such scenarios?

Exactly the same as before. Basically, the only thing that changed:
 whenever you test anything that requires a plugin to be loaded you **have to** use the `with_plugins` fixture.
Before, **sometimes** tests were passing even without this fixture, but it was a good idea to use it.


You don't have to treat `ckan.plugins` differently. You don't have to rewrite your tests. You don't have to change testing approach. Just include `with_plugins` if your test expect action\helper\template\auth_function\middleware\blueprint from your plugin enabled.

smotornyuk:
@pdelboca, I hope you don't mind reviewing this PR. You volunteered because you were away on a meeting:)

Just an overview of the change, because my first comment may be misleading:
 With this change, whenever you are testing functionality that is available only when some plugin is enabled, you have to use the `with_plugins` fixture.

My first comment just describes a couple of bugs, that inspired me for the PR. But it's better to ignore it because it looks like it overcomplicated things:)

pdelboca:
@smotornyuk I have been reviewing this. A couple of thoughts/questions.

1) Running `ckanext-sitesearch` tests against this branch will make most of them using plugins to fail (they are not strictly using `with_plugins` since they are defined in `test.ini` file). Test could be fixed of course to the new use of strict but should it be backported if it gonna break things/tests?
2) I personally don't like calling `plugins.unload_non_system_plugins()` when setting up the tests since it could be counterproductive by hiding side-effects of bugs. Setting up should not unset whatever is out there with a catch-all loaded function as a first step. I talk a little bit about it in the discussions (https://github.com/ckan/ckan/discussions/7561) and I think we are repeating the same pattern here. What are exactly unloading in `pytest_runtestloop` and why is something load if it is the setup?

smotornyuk:
1. Yes, I also thought about that. But the fact, that tests are running without an explicit `with_plugins` feels like a bug itself. And it may cause hidden bugs that are described in my first comment. So, it really depends on the way you see current behavior:
   * if we think that `with_plugins` should only enable missing plugins and it's ok if it leaves the environment in a somehow broken state(for example, changing the order of plugins), then the backport is not required, because this PR introduces a new feature
   * if we think that `with_plugins` means `enable plugins for the current tests` and implies that all other tests must be executed without plugins, then backport is required, because this PR is a bugfix


I do not insist on backport, but by breaking the tests, we'll help to improve them. Every test that is broken after this PR is potentially flaky and depends on the test execution order. Or even is not valid if it makes any assumptions about enabled plugins and their order.


2. As I mentioned in the first comment, the shortest way to hide side effects is to keep plugins enabled during the test if no `with_plugins` is added. 

Your main mistake is that you see every test as an interaction with a fully-functional CKAN application. It's not like that. Mainly, we are writing unit tests. Sometimes, module or integration tests. Rarely functional tests. What you expect is really required only for the last group, for the functional tests. For the generic unit test, we don't need a working CKAN app. As you are testing only a small piece of functionality, you enable only the required parts of the application via fixtures. If you save something to the database, you add `clean_db`. If you need plugins, you add `with_plugins`. If you want to test only a small helper, you(likely) don't need the whole application. 
Ideally, it should be possible to run a single test even if the application is broken, as long as the test itself does not use the application. For example, if you have a function `def sum(a, b): return a + b`, it should be testable without the proper `sqlalchemy.url` or even without the CKAN config file. But historically CKAN tests always were tightly coupled to the CKAN app, and it was too hard to rewrite all the tests into "correct" form. So we added a quick health check at the beginning of the test session, making sure that it's possible to create a CKAN application. Then we try to disable as much as possible, to make all tests initially isolated and independent(that's where we are unloading all the plugins). And after that, we are running the tests, expecting that every test can clearly specify its requirements using fixtures.

During tests, the config file describes the state of **enabled** features. I.e, if you are using `with_plugins`, the config file will tell you, what plugins are enabled. If you are not using something, don't make assumptions about it. Don't expect that some specific plugins are loaded if you don't use `with_plugins`.

amercader:
This is a tricky one, because the changes you suggest make obviously a lot of sense and make the tests much more predictable, but as we discussed, they do cause quite a lot breakage (as in tests failing)
I tested the patch in a few extensions and spatial was fine, but ckanext-dcat and ckanext-harvest had many failures because they rely on other plugins being loaded and they were written assuming "if they are in the test.ini file then we are fine".

Maybe writing tests that rely on other plugins is not that common, and of course we can fix the official extensions before 2.10.2 is released, but I still feel that this might be too disruptive for a patch release.

Two questions:
* Once this is merged and on 2.11, extensions that support 2.10 and 2.11 will have to adapt the tests anyway to get them passing in both versions right?
* Does this change mean that the `ckan.plugins` setting in test.ini is completely ignored?

PR #7647, opened 2023-06-13
Conditional DataStore Tab for Datapusher

JVickery-TBS:
feat(templates): conditional DataStore tab;

- Added helper for valid datapusher formats.
- Added condition to display the DataStore tab in Resource Edit.

### Features:

Conditions the DataStore tab in resource edit view for the Datapusher extension. Only show the tab if the resource is a supported datapusher format, or if the resource already has an existing datastore table (is datastore active).

- [ ] includes API changes

### DIFF
--- a/ckanext/datapusher/helpers.py
+++ b/ckanext/datapusher/helpers.py
@@ -29,3 +29,10 @@ def datapusher_status_description(status: dict[str, Any]):
         return captions.get(status['status'], status['status'].capitalize())
     else:
         return _('Not Uploaded Yet')
+
+
+def is_datapusher_format(resource_dict: dict[str, Any]):
+    supported_formats = toolkit.config.get('ckan.datapusher.formats')
+
+    return (resource_dict.get('format', '').lower() in supported_formats
+            and resource_dict.get('url_type') != u'datapusher')
--- a/ckanext/datapusher/plugin.py
+++ b/ckanext/datapusher/plugin.py
@@ -151,6 +151,7 @@ class DatapusherPlugin(p.SingletonPlugin):
             u'datapusher_status': helpers.datapusher_status,
             u'datapusher_status_description': helpers.
             datapusher_status_description,
+            u'is_datapusher_format': helpers.is_datapusher_format,
         }
 
     # IBlueprint
--- a/ckanext/datapusher/templates/package/resource_edit_base.html
+++ b/ckanext/datapusher/templates/package/resource_edit_base.html
@@ -12,5 +12,7 @@
 
 {% block inner_primary_nav %}
   {{ super() }}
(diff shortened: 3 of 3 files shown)


JVickery-TBS:
Cypress failed in last workflow. Could someone re-run the workflow for me? Does not look like the Cypress failed because of code in PR.

wardi:
for xloader/datapusher we also need to watch out for datastore-first resources. Any time the `url_type` is set to one of these values we should hide the tab as well (and ideally refuse to run automatic loading for these resources).

There's a new `datastore_rw_resource_url_types` helper that returns a list of the datastore-first url_type values:
https://github.com/ckan/ckan/pull/7618/files

JVickery-TBS:
@wardi so I can just change condition to `( h.is_datapusher_format(res.format) or res.datastore_active ) and ( res.url_type not in h.datastore_rw_resource_url_types() )` in the template, and that should be good?

JVickery-TBS:
As suggested by @ThrawnCA I have added the DataStore action link along with the other action links in the `resource_read` template.

Added `datastore_rw_resource_url_types` helper to the conditions for the edit tab and the manage link button.

ckanext/datapusher/templates/package/resource_read.html
```diff
@@ -0,0 +1,12 @@
+{% ckan_extends %}
+
+{% block action_manage %}
```
wardi, line 3: try inserting into the front of `action_manage_inner` so that it appears in the same order as the tabs when editing.

wardi: add similar links in `package/snippets/resource_item.html` at the front of `resource_item_explore_inner` and `package/snippets/resources.html` at the front of `resources_list_edit_dropdown_inner` so that this tab appears along with the others for faster navigation.

These were recently added as part of #7590

JVickery-TBS: @wardi cool! The conditional action link has been added to these template snippets in datapusher ext now. And `action_manage` updated to `action_manage_inner`

ckanext/datapusher/helpers.py
```diff
@@ -29,3 +29,9 @@ def datapusher_status_description(status: dict[str, Any]):
         return captions.get(status['status'], status['status'].capitalize())
     else:
         return _('Not Uploaded Yet')
+
+
+def is_datapusher_format(resource_format: str):
```
ThrawnCA, line 34: This helper could be expanded to cover the other two checks, `res.datastore_active` and `res.url_type not in h.datastore_rw_resource_url_types()`, since they're always used together.

JVickery-TBS: Helpers and conditions combined into one helper.

ThrawnCA (approved):

PR #7658, opened 2023-06-16
`required` flag adds `not_empty` validator to option

smotornyuk:
Automatically add the `not_empty` validator to any config option declared with `required: true`.


pdelboca:
I'm a little bit confused about this two options:
- What's the difference between `required` and having a `not_empty` validator? What are we trying to achieve by having this two?
- What happens if a config option has `required: false` and `not_empty` validator?

smotornyuk:
`required` is an idiomatic way to declare a mandatory option. It must be added to the option if the CKAN application cannot even start without it. Consider datastore - read and write DB connection strings are `required` because the datastore will break CKAN if they are missing.

If the option has `default`, it should not be `required`. Because CKAN can work even if the option is missing, because of the `default` value. But you can add `validators: not_empty` just to ensure, that user won't override the `default` with an empty string using the config file. Of course, it's up to the person who defines the declaration, but I think that `default` is the opposite of the `required`.

`validators: not_empty`, just as mentioned above - it's a good choice for an option, that should not be empty, as long as it's hard to make it empty(because of the `default`). If an option cannot be empty, but CKAN works even if it's not present in the config file, you are likely will use `validators: not_empty` instead of `required: true`

`required: true`, is similar to `type: int`. The latter just adds `int_validator` to the option. Look at the following definitions:
```yaml:
# Idiomatic
- key: some_number
  type: int
 
# Low level
- key: some_number
  validators: int_validator
```

They are the same, but the first one looks much friendlier. The same goes for `required`:
```yaml

- key: important_option
  required: true

- key: important_option
  validators: not_empty
```

The first definition takes your attention and puts an accent on the fact, that this option is required.
Second, just notifies you, that option should not be empty, but it's not important otherwise. They are the same, but the `required` flag helps you to express your intent.


The extra feature of the `required` flag: `ckan config declaration --core --minimal` prints an example of a minimal configuration that is required to run CKAN. It just shows all the options that have required flag.

---

> What happens if a config option has required: false and not_empty validator?

It will raise a validation error because of the `not_empty` validator. Every option that does not have an explicit `required` flag is `required: false`. `required` does not disable validators, it simplifies them and allows you to write a simpler definition. Have you worked with `ckanext-scheming`? Fields in ckanext-scheming also have a `required` flag and technically you can write in the dataset schema something like:
```yaml
  required: false
  validators: scheming_required not_empty
```
and it will work in a similar manner. I like the way it's done in ckanext-scheming and I assume that a lot of people are familiar with this extension, so I tried to make config declarations more or less similar to scheming, instead of making the learning curve steeper.

smotornyuk:
Yep, your final statement/question is absolutely correct. (don't forget that it's my vision, and I can be globally wrong 😆 )

Config declarations rather describe the config file than CKAN internals. From the config file's point of view, if an option has a default value, it's not required, because you don't have to add it to the config file. 
For CKAN internals we have validators:
* `validators: not_empty` without `required` tells that the config option is not important from the config file perspective(probably, because of `default`), but CKAN cannot work without it.
* `required: true` + `validators: not_empty` explicitly informs us that the config file is not complete without the option **and** CKAN cannot work without it. But it's too verbose.
* `required: true` without validators tells us that the option **MUST** be added to a config file and without it, the config file will be invalid. We can safely assume, that CKAN will not work with invalid config files and omit validators.

pdelboca:
I still don't like this approach, if we consider that "explicit is better than implicit" this does the opposite. We are implicitly adding validators.

@amercader I think we need a third vote for this one :)

wardi:
fwiw ckanext-scheming uses a `scheming_required` validator https://github.com/ckan/ckanext-scheming/blob/8646a9dce79aa0b5a46274271ca6330dc0870b92/ckanext/scheming/validation.py#L83-L91 that reads the configuration of the field and applies not_empty based on that. We could use something similar.. `config_required` that is included if no validators are listed but needs to be added by the user if they customize the validator list (if the user wants the required actually enforced)

smotornyuk:
In scheming `sheming_required` is convenient, because validators and field definitions can live in different places and be reused via snippets. Whenever a snippet defines validators, it makes sense to add `scheming_required`. In this way, whoever uses your snippet can write:
```yaml
 required: true # or false
 snippet: xxx
```
. And, of course, it's convenient as a default validator, when no validators are specified.

 But imagine the following definition:
```yaml
  required: true
  validators: scheming_required
```

What's the point of using the `scheming_required` validator together with the `required` flag in this case? (apart from marking filed as required in HTML). Why introduce another validator when we can just add `not_empty` and omit the `required` flag? And that's what we have in config declarations: `required` and `validators` always come together, we don't have snippets here. If we are going to explicitly specify `config_required` every time when `required: true` is added, it's better not to use `required: true` at all and just add `not_empty`.  

And there is another reason to use auto-modified validators. Look at this example:
```yaml
 type: list
 required: true
```
If we going to enforce manual addition of `not_empty` when `required: true` is set, then shouldn't we also enforce applying validators from `type: list` as well? So, the above definition must look like this:
```yaml
type: list
required: true
validators: not_empty aslist
```

Now, when we have validators added explicitly, both `type` and `required` don't have any value and we can just remove them to simplify the definition. 
Basically, if someone needs just validators - he can use them. If someone wants to use just `required` without validators, it should be also possible and must have some useful effect. Finally, if a person uses both `required` and `validators` at the same time, it should have a combined effect of two attributes. I.e: `required` enforces non-empty value, `validators: aslist` enforces a list. Together they must produce a rule that requres non-empty list.

---

@pdelboca, I don't agree with your explicit/implicit reference here. Do you think that tracking and activities, are bad, because they are implicit? According to your last statement, they are. Because we are not calling explicitly `activity_create` inside `package_create` and not making any direct requests into the tracking table whenever a user visits a page. 

But I think, you agree, that activities are explicit, even though they are created elsewhere. Because we have documentation of activity plugin, which says, that enabling `activity` creates a record for every package change. So, documentation is the point. And if we explicitly write that `required: true` adds `non_empty` validator, this feature will be explicit as well, right?

pdelboca:
@smotornyuk tracking activities are not implicit since we are explicitly calling for its creation using signal.

However, in this change we are setting a new validator in a `get` method. So, this method is not a simple `get_validators` it wil also set a new one if a condition is met. I do not expect a `get` method to `set` things if some condition is met. This is the kind of implicit logic that I think makes the code more difficult to follow and maintain.

```python
 def get_validators(self) -> str:
        """Return the string with current validators.
        """
        return self.validators
        validators = self.validators
        if self.has_flag(Flag.required) and "not_empty" not in validators:
            validators = f"not_empty {validators}"

        return validators
```

Anyhow, if others seems okay with this change we can go for it.

smotornyuk:
I see your point. It's not exactly describes situation, because `get_validators` does not "set" anything. After you call it, `validators` property remains unchanged. It behaves rather like [`Package.resources`](https://github.com/ckan/ckan/blob/master/ckan/model/package.py#L153-L156) - it takes certain collection) and alters it so that value is "correct".
```python
    def resources(self) -> list["Resource"]:
        return [resource for resource in
                self.resources_all
                if resource.state != 'deleted']
```
In `Package.resources` we are not returning literally all resources from the package. We are only returning whatever has most sence - active resources.
`get_validators` returns the list of validators and ensures that this list is correct. Namely, if option has a `required` flag, `get_validator` ensures that option cannot be empty. 
If we don't do it, can we say that option that has `required` flag but don't have `not_empty` validator is correct? From definition it looks like CKAN doesn't work without this option(because it's required), but if `not_empty` validator is missing, empty/missing value may actually slip into CKAN.

Do you have other suggestions? One from Ian would work if we have only `required` and `validators`. But because we have `type` on top of it, which also implies certain validators, situation becomes more complex. And if we are going to be 100% explicit, we have to disallow using `type` and `required` whenever `validators` is present, so that user is forced to set all the expected validators manually.

wardi:
I was thinking that the default set of validators when not given could be something like:
```yaml
validators: config_default config_required config_type
```
where these validators do different things based on the `default`/`configured_default`, `required` and `type` settings in the config declaration. This should mean that the majority of config declarations can go without specifying validators at all, and if validators are specified they could use some of these and replace others as preferred

smotornyuk:
Wouldn't it impose too many low-level details on developers? Even though, the better people know the system, the better it's for them, here additional knowledge wouldn't have any benefit. People will learn the details of declaration implementation, but they won't be able to use this knowledge anywhere. If we hide these implementation details, the tool will be simpler and whoever defines the option will be able to do the same things spending less time on learning.

In this way, people have to know what are default validators and keep an eye on them when changing the declaration. On top of this, it binds us in the future if we decide to change implementation details.

Let's forget about the fact, that `required` and `type` imply some validators and alter the actual `validators` property. Look at the feature as a user who knows no implementation details:

* Config declaration accepts the `validators` attribute. Whatever validators are mentioned there, they will be applied to the option during the validation stage.
* Config declaration accepts the `required` flag. Whenever this flag is enabled, CKAN won't start when the option is empty(forget about validators - CKAN just will not start, no matter what prevents it from starting)
* Config declaration accepts the `type` option. Whenever this option is provided, CKAN guarantees that in runtime the value of the option will have the specified type.

That's it. Imagine that these are 3 different parameters, that are not related to each other. You don't need extra details to use them - each of them is independent and you can master only one and never use others. We are not uncovering implementation details, so people won't rely on it and we can change implementation at any moment in the future. From the developer's point of view, there is no magic - the code is pretty clear if you take a look at the `get_validators` method.


This is my last argument, if it doesn't work, I give up:)

wardi:
@smotornyuk I think your argument is good. Your approach would also let us remove the `validators` from almost all definitions which I like better than requiring developers to know all the arcane details of our validators.

amercader:
I agree that the benefits of simplifying the config declaration outweigh the little magic involved in applying certain validators under the hood so I'm +1  on merging this

Issue #7671, opened 2023-06-22
Clean-up of configuration options


pdelboca:
### Intro

When working on #7662 I ran into some configs like: `search.facets.limit` or `search.facets.default` that didn't make a lot of sense to me. What's the difference between having 5, 7, 8? and What's the value of having a whole config for something that can be easily override in a template.

After that, while reading a book I found this chapter about [Avoid Preferences](https://basecamp.com/gettingreal/06.4-avoid-preferences) that really resonated with my experience while working with CKAN.

Taking into consideration that there's a difference between building a product and a framework I do think that we are over-using preferences to avoid difficult decision. 

### Point

I think CKAN would greatly benefit from lesser configuration options, both in terms of complexity and maintainability. Even more in the area of the front-end where we have preferences whose effect can be also achieved by overriding the template (a common practice in CKAN) 

### Proposal

Let's get rid of some of our configurations and lets make CKAN more lean in terms of it's codebase. Let's discuss which configuration options make us happy and make CKAN a good framework and which ones just doesn't make any sense.

### Proposed configurations to remove:
 - [ ] `search.facets.limit` and `search.facets.default`
 - [ ] `ckan.datasets_per_page`
 - [ ] `ckan.dumps_url`
 - [ ] `ckan.dumps_format`
 - [ ] `ckan.featured_groups`
 - [ ] `ckan.featured_orgs`
 - [ ] `ckan.default_group_sort`
 - [ ] `ckan.homepage_style` (why do we have 3 options?)
 - [ ] `ckan.template_head_end`
 - [ ] `ckan.template_footer_end`
 - [ ] `ckan.template_title_delimiter` What's the goal of this one?
 - [ ] `ckan.admin_tabs`
 - [ ] `ckan.user_list_limit`
 - [ ] `ckan.activity_list_limit`
 - [ ] ckan.activity_list_max`
 - [ ] ...
 - [ ] ...

Which preferences would you think it makes sense to remove?

wardi:
we could render template titles with a helper the same way we do many other titles, that would let us override this delimiter, provide smart translations and more

pdelboca:
@Pradyumn-NEC are you interested in working on some of this configs?

`ckan.default_group_sort` should be quite straight forward to remove. Do no forget to add a CHANGELOG file mentioning that this config has been removed will no longer be available.

Pradyumn-NEC:
[...]

PR #7673, opened 2023-06-23
Remove dataset dumps config options

pdelboca:
Fixes some of the issues in #7671 

### Proposed fixes:
This removes some old and no longer needed configuration options.


changes/7673.removal
```diff
@@ -0,0 +1 @@
+`ckan.dumps_url` and `ckan.dumps_format` config options have been removed.
```
wardi, line 1: add a note about extending package/search.html to achieve the same effect

pdelboca:
Done @wardi !

PR #7681, opened 2023-06-29
`ckan db upgrade` fixes

amercader:
I'm working on an upgrade from CKAN 2.5 to 2.10 and found errors that can be solved at the CKAN level

The most potentially controversial is the first commit (8d0d3b36fe61deb23ffcd24a3350fd38e7c5b467). Basically when calling the magic `get_site_user` action when starting up `ckan db upgrade` we get an error because the python model defines `User.last_active` and that is not present in the database (was introduced on CKAN 2.10). I've followed the same approach that we do to avoid race conditions on that particular situation and added an exception handling.

The rest of changes related to the separate script used to migrate the old revisions to activities and should be hopefully self-explanatory:
* [Ensure activity plugin is loaded before migrating old revisions](https://github.com/ckan/ckan/commit/888ac2770eb7ceaf0ceccc239475bf896114857c)
* [Support new plugin_data field in revision migrate script](https://github.com/ckan/ckan/commit/854fb6b19e0a7965ea5dad89a058ddd4a0106d2e)
* [Remove no longer supported SQLAlchemy key in revision migrate script](https://github.com/ckan/ckan/commit/5f48dc8868fab0012e9e696286a8869ffb0190b2)

I think anything that makes life easier for those trying to upgrade it's a good think, so I'm proposing to backport this

### DIFF
--- a/changes/7681.misc
+++ b/changes/7681.misc
@@ -1,0 +1,1 @@
+Fix errors when running the `ckan db upgrade` command
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -249,6 +249,11 @@ def update_config() -> None:
     if user_table_exists:
         try:
             logic.get_action('get_site_user')({'ignore_auth': True}, {})
+        except sqlalchemy.exc.ProgrammingError as e:
+            if "UndefinedColumn" in repr(e.orig):
+                log.debug("Old user model detected")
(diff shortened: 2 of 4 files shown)


amercader:
Unrelated failures fixed now

note: changes in environment.py don't apply to 2.10 and 2.9

PR #7685, opened 2023-06-30
Adding htmx to CKAN 🚀  (With an example)

### LINKED ISSUE #7524: Hypothesis: embracing an hypermedia approach could lead to an easier transition to a richer UX.
### Problem statement

Current front-end, while functional and extensible by plugins, have become tedious to maintain and work with.

### Hypothesis

Given that:
 - Lots of CKAN sites are maintained by small teams or even individuals, and
 - is fair to say that there is a significant movement of influential people raising concerns about client based apps and SPAs shortcomings and about them being overuse (see https://hypermedia.systems/ and https://hotwired.dev/), and
 - that a decoupled CKAN could handle this approach as one of the options for frontend technologies, then

It should be possible for CKAN to adopt an hypermedia approach to build a rich UX as a part of CKAN 3.0.

### Discovery work

It would be possible to code a little demo about integrating HTMX in CKAN to evaluate the feasibility of the approach. (There is already some tests ongoing: https://github.com/pdelboca/ckan/tree/demo-htmx) 

### Validation

Why the solution is trustworthy? It is build on already proven technologies and our requirements are not as complex to require a SPA. (Example: We do not need to build an alternative to Google Drive.)

What makes it strong? 
 - It would provide an easier transition (it is build on already existing stack of technology)
 - There are concise metrics on migrations to this technologies: https://htmx.org/essays/a-real-world-react-to-htmx-port/
 - Experiences suggest that this kind of technologies are good for small teams (https://www.youtube.com/watch?v=3GObi93tjZI)

Questions to consider:

- Is this change going to break current installations? **TBD**

- Can we provide a backwards compatibility? Yes

- How easy is gonna be for current implementations to migrate to this new release? **TBD**

- Do current versions of CKAN have the adequate resources/support to migrate to this new version? **Yes**

- Are we going to change the database schema? **No**

- Are we going to change the API? **No**

- Are we going to deprecate Interfaces? **No**

pdelboca:
Fixes #7524 

## Description

This PR integrates [htmx](https://htmx.org/) to CKAN  and refactors the **Follow**/**Unfollow** logic to have a clear and concrete example of how we can iteratively refactor our code to get all the benefits of HTMX.


changes/7685.feature
```diff
@@ -0,0 +1,3 @@
+Add htmx webasset to CKAN. For more information check the `documentation <https://docs.ckan.org/extensions/htmx.html>`_.
```
ThrawnCA, line 1: This URL is 404 Not Found for me.

pdelboca: Yes @ThrawnCA because it is a new documentation page that only exist in this PR :)

pdelboca: I'll try to add a relative link., though I'm not sure if they are available in the CHANGELOG files

doc/extensions/htmx.rst
```diff
@@ -0,0 +1,71 @@
...
+
+To make a request using htmx, you can use the ``hx-*`` html properties defined by
+the library. Check the `htmx examples <https://htmx.org/examples/>`_ for an
+overview of patterns and uses.
+
+----------------
+2. htmx in CKAN
+----------------
+
+For an exaple of how we are using htmx in CKAN, check the **Follow** / **Unfollow**
```
ThrawnCA, line 42: exaple -> example

doc/extensions/htmx.rst
```diff
@@ -0,0 +1,71 @@
...
+logic.
+
+The **Follow** button is a simple link that makes a POST request to the
+``/dataset/follow/<dataset-id>`` and then replaces the section ``package-info`` with
```
ThrawnCA, line 46: Should have the word "endpoint" after the URL.

ThrawnCA (changes requested):

ckan/views/dataset.py
```diff
@@ -893,60 +896,58 @@ def get(self, package_type: str, id: str) -> Union[Response, str]:
...
 
-def follow(package_type: str, id: str) -> Response:
-    """Start following this dataset.
-    """
-    context: Context = {
-        u'user': current_user.name,
-        u'auth_user_obj': current_user
-    }
-    data_dict = {u'id': id}
+def follow(package_type: str, id: str) -> str:
```
amercader, line 899: Perhaps it would be nice to clearly mark the "htmx view functions" so it's easy to see that these functions don't return a response but a string with the rendered html. 
For instance we could create an `@htmx` or `@for_htmx` decorator. That could do nothing for now, but in the future maybe it could check if the request has an htmx header (assuming the view only works with htmx). Or if the view can also render a full page we can define the template to be used if it's an htmx request (like this [pattern](https://github.com/spookylukey/django-htmx-patterns/blob/master/separate_partials_single_view.rst))

Or we could just add it to the docstrings, but a decorator feels like more flexible

amercader: Speaking of this,  do you think we will have single views handling htmx and regular requests?

doc/extensions/index.rst
```diff
@@ -41,3 +41,4 @@ features by developing your own CKAN extensions.
    translating-extensions
    flask-migration
    signals
+   htmx
```
amercader, line 44: I'd probably move this page to the Theming section, maybe after the Customizing CKAN's javascript page (which probably needs to be reviewed to mention the new htmx patterns)

doc/extensions/htmx.rst
```diff
@@ -0,0 +1,71 @@
+htmx
```
amercader, line 1:
```suggestion
Creating dynamic user interfaces with htmx
```

doc/extensions/htmx.rst
```diff
@@ -0,0 +1,71 @@
...
+    Events directly in HTML, using attributes, so you can build modern user
+    interfaces with the simplicity and power of hypertext."*
+    -- `htmx.org <https://htmx.org/>`_
+
+While not all CKAN templates have been updated to use htmx, you can use it
+in your own extensions to build modern user interfaces. htmx will be the core
+component in the implementation of the new CKAN UI, so you should expect more
+of it in future versions.
+
+---------
```
amercader, line 16: I think these docs are good primer but we need to provide a bit more guidance.

An Overview section should give a general overview of how htmx works (hx- attributes in templates, background requests to retrieve updated HTML code that gets replaced in the page, etc)

Next, an Implementation section should go step by step explaining the different parts you need to implement, and we can use the Follow/Unfollow example for that but rather than including lots of code without much comment, explaining a bit more what is happening at each step. The `request.htmx` reference is less important and should be included last. For instance, when you added:

> The **Follow** button is a simple link that makes a POST request to the
``/dataset/follow/<dataset-id>`` and then replaces the section ``package-info`` with
all the HTML returned by the endpoint.

We can add:

"Note how we use the ``hx-post`` and ``hx-target`` attributes in the HTML tag of the link to configure htmx to perform these actions"

Just to point the user to the actual changes. I don't think we need to include the whole snippet in the docs, and on the view function we just need to highlight that we are returning a renderer template instead of a response.

As we use different approaches like a single view function handling htmx and regular requests we can add these too to the docs.

I'm happy to help writing these.

amercader (commented):
Great work @pdelboca. This looks really straightforward and good to go, we just need to polish the documentation a bit

ckan/views/dataset.py
```diff
@@ -448,9 +447,6 @@ def read(package_type: str, id: str) -> Union[Response, str]:
             _(u'Dataset not found or you have no permission to view it')
         )
 
-    g.pkg_dict = pkg_dict
-    g.pkg = pkg
```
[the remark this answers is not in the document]
pdelboca: Reverted!

gulpfile.js
```diff
@@ -83,18 +83,10 @@ const qs = () =>
...
-    dest(__dirname + "/ckanext/textview/theme/public/vendor/")
-  )
-
-const highlightJsStyles = () =>
-  src(__dirname + "/node_modules/@highlightjs/cdn-assets/styles/a11y-light.min.css").pipe(
-    rename("a11y-light.css")).pipe(
-    dest(__dirname + "/ckanext/textview/theme/public/styles/")
-  )
-
-
```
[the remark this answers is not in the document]
smotornyuk: Initially, gulpfile was used for compiling less/scss into standard CSS. Later we added a hack, that copies updated jquery/higlights/bootstrap/etc. into CKAN vendor folders and that's exactly what you see here.

I'm working on a slightly different approach in #7691. IIRC, these lines that @pdelboca is removing are not working anyway, so there is no harm in removing them. Parts that are affected by these non-working lines will be fixed in #7691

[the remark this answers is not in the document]
smotornyuk: PS Moving updated libraries using gulpfile and, on top of this, doing this implicitly, relying on `npm`'s rules that are not very familiar to everyone is probably a wrong decision.
We discussed it briefly with Ian and agreed that using Makefiles for such kinds of tasks may be more efficient. In addition, it will give us a standard tool for adding maintenance commands, like transifex/pypi interaction

ckan/views/dataset.py
smotornyuk: We are using the following helper in our projects for new views that allow partial rendering with htmx:

```python
 def render(template: str, data: dict[str, Any] | None = None) -> str:
     """HTMX-aware version of tk.render.

     If a current request is made by HTMX, render a template with the same path from
     `htmx/` folder.
     When `HX-CKAN-Role` is present, apply it as a secondary extension. Use this 
     header if the single view can produce multiple partials.

     """
     if is_htmx_request():
         path, ext = os.path.splitext(template)

         if role := tk.request.headers.get("hx-ckan-role"):
             role = f".{role}"

         template = os.path.join("htmx", f"{path}{role}{ext}")

     return tk.render(template, data)
```

And, if you are using it in view instead of `tk.render`, the following happens when you are calling `render("group/read.html")`, for example:

* it it's a non-htmx request, `tk.render` is called as is 
* if it's a htmx request with no extra headers, `htmx/` added before the template name and we are rendering `tk.render("htmx/group/read.html")`. In this way, you can support rendering pages in the normal way and via HTMX. Basically, that's what @pdelboca is doing for the dataset search page.
* if there is an `hx-ckan-role: <ROLE>` header, we are using `<ROLE>` as a secondary extension, between the template name and template extension: `name.<ROLE>.html`. If HTMX request contains `hx-ckan-role: followers` header when requesting the group details page, the following template is rendered: `tk.render("htmx/group/read.followers.html")`. In this way, the view can update the main part of the page like in the previous point + update a number of separate segments(sidebar, header, footer)  separately.

pdelboca: I would go for a more explicit @for_htmx('template.html'') since it is more readable. It's true that the if pattern will appear a lot and the solution @amercader proposes seems logic.

Making `render` aware of `HTMX` and internally handling the logic makes it less transparent to the user. Is one of those double behavior function that are not explicit in the name and can be hard to follow up. Is a function in which the behavior does not change due to parameters but rather to a global object it does not have any control over. (Event when Flask uses Global as a pattern, I'm not always comfortable with it.)

@amercader regarding your second question. Using HTMX assumes that we are Okay with the principles of returning pieces of HTML to replace elements. Once you are okay with that principle, the library recommends to be [as specific as you want with your endpoints](https://hypermedia.systems/building-a-contacts-app-with-hyperview/#_hypermedia_notes_api_endpoints) since an endpoint will exist only to solve a particular UX/UI need on a given time. (Instead of a generic `package_show` that everyone can reuse). 

I will expect with time to have both a mix of views with @for_htmx decorator (that are useful in less dynamic views where we just need to refresh content) and several specific views with HTML snippets that we will use to make the UX more component oriented.

doc/extensions/index.rst
pdelboca: Done!

doc/extensions/htmx.rst
pdelboca: @amercader check the new changes and let me know if you like them!

ckan/common.py
```diff
@@ -147,13 +147,61 @@ def _get_request():
     return flask.request
 
 
+class HtmxDetails(object):
```
wardi, line 150:
```suggestion
class HTMXDetails(object):
```
would follow the python naming conventions better, e.g. https://docs.python.org/3/library/html.parser.html?highlight=html#html.parser.HTMLParser

wardi: unless htmx isn't initials for something... :thinking:

ckan/views/dataset.py
wardi: I like the decorator idea.

For many views it will make sense to have the same code run for htmx and non-htmx version, and this can be explicit in the view checking just before rendering.

For the htmx-only views an `@htmx_only` decorator would be nice for automatically 404ing non-htmx requests.

ckan/views/dataset.py
```diff
@@ -893,60 +896,58 @@ def get(self, package_type: str, id: str) -> Union[Response, str]:
...
-        h.flash_success(
-            _(u"You are now following {0}").format(package_dict[u'title'])
-        )
+        error_message = str(e.error_dict['message'])
+
+    extra_vars = {
+        'pkg': package_dict,
+        'am_following': am_following,
+        "current_user": current_user,
+        "error_message": error_message
```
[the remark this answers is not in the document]
wardi: Yeah we should make an effort to write tidy code or @smotornyuk is going to come along and enforce black as a pre-commit hook on all of us :smile:

doc/theming/htmx.rst
```diff
@@ -0,0 +1,156 @@
...
+of it in future versions.
+
+--------
+Overview
+--------
+
+``htmx`` is a library that allows you to use HTML attributes to make AJAX requests
+and update the DOM. It is a great alternative to Javascript frameworks like
+React or Vue, as it allows you to build dynamic user interfaces without writing
+any Javascript code.
```
wardi, line 23:
```suggestion
React or Vue, as it allows you to build dynamic user interfaces with regular flask
views and Jinja2 templates, allowing templates to be overridden by themes and other extensions. 
```

doc/theming/htmx.rst
```diff
@@ -0,0 +1,156 @@
...
+
+The library is very simple to use. You just need to add the ``hx-*`` attributes
+to your HTML elements to make them dynamic. For example, to make a link that
+makes a POST request to the ``/dataset/follow/<dataset-id>`` endpoint and
+replaces the HTML element with id ``package-info`` with all the HTML returned by
+the endpoint, you can do:
```
wardi, line 29:
```suggestion
the endpoint, you can write:
```

doc/theming/htmx.rst
```diff
@@ -0,0 +1,156 @@
...
+  2. A way to trigger a call to the endpoint when the user clicks on the button and replace the element with the new content.
+  3. A new endpoint that covers the backed logic and returns just enough HTML to replace the HTML element.
+
+
+1. HTML structure
+
+The HTML structure is very simple. We just need a div that contains the button and the counter. In our case, and
+to respect the current CKAN UX, we have the ``package/snippets/info.html`` snippet that is used to render this
+information. We just need to make sure that the HTML element that we want to replace has an id. In our case,
+we just add and id to it: ``package-info``.
```
wardi, line 74:
```suggestion
The HTML structure is very simple: a div that contains the button and the counter.
To respect the current CKAN UX we update the ``package/snippets/info.html`` snippet that is used to render this
information. We need to make sure that the HTML element that we want to replace has an id so
we add and id to it: ``package-info``.
```
I started by removing the 3 "just"s and then kept editing :smile:

doc/theming/htmx.rst
```diff
@@ -0,0 +1,156 @@
...
+In addition to the ``hx-post`` attribute, we also need to define the ``hx-target`` attribute. The ``hx-target``
+attribute defines the HTML element that will be replaced with the HTML returned by the endpoint. In our case,
+we want to replace the ``package-info`` element, so we can use the ``#package-info`` selector.
+
+3. The endpoint
+
+The last step is to implement the endpoint that will be called when the user clicks on the button. In our case,
+we want to call the ``/dataset/follow/<dataset-id>`` endpoint. This endpoint is already implemented in CKAN.
+We just need to make sure that, under this new context, it should return only the HTML that we want to display in the page
+instead of rendering again the whole dataset page.  We achieve that by making it sure that we return the snippet that
```
wardi, line 112:
```suggestion
We need to make sure that, under this new context, it should return only the partial HTML that we want to insert into the page
instead of rendering the whole dataset page again.  We achieve that by making it sure that we return the snippet that
```

doc/theming/htmx.rst
```diff
@@ -0,0 +1,156 @@
...
+
+    from ckan.common import request
+
+    if request.htmx:
+        # do something
+
+
+Calling ``request.htmx`` will return a HtmxDetails object that contains attributes
+for each one of the ``htmx`` attributes. For example, if you want to access the
+``hx-target`` attribute, you can do::
```
wardi, line 140:
```suggestion
``hx-target`` attribute, you can write::
```

gulpfile.js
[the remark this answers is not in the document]
wardi: Using make may not be more efficient but at least one person (me) will be less confused by them :smile:

changes/7685.misc
```diff
@@ -0,0 +1 @@
+`snippet/organization.html` has been moved to `organization/snippets/info.html` for consistency with Groups/Packages/Users.
```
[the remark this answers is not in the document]
pdelboca: Good catch!

In `master` both `snippet/group.html` and `group/snippets/info.html` exist. That's why I only needed to move `snippet/organization.html`.

I have been checking `snippet/group.html` but it is not used anywhere so I will remove it.

ckan/common.py
pdelboca: `htmx` are not initials and it is always referred as `htmx` (lower case) in comparison with `HTML`.

Their docs always refer as `htmx`. Tha's why I used `HtmxDetails`. That's how `django-htmx` calls it so I'm following them as well: https://django-htmx.readthedocs.io/en/latest/middleware.html

ckan/views/dataset.py
pdelboca: What should we do with this suggestions? I like the ideas but on the other hand I'm reluctant to start implementing tools for it before we actually start using it.

Would it be a better fit for another PR?

ckan/views/dataset.py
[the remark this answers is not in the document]
pdelboca: Chaos! 🔥

ckan/views/dataset.py
amercader: Let's port some more functionality to htmx and see what patterns emerge, and then it will be easier to decide for which option to settle

doc/theming/htmx.rst
```diff
@@ -68,10 +68,10 @@ What we need to achieve this behaviour is:
...
 1. HTML structure
 
-The HTML structure is very simple. We just need a div that contains the button and the counter. In our case, and
-to respect the current CKAN UX, we have the ``package/snippets/info.html`` snippet that is used to render this
-information. We just need to make sure that the HTML element that we want to replace has an id. In our case,
-we just add and id to it: ``package-info``.
+The HTML structure is very simple: a div that contains the button and the counter.
+To respect the current CKAN UX we update the ``package/snippets/info.html`` snippet that is used to render this
+information. We need to make sure that the HTML element that we want to replace has an id so
+we add and id to it: ``package-info``.
```
ThrawnCA, line 74: "add an id"

pdelboca: Thnx!

amercader:
@wardi @smotornyuk unless you want to have a final look I'm going to merge this as it looks like a great starting point

amercader:
great work @pdelboca . In your view, what are the next parts of the UI that could be migrated to htmx? Should we jump straight to search or are there smaller tasks that could be done?

kowh-ai:
Great work @pdelboca - Is it possible to get a list of **all** the parts of the UI, that are candidates for migration to htmx? I'm thinking a list (with maybe a ranking of how complex this could be) may make it easier for contributors to decide if they can help

pdelboca:
@amercader @kowh-ai I will think about it and make a list. There are some cases we can explore before jumping into the search.

PR #7578, opened 2023-05-08
7573 data api extensions

wardi:
Fixes #7573

### Proposed fixes:

Improve CKAN Data API dialog with syntax highlighting, multiple client languages and jinja2 blocks for expansion

### Features:

ckan-data-api

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [X] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

PR #7686, opened 2023-06-30
Update textview extension assets

pdelboca:
Some webassets are out of sync.

This PR adds some missing assets to the `textview` extension.

This is the output of executing `npm run postinstall`.

### DIFF
--- a/ckanext/textview/theme/public/styles/a11y-light.css
+++ b/ckanext/textview/theme/public/styles/a11y-light.css
@@ -1,0 +1,7 @@
+pre code.hljs{display:block;overflow-x:auto;padding:1em}code.hljs{padding:3px 5px}/*!
+  Theme: a11y-light
+  Author: @ericwbailey
+  Maintainer: @ericwbailey
+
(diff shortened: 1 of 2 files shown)


wardi:
I'd moved some of these to webassets directories, did I miss updating a file that pointed to the public folder? https://github.com/ckan/ckan/pull/7505

pdelboca:
@wardi thanks for the clarification!

Our `gulpfile.js` still copy this files to the theme folder. I'm gonna clean those lines in another PR. thanks!

wardi:
Thanks @pdelboca I didn't know about gulpfile.js
I've also added another copy of highlight.js for the data API dialog in https://github.com/ckan/ckan/pull/7578 this one built with just the 5 languages needed for the examples (bash, javascript, powershell, python and R)

Sounds like I should add this to the gulpfile as well, right?

pdelboca:
@wardi it is unclear.

I assume that we handle our frontend dependencies using gulp given that we have a command for it. The documentation its not quite clear either: https://docs.ckan.org/en/2.10/contributing/frontend/index.html?highlight=gulp#install-frontend-dependencies

Let's discuss it tomorrow in the CKAN Tech Meeting.

Issue #7690, opened 2023-07-04
npm install? npm ci? npm build?


wardi:
## CKAN version
master

## Describe the bug
We should update our front end documentation with the latest guidance for installing and upgrading JS dependencies.

What is a gulpfile and how do we update it? How do we specify build options? (e.g. customizing languages for highlight.js for the data API dialog)

smotornyuk:
Documentation updated in #7691. 

Now follows the wordy explanation, but I'll try to keep it short.

`gulp` is a task runner. It resembles Python's `click`. You register `tasks`(`commands` in `click`), and these tasks are executed by the interpreter. By default, `gulp` expects tasks to be defined in the `gulpfile.js` and we are relying on this default behavior. 
To execute tasks, you have to install GulpCLI globally: `npm i -g gulp-cli`. Now(assuming you've configured PATH properly), you have a `gulp` command, that searches for the `gulpfile.js` and executes tasks from it. You can switch to CKAN's root folder and run `gulp --tasks` - it will print all the available tasks.

We have 3 tasks defined in CKAN: 
* `build` - compile the main theme from SCSS into CSS
* `watch` - same as above, but starts watcher, which recompiles the theme after every change in the SCSS source
* `updateVendorLibs` - copy vendor libs from `node_modules`(which is the node's version of `<VENV>/lib/pythonX.Y/site-packages`) into expected public location(different for most of the libraries).

Any of these tasks can be invoked via `gulp <TASK>`. But, to hide implementation, simplify invocation and, potentially, allow switching to a different task-runner in the future, we are not using `gulp` directly. Instead, we are using nodeJS scripts. These are aliases(similar to python entry_points), which are defined in the `package.json`. 

Look at the [`script` attribute inside `package.json`](https://github.com/ckan/ckan/blob/master/package.json#L29-L32). It has 4 items: `test`, `watch`, `build`, `postinstall`. Every of them is an alias for the sh-script. To execute such a script, you have to run `npm run <ALIAS>`. For example, instead of `gulp updateVendorLibs` you can do [`npm run postinstall`](https://github.com/ckan/ckan/blob/master/package.json#L32). Or you can add a new item into `scripts`:
```json
"scripts": {
   ...,
   "hello": "echo hello world"
}
```
and invoke it via `npm run hello`.

[...]

PR #7691, opened 2023-07-04
[#7690] additional details about frontend interactions

smotornyuk:
* Fix fronted installation docs and explain why `npm ci` is better than `npm i`
* Mention gulpfile.js in the docs and add a link to gulp docs
* update gulp tasks, including datastore's highlight.js into scope


wardi:
highlight.js for the datastore needs some languages not in the default install (powershell and R) so I had built a custom one with only the languages required. How is that done now?

smotornyuk:
I explained the process in #7690. 
As for highlight.js - I'm not sure if we are doing it in the right way. First of all: a short list of supported languages. Now we use highlight.js with support of `javascript`, `python`, `R`, `PowerShell`, and cURL. If somebody extends the data API snippet and adds Ruby, this snippet won't be highlighted. The person will be forced to compile another version of highlight.js and include it in addition to our default version. 
The second problem: the text_view plugin. It also uses highlight.js. This PR will copy the same highlight.js source into the datastore and text view. But if both plugins are enabled, we'll have two different assets with the same content. I don't think that the data API button can be rendered for the text view, so we won't load them simultaneously. But still, it seems that defining the highlight.js asset inside CKAN core and using it in datastore and text view as a dependency is better than registering it twice.

In the original issue I explained how we can build custom version of highlight.js using 2 `npm run ...` aliases. How about creating a default `ckan/highlight.js` asset with support of languages that are used out of the box by CKAN snippets(datastore and textview) and explaining the way to build a different version of `highlight.js` and overriding `ckan/highlight.js` asset in extension, so that there will be always only one version of highlight.js loaded into the app and users are still allowed to add new languages

pdelboca:
I love having a Makefile, but can we remove `gulp` as well? If not we keep adding layers of automation over layers of automation for our build process.

Can we rely on the Makefile for our CSS build as well?

smotornyuk:
We can, but gulp does it much better, than bash. If we want to compile SCSS into CSS, bash will work. But gulp shines when we are using watchers or including sourcemaps into the development build of static files. 
So I think that keeping gulp worth the effort. But it should be used only for CSS/JS transformation and never for maintenance tasks that are not modifying frontend files.

I'll follow up with the second comment summarizing changes and ideas behind them

smotornyuk:
The scope was changed after discussion and that's what we have now.

Using `npm` for updating vendor libraries is not very clear for CKAN developers, because they are usually more familiar with python and bash than nodejs and its package manager. And using Gulp for copying over files doesn't sound like a proper solution anyway, so let's try a different approach.

In this PR I included a `Makefile`:
image

It contains a rule that copies vendor libraries from `node_modules` into public CKAN folders: `make vendor-copy`. 

But why use just a single rule if we have `Makefile`? I checked our release guide and added a number of additional rules for building the documentation, pulling/pushing translations, etc.

Eventually, we can encapsulate most of the non-click scripts inside a `Makefile`. It's more familiar than Gulp and suits better for generic tasks that do not include SCSS compilation. Maybe even the whole release process can be implemented as an interactive `make release` rule(but I don't want to go this far in the current PR).


@wardi, as for the original problem, multiple versions of highligh.js are included in the CKAN source, it is also handled via make-rule.
`make vendor-copy`, which copies updated vendor libraries into a public folder, automatically calls `hljs-build` rule. This rule installs all the build dependencies for Highlight.js(yes, they have not been installed automatically because of hljs policy) and builds Higlight.js for the predefined set of languages. If we want to extend this set, it can be done via one change inside Makefile.
Instead of copying Highlight.js into text_view and datastore plugin, I created `vendor/hljs` asset that is used by both plugins and can be potentially used by any external extension. In this way, we always have only one version of Highlight.js and don't see any errors in the JS console.
I updated documentation, explaining the way to include a new language support or a new theme for highlight.js, so every problem is covered.


As @pdelboca suggested, we can even go further and remove Gulp completely. But implementing a file watcher that minifies/includes sourcemaps into build CSS files is not a trivial task for bash and I afraid we'll end up with unreadable make-rule. That's why I suggest keeping Gulp, even though it adds extra layer to our toolset.
Or we can remove gulp and implement just a single make-rule that directly compiles SCSS into CSS. It will work in most cases, as usually we are not changing styles much. But the first person who is going to update/debug/rewrite CKAN theme will curse us a lot:)

pdelboca:
@smotornyuk great work! We can pilot this for the 2.11 release.

I'm still eager to remove gulp completely and let the future person in charge of UX start fresh, but I won't insist :D

smotornyuk:
Apart from the karma damage, personally, I have no reason to insist on any option.

Below are commands that can be used instead of gulpfile (`npm run build` and `npm run watch`). I mentioned at the meeting, that currently we are not using any fancy features of gulp, like media-query merging/sorting, so it's possible to replicate the current behavior. 

```Makefile

frontent-build: ## compile frontend
  npx sass --no-source-map \
    ckan/public/base/scss/main.scss:ckan/public/base/css/main.css \
    ckan/public/base/scss/main-rtl.scss:ckan/public/base/css/main-rtl.css 

frontent-watch:  ## Watch stylesheets and recompile when they change(with sourcemaps)
  npx sass \
    ckan/public/base/scss/main.scss:ckan/public/base/css/main.css \
    ckan/public/base/scss/main-rtl.scss:ckan/public/base/css/main-rtl.css 
```

I can include them in PR, if nobody is using Gulp and we agree to give the person who will work on frontend redevelopment a freedom to choose a toolset

doc/contributing/frontend/index.rst
```diff
@@ -373,3 +392,100 @@ available in all test contexts. Tests can be made asynchronous by using promises
...
+  {% asset "vendor/hljs-light-theme" %}
+
+  {# or dark accessible theme #}
+  {% asset "vendor/hljs-dark-theme" %}
+
+  {# or include external theme(check current Highlight.js version and update URL if necessary) #}
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/styles/docco.min.css">
+
+
+To reduce its size, Highlight.js was compileled for small subset of supported
```
wardi, line 466:
```suggestion
To reduce its size, Highlight.js was compiled for small subset of supported
```

wardi (approved):
approving but I'll defer to @pdelboca on gulp usage, I have no strong opinion either way

pdelboca:
@wardi I no longer have strong opinions on `gulp`, @smotornyuk made good points on why keeping it :)

This PR is already a massive improvement, we can revisit the `gulp` thing later

wardi:
my updated opinion on gulp after having used it is definitely "not a fan" but it's better than nothing. @smotornyuk would you resolve the conflicts so we can merge?

smotornyuk:
Updated

pdelboca:
@wardi do you have any more concerns around this PR? I have been reviewing this again as I still think is a great addition to organize our workflows.

I only found one thing when running `make translation-check-format`. I'm using Debian and it seems that `msgfmt` does not come installed out of the box. Not a big deal but just flagging it here

```bash
(ckan) ~/Repos/ckan  (highlightjs-gulp) [09:38] $ make translation-check-format 
find ckan/i18n/ -name "*.po"| xargs -n 1 msgfmt -c
xargs: msgfmt: No such file or directory
make: *** [Makefile:29: translation-check-format] Error 127
```

It was fixed by just installing gettext: `apt install gettext` so I do not think we must do anything about it.

PR #7697, opened 2023-07-07
datastore_upsert, datastore_create error fixes

### LINKED ISSUE #7683: datastore: InvalidDataError is an invalid data error
## CKAN version
all

## Describe the bug
using the wrong type of data for a column with the datastore raises an internal exception InvalidDataError from the API call. This exception isn't caught by the API view because it's not an expected exception type for APIs.

### Steps to reproduce
create a datastore table with e.g. and int type column and pass a string that can't be parsed as an int with the API. API responds with 500 server error.

e.g.
```bash
$ ckanapi action datastore_create \
    resource:'{"package_id":"an-existing-package"}' \
    fields:'[{"id":"num", "type":"int"}]'
...
$ ckanapi action datastore_upsert method=insert \
    resource_id=resource-id-from-above \
    records:'[{"num":"notanum"}]'
...
  File "/srv/app/src_extensions/ckan/ckanext/datastore/backend/postgres.py", line 1602, in upsert
    upsert_data(context, data_dict)
  File "/srv/app/src_extensions/ckan/ckanext/datastore/backend/postgres.py", line 1160, in upsert_data
    raise InvalidDataError(
ckanext.datastore.backend.InvalidDataError: The data was invalid: invalid input syntax for type integer: "notanum"
```

### Expected behavior
actions should raise a validation error so they can be caught by the API view and not cause a server error.

wardi:
Fixes #7683 

### Proposed fixes:
report postgres record errors consistently against records parameter

datastore_upsert method=insert: prevent 500 on invalid data datastore_create
datastore_create: invalid data errors now reported against records value (not "message")

### Features:

- [X] includes API changes


smotornyuk:
The master branch is fixed. The issue was caused by two PRs with conflicting changes(one removes the `cast` import because it's not used inside a module, and another calls `cast`) which were merged approximately at the same time.

I suspect now you need to merge the master into the current branch because GH action does not see changes from the master if I re-run the type-check job manually from the GitHub interface.

wardi:
@smotornyuk well, we're getting a different error now. That counts as progress right? :smile:

smotornyuk:
Another PR, but this time it was created 5 months ago and it passed type-check. Since then we've updated dependencies, which slightly change the type signature of the version parser. And because of this after the merge(which was completely legit, as all checks were green) we've got a problem. The only way to avoid such a problem is to review and merge everything immediately:)

I merged the master into the current branch and now it's green, so I won't test my luck and merge this PR

PR #7729, opened 2023-08-01
Updating config schema updates the sysadmin UI

### LINKED ISSUE #6951: Updating config schema does not update sysadmin UI
**CKAN version**

2.9

**Describe the bug**

If fields are removed from the configuration schema, eg by plugins implementing `IConfigurer`, the corresponding UI is still visible at /ckan-admin/config

**Steps to reproduce**

- Create an extension that implements IConfigurer and provides an `update_config_schema` function, returning an empty dictionary.
- Install and enable the extension.
- Log in as a sysadmin and go to /ckan-admin/config to verify that all the fields are still visible. Attempting to submit the form will result in an error.

**Expected behavior**

Any fields that are absent from the config schema should be absent from the form and help text at /ckan-admin/config

**Additional details**

This can be addressed by having an IConfigurer extension override `ckan/templates/admin/config.html`. However, this is a messy option. *Adding* a field can be done simply and cleanly by extending the `admin_form` block and calling `super()`, but *removing* a field requires overriding and recreating the entire block - which also has the side effect of erasing any additions made by other plugins that might want to extend that block.

It would be better to have fields automatically hide themselves when removed from the schema.

Gauravp-NEC:
Fixes #6951 

### Proposed fixes:
1.This PR fixes the config UI render issue, when config schema is updated via the extension.
2. Only the components which are specified in the schema are rendered with their helping text in admin/config.html form.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/admin/config.html
+++ b/ckan/templates/admin/config.html
@@ -4,6 +4,16 @@
 
 {% import 'macros/form.html' as form %}
 
+{% set site_title = data['ckan.site_title']%}
+{% set ckan_theme = data['ckan.theme'] %}
+{% set site_tag_line = data['ckan.site_description'] %}
+{% set site_about = data['ckan.site_about'] %}
+{% set site_intro_text = data['ckan.site_intro_text'] %}
+{% set site_custom_css =  data['ckan.site_custom_css']%}
+{% set field_url = 'ckan.site_logo' %}
+{% set is_upload = data[field_url] and not data[field_url].startswith('http') %}
+{% set is_url = data[field_url] and data[field_url].startswith('http') %}
+
 {% block primary_content_inner %}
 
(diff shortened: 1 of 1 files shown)


Gauravp-NEC:
@smotornyuk , I think `Lint / lint (pull_request) ` failure is not related to this PR. Please have a look.

image

smotornyuk:
Indeed. But this solution is overcomplicated and fragile. Why won't you just add a dedicated jinja2 block for every field?

smotornyuk:
Yes, exactly. But use some sort of common prefix, like `field_*` to separate these blocks from standard blocks of the template and avoid collisions in future

Gauravp-NEC:
@smotornyuk , Based on your above comment. I have done the changes in this PR. Please have a look. Thanks :)

PR #7502, opened 2023-03-28
Config from env vars must be normalized

avdata99:
`CONFIG_FROM_ENV_VARS` takes precedence over config file and extensions but those settings are not *normalized*

This is a work in progress.
Is this a bug? Am I missing something?

### Features:

- [x] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #7737, opened 2023-08-03
fix: cast environment variable to int

stefina:
By casting that environment variable as int it can be ensured that values read from ckanext-envvars will be read and used correctly

Fixes #

Otherwise it threw this exception for me:

 2023-08-03 13:59:34,452 ERROR [ckan.config.middleware.flask_app] '>' not supported between instances of 'int' and 'str'
   File "/usr/lib/python3.10/site-packages/flask/app.py", line 1516, in full_dispatch_request
     rv = self.dispatch_request()
   File "/usr/lib/python3.10/site-packages/flask/app.py", line 1502, in dispatch_request
     return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
   File "/usr/lib/python3.10/site-packages/flask/views.py", line 84, in view
     return current_app.ensure_sync(self.dispatch_request)(*args, **kwargs)
   File "/usr/lib/python3.10/site-packages/flask/views.py", line 158, in dispatch_request
     return current_app.ensure_sync(meth)(*args, **kwargs)
   File "/srv/app/src/ckan/ckan/config/middleware/../../views/resource.py", line 253, in post
     get_action(u'resource_create')(context, data)
   File "/srv/app/src/ckan/ckan/logic/__init__.py", line 551, in wrapped
   File "/srv/app/src/ckan/ckan/logic/action/create.py", line 342, in resource_create
     upload.upload(package.resources[-1].id,
   File "/srv/app/src/ckan/ckan/lib/uploader.py", line 338, in upload
     _copy_file(self.upload_file, output_file, max_size)
   File "/srv/app/src/ckan/ckan/lib/uploader.py", line 39, in _copy_file
     if current_size > max_size:
 TypeError: '>' not supported between instances of 'int' and 'str'

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/uploader.py
+++ b/ckan/lib/uploader.py
@@ -93,7 +93,7 @@ def get_max_image_size() -> int:
 
 
 def get_max_resource_size() -> int:
-    return config.get('ckan.max_resource_size')
+    return int(config.get('ckan.max_resource_size') or 10)
 
 
 class Upload(object):


smotornyuk:
It's a wrong approach. CKAN already [normalizes type of all config options](https://github.com/ckan/ckan/blob/master/ckan/config/environment.py#L122) in order to avoid typecasting everywhere in code. After this point keeping CKAN expects that plugins will do all the config modifications in a safe way.

ckanext-envvars [changes configuration during IConfigured.update_config hook](https://github.com/ckan/ckan/blob/master/ckan/config/environment.py#L136-L139) it's the responsibility of the extension to assign **correct** values into corresponding config options. For example, ckanext-envvars can call [normalize](https://github.com/ckan/ckan/blob/master/ckan/config/environment.py#L122C23-L122C41) once again, in the end of `update_config` hook

stefina:
@smotornyuk thanks a lot for looking at the PR and your advice. Happy to integrate a fix to where it makes sense.
While I am a little confused why we should need to call it twice (the version I am testing with is actually CKAN 2.10 so I assume that the function should be called for us) I will raise this issue in ckanext-envvars. 
Thanks a lot for your pointers!

bellisk:
After running into this issue again, I spent some time to find out what was really happening, and found that the problem was that we were setting the config option `ckan.ckan.max_resource_size` from the environment variable `CKAN_MAX_UPLOAD_SIZE_MB`. There is a merged PR to normalize config options set this way (https://github.com/ckan/ckan/pull/7502), but that change is only available in the `master` branch, not in v2.10.1, which we are using.

The workaround for us is to use the environment variable `CKAN__MAX_RESOURCE_SIZE` instead, which is interpreted as `ckan.max_resource_size` by ckanext-envvars and normalized there.

amercader:
@bellisk thanks for the info, the proper fix will be part of the upcoming CKAN 2.10.2 patch release

bellisk:
thanks @amercader, that's good to know!

amercader:
I'll close this as it has been addressed in #7502

PR #7753, opened 2023-08-15
Fix token expiration

avdata99:
Fixes the expiration date for API tokens
It looks like the `unit` for expiration API token is not actually in use
If we define `expire_in=10` and `unit=4` to try to expire in 10 days this will be expire in 40 seconds (10*4)
The UI for the API token creation works as expected but not from terminal.

### Features:

- [x] includes API changes

### DIFF
--- a/ckanext/expire_api_token/plugin.py
+++ b/ckanext/expire_api_token/plugin.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
-
+import logging
 from typing import Any
 from ckan.types import Schema
 from ckan.common import CKANConfig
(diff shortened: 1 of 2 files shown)


smotornyuk:
Why do you expect `unit = 4` to set duration in days? 
It's initially designed to hold a number of seconds in one `expire_in` unit. I.e, when when `expire_in = 10` and `unit = 4` produces 40 seconds, it's exactly what was expected. If you want to expire the API token in 10 days, send something like `unit= 86_400`

avdata99:
@smotornyuk My mistake, I misinterpreted the [unit selector](https://github.com/ckan/ckan/blob/master/ckanext/expire_api_token/templates/user/api_tokens.html#L5-L10)

<pre>
{% set unit_options = [
  {'value': 1, 'text': _('Second')},
  {'value': 60, 'text': _('Minute')},
  {'value': 3600, 'text': _('Hour')},
  {'value': 3600*24, 'text': _('Day')},
  ] %}
</pre>

PR #7772, opened 2023-09-01
Extract tracking to own core extension

pdelboca:
### Proposed fixes:

As discussed in the Dev Meeting, it would be nice to extract the tracking feature to its own extension. This PR adds the `tracking` core extension and migrates all the current functionality to it.

### Features:
 - CLI commands `ckan tracking ...` are maintained 
 - There is no substantial change in its usage/internals. (We can improve it later)
 - `include_tracking` parameter is deprecated tracking summary will be returned if the extension is enabled.

### TODO
 - [x] Fix tests circular dependency between `with_plugins` fixture and `migrate_db_for`
 - [x] Define how to track resources since overriding the whole template to add the tag class looks dirty.
 - [x] Move popular sort option to the extension
 - [x] Move CSS assets to extension
 - [x] Move models to the extension.
 - [x] Move tests to the extension.
 - [x] Move Popular icon feature to the extension.
 - [x] Clean core code


pdelboca:
@ckan/core the tracking functionality is tightly coupled to CKAN core and it will be difficult to get it as an independent extension without a heavy refactor. So I'm thinking that even when this PR has good intentions is gonna create more harm than good. I say this because I need to add several comments warning that part of the implementation still depends on core.

Mainly:
 - Tables in the database are still gonna be created in CKAN core. (We can avoid this but it was decided to keep it)
 - Indexed information is still defined in the main Solr schema which is in core.
 
On top of that, the current implementation is quite flaky. It's a javascript that adds an extra ajax POST request that then we filter in the middleware. The implementation also include a CSS class to trigger this javascript in the resources page. Without mentioning the weird two commands that we need to execute to process and update tables which make it buggy since we are filtering and sorting using Solr data that can be easily out of sync. Also, it is currently not working properly in `master`.

I'll add a `To Discuss` label to give some thoughts about it. I would like to see it out of Core, but this middle-point solution is not convincing me. Maybe we can delay this to `CKAN 3.0` and think a better approach or simply just remove it completely and make a better one in a new not-core extension.

pdelboca:
@smotornyuk could you take a look at this one? It is ready to review.

ckanext/tracking/plugin.py
```diff
@@ -0,0 +1,112 @@
...
+        return [tracking]
+
+    # IConfigurer
+    def update_config(self, config: CKANConfig) -> None:
+        toolkit.add_resource("assets", "tracking")
+        toolkit.add_template_directory(config, "templates")
+
+    # IMiddleware
+    def make_middleware(self, app: CKANApp, config: CKANConfig) -> Any:
+        app = TrackingMiddleware(app, config)
```
smotornyuk, line 34: In this way, you replace CkanFlaskApp with the TrackingMiddleware object for all plugins and the rest of the `make_flask_app` function. Imagine multiple plugins that register the middleware in the same way and the last middleware that uses `CKANApp.config` attribute:
```
def make_middleware(app, config):
  app.app.app.app.config ...
```

So instead of wrapping the app into a middleware, it's better to use [`after_request`](https://flask.palletsprojects.com/en/3.0.x/api/#flask.Flask.before_request) callback:

```
def make_middleware(app, config):
  app.after_request(function_that_does_something_for_tracking)
  return app
```

You'll need to adapt(and probably replace with a plain function) code of TrackingMidddleware, but it will keep the original value of the `app` and nothing will break in external extensions.

pdelboca: Thanks! Agree with refactoring it since I have encountered so far several times the app.app.app... stuff.

PR #7776, opened 2023-09-06
Strict configuration parsing by default

### LINKED ISSUE #7711: Default to strict config mode in 2.11- config options
As discussed in https://github.com/ckan/ckan/pull/7450#issuecomment-1597310416 we want default to `strict` mode for configuration options starting from CKAN 2.11. The code changes are easy enough, but I want to check what is the preferred way to handle it via config options.

In CKAN 2.10:

| `config.mode` | Behaviour |
| ------------------ | --- |
| Not set | Invalid values are left unprocessed |
| `default` | Invalid values are left unprocessed |
| `strict` | CKAN does not start unless all options are valid |

For CKAN 2.11:

**Option 1**

Just change the internal default. Leave the config options as they are. I'm very -1 on this one because `default` doesn't actually mean default.

| `config.mode` | Behaviour |
| ------------------ | --- |
| Not set | CKAN does not start unless all options are valid |
| `default` | Invalid values are left unprocessed |
| `strict` | CKAN does not start unless all options are valid |

**Option 2**

New `legacy` option to support the traditional behaviour (what was `default` in 2.10)

_Option 2a_

| `config.mode` | Behaviour |
| ------------------ | --- |
| Not set | CKAN does not start unless all options are valid |
| `default` | CKAN does not start unless all options are valid |
| `strict` | CKAN does not start unless all options are valid |
| `legacy` | Invalid values are left unprocessed  |

_Option 2b_

| `config.mode` | Behaviour |
| ------------------ | --- |
| Not set | CKAN does not start unless all options are valid |
| `default` | CKAN does not start unless all options are valid |
| `legacy` | Invalid values are left unprocessed  |

_Option 2c_

| `config.mode` | Behaviour |
| ------------------ | --- |
| Not set | CKAN does not start unless all options are valid |
| `strict` | CKAN does not start unless all options are valid |
| `legacy` | Invalid values are left unprocessed  |

**Option 3**

Drop the traditional behaviour entirely, deprecate and ignore the contents of `config.mode` and always assume `strict`

Any other options?
What are your preferences? @ckan/core

amercader:
Fixes #7711 
As discussed in https://github.com/ckan/ckan/pull/7450#issuecomment-1597310416

Changes in how the config is validated and parsed against the config declaration to default to the old "strict" mode (i.e. CKAN can not start with invalid config options values)

Changes in code are relatively simple, also included changes in documentation

### DIFF
--- a/changes/7677.removal
+++ b/changes/7677.removal
@@ -1,1 +1,1 @@
-`homepage_style` variable has been removed. `layout1.html` code has been moved into `home/index.html` as it will be the only layout available.
+ The ``ckan.homepage_style`` configuration options and the ``homepage_style`` variable have been removed. ``layout1.html`` code has been moved into ``home/index.html``, as it will be the only layout available.
--- a/changes/7776.feature
+++ b/changes/7776.feature
@@ -1,0 +1,4 @@
+The default behaviour starting from CKAN 2.11 is the old ``strict`` mode, where CKAN will not
+start unless **all** config options are valid according to the validators defined in the
+:ref:`configuration declaration <_declare-config-options>`. For every invalid config option,
+an error will be printed to the output stream.
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -136,9 +136,7 @@ class CKANConfig(MutableMapping):
         """
         if default is SENTINEL:
             default = None
-            if (
-                len(config_declaration._options) and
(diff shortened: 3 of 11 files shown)


smotornyuk:
Ready to merge this one. @wardi, any objections?

PR #7804, opened 2023-09-18
Fix Possible Member Duplication

JVickery-TBS:
fix(logic): member create behave like member save;

- Made `member_create` action method behave like `group_member_save` dictize method.

# Fixes

There is a possible scenario in which a member can get "duplicated". This happens if the member is deleted, and then re-added and the generated primary ID of the member happens to get selected before the old deleted member from the database table.

The `group_member_save` dictization method already handles this scenario. So this is just aligning the `member_create` logic action method to behave similarly to that.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -624,8 +624,8 @@ def member_create(context: Context,
                 capacity != u'admin':
             raise NotAuthorized("Administrators cannot revoke their "
                                 "own admin status")
-        if member.state != u'active':
-            member.state = u'active'
+        if member.state != 'active':
+            member.state = 'active'
     else:
         member = model.Member(table_name=obj_type,
                               table_id=obj.id,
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
Due to changes to CircleCI, someone with write permissions to this repo and who is added as a member to the CircleCI project will have to initialize the build/pipeline

ckan/logic/action/create.py
```diff
@@ -621,6 +624,8 @@ def member_create(context: Context,
                 capacity != u'admin':
             raise NotAuthorized("Administrators cannot revoke their "
                                 "own admin status")
+        if member.state != u'active':
+            member.state = u'active'
```
amercader, line 628:
```suggestion
        if member.state != 'active':
            member.state = 'active'
```

Can you commit this to trigger a new test run? I can't do it from the UI, thanks @JVickery-TBS

JVickery-TBS:
@amercader committed you suggestions. I am not part of the Circle CI project, nor have write perms on the ckan org or repos. So the Circle CI tests won't get triggered from my commits

PR #7807, opened 2023-09-18
download button for datastore-only resources

### LINKED ISSUE #7806: Datastore-only resources missing Download button
## CKAN version
all

## Describe the bug
When creating a datastore-only resource it is possible to add views like the datatable view that provides a filtered download option, but on the resource page itself no download button is offered.

### Steps to reproduce
~create a resource with `datastore_create`, view the resource info page.~ see https://github.com/ckan/ckan/pull/7807#issuecomment-1730206956 for proper steps to reproduce

### Expected behavior
A download button with the usual options (CSV, TSV, JSON, XML) should be visible

wardi:
Fixes #7806

### Proposed fixes:

- add download button for datastore-only resources
- fix surrounding template indentation

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7806.bugfix
+++ b/changes/7806.bugfix
@@ -1,0 +1,1 @@
+datastore-only resources now have a visible download button on the resource page
--- a/ckan/templates/package/resource_read.html
+++ b/ckan/templates/package/resource_read.html
@@ -34,50 +34,53 @@
                     <li>{% link_for _('Views'), named_route=pkg.type ~ '_resource.views', id=pkg.name, resource_id=res.id, class_='btn btn-default', icon='chart-bar' %}
                   {% endif %}
                 {% endblock action_manage %}
-            {% if res.url and h.is_url(res.url) %}
-              <li>
-                <div class="btn-group">
-                <a class="btn btn-primary resource-url-analytics" href="{{ res.url }}">
-                  {% if res.resource_type in ('listing', 'service') %}
-                    <i class="fa fa-eye"></i> {{ _('View') }}
-                  {% elif  res.resource_type == 'api' %}
-                    <i class="fa fa-key"></i> {{ _('API Endpoint') }}
-                  {% elif not res.has_views and not res.url_type == 'upload' %}
-                    <i class="fa fa-external-link"></i> {{ _('Go to resource') }}
-                  {% else %}
-                    <i class="fa fa-arrow-circle-down"></i> {{ _('Download') }}
-                  {% endif %}
-                </a>
-                {% block download_resource_button %}
-                  {% if res.datastore_active %}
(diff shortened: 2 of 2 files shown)


kowh-ai:
Testing the change, I’ve added a datastore-only resource as per the documentation [here](https://docs.ckan.org/en/2.10/maintaining/datastore.html#test-the-set-up)

`curl -X POST http://127.0.0.1:5000/api/3/action/datastore_create -H "Authorization: {YOUR-API-KEY}" -d '{"resource": {"package_id": "{PACKAGE-ID}"}, "fields": [ {"id": "a"}, {"id": "b"} ], "records": [ { "a": 1, "b": "xyz"}, {"a": 2, "b": "zzz"} ]}'`

It added OK and I see the table in the database. I restarted my CKAN container but I don’t see the “**Download**” button like I do on another resource I have added to this package/dataset. I have added some custom text to `resource_read.html` which does render on this page after a container restart, so I know changes to code are getting picked up during a container restart. I’ll keep digging. I must be doing something wrong
Screenshot 2023-09-21 at 1 41 59 pm
Screenshot 2023-09-21 at 1 42 13 pm

wardi:
Interesting. It's nothing you did but creating a datastore resource with datastore_create seems to have given it a URL, and this change assumes no URL will be set.

I submitted this as a fix I need for upcoming table designer work where the resource is created without a url value set, that's why you're not seeing the change.

wardi:
@kowh-ai Here are proper steps to reproduce:

1. Create a resource with no URL value set (just give it a name or description and click save)
2. `ckanapi action datastore_create resource_id=<the-resource-you-created> force=true`
3. visit the resource page (it should now have a download dropdown with this change)

kowh-ai:
@wardi thanks for that...will pick this up when I'm back from vacation end of next week...apologies for the delay

kowh-ai:
@wardi - nope, still doesn't show the "download" button after updating the existing resource

Screenshot 2023-10-02 at 11 12 00 am

wardi:
updating the existing resource won't clear the url, try with a new resource instead.

kowh-ai:
`ckanapi action datastore_create <etc>` requires an existing resource id ie: cannot create a new resource

The following 2 commands create the resource but there are no (Download, Goto Resource or Data API) buttons on the page

`ckanapi action resource_create package_id=26ecd3d3-6402-41ac-8dd0-106578c9c2b0 name=resource5`

`ckanapi action resource_create package_id=26ecd3d3-6402-41ac-8dd0-106578c9c2b0 name=resource6 \ upload@/srv/app/really.simple.csv force=true`

wardi:
Thanks @kowh-ai that seem like a slightly different issue but I'll try to fix that here too.

amercader:
I was able to reproduce and check the fix. @kowh-ai feel free to create a separate issue for that last behaviour

Issue #7117, opened 2022-10-03
Theme regression in 2.9.6


EricSoroos:
**CKAN version**
2.9.6

**Describe the bug**
On updating to ckan 2.9.6 from 2.9.5, an extension that used the app global main_css failed with a jinja template error:
```
ckan29                        |     {% asset main_css[6:-4] %}
ckan29                        | jinja2.exceptions.UndefinedError: 'ckan.config.middleware.flask_app.CKAN_AppCtxGlobals object' has no attribute 'main_css'
```

Ref: https://github.com/ckan/ckan/commit/f9084f909538a963b1121d5cfb1e3a1485324add

**Steps to reproduce**
Have an extension that refers to the old `main_css` asset or `h.get_rtl_css`

**Expected behavior**
Not an error -- point releases should not remove portions of the publicly visible API.

PR #7808, opened 2023-09-21
Upgrade requirements September 2023

amercader:
Fixes #7717

Main points:

* Added certifi package to requirements.in due to a security advisory (#7117) (we were getting this through requests, which requires an older minimum version), only with a minimum version
* Updated the Sphinx build command (`python setup.py build_sphinx` -> `sphinx-build doc build/sphinx`) as the setuptools integration has been removed in Sphinx 7


kowh-ai:
@amercader - will get to this next week, on vacation for a few days

kowh-ai (approved):
Looks good @amercader

PR #7839, opened 2023-09-27
Fix Octet Streaming of Datastore Dump

JVickery-TBS:
fix(views): datastore ext dump octet streaming;

- Fixed view method for datastore dump to actually stream data in Flask response.
- Fixed possible issue in sql return.

# Fixes

Datastore dump does not actually stream the data. The view method has now be re-written to properly stream the data via Flask Response.

Also includes small condition for possible issue in the return of some sql stuff.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -181,10 +181,15 @@ def dump(resource_id: str):
         else:
             paginate_by = PAGINATE_BY
 
+        headers = {}
+        if content_type:
+            headers['Content-Type'] = content_type
+        if content_disposition:
+            headers['Content-disposition'] = content_disposition
+
         return Response(stream_dump(offset, limit, paginate_by, result),
                         mimetype=u'application/octet-stream',
-                        headers={'Content-Type': content_type,  # type: ignore
-                                 'Content-disposition': content_disposition})
+                        headers=headers)  # type: ignore
     except ObjectNotFound:
         abort(404, _(u'DataStore resource not found'))
 
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -112,8 +112,6 @@ def dump(resource_id: str):
     else:
         abort(404, _('Unsupported format'))
 
-    output_stream = StringIO()
-
     headers = {}
     if content_type:
         headers['Content-Type'] = content_type
@@ -122,7 +120,6 @@ def dump(resource_id: str):
 
     try:
         return Response(dump_to(resource_id,
-                                output_stream,
                                 fmt=fmt,
                                 offset=offset,
                                 limit=limit,
@@ -206,10 +203,12 @@ class DictionaryView(MethodView):
 
 
 def dump_to(
-    resource_id: str, output_stream: Any, fmt: str, offset: int,
+    resource_id: str, fmt: str, offset: int,
     limit: Optional[int], options: dict[str, Any], sort: str,
-    search_params: dict[str, Any], user: str, is_generator: bool = True
+    search_params: dict[str, Any], user: str
 ):
+    output_buffer = StringIO()
+
     if fmt == 'csv':
         writer_factory = csv_writer
         records_format = 'csv'
@@ -227,9 +226,9 @@ def dump_to(
 
     bom = options.get('bom', False)
 
-    def start_stream_writer(output_stream: StringIO,
+    def start_stream_writer(output_buffer: StringIO,
                             fields: list[dict[str, Any]]):
-        return writer_factory(output_stream, fields, bom=bom)
+        return writer_factory(output_buffer, fields, bom=bom)
 
     def stream_result_page(offs: int, lim: Union[None, int]):
         return get_action('datastore_search')(
@@ -247,7 +246,7 @@ def dump_to(
 
     def stream_dump(offset: int, limit: Union[None, int],
                     paginate_by: int, result: dict[str, Any]):
-        with start_stream_writer(output_stream, result['fields']) as output:
+        with start_stream_writer(output_buffer, result['fields']) as output:
             while True:
                 if limit is not None and limit <= 0:
                     break
@@ -255,10 +254,10 @@ def dump_to(
                 records = result['records']
 
                 output.write_records(records)
-                output_stream.seek(0)
-                yield output_stream.read()
-                output_stream.truncate(0)
-                output_stream.seek(0)
+                output_buffer.seek(0)
+                yield output_buffer.read()
+                output_buffer.truncate(0)
+                output_buffer.seek(0)
 
                 if records_format == 'objects' or records_format == 'lists':
                     if len(records) < paginate_by:
@@ -273,8 +272,8 @@ def dump_to(
                         break
 
                 result = stream_result_page(offset, limit)
-        output_stream.seek(0)
-        yield output_stream.read()
+        output_buffer.seek(0)
+        yield output_buffer.read()
 
     result = stream_result_page(offset, limit)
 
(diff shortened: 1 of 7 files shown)


wardi:
Love to see this @JVickery-TBS

Datastore dump for large datasets has been broken since 2.9 and really needs this fix.

JVickery-TBS:
Okay typing should be good now. And added a fix to `member_dump` from group view as I contributed it using the datastore csv_writer which now changes with this PR

ckanext/datastore/blueprint.py
```diff
@@ -59,6 +61,11 @@ def dump_schema() -> Schema:
 
 
 def dump(resource_id: str):
+    try:
+        resource_id = resource_id_validator(resource_id)  # type: ignore
+    except Invalid:
+        abort(404, _(u'DataStore resource not found'))
```
amercader, line 67: `resource_id_validator` only checks that the input string has valid values and length, not that the resource exists.

If we want to check if a resource has a Datastore table we can do:

```suggestion
    try:
        get_action("datastore_search")({}, {"resource_id": resource_id})
    except ObjectNotFound:
        abort(404, _("DataStore resource not found"))
    
```

Unless there's a better way @wardi ?

ckanext/datastore/blueprint.py
```diff
@@ -67,30 +74,127 @@ def dump(resource_id: str):
             )
         )
 
-    response = make_response()
-    response.headers[u'content-type'] = u'application/octet-stream'
+    fmt = data[u'format']
```
amercader, line 77: Let's drop all the `u` prefixes in new code :pray:

amercader (commented):
This looks good to me, just added a couple of comments before merging

ckanext/datastore/blueprint.py
```diff
@@ -67,30 +74,127 @@ def dump(resource_id: str):
...
+        k: v
+        for k, v in data.items()
+        if k in [
+            u'filters', u'q', u'distinct', u'plain', u'language',
+            u'fields'
+        ]
+    }
 
-    try:
-        dump_to(
```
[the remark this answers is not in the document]
JVickery-TBS: @amercader the `dump_to` is still used by the CLI command. Should I remove it and just rework the new code into a function that the blueprint method AND the cli method can call?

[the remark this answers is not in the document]
amercader: @JVickery-TBS preferably yes, let's consolidate one function and avoid duplication, but if it's complicated at least we can move the `dump_to` function to the cli module where it is used.

ckanext/datastore/blueprint.py
wardi: we could skip the check here and convert NotFound/ValidationErrors from the later datastore_search call to 404's

wardi: If catching the errors is too late (headers/data have already been sent) then datastore_search here would work but at least add a `"limit": 0` so that it returns more quickly.

ckanext/datastore/blueprint.py
JVickery-TBS: Done!

ckanext/datastore/blueprint.py
JVickery-TBS: Did `datastore_search` with limit 0. Also added not supported format response

ckanext/datastore/blueprint.py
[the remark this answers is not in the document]
JVickery-TBS: Tried to rework the new code changes into the `dump_to` method. But my local setup for ckan core is not really working with uploading datastore tables. So just going to make some changes on a different branch in another setup to test to make sure that this is working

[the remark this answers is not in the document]
JVickery-TBS: @amercader okay had to change a couple things but got it working nicely now. So the code is mainly in the `dump_to` method now.

I also saw a `FIXME` for the cli dump command regarding the `bom` flag. That is now removed and conditioned based on format

ckanext/datastore/cli.py
```diff
@@ -86,13 +86,17 @@ def permissions_sql(maindb: str, datastoredb: str, mainuser: str,
...
-         offset: int, limit: int, bom: bool):
+         offset: int, limit: int):
     u'''Dump a datastore resource.
     '''
     flask_app = ctx.meta['flask_app']
+    user = logic.get_action('get_site_user')(
+            {'ignore_auth': True}, {})
+    bom = False
+    if format == 'csv' or format == 'tsv':
+        bom = True
```
wardi, line 99: not everyone wants a bom in their csv/tsv, I think you can leave the option as it was.

ckanext/datastore/blueprint.py
```diff
@@ -163,63 +206,103 @@ def post(self, id: str, resource_id: str):
...
+                    paginate_by: int, result: dict[str, Any]):
+        with start_stream_writer(output_stream, result['fields']) as output:
+            while True:
+                if limit is not None and limit <= 0:
+                    break
+
+                records = result['records']
+
+                output.write_records(records)
+                output_stream.seek(0)
```
wardi, line 258: streams are by definition file-like objects that you can't call seek on, so this should be renamed `output_buffer` or similar.

Since all the users of dump_to either create a StringIO to pass as this variable or pass a file with `is_generator=False` I think it's better to remove both parameters, create the StringIO object inside the function and always use it as a generator.

You can stream the `dump_to` generator to a file with
```python
for block in dump_to(...):
    myfile.write(block)
```
instead of passing in myfile and is_generator=False

ckanext/datastore/cli.py
JVickery-TBS: Reverted

ckanext/datastore/blueprint.py
JVickery-TBS: @wardi okay cool, yeah I did not know how to get a generator to write to a file. This has been done now and works

amercader:
@JVickery-TBS some related tests are failing:

https://app.circleci.com/pipelines/github/ckan/ckan/5846/workflows/c3297949-45cb-4f09-8798-19f858e1c68e/jobs/13838/tests

I think circle ci  working needs to be fixed on your end, perhaps authorizing GitHub again?

JVickery-TBS:
@amercader sadly it was a recent change to CircleCI. Users need to have write permissions to the github repo, and also be added to the project in CircleCI to initialize a CircleCI build/workflow.

I had to talk to CircleCI support about it as we recently moved to CircleCI. And of course the week we started using it was the week they made this change. They said it was for security purposes so that anyone cannot just initialize a CircleCI build.

But thanks for running the workflow, I will update these tests to work with the new code

JVickery-TBS:
@amercader Okay the tests should hopefully work now. At least they worked locally for me. Seems like github is just taking a while to update this PR with the new push. But hopefully it updates sometime soon

amercader:
@JVickery-TBS Still failing I'm afraid:
https://app.circleci.com/pipelines/github/ckan/ckan/5855/workflows/527a79bb-8019-41e9-b881-591984d6a40d/jobs/13846/tests

JVickery-TBS:
@amercader Okay Ian came to my rescue and this should all be working now. Had to move the StringIO/BytesIO stuff into the writers as the XML one wanted Bytes and the others String. So you will no longer need to supply a file-like object to the writer methods, they will make one for you and yield the bytes in write records

amercader:
Thanks @JVickery-TBS , and it looks you sorted out your circleci issues! 

As this is quite a big change, would you mind submitting a PR with the same changes against dev-v2.10 to make backporting easier? Many thanks

PR #7853, opened 2023-10-05
SECRET_KEY: fall back to beaker.session.secret

wardi:
### Proposed fixes:

- default secret_key to beaker.session.secret for easier config migration/sharing configurations between ckan versions
- a small config documentation fix

### Features:

- [ ] includes API changes


wardi:
@amercader interested in your thoughts on this one. If the uppercase SECRET_KEY is a flask convention should we change envvars so it can generate this option instead?

amercader:
I think we should align with Flask and keep the upper case `SECRET_KEY` and make ckanext-envvar catch up and support uppercase config options as we now have plenty of those declared in the config declaration (coming from [Flask-login](https://github.com/ckan/ckan/blob/70f1ee5d7c94cf121cd1c4562dfd325f5c52f339/ckan/config/config_declaration.yaml#L743), [WTForms](https://github.com/ckan/ckan/blob/70f1ee5d7c94cf121cd1c4562dfd325f5c52f339/ckan/config/config_declaration.yaml#L676), etc) which would be great to support.

I made a first try in ckanext-envvars, using the config declaration if present (ie CKAN>=2.10) to keep those keys declared as they are when parsing the env vars. This works well in CKAN 2.11 (master) but not in CKAN 2.10. The problem is that `SECRET_KEY` is declared as `internal: true` in CKAN 2.10 which I think prevents it from being returned when doing `config_declaration.iter_options()`.

https://github.com/okfn/ckanext-envvars/pull/9

I'll keep investigating but does this seem like a good approach to explore?

amercader:
[ckanext-envvars 0.0.4](https://github.com/okfn/ckanext-envvars) supports upper case config options

wardi:
@amercader I merged #7884 so we can see the tests passing, once we merge that to master this will be easier to review

PR #7857, opened 2023-10-11
Fix queries when state is in fq_list (2.9 version) #7736

### LINKED ISSUE #7736: Package Search breaks when '+state' is in fq_list
## CKAN version
2.9.x, at least

## Describe the bug

This call:
`get_action('package_search')(None, {'fq_list':['+state:((anything other than active))']})`

results in a solr query like:
`...&fq=state:((anything other than active))&fq=%2Bsite_id:"default"&fq=%2Bstate:active&...`

due to https://github.com/ckan/ckan/blob/2.9/ckan/lib/search/query.py#L332C11-L332C11

```
        if 'fq' in query:
            fq.append(query['fq'])
        fq.extend(query.get('fq_list', []))

        # show only results from this CKAN instance
        fq.append('+site_id:%s' % solr_literal(config.get('ckan.site_id')))

        # filter for package status
        if not '+state:' in query.get('fq', ''):
            fq.append('+state:active')
```

because the filter for package status is only checking the passed in `fq`, and not the `fq_list`.

### Expected behavior

I'd expect that `fq_list` is a more convenient version of `fq`.

I think a fix for this would be:
```
        # filter for package status
        if not any('+state:' in _item for _item in fq):
            fq.append('+state:active')
```

cormachallinanderilinx:
Fixes #7736

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -337,7 +337,7 @@ class PackageSearchQuery(SearchQuery):
         fq.append('+site_id:%s' % solr_literal(config.get('ckan.site_id')))
 
         # filter for package status
-        if not '+state:' in query.get('fq', ''):
+        if not any('+state:' in _item for _item in fq):
             fq.append('+state:active')
 
         # only return things we should be able to see


EricSoroos:
included in main #7905.

PR #7862, opened 2023-10-12
Update Datapusher to use new IResourceController hook

markstuart:
### Proposed fixes:

[CKAN 2.10 updated from `before_update` to `before_resource_update` in IResourceController](https://github.com/ckan/ckan/blob/master/CHANGELOG.rst#removals-and-deprecations)

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #7866, opened 2023-10-17
Fix user_show for ckan.auth.public_user_details

### LINKED ISSUE #7838: `user_show` returns `NotFound` before checking authentication for anonymous users
## Describe the bug

The `ckan.logic.action.get.user_show` action returns a [NotFound error](https://github.com/ckan/ckan/blob/ckan-2.10.1/ckan/logic/action/get.py#L1466) before checking any auth function. This also breaks [`ckan.auth.public_user_details`](https://docs.ckan.org/en/2.10/maintaining/configuration.html#ckan-auth-public-user-details).

### Steps to reproduce
Call `user_show` anonymously. Instead of "access denied", "not found" is returned.

### Expected behavior
For anonymous users, an "access denied" error should be returned. Also, the configuration variable [`ckan.auth.public_user_details`](https://docs.ckan.org/en/2.10/maintaining/configuration.html#ckan-auth-public-user-details) seems to have no effect at all anymore.

markstuart:
Fixes #7838

### Proposed fixes:

Allows _check_access to evaluate the ckan.auth.public_user_details config so that it can return the correct 403 response and not leak information about what users are in the system.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/get.py
+++ b/ckan/logic/action/get.py
@@ -1444,13 +1444,13 @@ def user_show(context: Context, data_dict: DataDict) -> ActionResult.UserShow:
     else:
         user_obj = None
 
-    if not user_obj:
-        raise NotFound
-
-    context['user_obj'] = user_obj
+    if user_obj:
+        context['user_obj'] = user_obj
 
     _check_access('user_show', context, data_dict)
 
+    if not user_obj:
+        raise NotFound
 
     # include private and draft datasets?
     requester = context.get('user')


markstuart:
typecheck failing test is fixed in #7862

Zharktas:
PR is made against the wrong branch, should be done against master and cherry-picked from the to dev-v2.10

pdelboca:
Hello @markstuart we are moving forward with the release of CKAN 2.11 and we would love to have this fix.

Could you re create the PR but targeting master instead? If you do not have time, let me know and I would create a new one. (I don't want to steal your contribution)

ckan/logic/action/get.py
```diff
@@ -1444,13 +1444,13 @@ def user_show(context: Context, data_dict: DataDict) -> ActionResult.UserShow:
     else:
         user_obj = None
 
-    if not user_obj:
-        raise NotFound
-
-    context['user_obj'] = user_obj
+    if user_obj:
+        context['user_obj'] = user_obj
```
markstuart, line 1448: When updating this PR to target master, I found that the Context object no longer accepts None for the 'user_obj' property, so I've added this simple check. This is untested as I don't have a CKAN instance running on `master` locally.

markstuart:
@pdelboca here you go, hopefully that works as expected. I'm very stretched for time right now, so I'd appreciate it if you could cherry-pick back to dev-v2.10 as suggested by @Zharktas.

pdelboca:
Thanks @markstuart ! We'll backport it during the release process.

PR #7868, opened 2023-10-19
Tighten validation for id fields

### LINKED ISSUE #7604: Tighten validation for `id` fields
Having loose or inconsistent validation on `id` fields can lead to confusing behaviour or security concerns. Let's do a review of the various entities with `id` field in the model and apply the same validation to all of them (if possible). Validators should include:

* "Can only be changed by a sysadmin": custom ids only make sense in the context of specific scenarios like migrations, it make sense that only sysadmins (or plugins that ignore auth) touch them
* "Value is a valid UUID". In the past we've allowed different values here to support potential use cases but it's probably safer to enforce UUIDs for all ids and use custom fields if needed
* "Id does not already exist when creating": to prevent taking over other entities

Any that I missed?

Some of these are already applied to some entities, but it would be good to be consistent across all models
* Dataset
* Resource
* Resource View
* Groups (org/groups)
* User
* Activity
* Extras
* Tags (?)
* ...

shashigharti:
Fixes #7604

### Proposed fixes:
- In progress
- Adds validation to id fields of the core entities

### Features:

- [ X] includes tests covering changes
- [ ] includes API changes


ckan/logic/validators.py
```diff
@@ -233,6 +232,33 @@ def resource_id_does_not_exist(
...
 
+def id_does_not_exist(value: str, context: Context) -> Any:
+    """Ensures that the value is not used as a ID or name.
+    """
+
+    model = context['model']
+    session = context['session']
+    entities_map = {
+        'resource_view': model.ResourceView,
+        'group': model.Group
```
amercader, line 243: @shashigharti I think in this case it's safer to have individual validators for each entity type even if that means many different functions. Relying on the keys to determine the entity type won't always be safe. And it will make overriding specific validators easier for plugins.
So let's go for `resource_view_id_does_not_exist`, `group_id_does_not_exist`, etc

amercader:
Closed in favour of #8069

PR #7877, opened 2023-10-26
Drop support for CONFIG_FROM_ENV_VARS

pdelboca:
I'm working on a new deployment of CKAN and while doing it I'm finding how we handle the configuration file and the environment variables a little bit confusing (comparing with more plain Django or Flask apps).

## Context

Usually in normal Flask/Django applications the workflow is the following:
 - Set some environment variables in the server
 - Read those environment variables when starting the application
 - Update the configuration object with the values
 - Run the application
 
In the practice this means that we BUILD the Dockerfile and then whatever system we use to run the application sets the environment variables as secrets. Configuration is set at RUN time before starting.

However, CKAN's "official" workflow is a little bit different:
 - Our Dockerfile creates the `ckan.ini` that will contain the values (this means something is happening at BUILD time)
 - Then at RUN time we create the config object using the `ckan.ini` file
 - If some ENV variables are set we override some of the values in the config object
 - If we have `ckanext-envvars` activated then we will again override the config object.
 
This means that somethings can be handled at BUILD time (when we generate the `ckan.ini` file, but it can also be handled at RUN time (for some variables at least)

Also, our official Docker image checks the connection to the DB using `CKAN_SQLALCHEMY_URL` but at the same time with `ckanext-envvars` we are recommending it to be `CKAN___SQLALCHEMY_URL`.
 
### Proposed fixes:

This PR proposes to drop support for the `CONFIG_FROM_ENV_VARS` functionality to reduce a little bit the scenarios and make the initialization of CKAN more predictable and easier to understand.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -79,43 +79,11 @@ def load_environment(conf: Union[Config, CKANConfig]):
     build_js_translations()
 
 
-# A mapping of config settings that can be overridden by env vars.
-# Note: Do not remove the following lines, they are used in the docs
-# Start CONFIG_FROM_ENV_VARS
-CONFIG_FROM_ENV_VARS: dict[str, str] = {
-    'sqlalchemy.url': 'CKAN_SQLALCHEMY_URL',
-    'ckan.datastore.write_url': 'CKAN_DATASTORE_WRITE_URL',
-    'ckan.datastore.read_url': 'CKAN_DATASTORE_READ_URL',
-    'ckan.redis.url': 'CKAN_REDIS_URL',
-    'solr_url': 'CKAN_SOLR_URL',
-    'solr_user': 'CKAN_SOLR_USER',
-    'solr_password': 'CKAN_SOLR_PASSWORD',
-    'ckan.site_id': 'CKAN_SITE_ID',
-    'ckan.site_url': 'CKAN_SITE_URL',
-    'ckan.storage_path': 'CKAN_STORAGE_PATH',
-    'ckan.datapusher.url': 'CKAN_DATAPUSHER_URL',
-    'smtp.server': 'CKAN_SMTP_SERVER',
-    'smtp.starttls': 'CKAN_SMTP_STARTTLS',
-    'smtp.user': 'CKAN_SMTP_USER',
-    'smtp.password': 'CKAN_SMTP_PASSWORD',
-    'smtp.mail_from': 'CKAN_SMTP_MAIL_FROM',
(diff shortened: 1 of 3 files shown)


smotornyuk:
Using environment variables for interpolation:

1. Set an option from the config file to interpolated value that starts with `CKAN_`:
   ```ini
   ckan.site_title = %(CKAN_TITLE_FROM_ENV)s   
   ```
2. An attempt to run the server without corresponding envvar will produce an error:
   ```log
   configparser.InterpolationMissingOptionError: Bad value substitution: option 'ckan.site_title' 
     in section 'app:main' contains an interpolation key 'CKAN_TITLE_FROM_ENV' which is not 
     a valid option name. Raw value: '%(CKAN_TITLE_FROM_ENV)s'
   ```
3. Export the variable before starting the server in order to fix the error:
   ```sh
   CKAN_TITLE_FROM_ENV="CLI title" ckan run -t
   ```

Environment variables are collected [here](https://github.com/ckan/ckan/blob/master/ckan/cli/__init__.py#L31-L33). The only requirement: envvar's name must begin with `CKAN_`(uppercased). Everything else doesn't matter. For example, you don't need to use uppercase for the rest of the variable name, nor do you need any resemblance between the variable name and the option name. 

The differences between this feature and ckanext-envvars:

* ckanext-envvars uses `CKAN__` prefix(double underscore). To avoid conflicts, make sure that you have one underscore after `CKAN` if you want to use interpolation and two underscores if you want to use ckanext-envvars.
* ckanext-envvars transforms the rest of the variable name into the config option name(there is an exception for declared options). `CKAN__A__B__C` turns into `a.b.c`. The final config option unconditionally overrides any existing option in the config. Core interpolation works only if you use the whole envvar's name in the interpolation expression. For example, if you just export `CKAN_TITLE=1` - nothing will change. You have to add the following expression in front of any config option: `ckan.option.name = %(CKAN_TITLE)s`


As a result, the use cases are the following:

* if you want to use envvar value when it is present, and use the default value from the config file - use ckanext-envvar. This is more flexible, but requires an extra extension
* If you want to use explicitely specified environment variable and fail when it's missing - use core interpolation. This is more restrictive, but works out of the box.

pdelboca:
@kowh-ai this one is ready to review. Let me know what are your thoughts!

pdelboca:
@kowh-ai can you double check with your Dockerimages as well?

kowh-ai:
> @kowh-ai can you double check with your Dockerimages as well?

Yes, I'll need to build a new base image from master and then use that in the ckan compose service

PR #7878, opened 2023-10-26
Fix performance for users following many groups/orgs/users

EricSoroos:
Essentially, the previous code uses a set of unions of individual queries that blows up pretty badly when users follow many items. This PR changes the form from

```
select  from activity where id=1
union 
select from activity where id=2
...
```
to 
``` 
select from activity where id in (1,2,3)
```

A sample query: 
[bad_query (1).log](https://github.com/ckan/ckan/files/13178166/bad_query.1.log)

This is the backport -- diff doesn't apply cleanly to 2.10 due to typing -- will do post discussion. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/model/activity.py
+++ b/ckan/model/activity.py
@@ -170,6 +170,10 @@ def user_activity_list(user_id, limit, offset):
 
     return _activities_at_offset(q, limit, offset)
 
+def _to_list(vals):
+    if isinstance(vale, (list, tuple)):
+        return vals
+    return [vals]
 
 def _package_activity_query(package_id):
     '''Return an SQLAlchemy query for all activities about package_id.
@@ -177,7 +181,7 @@ def _package_activity_query(package_id):
     '''
     import ckan.model as model
     q = model.Session.query(model.Activity) \
-        .filter_by(object_id=package_id)
+        .filter(model.Activity.object_id.in_(_to_list(package_id)))
     return q
 
 
@@ -210,11 +214,7 @@ def _group_activity_query(group_id, include_hidden_activity=False):
     '''
     import ckan.model as model
 
-    group = model.Group.get(group_id)
-    if not group:
-        # Return a query with no results.
-        return model.Session.query(model.Activity).filter(text('0=1'))
-
+    groups = _to_list(group_id)
     q = model.Session.query(
         model.Activity
     ).outerjoin(
@@ -236,17 +236,17 @@ def _group_activity_query(group_id, include_hidden_activity=False):
         # desired but is consistent with legacy behaviour.
         or_(
             # active dataset in the group
-            and_(model.Member.group_id == group_id,
+            and_(model.Member.group_id.in_(groups),
                  model.Member.state == 'active',
                  model.Package.state == 'active'),
             # deleted dataset in the group
-            and_(model.Member.group_id == group_id,
+            and_(model.Member.group_id.in_(groups),
                  model.Member.state == 'deleted',
                  model.Package.state == 'deleted'),
                  # (we want to avoid showing changes to an active dataset that
                  # was once in this group)
             # activity the the group itself
-            model.Activity.object_id == group_id,
+            model.Activity.object_id.in_(groups),
         )
     )
 
@@ -265,10 +265,9 @@ def _organization_activity_query(org_id, include_hidden_activity=False):
     '''
     import ckan.model as model
 
-    org = model.Group.get(org_id)
-    if not org or not org.is_organization:
-        # Return a query with no results.
-        return model.Session.query(model.Activity).filter(text('0=1'))
+    orgs = [model.Group.get(org_id) for org in _to_list(org_id)]
+    orgs = [org for org in orgs if org and org.is_organization]
+    orgs = [org.id for org in orgs]
 
     q = model.Session.query(
         model.Activity
@@ -285,8 +284,8 @@ def _organization_activity_query(org_id, include_hidden_activity=False):
         # to a org but was then removed will not show up. This may not be
         # desired but is consistent with legacy behaviour.
         or_(
-            model.Package.owner_org == org_id,
-            model.Activity.object_id == org_id
+            model.Package.owner_org.in_(orgs),
+            model.Activity.object_id.in_(orgs)
         )
     )
     if not include_hidden_activity:
@@ -337,9 +336,7 @@ def _activities_from_users_followed_by_user_query(user_id, limit):
         # Return a query with no results.
         return model.Session.query(model.Activity).filter(text('0=1'))
 
-    return _activities_union_all(*[
-        _user_activity_query(follower.object_id, limit)
-        for follower in follower_objects])
+    return _user_activity_query([follower.object_id for follower in follower_objects], limit)
 
 
 def _activities_from_datasets_followed_by_user_query(user_id, limit):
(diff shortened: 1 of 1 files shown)


EricSoroos:
Note, there's a new commit here with an additional optimization:

Change the join to a subselect.  This prevents a non-indexable seq
scan on the activities table, and reduces it to an indexable subselect
on package yielding ids that are indexable on activities.

Changes the query from
```
explain analyze SELECT activity.id AS activity_id, activity.timestamp AS activity_timestamp, activity.user_id AS activity_user_id, activity.object_id AS activity_object_id, activity.revision_id AS activity_revision_id, activity.activity_type AS activity_activity_type, activity.data AS activity_data
 FROM activity LEFT OUTER JOIN package ON package.id = activity.object_id AND package.private = false
  WHERE (package.owner_org IN ('7f8304e3-da4c-458b-ae1f-8088c307f7db') OR activity.object_id IN ('7f8304e3-da4c-458b-ae1f-8088c307f7db')) AND activity.user_id NOT IN ('13507bbd-b7ee-492c-bf6e-2d21ef76b878') ORDER BY activity.timestamp DESC
  LIMIT 31;
                                                                               QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=653140.89..653144.50 rows=31 width=720) (actual time=2991.648..2991.837 rows=31 loops=1)
   ->  Gather Merge  (cost=653140.89..653288.60 rows=1266 width=720) (actual time=2991.648..2991.834 rows=31 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Sort  (cost=652140.86..652142.44 rows=633 width=720) (actual time=2985.780..2985.784 rows=15 loops=3)
               Sort Key: activity."timestamp" DESC
               Sort Method: quicksort  Memory: 68kB
               ->  Hash Left Join  (cost=651.25..652111.41 rows=633 width=720) (actual time=821.912..2985.684 rows=20 loops=3)
                     Hash Cond: (activity.object_id = package.id)
                     Filter: ((package.owner_org = '7f8304e3-da4c-458b-ae1f-8088c307f7db'::text) OR (activity.object_id = '7f8304e3-da4c-458b-ae1f-8088c307f7db'::text))
                     Rows Removed by Filter: 1241713
                     ->  Parallel Seq Scan on activity  (cost=0.00..646823.48 rows=1765256 width=720) (actual time=0.148..2665.750 rows=1241733 loops=3)
Filter: (user_id <> '13507bbd-b7ee-492c-bf6e-2d21ef76b878'::text)
                           Rows Removed by Filter: 517703
                     ->  Hash  (cost=534.82..534.82 rows=9314 width=73) (actual time=3.886..3.887 rows=9314 loops=3)
                           Buckets: 16384  Batches: 1  Memory Usage: 1092kB
                           ->  Seq Scan on package  (cost=0.00..534.82 rows=9314 width=73) (actual time=0.009..2.386 rows=9314 loops=3)
                                 Filter: (NOT private)
                                 Rows Removed by Filter: 568
 Planning time: 2.267 ms
 Execution time: 2991.969 ms
```

to

```
explain analyze SELECT activity.id AS activity_id, activity.timestamp AS activity_timestamp, activity.user_id AS activity_user_id, activity.object_id AS activity_object_id, activity.revision_id AS activity_revision_id, activity.activity_type AS activity_activity_type, activity.data AS activity_data
FROM activity
    WHERE activity.object_id in (
      select package.id from package
        where package.private = false
        and package.owner_org IN ('7f8304e3-da4c-458b-ae1f-8088c307f7db')
      union select '7f8304e3-da4c-458b-ae1f-8088c307f7db')
    AND activity.user_id NOT IN ('13507bbd-b7ee-492c-bf6e-2d21ef76b878') ORDER BY activity.timestamp DESC
 LIMIT 31

                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=13430.11..13430.18 rows=31 width=831) (actual time=0.157..0.161 rows=31 loops=1)
   ->  Sort  (cost=13430.11..13436.01 rows=2362 width=831) (actual time=0.157..0.159 rows=31 loops=1)
         Sort Key: activity."timestamp" DESC
         Sort Method: quicksort  Memory: 134kB
         ->  Nested Loop  (cost=58.69..13359.79 rows=2362 width=831) (actual time=0.040..0.122 rows=61 loops=1)
               ->  HashAggregate  (cost=15.66..15.70 rows=4 width=32) (actual time=0.020..0.021 rows=3 loops=1)
                     Group Key: package.id
                     ->  Append  (cost=4.31..15.65 rows=4 width=32) (actual time=0.014..0.017 rows=3 loops=1)
                           ->  Bitmap Heap Scan on package  (cost=4.31..15.60 rows=3 width=37) (actual time=0.014..0.016 rows=2 loops=1)
                                 Recheck Cond: (owner_org = '7f8304e3-da4c-458b-ae1f-8088c307f7db'::text)
                                 Filter: (NOT private)
                                 Heap Blocks: exact=2
                                 ->  Bitmap Index Scan on idx_pkg_owner_org  (cost=0.00..4.31 rows=3 width=0) (actual time=0.009..0.009 rows=2 loops=1)
                                       Index Cond: (owner_org = '7f8304e3-da4c-458b-ae1f-8088c307f7db'::text)
                           ->  Result  (cost=0.00..0.01 rows=1 width=32) (actual time=0.000..0.001 rows=1 loops=1)
               ->  Bitmap Heap Scan on activity  (cost=43.03..3330.11 rows=590 width=831) (actual time=0.013..0.029 rows=20 loops=3)
                     Recheck Cond: (object_id = package.id)
                     Filter: (user_id <> '13507bbd-b7ee-492c-bf6e-2d21ef76b878'::text)
                     Rows Removed by Filter: 1
                     Heap Blocks: exact=48
                     ->  Bitmap Index Scan on idx_activity_object_id  (cost=0.00..42.89 rows=844 width=0) (actual time=0.010..0.010 rows=22 loops=3)
                           Index Cond: (object_id = package.id)
 Planning time: 0.258 ms
 Execution time: 0.196 ms
(24 rows)

Also benefits from an index:

create index idx_pkg_owner_org on package(owner_org);
```

EricSoroos:
Not sure if we can add the index in a point release, because we'd wind up forking the alembic history when upgrading to the next major?

ckan/model/activity.py
```diff
@@ -170,14 +172,18 @@ def user_activity_list(user_id, limit, offset):
 
     return _activities_at_offset(q, limit, offset)
 
+def _to_list(vals):
+    if isinstance(vals, (list, tuple)):
+        return vals
+    return [vals]
```
pdelboca, line 178: @EricSoroos can you add a small docstring to this function? What it does is clear, however it is not clear why we need it. I found this internal functions hard to maintain in the future if it is not clear why are there.

Looks that sometimes `vals` it is a tuple, others a list, and sometimes just a value. Where you able to identify why is that case? The variable names refers to `user_id` or `package_id` which makes you think it is singular but I guess someone is passing lists to it?

EricSoroos: It's probably just a list, but I don't want to be returning [(vals,)].

It was user_id or package_id, but I've overloaded those items to potentially have a list instead to minimize the changes here, so we're doing a  `field in (values)`, but those values could still be singular string uids.

wardi:
> Not sure if we can add the index in a point release, because we'd wind up forking the alembic history when upgrading to the next major?

Isn't handling migration branches a thing that alembic does well?

ThrawnCA:
> Not sure if we can add the index in a point release, because we'd wind up forking the alembic history when upgrading to the next major?

It could at least be added in #7903, right?

duttonw:
We have implemented this on data.qld.gov.au and it did improve things somewhat but its still not good enough. This also affects the email notifications for followed items where activity changes.

Issue #6955, opened 2022-07-05
Page visibility is cached beyond login/logout


ThrawnCA:
**CKAN version**

2.9

**Describe the bug**

Accessing a restricted page such as a private dataset will cache the page; returning to that page will initially show the cached version even if access has changed due to login/logout.

**Steps to reproduce**

* Log in as a privileged user.
* Go to a private dataset.
* Log out.
* Use the 'Back' button to return to the dataset. The page will initially be visible, but will deny access when reloaded.

This also works in reverse:

* Enter the URL of a private dataset while logged out. You will receive an error.
* Log in.
* Use the 'Back' button to return to the page. You will see the error again, until the page is reloaded.

**Expected behavior**

Access controls on a page should apply immediately after logging in/out.

**Additional details**

This becomes even more apparent with plugins like ckanext-datarequest; browsing to the data request list page, clicking the login button there, and logging in, will result in returning to a data request list page that looks logged out, still having a login button, until it is reloaded.

Zharktas:
As far as I know, browsers don't reload the page when back button is pressed, they only execute the javascript that was already there. As CKAN uses server side rendered templates, instead of modern frontend rendering, there isn't much that can be done for cached pages.

ThrawnCA:
The ckanext-datarequests example doesn't use the Back button, though...

wardi:
@ThrawnCA If this is a concern for your site one fix could be to use response headers to disable caching for all pages when a user is logged in. Maybe that's what web sites do when they have a "I am logging in from a shared computer" option on their login screens.

Zharktas:
The login workflow and cookie settings might also relate. If the workflow is doing proper redirects and you are using external authentication service, you need to use Lax in `who.samesite` as Strict prevents sending the cookie after redirects in external service.

ThrawnCA:
There's no external authentication provider at https://www.data.qld.gov.au which is where we're encountering this.

ThrawnCA:
Further investigation has revealed that CKAN uses "Cache-Control: private" whenever a user is logged in, meaning that pages will be cached in the browser but not publicly.

This ensures that sensitive pages will remain secure, but it does not address the issue of highly dynamic pages that must be loaded fresh every time, such as `/user/_logout`.

ThrawnCA:
> @ThrawnCA If this is a concern for your site one fix could be to use response headers to disable caching for all pages when a user is logged in.

[...]

PR #7892, opened 2023-11-03
disable all caching on _logout URL, #6955

ThrawnCA:
By default, authenticated pages are stored in browser cache, which is secure enough - but we don't want _logout cached at all, because despite being a GET request, it has a crucial side effect (deleting the session cookie).

### Proposed fixes:

Disable all caching on `/user/_logout`

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/7892.bugfix
+++ b/changes/7892.bugfix
@@ -1,0 +1,1 @@
+Change logout link into a POST submit so it interacts properly with caching and CSRF
--- a/ckan/templates/header.html
+++ b/ckan/templates/header.html
@@ -39,10 +39,12 @@
         </li>
         {% endblock %} {% block header_account_log_out_link %}
         <li>
-          <a href="{{ h.url_for('user.logout') }}" title="{{ _('Log out') }}">
-            <i class="fa fa-sign-out" aria-hidden="true"></i>
-            <span class="text">{{ _('Log out') }}</span>
-          </a>
+          <form action="{{ h.url_for('user.logout') }}" method="post">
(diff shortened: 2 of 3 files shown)


pdelboca:
Hello @ThrawnCA ! Thanks for reporting this!

I have been reviewing this one and debating it in the tech team meeting. The root cause of this is that the call is still a `GET` when it should be a `POST` request.

Are you interested on changing this PR for an implementation that changes the call to logout to a POST? I'm thinking something like this:
```jinja
<form action="{{ h.url_for('user.logout') }}" method="post">
   <input class="btn btn-link" type="submit" value="Log out">
   {{ h.csrf_input() }}
</form>
```

And then the change in the route registration to accept POST.

ThrawnCA:
> Are you interested on changing this PR for an implementation that changes the call to logout to a POST? I'm thinking something like this:

Done. We have deployed this on our DEV environment, dev.data.qld.gov.au, and it appears to be working correctly. The icon is slightly larger than before, and the button doesn't display a URL on hover; otherwise the user experience is essentially unchanged.

pdelboca:
Thanks @ThrawnCA !

ThrawnCA:
@pdelboca It looks like converting logout to a POST request has had an unfortunate side effect: Since pages are cached too aggressively (see #6955), it is possible for the logout button in the header to have a stale CSRF token.

The whole issue of caching in CKAN 2.10, and how it interacts with logging in and out, needs to be revisited.

PR #7893, opened 2023-11-06
Switch from beaker to flask-session

smotornyuk:
Remove beaker and use flask + flask-session instead


ckan/config/middleware/__init__.py
```diff
@@ -34,3 +36,23 @@ def make_app(conf: Union[Config, CKANConfig]) -> CKANApp:
     _internal_test_request_context = flask_app._wsgi_app.test_request_context()
 
     return flask_app
+
+
+class CKANSecureCookieSessionInterface(SecureCookieSessionInterface):
+    """Flask cookie-based sessions with expration support.
```
pdelboca, line 42:
```suggestion
    """Flask cookie-based sessions with expiration support.
```

pdelboca:
@smotornyuk this one is looking good! and I would love to merge it.

What was the decision on the Tech Team?  Are we going to provide support for old config names?

ckan/config/config_declaration.yaml
```diff
@@ -258,99 +242,141 @@ groups:
...
+            supported by `Flask-Session
+            <https://flask-session.readthedocs.io/en/latest/config.html>`_:
+            ``redis``, ``memcached``, ``filesystem``, ``mongodb``. Keep in mind
+            that certaint session types `require additional Python packages
+            <https://flask-session.readthedocs.io/en/latest/interfaces.html>`_.
+
+            By default, attempt to use non-existing ``SESSION_TYPE`` cause a
+            runtime error upon visiting any endpoint. But if
+            ``ckan.session.custom_type.<TYPE NAME>.import_path`` config option
+            is present, its value used as an import path for the custom session
```
pdelboca, line 278:
```suggestion
            is present, its value is used as an import path for the custom session
```

ckan/config/middleware/__init__.py
```diff
@@ -34,3 +39,37 @@ def make_app(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+        pass
+
+    def open_session(self, app: CKANApp, request: Request):
+        session = super().open_session(app, request)
+        if session:
+            # Cookie-based sessions expire with the browser's session. The line
+            # below changes this behavior, extending session's lifetime by
+            # `PERMANENT_SESSION_LIFETIME` seconds. `SESSION_PERMANENT` option
+            # is used as indicator of permanent sessions by flask-session
+            # package, so we also should rely on it, for predictability.
```
pdelboca, line 58: I would move this as docstring of the class.

Also, why are we moving out of the default behavior? Just for predictability or is there any CKAN related feature that requires this? If so, it would be nice to document it.

ckan/config/middleware/__init__.py
```diff
@@ -34,3 +39,37 @@ def make_app(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+            session.setdefault("_permanent", app.config["SESSION_PERMANENT"])
+
+        return session
+
+
+class CKANRedisSessionInterface(RedisSessionInterface):
+    """Flask-Session redis-based sessions with CKAN's Redis connection.
+    """
```
pdelboca, line 66: I would document here as well why do we need this level of customization. (For future maintainers)

pdelboca (changes requested):

ckan/config/middleware/__init__.py
smotornyuk: I prefer to keep this explanation directly around this line because it doesn't change the general behavior of the session storage, just adds a small feature. And usually, features are described in the place, they are implemented. But yes, it has sense to mention this in the class documentation as well, to explain the puprose of the whole custom class

> why are we moving out of the default behavior
I think you misunderstood the idea. We are not moving out of the default behavior. We are adding **a new feature** and immitating default behavior for it. In the comment I explained exactly it:
1. cookie-session expires with browser session out of the box
2. we change this behavior in the line below the comment
3. To be predictable(i.e. be similar to other session backends), we rely on `SESSION_PERMANENT` flag

> Also, why we need custom logic for this
Once again, misundestanding? You pointing at documentation of `flask-session` but asking about `flask`'s core functionality. `flask-session` describes only it's own behavior. Whatever it says about `PERMANENT_SESSION_LIFETIME` is not necessarily true for core `flask`. And that's exactly why we need custom logic - to immitate behavior of `flask-session` for `flask`'s core cookie-based session

ckan/config/config_declaration.yaml
```diff
@@ -258,99 +242,141 @@ groups:
...
+          reliably keep the session from expiring, but uses more
+          bandwidth. Non-permanent sessions are not affected.
 
-      - key: beaker.session.validate_key
-        validators: not_empty
-        placeholder_callable: secrets:token_urlsafe
-        validators: configured_default("SECRET_KEY",None)
-        callable_args:
-          nbytes: 20
+      - key: SESSION_PERMANENT
```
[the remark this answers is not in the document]
smotornyuk: Yes, I am also interested in why you are so sure, that ours defaults to the browser's session?:)

ckan/config/middleware/__init__.py
smotornyuk: Thanks, good catch

ckan/config/config_declaration.yaml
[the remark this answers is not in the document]
pdelboca: The config description says ` When disabled, session expires with the browser session.` That's why I am asking why is that behavior instead of Flask's default.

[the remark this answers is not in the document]
smotornyuk: PS and you mention that "our defaults to browser session". I suspect that you were just distracted by someone from a review of this PR, so let us see why it's not so:

```yaml
- key: SESSION_PERMANENT
  default: true
```
Looks like by default this option IS enabled. Let's check what the description says:

>  When **disabled**, the session expires with the browser session. When **enabled**, the session's expiration time is set to ``PERMANENT_SESSION_LIFETIME`` seconds.

The first sentence that mentions browser session is not our case, because we enable this flag by default. The second sentence explains that our session's lifetime is controlled by `PERMANENT_SESSION_LIFETIME`. So, yes, we are not controlled by browser sessions by default.

pdelboca:
Seems okay @smotornyuk !

I'm just adding @wardi and @amercader for an extra pair of eyes.

mhazan01:
any update on this   ? the beaker cache sessions exploding is killing my server.

ckan/config/middleware/flask_app.py
```diff
@@ -69,6 +69,24 @@
...
+        If config file contains custom import path for ``SESSION_TYPE``, use it
+        to allow session customizations. If custom import path is missing,
+        initialize one of interfaces provided by flask-session module.
+
+        """
+        session_type = app.config["SESSION_TYPE"]
+        option_name = f"ckan.session.custom_type.{session_type}.import_path"
+
+        if path := config.get(option_name):
+            return import_string(path)(app)
```
wardi, line 85: Importing arbitrary python code from the INI/environment variables doesn't align with ckan's current pattern of implementing interfaces and enabling plugins. Is there a reason for this different approach here?

wardi: If there isn't an immediate need for this interface I suggest we drop it and discuss the best way to implement pluggable session types. `import_string(s)` is roughly the same as `exec(s)` so I'm leery about including it here.

amercader: Yes, perhaps adding support for custom interfaces is premature. The whole point of Flask-Session is that provides many options already out of the box.
I was also going to suggest that we remove the [`ckan.session.custom_type.cookie.import_path`](https://github.com/ckan/ckan/pull/7893/files#diff-5752be3d55839c292941807dff4cb1943222e276a2d2a64888e23f5b49104628R343) and [`ckan.session.custom_type.redis.import_path`](https://github.com/ckan/ckan/pull/7893/files#diff-5752be3d55839c292941807dff4cb1943222e276a2d2a64888e23f5b49104628R351) options. The wrappers @smotornyuk introduced make perfect sense in the context of CKAN, I don't think they need to be customizable.

amercader:
@smotornyuk if you can have a quick look at https://github.com/ckan/ckan/pull/7893/commits/4325d3a9e8e372364a9537341164f4416d662c04 to make sure I haven't missed anything that would be great. I moved the session interface classes to another module to avoid circular import errors.

smotornyuk:
Everything is good. There is a mention of custom interfaces in changelog: "See :ref:`SESSION_TYPE` for alternative backends available, including **custom ones**.", but it may just refer to all 5 session storages supported by flask-session, so it's ok to keep this wording unchanged.

PR #7897, opened 2023-11-08
[7720]Fix group/org index views unsafe and slow 

### LINKED ISSUE #7720: group/org index views unsafe and slow
## CKAN version
master

## Describe the bug
The group/org index view makes use of mutable global state and repeats expensive operations for no reason

### Steps to reproduce
Navigate to /group/ or /organization/ on a ckan site with many groups/orgs

### Expected behavior
Not be unnecessarily slow or unsafe when running with threads

### Additional details
Global mutable state (unsafe with threads and generally a bad idea):
https://github.com/ckan/ckan/blob/9956f92925740957b119bd7c6719ee202eb51082/ckan/views/group.py#L124-L131

Duplicated, expensive `group_list` calls:
https://github.com/ckan/ckan/blob/9956f92925740957b119bd7c6719ee202eb51082/ckan/views/group.py#L173

https://github.com/ckan/ckan/blob/9956f92925740957b119bd7c6719ee202eb51082/ckan/views/group.py#L197-L205

Current implementation of `group_list` calls group/organization_show for each item returned so output schema validation can be done, calling these *twice* makes performance even worse for no clear benefit.

TomeCirun:
Fixes #7720 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/group.py
+++ b/ckan/views/group.py
@@ -45,7 +45,7 @@ log = logging.getLogger(__name__)
 lookup_group_plugin = lib_plugins.lookup_group_plugin
 lookup_group_controller = lib_plugins.lookup_group_controller
 
-is_org = False
+# is_org = False
 
 
 def _get_group_template(template_type: str,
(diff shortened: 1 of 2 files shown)


ckan/logic/action/get.py
```diff
@@ -338,12 +339,16 @@ def _group_or_org_list(
     all_fields = asbool(data_dict.get('all_fields', None))
 
     if all_fields:
+        if set_max_limit:
```
amercader, line 342: @TomeCirun @wardi I was the one who introduced the two `group_list` calls in the group index view function (then controller) in https://github.com/ckan/ckan/pull/2554, believe it or not to improve performance :)

The reasoning was that for the index page we need the group details for all items in that page (`all_fields=True`) but also we need the total number of groups to be able to paginate properly (2on call with `all_fields=False`). I know it's not elegant but going back to a single call with `all_fields=True` without limit will reduce performance for sure.

@wardi can you think of an alternative approach?

amercader:
@TomeCirun I took the liberty of extract and add the "remove mutable global state" bit of this PR into https://github.com/ckan/ckan/pull/8359, as it is needed fo the 2.11 release

TomeCirun:
Hi @amercader @wardi  ,

I’ve come up with a more elegant solution for pagination in the `group.index` blueprint. Please review my latest commit. If you agree with the changes, I can create a separate PR to focus solely on improving the index functionality, as you’ve already addressed the `remove mutable global state` part.

Thanks!

PR #7903, opened 2023-11-09
Fix performance for users following many groups/orgs/users #7878

EricSoroos:
Master version or  #7878

### Proposed fixes:
Replace unions with in (list).

Also replace an outer join with a subselect, so that it can be indexed.

### Features:

- [ ] includes API changes


smotornyuk:
More or less proper type definitions exist only for sqlalchemy v2.0, so the best you can do now - add inline comment `  # type: ignore` for every problematic line. [like here](https://github.com/ckan/ckan/blob/master/ckan/authz.py#L347)

EricSoroos:
Thanks. Hopefully that does it.

ckanext/activity/model/activity.py
```diff
@@ -15,6 +15,8 @@
     not_,
     union_all,
     text,
+    select,
+    literal
```
ThrawnCA, line 19: It would be better to have a trailing comma here, so that adding more lines (as demonstrated in this change) doesn't require changing an existing line.

ckanext/activity/model/activity.py
```diff
@@ -444,28 +451,35 @@ def _organization_activity_query(org_id: str) -> QActivity:
...
-            or_(
-                model.Package.owner_org == org_id, Activity.object_id == org_id
+            #
+            # Use subselect instead of outer join so that it can all
+            # be indexable
+            Activity.object_id.in_(  # type: ignore
+                select([model.Package.id])  # type: ignore
+                .where(
+                    and_(
+                        model.Package.private == False,  # noqa
```
wardi, line 475: activities for private datasets are supported now

ThrawnCA:
https://github.com/qld-gov-au/ckan/pull/152 has some optimisation based on not loading the Data column unnecessarily.

wardi (changes requested):
These look good and would love to have in 2.11 just need to remove the private check

wardi (approved):

wardi:
> [qld-gov-au#152](https://github.com/qld-gov-au/ckan/pull/152) has some optimisation based on not loading the Data column unnecessarily.

@ThrawnCA this sounds great too, would you submit that as a separate PR if it improves performance when combined with this change?

PR #7917, opened 2023-11-16
Datapusher: Use IDomainObjectModification Implementation

JVickery-TBS:
# Fixes

The `notify` method of the `IResourceUrlChange` assumes that the Resource exists as it calls the `resource_show` action method. But, at this point, other extensions can override/extend action methods to delete a Resource from the session. This namely happens with the Validation extension running in sync mode: if the validation fails, it will delete the Resource from the session and it won't get created but then Datapusher's `notify` method would fire off and not find the Resource in `resource_show`

Just changing this to use `IDomainObjectModification` will make is handle the exceptions automagically: https://github.com/ckan/ckan/blob/master/ckan/model/modification.py#L67

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datapusher/plugin.py
+++ b/ckanext/datapusher/plugin.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 from ckan.common import CKANConfig
 from ckan.types import Action, AuthFunction, Context
 import logging
-from typing import Any, Callable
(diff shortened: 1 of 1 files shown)


KatiRG:
some context: https://github.com/frictionlessdata/ckanext-validation/issues/94

PR #7918, opened 2023-11-17
Remove Recline view plugins

amercader:
Following up on the deprecation work done in CKAN 2.10 (#7078) this finally removes the venerable Recline-based view plugins from core.

Users should use the DataTables view plugin or other more advanced extensions:

https://docs.ckan.org/en/2.10/maintaining/data-viewer.html#other-view-plugins


pdelboca:
Thanks @amercader ! I have tested this one and looks good to go!

PR #7923, opened 2023-11-20
Use Pluggable Helpers for Views

JVickery-TBS:
fix(views): pluggable helpers;

- Use `toolkit.h` instead of `lib.helpers`.

### Proposed fixes:

View methods should use the pluggable helpers from `ckan.plugins.toolkit` instead of directly from `ckan.lib.helpers`.

### Features:

- [ ] includes API changes


smotornyuk:
@JVickery-TBS , this change has no problems, but it can create a misleading example. 

toolkit exists for extensions, so that they can use a single entrypoint to get most used CKAN features. If something changes inside CKAN, we slightly alter toolkit, and plugins still work without any changes.

Inside `ckanext` folder of CKAN it's ok to use the toolkit, but inside `ckan` itself, it's better to avoid it as much as possible. For example, instead of `toolkit.h` you can use `from ckan.lib.helpers import helper_functions as h`.

In this case, there will be no difference, because you are only updating `views` module, that is not used by toolkit.

But imagine that you do the similar change to [base module](https://github.com/ckan/ckan/blob/master/ckan/lib/base.py#L20). You import toolkit from `base`, but toolkit itself [relies on `base`](https://github.com/ckan/ckan/blob/master/ckan/plugins/toolkit.py#L16) so we'd cause a circular import. So in this case the only valid option is to use `from ckan.lib.helpers import helper_functions as h`. And for consistency(and in case toolkit start importing views in future), CKAN core should never use `toolkit` and prefer `from ckan.lib.helpers import helper_functions as h` instead

JVickery-TBS:
@smotornyuk does that mean that developers are unable to extend or override the flash helpers used throughout the views modules? or should modifying helpers in `ITemplateHelpers` implementations affect helpers used via `from ckan.lib.helpers import helper_functions as h`??

From my testing, I was unable to extend the flash message helpers like that, but I can try again and debug some more?

smotornyuk:
`toolkit.h` is an alias for `helpers.helper_functions`. So whatever `ITemplateHelpers` does, it will be reflected on `helper_functions`.

You probably were confused by popular(though, wrong) import: `import ckan.lib.helpers as h`. In this case, we are importing `helpers` module directly, and `h.flash_notice` refers to the original implementation. You cannot call any custom helper registered via `ITemplateHelpers`, nor you'll see any overrides/chained_helpers, because `h` contains only the original code from CKAN.

The correct form of this import is `from ckan.lib.helpers import helper_functions as h`. Here we are not importing `helpers` module, but instead import only one member from it. This `helper_functions` object - it's a dynamic collection of all helper function available in CKAN. Via it you can access all helpers registered via `ITemplateHelpers` and you also will see all overrides/chained functions for existing actions. This `helper_functions` is exactly what is aliased as `h` inside toolkit

JVickery-TBS:
@smotornyuk ohh I see! Okay thanks! That make a lot of sense.

Okay I will update this PR to use `from ckan.lib.helpers import helper_functions as h`

PR #7925, opened 2023-11-21
[7915]Fix HTML5 validation fails on resource uploads

### LINKED ISSUE #7915: HTML5 validation fails on resource uploads
## CKAN version

2.10

## Describe the bug

Uploaded resources can get into a state where HTML5 validation fails, complaining that the value of `field-resource-url` is not a proper URL.

### Steps to reproduce

- Create a dataset.
- Create a resource by uploading a file.
- Edit the resource and save it without making changes.
- Edit and save the resource again. The "Update Resource" button will silently fail.

### Expected behavior

Upload URLs should not be subject to HTML5 validation, and if they are, it should be visible rather than silent.

### Additional details

The failure can be made more visible by inspecting the page and disabling the CSS rule that hides `.resource-upload-field > div.select-type`. Clicking on Update Resource will then give a proper error message attached to the URL field, albeit the error should not occur.

The error will temporarily go away if the dataset, rather than the resource, is edited and saved. However, it will return if resources continue to be edited and saved.

TomeCirun:
Fixes #7915 

### Proposed fixes:
After debugging for a while, I found two potential solutions for the issue:
 
1). Change Input Type (Implemented in my PR): This approach involves altering the type of the hidden input to text, effectively bypassing HTML5 validation. The type reverts to 'url' when the upload is cleared, and the 'LINK' option is chosen.

2). Removing this [check](https://github.com/ckan/ckan/blob/master/ckan/lib/dictization/model_dictize.py#L132): An alternative option is to eliminate the "`for_edit`" check, allowing the complete qualified URL to be returned.

Please let me know your preference for the approach to proceed with.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/modules/resource-upload-field.js
+++ b/ckan/public/base/javascript/modules/resource-upload-field.js
@@ -1,10 +1,21 @@
 this.ckan.module('resource-upload-field', function (jQuery) {
   var _nameIsDirty = !! $('input[name="name"]').val();
+  var urlField = $('#field-resource-url');
   return {
     initialize: function() {
       $('input[name="name"]').on('change', function() {
         _nameIsDirty = true;
       });
+
+      // Change input type to text if Upload is selected
+      if ($('#resource-url-upload').prop('checked')) {
+        urlField.attr('type', 'text');
+      }
+
+      // revert to URL for Link option
+      $('#resource-link-button').on('click', function() {
+        urlField.attr('type', 'url');
+      }) 
 
       $('#field-resource-upload').on('change', function() {
         if (_nameIsDirty) {


ThrawnCA:
I would lean toward making the least change, so apply url validation only if we actually expect a URL. That way, the field contents are the same as they were before, we just partially roll back the extra validation that 2.10 added.

ThrawnCA:
You need a changelog entry to make towncrier happy. Do you want to just cherry-pick https://github.com/qld-gov-au/ckan/commit/097e77d34c9e0b8b195c7f75384a092333ae58d1 ?

amercader:
Thanks @ThrawnCA and @TomeCirun, I added the changelog entry

Chealer:
Thank you very much @TomeCirun and @amercader. This was backported to 2.10 via 2d60c7f and fixed in CKAN 2.10.3.

Chealer:
I tried to cleanup the current workaround with the following:
```
// Workaround #7915 (HTML5 validation silently fails on some resource uploads)
$( "#resource-edit" ).on( "submit", function( event ) {
    if ($('#resource-url-upload').prop('checked')) {
      // Make the irrelevant URL field type "text" rathen than "url", so the form can be submitted even if the field's value is not a valid URL
      jQuery('#field-resource-url').attr('type', 'text');
    }
});
```
This unfortunately fails (does nothing) because the `submit` event is fired **after** built-in form validation.
A clean workaround would require [an event fired *before* form submission, which doesn't exist](https://stackoverflow.com/questions/6871428/event-in-html5-form-raised-before-validation-of-input-fields).

The following could work, but there may be browsers where `click` doesn't always fire before `submit`:
```
// Workaround #7915 (HTML5 validation silently fails on some resource uploads)
$( "#resource-edit" ).on( "click", function( event ) {
  // If the resource is not a link, make the irrelevant URL field's type "text" rathen than "url", so that the form can be submitted even if the field's value is not a valid URL.
  jQuery('#field-resource-url').attr('type', $('#resource-url-upload').prop('checked') ? 'text' : 'url');
});
```

If the current approach is kept, it would at least clarify to comment the added workaround with something like:
```
// Begin workaround #7915 (HTML5 validation silently fails on some resource uploads)
      // Change input type to "text" if Upload is selected, to disable URL validation
      if ($('#resource-url-upload').prop('checked')) {
        urlField.attr('type', 'text');
      }

      // revert to "url" when the user selects the Link option
      $('#resource-link-button').on('click', function() {
        urlField.attr('type', 'url');
      }) 
// End workaround
```

PR #7932, opened 2023-11-22
[7836]Fix Cannot create API token via API (hen and egg)

### LINKED ISSUE #7836: Cannot create API token via API (hen and egg)
## Describe the bug
CKAN 2.9 supported API keys which made it possible to create a user via the API (`ckan.auth.create_user_via_api = true`) and automatically return the user's API key (`ckan.auth.create_default_api_keys = true`).

Since CKAN 2.10 does not support API keys, it is not possible anymore to use the API anonymously to create users and start working as them. An API token is required to create an API token for a user. So users would have to log in via the web interface first to create an API token.

TomeCirun:
Fixes #7836 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -1043,7 +1043,7 @@ def user_create(context: Context,
     if with_apitoken:
         # Create apitoken for user.
         api_token = _get_action("api_token_create")(
-            context, {"user": user.name, "name": "test"}
+            context, {"user": user.name, "name": "default"}
         )
         user_dict["token"] = api_token["token"]
 
--- a/ckan/logic/action/create.py
+++ b/ckan/logic/action/create.py
@@ -965,7 +965,9 @@ def user_create(context: Context,
                 }
             }
     :type plugin_extras: dict
-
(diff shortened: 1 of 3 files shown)


TomeCirun:
@smotornyuk done

ckan/logic/action/create.py
```diff
@@ -1032,10 +1035,18 @@ def user_create(context: Context,
...
     dashboard = model.Dashboard(user.id)
+
     session.add(dashboard)
     if not context.get('defer_commit'):
         model.repo.commit()
 
+    if with_apitoken:
+        # Create apitoken for user.
+        api_token = _get_action("api_token_create")(
+            context, {"user": user.name, "name": "default"}
```
smotornyuk, line 1046: It will cause an AuthorizationError because you are potentially using anonymous context

smotornyuk:
Found a problem - anonymous user will get an error during api request

ckan/logic/action/create.py
TomeCirun: @smotornyuk done

PR #7937, opened 2023-11-23
[7603]Replace Beaker as session management library

### LINKED ISSUE #7603: Replace Beaker as session management library
CKAN has historically used the [Beaker](https://github.com/bbangert/beaker) library to manage sessions. It was agreed that we should explore moving to an alternative due to concerns about maintenance and security, and to better align with the rest of the Flask ecosystem.

Perhaps the best alternative is [Flask-session](https://flasksession.readthedocs.io), which has recently become part of the [Pallets ecosystem](https://github.com/pallets-eco/flask-session).

The code changes would probably be [easy enough](https://flasksession.readthedocs.io/en/latest/quickstart.html), but we would need to review all the session/beaker related init steps (like [these](https://github.com/ckan/ckan/blob/d182703d28640f1a241a7e420fd996439b8fc378/ckan/config/middleware/flask_app.py#L175)) to remove what is no longer needed. Ideally everything would be set up via [config options](https://flasksession.readthedocs.io/en/latest/config.html). We can decide if we want to keep some form of compatibility between the existing config options (`beaker.session.*`) and the new ones.

We should also take the opportunity to think of the best default configuration for a new CKAN instance, document options properly and provide some examples.

In https://github.com/ckan/ckan/pull/7464 we are already moving to defaulting to client-based sessions on new sites, which is the same default that Flask has. Flask-session would enable server-side sessions for those users that would rather keep that data in the server (Flask-session supports the same backends as Beaker, file-system, redis, database, etc)

TomeCirun:
Fixes #7603 

### Proposed fixes:
more info soon...

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -88,17 +88,7 @@ groups:
         internal: true
       - key: SERVER_NAME
         internal: true
-      - key: SESSION_COOKIE_DOMAIN
-        internal: true
-      - key: SESSION_COOKIE_HTTPONLY
-        internal: true
-      - key: SESSION_COOKIE_NAME
-        internal: true
       - key: SESSION_COOKIE_PATH
-        internal: true
-      - key: SESSION_COOKIE_SAMESITE
-        internal: true
-      - key: SESSION_COOKIE_SECURE
         internal: true
       - key: SESSION_REFRESH_EACH_REQUEST
         internal: true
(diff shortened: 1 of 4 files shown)


amercader:
@TomeCirun are you aware of the ongoing work on https://github.com/ckan/ckan/pull/7893?

TomeCirun:
Nope :( I looked at the ticket and there wasn't any linked PR... @amercader my bad I'm sorry

PR #7938, opened 2023-11-24
Use current CKAN version in cookiecutter tests runner template

amercader:
Replace the hardcoded Docker tags with the actual current version number, so when the extension is generated they point to the correct version.

TODO: create major.minor tags for the ckan images

### DIFF
--- a/changes/7938.minor
+++ b/changes/7938.minor
@@ -1,0 +1,1 @@
+Use correct Docker tag in generated extensions test runner configuration file.
--- a/ckan/cli/generate.py
+++ b/ckan/cli/generate.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import contextlib
 import os
 import json
+from packaging.version import parse as version_parse
 import shutil
 from typing import Optional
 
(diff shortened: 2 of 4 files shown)


amercader:
@kowh-ai I don't think we are pushing `major.minor` tags to Docker Hub (eg `2.10`, `2.10-dev`) for the `ckan/ckan-base` and `ckan/ckan-dev`. I think they are useful for cases like this one where you always want to target the latest patch release.

What do you think?

kowh-ai:
Here is a snaphot of the DockerHub `ckan/ckan-base` repo with the tags currently used:

Screenshot 2023-11-24 at 1 19 25 pm

So you think we should also push extra tags eg: `ckan/ckan-base:2.10` when a new patch version number is released? So a new 2.10 image (built from **2.10.2**) will override the previous 2.10 image (built from **2.10.1**)

Fine by me

amercader:
@kowh-ai precisely, just like [solr](https://hub.docker.com/_/solr), [postgres](https://hub.docker.com/_/postgres), [alpine](https://hub.docker.com/_/alpine), [python](https://hub.docker.com/_/python) etc do it

kowh-ai:
@amercader - created a PR in the ckan-docker-base repo: https://github.com/ckan/ckan-docker-base/pull/37 to address this. Let me know what you think

kowh-ai (approved):

PR #7947, opened 2023-11-28
[#7945] Translate aria-label attributes

### LINKED ISSUE #7945: aria-labels without translations
## CKAN version
2.10

## Describe the bug
There are some aria-labels that are not covered by the translations, causing a mis-match in the UI language for people that are using screen readers.

### Steps to reproduce
Using a CKAN instance with a German localization (100% translated according to transifex), such as https://opendata.leipzig.de:
- shrink screen to "smartphone" size: Navigation menu button is labeled "Toggle navigation"
- fail to log in, error message close button is labeled "Close"

### Expected behavior
Navigation elements have localized labels.

## Comments
I might expand this list if more missing label translations are found. I'm also looking into transifex and whether I can get this fixed myself.

amercader:
Fixes #7945

For HTML templates we just wrap the value in the `{{ _() }}` function so they get extracted. For markup that is generated from JS modules, we need to find the attribute and replace the value with the translated string. "Close" was not extracted up until now, so once this is merged we'll need to run `extract_messages` so it goes to the pot file > Transifex > po files (in 2.10.2 and 2.11.0)


ckanext/datastore/assets/js/api-info.js
```diff
@@ -64,6 +64,7 @@ this.ckan.module('api-info', function (jQuery) {
       this.loadTemplate().done(function (html) {
         module.modal = jQuery(html);
         module.modal.find('.modal-header').append('<button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>');
+        module.modal.find('.btn-close').attr("aria-label", this._('Close'));
```
pdelboca, line 67: This line in raising:

```
Uncaught TypeError: this._ is not a function
    show http://localhost:5000/storage/webassets/webassets-external/84c5d34080f70b747a647073a4d19589_api-info.js:67
    jQuery 4

```

pdelboca (commented):
@amercader looks good! Check my comment, probably in that context we cannot call to `this._()`

amercader:
Thanks @pdelboca , should be all good now

PR #7967, opened 2023-12-07
2023-12 dependency update(Flask v3)

smotornyuk:
Update dependencies to the latest versions

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -32,7 +32,7 @@ from ckan.common import asbool, config, current_user
 from flask import flash
 from flask import get_flashed_messages as _flask_get_flashed_messages
 from flask import redirect as _flask_redirect
-from flask import _request_ctx_stack
+from flask.globals import request_ctx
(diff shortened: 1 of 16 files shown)


ckan/lib/helpers.py
```diff
@@ -311,7 +311,7 @@ def _get_auto_flask_context():
     from ckan.config.middleware import _internal_test_request_context
 
     # This is a normal web request, there is a request context present
-    if _request_ctx_stack.top:
+    if request_ctx:
```
pdelboca, line 314: Flask has a specific function to test this. Can we use it instead?

https://flask.palletsprojects.com/en/3.0.x/api/#flask.has_request_context

pdelboca (approved):
Looks good and working @smotornyuk !

Just one small comment and we are ready to go 🚀

ckan/lib/helpers.py
smotornyuk: Sure, one moment

PR #7968, opened 2023-12-07
Reworked Resource Plugin Hooks for Better Framework

JVickery-TBS:
Feature/resource plugin hooks

### Proposed fixes:

The Resource plugin hooks from `IResourceController` only happen during the `resource_update`, `resource_create` and `resource_delete` action methods. However, users are able to create, update, and delete Resources via the API through the endpoint `api/action/package_update`. In this case, none of the `IResourceController` plugin hooks happen.

As the resource action methods all call `package_update` anyways, I have moved all of the `IResourceController` code into the `package_update` and `package_create` methods.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -329,7 +329,7 @@ def package_update(
 
                 if resource_id in updated_resource_ids:
                     if updated_resource_id and \
-                      resource_dict.get('id') != updated_resource_id:
+                      new_resource.get('id') != updated_resource_id:
                         continue
                     if resource_id in old_resource_dicts:
                         plugin.before_resource_update(


JVickery-TBS:
@wardi okay I had to add some more code in here. Had to make sure that the after hooks weren't running for every single thing every time. And a small fix for resource delete hooks too.

I also made it so the code is not calling the dictize functions, used `package_show` instead and looped through the resource dicts in there. However, not sure if that is better or worse than calling the dictize functions per resource as not all resources are going to have their hooks executed?

but here are some simple stats on the execution times of the `resource_update` and `package_update` action methods:

Updating a single Resource belonging to a Dataset with 345 Resources:
- Current Code: 3.462445020675659
- PR w/ calling dictize functions: 5.979736328125
- PR without dictize functions (using package_show): 4.322240591049194

Notes: because the `resource_update` has a context key,value to specify its ID, we are doing hooks on every single resource. However, there is still all the logic to get the current resource ids etc so it does still do more things than the current code...but we might be able to get around this by conditioning that context key,value more? Like, if the context for `resource_update` is set, don't go looping through resources we don't need to?

Updating a Dataset with 345 Resources (vie the web form):
- Current Code: 1.2072818279266357
- PR w/ calling dictize functions: 1.055083990097046
- PR without dictize functions (using package_show): 2.5762579441070557

Notes: the view method uses `allow_partial_update`, so it doesn't actually send any Resources in the dict. So there is not really any resources to handle. Again, we can probably condition this `allow_partial_update` context key more to prevent unnecessary looping through the resources??


I did decide to see if doing a `package_patch` API call would work. So I ran the `package_patch` without supplying `resources` list. So the `resources` will get added into the `data_dict` in the `package_patch` action method, and sent to the `package_update` method. So at this point, `package_update` will have every single resource in the `data_dict`. So we can expect some major loopage here.

Updating a Dataset with 345 Resources (via package_patch API call):
- Current Code: 4.102233648300171
- PR w/ calling dictize functions: 11.80790114402771
- PR without dictize functions (using package_show): 8.980406761169434

Notes: yeah...so as you said, calling dictize methods on all the things hurts. The usage of `package_show` is also pretty painful here too.

Overall, I think we for sure need a way to see if a Resource is actually different at the beginning of the `package_update` action method.

I am going to try to condition some of these loops with the `allow_partial_update` and `_updating_resource` context keys.

I am also thinking that in `package_patch` action method, we can probably check if `resources` is in the `data_dict`, and if it is not, then we can do an `allow_partial_update` context, and exclude getting the resources in the `package_show` call in the `package_patch` method.

JVickery-TBS:
@wardi okay well doing `allow_partial_update` in `package_patch` if no resources are supplied helps for sure.

But as for the rest of the code that is now in the `package_update` method, I feel like it is terrible and will not work in every case. It is also just extremely bloated now, and will make future development a nightmare. We might have to do this a different way.

I feel like Resources should have their own `plugin_validate` and the Resource action methods should not even call the package update. Moreover, it would be optimal if the Package methods did not even accept a list of resources. Is this all just so the dataset object in SOLR gets updated with all of the Resource dicts? I just don't know how it would be possible for us to know which resources have been updated, deleted, or created within the `package_update` action method to make this code not suck.

JVickery-TBS:
https://github.com/JVickery-TBS/ckan/tree/feature/resource-plugin-hooks-v2 here is another branch that just has some comments and new context keys for a general idea of the concept

JVickery-TBS:
@wardi did we want to move this PR to an Issue for discussion? Or just add the label "Awaiting tech team feedback"??

PR #7969, opened 2023-12-08
Invalidate page caches when cookies change, #6955

ThrawnCA:
Mitigates #6955

### Proposed fixes:

Add 'Cookie' to the 'Vary' response header, to tell browsers that page caches should be dropped if cookies change.

This is necessary because the page header changes based on being logged in or out, along with page elements like Manage buttons. Also, pages containing forms should not be cached if the CSRF token changes.

### Features:

- [ ] includes API changes


ThrawnCA:
@wardi This has now changed thanks to input from @EricSoroos

ckan/views/__init__.py
```diff
@@ -62,6 +63,10 @@ def set_cache_control_headers_for_response(response: Response) -> Response:
     else:
         response.cache_control.private = True
 
+    # Invalidate cached pages upon login/logout
+    if limit_cache_by_cookie:
+        response.vary.add("Cookie")
+
```
wardi, line 69: This should be in the if: block above because otherwise caching is disabled, right?

ThrawnCA: No. As of CKAN 2.10, "disabled" isn't really disabled; it's just "private to the browser". Which is secure, but doesn't fix the problem of pages being cached beyond logout.

EricSoroos: Is CSRF going to generally be applicable for non-logged in users? Or rather, do we set that cookie for pages where there's no form for non-logged in users?  If we do set a csrf cookie on everything, then this essentially blows caching for all html responses. 

In general, I think there are three cases:

* Caching Allowed (public pages, assets, especially the home page when not logged in)
* Private caching allowed (e.g. logged in + back button, on most of the site)
* Caching disabled (login/logout redirect targets)

ThrawnCA: > If we do set a csrf cookie on everything, then this essentially blows caching for all html responses.

If there is a constantly changing CSRF token in the forms, then caching them will break the site. You'll have users submitting stale tokens and getting errors.

It's okay if the token is generated once per session, or once every 30 minutes. But if it's created anew for each request, then caching is not an option.

> Caching Allowed (public pages, assets, especially the home page when not logged in).

As soon as you *do* log in, though, every cached page needs to be invalidated for you.

> Private caching allowed (e.g. logged in + back button, on most of the site)

This needs to be invalidated as soon as you log out. 

> Caching disabled (login/logout redirect targets)

Well, those should have been POST requests all along, which takes care of it.

ThrawnCA:
When it comes down to it, almost any page can change if you log in or out, if only in the header. So caching ought to be invalidated if your cookies change.

ThrawnCA:
@wardi There are probably more elegant ways to implement this, but I do think something like it is required. It's been working well on data.qld.gov.au for months now.

ThrawnCA:
@wardi @EricSoroos Bump. This or something like it is still needed, because the 'private' cache behaviour (ensuring data doesn't leave the browser) is inadequate to handle basic scenarios like logging out, or CSRF tokens being rotated.

PR #8012, opened 2024-01-12
[WIP] Add tests whether jwt secret getter actually raises error

toothstone:
I have a suspicion that the exception is not actually raised.

### Features:

- [ ] includes API changes


toothstone:
From the short glance I had at the failover code I got the impression that there are deeper issues with the type annotations vs what is actually handled, given that the tests here indeed didn't get to raise the expected exception, but a TypeError.

PR #8023, opened 2024-01-22
[#7999] Optional session.validate_key

### LINKED ISSUE #7999: Breaking Change in CKAN v2.10.3: Missing beaker.session.validate_key Parameter Causes Errors
## CKAN version
2.10.3
## Describe the bug
We have recently encountered a breaking change in CKAN version 2.10.3 related to the `beaker.session.validate_key` configuration parameter. In our setup, CKAN runs under an NGINX Proxy server, and we use Docker for deployment.

**Previous Behavior (v2.10.1):**
In CKAN version 2.10.1, which we were using, the command for generating the `ckan.ini` configuration file did not include the `beaker.session.validate_key` parameter. This led us to believe that the parameter was not mandatory or automatically handled by CKAN.

**Current Issue (v2.10.3):**
Upon updating to version 2.10.3, we faced accessibility issues with CKAN. It could not accept requests due to an error stemming from the absence of the `beaker.session.validate_key` parameter in the configuration. This parameter is now enforced as mandatory in `config_declaration.yaml` as per commit `06d72d2d8cfb61e4773097cd80f768a323b1ffef`.

**Impact:**
This change has caused significant disruption as our application, which was expected to run smoothly post-upgrade, failed to do so. It took us a considerable amount of time to identify the root cause of the issue.

### Proposed Solution:

1. Update Documentation: Clearly document this change in the migration guide for version 2.10.3 to alert users about the need to include `beaker.session.validate_key` in their `ckan.ini` file.
2. Backward Compatibility: Consider implementing a fallback mechanism in CKAN for this parameter, where it generates a default value if not explicitly provided in the configuration. This would ensure smoother transitions for existing deployments.
3. Improved Communication: It would be beneficial for the CKAN community to proactively communicate such breaking changes in future releases.

smotornyuk:
Fixes #7999

:information_source: PR into dev-v2.10. beaker is replaced by flask-sessions upstream so this change makes no sense for the master branch

When `beaker.session.validate_key` is empty, use `beaker.session.secret` as the default

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -318,7 +318,7 @@ groups:
           The dbm option is not recommended as it is not thread-safe.
 
       - key: beaker.session.validate_key
-        validators: not_empty
+        validators: configured_default("beaker.session.secret",None) not_empty
         placeholder_callable: secrets:token_urlsafe
         callable_args:
           nbytes: 20
--- a/changes/8023.bugfix
+++ b/changes/8023.bugfix
@@ -1,0 +1,1 @@
+CKAN does not start without ``beaker.session.validate_key`` option introduced in v2.10.3


ckan/config/config_declaration.yaml
```diff
@@ -318,7 +318,7 @@ groups:
           The dbm option is not recommended as it is not thread-safe.
 
       - key: beaker.session.validate_key
-        validators: not_empty
+        validators: configured_default("beaker.session.secret",None) not_empty
```
ThrawnCA, line 321: Missing space after comma

Suryakandukuri: This gave rise to the below error.

```
ckan.logic.UnknownValidator: Validator `configured_default("beaker.session.secret",` does not exist
unable to load app 0 (mountpoint='') (callable not found or import error)
*** no app loaded. going in full dynamic mode ***
  File "/usr/lib/ckan/default/src/ckan/ckan/logic/__init__.py", line 848, in get_validator
    return _validators_cache[validator]
KeyError: 'configured_default("beaker.session.secret",'
```

smotornyuk: @ThrawnCA , no, that's intended syntax. `validators` evaluated using a strategy from ckanext-scheming: split `v1 v2 v3` on whitespace and then convert every chunk into validator. If you add a space after the comma, you'll get three validators: `configured_default("...",`, `None)` and `not_empty`. And an error, because first two validators do not exist. 

@Suryakandukuri attached example of error that I'm talking about

tino097: @smotornyuk maybe we could add this explanation about spaces as a note in the docs or in the changelog?

ThrawnCA: It seems rather incongruous for the parser to recognise parenthesised function calls, and yet not be able to apply the parentheses when parsing spaces.

smotornyuk: Just a bit of explanation, why we are not improving it. The whole "validators" idea for config declaration was taken from ckanext-scheming, because it's arguably most used extension for schema modifications. We don't wan't to use different logic, just to share concepts.
It's possible to write a recursive regular expression using 3rd party library, but I prefer not to add new dependencies, especially for such a small task.
As an alternative, we can make a smarter divider, that counts all the quotes, brackets, parenthesis and splits the validator line only when we have all these symbols in balanced state.

But keep in mind, that our goal it to be scheming-friendly. Any change we do here must be adopted by scheming, so that people remain on the same ground. The question is: does this small inconvenience worth keeping two copies of more complex lexer in two different repos?

If in future scheming integrated into CKAN, it may have sense. Or if we see that people often use validator factories with multiple arguments(I think we have just two such validators in CKAN core and they are not the most popular ones). Otherwise, possibility of adding a space between arguments doesn't worth efforts. At least, I don't have time for making this, but I'll definitely will support a PR with such change

PR #8031, opened 2024-01-26
Remove nginx cache as beaker cookies are not cacheable

Zharktas:
Cookies always exist so caching based on cookies doesn't work. We really shouldn't be provide nginx cache instructions anyway, its out of scope of ckan.

### Proposed fixes:
Remove nginx cache from docs.

### Features:

- [ ] includes API changes

### DIFF
--- a/doc/maintaining/installing/deployment.rst
+++ b/doc/maintaining/installing/deployment.rst
@@ -177,7 +177,6 @@ following contents:
 
 .. parsed-literal::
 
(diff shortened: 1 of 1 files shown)


paulmueller:
Please cross-check with #8032 (I think `remember_token` is the new auth cookie)

Zharktas:
@ckan/core fyi: this also needs to fixed in https://github.com/ckan/ckan-packaging/blob/master/common/etc/nginx/sites-available/ckan

kowh-ai (approved):

PR #8033, opened 2024-01-26
AGPLv3 added as LICENSE

### LINKED ISSUE #8030: Can you please include a copy of the AGPLv3 with the code base?
## Describe the bug

It is awesome you have chosen the AGPLv3 for CKAN. However, the AGPLv3 requires that a copy of the license be provided with the source (Section 4). The link in the README is not quite sufficient. I know this seems trivial, and I don't think CKAN (as author and copyright holder) is necessarily obligated* to include the license. However, anybody downstream who clones the CKAN repo and distributes it would be in violation of the license if they don't download a copy of the AGPLv3 and include it in the code base.

*I don't know if this is 100% correct because I am not a lawyer and this is definitely not legal advice.

Can you please include a copy of the AGPLv3 in the code base?

Thanks in advance.

~willtop43

hwpplayer1:
Fixes #8030 

### Proposed fixes:

I added AGPLv3 license to root of the repository

Thanks and happy hacking !

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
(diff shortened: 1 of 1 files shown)


wardi:
This already exists as LICENSE.txt

Issue #8034, opened 2024-01-29
Activity Stream Changes link


blagojabozinovski:
## CKAN version
CKAN 2.10.3

## Describe the bug
Activity Stream Changes link breaks

### Steps to reproduce
Login to CKAN and create a new dataset and add resource. Go to Activity stream and hit the oldest Changes link

### Expected behavior
Show the Changes

### Additional details
Full stack trace of the error raised
Traceback (most recent call last):
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask/app.py", line 2076, in wsgi_app
    response = self.handle_exception(e)
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask_debugtoolbar/__init__.py", line 157, in dispatch_request
    return view_func(**req.view_args)
  File "/home/blagoja/Projects/CKAN/kepa/ckan/ckanext/activity/views.py", line 404, in package_changes
    return tk.render(
  File "/home/blagoja/Projects/CKAN/kepa/ckan/ckan/lib/base.py", line 107, in render
    return flask_render_template(template_name, **extra_vars)
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask/templating.py", line 147, in render_template
    return _render(
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/flask/templating.py", line 128, in _render
    rv = template.render(context)
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/home/blagoja/Envs/kepa02/lib/python3.8/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/home/blagoja/Projects/CKAN/kepa/ckan/ckanext/activity/templates/package/changes.html", line 1, in top-level template code
    {% extends "package/base.html" %}
[...]

PR #8035, opened 2024-01-29
Set dataset type in new resource template

blagojabozinovski:
Fixes #

### Proposed fixes:
Proposed fix to this [issue](https://github.com/ckan/ckan/issues/8034)


wardi:
It would be better to get the type from the db instead of the request parameter (IIUC) but this fix is small and prevents a failure on the url generation.

PR #8053, opened 2024-02-05
Allow alternate Solr query parsers via allow list

amercader:
Back in [2018](https://github.com/ckan/ckan/commit/1d9a5265ddd73afc599657e126ab96310559a4c1), due to a vulnerability in Solr's xml query parser the decision was taken to disallow all [local parameters](https://solr.apache.org/guide/solr/latest/query-guide/local-params.html) in Solr queries.

But alternate query parsers defined via local parameters can provide very interesting functionalities, like the [Dense Vector Search](https://solr.apache.org/guide/solr/latest/query-guide/dense-vector-search.html) leveraged by [ckanext-embeddings](https://github.com/amercader/ckanext-embeddings).

This PR introduces a less draconian approach where a with new `ckan.search.solr_allowed_query_parsers` config option you can define one or more allowed query parsers.

The type of query parser can be defined either via a short form or explicitly with the `type` local param, both are taken into account:

        q={!knn field=vector ...}...
        q={!type=knn field=vector ...}...

I'd like to propose this for backporting to 2.10.x

### DIFF
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -5,7 +5,7 @@ import re
 import logging
 from typing import Any, NoReturn, Optional, Union, cast, Dict
 from pyparsing import (
-    Word, QuotedString, Suppress, OneOrMore, Group, alphas, alphanums
+    Word, QuotedString, Suppress, OneOrMore, Group, alphanums
 )
 from pyparsing.exceptions import ParseException
 import pysolr
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -103,8 +103,8 @@ def _parse_local_params(local_params: str) -> list[Union[str, list[str]]]:
     {!type=dismax qf=myfield v='some value'} -> [['type', 'dismax'], ['qf', 'myfield'], ['v', 'some value']]
 
     """
-    key = Word(alphas)
-    value = QuotedString('"') | QuotedString("'") | Word(alphanums)
+    key = Word(alphanums + "_")
+    value = QuotedString('"') | QuotedString("'") | Word(alphanums + "_$")
     pair = Group(key + Suppress("=") + value)
     expression = Suppress("{!") + OneOrMore(pair | key) + Suppress("}")
 
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -95,6 +95,7 @@ def _get_local_query_parser(q: str) -> str:
     local parameters, .e.g. q={!child ...}...
     """
     qp_type = ""
+    q = q.strip()
     if not q.startswith("{!"):
         return qp_type
 
@@ -412,7 +413,7 @@ class PackageSearchQuery(SearchQuery):
         query.setdefault("df", "text")
         query.setdefault("q.op", "AND")
         try:
-            if query["q"].startswith("{!"):
+            if query["q"].strip().startswith("{!"):
                if not _get_local_query_parser(query["q"]) in config["ckan.search.solr_allowed_query_parsers"]:
                 raise SearchError("Local parameters are not supported.")
         except KeyError:
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -99,7 +99,9 @@ def _get_local_query_parser(q: str) -> str:
     if not q.startswith("{!"):
         return qp_type
 
-    parts = q.lstrip("{!").rstrip("}").split(" ")
+    local_params = q[:q.index("}")]
+
+    parts = local_params.lstrip("{!").rstrip("}").split(" ")
     if "=" not in parts[0]:
         # Most common form of defining the query parser type e.g. {!knn ...}
         qp_type = parts[0]
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -412,12 +412,19 @@ class PackageSearchQuery(SearchQuery):
 
         query.setdefault("df", "text")
         query.setdefault("q.op", "AND")
-        try:
-            if query["q"].strip().startswith("{!"):
-               if not _get_local_query_parser(query["q"]) in config["ckan.search.solr_allowed_query_parsers"]:
-                raise SearchError("Local parameters are not supported.")
-        except KeyError:
-            pass
+
+        def _check_query_parser(param: str, value: Any):
+            if isinstance(value, str) and value.strip().startswith("{!"):
+                if not _get_local_query_parser(value) in config["ckan.search.solr_allowed_query_parsers"]:
+                   raise SearchError(f"Local parameters are not supported in param '{param}'.")
+
+        for param in query.keys():
+            if isinstance(query[param], str):
+                _check_query_parser(param, query[param])
+            elif isinstance(query[param], list):
+                for item in query[param]:
+                    _check_query_parser(param, item)
+
 
         conn = make_connection(decode_dates=False)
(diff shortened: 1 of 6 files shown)


amercader:
@wardi what do you think about getting this backported to 2.10? It would make easier to use stuff like ckanext-embeddings

ckan/lib/search/query.py
```diff
@@ -89,6 +89,26 @@ def convert_legacy_parameters_to_solr(
     return solr_params
 
 
+def _get_local_query_parser(q: str) -> str:
+    """
+    Given a Solr `q` parameter, extract any custom query parsers used in the
+    local parameters, .e.g. q={!child ...}...
+    """
+    qp_type = ""
+    if not q.startswith("{!"):
```
wardi, line 98: do we need to be more careful with this, e.g. is whitespace before `{!` ignored?

wardi:
@amercader yes I think that's a great idea. I need to learn more about solr syntax to fully review this one but I'll try to make it a priority

ckan/lib/search/query.py
```diff
@@ -89,6 +89,27 @@ def convert_legacy_parameters_to_solr(
...
+    """
+    Given a Solr `q` parameter, extract any custom query parsers used in the
+    local parameters, .e.g. q={!child ...}...
+    """
+    qp_type = ""
+    q = q.strip()
+    if not q.startswith("{!"):
+        return qp_type
+
+    parts = q.lstrip("{!").rstrip("}").split(" ")
```
wardi, line 102: local parameters can be combined with parameter values by putting the value after the `}` https://solr.apache.org/guide/solr/latest/query-guide/local-params.html#specifying-the-parameter-value-with-the-v-key so we should split on `}` here instead, right?

ckan/lib/search/query.py
```diff
@@ -89,6 +89,27 @@ def convert_legacy_parameters_to_solr(
...
+    if "=" not in parts[0]:
+        # Most common form of defining the query parser type e.g. {!knn ...}
+        qp_type = parts[0]
+    else:
+        # Alternative syntax e.g. {!type=knn ...}
+        type_part = [p for p in parts if p.startswith("type=")]
+        if type_part:
+            qp_type = type_part[0].split("=")[1]
```
wardi, line 110: what does multiple `type=` parameters do? Is it safe to assume solr uses the first?

wardi:
The docs suggest that local params apply to parameters other than q. If that's true we should make sure we're filtering those as well.

Also it's possible to use parameter de-referencing as part of queries, https://solr.apache.org/guide/solr/latest/query-guide/local-params.html#parameter-dereferencing could that expose anything sensitive from other parameters that we're passing to solr?

ckan/lib/search/query.py
wardi: Also we can't split on " " because spaces might appear in a quoted parameter something like `q={!v='lies type= here' type=dismax}` I don't know if we can avoid a full parser :-(

amercader:
@wardi I pushed some changes to make the parsing more robust and added lots of test cases, see below 

> local parameters can be combined with parameter values by putting the value after the } https://solr.apache.org/guide/solr/latest/query-guide/local-params.html#specifying-the-parameter-value-with-the-v-key so we should split on } here instead, right?

Handled in [ab440f3](https://github.com/ckan/ckan/pull/8053/commits/ab440f36751c00da4624e8587be65eac520f0601)

> what does multiple type= parameters do? Is it safe to assume solr uses the first?

Yes, Solr ignores the following ones

> The docs suggest that local params apply to parameters other than q. If that's true we should make sure we're filtering those as well.

Done in [69a9ede](https://github.com/ckan/ckan/pull/8053/commits/69a9ede8120f158cfb2e931951bf7ad4ef456e46)

> Also it's possible to use parameter de-referencing as part of queries, https://solr.apache.org/guide/solr/latest/query-guide/local-params.html#parameter-dereferencing could that expose anything sensitive from other parameters that we're passing to solr?

I don't know. Can you think of an example? Maybe it's fair to hand the responsibility to handle this to whichever extension uses the custom queries with custom parsers

> Also we can't split on " " because spaces might appear in a quoted parameter something like q={!v='lies type= here' type=dismax} I don't know if we can avoid a full parser :-(

I confess this is the only case that the parser function [won't cover ](https://github.com/ckan/ckan/pull/8053/files#diff-30a3f7438582547c578f69519b4bed402d5b58a27f4a1d5151e15d21ad8a16d5R107).

wardi:
@amercader WDYT about rejecting any local params that contain `'`? That would prevent the fake type example above.

ckan/lib/search/query.py
```diff
@@ -89,29 +92,48 @@ def convert_legacy_parameters_to_solr(
...
+    Given a Solr parameter, extract any custom query parsers used in the
     local parameters, .e.g. q={!child ...}...
     """
     qp_type = ""
     q = q.strip()
     if not q.startswith("{!"):
         return qp_type
 
-    local_params = q[:q.index("}")]
+    local_params = q[:q.index("}") + 1]
```
wardi, line 123: in case there's a `}` inside a quoted string in the local params
```suggestion
    local_params = q[:q.rindex("}") + 1]
```

wardi: Also a possible ValueError when missing

ckan/lib/search/query.py
```diff
@@ -89,29 +92,48 @@ def convert_legacy_parameters_to_solr(
...
 
-    parts = local_params.lstrip("{!").rstrip("}").split(" ")
-    if "=" not in parts[0]:
+    parts = _parse_local_params(local_params)
```
wardi, line 125: turn ParseExceptions here and ValueErrors above into SearchQueryError?

amercader:
@wardi As you can see I ended up writing a proper parser :man_shrugging: This needs a bit more work as I need to allow more characters in keys and values but should be easy to fix and test.
If you are happy with this approach I think it's the safest going forward and for backporting.
This of course relies on pyparsing, which we already have as an indirect requirement (rq -> redis -> packaging -> pyparsing) but of course if we use it ourselves we should add it as an explicit requirement in `requirements.in` don't you think?

amercader:
All done. I've pinned pyparsing to the version we currently have in reqs.txt, it could be upgraded but let's do that in a separate PR

wardi:
This is great. Will you leave out the requirements change from the backport?

amercader:
I think so, it's unlikely that the requirements change so much in the release branches that pyparsing is not available anymore so it's safe to rely on itbbeong present

amercader:
@wardi merging this unless you have any final comments

wardi (approved):

amercader:
I reverted the backport to 2.9 because in the versions we use there pyparsing is not a requirement :(
Let's discuss what to do, for now it will only go to 2.10.4

wardi:
For the backport could we reject any local params that contain `'` or does that prevent your feature from working?

amercader:
By my feature you mean the semantic search? That should be fine because the local params look like `{!knn field=vector topK=10}` but anything like `{!bool must='some text'}` would fail (and we should also check for double quotes I guess. It sucks to have this half-backed support but I guess it's definitely better than what we have now and doesn't require us to add new requirements to 2.9

So for 2.9 basically I'll take the code in https://github.com/ckan/ckan/pull/8053/commits/ab440f36751c00da4624e8587be65eac520f0601 (just switching to pyparsing), add the quotes checks and adapt the tests

PR #8069, opened 2024-02-09
Tighten validation for id fields (followup)

### LINKED ISSUE #7604: Tighten validation for `id` fields

amercader:
Fixes #7604 

Followup to the great work by @shashigharti in #7868 

This opened several cans of worms, here's a summary of the work so far (still WIP)

| Entity              | `empty_if_not_sysadmin` | `{entity}_id_does_not_exist` | `id_validator` (Valid UUID) |
| ------------------- | ----------------------- | ---------------------------- | --------------------------- |
| Dataset             | existing                | existing                     | added                       |
| Resource            | added                   | existing                     | existing                    |
| Resource View       | added                   | added                        | added                       |
| Groups (org/groups) | added                   | added - failing :x:                         | added                       | 
| User                | added                   | added                        | added                       |
| Extras              | added                   | n/a [1]                      | added                       |
| Activity            | n/a [2]                 | n/a                          | n/a                         |
| Tag                 | n/a [3]                 | n/a                          | n/a                         |

[1] model_save uses the package_id/group_id + extra key to create / update extras rows, besides there's one
single schema for dataset and group extras so it's not easy to check for existing ids in the relevant table

[2] Activity ids can not be set by any user via the API

[3] Tag ids can not be set by any users via the API

Two things to discuss

1. Id validator. The only version of this we have right now is the `resource_id_validator` which was introduced fairly recently:
2.


amercader:
Update on the current status:

* UUIDs are now enforced in id fields (main commit is https://github.com/ckan/ckan/pull/8069/commits/c48bc7174529f0515a0c57b62a80736d026fc080, with some followups just to fix tests)
* Simplified schemas handling in `IGroupForm` (https://github.com/ckan/ckan/pull/8069/commits/8922684422707ad3f761988915f11c68d388bde1)
* User schemas just required minor tweaks, I'll leave for a future PR simplifying them and implementing proper `IUserForm` support
* Resources need a special case, because of the way resources are created and updated (several action paths that all end up in `package_update`). Package updates will use the `update_package_schema` but it is difficult (impossible?) to use separate schemas for newly added resources (i.e. a create schema) and for existing ones (i.e. an update one), which might even be present in the same input data_dict. So we can not use the pattern of using a create schema with a "id can not exist" / "empty if not sysadmin" validator and an update schema with "id must exist" validator. So I think that for resources alone, we will need to remove the `empty_if_not_sysadmin` validator. I think that the safeguards provided by `resource_id_does_not_exist` should be safe enough (This validator should actually be called `resource_id_does_not_exist_in_another_dataset`)

ckan/logic/validators.py
```diff
@@ -282,12 +319,11 @@ def resource_id_exists(value: Any, context: Context) -> Any:
...
 
-def resource_id_validator(value: Any) -> Any:
-    pattern = re.compile("[^0-9a-zA-Z _-]")
-    if pattern.search(value):
-        raise Invalid(_('Invalid characters in resource id'))
-    if len(value) < 7 or len(value) > 100:
-        raise Invalid(_('Invalid length for resource id'))
+def uuid_validator(value: Any) -> Any:
+    try:
+        uuid.UUID(value, version=4)
```
smotornyuk, line 324: UUIDs are case-insensitive. How about returning `str(uuid.UUID(value, version=4))` to normalize the value? In this way, we'll avoid creating entries with the same UUID written in different cases (we have a similar problem for user emails)

wardi: :100: this. A decade ago we imported an organization list with uppercase UUIDs and now they're mixed with new orgs created with lowercase UUIDs. So untidy.

amercader: This makes sense, but if we change the value returned we might change the case of some existing upper case UUIDs when editing objects, wouldn't that cause potential issues?

amercader: To add to this, this would only be an issue for resources and extras (both use the same schema for creates and updates for reasons explained above). The rest of entities don't use `uuid_validator` on updates. So if an old system created resource ids like `13857D04-A35B-4834-93F1-5DD98D401D44`, when updating the resource in the new CKAN version it would change to `13857d04-a35b-4834-93f1-5dd98d401d44`. Would that be an issue assuming it is mentioned in the changelog?

amercader: Confirmed, Bad Things™ happen when running this code in existing upper case resource ids:

```
ckanapi action package_create name=test_resource owner_org=test

ckanapi action resource_create package_id=test_resource id=C246F35C-7EAE-4F05-B971-99ED3A878B07

# Switch to code running with the uuid_validator version normalizing the value:

ckanapi action resource_update id=C246F35C-7EAE-4F05-B971-99ED3A878B07 description=test

ERROR [ckan.logic] Resource C246F35C-7EAE-4F05-B971-99ED3A878B07 exists but it is not found in the package it should belong to.
[...]
  File "/home/adria/dev/pyenvs/ckan-py3.9/ckan/ckan/logic/action/update.py", line 119, in resource_update                                 
    resource = _get_action('resource_show')(context, {'id': id})     
  File "/home/adria/dev/pyenvs/ckan-py3.9/ckan/ckan/logic/__init__.py", line 578, in wrapped                                              
  File "/home/adria/dev/pyenvs/ckan-py3.9/ckan/ckan/logic/action/get.py", line 1098, in resource_show                                     
    raise NotFound(_('Resource was not found.'))                     
ckan.logic.NotFound: Resource was not found.  
```

Looking at the DB, the existing upper case id resource was deleted and a new resource was created for the lower case version. There's a `resource_show` call at the end of `resource_update` which fails because it tries to search for the old upper case id (which is now deleted) in the dataset dict and fails.

So all in all, I don't think we can introduce this change with an extra database migration that lower cases all ids, and even then we might break external systems that rely on the existing upper case ids (although this might be fine if we document it)

amercader:
@wardi @smotornyuk any more comments on this?

ckan/logic/validators.py
wardi: So we're not going ahead with normalizing UUIDs for this iteration, right?

amercader: I think that if we want to normalize UUIDs we need to introduce a migration that changes all existing UUIDS to lowercase, or at least the resource ones to prevent the error I described, and clearly document it. I don't think its worth doing it given the potential for breakages. But what do you think?

wardi: Agreed, that clean up can be part of a future release. Maybe migrate our columns to the proper UUID type at the same time for better performance.

amercader:
Not sure why a ton of tests started failing after merging master :sob: I'll investigate

amercader:
@smotornyuk [these typing errors](https://github.com/ckan/ckan/actions/runs/9078272094/job/24944949228?pr=8069) are related to the same changes we discussed in https://github.com/ckan/ckan/pull/7976/files#r1597518095

But in this case the methods like `create_group_schema()` are expected to be present. Do we need to add these methods to `IGroupForm` (right now they come from the [`DefaultGroupForm`](https://github.com/ckan/ckan/blob/48cde8e5f974606862f4397a2989b3d4085787c4/ckan/lib/plugins.py#L494))? or just `type: ignore`?

amercader:
All is green again, this is ready to go

changes/8069.migration
```diff
@@ -0,0 +1,3 @@
+* Only sysadmins can now set the ``id`` field of Datasets, Groups, Organizations, Users, Resource Views and Extras
+* If provided, the value of the ``id`` field needs to be a valid UUID v4 string. Sites using custom ids that are not UUIDs can uextend the relevant schema to override the validation on the ``id`` field, but are strongly encouraged to use a separate custom field to store the custom id instead.
```
wardi, line 2:
```suggestion
* If provided, the value of the ``id`` field needs to be a valid UUID v4 string. Sites using custom ids that are not UUIDs can extend the relevant schema to override the validation on the ``id`` field, but are strongly encouraged to use a separate custom field to store the custom id instead.
```

ckan/lib/plugins.py
```diff
@@ -490,10 +491,25 @@ def bulk_process_template(self) -> str:
...
+
+    @deprecated(
+        "Use either `create_group_schema()` or `update_group_schema()`",
+        since="2.11.0"
+    )
     def form_to_db_schema_options(self,
                                   options: dict[str, Any]) -> dict[str, Any]:
-        ''' This allows us to select different schemas for different
-        purpose eg via the web interface or via the api or creation vs
+        ''' [Deprecated] This allows us to select different schemas for
```
wardi, line 511: Not only deprecated, these won't be called at all any more even as a fall-back, right?

ckan/logic/schema.py
```diff
@@ -370,6 +377,16 @@ def default_extras_schema(ignore: Validator, not_empty: Validator,
     }
 
 
+@validator_args
+def default_show_extras_schema(ignore: Validator):
+
+    schema = default_extras_schema()
+
+    schema["id"] = [ignore]
```
wardi, line 385: does this strip the ids out of the extras on show? If so I love it.

ckan/lib/plugins.py
wardi: (looks like ckanext-scheming uses the `validate` method of `IGroupForm` it won't be affected by the change)

ckan/logic/validators.py
```diff
@@ -232,6 +232,43 @@ def resource_id_does_not_exist(
...
+def resource_view_id_does_not_exist(value: str, context: Context) -> Any:
+    """Ensures that the value is not used as a ID or name.
+    """
+
+    model = context['model']
+    session = context['session']
+
+    result = session.query(model.ResourceView).get(value)
+    if result:
+        raise Invalid(_('ResourceView id already exists'))
```
wardi, line 269: Note that these don't completely protect us. If a user is sending multiple requests to create something with the same id it's possible for one to overwrite the other unless we pass a flag through to the save code to have postgres only create the record, on commit we'll reliably get an error if there was a conflict on one of the id fields. That's going to be too late for adding an error to the validation but at least we could prevent *bad things* from happening.

OTOH maybe all that's not worth the effort if the only users that can pass ids are sysadmins.

ckan/logic/schema.py
amercader: Yes! But that has been the case for a really long time https://github.com/ckan/ckan/commit/5cd0401e13bad2c2a3a279b13d141fc524191ac4

ckan/lib/plugins.py
amercader: Correct. I thought of leaving them for a version in case some plugin was using them but that's seems unlikely and maybe we can just remove them?

ckan/logic/validators.py
amercader: Do you mean a user sending multiple requests to create new objects with the same new id? If so they will get a postgres PK error even if they bypass this right? What I mean is wouldn't our current code already do that?

ckan/lib/plugins.py
wardi: The polite thing would be to have a version released with the methods deprecated and working as a fall-back then remove them next release.

Marking them deprecated but actually broken doesn't make sense. Better to remove them and raise an error when a plugin tries to use them.

ckan/logic/validators.py
wardi: For example if two resources are created at the same time with the same id it's possible that
1. both pass validation at the same time (id doesn't exist)
2. one completes `resource_dict_save` to create the resource
3. the other begins `resource_dict_save` which turns into an update of the resource from step 2 because of https://github.com/ckan/ckan/blob/d77ceb7f0a4065d13781ed4673ba9a69ab7f9e97/ckan/lib/dictization/model_save.py#L24-L37

A fix is adding `create`/`update` parameters on all our `..._dict_save` functions (instead of this upsert-like behavior embedded in them) so that we will reliably a PK error in this case.

I could make a separate issue for this because this PR as it stands is a big improvement even without fixing our dict_save functions.

amercader: Yeah let's do that in a separate PR. I'm sure we can include it in 2.11

wardi: :+1: made an issue for it #8230

ckan/lib/plugins.py
amercader: Just to make sure @wardi , you are suggesting that in the actions, if an `IGroupForm` does not have one of the new methods (eg `IGroupForm.create_group_schema()` it should try to call the old equivalent, eg `IGroupForm.form_to_db_schema()` right?

wardi: yes, with the deprecation warning and remove it the next release

amercader: @wardi see https://github.com/ckan/ckan/pull/8069/commits/bd720fd0c959e56eb237c6a787eee75d7a09ca73

ckan/logic/action/create.py
```diff
@@ -724,8 +724,18 @@ def _group_or_org_create(context: Context,
...
+            context.get("schema") or group_plugin.create_group_schema()
+        )
+    except AttributeError:
+        # TODO: remove these fallback deprecated methods in the next release
+        try:
+            schema = getattr(group_plugin, "form_to_db_schema_options")({
+                'type': 'create', 'api': 'api_version' in context,
+                'context': context})
+        except AttributeError:
+            schema = group_plugin.form_to_db_schema()
```
wardi, line 738: these will swallow AttributeErrors from inside the function calls which could be a source of very confusing bugs. Maybe some `hasattr` checks would be better?

amercader: Yes, absolutely. My only defense is that I just copied the code that was previously there :) See https://github.com/ckan/ckan/pull/8069/commits/dbbf0fa005ab88763f8f32204d542860a82f998b

wardi (approved):

PR #8077, opened 2024-02-13
Fix Password Reset Keys w/ Multiple Accounts

JVickery-TBS:
fix(user): fix reset keys multiple;

- Fix multiple user account reset keys not getting committed to DB.

### Proposed fixes:

The password reset gen method used `commit_and_remove` which prevented any other DB things from happening. So when in the loop to send multiple reset emails, only the first user object would get the reset key inserted into the DB.

This is only an issue with resetting password with an email that is used on multiple accounts. So will for sure need backporting.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8077.bugfix
+++ b/changes/8077.bugfix
@@ -1,0 +1,2 @@
+Fixed an issue when resetting password with
+an email address belonging to multiple user accounts.
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -283,7 +283,7 @@ def send_invite(
 
 def create_reset_key(user: model.User):
     user.reset_key = make_key()
-    model.repo.commit_and_remove()
+    model.repo.commit()
 
 
 def make_key():
--- a/ckan/views/user.py
+++ b/ckan/views/user.py
@@ -698,22 +698,18 @@ class RequestResetView(MethodView):
                 # user, as that would reveal the existence of accounts with
                 # this email address)
                 for user_dict in user_list:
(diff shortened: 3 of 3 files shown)


duttonw:
Isn't there a change on user registrations to restrict only 1 user per email address. Yes for legacy systems this is required this is also having conflicts on the db update see: https://github.com/ckan/ckan/commit/ced27754b48654356d5be2aa61c73f041b674714 
For our www.data.qld.gov.au instance we needed to set the unique = false to have it be successful: https://github.com/qld-gov-au/ckan/blob/qgov-master-2.10.1/ckan/migration/versions/102_ff13667243ed_create_conditinal_index_on_user.py#L24

JVickery-TBS:
@duttonw yes, CKAN 2.10 has a database constraint on the user emails. But 2.9 does not have that DB constraint, it just has some validation to prevent duplicate email accounts, but you are still allowed to have multiple same emails in the user DB table. So this is more of a fix for older versions of CKAN. (and I guess in your case it would fix your setup where you have allowed non-unique emails in the DB)

CKAN has a command `ckan db duplicate_emails` which lists all of the duplicate email user accounts. So I think the idea is that one would have to fix all of those before upgrading to 2.10

wardi:
@JVickery-TBS would you make this pr against dev-v2.9 instead of master so we can review it as part of the next patch release

JVickery-TBS:
@wardi https://github.com/ckan/ckan/pull/8079

PR #8079, opened 2024-02-15
Fix Password Reset Keys w/ Multiple Accounts (2.9 Patch)

JVickery-TBS:
fix(user): fix reset keys multiple;

- Fix multiple user account reset keys not getting committed to DB.

### Proposed fixes:

The password reset gen method used `commit_and_remove` which prevented any other DB things from happening. So when in the loop to send multiple reset emails, only the first user object would get the reset key inserted into the DB.

This is only an issue with resetting password with an email that is used on multiple accounts. So will for sure need backporting.

See: https://github.com/ckan/ckan/pull/8077 for master PR

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/mailer.py
+++ b/ckan/lib/mailer.py
@@ -204,7 +204,7 @@ def send_invite(user, group_dict=None, role=None):
 
 def create_reset_key(user):
     user.reset_key = text_type(make_key())
-    model.repo.commit_and_remove()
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
@wardi yeah multiple emails get sent like it was before, only difference is that the Reset Keys actually get put into the DB so they work.

PR #8095, opened 2024-03-01
Get the DEFAULT_FORMATS from the ini file

tino097:
This is already in master but as we are moving from Datapusher to XLoader or Datapusher Plus with most of the portals and still using 2.9 CKAN we could have it in 2.9 too

### DIFF
--- a/ckan/cli/views.py
+++ b/ckan/cli/views.py
@@ -322,7 +322,7 @@ def _add_default_filters(search_data_dict, view_types):
             u"recline_grid_view",
             u"recline_graph_view",
             u"recline_map_view",
-            u"datatablesplus_view",
+            u"datatables_view",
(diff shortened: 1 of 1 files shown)


ckan/cli/views.py
```diff
@@ -307,6 +322,7 @@ def _add_default_filters(search_data_dict, view_types):
             u"recline_grid_view",
             u"recline_graph_view",
             u"recline_map_view",
+            u"datatablesplus_view",
```
amercader, line 325: What is the `datatablesplus_view` ? Shouldn't this be `datatables_view`?

tino097: ups, i tell the secret im working on :) 
thanks i missed it while testing

PR #8098, opened 2024-03-05
[7415]Resource does not refresh the format

### LINKED ISSUE #7415: Replacing a file in an existing resource does not refresh the format

TomeCirun:
Fixes #7415 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -8,6 +8,8 @@ import logging
 import datetime
 import time
 import json
+import mimetypes
+import os
 from typing import Any, Union, TYPE_CHECKING, cast
 
 import ckan.lib.helpers as h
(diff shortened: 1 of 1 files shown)


JVickery-TBS:
@TomeCirun For this, you will want your code (or something similar to your code) in a validator and not in the `resource_action` method.

Are you able to move your code into a validator and update the resource schema?

You will probably want to change the `if_empty_guess_format` validator to something else (https://github.com/ckan/ckan/blob/master/ckan/logic/schema.py#L58)

And the validator is defined here: https://github.com/ckan/ckan/blob/master/ckan/logic/validators.py#L907

For some reference, you can see what I have done in our plugin for a validator which would always guess the format for Resources. https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/validators.py#L447 and https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/logic.py#L262

Our code is a bit more crazy and you would not need a lot of that as we have some custom logic and use the Scheming extension and limit the formats.

Thanks!

TomeCirun:
Hi @JVickery-TBS , Could you please review the changes I've made and let me know if there's anything else I should address for this issue?

Thanks!

PR #8110, opened 2024-03-11
definition list for data dictionary view

### LINKED ISSUE #8108: Custom Data Dictionary preview
## CKAN version
master

## Describe the bug
IDataDictionaryForm lets us expand on the data dictionary form and validation but not the Data Dictionary preview in the resource pages

### Steps to reproduce
Use IDataDictionaryForm to add fields to the data dictionary.

### Expected behavior
New fields should be visible in the resource page, not only the DataStore types, labels and descriptions

### Additional details

wardi:
Fixes #8108 

### Proposed fixes:
Switch data dictionary from a table display to a definition list to allow additional information to be added for each field.

Current:

data-dictionary-table

New:

data-dictionary-dl

New with additional information per field:

data-dictionary-extra

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8110.feature
+++ b/changes/8110.feature
@@ -1,0 +1,3 @@
+Use a definition list for the Data Dictionary view on resource pages to allow
+extra information for each field. Update example_idatadictionaryform plugin to
+display extra information.
--- a/ckanext/datastore/assets/css/datastore.css
+++ b/ckanext/datastore/assets/css/datastore.css
@@ -1,3 +1,7 @@
+.accordion-button.no-after::after {
+  background-image: none;
+}
+
 input#lang-curl ~ div .example-curl {
   display: none;
 }
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -2315,7 +2315,12 @@ class DatastorePostgresqlBackend(DatastoreBackend):
             info['meta']['aliases'] = aliases
 
             # get the data dictionary for the resource
-            data_dictionary = datastore_helpers.datastore_dictionary(id)
+            with engine.connect() as conn:
+                data_dictionary = _result_fields(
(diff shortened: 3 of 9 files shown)


wardi:
@smotornyuk here's the change with an accordion added that I think is an general improvement. I've added the type information to the header so it's visible even when not expanded. WDYT? (and am I using columns wrong for this? adding a `class="row"` around them breaks completely)

data-dictionary-accordion

smotornyuk:
Looks good. As for `col*` outside a `row`, strictly speaking, it's a violation. But, unless you want to write explicit styles or add a dozen extra classes to balance the position of elements, I'd leave `col` here. 

By itself, `col*` class will not work and you should not rely on it outside of `row`. But this time it's working because your `col*` elements are direct descendants of `accordion-button` element, which shares some properties of `row`. 
To put it simple, `col*` will work as direct descendant of any element with `display:flex` (`accordion-button` has this property). So whenever you are using `col*`, you can add `d-flex` class to its parent to make it work. 

It's better not to rely on this behavior, because it's specific for Bootstrap v4 and v5 an may change in future. But we are going to update styles anyway, so we can tolerate this inconvenience.


The only thing that requires changes is `-sm-` part of the class. Use just `col-*`, i.e `col-1` instead of `col-sm-1` and the same for any `-sm-` part. `-sm-` classes are applied only to screens wider than 570px, so accordions may look worse on mobile. `col-` + number, without intermediate parts, is applied to any width so you'll have approximately the same appearance on any screen.

wardi:
Thank you @smotornyuk . Here's a fix that removes those classes and the accordion functionality from fields that have no data dictionary labels/descriptions/extra information.

Here the second field has no title/description/extra so it can't be clicked and has no arrow on the right:

data-dictionary-disabled-dropdown

PR #8128, opened 2024-03-20
Support CSS URLs when Using root_path and a Reverse Proxy

JVickery-TBS:
fix(css): uris with root_path;

- Support CSS urls when using `ckan.root_path` and/or a reverse proxy.

### Proposed fixes:

The main css files use `url` values with relative paths, which results in them 404ing if someone has set up CKAN using a `ckan.root_path` along with a proxy.

I have added CSS variables in a `root` css element in the `base.html` template so we can call the `url_for_static` helper. And used the variables throughout the stylesheets.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/css/main-rtl.css
+++ b/ckan/public/base/css/main-rtl.css
@@ -11385,7 +11385,7 @@ a.tag:hover {
   padding-left: 0px;
   min-height: 205px;
   padding-top: 0.75rem;
-  background: #fbfbfb url("../../../base/images/bg.png");
+  background: #fbfbfb var(--background-image);
   border: 1px solid #ddd;
   border-width: 1px 0;
 }
@@ -13108,7 +13108,7 @@ td.diff_header {
 
 .main {
   padding: 20px 0;
-  background: #eee url("../../../base/images/bg.png");
+  background: #eee var(--background-image);
 }
 
 .main:after,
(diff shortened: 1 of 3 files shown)


wardi:
@smotornyuk you mentioned that the `cssrewrite` filter should be able to fix static urls. How does that work?

JVickery-TBS:
https://webassets.readthedocs.io/en/latest/builtin_filters.html#cssrewrite 

Not sure where in the ckan code to put something like this?

smotornyuk:
It's already been added to [main/main-rtl assets](https://github.com/ckan/ckan/blob/master/ckan/public/base/css/webassets.yml#L2)

Ideally, it must rewrite URLs automatically when the asset is assembled. I wonder if it rewrites the URL but does it in a wrong way, or is not applied at all. I'm sure this filter worked before for some files(and that's why it's added everywhere).

If possible, it's better to debug the problem with the filter and fix it, or register a custom filter that does something similar appropriately. @JVickery-TBS, your fix is good for this particular case, but we need something that will work for every extension that relies on relative paths. I'd say that auto-fixing with cssrewrite looks tidier, so let's try debugging it and finding the problem.

JVickery-TBS:
@smotornyuk from my understanding of reading https://webassets.readthedocs.io/en/latest/builtin_filters.html#cssrewrite, it rewrites based on the location of the output path. But if using CKAN's `root_path` with a reverse proxy, the `root_path` does not actually exist and webassets would not include the `ckan.root_path` in the rewrite.

The filter works fine as far as I can tell. Is it just that I have not set up `ckan.root_path` correctly? Does this directory actually have to exist on the server and all of the source files have to be in it?

PR #8146, opened 2024-04-03
Requirements april 2024

amercader:
[Some failures](https://app.circleci.com/pipelines/github/ckan/ckan/6552/workflows/10d2e733-c157-4e1c-99b0-6438972fbeb6/jobs/14544/tests), which is to be expected. A quick glance seems to suggest three areas to fix:

- [ ] Queue related tests (rq or redis?)
- [ ] Test classes properties (eg `AttributeError: 'TestPackageMemberCreateAuth' object has no attribute 'dataset'`) - Probably a pytest change
- [ ] Building the docs


requirements.txt
```diff
@@ -47,80 +47,74 @@ flask-login==0.6.3
...
-mypy==0.971
+mypy==1.9.0
     # via sqlalchemy
-mypy-extensions==0.4.3
+mypy-extensions==1.0.0
     # via mypy
 nose==1.3.7
     # via pyutilib
-packaging==21.3
-    # via redis
```
smotornyuk, line 88: This package is still required by sphinx in form `packaging>=21.0`
Any new CKAN setup(inside test containers) will get the latest packaging version - v24. But all existing instances will continue to use previously pinned v21.3(unless you install dev-requirements with eager update strategy)

v21 of packaging had LegacyVersion class(it was removed in [v22](https://packaging.pypa.io/en/stable/changelog.html#id5)) and it will cause a type error because you removed `if not isinstance(version, Version):` check from app_globals. 

Let's add `packaging>=22` or `==24` to dev-requirements to be sure that no `LegacyVersion` appears in code

amercader: Actually, we are using `packaging` in [app_globals](https://github.com/ckan/ckan/blob/a63839511e2d0cf93e3ff6743f00f694accb5f22/ckan/lib/app_globals.py#L9), [cli](https://github.com/ckan/ckan/blob/a63839511e2d0cf93e3ff6743f00f694accb5f22/ckan/cli/generate.py#L7) and the [toolkit](https://github.com/ckan/ckan/blob/a63839511e2d0cf93e3ff6743f00f694accb5f22/ckan/plugins/toolkit.py#L18) so it should be a first level requirement in requirements.in, right? Otherwise we'll rely on people installing dev-requirements.txt, which should not be necessary. I'll add it.

amercader:
Typing error unrelated, see https://github.com/ckan/ckan/pull/8196#issuecomment-2074629480

PR #8149, opened 2024-04-03
Catch error in datastore

avdata99:
Catch an error in datastore to avoid 500 error in POSTs to `datatables/ajax/<resource_view_id>`
The `datastore_search` action raises this error, it's captured later in this function but not here

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datatablesview/blueprint.py
+++ b/ckanext/datatablesview/blueprint.py
@@ -9,7 +9,7 @@ from flask import Blueprint
 
 from ckan.common import json
 from ckan.lib.helpers import decode_view_request_filters
-from ckan.plugins.toolkit import get_action, request, h
+from ckan.plugins.toolkit import get_action, request, h, ObjectNotFound
 import re
 
 datatablesview = Blueprint(u'datatablesview', __name__)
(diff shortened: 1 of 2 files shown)


ckanext/datatablesview/blueprint.py
```diff
@@ -59,13 +59,16 @@ def ajax(resource_view_id: str):
...
+    try:
+        unfiltered_response = datastore_search(
+            {}, {
+                u"resource_id": resource_view[u'resource_id'],
+                u"limit": 0,
+                u"filters": view_filters,
+            }
+        )
+    except ObjectNotFound:
+        return json.dumps({'error': 'Object not found'})
```
wardi, line 71: Is this the response datatables expects on errors? Does it render the text in a dialog box or something?

Should probably also set the response status to 404.

avdata99: @wardi This function is a response for a POST and only returns JSON with no status (even for errors).
https://github.com/ckan/ckan/blob/master/ckanext/datatablesview/blueprint.py#L141
I just add statu codes for all responses

avdata99: > Should also catch NotAuthorized here

Added here https://github.com/avdata99/ckan/commit/ff1b9d758d502fbabbe3707a4efd613bbee97a16

avdata99:
Hi @wardi. I just see this error in another CKAN instance.
Please if you have some time take a look and let me know if more changes are required here.
Sorry to bother you with this again.

PR #8152, opened 2024-04-04
Fixing activity list permissions

avdata99:
### Proposed fixes:

It looks like we are not properly handling permission for the activities lists

If we open the user activities list we were defining `{% set can_show_activity_detail = h.check_access('activity_list', {'id': id, 'include_data': True, 'object_type': object_type}) %}` at a list level (no for each activity)

So `can_show_activity_detail` means that the user can `user_update` which is only appropriate to show _user_ activities.
But if the user changed a dataset, we want to ask for `package_update` instead before showing the _Changes_ option in the list.

I think each activity require to check different permissions before showing 

The _View this version_ link points to `activity.package_history` which require the `activity_show` permission
The _Changes_ link points to `activity.package_changes` which require the `activity_diff` permission

So in a user activity (also in other activities lists) list we can see links that point to URLs that will return 403

### Features:

- [ ] includes API changes


avdata99:
Cleaning, too old

PR #8153, opened 2024-04-04
Hide activity stream if the user is not allowed

avdata99:
Fixes 403 errors for users trying to see activity stream but do not have the permission to do it

### Proposed fixes:

If a user do not have permission for user/package/org/group activity list the link must not be visible

### Features:

- [ ] includes API changes


avdata99:
Cleaning, too old

PR #8155, opened 2024-04-04
Limit Sysadmin Updating (no self modify, no system user modify)

JVickery-TBS:
feat(validator): sysadmin update;

- Added extra validator to updating a user's sysadmin value.

### Proposed fixes:

Adds an extra validator to the user schema which limits the capabilities of modifying a User's sysadmin value. Limits it so you cannot modify your own sysadmin value, or the system user's sysadmin value.

### Features:

- [ ] includes API changes


JVickery-TBS:
Wow okay, finally got all the tests working.

ckan/logic/validators.py
```diff
@@ -562,6 +562,59 @@ def ignore_not_sysadmin(key: FlattenKey, data: FlattenDataDict,
...
+def limit_sysadmin_update(key: FlattenKey, data: FlattenDataDict,
+                          errors: FlattenErrorDict, context: Context) -> Any:
+    """
+    Should not be able to modify your own sysadmin privs, or the system user's
+    """
+    value = data.get(key)
+
+    if value is None:
+        # sysadmin key not in data, return here
+        data.pop(key, None)
```
amercader, line 574: I think validators should not remove the key from the validated data unless they are of type "ignore if", where you actually want to remove it from further validation. In this case, other validators might need the key to still be there, even if empty. Same for the other places where the key is popped

ckan/logic/validators.py
```diff
@@ -562,6 +562,59 @@ def ignore_not_sysadmin(key: FlattenKey, data: FlattenDataDict,
...
+    if not contextual_user:
+        # auth_user_obj has not been set, try to get it from user context
+        contextual_user = context['model'].User.get(contextual_user_name)
+
+        if contextual_user:
+            contextual_user_name = contextual_user.name
+        elif not context.get('ignore_auth'):
+            # using a fake user and not ignoring auth, silently fail here
+            data.pop(key)
+            return
```
amercader, line 590: This whole block can probably be simplified to this right?
I don't understand the `ignore_auth` bit. This validator should be not related to authorization, unless I'm missing something

```suggestion
    if not contextual_user and contextual_user_name:
        contextual_user = context['model'].User.get(contextual_user_name)

    if not contextual_user:
        return
```

ckan/tests/logic/action/test_patch.py
```diff
@@ -234,3 +235,26 @@ def test_resource_patch_for_update(self):
         with mock.patch.dict('ckan.logic._actions', {'package_show': mock_package_show}):
             helpers.call_action('resource_patch', id=resource['id'], description='hey')
             assert mock_package_show.call_args_list[0][0][0].get('for_update') is True
+
+    def test_user_patch_sysadmin(self):
```
amercader, line 239: Can you add a test for sysadmins being able to change another user sysadmin status? Just for completeness

amercader (commented):
This looks great @JVickery-TBS , just some comments

ckan/logic/validators.py
JVickery-TBS: Ah okay, sounds good. Will update to not have any pops in this validator

ckan/logic/validators.py
JVickery-TBS: @amercader yeee it was kind of a bad workaround to this piece of code in the Reset Password view: https://github.com/ckan/ckan/blob/master/ckan/views/user.py#L809

That piece of code in the reset password view could probably be moved into a validator for the user schema??

ckan/tests/logic/action/test_patch.py
JVickery-TBS: extra assertions have been added

JVickery-TBS:
ooohh @amercader I am suddenly recalling another reason why I did the `ignore_auth` check thing. It is because the tests use a fake username `127.0.0.1` with ignore auth

So should I check the ignore_auth thing still if the provided user does not exist in the DB? Or is there a better way to do that?

amercader:
no worries @JVickery-TBS this is on my list to have a closer look. I'll figure out what's the best approach

amercader:
@JVickery-TBS Finally got around to have a look at this.
I reverted your changes to `call_action` and relaxed the checks done at the new validator level (https://github.com/ckan/ckan/commit/b65cc13ac90a0542d02dda9befc1fb28a4916247) as the presence of `context["user"]` or `context["user_obj"]` is tested
elsewhere (with `ignore_if_not_sysadmin` validator and the check for the input `id` in `user_update` respectively). All tests are [green](https://app.circleci.com/pipelines/github/ckan/ckan/7248/workflows/ddefb4ae-596e-4c81-bce9-c2a3f7bd9305/jobs/15268) now.

PR #8160, opened 2024-04-06
Add HTMX to activity stream

### LINKED ISSUE #8129: Leverage HTMX in CKAN's Activity Stream
## CKAN version
master

## Details

The current UI to display activity stream is a really good place to leverage HTMX. This will require some refactor on how the templates but both the `pagination` and the `filters` could be improved.

## Tasks
 - [ ] Use an `hx-get` attribute in our **Activity type** `select` element to apply the filters.
 - [ ] Use as well `hx-get` to move between pages instead of full page reload.

## Screenshot of the view to refactor

image

aleeexgreeen:
Fixes #8129 

### Proposed fixes:

Use HTMX in activity stream's select element and navigating between pages instead of full page reload.

### Features:

- [ ] includes API changes


pdelboca:
Hello @aleeexgreeen ! As a first approach it is good but let's be a little bit more specific on the interaction to take full advantage of HTMX.

The "problem" with this PR is that we are rendering all the full template again and then replacing the body. Re-rendering the whole template might not be ideal since it needs to call a lot of logic for things that are not going to change (followers, dashboard, login menu, breadcrumb, etc).

The pattern we want to follow here is the following:
```python
def my_view(request):
    if request.htmx:
        template_name = "partial.html"
    else:
        template_name = "complete.html"
    return render(template_name, ...)
```

Usually, while working with HTMX we look for two workflows:
 1. The first request will render the whole template
 2. The subsequent HTMX interactions will render just snippets and replace parts of the page.
 
So, what can we do to achieve this? We will need to do some work on out templates. 
1. Make `activity_stream.html` a snippet (do not extend from `organization/read_base.html`)
2. Move the subtitle block to `organization/read_base.html`
3. Leave only `{% block primary_content_inner %}` in `organization/activity_stream.html`
4. Call `{% snippet activity_stream.html, ...., ...., .... %}` from `organization/read_base.html`.
5. In the `group_activity` view, at the end, add an if statement to return `render("activity_stream.html", ...)` instead of the full template if it is an HTMX request.
6. In the template, replace the `<div class="module-content">...</div>` instead of all the `body`

Does it makes sense? Our current logic to define templates is súper messy, so some homework for you will be to mimic (or improve!) the logic behind `_get_group_template("activity_template", group_type).

aleeexgreeen:
@pdelboca Hi! Thanks for the guidance, it all made sense! I came across a couple problems/questions that I wanted to bring up to you. 

1. Moving the subtitle block to `organization/read_base.html` instead of extending it from `organization/read_base.html` causes the organization page (/organization/<organization_name>) to have the activity stream subtitle
2. The groups' and organizations' activity streams are nearly identical, except for some parameters (`object_type` and `blueprint`) being passed to the snippets. I'm thinking using `group_type` as `object_type` and adding an if statement or line in `group_activity` such as `blueprint = "activity.{}_activity".format(group_type)` for blueprint to be added to extra_vars, would work. Does that make sense and seem like a good idea?

I've pushed my progress so far, so you could take a look

aleeexgreeen:
@pdelboca Whats the correct way to use `request.htmx`? Check types keeps failing for me

pdelboca:
@aleeexgreeen hello! You are using it correctly, is there a problem with the types definition that didn't got detected when merging the HTMX PR.

@smotornyuk The problem is that our CKANRequest does not have a proper type, so adding an attribute to it will fail because Flask's request do not have it (or at least that's what I'm understanding of the error).

Do you think it is time to add our own CKANRequest type?

smotornyuk:
Good point. I just checked signatures and noticed that we can switch to CKANRequest with minimal changes to the current Flask version. 
So, if you do not backport this change to 2.10, yes, let's go with CKANRequest.

Here's the patch for satisfying type checker:
```diff
diff --git a/ckan/common.py b/ckan/common.py
index 32aaf51f2..0f04eccad 100644
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -212,7 +212,7 @@ class CKANRequest(LocalProxy[Request]):
         request.args
 
         '''
-        return cast(flask.Request, self).args
+        return self.args
 
 
 def _get_c():
@@ -327,7 +327,7 @@ local("config_declaration")
 config_declaration = local.config_declaration = Declaration()
 
 # Proxies to already thread-local safe objects
-request = cast(flask.Request, CKANRequest(_get_request))
+request = CKANRequest(_get_request)
 # Provide a `c`  alias for `g` for backwards compatibility
 g: Any = LocalProxy(_get_c)
 c = g
diff --git a/ckan/lib/captcha.py b/ckan/lib/captcha.py
index d4f0227c9..eecc852d5 100644
--- a/ckan/lib/captcha.py
+++ b/ckan/lib/captcha.py
@@ -2,11 +2,10 @@
 
 import requests
 
-from ckan.common import config
-from ckan.types import Request
+from ckan.common import config, CKANRequest
 
 
-def check_recaptcha(request: Request) -> None:
+def check_recaptcha(request: CKANRequest) -> None:
     '''Check a user's recaptcha submission is valid, and raise CaptchaError
     on failure.'''
     recaptcha_private_key = config.get('ckan.recaptcha.privatekey')
diff --git a/ckan/views/admin.py b/ckan/views/admin.py
index 3a6a96322..248a41a68 100644
--- a/ckan/views/admin.py
+++ b/ckan/views/admin.py
@@ -86,7 +86,7 @@ class ConfigView(MethodView):
 
     def post(self) -> Union[str, Response]:
         try:
-            req: dict[str, Any] = request.form.copy()
+            req = request.form.copy()
             req.update(request.files.to_dict())
             data_dict = logic.clean_dict(
                 dict_fns.unflatten(
```

aleeexgreeen:
@smotornyuk @pdelboca It worked, thank you!

@pdelboca Hi! I've made some modifications and this should be ready for your re-review, thanks!

pdelboca:
Thanks @aleeexgreeen for this PR! It looks great :)

Issue #7952, opened 2023-11-29
Python 3.12 support / Pyutillib replacement


amercader:
Changes are needed for CKAN to run on Python 3.12
*  There are a couple of requirements listed in `requirements.txt` that are incompatible (`greenlet` needs to be bumped to 3.0.1 and lxml to 4.9.3)
* Pyutilib does not work on py3.12:
```
ckan generate config
Traceback (most recent call last):
  File "/home/adria/.pyenv/versions/ckan-ml/bin/ckan", line 5, in <module>
    from ckan.cli.cli import ckan
  File "/home/adria/dev/pyenvs/ckan-ml/ckan/ckan/cli/cli.py", line 12, in <module>
    import ckan.plugins as p
  File "/home/adria/dev/pyenvs/ckan-ml/ckan/ckan/plugins/__init__.py", line 3, in <module>
    from ckan.plugins.core import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adria/dev/pyenvs/ckan-ml/ckan/ckan/plugins/core.py", line 15, in <module>
    from pyutilib.component.core import PluginGlobals, implements
  File "/home/adria/.pyenv/versions/3.12.0/envs/ckan-ml/lib/python3.12/site-packages/pyutilib/component/__init__.py", line 18, in <module>
    from . import executables
  File "/home/adria/.pyenv/versions/3.12.0/envs/ckan-ml/lib/python3.12/site-packages/pyutilib/component/executables/__init__.py", line 13, in <module>
    from pyutilib.component.executables.executable import IExternalExecutable, ExternalExecutable
  File "/home/adria/.pyenv/versions/3.12.0/envs/ckan-ml/lib/python3.12/site-packages/pyutilib/component/executables/executable.py", line 10, in <module>
    import pyutilib.misc
  File "/home/adria/.pyenv/versions/3.12.0/envs/ckan-ml/lib/python3.12/site-packages/pyutilib/misc/__init__.py", line 21, in <module>
    from pyutilib.misc.import_file import import_file, run_file
  File "/home/adria/.pyenv/versions/3.12.0/envs/ckan-ml/lib/python3.12/site-packages/pyutilib/misc/import_file.py", line 11, in <module>
    import imp
ModuleNotFoundError: No module named 'imp'

```

[...]

PR #8171, opened 2024-04-09
Adapt test for Python 3.11 support

amercader:
In Python 3.11 `urlparse` got more strict with the values it considers valid so this "URL" used in our tests is no longer correctly parsed:

'sh+eme://[net:loc]:12345/a/path?a=b&c=d'

This should help turn some of the versions green here (like CKAN 2.10 on Python 3.11):

https://github.com/ckan/ckan-python-monitor

For reference:
https://github.com/python/cpython/issues/103848


amercader:
Credit to @EricSoroos for first noticing it here: https://github.com/ckan/ckan/issues/7952#issuecomment-1835754472

amercader:
@wardi @smotornyuk I added the ipv6 test as discussed

wardi (approved):

PR #8191, opened 2024-04-18
Backport action

### LINKED ISSUE #8181: Automating backports via GitHub Actions
As part of a wider effort to reduce the burden of maintenance and increase security (see also #8143) I'd like to propose the integration of an automated process to backport merged pull request to the release branches. I believe this would have a big impact because:
* Manually cherry-picking PRs marked for backporting is time consuming and error prone
* Backporting early means less conflicts and that the release branch is a better snapshot of the next release

I've been testing different approaches in my personal fork, and want to give a full summary of my findings

Here is the full workflow definition I'm currently using in my fork:

https://github.com/amercader/ckan/blob/master/.github/workflows/backports.yml

### What to use

There are two main approaches to automating backports, bots that listen to GitHub webhooks (like [Python's](https://github.com/python/miss-islington)) or GitHub Actions. The former are generally a bit more complex and tailored to specific projects, while actions have a limited set of functionalities but are much easier to integrate.

There are [several backport actions](https://github.com/marketplace?category=&type=actions&verification=&query=sort%3Apopularity-desc+backport) available in the Marketplace but I chose [korthout/backport-action](https://github.com/korthout/backport-action) because it was well documented, well written and actively maintained.

### How it works

The setup is easy enough, just a new workflow to start a GitHub action whenever a Pull request is merged. If that PR has a _"Backport {branch}"_ label, the action will cherry-pick the commits from the merged PR:
* If they merge cleanly into the target branch, it will create a new PR against the target branch
* If they can't be merged, it will add a comment to the original merged PR with the instructions to fix it manually

### Permissions and risks

The workflow is triggered by the [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event, because the token needs write access to create the PRs and add comments. We need to be mindful of this, and never label for backporting PRs from external forks that could modify the workflows, access the action secrets etc (basically that modify the `.github` folder).
Of course the action itself could present a risk, and there is no sure way to avoid this other than reviewing its source code. I've reviewed the [source](https://github.com/korthout/backport-action/tree/main/src), which is is straightforward and well written but others can review it too.

### Customizing

I added a couple of modifications to the default behaviour of the action:

1. By default, the action runs using the `GITHUB_TOKEN` available to all actions. This token does not start new workflow runs for the events it triggers. This means that PR checks including tests won't be created in backport PRs:

   https://github.com/amercader/ckan/pull/4

   This serioulsy limits the usefulness of it, as we need to make sure that tests pass in the release branch as well.The alternative is to use a Personal Access Token (PAT), which does trigger new workflows. Note how in this case the PR was created by my user (as I used one of my own PATs) 

   https://github.com/amercader/ckan/pull/7

    To not impersonate one of our personal accounts, we can create a new `ckanbot` GH account, limit its access to the main `ckan/ckan` repo, and create a PAT with further limits the scope of the token to only what the actions needs

2. By default, if the backport fails, the action will add a comment to the original PR. I think this is easy to miss (we get a lot of Github notifications!):

    https://github.com/amercader/ckan/pull/5

   To mitigate this I added [another job](https://github.com/amercader/ckan/blob/master/.github/workflows/backports.yml#L37) to the workflow that will add a "Backport failed" label to the original PR. This way it is easier to filter all the pending backports when we are doing a release.

    https://github.com/amercader/ckan/pull/8

3. If you merge the PR and forgot to add the "Backport {branch}" label there is no way to retrigger the process. One way to deal with this is to also trigger the backport when you add a comment starting with `/backport` (or whatever we choose). This is documented in the action (see [here](https://github.com/amercader/ckan/blob/8fce75a1b452d17798aaf9ca513001c4ab1897ce/.github/workflows/backports.yml#L23)) but I think we need a way to limit the allowed users that can trigger it to ones with write access

Of course we can tweak other things like the title or description of the backport PR, whether to copy the assignees etc, but this should just be changing configuration options.

### Next steps

First of course, everybody needs to be happy with this approach. If so we then:

- [ ] Create and configure the `ckanbot` account, giving it the necessary permissions only
- [ ] Create PR with the final backports.yml workflow config
- [ ] Create new labels (eg "Backport dev-v2.10", "Backport failed")
- [ ] Start testing and tweaking it!

amercader:
Fixes #8181
As discussed in the issue, this PR includes the configuration file for an action that:

* When a PR that has a label like "Backport <branch>" is merged,   it will trigger a backport action
* If the PR commits merge cleanly into the target branch, a new PR will   be created against it, assigned to the same user as the merged one.  The usual checks will be run on the new PR
* If the commits don't merge cleanly, a comment will be posted on the orginal PR with the manual commands to fix the conflicts, and the PR will be labelled with "Backport failed"
* Additionally, Tech Team members can trigger a backport on open or already closed PRs adding a comment starting with `/backport` (and adding the relevant label)

Some more details:
* The action is run using a token from the @ckanbot account, this needs to be stored in the repository secrets (`BACKPORT_ACTION_PAT`)
* There are also two public variables (`TECH_TEAM_USER_IDS` and  `CKANBOT_USER_ID`) that need to be added to the repository variables.
* The action used source code code can be found at https://github.com/korthout/backport-action

### DIFF
--- a/.github/workflows/backports.yml
+++ b/.github/workflows/backports.yml
@@ -1,0 +1,73 @@
+name: Backport merged pull request
+on:
+  pull_request_target:
+    types: [closed]
+  issue_comment:
+    types: [created]
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+    # Only run when a pull request is merged
+    # or when a comment starting with `/backport` is created by a Tech Team member
+    if: |-
+      ${{
+        (
+          github.event_name == 'pull_request_target' &&
+          github.event.pull_request.merged
+        ) || (
+          github.event_name == 'issue_comment' &&
+          github.event.issue.pull_request &&
+          contains(fromJSON(vars.TECH_TEAM_USER_IDS), github.event.comment.user.id) &&
+          startsWith(github.event.comment.body, '/backport')
+        )
+      }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Create backport pull requests
+        uses: korthout/backport-action@v2
+        with:
+          # Token to authenticate requests to GitHub. This is a Personal Access Token
+          # from the ckanbot user
+          github_token: ${{ secrets.BACKPORT_ACTION_PAT }}
+          # Run when there is one or more "Backport <branch>" labels,
+          # excluding "Backport pending"
+          label_pattern: "Backport (?!pending)([^ ]+)$"
+          merge_commits: skip
+          copy_assignees: true
+          pull_description: |-
+            This is an automated backport pull request 🏗️.
+
+            ### Details
+            |     |     |
+            | --- | --- |
+            | Original pull request | #${pull_number} (${pull_title}) |
+            | Original author | @${pull_author} |
+            | Target branch | **${target_branch}** |
+
+            Please make sure that all relevant checks pass before merging it.
+  backport-labels:
+    name: Backport failed labels
+    runs-on: ubuntu-latest
+    if: |-
+      ${{
+        github.event_name == 'issue_comment' &&
+        github.event.issue.pull_request &&
+        github.event.comment.user.id == vars.CKANBOT_USER_ID &&
+        startsWith(github.event.comment.body, 'Backport failed for')
+      }}
+    steps:
+      - name: Add Backport failed label to PR
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.BACKPORT_ACTION_PAT }}
+          script: |
+            await github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              labels: ["Backport failed"]
+            });
--- a/changes/8191.misc
+++ b/changes/8191.misc
@@ -1,0 +1,1 @@
+Add a GitHub for automated backports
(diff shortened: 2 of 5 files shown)


Zharktas:
Is there a reason for using personal access token instead of github token in this ? PATs are somewhat annoying as they tend to expire (which is a good thing, otherwise no one would ever rotate them). But using them would add manual labor to rotate them for this action to work.

amercader:
@Zharktas we can choose any of the two types of tokens, in fact for the initial testing I used one of the "legacy" tokens. The new fine-grained tokens allow you to restrict permission to one particular repo and limit the permissions within this, rather than just impersonate an account, but yes they enforce an expiration day.
Given that the ckanbot account only has access to the ckan/ckan repo (via the ckanbot team) that might be limited enough to use legacy tokens. Personally I think it's fine to use expiring PATs if we set them to a year or so, we'll get a notification, generate a new one and update the settings. If we don't we won't break anything, the next back port action will fail, we update the token and rerun it using the `/backport` comment

wardi (approved):

amercader:
@Zharktas any objections to merge this? we can experiment with different types of tokens once this is merged. I'd like to start testing it with the new dev-v2.11 branch

amercader:
The backport actions seems to be working fine. See e.g. https://github.com/ckan/ckan/pull/8244

I had to tweak the permissions of the token to get it working and added all the settings needed in the docs: https://github.com/ckan/ckan/pull/8245

Zharktas:
Just to note, i had no objections for this, as the rotate cycle is somewhat reasonable 👍

PR #8094, opened 2024-02-29
[8002]Fix dataset respond 200 when solr down

TomeCirun:
Fixes #8002 

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #8196, opened 2024-04-23
Less details in Solr connection error message

FuhuXia:
Fixes #
Use a generic error message, not connection details in Solr error message. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/search/query.py
+++ b/ckan/lib/search/query.py
@@ -470,7 +470,7 @@ class PackageSearchQuery(SearchQuery):
                     raise SearchQueryError('Invalid "sort" parameter')
 
                 if "Failed to connect to server" in e.args[0]:
-                    raise SolrConnectionError("Connection Error", message=e.args[0])
+                    raise SolrConnectionError("Connection Error", message="Failed to connect to Solr server")
 
             raise SearchError('SOLR returned an error running query: %r Error: %r' %
                               (query, e))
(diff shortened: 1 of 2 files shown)


ckan/lib/search/query.py
```diff
@@ -470,7 +470,7 @@ def _check_query_parser(param: str, value: Any):
                     raise SearchQueryError('Invalid "sort" parameter')
 
                 if "Failed to connect to server" in e.args[0]:
-                    raise SolrConnectionError("Connection Error", message=e.args[0])
```
amercader, line 473: Can we get the details logged with a `log.warning(e.args[0])`? This might help maintainers debug the actual issue

amercader:
There's another issue with the code merged in #8094 . The `SolrConnectionError` exception class for some reasons [inherits from urllib3's `NewConntectionError`](https://github.com/ckan/ckan/blob/a63839511e2d0cf93e3ff6743f00f694accb5f22/ckan/lib/search/common.py#L33). This expects a [connection](https://github.com/urllib3/urllib3/blob/9961d14de7c920091d42d42ed76d5d479b80064d/src/urllib3/exceptions.py#L142) as a first parameter, not a string.

This is causing [typing errors](https://github.com/ckan/ckan/actions/runs/8814539881/job/24194605627?pr=8157) and will probably cause other problems when we actually raise this exception. I don't see why we need to extend urllib3 class and we can't simply do the following like the rest of the search exceptions:

```python
class SolrConnectionError(Exception):
    pass
```

@FuhuXia do you mind updating the exception class as part of this PR and check that the behaviour of #8094 is kept? Thanks

ckan/lib/search/query.py
FuhuXia: Changes were made so that 
**Site Visitors** sees `"Solr returned an error while searching."`, 
**Site Maintainers** sees `"Connection Error: Failed to connect to Solr server."`

PR #8209, opened 2024-05-02
Fix Datatables Last View

JVickery-TBS:
fix(js): datatables last view;

- Changed local storage key to suffix with view ID.

Fixes #

### Proposed fixes:

The Datatables View plugin uses a local storage key `lastView` to get the last view type (list or table) and will always use that when displaying a table. This breaks the functionality of the `responsive-flag`.

Suffixing the local storage key with the view ID will fix this issue (at least when looking at different datatable views across a site). The JS code here already suffixes some local storage keys with the view ID, so this is really just aligning with that.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8209.bugfix
+++ b/changes/8209.bugfix
@@ -1,0 +1,2 @@
+Datatables View's `lastView` local storage key is now suffixed with the view ID,
+allowing each datatable view to respect its given `responsive-flag`.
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
Whoever gets assigned this, the failure of `Check types` is not related to this PR's code.

PR #8212, opened 2024-05-03
:construction_worker: IUser Form implementation :construction_worker: 

### LINKED ISSUE #6070: Spec: IUserForm

tino097:
Fixes #6070 

This PR is still WIP and im hopping to finish it soon after the few things we would need to decide on. 

Becase it was really tricky to separate the private and public data in same column, i decided to add separate one called, for now, `plugin_data` (this was disscused some time [ago](https://github.com/ckan/ckan/issues/4147#issuecomment-1158014489))

However, I'm still figuring out an effective method to provide information which plugin is providing custom data so this solution is currently using `public` namespace. 

@ckan/core I would appreciate any suggestions or feedback.


tino097:
So i will need some suggestion about the naming of the `plugin_data` 

I considered few options as `plugin_fields`, `extra fileds`, `custom_data` but im not sure if those can be explainatory enough

ckan/model/user.py
```diff
@@ -59,7 +59,8 @@ def set_api_key() -> Optional[str]:
         Column('sysadmin', types.Boolean, default=False),
         Column('state', types.UnicodeText, default=core.State.ACTIVE),
         Column('image_url', types.UnicodeText),
-        Column('plugin_extras', MutableDict.as_mutable(JSONB))
+        Column('plugin_extras', MutableDict.as_mutable(JSONB)),
+        Column('plugin_data', MutableDict.as_mutable(JSONB))
```
wardi, line 63: IMHO we should rename plugin_extras into `plugin_data` here. There's no reason to have two places to store namespaced plugin-specific data

ckanext/example_iuserform/plugin.py
```diff
@@ -0,0 +1,69 @@
...
+            cast(
+                ValidatorFactory,
+                tk.get_validator("store_plugin_data")("example_iuserform"),
+            ),
+            tk.get_validator("ignore_missing"),
+        ]
+        return f
+
+    def show_user_schema(self) -> Dict[str, Any]:
+        schema = super(ExampleIUserFormPlugin, self).show_user_schema()
```
wardi, line 51: The schema should be passed as an argument so that multiple IUserForm plugins can add their own fields

ckan/lib/dictization/model_dictize.py
```diff
@@ -617,6 +616,10 @@ def user_dictize(
             result_dict['plugin_extras'] = copy.deepcopy(
                 plugin_extras) if plugin_extras else plugin_extras
 
+    # check and update custom user data
+    plugin_data = result_dict.pop('plugin_data', None)
+    result_dict['plugin_data'] = copy.deepcopy(
+        plugin_data) if plugin_data else plugin_data
```
wardi, line 622: if plugin_data is returned in the result_dict then all values will be available through the API without validation controlling what is entered or returned. Instead of `include_plugin_extras`, a `plugin_data` dict parameter could be passed that `user_dictize` could fill, and that dict could be included in the context of validators so that store_plugin_data and load_plugin_data would have access to them.

ckan/model/user.py
tino097: my understanding was that we would like to have separate space to store private and public data. 
so `plugin_extras` would be place to keep plugin related data and it would be available only for sysadmins 
and `plugin_data` would be stored in the user_extras table had we proceeded with that approach.

ckan/lib/dictization/model_dictize.py
tino097: sorry im not sure what you mean here, can we discuss on the next meeting or you could give me more details in the comments

tino097:
For the option to keep the data in one column, when we will have multiple extensions implementing this interface, we would endup having following structure 

```json
    "plugin_data": {
                       "example_plugin1": {
                                "private": {
                                     "key1": "value1",
                                     "key2": "value2"
                                },
                                "public": {
                                     "key1": "value1",
                                     "key2": "value2"
                                }
                       },
                        "example_plugin2": {
                                "private": {
                                     "key1": "value1",
                                     "key2": "value2"
                                },
                                "public": {
                                     "key1": "value1",
                                     "key2": "value2"
                                }
                       }
           } 
```

With this mixed public and private data per extension, i will have several questions and considerations 

   * How should validators be structured to handle both public and private data? 
   * How we will define the default schema  ?

Any feedback and suggestions are appriciated

smotornyuk:
My suggestion comes from Ian's comment [here](https://github.com/ckan/ckan/pull/8212#discussion_r1596030333)

You are now exposing the whole `plugin_data` on the model dictization stage. And, in addition, you have `load_plugin_data` which can copy some of the `plugin_data` items on the top level of the user dictionary. For example, if I have the `plugin_data: {custom: {KEY: VALUE}}` and I patch `show_user_schema` with `{KEY: [load_plugin_data('custom')]}`, I'll get the following result:

```python
user == {
  "name": "sergey",
  "KEY": "VALUE",
  "plugin_data": {"custom": {"KEY": "VALUE"}}
}
```

You see, the whole `plugin_data` is unconditionally shown and any attempt to use `load_plugin_data` leads to data duplication.


Now we go back to Ian's idea. Imagine that the whole `plugin_data` **is never shown**, i.e., it's always private. There are no `private/public` sections inside plugin data - the whole `plugin_data` attribute is hidden.

If you want to make something "public", you just add the `load_plugin_data` validator to the `show_user_schema`. As a result, a specific key from the *private* `plugin_data` gets copied on the top-level of the user dictionary and becomes *public*.

And here comes that point that was unclear to you before. If you just do as I suggested before, you'll see that `load_plugin_data` cannot pluck properties from `plugin_data`. As you are removing `plugin_data` on the dictization level to make it invisible, `data.get("plugin_data")` line from the `load_plugin_data` validator returns nothing.

To overcome this, you can move `plugin_data` into context on dictization level:

```python
def user_dictize(...):
  ...
  context['plugin_data'] = result_dict.pop('plugin_data', {})
```

And then, inside the `load_plugin_data` validator, you can access `plugin_data` from the context(instead of `data`):

```python
def load_plugin_data(key, data, error, CONTEXT):
   plugin_data = CONTEXT.get("plugin_data")
   ...
```

tino097:
@wardi  ive followed the other `extras` implementations.

PR #8226, opened 2024-05-15
snippet performance improvement

### LINKED ISSUE #6146: snippet performance issue
**CKAN version**
all

**Describe the bug**
Heavy use of template `{% snippet %}`  slows page rendering times compared to alternatives like macros or `{% include %}`

**Steps to reproduce**
Compare the rendering time of a page using snippets with one that uses macros or `{% include %}` to display the same information. The activity feed, the search results page and customized dataset forms when using ckanext-scheming are examples of pages with lots of snippets.

**Expected behavior**
Render time for `{% snippet %}` should be similar to macros and `{% include %}`

**Additional details**
`{% snippet %}` is a great tag worth saving, not just for compatibility.

Unlike `{% include %}` snippets support passing parameters, and unlike macros snippet parameters can be added over time without causing errors. Snippets are easier for developers to use than deeply extended templates with blocks selectively overridden.

The performance issue may be caused by the way jinja2 converts templates into byte code - a template passed to `render()` needs to be interpreted and can't use the byte code cache, unlike extended templates or imported macros. Our `{% snippet %}` tag calls `render()` returning the resulting string so snippets are re-interpreted every time they are used, even if it's the exact same snippet used in a loop. 

A possible fix is to have the snippet tag automatically convert snippets into a macro with the parameters passed by the caller. The macro-ified snippets could be generated with a dynamic template loader instead of creating them on the filesystem.  The snippet tag would convert itself into an `{% import %}` and a call to the macro generated.

wardi:
Fixes #6146 

### Proposed fixes:
reduce separate rendering steps for `{% snippet %}` by converting to:
```jinja2
{% with %}
    {% import ... as snippet %}
    {{ snippet(...) }}
{% endwith %}
```
and a generated template with
```jinja2
{% macro snippet(...) %}
    {% include ... %}
{% endmacro %}
```
for fewer extra stack frames and better performance.

> [!NOTE] 
> does not improve performance of `h.snippet` and other helpers that call it, like `h.user_image`, `h.follow_button`, `h.csrf_input`

```mermaid
---
config:
    xyChart:
        height: 300
---
xychart-beta horizontal
    title "/testing/primer performance (23 snippets)"
    x-axis ["old", "new"]
    y-axis "Speed (responses/s)" 0 --> 11
    bar [5.18, 6.25]
```

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -21,7 +21,6 @@ from werkzeug.exceptions import (
     Unauthorized,
     Forbidden
 )
-from werkzeug.local import LocalProxy
 
 from flask_babel import Babel
 
@@ -180,6 +179,7 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
         'h': h.helper_functions,
         'ungettext': ungettext,
         'current_user': current_user,
+        'c': g,  # backwards compat. with old Pylons templates
     })
 
     # Common handlers for all requests
@@ -188,9 +188,6 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     # the `before_request` callback
     app.before_request(ckan_before_request)
     app.after_request(ckan_after_request)
-
-    # Template context processors
-    app.context_processor(c_object)
 
     # Babel
     _ckan_i18n_dir = i18n.get_ckan_i18n_dir()
@@ -403,13 +400,6 @@ def ckan_after_request(response: Response) -> Response:
     log.info(' %s %s render time %.3f seconds' % (status_code, url, r_time))
 
     return response
-
-
-def c_object() -> dict[str, LocalProxy[Any]]:
-    u'''
-    Expose `c` as an alias of `g` in templates for backwards compatibility
-    '''
-    return dict(c=g)
 
 
 class CKAN_AppCtxGlobals(_AppCtxGlobals):  # noqa
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -182,6 +182,9 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     app.jinja_env.filters['empty_and_escape'] = \
         jinja_extensions.empty_and_escape
 
+    # Add template helpers
+    app.jinja_env.globals.update(helper_functions())
+
     # Common handlers for all requests
     #
     # flask types do not mention that it's possible to return a response from
@@ -190,7 +193,6 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     app.after_request(ckan_after_request)
 
     # Template context processors
-    app.context_processor(helper_functions)
     app.context_processor(c_object)
 
     app.context_processor(_ungettext_alias)
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -99,14 +99,6 @@ class I18nMiddleware(object):
         return self.app(environ, start_response)
 
 
-def _ungettext_alias():
-    u'''
-    Provide `ungettext` as an alias of `ngettext` for backwards
-    compatibility
-    '''
-    return dict(ungettext=ungettext)
-
-
 def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     """ This has to pass the flask app through all the same middleware that
     Pylons used """
@@ -182,8 +174,13 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     app.jinja_env.filters['empty_and_escape'] = \
         jinja_extensions.empty_and_escape
 
-    # Add template helpers
-    app.jinja_env.globals.update(helper_functions())
+    # globals work in imported and included templates (like snippets)
+    # whereas context processors do not
+    app.jinja_env.globals.update({
+        'h': h.helper_functions,
+        'ungettext': ungettext,
+        'current_user': current_user,
+    })
 
     # Common handlers for all requests
     #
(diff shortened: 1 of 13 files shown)


wardi:
It's a little concerning that `ckan/tests/lib/search/test_query.py::TestPackageQuery::test_name_multiple_results` is a flaky test :thinking:

wardi:
Another benefit is the stack traces and profile outputs of rendering templates are much cleaner. Here's the existing code with a {% snippet %} that inserts 8 stack frames and makes profiling templates almost impossible.
snakeviz-render_snippet

After this change there are 0 additional stack frames and easy profiling:

snakeviz-fastsnippet

smotornyuk:
Suggestion for simpler migration/detecting cached templates.

In addition to currently supported forms of snippet
* `{% snippet "name.html" %}`
* `{% snippet "name.html", var=value %}`

let's add another one: `{% snippet "name.html" cached %}`. This form will be completely identical to `{% snippet "name.html" %}` and does not change behavior.


In addition, let's add config option or environment variable that reports any usage of simplest form `{% snippet "name.html" %}`. 

In this way a developer can enable this switch and see all the templates, that are cached and may be outdated, just as `promotedt.html` in your PR. After analyzing the snippet, the developer will add `cached`(if it's safe to cache) or `_anything=none`(if the template relies on a variable) to the snippet tag.

Because of `cached` keyword, developer can easily filter out processed snippets. As this keyword doesn't actually change anything, it can be kept in future, or removed via regexp when migration is finished.

wardi:
@smotornyuk has concerns about the caching change and requested a new config option and syntax something like
```
{% snippet 'foo.html' cached %}
```
to indicate whether a snippet should be eligible for caching. The new config option could raise an exception if a snippet with no parameters is missing a `cached` or `uncached` indicator.

In the future we could add caching for snippets with paramers too, e.g.
```
{% snippet 'foo.html' a=name, b=value cached %}
```
where the snippet is determined completely by value of its parameters.

wardi:
@smotornyuk @amercader I've removed caching for snippets without parameters. It seems jinja2 caching ignores everything, including the language setting, making it useless on many sites and not worth the breaking change for the remaining ones.

wardi:
Just found a problem with variables in the parent templates leaking into the snippets. Working on a different approach using `macro` and `import` instead.

ckan/lib/jinja_extensions.py
```diff
@@ -261,20 +273,61 @@ def make_call_node(*kw: Any) -> Any:
...
                [, <keyword>=<value>]... %}
 
-    see lib.helpers.snippet() for more details.
+    renders <template_name> with <keyword>=<value> as local variables
     '''
 
-    tags = set(['snippet'])
+    tags = {'snippet'}
+
+    def parse(self, parser: Parser):
```
amercader, line 287: I must confess that I don't fully grasp this code. Would this be a fair summary?
Here we parse the `{% snippet 'path/to/snippet.html', arg1=test %}` string, and turn it into a node tree equivalent to:

```
{% with %}
    {% import ... as snippet %}
    {{ snippet(...) }}
{% endwith %}
```

And then this will be picked up in the `CkanFileSystemLoader.parse()` method above which will turn it into :

```
{% macro snippet(...) %}
    {% include ... %}
{% endmacro %}```

Which is what will be finally rendered by Jinja. Or is this the other way around?
Sorry I find this a bit confusing

File: ckan/config/middleware/flask_app.py [hunk 2]
```diff
@@ -190,11 +190,8 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     app.after_request(ckan_after_request)
 
     # Template context processors
-    app.context_processor(helper_functions)
     app.context_processor(c_object)
```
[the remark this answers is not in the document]
wardi: I didn't run into any failing tests, but some extensions might depend on it so I'll move it too

ckan/lib/jinja_extensions.py
wardi: That's right. The snippet `with`/`import` node tree references a specially named wrapper template `original_name§arg1,arg2.html` that gets generated by `CkanFileSystemLoader.get_source()` containing `macro`/`include` tags.

The imported wrapper template is necessary to create a separate namespace and convert `arg1`, `arg2`… into local variables for the snippet template.

wardi:
The old code has 7 stack frames between snippet tag and the actual snippet html:

```python-traceback
   File "/srv/app/src_extensions/ckan/ckan/templates/development/primer.html", line 21, in block 'secondary_content'
     {% snippet 'development/snippets/nav.html', heading='Navigation' %}
   File "/srv/app/src_extensions/ckan/ckan/lib/jinja_extensions.py", line 277, in _call
     return base.render_snippet(*args, **kwargs)
   File "/srv/app/src_extensions/ckan/ckan/lib/base.py", line 69, in render_snippet
     output = render(template_name, extra_vars=kw)
   File "/srv/app/src_extensions/ckan/ckan/lib/base.py", line 104, in render
     return flask_render_template(template_name, **extra_vars)
   File "/usr/lib/python3.10/site-packages/flask/templating.py", line 150, in render_template
     return _render(app, template, context)
   File "/usr/lib/python3.10/site-packages/flask/templating.py", line 131, in _render
     rv = template.render(context)
   File "/usr/lib/python3.10/site-packages/jinja2/environment.py", line 1304, in render
     self.environment.handle_exception()
   File "/usr/lib/python3.10/site-packages/jinja2/environment.py", line 939, in handle_exception
     raise rewrite_traceback_stack(source=source)
   File "/srv/app/src_extensions/ckan/ckan/templates/development/snippets/nav.html", line 1, in top-level template code
```

Now there are only 2 stack frames (including the generated intermediate `development/snippets/nav§heading.html` template)
```python-traceback
   File "/srv/app/src_extensions/ckan/ckan/templates/development/primer.html", line 21, in block 'secondary_content'
     {% snippet 'development/snippets/nav.html', heading='Navigation' %}
   File "/usr/lib/python3.10/site-packages/jinja2/runtime.py", line 782, in _invoke
     rv = self._func(*arguments)
   File "development/snippets/nav§heading.html", line 1, in template
   File "/srv/app/src_extensions/ckan/ckan/templates/development/snippets/nav.html", line 1, in top-level template code
```

amercader (approved):
This looks good to me @smotornyuk, but feel free to have another look

Once the last changes are done I'll merge it

PR #8228, opened 2024-05-16
User autocomplete by email

tino097:
We have option as sysadmins to search by [email](https://github.com/ckan/ckan/blob/master/ckan/model/user.py#L306) but when using the autocomplete endpoint we never sent the current user.

This should fix that and sysadmins could add the members by email. This is helpful when users are logging via sso or saml and usernames could be automaticly generated


tino097:
hey @amercader 
I reverted back to the original class method for dynamic searching and we will just pass the username to check if the user doing the search is sysadmin and as one, he could use `user_autocomplete` to search users by email

amercader:
This looks good @tino097, can you add a changelog entry to describe the new feature?

PR #8253, opened 2024-06-10
[5497]Fix cannot create draft datasets via API

### LINKED ISSUE #5497: Cannot create "draft" datasets via API
**CKAN version**
2.8

**Describe the bug**
I would like to create a draft dataset from the API, but setting `{"state": "draft"}` has no effect.

**Steps to reproduce**
```python
import time
import requests

url = "https://demo.ckan.org/api/3/action/package_create"
dataset_dict = {'title': 'A test dataset ',
                        'name': 'unique_name_{}'.format(int(time.time())),
                        'state': 'draft',  # this is the important part
                        'author': 'Test User'}
headers = {'Authorization': 'YOURAPIKEY'}
req = requests.post(url, data=dataset_dict, headers=headers)
print(req.json()["result"]["state"])
# prints "active"
```

**Expected behavior**
I would have expected that the state is `"draft"`. This is possible via the web interface (by aborting the dataset creation process during the resource upload step). But it is apparently not possible via the API.

If there is a workaround, I would also be interested in that.

TomeCirun:
Fixes #5497 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/schema.py
+++ b/ckan/logic/schema.py
@@ -132,11 +132,12 @@ def default_create_package_schema(
         package_name_validator: Validator, strip_value: Validator,
         if_empty_same_as: ValidatorFactory,
         email_validator: Validator, package_version_validator: Validator,
-        ignore_not_package_admin: Validator, boolean_validator: Validator,
+        boolean_validator: Validator,
         datasets_with_no_organization_cannot_be_private: Validator,
         empty: Validator, tag_string_convert: Validator,
         owner_org_validator: Validator, json_object: Validator,
         ignore_not_sysadmin: Validator, uuid_validator: Validator) -> Schema:
+    
     return {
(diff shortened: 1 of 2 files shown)


wardi:
There are a few things going on here.
- `ignore_not_package_admin` seems to be broken: it was designed to defer to `package_change_state`
- `package_change_state` defaults to the same checks as `package_update`
- it's currently only allowing sysadmins to set the value for some reason

Since this is the only user of these functions (and because they're not working) let's remove them at the same time.

Also the `state` field is very important to datasets so we shouldn't allow it to be set to any value.

Let's add a new validator to make sure the state set is valid (and to allow sites to override this behaviour including limiting who can set the sate if they choose)

@TomeCirun how does that sound?

wardi:
We should also remove the `allow_state_change` context variable code, since this won't work without it

PR #8261, opened 2024-06-13
Group update fails when using name instead of id

smotornyuk:
Group update schema contains `group_id_exists` validator for `id` field. When using `id=name` during an update, this causes a validation error.

Using `group_id_or_name_exists` instead solves the problem


wardi:
@smotornyuk test failure looks like it might be related to the change. Tests need an update?

smotornyuk:
Yes, you are right. It was possible to update group and set an empty name, because `update_group_schema` has `not_missing` to filter `None` and validator for uniqueness, which don't complain because there are no groups with empty name :)

PR #7672, opened 2023-06-23
Remove template_head_end and template_footer_end configs

pdelboca:
Fixes some of the points in #7671 

### Proposed fixes:
Removes `template_head_end` and `template_footer_end` configuration options.

PR #8264, opened 2024-06-13
add missing head_extras block

snokamedia:
added missing head_extras block used in default templates ckan/templates/package/read_base.html and ckan/templates/package/resource_read.html

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/base.html
+++ b/ckan/templates/base.html
@@ -83,6 +83,10 @@
       </style>
       {%- endif %}
     {% endblock %}
+
+    {%- block head_extras %}
+    {% endblock %}
+
   </head>
 
   {# Allows custom attributes to be added to the <body> tag #}


amercader:
Thanks @snokamedia this makes sense. Can you restore the `body_extras` block as well? It is still used in at least [one extension](https://github.com/IATI/ckanext-iati/blob/bdd4694e8ee8c0ed4207a2e635873e2ba7ac3f02/ckanext/iati/theme/templates/package/read_base.html#L92) out there. It should go [here](https://github.com/ckan/ckan/pull/7672/files#diff-af6f9df2b3503fc59a8747db5a453ebd3dc140172c9ad746ebfe9e9712795702L118).

These blocks were removed in #7672 . The removal of the config options made sense but the blocks should have been left there as they were used by other templates.

Issue #8275, opened 2024-06-18
No organization or group listed in user organization/group tab


blagojabozinovski:
## CKAN version
CKAN 2.11.0a0
## Describe the bug
No organization or group is listed where the logged in user is member in user organization tab or user group tab
![Screenshot from 2024-06-18 13-53-56](https://github.com/ckan/ckan/assets/62408051/a0ad1233-52ab-49e7-a6c8-8a6eb0bf2358)
![Screenshot from 2024-06-18 13-54-04](https://github.com/ckan/ckan/assets/62408051/e5f5ff41-6520-40c0-8caa-93fe9ac3113e)

### Steps to reproduce
Go to the following url 
https://<site_url>/user/<username>/organizations
https://<site_url>/user/<username>/group

### Expected behavior
Show list of groups and organizations where the logged in user is member
Also not clear why the helper h.default_group_type returns group_type = Group for Organization tab on the second printscreen

PR #8277, opened 2024-06-18
Set orgs_available in user read_base template

blagojabozinovski:
Fixes #
Fixes the no organization or group listed in user organization/group tab [issue](https://github.com/ckan/ckan/issues/8275)
### Proposed fixes:
Proposed fix for the following [issue](https://github.com/ckan/ckan/issues/8275)
The orgs_available variable is not passed to read_organizations.html or read group.html template 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/user/read_base.html
+++ b/ckan/templates/user/read_base.html
@@ -1,6 +1,16 @@
 {% extends "page.html" %}
 
 {% set user = user_dict %}
+{% set orgs_available = h.organizations_available(permission='manage_group',
+  include_dataset_count=True,
+  include_member_count=True,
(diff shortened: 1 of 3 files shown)


wardi:
This would slow down almost every page on the site, please make the change in just the templates or views that need it.

wardi (changes requested):
This would slow down almost every page on the site, please make the change in just the templates or views that need it.

blagojabozinovski:
> This would slow down almost every page on the site, please make the change in just the templates or views that need it.

@wardi Good point. Can You please check if it's OK now? I think these are the only templates that need it.

PR #8293, opened 2024-06-21
IMailer Implement

JVickery-TBS:
Framework for the `mailer.py` script.

Branched off of (and requires): https://github.com/ckan/ckan/pull/8200 

### Features:

- [ ] includes API changes


smotornyuk:
And as extra option, you can even extract core mailing logic into separate plugin, like activities and explain in documentation, that this plugin must be enabled in order to send mails. In this way you won't have a special case for the native mailer logic and can treat everything as implementation of INotifier

smotornyuk:
Sorry for another comment. I just thought, that ability to override `get_invite_body`/`send_invite_body` may be quite useful. Yes, you still can intercept any message with notify_recipient, but what if you don't want to spend extra time on rendering the template, especially if you are not going this rendered `body`?

But instead of registering a lot of methods(treating them in special way, while any other custom type of notification from plugin wouldn't have such hooks), yfor ou can add second method to interface: `notify_about_topic(TOPIC, details)`. Then you rewrite the couple of places in core, like `mailer.send_reset_link(user_obj)` to `notify_about_topic('password_reset_link', {user: user})` and any plugin can handle all types of notifications using this generic extension point instead of defining 4-5 different methods. Even more. Other plugins can also call `notify_about_topic("object_uploaded_to_s3", details)` and your interface will handle these custom notification in the same way it handles native notification of CKAN

JVickery-TBS:
@smotornyuk awesome, thanks for all of the feedback! So basically I should be adding the `notify_about_topic` into the `INotifier` (https://github.com/ckan/ckan/pull/8200) implement class?

Yeah the idea around this was to be able to override the mail methods prior to when `mail_recipient` gets called throughout the different methods.

So would the code in core change from something like: `mailer.send_reset_link(user_obj)` to:
```
for plugin in plugins.PluginImplements('INotifier'):
    notify_about_topic('password_reset_link', {'user': user})
else:
    mailer.send_reset_link(user_obj)
```
And then there would be no code changes to `mailer.send_reset_link`?

smotornyuk:
I think, that's the optimal solution. If you register catch-all notifier, it just prevents `else` route and handles all the notifications, as long as you don't have other implementations.

If you don't have any implementation, CKAN's default logic is used.

If you have multiple notification plugins, let's say `sms` and `fax`, they can even check if they need to handle notification via config

```ini
ckan.notification.topic.password_reset_link.handler = sms
ckan.notification.topic.invite_user.handler = fax

## or ##
ckanext.sms_notifier.handle_reset_link = true
ckanext.fax_notifier.handle_invite = true
```

Finally, if you have custom notifier and want native CKAN's mailer to handle only invite mail, it's easy to write small plugin that wraps CKAN's mailer and do the config trick mentioned above.


I think, it covers all the options. And yes, in this way we can avoid modifying existing functions.

JVickery-TBS:
@smotornyuk is this something that will need to support backward compatibility? for other plugins that directly call `mailer.mail_recipient`? or does that not matter?

JVickery-TBS:
@smotornyuk okay got something going now in here which I think covers what you talked about above.

https://github.com/open-data/ckanext-gcnotify/blob/feature/imailer/ckanext/gcnotify/plugin.py <- I am testing out the changes in a branch of mu GC Notify extension and it is working nicely.

Just gunna clean up lint now, and I will add/fix up the automated tests once we have actually landed on everything here.

smotornyuk:
`mail_user` is exposed via toolkit, so it must be the same as before. `mail_recipient` is not documented in CKAN docs, so it can be changed. But if you keep it the same as before, so that nobody has to rewrite code during update, it would be perfect

JVickery-TBS:
@smotornyuk @wardi okay the `mail_user` method should be the same, no changes there. I have updated the test coverage and added in new ones for `iNotifier.notify_about_topic`

wardi:
Sorry for leaving this so long, @JVickery-TBS would you resolve the conflicts so we can include this in the next release?

JVickery-TBS:
@wardi conflicts resolved. But getting the `post` method of `ckan/views/user.py:649:9` is too complex :/

How would I solve that? I think it is because of the new loop:
```
for plugin in plugins.PluginImplementations(plugins.INotifier):
```

JVickery-TBS:
@wardi okay I put the new code piece into a new class method in `RequestResetView` which solved complexity issue. There is just this codecov issues now

JVickery-TBS:
The codecov/patch might be a false negative actually, idk if it is reading the code from the `ckanext/example_inotifier` or not

PR #8294, opened 2024-06-22
[6150]Better accesibility for text rendered as title attributes

### LINKED ISSUE #6150: Better accesibility for text rendered as `title` attributes
Across our templates we use the `title` attribute to provide additional information when hovering with the mouse, eg:

https://github.com/ckan/ckan/blob/0ceebd17bcbd435b156e776df20593ed41af3ffd/ckan/templates/header.html#L10

https://github.com/ckan/ckan/blob/0ceebd17bcbd435b156e776df20593ed41af3ffd/ckan/templates/user/read_base.html#L81

This [is strongly discouraged](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns) as it prevents a whole subset of users to access this information. We should incorporate more accessible ways of conveying this information. That might involve different approaches depending on the template and usage, including adding new markup elements for the extra information, use of `aria` attributes and elements, etc

This article provides a good starting point:

https://inclusive-components.design/tooltips-toggletips/

Any guidance and additional resources on this front would be greatly appreciated

TomeCirun:
Fixes #6150 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/javascript/main.js
+++ b/ckan/public/base/javascript/main.js
@@ -101,3 +101,16 @@ $(function() {
     $("body").removeClass("filters-modal");
   });
 });
+
+// Initialize Bootstrap 5 tooltips for elements with data-bs-toggle="tooltip", 
+// using aria-label for tooltip content.
+$('[data-bs-toggle="tooltip"]').each(function () {
+  var tooltipTriggerEl = this;
+  var ariaLabel = $(tooltipTriggerEl).attr('aria-label');
+
+  new bootstrap.Tooltip(tooltipTriggerEl, {
+    title: ariaLabel,
+    placement: "bottom",
+    delay: { show: 400, hide: 200 }
+  });
+});
--- a/ckan/templates/snippets/context/group.html
+++ b/ckan/templates/snippets/context/group.html
@@ -3,7 +3,7 @@
 <div class="context-info">
   {% if description != 'null' %}
     {% set description = description|striptags %}
-    <p class="about" title="{{ description }}">
+    <p class="about" aria-label="{{ description }}" data-bs-toggle="tooltip">
       {{ description|truncate(160) }}
     </p>
   {% endif %}
--- a/ckan/templates/snippets/search_form.html
+++ b/ckan/templates/snippets/search_form.html
@@ -68,7 +68,7 @@
               {%- else -%}
                 {{ h.list_dict_filter(search_facets_items, 'name', 'display_name', value) }}
               {%- endif %}
-              <a href="{{ facets.remove_field(field, value) }}" class="remove" title="{{ _('Remove') }}"><i class="fa fa-times"></i></a>
+              <a href="{{ facets.remove_field(field, value) }}" class="remove" aria-label="{{ _('Remove') }}" data-bs-toggle="tooltip"><i class="fa fa-times"></i></a>
             </span>
           {% endfor %}
         {% endfor %}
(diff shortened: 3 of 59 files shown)


smotornyuk:
This solution would break(and initialize twice) any other tooltip that does not rely on aria-attributes.

smotornyuk:
And it worth reviewing the role of title attribute. For example, I doubt [this link](ckan/templates/group/snippets/feeds.html) really needs tooltip or the `title` attribute

ckan/templates/snippets/context/group.html
```diff
@@ -3,7 +3,7 @@
 <div class="context-info">
   {% if description != 'null' %}
     {% set description = description|striptags %}
-    <p class="about" title="{{ description }}">
+    <p class="about" aria-label="Group description">
```
smotornyuk, line 6: Don't use arial-label on static text. This attribute is designed for links/buttons/toggles.

ckan/templates/user/snippets/api_token_list.html
```diff
@@ -38,7 +38,7 @@
               <form action="{{ action }}" method="POST">
                 {{ h.csrf_input() }}
                 <div class="btn-group">
-                  <button type="submit" href="{{ action }}" class="btn btn-danger btn-sm" title="{{ _('Revoke') }}" data-module="confirm-action">
+                  <button type="submit" href="{{ action }}" class="btn btn-danger btn-sm" aria-label="{{ _('Revoke ApiToken') }}" data-bs-title="{{ _('Revoke ApiToken') }}" data-bs-toggle="tooltip" data-module="confirm-action">
```
smotornyuk, line 41: The term `ApiToken` is used in code. In normal language, you should not use random case swaps **inSiDe the WoRd**. 
Check the heading of the table: we are using `token` in UI when referring to this thing.

ckanext/activity/templates/package/snippets/resources.html
```diff
@@ -2,11 +2,12 @@
 
 
 {% block resources_list %}
+  {{ is_activity_archive}}
```
smotornyuk, line 5: Do you really need this variable printed to the page?

ckan/templates/snippets/search_form.html
```diff
@@ -68,7 +68,7 @@ <h1>
               {%- else -%}
                 {{ h.list_dict_filter(search_facets_items, 'name', 'display_name', value) }}
               {%- endif %}
-              <a href="{{ facets.remove_field(field, value) }}" class="remove" title="{{ _('Remove') }}"><i class="fa fa-times"></i></a>
+              <a href="{{ facets.remove_field(field, value) }}" class="remove" aria-label="{{ _('Remove applied filter: {value}').format(value=value) }}" data-bs-title="{{ _('Remove applied filter: {value}').format(value=value) }}" data-bs-toggle="tooltip"><i class="fa fa-times"></i></a>
```
smotornyuk, line 71: I recommend leaving only `Remove` in inside data-bs-title. Aria-label is good, but tooltip for user that can see the item is too verbose

ckan/public/base/javascript/main.js
```diff
@@ -101,3 +101,8 @@ $(function() {
     $("body").removeClass("filters-modal");
   });
 });
+
+// Initialize tooltips using Bootstrap
+$('[data-bs-toggle="tooltip"]').each(function (index, element) {
```
smotornyuk, line 106: Do not execute code immediately, wait till the document is ready, like in the snippet above.

smotornyuk (changes requested):

PR #8295, opened 2024-06-24
2024-06. Update pyright and types

smotornyuk:
Update pyright to the latest version


ckan/types/model.py
```diff
@@ -1,18 +1,19 @@
...
+from sqlalchemy.orm.query import Query as _Query
 from sqlalchemy import Table
 
 if TYPE_CHECKING:
     import ckan.model as _model  # noqa
 
 
 AlchemySession = ScopedSession
-Query = Query
+Query = _Query
```
wardi, line 16: why?

smotornyuk: Current limitation of pyright. When pyright analyzes `Query`, it sees that this variable has value `_Query, tracks it back to the sqlalchemy source, correctly producing the type.

In the previous form, when pyright saw `Query`, it looked at the value of this variable and saw `Query`. I.e., a variable that has a value of itself. So pyright assumed that it has type `Any`, without any attempt to track the source of the original variable.

In previous versions of pyright, the parsing order was different, that's why we used successfully `Query = Query` before. But the old parsing order had problematic edge cases, so it was changed to a more restrictive, but safer order.

Alternatively, we can import value from sqlalchemy and remove `Query = Query`. But it will cause an "unused import" error instead, that's why the assignment exists in the first place.

smotornyuk:
@wardi , I found a nicer way to avoid the problem. Instead of creating the alias, I added `__all__` to re-export `Query` without assigning it to module level variable.
```python
__all__ = [
    "Model", "AlchemySession", "Query",
]
```

PR #8299, opened 2024-06-25
[8202]Change timestamp format for 'timestamp' field type

### LINKED ISSUE #8202: datastore_search truncates timestamps to seconds
## CKAN version
master

## Describe the bug
timestamps stored in postgres can store microsecond values but `datastore_search` truncates to seconds
https://github.com/ckan/ckan/blob/ce6f837e4dbbc8aaa1ad5002e13ccf79d354061b/ckanext/datastore/backend/postgres.py#L2078

### Steps to reproduce
datastore_create or load data with smaller than 1s precision

### Expected behavior
Data stored (at least up to postgres accuracy) should be returned by datastore_search

### Additional details
We use a trigger to record created/modified timestamps on rows and not returning full accuracy makes our tests fragile

PavloMalko:
Ad milliseconds into timestamp field

Fixes #8202 

### Features: add ms value into filed with 'timestamp' type.

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -2078,7 +2078,7 @@ class DatastorePostgresqlBackend(DatastoreBackend):
             if typ == 'nested':
                 fmt = "coalesce(({0}).json,'null')"
             elif typ == 'timestamp':
-                fmt = "to_char({0}, 'YYYY-MM-DD\"T\"HH24:MI:SS')"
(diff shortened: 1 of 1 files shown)


wardi:
notes from meeting:
- add a new configuration option for datastore search that defaults to returning MS in timestamps
- add note to changelog about config option for sites that want to maintain the old behaviour
- don't make format configurable, APIs should return dates in ISO format

Thank you!

Issue #8139, opened 2024-04-01
`datetime.utcnow` deprecation


smotornyuk:
Using `datetime.utcnow` [is not recommended](https://docs.python.org/3.9/library/datetime.html#datetime.datetime.utcnow) because different libraries(and `datetime` itself) often use dates without timezone as local time. And since python v3.12 `.utcnow` is deprecated.

<details><summary>There are 47 explicit occurrences of `.utcnow` in code</summary>

```
ckan/lib/dictization/model_save.py:54:33: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/lib/uploader.py:159:37: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/lib/uploader.py:266:41: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/logic/action/create.py:702:29: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/logic/action/update.py:310:31: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/api_token.py:72:28: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/dashboard.py:30:44: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/dashboard.py:31:32: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/follower.py:34:25: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/user.py:34:17: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/user.py:349:36: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
ckan/model/user.py:352:32: DTZ003 The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
[...]

PR #8305, opened 2024-06-28
Add timezone support to date helpers

smotornyuk:
Because of naive `datetime` deprecation(#8139), we'll face more dates with timezone in the future.

`date_str_to_datetime` helper does not support timezones. But this can be easily fixed, as since python 3.7, `datetime` has `fromisoformat` helper that parses some of ISO 8601 representations, including the one, with timezone specification.

I left the original implementation unchanged, so all previously parsed values are still supported. 

As a bonus, parsing any ISO 8601 value(`YYYY-MM-DD HH:MM:SS` with or without timezone) is 20 times faster than before.

### DIFF
--- a/changes/8305.feature
+++ b/changes/8305.feature
@@ -1,0 +1,1 @@
+`date_str_to_datetime` helper accept values with timezone information.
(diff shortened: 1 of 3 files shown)


amercader:
@wardi you need to add the `Backport <branch>` labels before commenting otherwise the action doesn't know where to backport

PR #8322, opened 2024-07-04
[8189]clean activity history command/action

### LINKED ISSUE #8189: clean activity history command/action
## CKAN version
all

## Describe the bug
Activity history grows without bounds. There should be a way to remove old activity history, either a CLI something like https://github.com/open-data/ckanext-canada/blob/c2808d9c318e19d5f0c3b4c2f53073e9576d6b8f/ckanext/canada/commands.py#L740-L756 and/or a similar action.

### Steps to reproduce
Run a ckan site for a long time.

### Expected behavior
Database not expand forever

### Additional details

TomeCirun:
Fixes #8189 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/activity/logic/action.py
+++ b/ckanext/activity/logic/action.py
@@ -6,6 +6,7 @@ import logging
 import datetime
 import json
 from typing import Any, Optional
+from dateutil import tz
 
 import ckan.plugins.toolkit as tk
 from ckan import authz
@@ -20,12 +21,11 @@ from ..model import activity as model_activity, activity_dict_save
 log = logging.getLogger(__name__)
 
 
-def _get_user_permission_labels(
-    context: Context
-):
-    if not authz.is_sysadmin(context.get('user')):
+def _get_user_permission_labels(context: Context):
+    if not authz.is_sysadmin(context.get("user")):
         return get_permission_labels().get_user_dataset_labels(
-            context['auth_user_obj'])
+            context["auth_user_obj"]
+        )
     else:
         return None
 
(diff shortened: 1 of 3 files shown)


ckan/cli/clean.py
```diff
@@ -92,3 +93,57 @@ def users(force: bool):
...
+            {"ignore_auth": True}, {}
+        )
+        context: Context = {"user": site_user["name"]}
+        data_dict: dict[str, Any] = {
+            "start_date": start_date,
+            "end_date": end_date,
+            "offset_days": offset_days,
+        }
+
+        result = logic.get_action("activity_delete_by_date_range_or_offset")(
```
JVickery-TBS, line 144: @TomeCirun sorry, another quick comment here. Just went to see what I had in our plugin to do something similar. I did use a `--quiet` flag in my click command. And paired that with a `click.confirm` which would prompt the user in the CLI to do the y/n answer if they want to delete x amount of activities.

Would be nice to have as Activities are not stately in the database, so they do not have `active`, `draft`, `deleted`. So once this new `activity_delete` action runs, they are gone for good. So would be good to prompt the user to confirm this action.

JVickery-TBS:
@TomeCirun if your up for it, it would be cool to add Activities as a data type to be purged in the `templates/admin/trash.html` view.

A quick look at the `views.admin.TrashView`, it does not look like it would be easy to extend in the `ckanext/activity` so might be best to do the `plugin_loaded("activity")` and add it into Core there. @wardi could you confirm??

wardi:
We should adjust `templates/admin/trash.html` so it can be extended by any plugin cleanly, then add the required template and blueprints to `ckanext/activities`

JVickery-TBS:
@wardi should the new `clean activities` sub command here be inside the ckanext/activity plugin? Could import `from ckan.cli.clean import clean` and use the decorator inside of the `activity` plugin to add a sub command.

wardi:
@JVickery-TBS sounds like a good suggestion

PR #8330, opened 2024-07-08
Requirements upgrade July 2024

amercader:


amercader:
These are all fairly uncontroversial, and include minor security upgrades

PR #8342, opened 2024-07-12
Datastore Field Name Validation and Control Characters

JVickery-TBS:
feat(logic): ds field name validator;

- Added Datastore field name validator to ds create sub-schema.
- Added catch to conrtol characters in ds field names.

Fixes #

### Proposed fixes:

Adds a validator to the datastore create sub-schema for field IDs. And also prevents control characters from being used in datastore field names.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/helpers.py
+++ b/ckanext/datastore/helpers.py
@@ -14,6 +14,8 @@ import ckan.common as converters
 import ckan.plugins.toolkit as tk
 from ckan.types import Context
 
+import unicodedata
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


JVickery-TBS:
@wardi looks like the new control character thing would prevent escaping things like single quotes:
```
ckan.logic.ValidationError: None - {'fields': [{'id': ['"a;"\' x" is not a valid field name']}]}
```

wardi:
It's not the control character check causing the test failure, it's the check for `"`, which isn't valid in postgres. I think the test should be updated to remove the double quote from the column name. Maybe it only worked because it was testing with sqlite? There's no reason to run tests with sqlite anymore so that could be changed too.

wardi (changes requested):
If we remove the double quote from the failing test this should be good to go.

JVickery-TBS:
@wardi `TestSearchIndex.test_search_index_rebuild` is failing not too sure why, could not manage to trace it

changes/8342.feature
```diff
@@ -0,0 +1 @@
+Datastore field names/IDs are now validated at the scheming level. Control characters are no longer allowed in Datastore field names/IDs.
```
wardi:
```suggestion
Datastore field names/IDs are now validated at the schema level. Control characters are no longer allowed in Datastore field names/IDs.
```

PR #8349, opened 2024-07-18
[dev-v2.10] Security upgrades July 2024

amercader:
As discussed in previous meetings.

This bumps the minimum Python version for CKAN 2.10 to Python 3.8


.github/workflows/pyright.yml
```diff
@@ -11,16 +11,18 @@ jobs:
...
         with:
           node-version: ${{ env.NODE_VERSION }}
       - name: Install python deps
         run: pip install -r requirements.txt -r dev-requirements.txt -e.
       - name: Install node deps
         run: npm ci
       - name: Check types
-        run: npx pyright
+        run: |
+          npm list pyright
```
wardi, line 27: so we get the pyright version in the run log? makes sense

PR #8354, opened 2024-07-18
`ckan.download_proxy` support for the `resourceproxy` plugin

### LINKED ISSUE #7422: Download_proxy documentation appears to be lost
## CKAN version
2.10, master

## Describe the bug
Documentation from https://github.com/ckan/ckan/pull/6517 seems to be missing.

https://docs.ckan.org/en/2.10/maintaining/configuration.html#resource-proxy-settings

amercader:
* Added back the documentation for `ckan.download_proxy` , probably lost during the switch to config declaration. Fixes #7422 
* Added support for `ckan.download_proxy` to `resourceproxy` following the same pattern used in [xloader](https://github.com/ckan/ckanext-xloader/pull/127), [archiver](https://github.com/ckan/ckanext-archiver/pull/78), etc

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1038,10 +1038,15 @@ groups:
               they're given in the config file, regardless of which Python modules
               they're implemented in.
 
-      - key: ckan.resource_proxy.timeout
-        type: int
-        default: 5
-        description: Timeout in seconds to use on Resource Proxy requests.
+      - key: ckan.download_proxy
+        default: None
+        example: http://proxy:3128
+        description: |
+          Specifies a HTTP proxy to be used by extensions such as XLoader or Archiver
+          when downloading remote files. This may be useful for enabling access to
+          restricted network locations, or restricting access to privileged ones, eg
+          preventing `Server Side Request Forgery <https://feeding.cloud.geek.nz/posts/restricting-outgoing-webapp-requests-using-squid-proxy/>`_.
+          It will not be used by CKAN core.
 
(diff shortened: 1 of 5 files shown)


ckan/config/config_declaration.yaml
```diff
@@ -1038,10 +1038,15 @@ groups:
               they're given in the config file, regardless of which Python modules
               they're implemented in.
 
-      - key: ckan.resource_proxy.timeout
```
EricSoroos, line 1041: Why is this removing the timeout?

ckan/config/config_declaration.yaml
```diff
@@ -1038,10 +1038,15 @@ groups:
...
-        type: int
-        default: 5
-        description: Timeout in seconds to use on Resource Proxy requests.
+      - key: ckan.download_proxy
+        default: None
+        example: http://proxy:3128
+        description: |
+          Specifies a HTTP proxy to be used by extensions such as XLoader or Archiver
```
wardi, line 1045:
```suggestion
          Specifies a HTTP proxy to be used by extensions such as Resource Proxy, XLoader or Archiver
```

wardi:
It's worth adding a note in https://docs.ckan.org/en/2.10/maintaining/data-viewer.html#resource-proxy that users should install and configure a proxy to prevent internal network resources from being exposed with the Resource Proxy

amercader:
good point @wardi , done in [7f1a978](https://github.com/ckan/ckan/pull/8354/commits/7f1a9780d28db28e8777a29b2953eb739164b03e)

Issue #7220, opened 2022-11-18
setup.cfg (find) module autodiscover does not include all modules


spwoodcock:
**CKAN version**
master / dev-v2.10

**Describe the bug**
- Configuration for packaging was moved from setup.py to setup.cfg between 2.9 --> 2.10.
- The new configuration does not include all modules under the `ckan` directory during install.
- After debugging I found it was due to the `options.packages.find.include` configuration setting: https://github.com/ckan/ckan/blob/master/setup.cfg#L36

**Steps to reproduce**
1. Install either the master or dev-v2.10 branch:
`pip install git+https://github.com/ckan/ckan`
`pip install git+https://github.com/ckan/ckan@dev-v2.10`

2. Attempt to run CKAN:
`ckan -c config.ini run --host 0.0.0.0`

3. Get an error that module ckan.cli.cli does not exist.

**Expected behavior**
- CKAN has all required modules installed and the application server starts.

**Solution**
- Removing the setting solves the issue - I will create a PR.
  - This could be a bug in setuptools, or the config is being used incorrectly.
  - I can't see any implications of removing the config option.
  - I think it would be better to use the `exclude` option for anything that specifically requires excluding (default: all modules are auto-discovered).

pdelboca:
I'll add some extra information. Currently, using `pip install .` or `pip install -e .` will differ in the modules it install. Here is the diff:
```
 ~/Desktop$ diff editable.txt non-editable.txt 
15d14
< Found submodule activity (is a package: True)
21d19
< Found submodule example_blanket_implementation (is a package: True)
25d22
< Found submodule example_humanizer (is a package: True)
36d32
< Found submodule example_isignal (is a package: True)
47a44
> Found submodule test_tag_vocab_plugin (is a package: False)
```

This is a problem, with executing `pip install .` does not add `activity` plugin?


For `pip install .`:
```
Found submodule authz (is a package: False)
Found submodule cli (is a package: True)
Found submodule common (is a package: False)
Found submodule config (is a package: True)
Found submodule exceptions (is a package: False)
Found submodule i18n (is a package: True)
Found submodule lib (is a package: True)
Found submodule logic (is a package: True)
Found submodule migration (is a package: True)
Found submodule model (is a package: True)
Found submodule plugins (is a package: True)
Found submodule tests (is a package: True)
Found submodule types (is a package: True)
Found submodule views (is a package: True)
Found submodule audioview (is a package: True)
Found submodule chained_functions (is a package: True)
[...]

PR #8359, opened 2024-07-22
[#7220] Remove global var and magic methods in group blueprint

amercader:
Ref #7720 

This addresses issue number 1 in #7720 (making use of mutable global state). And replaces global vars and magic methods with explicit `is_organization` params and calls to the relevant org / group actions.

Usage of global state started to give consistent test failures after changes in https://github.com/ckan/ckan/pull/8357 to support py 3.11 and 3.12

Looks like a lot of changes but they are very repetitive, it might be even safe to backport to 2.10

### DIFF
--- a/ckan/views/group.py
+++ b/ckan/views/group.py
@@ -2,10 +2,8 @@
 from __future__ import annotations
 
 import logging
-import re
 from collections import OrderedDict
 from typing import Any, Optional, Union
-from typing_extensions import Literal
 
 from urllib.parse import urlencode
 import csv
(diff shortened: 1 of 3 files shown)


ckan/views/group.py
```diff
@@ -142,7 +115,8 @@ def index(group_type: str, is_organization: bool) -> str:
     }
 
     try:
-        assert _check_access(u'group_list', context)
+        action_name = 'organization_list' if is_organization else 'group_list'
+        assert check_access(action_name, context)
```
wardi, line 119: seems odd that we assert `check_access` since this function either returns True or raises a NotAuthorized, right?

PR #8387, opened 2024-08-08
Replace hardcoded "public" schema with dynamic search_path in SQLAlchemy URL

### LINKED ISSUE #8008: CKAN should not assume the database schema is always 'public'
## CKAN version
2.10

## Describe the bug
File ckan/model/group.py contains queries which specify the 'public' schema. This breaks if CKAN was installed in a different schema.

### Steps to reproduce
- Install CKAN in a database schema that is not 'public';
- Login;
- Create an organization;
- Logout and login again;
- Check the error in the log (table 'member' is not found).

### Expected behavior
CKAN should work when not installed in public schema.

### Additional details
Error: "(psycopg2.errors.UndefinedTable) relation "public.member" does not exist"

brijeshnec:
Fixes #8008

### Proposed fixes:
This removes the hardcoded "public" schema from the SQLAlchemy URL configuration and replaces it with a dynamic schema selection using the search_path parameter.

Updated SQLAlchemy URL example:
sqlalchemy.url = postgresql://ckan_default:pass@localhost/ckan_default?options=-csearch_path=schemaname

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/model/group.py
+++ b/ckan/model/group.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import datetime
+import urllib.parse
 from typing import Optional, Union, overload
 from typing_extensions import Literal, Self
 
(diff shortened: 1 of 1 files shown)


wardi:
does removing `public.` from these statements not achieve the same goal?

PR #8396, opened 2024-08-13
simpler package indexing, no system plugins

### LINKED ISSUE #8395: simplify package search index
## CKAN version
master

## Describe the bug
- `SynchronousSearchPlugin` is the last of CKAN's "system plugins" — plugins that aren't actually plugins (they are always loaded and various parts of core depend on them)
- Package indexing is implemented using the `IDomainObjectModification` interface, a very low level DB interface. It catches modifications to many DB models and in the case of Packages, calls the `package_show` action and hands the returned data off to be indexed.

This is complicated, obscure and inelegant because it uses a low-level database hook to jump up to the logic layer to get a dict version of a package, when that data is already available in the logic layer.

There's also a special case in the implementation of IDomainObjectModification notifications that swallows all exceptions other than SearchIndexErrors, making indexing difficult to debug.

### Steps to reproduce
Try to trace debug package indexing, have a bad time

### Expected behavior
There shouldn't be any plugins that aren't plugins. Package indexing should be simple.

### Additional details
If we index packages from the logic layer before committing we can reuse package data that we've already generated as our response data for faster response times.

wardi:
Fixes #8395 

### Proposed fixes:
- remove SynchronousSearchPlugin
- add indexing code to logic layer
- remove exception-swallowing in IDomainObjectModification notify code

### Features:

- [ ] includes API changes


wardi:
@Zharktas @pdelboca merged and ready for review again

ckan/config/environment.py
```diff
@@ -65,7 +65,7 @@ def load_environment(conf: Union[Config, CKANConfig]):
         warnings.filterwarnings('ignore', msg, sqlalchemy.exc.SAWarning)
 
     # load all CKAN plugins
-    p.load_all()
+    p.load_all(force_update=True)
```
pdelboca, line 68: Why is force_update required here?

ckan/plugins/core.py
```diff
@@ -136,20 +128,21 @@ def plugins_update() -> None:
...
 
-    load(*plugins)
+    load(*plugins, force_update=force_update)
 
 
 def load(
-        *plugins: str
+        *plugins: str,
+        force_update: bool = False,
 ) -> Plugin | list[Plugin]:
```
pdelboca, line 146: It would be nice to document here what force_update is and why we needed.

ckan/logic/action/create.py
```diff
@@ -306,25 +307,32 @@ def resource_create(context: Context,
...
         raise ValidationError(error_dict)
 
-    resource = updated_pkg_dict['resources'][-1]
+    #  Run package show again to get out actual last_resource
+    both_data = _get_action('package_show')(
+        logic.fresh_context(context, ignore_auth=True, use_cache=False),
+        {'id': package_id, 'use_default_schema': 'both'}
+    )
+    # update index because of defer_commit use above
+    logic.index_update_package_dict(both_data)
```
pdelboca, line 324: What's the specific relation between defer_commit and updating the index?

ckan/lib/search/__init__.py
```diff
@@ -199,23 +155,16 @@ def rebuild(package_id: Optional[str] = None,
...
     if package_id:
-        pkg_dict = logic.get_action('package_show')(context, {
-            'id': package_id
-        })
-        log.info('Indexing just package %r...', pkg_dict['name'])
-        package_index.remove_dict(pkg_dict)
-        package_index.insert_dict(pkg_dict)
+        log.info('Indexing just package %r...', package_id)
+        logic.index_remove_package(package_id)
+        logic.index_update_package(context, package_id)
```
pdelboca, line 163: We are calling `remove` and then `update`. Does this mean that `update` also creates? Or should we call to `create`?

pdelboca (commented):
I like the PR @wardi ! I think it makes the codebase way more explicit and easier to follow when working on indexing.

It confuses me a little bit the `both` parameter, I think it would be nice to document it and I also left some comments on some lines that I think could use a little bit more of explanation about what's going on.

ckan/config/environment.py
wardi: load_environment is used in tests which sometimes need to clear all the plugins by having an empty plugin list

ckan/plugins/core.py
wardi: added docstrings to `load` and `load_all`

ckan/logic/action/create.py
wardi: Thanks for catching this. Removed because it was no longer required.

ckan/lib/search/__init__.py
wardi: Just me blindly following the old code. I've dropped the `remove` to match the behaviour below.

wardi:
@pdelboca thanks, how does it look now?

amercader:
Had a look at this as it directly relates to the work I'm doing on rethinking the search.

This PR looks great but I find the `both` option and including the validated data in `with_custom_schema` a bit confusing, and that it complicates the actions code. If the indexing code is the only one that requires both versions of the dict, could we perhaps make two `package_show` calls in `index_update_package()`, one with `use_default_schema=True` and one with `use_default_schema=False` (both)? Or alternatively validate the  dict directly like the [old code used to do](https://github.com/ckan/ckan/pull/8396/files#diff-99e591024dd20cd60b6dc8fdcc609171fd03af439787f702ea5c62ec66ce6178L114)

amercader:
For the current version we need to keep things as they are but see https://github.com/ckan/ckan/discussions/8444#discussioncomment-12501670 for some thoughts on data_dict/validated_data_dict going forward

wardi:
@amercader it's a significant performance hit on datasets with many resources to do all the package_show work twice, that's why I went with slightly uncomfortable "both" option.

amercader:
@wardi I understand how running package_show twice may affect performance, but if we only focus on validating the data dict, that needs to happen regardless before indexing, right? So if we call package_show once with use_default_schema=True and use_cache=False, and then on `index_update_package()` we validate it with

```
 schema = package_plugin.show_package_schema()
 validated_pkg_dict, _errors = lib_plugins.plugin_validate(...)
```

Wouldn't that take roughly the same time? 

My concern other than the added complexity in `package_show` (mostly, but also in `package_create` and `package_update`) is that this is exposed to the public API, so any user can pass `use_default_schema=both` on a normal package_show call and will get a dict with:

```
{
    "id": "xxxx",
    "name" "test-dataset",
    # ...
    "validated_data_dict": "<JSON dump of the validated_data_dict>",
    "with_custom_schema": {
       # A copy of the same dict
    }
}
```

The `with_custom_schema` is expected, but the `validated_data_dict` comes from using the [cached `data_dict`](https://github.com/ckan/ckan/pull/8396/files#diff-de994d21743eecf63b0983d3dd5c1254224ae84ac8d2b99a00304d0bc5d1650dR1032), which surprise, surprise also contains a `validated_data_dict` :sweat_smile: 

I know context vars are frown upon but this seems like a reasonable thing to hide from the public API. And perhaps it would be the perfect time to introduce an explicit  `for_indexing` key as we discussed.

wardi:
> @wardi I understand how running package_show twice may affect performance, but if we only focus on validating the data dict, that needs to happen regardless before indexing, right? So if we call package_show once with use_default_schema=True and use_cache=False, and then on `index_update_package()` we validate it with
> 
> ```
>  schema = package_plugin.show_package_schema()
>  validated_pkg_dict, _errors = lib_plugins.plugin_validate(...)
> ```
> 
> Wouldn't that take roughly the same time?

Validating first with the default schema and then with a custom schema isn't the same as validating only with the custom schema, so I couldn't use that approach.
 
> My concern other than the added complexity in `package_show` (mostly, but also in `package_create` and `package_update`) is that this is exposed to the public API, so any user can pass `use_default_schema=both` on a normal package_show call and will get a dict with:
> 
> ```
> {
>     "id": "xxxx",
>     "name" "test-dataset",
>     # ...
>     "validated_data_dict": "<JSON dump of the validated_data_dict>",
>     "with_custom_schema": {
>        # A copy of the same dict
>     }
> }
> ```
> 
> The `with_custom_schema` is expected, but the `validated_data_dict` comes from using the [cached `data_dict`](https://github.com/ckan/ckan/pull/8396/files#diff-de994d21743eecf63b0983d3dd5c1254224ae84ac8d2b99a00304d0bc5d1650dR1032), which surprise, surprise also contains a `validated_data_dict` 😅

That's a bug for sure :-)

> I know context vars are frown upon but this seems like a reasonable thing to hide from the public API. And perhaps it would be the perfect time to introduce an explicit `for_indexing` key as we discussed.

If we go with a `for_indexing` option let's make it a normal parameter. There's nothing that is returned that is internal or tied to access from a plugin so why would we want to prevent someone from e.g. building an external service that indexes datasets.

If we decide to drop the default schema version of data from the search index this all gets much simpler. Do you know if the primary consumer of this data version is the harvest extension or anything else?

wardi:
@amercader `use_default_schema=both` has been removed

amercader:
Thanks @wardi , I think this is a massive improvement and lays the foundation for a newer and better search

ckan/lib/search/index.py
```diff
@@ -109,21 +109,17 @@ def index_package(self,
...
 
         # Index validated data-dict
-        package_plugin = lib_plugins.lookup_package_plugin(
-            pkg_dict.get('type'))
-        schema = package_plugin.show_package_schema()
-        validated_pkg_dict, _errors = lib_plugins.plugin_validate(
-            package_plugin,
-            {'model': model, 'session': model.Session},
-            pkg_dict, schema, 'package_show')
+        assert 'with_custom_schema' in pkg_dict, (
```
ThrawnCA, line 115: Why is this parameter now mandatory?

What is code that previously called eg `PackageSearchIndex.update_dict(<data>)` now meant to do? (NB `ckanext-xloader` breaks as a result of this, and I'm unsure what fixing it would look like.)

ckan/logic/action/update.py
```diff
@@ -440,6 +438,17 @@ def package_update(
 
             item.after_dataset_update(context, data)
 
+    return_id_only = context.get('return_id_only', False)
+
+    if return_id_only and context.get('defer_commit'):
+        return pkg.id
+
+    pkg_default, pkg_custom = logic.package_show_default_and_custom_schemas(
```
ThrawnCA, line 446: This is problematic when code has a need to skip `package_show` and go directly to the model, eg https://github.com/ckan/ckan/issues/3245#issuecomment-248298353

How can that code now retrieve the two dict versions?

wardi: `package_update` now recognizes when dataset and resource fields are unchanged skipping much of the wasted work it used to do. As a result `resource_patch` can now be used to efficiently update a single resource's extra field like `datastore_active` without needing to update the model directly.

ckan/lib/search/index.py
wardi: We might want to add an action to force a package to be reindexed (e.g. so that `ckan search-index rebuild` could be done with API calls), but as long as you're using any of the normal API endpoints to update a package this wouldn't normally be required.

ThrawnCA: What if you _aren't_ using the normal API endpoints?

ckanext-ytp-comments updates the index via a `before_dataset_index` hook so that comments associated with the dataset are searchable. So, when a comment is added or changed, it triggers a reindex of the dataset, not directly altering fields of the dataset, merely ensuring that the hook runs. But with this change, that breaks.

This change has made `PackageSearchIndex.index_package` practically unusable by plugins, which is not good.

ThrawnCA: I don't think it's appropriate to make a parameter mandatory, when the value for that parameter is not supplied by `package_show`.

wardi: Internally we have a tricky situation: we want to avoid dictizing a package twice in the common case, but for now we still need to store packages both with the original schema and any custom schema[1].

To replicate the way package_create and package_update index a package from an extension you could copy these two lines of code:

https://github.com/ckan/ckan/blob/8b1d920a5715eaaefc27ef621ce51e7cb2ff453c/ckan/logic/action/create.py#L228-L232

As I mentioned earlier it would be nicer to make this a separate action so that extensions indexing extra data don't need to care about the internal implementation. Such a PR is welcome.

1: IIRC this is for ckanext-harvest to be able to interpret package data from any ckan regardless of schema. We may choose to stop storing the default-schema package versions in solr in 3.0 to greatly simplify indexing.

ThrawnCA: Our team would definitely see value in an API action that triggered update hooks without changing data.

Would it be better as a new action, like `package_touch`, or as a "proceed even if data is unchanged" flag on the existing `package_update`?

wardi: There isn't room to add a flag to `package_update` because a regular parameter might conflict with custom fields, and context parameters don't make sense for things like this that should work as normal API calls.

Also `package_update` behaviour is now "do nothing (including reindexing) if there were no changes" so a new action makes more sense IMO.

I like `package_reindex` because it would do exactly what it says. We could use the same permission check as `package_update`.

`package_reindex` should be documented as intended for use only when data indexed along with the package has been updated, because `package_update` and friends reindex automatically.

We could use it immediately in the search-index cli code instead of:
https://github.com/ckan/ckan/blob/8b1d920a5715eaaefc27ef621ce51e7cb2ff453c/ckan/lib/search/__init__.py#L146

Maybe `package_create` and `package_update` should call it as well so that if a plugin overrides the action, their version will be used in all cases.

> [!NOTE]
> This is also related to the long-planned pluggable search interface work.  Pluggable search will allow swapping out solr and will support indexing all the things, not only packages, so maybe the action should be generic, like an `index_update` action with `entity="package"` and `id=package_id` parameters. There's nothing preventing us from starting with `package_reindex` for now, though.

> [!TIP]
> We should be creating  background jobs to call `package_reindex` for each package in an organization when that organization's title changes, but that could be a separate PR.

ThrawnCA: > I like `package_reindex` because it would do exactly what it says.

There are cases where I would like it to also trigger plugin hooks as if it were a full update, though...

ThrawnCA: Okay, I have found an approach that seems to work across versions, but it's possible that there's still something problematic. Any feedback?

    try:
        # if #9440 is implemented
        toolkit.get_action('package_reindex')(context_, {'id': package_id})
    except KeyError:
        if hasattr(logic, 'index_update_package'):
            # CKAN 2.12
            logic.index_update_package(context_, package_id)
        else:
            # older CKAN
            package_index = PackageSearchIndex()
            package = toolkit.get_action('package_show')(context_, {'id': package_id})
            package_index.index_package(package, defer_commit=False)

PR #8398, opened 2024-08-15
More lenient line length setting for linting

amercader:
I don't want to open a can of worms here, but can we push this to at least 88 (what Black uses)?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,5 @@
 [flake8]
+max-line-length=88
 exclude =
         .*
         ./contrib


pdelboca:
@amercader in order to be able to push backports we will need to backport this as well. We are now accepting in `master` parameters that `dev-v2.11` do not accept and therefore we cannot merge.

See https://github.com/ckan/ckan/pull/8516

Any thoughts?

amercader:
/backport 

@pdelboca Yes sure, we can definitely backport this

amercader:
@pdelboca this is now backported

PR #8401, opened 2024-08-15
compatibility fix for jsonb extras feature

wardi:
Fix for incompatibility introduced by #8288 

@amercader does this resolve the issue you raised in gitter?

### Proposed fixes:

convert extras values to strings for backwards compatibility

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/converters.py
+++ b/ckan/logic/converters.py
@@ -6,6 +6,7 @@ from typing import Any
 import ckan.model as model
 import ckan.lib.navl.dictization_functions as df
 import ckan.logic.validators as validators
+from ckan.lib.navl.validators import unicode_safe
 
 from ckan.common import _, aslist
 from ckan.types import (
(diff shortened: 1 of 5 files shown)


amercader:
Not quite, It does for the case when you are creating a dataset using extras directly:

```
        data_dict = {
            "name": "test-non-string-extras",
            "extras": [
                {
                    "key": "some_number",
                    "value": 1.5
                }
            ]
        }
```
But it does not fix when you are using a scheming field:

```
        data_dict = {
            "name": "test-non-string-extras",
            "some_number": 1.5
        }
```
My guess is that using `convert_to_extras` bypasses the validation on the extra `value` (and probably `key`)

amercader:
@wardi your latest changes fixed the issue with scheming fields. I added a test case to cover this scenario in [2f37de4](https://github.com/ckan/ckan/pull/8401/commits/2f37de46256c460eabeb504563327c6e86b8e95c)

amercader:
No, doesn't need one as it hasn't been released, merging now

PR #8410, opened 2024-08-26
Upgrade SQLAlchemy to v2

smotornyuk:
When CKAN v2.11 is out and its code is completely compatible with SQLAlcheny v2, extension maintainers can start upgrading their extensions. I'll add a page to the wiki with upgrade tips.

As for the core, we can now upgrade to SQLAlchemy v2. The main part of this upgrade includes only changes to the version inside the requirements file. It's enough to install the latest SQLAlchemy and nothing will be broken. 

But v2 of SQLAlchemy also includes built-in type-definition, so we no longer need any mypy or sqlalchemy-stubs addons. So the secondary feature of this PR is making our typing compatible with new types definitions. 

:information_source: When upgrading CKAN locally, you have to remove external type definitions, or they will cause incorrect type detection. Run `pip uninstall -y sqlalchemy2-stubs sqlalchemy-stubs`

For anyone who is going to review this PR, here are the most common types of changes:

* instead of `== False` it's better to use `== false()`. The result is the same, but the latter is accepted by linters. I changed this expression only inside files that were changed due to other reasons.
* `model.Session.query` is properly typed now, so we don't need explicit types when assigning its result into a variable. `x: Query[...] = model.Session.query(...)` transformed into `x = model.Session.query(...)`
* `model.Session.query` now always returns a sequence of the `Row` object. `list[tuple[TYPE]]` transformed into `Sequence[Row[tuple[TYPE]]]`. Sometimes `list` is used instead of the `Sequence`, to reduce number of changes. But for new code it's recommended to use `Sequence`.
* `.as_mutable(JSONB)` - this type signature is not yet completed and requires `# type: ignore` comment
* Whenever model's classmethod returns `Self`, `.query(cls)` must be used instead of static form, like `.query(Tag)`, because `Tag` class is not final and can be extended by child class
* `pgcode` property exists in SQLAlchemy exceptions only when working with PostgreSQL. Type checker doesn't know that we are not going to use other DB engines, so any access to `pgcode` must be wrapped into a typecast expression.

### DIFF
--- a/ckan/config/environment.py
+++ b/ckan/config/environment.py
@@ -214,7 +214,7 @@ def update_config() -> None:
     # to eliminate database errors due to stale pooled connections
     config.setdefault('sqlalchemy.pool_pre_ping', True)
     # Initialize SQLAlchemy
-    engine = engine_from_config(config)
+    engine = engine_from_config(dict(config))
     model.init_model(engine)
 
     for plugin in p.PluginImplementations(p.IConfigurable):
--- a/ckan/lib/dictization/__init__.py
+++ b/ckan/lib/dictization/__init__.py
@@ -130,7 +130,7 @@ def table_dict_save(table_dict: dict[str, Any],
     if id:
         obj = session.query(ModelClass).get(id)
 
-    if not obj:
+    if not obj and isinstance(table, Table):
         unique_constraints = get_unique_constraints(table, context)
         for constraint in unique_constraints:
             params = dict((key, table_dict.get(key)) for key in constraint)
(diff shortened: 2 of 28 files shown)


amercader:
@wardi @smotornyuk I merged master and got a lot of [scary typing failures](https://github.com/ckan/ckan/actions/runs/12083078508/job/33695472133) that luckily went away after adapting the tracking model definition (https://github.com/ckan/ckan/pull/8410/commits/6325fa4da2928c23791ba512e9c7ea7de17af3f8)

This PR looks good to me. I think it would be good to merge this and start working with it on master going forward to identify potential issues way ahead of a 2.12 release.

amercader:
@wardi @smotornyuk ping any objections to merge this?

PR #8411, opened 2024-08-26
Remove v2.11 changelog entries from upstream

smotornyuk:
These changelog entries were used to generate the v2.11 changelog and are no longer required.


amercader:
Sorry, that should be part of the release process. I guess you deleted the files that are both on the dev-v2.11 branch and master and kept the ones only in master right? We should automate this with a command and make it part of the release instructions

smotornyuk:
Yes. And I checked that all remaining files are indeed not mentioned in changelog.

The challenging part here is synchronizing the master branch with the release branches. When the changelog is compiled inside the release branch, the CLI command suggests removing fragments and then we can commit changes. But as this happens inside a release branch, we have to keep a list of removed files and then remove these files from the master branch.

As an alternative, whenever we create a new release branch, we can remove everything from the `changes` folder inside the master branch - all these fragments are included in the release branch and will appear in the changelog. It leaves us only with backported PRs, that are created after branching. For them, we can remove changelog fragment from master branch whenever backport is successfully merged into release branch via git workflow

Issue #265, opened 2013-01-10
Exception: The CKAN and datastore database are the same.


seanh:
The default values for the datastore settings in the ini file template don't work:

```
ckan.datastore.write_url = postgresql://ckanuser:pass@localhost/ckantest
ckan.datastore.read_url = postgresql://readonlyuser:pass@localhost/ckantest
```

gives:

```
Traceback (most recent call last):
  File "/home/seanh/.virtualenvs/ckan/local/lib/python2.7/site-packages/nose/suite.py", line 208, in run
    self.setUp()
  File "/home/seanh/.virtualenvs/ckan/local/lib/python2.7/site-packages/nose/suite.py", line 291, in setUp
    self.setupContext(ancestor)
  File "/home/seanh/.virtualenvs/ckan/local/lib/python2.7/site-packages/nose/suite.py", line 300, in setupContext
    self.config.plugins.startContext(context)
  File "/home/seanh/.virtualenvs/ckan/local/lib/python2.7/site-packages/nose/plugins/manager.py", line 99, in __call__
    return self.call(*arg, **kw)
  File "/home/seanh/.virtualenvs/ckan/local/lib/python2.7/site-packages/nose/plugins/manager.py", line 167, in simple
    result = meth(*arg, **kw)
  File "/home/seanh/Projects/ckan/ckan/ckan/ckan_nose_plugin.py", line 40, in startContext
    plugin.configure(config)
  File "/home/seanh/Projects/ckan/ckan/ckanext/datastore/plugin.py", line 59, in configure
    self._check_separate_db()
  File "/home/seanh/Projects/ckan/ckan/ckanext/datastore/plugin.py", line 134, in _check_separate_db
    raise Exception("The CKAN and datastore database are the same.")
Exception: The CKAN and datastore database are the same.
```

Issue #224, opened 2012-12-19
css - Main menu active


tobes:
@johnmartin 

On master when I go to a url such as http://localhost:5000/organization the main menu item should be highlighted eg organizations in this example

The reason is that the html is `<li class="active"><a>...` before it was `<li><a class="active">...` 

I thought we had sorted this previously.

thanks

PR #8419, opened 2024-09-02
[Bugfix/datapusher 265] Disable the button for unsupported file type formats

Nisha1293:
Currently, the "Upload to DataStore" and "Delete from DataStore" buttons are enabled by default for all file formats. But, DataStore only supports tabular data. 

Therefore, for file formats such as JSON and PDF, these buttons should be disabled.

This PR addresses the following issues:

[Issue #224](https://github.com/ckan/datapusher/issues/224)
[Issue #265](https://github.com/ckan/datapusher/issues/265)

### DIFF
--- a/ckanext/datapusher/views.py
+++ b/ckanext/datapusher/views.py
@@ -46,7 +46,7 @@ class ResourceDataView(MethodView):
             toolkit.g.pkg_dict = pkg_dict
             toolkit.g.resource = resource
 
-            # Determine if the upload button and delete button 
+            # Determine if the upload button and delete button
             # should be enabled or disabled
             resource_format = resource.get('format', '').strip()
             supported_formats = toolkit.config.get('ckan.datapusher.formats')


amercader:
This could be replicated in the xloader extension as well

wardi:
It should still be possible to delete datastore tables if they exist but an unsupported format file was later uploaded to the same resource.

Nisha1293:
> It should still be possible to delete datastore tables if they exist but an unsupported format file was later uploaded to the same resource.

@wardi I am  working on this point.
In the current code, there is one issue: if data is not available in the datastore for the existing resource and the user clicks the "Delete from DataStore" button, it results in a 500 Internal Server Error.

I am updating the my patch request and it will also address this issue. 
The revised code will ensure that the "Delete from DataStore" button is only functional if data is available for the existing resource, regardless of whether the data was uploaded before or after an unsupported format file was added. If no data is available, the button will be disabled to prevent errors.

ckanext/datapusher/views.py
```diff
@@ -46,6 +46,16 @@ def get(self, id: str, resource_id: str):
...
 
+            # Determine if the upload button and delete button
+            # should be enabled or disabled
+            resource_format = resource.get('format', '').strip()
+            supported_formats = toolkit.config.get('ckan.datapusher.formats')
+            is_upload_supported = (
+                resource_format in supported_formats or
+                resource_format.lower() in supported_formats
+            )
+            is_delete_supported = resource.get('datastore_active', False)
```
tino097, line 57: this looks like it will prevent deleting all the resources that have been uploaded to the portal

Nisha1293: delete from datastore button will be disable only when file is not present to datastore. 
In current scenario, if data is  .not present is datastore, and the user clicks the "Delete from DataStore" button, it results in a 500 Internal Server Error.

PR #8421, opened 2024-09-05
package_update: faster dataset metadata update

### LINKED ISSUE #8407: Faster dataset metadata updates with package_update
Building on performance improvements in #8360:

We can define the dataset fields on which resource validation may depend so we can skip validation and dictization of all resources when only other dataset fields change. This will significantly improve performance of `package_update` for datasets with many resources when editing dataset fields.

Let's add a `resource_validation_dependencies` method to IDatasetForm that will allow plugins like `ckanext-scheming` to return a list of dataset field names that may affect resource validation. The default would be an empty list, but for the upcoming https://github.com/ckan/ckanext-scheming/pull/418 the list would include `'state'` so that dataset state changes can enforce/skip required checks on fields in resources.

Let's also formalize updating dataset metadata without changing resources when a user does not pass a `resources` list to `package_update`. Currently this is only possible from the UI with a `partial_update` context parameter. There's no reason to limit this functionality to internal users.

wardi:
Fixes #8407

### Proposed fixes:
- faster dataset metadata update with `IDatasetForm.resource_validation_dependencies` for declaring dataset fields that may affect resource validation (default: none)
- allow sysadmin to set `metadata_modified` value on datasets - useful for harvesting or mirroring ckan sites
- only update `metadata_modified` when dataset or resource metadata has changed
- remove `allow_partial_update` context parameter and instead allow normal API users to:
  - update dataset metadata without passing `resources`
  - update group/org metadata without passing `users`, `datasets` and `groups`

| 100-resource dataset action | old | new | improvement |
| --- | --- | --- | --- |
| package_patch (dataset field) | .729 | .381 | 90% |
| package_patch (no change after validation) | .537 | .046 | 1000% |

### Features:

- [X] includes API changes


wardi:
This just needs a few more tests

ckan/logic/action/update.py
```diff
@@ -290,26 +292,51 @@ def package_update(
...
+    if data_dict.get('metadata_modified'):
+        dataset_changed = original_package != data_dict
+    else:
+        dataset_changed = dict(
+            original_package, metadata_modified=None, resources=None
+            ) != dict(data_dict, metadata_modified=None, resources=None)
+    if not dataset_changed and original_package.get('resources'
+            ) == data_dict.get('resources'):
+        # flag no change for caller
+        context['package_update_unchanged'] = 'before validation'
```
wardi, line 312: Not in love with this part but it's the simplest possible way I could think to let the activity extension skip creating useless activity entries. I don't like that this is a new Context value that's a really specific case not a general solution.

Alternatives I considered:
- Create a generic way to return extra information in the API envelope via the context that could be used by other actions. This would be useful for API users so they would know if an update actually occurred after any of the package or resource updating actions.
- Create a new interface for the activity plugin that could be passed both the `original_package` and updated `data_dict` from `package_update` so it doesn't have to call `package_show` itself or and the generic signal handling.

wardi:
Performance measurements added to description

changes/5713.feature
```diff
@@ -1,5 +1,10 @@
 Update resources faster by selectively skipping resource validation and
-database updating for unchanged resources.
+database updating for unchanged resources. Now activities are created and
+metadata_modified updated only if there is a real change. metadata_modified
+may now be set by sysadmins (useful for harvesting or mirroring).
+
+The allow_partial_update context parameter has been removed, now normal API
+users may call package_update without passing resources.
```
amercader, line 7: What will be the effect of this?
```suggestion
users may call package_update without passing resources. In this case the existing resources
will remain untouched instead of being deleted.
```

ckan/logic/action/update.py
```diff
@@ -387,47 +409,56 @@ def package_update(
             and plugin_data
         )
 
-    pkg = model_save.package_dict_save(
+    nested = model.repo.session.begin_nested()
```
amercader, line 412: Can you explain why this is needed?

ckan/logic/action/update.py
amercader: We are getting rid of an ugly context var and introducing an ugly context var so it evens out :)

> Create a generic way to return extra information in the API envelope via the context that could be used by other actions. 

You mean something like `context["extra_data"]["package_update_unchanged"] = "before validation"`?
This looks like encouraging sharing info in the context, so perhaps I prefer a specific key that is clear where it comes from. Unless I misunderstood?

> This would be useful for API users so they would know if an update actually occurred after any of the package or resource updating actions.

Not sure what you mean here. You mean an update happening e.g. in a chained action?

wardi: @amercader I'm thinking something like `context['updated_entities']['package'] = [id]` and then expose that `updated_entities` in the API response envelope or something so that API users can know when something has been modified as a result of the action they call. Wouldn't need to implement it for everything right away but it would naturally extend to resources, groups, orgs, users etc. and bulk actions that may update many things at once.

wardi: As discussed at meeting I'll implement this as `context['changed_entities']` to potentially encompass creation and deletion in the future, add this value to the response envelope for API calls and document it in the changelog.

ckan/logic/action/update.py
wardi: This lets us commit or rollback just the changes to this package without affecting the outer transaction in case `defer_commit=True` is passed in the context

JVickery-TBS:
@wardi @amercader I am not sure if this actually requires a migration script or not?

In previous versions of CKAN, when deleting a Resource, the position stays as it's initial number and the other Resources do not get changed.

E.g. if you have 4 Resources, order 1 2 3 4. And you delete the 3rd one, you will now have 3 Resources but they keep their `position` 1 2 and 4. THIS is how older CKAN versions worked right?

So here is the issue with this new code that I have run into. We have packages with all these deleted resources and they have orders like 1 2 and 4. Then this new code is gunna raise IndexErrors when looping package_update and resource_patch as it loops on resource index / resource position key/value pairs. And because the new code kind of skips over actually doing package_dict_save and committing if nothing has changed, and you cannot query deleted resources, I have not come up with a super simple way of mitigating this issue.

We DID undelete a resource, and then delete it in UI and that forced all the resources to get correct positions. I am gunna try to think of a way to do a migration script for this if that makes sense??

JVickery-TBS:
Maybe not...sorry haha. I didn't backport this line: https://github.com/ckan/ckan/commit/e8ec7fb06325f792a2159b2d9f92c1cb664a8fa2 and that seemingly cause datasets to magically delete resources. and then not update positions. and then things blew up.

PR #8436, opened 2024-09-13
[UI-Revamp] Dataset create page

steveoni:
Fixes #

### Features:
Implement UI for dataset page based on https://www.figma.com/design/oI3wDSdZ0cHZLh6CrfYMnb/CKAN-3.0-Design-System?node-id=1887-11502&node-type=INSTANCE&m=dev

- [ ] includes API changes

### DIFF
--- a/ckan/public/base/css/main-rtl.css
+++ b/ckan/public/base/css/main-rtl.css
@@ -11927,12 +11927,12 @@ td.diff_header {
   color: #6c757d;
 }
 
-.toolbar .home a {
+/* .toolbar .home a {
   text-decoration: none;
-}
+} */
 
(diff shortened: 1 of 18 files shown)


steveoni:
Note:

* The last stage of the create dataset  needs a refined way to display dataset metadata and its resources
* some of the icons used a replica or look-alike from font-awesome
* some UI elements are left over from the previous ckan ui. such as the `Required` notes at the bottom of the first stage form.

wardi:
This work should be part of the new public folder/templates created by #8470 not the default ones. We'll have at least one release with the option to enable the new UI but it won't be the default because it's likely to break any template customization done in other extensions

PR #8449, opened 2024-09-20
User details break group display

### LINKED ISSUE #8438: ckan.auth.public_user_details breaks group display
## CKAN version
2.11

## Describe the bug
Setting the option `ckan.auth.public_user_details = false`  to hide members breaks the display of groups for public and registered members (but not sysadmin)
Error-log:  

>   File "/usr/lib/ckan/default/src/ckan/ckan/logic/action/get.py", line 178, in member_list
>     _check_access('group_show', context, data_dict)
>   File "/usr/lib/ckan/default/src/ckan/ckan/logic/__init__.py", line 391, in check_access
>     raise NotAuthorized(msg)
> ckan.logic.NotAuthorized: User xxxxxx not authorized to read group be2b8ec2-f0ca-4746-8d3...

Note the [docs ](https://docs.ckan.org/en/2.11/maintaining/configuration.html#ckan-auth-public-user-details) do say it will " will raise a Not Authorized exception." which does appear to be true but there is no graceful trapping of the exception.  The page is killed off part way through rendering (after the side bar generated but before CSS and main page (screenshot below)

### Steps to reproduce

1. set `ckan.auth.public_user_details = false`  and cycle the server
2. log out
3. go to the groups menu and display groups
4. click on a group

### Expected behavior
I was expecting either:

1. the interface tab with the members list to be hidden, but the group data and about tabs  to remain visible  (preferred)
or...
2. a graceful error message saying the group was for members only

### Additional details
related issue  [Hide list of users by default #3441](https://github.com/ckan/ckan/issues/3441)

screen shot

image

tino097:
Put the member_list method in try block and return 0 when not Authorized

Fixes #8438

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2018,7 +2018,10 @@ def member_count(group: str) -> int:
         u'id': group,
         u'object_type': u'user'
     }
-    return len(logic.get_action(u'member_list')(context, data_dict))
+    try:
+        return len(logic.get_action(u'member_list')(context, data_dict))
+    except logic.NotAuthorized:
+        return 0
 
 
 @core_helper


amercader:
Can you add a little changelog entry @tino097 ? Otherwise this is good to go

PR #8437, opened 2024-09-16
[#8316] Fix exception in `ckan generate extension` command

amercader:
Fixes #8316

In our post_gen_project hook, we were dumping the cookiecutter object as a Python object but using the `jsonify` filter. Cookicutter 2.6 introduced a new item in the context which is None by default, and this got translated as `null` when dumping it, which is not valid Python syntax.

I've refactored the hook a bit so it dumps the context in a bit more safely manner.

- [x] includes bugfix for possible backport

Issue #8382, opened 2024-08-01
Transition to pyproject.toml


Zharktas:
In https://github.com/ckan/ckan/pull/6620 setup.py was largely simplified but still some bits remain in old setup.py files. Now with lates pip and setuptools installing extensions as editable for development, there are following warnings:

`DEPRECATION: Legacy editable install of <some_extension> from file:///srv/app/ckanext/<some_extension> (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457`

Should we try to solve the remaining issues and start transitioning to pyproject.toml configurations in extensions and core or do we want to maintain setup.py/setup.cfg configurations ?

duttonw:
ckan core may be simplish, but when you start to add all the extensions, the order of each dependency install starts to matter.  I'd love to see us following and promoting a 'best of breed' way of installing and allowing the mangability of direct from git/tag.  

Unsure if Python has this but in the npm project ecosystem you can point to a repo/branch|commit|tag and it will download, and run the 'prepare' tag to do local "snapshot" style include.

EricSoroos:
@duttonw 

Install order shouldn't matter, and can't matter if the requirements are specified and obeyed. It only matters if you're installing one dependency, and then separately installing another extension with conflicting requirements. If you can't install them together, with one resolution, technically you've got a broken system and one of the extensions is operating outside of it's desired versions. The python system may refuse to run in this case.

There's a definite balancing act with extensions and pinning versions:
1) Don't ever pin a version of a dependency that's part of core ckan.  Requiring a range of versions on CKAN is appropriate.
2) Avoid tight pins on anything that's common and 'stable', e.g. six, requests, pillow.
3) Only have tight/exact pins for things that you're particularly dependent on. 
4) pip install ckanext-foo must install the requirements, it should never be `pip install ckanext-foo ; pip install -r ckanext-foo/requirements.txt`

Then, when someone installs, the pip dependency solver can do the best job getting some combination of all of the requirements. 

Finally, when the person responsible for the system gets a tested build, it's then possible to freeze versions and install from that list, essentially giving the solver the required solution, rather than letting it build up. 

[...]

PR #8456, opened 2024-09-24
[#8382] Packaging updates in the extensions template

amercader:
Refers to #8382 (this also includes #8437)

As discussed in #8382, this PR:

* Updates the documentation to remove instances of the deprecated `python setup.py develop` and `python setup.py sdist upload` commands, and to not recommend editable installs unless it's a development install
* Adds a new pyproject.toml file to the extension template which includes all project information, except for:
    * setuptools config options, which go to `setup.cfg`
    * Babel config options, which still hasn't mature pyproject.toml support
* Adds the `[build-system]` section to CKAN core's pyproject.toml

There are no changes with relation to namespace packages or requirements in this PR

### DIFF
--- a/contrib/cookiecutter/ckan_extension/hooks/post_gen_project.py
+++ b/contrib/cookiecutter/ckan_extension/hooks/post_gen_project.py
@@ -50,9 +50,8 @@ def recut():
     keywords = context['keywords'].strip().split()
     keywords = [keyword for keyword in keywords
                 if keyword not in ('ckan', 'CKAN', 'A', 'space',
-                                   'seperated', 'list', 'of', 'keywords')]
+                                   'separated', 'list', 'of', 'keywords')]
     keywords.insert(0, 'CKAN')
-    keywords = u' '.join(keywords)
     context['keywords'] = keywords
 
     # Double check 'project_shortname' and 'plugin_class_name'
--- a/contrib/cookiecutter/ckan_extension/hooks/post_gen_project.py
+++ b/contrib/cookiecutter/ckan_extension/hooks/post_gen_project.py
@@ -36,7 +36,15 @@ def recut():
     setup_template = 'setup.py'
 
     # get context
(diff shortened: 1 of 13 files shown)


contrib/cookiecutter/ckan_extension/hooks/post_gen_project.py
```diff
@@ -36,15 +36,22 @@ def recut():
     setup_template = 'setup.py'
 
     # get context
-    context = {{ cookiecutter | jsonify }}
+    context = {}
+
+{%- for key, value in cookiecutter.items() -%}
+{% if value is not none %}
+    context["{{key}}"] = {{ value|tojson }}
```
wardi, line 43: True and False as well as None inside something like a list or dict would lead to problems too. We'd want `repr` here but I don't think we get that in jinja2

contrib/cookiecutter/ckan_extension/{{cookiecutter.project}}/setup.py
```diff
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 from setuptools import setup
 
+# Note: Do not add new arguments to setup(), instead add setuptools
```
wardi, line 4: maybe add a link to the issue explaining why

contrib/cookiecutter/ckan_extension/hooks/post_gen_project.py
```diff
@@ -36,15 +36,22 @@ def recut():
...
+{% else %}
+    context["{{key}}"] = None
+{% endif %}
```
wardi, line 46: This might cover the edge cases better

```suggestion
    context["{{key}}"] = {{ value.__repr__() | safe }}
```

amercader:
@wardi I've reverted the install command back to editable as discussed in the dev meeting, this is ready for review

PR #8458, opened 2024-09-25
Pluggable DataStore Formats

JVickery-TBS:
feat(dev): datastore dump implement;

- Pluggable datastore dump formats.

### Proposed fixes:

Interfacing for DataStore dump formats.

### Features:

- [ ] includes API changes


ckanext/datastore/blueprint.py
```diff
@@ -36,17 +34,28 @@
...
 
+def dump_formats():
+    dump_formats = [
+        formats.CSV(),
+        formats.TSV(),
+        formats.JSON(),
+        formats.XML(),
+    ]
+    for plugin in p.PluginImplementations(IDatastoreDump):
+        dump_formats.append(plugin())
```
wardi, line 50: might want to `extend` here so that a plugin can define more than one format or pass in the current list and have the plugin return a modified list so that plugins could override/remove other formats (like removing the CSV/TSV ones and replacing them with ones that set `bom=True`)

ckanext/datastore/blueprint.py
```diff
@@ -92,25 +101,17 @@ def dump(resource_id: str):
...
-                                    name=resource_id)
-        content_type = b'application/json; charset=utf-8'
-    elif fmt == 'xml':
-        content_disposition = 'attachment; filename="{name}.xml"'.format(
-                                    name=resource_id)
-        content_type = b'text/xml; charset=utf-8'
+    for d in dump_formats():
+        if fmt == d.get_format():
+            format_class = d
+            break
```
wardi, line 107: a `{format_name: format_class}` dict would be better for the formats so we can look them up or get all the format names instantly.

ckanext/datastore/blueprint.py
```diff
@@ -223,23 +224,13 @@ def post(self, id: str, resource_id: str):
...
-        writer_factory = json_writer
-        records_format = 'lists'
-    elif fmt == 'xml':
-        writer_factory = xml_writer
-        records_format = 'objects'
-    else:
+    writer_factory = fmt_class.get_writer
+    records_format = fmt_class.get_records_format()
+    if not records_format:
         assert False, 'Unsupported format'
```
wardi, line 234:
```suggestion
```

ckanext/datastore/formats.py
```diff
@@ -0,0 +1,209 @@
...
+    def write_records(self, records: list[Any]) -> bytes:
+        self.output.write(records)  # type: ignore
+        self.output.seek(0)
+        output = self.output.read().encode('utf-8')
+        self.output.truncate(0)
+        self.output.seek(0)
+        return output
+
+    @contextmanager
+    def get_writer(self, fields: list[dict[str, Any]], bom: bool = False):
```
wardi, line 38: We need a way for a site to choose that the bom is set to True for these, so it could be a parameter to `__init__` instead

ckanext/datastore/interfaces.py
```diff
@@ -232,3 +235,82 @@ def update_datastore_info_field(
         in the data dictionary page.
         """
         return field
+
+
+class IDatastoreDump(interfaces.Interface):
```
wardi, line 240: Interfaces shouldn't have any code in them, this could be a normal base class (e.g. `DatastoreDumpFormat`) instead

ckanext/datastore/interfaces.py
```diff
@@ -232,3 +235,82 @@ def update_datastore_info_field(
...
+        if columns:
+            self.id_col = columns[0] == '_id'
+            if self.id_col:
+                columns = columns[1:]
+        self.columns = columns
+
+
+    def __call__(self, output: Optional[Union[StringIO, BytesIO]]=None, columns: Optional[list[str]]=None):
+        self.__init__(output, columns)
+        return self
```
wardi, line 255: If we call an instance the `__init__` is called again with new parameters? That's an unusual pattern. There's almost no cost to creating a new instance, and if there was a normal method called `reset` or `set_parameters` would be better.

PR #8481, opened 2024-10-14
Remove jQuery truncator plugins

amercader:
This hasn't been used for a while and was creating code scanning issues


amercader:
@aleeexgreeen as you start migrating pages in the new theme it would be good to remove anything that is no longer used, e.g jquery plugins, ckan modules, etc

PR #7934, opened 2023-11-22
[7723]Fix user's email is not unique

TomeCirun:
Fixes #7723

### Features:

- [x] includes tests covering changes
- [ ] includes updated documentation
- [ ] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #8491, opened 2024-10-23
Redefine `by_email`

avdata99:
Even after #7934, #6916, #5100 and probably others we can't never be sure about duplicated email.
Saml2 and other extensions trust in the `User.by_email` function to get the user by email. Usually we expect that function to return `None` or a unique user, but if there are duplicated emails, we can expect an error.

Migration `102_ff13667243ed_create_conditinal_index_on_user` mark as unique a combination of `email` and `state` but not takes into consideration upper and lower case for the email.

I think is required to redefine the `User.by_email` function or add another one.

If you have an active and an inactive user with the same email, this function can return the inactive one.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8491.misc
+++ b/changes/8491.misc
@@ -1,0 +1,10 @@
+After [#7934](https://github.com/ckan/ckan/pull/7934),
+[#6916](https://github.com/ckan/ckan/pull/6916)
+[#5100](https://github.com/ckan/ckan/pull/5100) and many other we can't never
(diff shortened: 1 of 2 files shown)


wardi:
It sounds like the we should fix the index to apply a unique index on emails in lowercase so that we can't have an exception raised by this method.

IIUC technically the local part of emails can handle uppercase and lowercase differently but it's reasonable to prevent emails differing only by case in ckan.

wardi:
should be resolved by #8626

PR #8492, opened 2024-10-24
[Midnight-Blue] header

aleeexgreeen:
### UI Changes:

- Removes login button when user is logged in.
- Adds new ckan logo
- Center aligns the navigation list
- Introduces brand colours and new gray colours
- Adds `aria-current` attribute to the active navigation item, so screen reader users are aware of which tab is currently active.

** I reformatted `header.html` so it is easier to read/work with 
** There is no design for the account masthead portion yet
### Features:

- [ ] includes API changes


ckan/templates-midnight-blue/header.html
```diff
@@ -1,125 +1,119 @@
...
-      </div>
-    </nav>
-  </div>
-</header>
-{% endblock %}
+            <ul class="navbar-nav mb-2 mb-lg-0 col-md-10">
+              {% block header_site_navigation_tabs %}
+                {% set org_type = h.default_group_type('organization') %}
+                {% set group_type = h.default_group_type('group') %}
+                {{ h.build_nav_main(
```
amercader, line 95: In general, for the new theme we want to try to move away from python helpers _that output HTML_ (not python helpers in general) in favour of straight Jinja code or Jinja macros. This makes it much easier for plugins to customize and avoids the spaghetti code we now have.

.e.g. `build_nav_main` calls `_make_menu_item` calls `_link_active` and `_link_to`, which calls `_link_class`, `_create_link_text` and `_link_aria_current` to finally call `link_to`. 

And all this to output:

```html
<li class="active"><a aria-current="page" href="/dataset/">Datasets</a></li>
<li><a href="/organization/">Organizations</a></li>
<li><a href="/group/">Groups</a></li>
<li><a href="/about">About</a></li>
```

The same could be achieved with something like (`h.page_is_active()` does not exist, we can copy whatever logic is used now):

```
<li {% if h.page_is_active() %}class="active{% endif %}>
  <a {% if h.page_is_active() %}aria-current="page"{% endif %} href="{% h.url_for(dataset_type ~ '.search') }">
  {% h.humanize_entity_type('package', dataset_type, 'main nav') or _('Datasets') %}
  </a>
</li>
...
```
If we wanted to make it slightly more convenient we can create a Jinja macro in `ckan/templates/macros/nav_link.html`:

```
{% macro nav_link(url, text): %}
<li {% if h.page_is_active() %}class="active{% endif %}>
  <a {% if h.page_is_active() %}aria-current="page"{% endif %} href="{{ url }}">
  {{ text }}
  </a>
</li>
```
and then call it from `header.html`:

```
{% import 'macros/nav_link.html' as nav_link %}

{{ nav_link(h.url_for(dataset_type ~ '.search'),  h.humanize_entity_type('package', dataset_type, 'main nav') or _('Datasets')) }}
{{ nav_link(h.url_for(org_type ~ '.index'),  h.humanize_entity_type('organization', org_type, 'main nav') or _('Organizations')) }}
{{ nav_link(h.url_for(group_type ~ '.index'),  h.humanize_entity_type('group', group_type, 'main nav') or _('Grpups')) }}
{{ nav_link(h.url_for('home.about'), _('About')) }}
```

@aleeexgreeen hope this makes sense, I'm keen to hear your thoughts on this, what form do you think is more convenient/user friendly?

amercader (commented):
@aleeexgreeen see comments, also

Can you align these vertically? And shouldn't all items have the same color as the log out link as they are all clickable icons?
Screenshot 2024-10-25 at 14-15-08 Dataset - CKAN

The logo looks very low res and cut on the right. Perhaps @thegostev or @Starl3n can provide a higher res one?
Screenshot 2024-10-25 at 12-58-14 Welcome - CKAN

ckan/lib/helpers.py
```diff
@@ -770,14 +770,26 @@ def _create_link_text(text: str, **kwargs: Any):
...
     title = kwargs.pop('title', kwargs.pop('title_', None))
+    aria_current = _link_aria_current(kwargs)
+
     return link_to(
         _create_link_text(text, **kwargs),
         url_for(*args, **kwargs),
         cls=cls,
-        title=title
+        title=title,
+        aria_current = aria_current
```
amercader, line 792: This should fix the lint failure

```suggestion
        aria_current=aria_current
```

ckan/public-midnight-blue/base/css/main.css
```diff
@@ -18,96 +18,97 @@
   --bs-teal: #20c997;
```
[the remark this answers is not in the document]
aleeexgreeen: I'm using `npm run build`, are they performing different functions?

ckan/templates-midnight-blue/header.html
aleeexgreeen: Hi! It makes a lot of sense to do it this way, it will simplify the process of adding attributes and elements to the links in the future. 

I think going with the macro is the best idea, I can see it being used for breadcrumbs, tabs and potentially pagination as well, which all require an `aria-current` attribute for better accessibility (since the tabs being used are links `aria-current="page"` is appropriate to use instead of `aria-selected="true"`). 

The macro solution is super convenient if we can pass additional classes, an icon, the url, and the link text (those are the only ones I can think of right now). It will be used in place of helpers `build_nav_main`, `build_nav_icon` and `build_nav` , making `_make_menu_item` obsolete, so I'm all for this option.

aleeexgreeen:
> Can you align these vertically? And shouldn't all items have the same color as the log out link as they are all clickable icons? 

I vertically aligned the items for now. I haven't received a design for the masthead yet, so depending on if I get one or not, this bar may look completely different. I've changed the link colour as well

amercader:
@aleeexgreeen how's it going? do you need some help with this?

aleeexgreeen:
> @aleeexgreeen how's it going? do you need some help with this?

Hi! Sorry, was waiting to hear back about the header image. I've updated it, but it doesn't seem to be any better quality wise. I think I will try an svg version.

In terms of the macros, I have come up with something that works for the majority of cases, it just does not work for individual group and organization pages, because of endpoint requirements. If I were to pass the url, it would work to create the link but since I've just taken the way `build_nav_main()` to see if the page is active, passing the URL wouldn't work to check if the page is active.

Another thing, using the macros rather than the python helper means that the nav menu item of the organization and groups page no longer are active when on an individual group or organization. Is this a problem? The ckan demo site doesn't have the organization/group nav item active when on an individual/group as well, so I'm not sure if this was a new feature. Example: `templates` and `templates-midnight-blue`

Screenshot 2024-11-11 at 1 16 20 PM
Screenshot 2024-11-11 at 1 16 49 PM

amercader:
hi @aleeexgreeen sorry for the late reply on this, I got pulled over to other stuff

TL;DR: I think keeping the navigation links when on individual dataset/orgs/groups is important, demo is broken, check the diff below for restoring this behaviour

---

More detailed explanation:

* To handle these kind of cases where we want to keep the link active in a page that is not the linked one in the navigation link the old functions use the `highlight_controllers` parameter (which is obvs not very self-explanatory). You could pass e.g. `highlight_controllers=['organization']` and then in all org pages (index, individual orgs, edit org, etc, i.e. all `organization.*` endpoints) the nav link would be active. 
* The diff below restores this param, adding it to the macro and the `page_is_active()` helper,which then passes to the existing `_link_active()`
* On naming, "controllers" and "actions" are leftover terms from an ancient web framework that CKAN used before, Pylons. In the Flask world these are blueprints and endpoints. We are not touching the existing functions but for new macros and helpers we can use the new terms (they will go when we remove support for the old templates and with them the old python helpers). That's why I called the param `active_blueprints` in the macro and new helper. In `page_is_active` we map the new terms to the old terms so nothing needs to be changed there.

Let me know if this makes sense






```diff
diff --git a/ckan/lib/helpers.py b/ckan/lib/helpers.py
index b717446a8..25afe7e93 100644
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -740,12 +740,15 @@ def get_flashed_messages(**kwargs: Any):
 
 
 @core_helper
-def page_is_active(menu_item: str) -> bool:
+def page_is_active(
+        menu_item: str, active_blueprints: Optional[list[str]] = None) -> bool:
     '''Returns whether the current link is the active page or not'''
-    controller, action = menu_item.split('.')
+    blueprint, endpoint = menu_item.split('.')
+
     item = {
-        'action': action,
-        'controller': controller
+        'controller': blueprint,
+        'action': endpoint,
+        'highlight_controllers': active_blueprints,
     }
     return _link_active(item)
 
diff --git a/ckan/templates-midnight-blue/header.html b/ckan/templates-midnight-blue/header.html
index 46763d53f..2be41ba8e 100644
--- a/ckan/templates-midnight-blue/header.html
+++ b/ckan/templates-midnight-blue/header.html
@@ -94,9 +94,9 @@
                 {% set org_type = h.default_group_type('organization') %}
                 {% set group_type = h.default_group_type('group') %}
 
-                {{ nav.link(dataset_type ~ '.search', h.humanize_entity_type('package', dataset_type, 'main nav') or _('Datasets')) }}
-                {{ nav.link(org_type ~ '.index', h.humanize_entity_type('organization', org_type, 'main nav') or _('Organizations')) }}
-                {{ nav.link(group_type ~ '.index', h.humanize_entity_type('group', group_type, 'main nav') or _('Groups')) }}
+                {{ nav.link(dataset_type ~ '.search', h.humanize_entity_type('package', dataset_type, 'main nav') or _('Datasets'), active_blueprints=[dataset_type, dataset_type + '_resource']) }}
+                {{ nav.link(org_type ~ '.index', h.humanize_entity_type('organization', org_type, 'main nav') or _('Organizations'), active_blueprints=[org_type]) }}
+                {{ nav.link(group_type ~ '.index', h.humanize_entity_type('group', group_type, 'main nav') or _('Groups'), active_blueprints=[group_type]) }}
                 {{ nav.link('home.about', _('About')) }}
               {% endblock %}
             </ul>
@@ -115,4 +115,4 @@
       </nav>
     </div>
   </header>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
diff --git a/ckan/templates-midnight-blue/macros/nav_link.html b/ckan/templates-midnight-blue/macros/nav_link.html
index 69b383bb0..2f0838508 100644
--- a/ckan/templates-midnight-blue/macros/nav_link.html
+++ b/ckan/templates-midnight-blue/macros/nav_link.html
@@ -1,18 +1,19 @@
 {#
     Builds links for navigation
-    
+
     nav_item  - the name of the defined nav item defined in config/routing as the named route of the same name
     title      - text used for the link
     icon       - icon name used for link
-    
+
     Example:
-    
+
     {% import 'macros/nav_link.html' as nav %}
     {{ nav.link('home.about', _('About')") }}
-    
+
 #}
-{% macro link(nav_item, title, icon="") %}
-{% set active_page = h.page_is_active(nav_item) %}
+{% macro link(nav_item, title, active_blueprints=[],  icon="") %}
+
+{% set active_page = h.page_is_active(nav_item, active_blueprints=active_blueprints) %}
 
   <li {% if active_page %}class="active"{% endif %}>
     <a {% if active_page %}aria-current="page"{% endif %} href="{{ h.url_for(nav_item) }}">
@@ -22,4 +23,4 @@
       {{ title }}
     </a>
   </li>
-{% endmacro %}
\ No newline at end of file
+{% endmacro %}
```

aleeexgreeen:
@amercader that makes complete sense, and just a heads up the `highlight_controllers` behaviour doesn't work for the activity stream tab of the individual pages (I can look into that at some point).

Two things,
1. The SVG was somehow worse quality wise, so this newest logo is the best we can do quality wise, unless there is a high def logo somewhere?
2. I added the param `url` to the macro. It's for building the tab URLs on the individual pages, because just passing the named route and id isn't working. So for example the groups tabs would be look something like this in `read_base.html`. 
```
{{ nav.link(group_type + '.read', h.humanize_entity_type('package', dataset_type, 'content tab') or _('Datasets'), url=h.url_for(group_type + '.read', id=group_dict.name)), icon='sitemap'}}
{{ nav.link(group_type + '.members', _('Members'), url=h.url_for(group_type + '.members', id=group_dict.name), icon='users') }}
{{ nav.link(group_type + '.about', _('About'), url=h.url_for(group_type + '.about', id=group_dict.name)), icon='info-circle' }}
```
Thoughts?

amercader:
Re the logo don't worry, I'll try to get a better one and we'll change it later on

Re the URL I don't like the two redundant parameters, let's just replace the `nav_item` one with `url` in the `nav.link()` macro:
In header.html:

```
{{ nav.link(h.url_for(dataset_type ~ '.search'), h.humanize_entity_type('package', dataset_type, 'main nav') or _('Datasets'), active_blueprints=[dataset_type, dataset_type + '_resource']) }}
{{ nav.link(h.url_for(org_type ~ '.index'), h.humanize_entity_type('organization', org_type, 'main nav') or _('Organizations'), active_blueprints=[org_type]) }}
{{ nav.link(h.url_for(group_type ~ '.index'), h.humanize_entity_type('group', group_type, 'main nav') or _('Groups'), active_blueprints=[group_type]) }}
{{ nav.link(h.url_for('home.about'), _('About')) }}

```


In individual pages:

```
{{ nav.link(h.url_for(group_type + '.read', id=group_dict.name), h.humanize_entity_type('package', dataset_type, 'content tab') or _('Datasets'), icon='sitemap') }}
{{ nav.link(h.url_for(group_type + '.members', id=group_dict.name), _('Members'), icon='users') }}
{{ nav.link(h.url_for(group_type + '.about', id=group_dict.name), _('About'), icon='info-circle') }}
```

But for that we need to modify our `page_is_active` helper to support both endpoints and URLs:

```diff
diff --git a/ckan/lib/helpers.py b/ckan/lib/helpers.py
index 25afe7e93..ac69e47c3 100644
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -32,7 +32,7 @@ from dominate.util import raw as raw_dom_tags
 from markdown import markdown
 from bleach import clean as bleach_clean, ALLOWED_TAGS, ALLOWED_ATTRIBUTES
 from ckan.common import asbool, config, current_user
-from flask import flash, has_request_context
+from flask import flash, has_request_context, current_app
 from flask import get_flashed_messages as _flask_get_flashed_messages
 from flask import redirect as _flask_redirect
 from flask import url_for as _flask_default_url_for
@@ -739,10 +739,31 @@ def get_flashed_messages(**kwargs: Any):
     return _flask_get_flashed_messages(**kwargs)
 
 
+@core_helper
+def endpoint_from_url(url: str) -> str:
+    try:
+        urls = current_app.url_map.bind("")
+        match = urls.match(url)
+        endpoint = match[0]
+    except RuntimeError:
+        endpoint = ""
+
+    return endpoint
+
+
 @core_helper
 def page_is_active(
         menu_item: str, active_blueprints: Optional[list[str]] = None) -> bool:
-    '''Returns whether the current link is the active page or not'''
+    '''
+        Returns whether the current link is the active page or not. Accepts a
+        route (e.g. 'group.index') or a URL (e.g. '/group')
+        `active_blueprints` contains a list of additional blueprints that should be
+        considered active besides the one in `menu_item`
+    '''
+
+    if menu_item.startswith("/"):
+        menu_item = endpoint_from_url(menu_item)
+
     blueprint, endpoint = menu_item.split('.')
 
     item = {

```

aleeexgreeen:
@amercader Thanks so much!!

amercader:
Great work @aleeexgreeen , looking forward to the next PR

PR #8499, opened 2024-10-30
[8480] fix tracking refer to potentially missing fields

### LINKED ISSUE #8480: Tracking after_dataset_search treats some fields as mandatory
## CKAN version

2.11+

## Describe the bug

The `tracking` extension `after_dataset_search` counts on `id` and `resources` fields of a search result package. Because we have the `fl` param, we can't guarantee that they will exist in a result dict.

### Steps to reproduce
1. Enable `tracking` extension
2. Try to access the `package_search` API endpoint providing the `fl` param that excludes the mentioned fields.
http://127.0.0.1:5000/api/action/package_search?fl=title

### Expected behavior
I'm not sure what is the desired behavior here. I assume we can skip the tracking if we can't identify the dataset. As for resources, just fallback for an empty list.

mutantsan:
Fixes #8480

### Proposed fixes:

Don't try to track the dataset if we can't identify it (no `id`) 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/tracking/plugin.py
+++ b/ckanext/tracking/plugin.py
@@ -36,33 +36,33 @@ class TrackingPlugin(p.SingletonPlugin):
 
     # IPackageController
     def after_dataset_show(
-            self,
-            context: Context,
-            pkg_dict: "dict[str, Any]"
-            ) -> "dict[str, Any]":
+        self, context: Context, pkg_dict: "dict[str, Any]"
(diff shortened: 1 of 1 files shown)


amercader:
This looks good @mutantsan , can you add the changelog entry?
Is this an issue in CKAN 2.10?

Also I'm just curious in what scenario datasets reach this point without an id.

PR #8502, opened 2024-11-01
Fix history views if a resource has been deleted

### LINKED ISSUE #8501: Error on viewing any versions of a deleted resource or its dataset prior to the resource being deleted
## CKAN version
2.11.1

## Describe the bug
If a dataset has ever had a resource deleted from it, trying to view a previous version of the dataset, or that resource, returns an Internal Server Error.

### Steps to reproduce

1. Sign in to CKAN as a sysadmin
2. Create a dataset with a resource
3. Delete the resource
4. Navigate to the dataset's activity stream, http://example.org/dataset/activity/{DATASET_NAME}
5. Click the "View this version" link under the second activity in the list, to see the state of the dataset before the resource was deleted: http://example.org/dataset/{DATASET_ID}/history/{ACTIVITY_ID}
6. Get an Internal Server Error page instead
7. Navigate to the state of the resource at the same point in time: http://example.org/dataset/{DATASET_ID}/resources/{RESOURCE_ID}/history/{ACTIVITY_ID}
8. Again, get an Internal Server Error page instead

### Expected behavior
The previous states of both dataset and resource should be viewable, as the states of their metadata are saved in each change.

### Additional details
The error seems to arise because in both the `resource_history` view and the `package_history` view, we check whether each resource has any resource views by calling `resource_view_list` with its id. If a resource has been deleted, `resource_view_list` raises a `ckan.logic.NotFound` error, which is uncaught.

bellisk:
Fixes #8501

### Proposed fixes:
Catch `ckan.logic.NotFound` errors that are raised when trying to list the resource views for a deleted resource. This fixes the Internal Server Error response when trying to view the history of a deleted resource or its package prior to the deletion.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/activity/tests/test_views.py
+++ b/ckanext/activity/tests/test_views.py
@@ -716,6 +716,95 @@ class TestPackage:
             status=404,
         )
 
+    def test_read_dataset_as_it_used_to_be_after_deleting_resource(self, app):
+        user = factories.User()
+        dataset = factories.Dataset(title="Dataset title")
+        resource = factories.Resource(package_id=dataset["id"])
+        activity = (
+            model.Session.query(Activity)
+            .filter_by(object_id=dataset["id"])
+            .one()
+        )
+
+        helpers.call_action(
+            "resource_delete",
+            context={"user": user["name"]},
+            id=resource["id"],
+        )
+
+        sysadmin = factories.SysadminWithToken()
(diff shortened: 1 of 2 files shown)


bellisk:
Ready for review.

tino097 (approved):

tino097:
Thanks @bellisk

PR #8503, opened 2024-11-03
[UI revamp] Dataset create page

steveoni:
Fixes #

### Features:
Implement UI for dataset page based on https://www.figma.com/design/oI3wDSdZ0cHZLh6CrfYMnb/CKAN-3.0-Design-System?node-id=1887-11502&node-type=INSTANCE&m=dev

- [ ] includes API changes


wardi:
I'm looking at the new dataset editing step navigation and I find it a bit confusing.

Screenshot from 2024-11-12 16-02-15

What are the green check marks and what do they mean? Should the current active step be indicated somehow, like with a different color and an arrow sort of like the current resource-editing navigation?

The preview step seems like a simplified version of the normal "show" page but with a "Submit Data" button to mark the package as active.

What is the Drafts section for? "Save to drafts" makes no change and navigates to the normal "show" page.

There's also no way to navigate directly to any of the resources except for the last one from this new UI, you need to navigate to the normal "show" page and use an "edit resource" link from the drop-down.

Why don't we:
1. remove the unused drafts drop-down
2. show all created resources on the left like we do when editing active dataset resources so users can navigate between them
3. replace the preview step in the left pane with "preview" and "publish" buttons. The "preview" button would navigate to the normal "show" page.
4. use this same UI for editing published resources so that the look of everything doesn't change suddenly after publishing. The "publish" button could be hidden or replaced with a "unpublish" button to return to a draft state.

How does that sound?

wardi:
Yes I think both the "Step 3 Preview" and "Next: Preview Dataset" parts of the UI change should be reconsidered.

The "Preview" page AFAICT is a modified version of the normal show page but it's missing the functionality of the show page and isn't an actual preview of how the dataset will appear. Users would be better served by navigating to the normal show page if they want a preview.

steveoni:
let me know if this okay:

image


image

aleeexgreeen:
Hi @steveoni, it’s looking good! I noticed there were no `.scss` files in your “files changed”, can you add your css changes to the scss files? If I or anyone else were to rebuild the scss, all your changes would be lost. If you pull the header changes from master, all the brand colours and greys are defined.

In terms of UI, can we remove the white background of `accordion-item` and revert the “upload” and “link” buttons back to `btn-default` and change the outline colour? For the select2 dropdowns, when they are active the input area is grey, can we change that to white? I also noticed opening an accordion, closes another one, can we make it so that the accordions are independent to one another?

steveoni:
> Hi @steveoni, it’s looking good! I noticed there were no `.scss` files in your “files changed”, can you add your css changes to the scss files? If I or anyone else were to rebuild the scss, all your changes would be lost. If you pull the header changes from master, all the brand colours and greys are defined.
> 
> In terms of UI, can we remove the white background of `accordion-item` and revert the “upload” and “link” buttons back to `btn-default` and change the outline colour? For the select2 dropdowns, when they are active the input area is grey, can we change that to white? I also noticed opening an accordion, closes another one, can we make it so that the accordions are independent to one another?

this are now resolved

ckan/templates-midnight-blue/package/base_form_page.html
```diff
@@ -15,22 +34,79 @@
 
 {% block secondary_content %}
   {% block info_module %}
-    <section class="module module-narrow module-shallow">
```
aleeexgreeen, line 18: The `module module-narrow module-shallow` classes provided the padding/margins for mobile, we should add this back, as there is no longer any padding on mobile.

ckan/public-midnight-blue/base/scss/_toolbar.scss
```diff
@@ -34,7 +37,7 @@
 }
 
 .toolbar .breadcrumb .active {
-    color: $breadcrumb-active-color;
```
aleeexgreeen, line 37: This is defined in `_variables.scss` we can change it's value to $primary and use it as the color in `.toolbar .breadcrumb .active a, .toolbar .breadcrumb a.active` and remove `.toolbar .breadcrumb .active` because it doesn't do anything currently.

ckan/public-midnight-blue/base/scss/_module.scss
```diff
@@ -22,8 +28,6 @@
 }
 
 .module-content {
-    padding: $grid-gutter-width;
```
aleeexgreeen, line 25: Can we add this back? This is needed around the site.

aleeexgreeen (commented):
Thanks for making the changes @steveoni! I just have a couple notes on the scss. Also, wherever possible we should use the defined variables in `_bootstrap-variables.scss` and `_variables.scss` so the UI is as consistent as possible (font-weights, border-radius, colors etc..)

steveoni:
@aleeexgreeen now resolved

wardi:
Thanks @steveoni this looks much better.

I did notice that the preview button isn't working though, it's acting the same as the publish button (setting state=active)

Why is the list of resources created collapsed? Wouldn't it be better to see them as you add them?

The edit screens (when state=active) still look different than these screens (when state=draft), why not align them so that when creating datasets we can create views and fill out data dictionaries like we can after a dataset is published? I don't think there's value in having a different look and feel when creating vs. when editing.

In the resource view there seems to be a CSS issue with the selected resource (may not be related to this change):
Screenshot from 2025-01-07 16-23-55

This side bar with a list of resources and quick ways to jump to the edit, view or data dictionary for each should be carried through all the dataset pages except for the main dataset view/preview page.

We could add a link to the dataset page and the edit dataset page (with another little wrench icon) above the list of resources so that we consistent navigation throughout all the pages for working with a dataset. I think this would be better for usability and less confusing. Something like:

Screenshot from 2025-01-07 16-41-38

WDYT?

steveoni:
> Thanks @steveoni this looks much better.
> 
> I did notice that the preview button isn't working though, it's acting the same as the publish button (setting state=active)
> 
> Why is the list of resources created collapsed? Wouldn't it be better to see them as you add them?
> 
> The edit screens (when state=active) still look different than these screens (when state=draft), why not align them so that when creating datasets we can create views and fill out data dictionaries like we can after a dataset is published? I don't think there's value in having a different look and feel when creating vs. when editing.
> 
> In the resource view there seems to be a CSS issue with the selected resource (may not be related to this change): Screenshot from 2025-01-07 16-23-55
> 
> This side bar with a list of resources and quick ways to jump to the edit, view or data dictionary for each should be carried through all the dataset pages except for the main dataset view/preview page.
> 
> We could add a link to the dataset page and the edit dataset page (with another little wrench icon) above the list of resources so that we consistent navigation throughout all the pages for working with a dataset. I think this would be better for usability and less confusing. Something like:
> 
> Screenshot from 2025-01-07 16-41-38
> 
> WDYT?

i will check the preview dataset, it's meant to redirect to the dataset read page.

and for the edit; I only made this pr mainly based on the Dataset Create page. But if you want the dataset edit page to also be modified under this pr, that's okay. i can go ahead if that's the case

steveoni:
The resource view page updated
image

Edit package page

image

So I'm unsure if I should edit the existing content of the `Dataset` accordion section: maybe the `Edit metadata` and `Reorder resources`  tab should be moved into the accordion, and where do we place the follow button?

For the **Preview button** I tested it again and it works how it should.

Also, some of these components carry over to the dataset view page, so I added some updates over there

image

wardi:
It looks like there's no way to publish a dataset *after un-publishing* it?
Screenshot from 2025-02-24 11-21-02
the "Publish" button no longer appears at the bottom of any resource.

The edit resource view looks quite different after a dataset has been published:
Screenshot from 2025-02-24 11-29-17

The published resource edit page has a clear dark background color showing which resource is being edited, and the unpublished one has easy navigation back to the dataset editing form.

How about adding that clear indication of which resource is being edited and the easy navigation back to editing the dataset to both versions?

steveoni:
unpublish button fixed- style updated for active resource

image

And here is the edit page. Let me know what to add or remove. Compared to the unpublish page, this page is bound to be different in terms of the action button. For example, an 'Unpublish' button here wouldn't make sense, as we can't determine whether it applies to the resource or the entire dataset.

image

wardi:
@steveoni please take a look at the failing tests and lint errors, and add a changelog file.

steveoni:
> @steveoni please take a look at the failing tests and lint errors, and add a changelog file.

these are now resolved

wardi:
I understand this is holding up other work so I will merge as-is, but I still see some issues that need to be fixed in future PRs:

1. When creating a dataset the following text is shown, but this isn't accurate. Resources are shown here once they are created while the dataset is still in draft mode.
Screenshot from 2025-03-25 19-33-17

2. If a resource is created without a name it appears as a collapsed bubble instead of the normal "Unnamed resource" text. This appeas to get fixed only after the dataset is published.
Screenshot from 2025-03-25 19-33-47

The collapsed resource is overlapped by following resources that are created:
Screenshot from 2025-03-25 19-34-29

3. Clicking preview on the resource page still publishes the dataset (removes the draft state).

PR #8516, opened 2024-11-07
[Backport dev-v2.11] Tracking extension fix for 2.11+

ckanbot:
This is an automated backport pull request 🏗️.

### Details
|     |     |
| --- | --- |
| Original pull request | #8479 (Tracking extension fix for 2.11+) |
| Original author | @tino097 |
| Target branch | **dev-v2.11** |

Please make sure that all relevant checks pass before merging it.

PR #8523, opened 2024-11-11
Bugfix #5468 page view tracking sub path issue 

### LINKED ISSUE #5468: Page view tracking of datasets is not working if ckan is running at a subpath

Nisha129312:
Fixes #5468 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/tracking/cli/tracking.py
+++ b/ckanext/tracking/cli/tracking.py
@@ -5,6 +5,7 @@ import csv
 
 from typing import NamedTuple, Optional
 
+import re
 import click
 import sqlalchemy as sa
 from sqlalchemy import desc, func, select, cast
@@ -14,6 +15,7 @@ from ckan.model import Package
 from ckan.model.meta import Session as session
 import ckan.logic as logic
 from ckan.cli import error_shout
+from ckan.common import config
 
 from ckanext.tracking.model import (TrackingSummary as ts,
                                     TrackingRaw as tr)
@@ -142,10 +144,12 @@ def update_tracking(summary_date: datetime.datetime):
     Update the tracking_summary table with data from tracking_raw
     '''
     package_url = "/dataset/"
+    root_path = config.get('ckan.root_path')
+    root_path = re.sub('/{{LANG}}', '', root_path)
     session.query(ts).filter(ts.tracking_date == summary_date).delete()
     tracking_tmp = (
         session.query(
-            tr.url,
+            func.replace(tr.url, root_path, '').label("tracking_url"),
             tr.user_key,
             cast(tr.access_timestamp, sa.Date)
             .label("tracking_date"),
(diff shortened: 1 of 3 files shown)


ckanext/tracking/cli/tracking.py
```diff
@@ -142,10 +144,12 @@ def update_tracking(summary_date: datetime.datetime):
     Update the tracking_summary table with data from tracking_raw
     '''
     package_url = "/dataset/"
+    root_path = config.get('ckan.root_path')
+    root_path = re.sub('/{{LANG}}', '', root_path)
```
tino097, line 148: hey @Nisha129312 
can we check for the `ckan.root_path` if it was set and in that case to do all replacing ? If we dont have `root_path` set we are doing unnecessary replace in later in the code

Nisha129312: @tino097 I have done the changes as per your comment. Can you please now review it.

tino097: Maybe you could try something like this 
```
rp = config.get('ckan.root_path', '')
root_path = re.sub('/{{LANG}}', '', rp) if rp else ''
url = (func.replace(tr.url, root_path, '').label("tracking_url") if root_path else tr.url)
```
and use `url` in the query

tino097:
We need to backport https://github.com/ckan/ckan/pull/8516 first 

@pdelboca

Artlaninc89:
> Fixes #5468 
> 
> 
> 
> ### Proposed fixes:
> 
> 
> 
> 
> 
> 
> 
> ### Features:
> 
> 
> 
> - [X] includes tests covering changes
> 
> - [X] includes updated documentation
> 
> - [X] includes user-visible changes
> 
> - [X] includes API changes
> 
> - [X] includes bugfix for possible backport
> 
> 
> 
> Please [X] all the boxes above that apply
> 
>

Artlaninc89:
> Fixes #5468 
> 
> 
> 
> ### Proposed fixes:
> 
> 
> 
> 
> 
> 
> 
> ### Features:
> 
> 
> 
> - [X] includes tests covering changes
> 
> - [X] includes updated documentation
> 
> - [X] includes user-visible changes
> 
> - [X] includes API changes
> 
> - [X] includes bugfix for possible backport
> 
> 
> 
> Please [X] all the boxes above that apply
> 
>

PR #8530, opened 2024-11-14
configurable datastore FTS indexes

### LINKED ISSUE #5847: Overly aggressive indexing strategy greatly increases datastore storage requirements

wardi:
Fixes #5847

### Proposed fixes:
Disable automatic FTS indexes of text fields in the datastore through a new configuration option

### Features:

- [ ] includes API changes


amercader:
This looks good @wardi 

IIRC there needs to be a similar PR for backporting to 2.11 (and 2.10?) with the default value `ckan.datastore.default_fts_index_field_types` inverted to keep compatibility right?

wardi:
@amercader yes, I will create those separately. I also want to have a cli command to update the FTS indexes based on the configuration option so users have an easy way to reclaim space on an existing installation.

amercader:
@wardi is the CLI command meant to be in this PR or can I merge it?

wardi:
bumped back to draft so I can finish the CLI

wardi:
@amercader ready for review again, will work on 2.11 backport

ckanext/datastore/cli.py
```diff
@@ -236,5 +218,33 @@ def upgrade():
...
+def fts_index(timeout: float):
+    '''Use to create or remove full-text search indexes after changing
+    the ckan.datastore.default_fts_index_field_types setting.
+    '''
+    site_user = logic.get_action('get_site_user')({'ignore_auth': True}, {})
+    resource_ids = get_all_resources_ids_in_datastore()
+
+    for i, resid in enumerate(get_all_resources_ids_in_datastore(), 1):
+        print(f'\r{resid} [{i}/{len(resource_ids)}] ...', end='')
+        logic.get_action('datastore_create')(
```
amercader, line 241: We don't need to do anything now but it feels a bit uneasy to call `datastore_create` without params in order to trigger the indexes updating, as this will also do anything that a `datastore_create` call does. Maybe now that is just the indexes but who knows if something else is added in the future. A specific `datastore_update_indexes` call would be safer but again, we can improve this in the future.

wardi: I was also uneasy with this, except that creating indexes is part of what `datastore_create` does automatically, so I don't know how you would move that functionality to a separate call without breaking the API.

PR #8533, opened 2024-11-15
Fix translation issue for new field in resource extra information

### LINKED ISSUE #8532: CKAN 2.10.5: Additional Information Field Names on Resource Page Not Translated
## CKAN version

2.10.5

## Describe the bug

I developed an extension plugin and added a new field to the "Additional Information" section on the resource page in CKAN. However, even after changing the language, the field name in the "Additional Information" section does not get translated.

追加情報

### Steps to reproduce

１．Use the `before_resource_show` function in the `plugin.py` file to add a new field to the "Additional Information" section on the resource page:
```
def before_resource_show(self, resource_dict: dict[str, Any]):
        resource_dict['Downloads'] = (
                get_resource_download_count()
        )

        return resource_dict
```

２．Implement Japanese translations according to the [CKAN i18n documentation](https://docs.ckan.org/en/2.10/contributing/i18n.html?highlight=babel).
The `po` file created for translation is as follows:
```
msgid "Downloads"
msgstr "ダウンロード数"
```

３．Switch to Japanese on the resource screen using the `language selector`(language_selector.html).

４．Check if the field name in the "Additional Information" section changes from "Downloads" to "ダウンロード数."

### Expected behavior

The field names in the "Additional Information" section on the resource page should be translated according to the selected language.

### Additional details

In the "Additional Information" section of the dataset (package) page, field names appear to be translated using `_(...)`, as shown below:
```
<th scope="row" class="dataset-label" property="rdfs:label">{{ _(key|e) }}</th>
```
However, it seems that the translation handling for the field names in the "Additional Information" section on the resource page is not implemented, as shown below:
```
<th scope="row">{{ key | capitalize }}</th>
```
Shouldn’t the field names in the "Additional Information" section of the resource page support translation in the same way as the dataset (package) page?

TakumiMiyazawaAbel:
Fixes #8532

### Proposed fixes:

I will modify the template to ensure that the keys in the additional information on the resource screen are translated.
`ckan/templates/package/resource_read.html`

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/package/resource_read.html
+++ b/ckan/templates/package/resource_read.html
@@ -169,7 +169,7 @@
               </tr>
               {% for key, value in h.format_resource_items(res.items()) %}
                 {% if key not in ('created', 'metadata modified', 'last modified', 'format') %}
-                  <tr class="toggle-more"><th scope="row">{{ key | capitalize }}</th><td>{{ value }}</td></tr>
+                  <tr class="toggle-more"><th scope="row">{{ _(key | capitalize) }}</th><td>{{ value }}</td></tr>
                 {% endif %}
               {% endfor %}
             </tbody>
--- a/changes/8532.bugfix
+++ b/changes/8532.bugfix
@@ -1,0 +1,1 @@
+Modified the extension so that the key of the new field added to the resource's additional information is translated according to the selected language.


ckan/templates/package/resource_read.html
```diff
@@ -169,7 +169,7 @@ <h2>{{ _('Additional Information') }}</h2>
               </tr>
               {% for key, value in h.format_resource_items(res.items()) %}
                 {% if key not in ('created', 'metadata modified', 'last modified', 'format') %}
-                  <tr class="toggle-more"><th scope="row">{{ key | capitalize }}</th><td>{{ value }}</td></tr>
+                  <tr class="toggle-more"><th scope="row">{{ _(key | capitalize) }}</th><td>{{ value }}</td></tr>
```
wardi, line 172: Resources can have arbitrarily-named extra fields added by users but translations are only added by site operators or administrators (it using the multilingual plugin). This could lead to surprising behavior if a user-provided extra field happened to match a translated English term.

Instead of a change like this in core consider overriding the `{% block resource_additional_information_inner %}` in the same extension where you add the download count to display the downloads label correctly.

wardi: Another option is to use an extension like ckanext-scheming to define your custom resource field. ckanext-scheming handles overriding the resource information and showing labels in the correct language, you could use an `output_validator` to generate the downloads value instead of `before_resource_show` which would let the value be cached which should be much faster.

TakumiMiyazawaAbel: > Resources can have arbitrarily-named extra fields added by users but translations are only added by site operators or administrators (it using the multilingual plugin). This could lead to surprising behavior if a user-provided extra field happened to match a translated English term.
> 
> Instead of a change like this in core consider overriding the `{% block resource_additional_information_inner %}` in the same extension where you add the download count to display the downloads label correctly.

Thank you for pointing that out. Could this issue also occur on the dataset page? On the dataset page, we have already implemented similar fixes as described in this PR.

Would it be possible to revise the approach to apply the same handling as on the dataset page?

wardi: Yes ckanext-scheming can also be used to define custom dataset fields with labels translated into multiple languages.

I see that package extras *do* have their keys sent to gettext but unlike resource extras are not capitalized and are escaped first, which is even more inconsistent. My preference would be to remove the inconsistent behavior from both and perhaps give users a helper to format extra key names so that themes can override it however they see fit.

@amercader WDYT? Do you know if we're sending user data through gettext in other places (maybe this is intended behavior when using the multilingual plugin?)

amercader: @wardi yes, this makes sense also Iooks like there is some broken `resource` property being set as well:

https://github.com/ckan/ckan/blob/328bf9e6146bf00fdca477127d47fe95c1e630fb/ckan/templates/package/snippets/additional_info.html#L85

TakumiMiyazawaAbel: @wardi @amercader 
I understand that this PR addresses an important fix for translation issues in resource extra fields, but it seems that the discussion has stalled recently. Could you kindly provide an update on the current status or the next steps planned for this PR? Additionally, if further discussion is needed regarding how this PR impacts dataset pages or other related functionalities, please let me know. Thank you for your time and support.

PR #8538, opened 2024-11-18
8537: Only show member_count in a group list item if we have that value

### LINKED ISSUE #8537: All groups in dataset groups tab show "0 members"
## CKAN version
2.11.0

## Describe the bug
On the Groups tab of a dataset page, all the groups that are listed for the dataset are shown as having 0 members, even when this is not the case. This is the same whether the user is logged in or not.

### Steps to reproduce
1. Go to any dataset page and click on the Groups tab.
2. If the dataset does not belong to any groups, add it to some.
3. See that all the group tiles on the page include the text `0 Members`.

### Expected behavior
The group tiles should show the correct number of members for each group. Alternatively, it would be better not to show any member information than to show false information.

### Additional details
groups

bellisk:
Fixes #8537 

### Proposed fixes:

When displaying a group list on the group search page, or on a user's page, we have the member_count for each group (as of https://github.com/ckan/ckan/pull/7007). On the groups tab of a dataset page, we don't get the member_count for the groups. In this case, we should rather display nothing than '0 Members'.

Getting the member count for each group and displaying the right number seemed expensive for little utility.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8537.bugfix
+++ b/changes/8537.bugfix
@@ -1,0 +1,1 @@
+Fix showing '0 members' for all groups on a dataset page.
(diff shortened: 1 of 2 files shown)


tino097:
Thanks @bellisk

PR #8559, opened 2024-12-02
Fixes in the PyPI publish workflow

amercader:
I separated the PyPI and Test PyPI workflows in two separate files (and also two different GitHub environments) for ease of configuration and security. 
Added better event checks (on tag push and push to master) 

The publishing to Test PyPI on each merge requires another step to add a timestamp to the version number otherwise the upload fails because of the existing version.


kowh-ai (approved):
Looks good @amercader however tbh my knowledge here is limited

Issue #7722, opened 2023-07-31
Updating an organization's data doesn't update the the organization data for its packages


bellisk:
## CKAN version
2.8

## Describe the bug
After updating an organization, the organization's packages are not updated, meaning that the data in the packages' `organization` field is not up-to-date.

### Steps to reproduce
- Edit an organization and change data in one field, e.g. description.
- Get the organization's packages from the API: `/api/3/action/package_search?fq=organization:{ORG_NAME}`
- Check the field `organization` for any of the packages and see that it contains old data, not the new change

### Expected behavior
I expect the organization data on datasets to reflect the current state of the organization that owns the datasets.

### Additional details
Looking in the code, I see that this behaviour seems to have been deliberately added back in 2012. With a remove of 11 years, I can't make out the rationale for the change, and I wonder if it's still relevant or if this is something we could change.

In the function `group_dict_save`, the constant `prevent_packages_update` was added with the following comment. I'm not sure what the 'new org rules' were in 2012 or why we should avoid updating packages with `group_edit` for organizations.

https://github.com/ckan/ckan/blob/9956f92925740957b119bd7c6719ee202eb51082/ckan/lib/dictization/model_save.py#L416-L426

In `_group_or_org_save`, we explicity set `prevent_packages_update` to `False` for organizations, as described in the the above snippet.

https://github.com/ckan/ckan/blob/9956f92925740957b119bd7c6719ee202eb51082/ckan/logic/action/update.py#L696-L699

I would prefer to change this behaviour to match my expectations, but I'll be happy if someone can explain to me why we do it this way. :)

fdgneto:
Hi @bellisk, I'm using CKAN 2.10 and I'm having the same issue, have you found a solution?

wardi:
This is a duplicate of #5842 see some of the workarounds discussed there

Issue #5842, opened 2021-01-20
Changing organisation name via API does not update Solr, breaks pages


ThrawnCA:
**CKAN version**

2.8.4

**Describe the bug**

Updating an organisation name via the `organization_update` API does not updated the associated Solr records for the organisation's datasets, which in turn causes errors (in `snippets/organization.html`) when viewing those dataset pages.

**Steps to reproduce**

- Create an organisation called "foo".
- Create a dataset in the organisation named "baz".
- Use the `organization_update` API to alter the organisation name to "meh".
- Visit `/dataset/baz`

**Expected behavior**
The dataset should be displayed with the new organisation name.

**Additional details**

Running a Solr index rebuild on the affected dataset(s) resolves the error.

ThrawnCA:
Triggering an index update for every affected dataset could be quite a bit of work, of course, if an organisation contains many datasets. A cheaper alternative would be to update `snippets/organization.html` so it gracefully handles the renaming. Currently, it seems to be looking up the organisation associated with a package by using the organisation name, not UUID - so when the name changes, it breaks.

ThrawnCA:
It appears that this does not always affect every dataset in the organisation; some will work while others will break. I haven't yet identified a pattern, although private datasets seem more likely to be affected (but public datasets _can_ break too).

duttonw:
I have replicated this on https://demo.ckan.org/ following the steps above

1. original org https://demo.ckan.org/organization/abcdd
2. dataset createdhttps://demo.ckan.org/dataset/tadsf
3. org changed to 123abc including url https://demo.ckan.org/organization/123abc
4. https://demo.ckan.org/dataset/tadsf is now broken

```
Server Error
An internal server error occurred
```

amercader:
This is a long standing issue. Basically if you update an organization's name you need to reindex all of its dataset so the Solr index gets the new value.
As this is potentially a long task depending on the number of datasets it should be implemented as a [background job](https://docs.ckan.org/en/2.9/maintaining/background-tasks.html).

Here's a quick spec in case someone wants to contribute it:

* Create a new `ckan/jobs.py` module (we still don't have builtin jobs), with a function named `update_organization_name_on_datasets` or similar
* This functions takes an organization id as parameter
* The function calls `package_search` with a `fq`: `owner_org:<organization_id>` filter to get all datasets that belong to the org, and then calls `ckan.lib.search.rebuild()` with their ids. Things to check: results pagination, job timeout, etc
[...]

PR #8561, opened 2024-12-02
Reindex Packages UI and After Org/Group Update

JVickery-TBS:
Related issues:

- https://github.com/ckan/ckan/issues/7722
- https://github.com/ckan/ckan/issues/5842

Fixes #

### Proposed fixes:

After updating a group or organization, any datasets that are in the SOLR index will not get the new title or name until a reindex of the package.

This adds in a background job if the org or group title or name change, to reindex all of the records in that org or group. It also adds in some views for admins to enqueue these background jobs, and shows the progress with a new progress bar module.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/search_index.py
+++ b/ckan/cli/search_index.py
@@ -41,13 +41,21 @@ def rebuild(
     u''' Rebuild search index '''
     from ckan.lib.search import rebuild, commit
     try:
-
-        rebuild(package_id,
-                only_missing=only_missing,
-                force=force,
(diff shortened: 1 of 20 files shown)


amercader (commented):
Thanks @JVickery-TBS , I left some initial comments. I'm not 100% convinced with the Group / Org admin UI to reindex datasets. I feel like most users will struggle to understand what is it it does or even find it. But let's keep it for now.

Issue #8427, opened 2024-09-10
Deprecation warnings in bootstrap scss


aleeexgreeen:
## CKAN version
master

## Describe the bug
Deprecation warnings from `ckan/public/base/vendor/bootstrap/scss` appear when building the scss. Deprecations are for declarations using percentages in global `abs()` and for passing a number without a unit. There are about 8 warnings.

### Steps to reproduce
Run `npm run build`

### Additional details
```
Deprecation Warning: Passing percentage units to the global abs() function is deprecated.
In the future, this will emit a CSS abs() function to be resolved by the browser.
To preserve current behavior: math.abs(100%)
To emit a CSS abs() now: abs(#{100%})
More info: https://sass-lang.com/d/abs-percent
   ╷
57 │   $dividend: abs($dividend);
   │              ^^^^^^^^^^^^^^
 ckan/public/base/vendor/bootstrap/scss/vendor/_rfs.scss 57:14         divide()
```

```
Deprecation Warning: $weight: Passing a number without unit % (100) is deprecated.
To preserve current behavior: $weight * 1%
More info: https://sass-lang.com/d/function-units
    ╷
201 │   @return mix(rgba($foreground, 1), $background, opacity($foreground) * 100);
    │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ckan/public/base/vendor/bootstrap/scss/_functions.scss 201:11            opaque()
```

PR #8569, opened 2024-12-08
[Midnight blue] Login Page

aleeexgreeen:
### Proposed fixes:

- New design layout and scss for login page and pages associated with logging in
- Alerts scss
- Adds icon option for text input
- Changes default alert class to danger because of #8427 and #7901
- Adds `role="alert"` to alerts [more details](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)
- Adds boat logo

### Features:

- [ ] includes API changes


amercader:
I understand this is part of the new design, but I'm not convinced about including the CKAN logo in the login/logout pages

* For the standard theme it creates an akward duplication with the header logo
* On most sites, users will want to change the logo the their own organization one. These would add an extra step.

Let's see what others think but I'd vote for removing it. 



Screenshot 2024-12-09 at 12-53-38 Login - CKAN
Screenshot 2024-12-09 at 12-53-06 Logged Out - CKAN

aleeexgreeen:
Hi @amercader ! I have a modified version of the gulpfile in my local, but it only builds the `midnight-blue` theme (I've changed public to public-midnight-blue). It looks like we may have to create separate `build` and `watch` functions for midnight-blue.

> I think this function should be updated to also build the midnight blue theme.

I've tried having them both in the build and watch functions, but without something distinguishing in which theme the files were changed, all css files are updated with the changed code. That was from a quick test though, so I may be wrong!

wardi:
+1 on removing the duplicated logo. Do we need some other artwork like a big key icon or something there?

thegostev:
Voting for removing the logo, leaving one in the header.

PR #8590, opened 2024-12-17
faster datastore_info: include_meta, include_fields_schema parameters

### LINKED ISSUE #8589: datastore_info fast path
## CKAN version
all

## Describe the bug
`datastore_info` unconditionally queries and returns the row count, database size, index size, aliases, db field schema information. These can be expensive to generate and are unused by many callers.

### Steps to reproduce
Call `datastore_info` on a large table or a table with many fields

### Expected behavior
Return field information efficiently

wardi:
Fixes #8589 

### Proposed fixes:
Add new datastore_info parameters:
- include_meta
- include_fields_schema

both default to True for backwards compatibility, but when set to False those sections of the datastore_info response will be left out. For a datastore table with ~11M rows this drops run time from 1.71s to 0.07s, and can be applied everywhere the API is used internally to improve page render times.

### Features:

- [X] includes API changes

### DIFF
--- a/changes/8589.feature
+++ b/changes/8589.feature
@@ -1,0 +1,3 @@
+datastore_info: new include_meta and include_fields_schema parameters that
+may be set to False to avoid computing table size, index size, row count,
+aliases or per-field index, unique, notnull status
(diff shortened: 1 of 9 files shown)


wardi:
Hi @smotornyuk please take a look at this one when you have time

smotornyuk:
Sorry for delay. Looks good, merging it

PR #8597, opened 2024-12-24
datatables: skip extra draw calls on load

### LINKED ISSUE #8595: datatables is view very slow for large tables
## CKAN version
master, 2.11, 2.10

## Describe the bug
datatables [draw](https://datatables.net/reference/api/draw%28%29) calls are used in 7 places in our [datatablesview.js](https://github.com/ckan/ckan/blob/master/ckanext/datatablesview/assets/datatablesview.js) script. Each will cause a new ajax call that runs datastore_search on the back end.

On initial page load this ajax call is run *3 times*, so if a table takes 5s to return datastore_search results the datatable view won't be rendered until more than 15s have passed.

### Steps to reproduce
render a datatable view of a large datastore table

### Expected behavior
ckan should only trigger expensive calls to the datatable ajax back end when necessary.

### Additional details

slow-datatable

wardi:
Fixes #8595

### Proposed fixes:
Skip extra second and third draw call on page load to cut time waiting for duplicated ajax responses.

Was ~17s for view of 11M row table, now ~9s

less-slow-datatables

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datatablesview/assets/datatablesview.js
+++ b/ckanext/datatablesview/assets/datatablesview.js
@@ -114,7 +114,6 @@ function fitColText () {
   if (gcurrentView === 'list') {
     dt.responsive.recalc()
   }
-  dt.columns.adjust().draw(false)
 }
 
 // ensure element id is valid
@@ -576,7 +575,6 @@ this.ckan.module('datatables_view', function (jQuery) {
               }
             }
           })
-          api.draw(false)
         }, // end stateLoadParams
         stateSaveParams: function (settings, data) {
           // this callback is invoked when saving state info


wardi:
@Zharktas I've identified and removed the surrounding code that depended on these extra redraw calls, this is ready for another look.

JVickery-TBS:
@wardi yeah, I think if you upgrade the DataTables vendor version, the auto adjust columns works a lot better. It should never have to redraw just to adjust column sizes. It should only do `table.columns.adjust()` during initialization as I understand it. 

The responsive plugin in a later version is probably what handles the column adjustments better.

PR #8603, opened 2024-12-31
faster datatables with estimated totals

wardi:
Second, complementary change for #8595 

### Proposed fixes:

Use estimated totals when tables are larger than 100000 rows.

Drops default unfiltered ajax call run time on an 11M row dataset from 4000ms to 150ms.

### Features:

- [ ] includes API changes


wardi:
@kowh-ai I've added "(estimated)" text to the total displayed when it was estimated in the data tables view.

wardi:
@kowh-ai tests are green, this is ready for another look

kowh-ai (approved):
AFAICT this looks solid. Happy to merge

PR #8606, opened 2025-01-07
Update tag-name-validation to less restrictive tag handling

splattater:
Update validation to allow more special chars (e.g. punctuation) Add test cases for valid and invalid tags

As discussed in https://github.com/ckan/ckan/issues/2780#issuecomment-225540769 and suggested in
https://github.com/open-data/ckanext-canada/blob/wet4-scheming/ckanext/canada/validators.py#L52-L92
this commit applies the validation for tag names and extends the tests for the new cases.

This was motivated by the possibility to add scales ("1:5000") of spatial datasets as tags.

Fixes #

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -9,6 +9,7 @@ import mimetypes
 import string
 import json
 import uuid
+import unicodedata
 from typing import Any, Container, Optional, Union
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


wardi:
Choosing characters to allow in tags is a very site-specific decision. I don't see a reason to change ckan's default rules for acceptable punctuation, especially because it may be surprising to users when they upgrade.

It's recommended to override the `tag_name_validator` using the IValidators interface from a theme plugin to customize the rules for tags for a particular site.

splattater:
Would it be a better solution to make it a configuration option with the current setting as default setting?

toothstone:
Since I kicked off this PR with my user story, I would like to chime in with some context:
Primary goal here is to achieve DCAT-AP conformity for Open Data CKAN instances when handling keywords aka tags. There are no charset limits on DCAT keywords, so we can't sensibly map them onto the heavily restricted CKAN tags, as this leads to harvesters rejecting valid DCAT entries.

Because tag restrictions have been coming up again before and there seemed to be general consensus/maintainer approval that this should be addressed directly at the CKAN core, this PR was placed here. It is of course also an option to move it (fully or partially) into ckanext-dcat.

Making this configurable is IMHO the obvious solution for backwards compatible behaviour (ckanext-dcat could just turn it on by default then), one more change I would propose to make adoption easier, is to have the new validation strictly **more** accepting than the old, which of course should be ensured via extensive tests and a test-driven commit order. On first glance there seem to be new restrictions regarding e.g. double spaces.

The "treat tags with commas as list" behaviour needs to be removed/disabled for DCAT conformity, so there also won't be a need to disallow commas. In fact, users have stated that this is problematic behaviour: https://github.com/ckan/ckanext-dcat/issues/253 - and patching it out had no adverse effects.

I hope we can get some improvements for the CKAN ecosystem out of this, one way or another :)

wardi:
(adding a comment so I raise this for discussion at the next dev call)

gerbsen:
hey @wardi, thank you for the consideration and raising this issue at your dev meeting. May I ask how it went?

LorenzBuehmann:
Hi @wardi and other devs,

we think we addressed the requested task:

- [x] provide a new configuration option to set the regex applied to validate tags with the default tag_name_validator
  - we added a config option `ckan.validation.tag_name.regex` to set a custom regex
  - if nothing has been set, the default regex is being used
- [x] document this configuration option with an example of how to set it back to the old (current) standard for valid tags
  - we added a corresponding section in the `config_declaration.yaml` file
- [x] new allowed characters should still be somewhat limited e.g. to only printable characters
  - even when a custom regex would allow non-printable chars, a check via `isPrintable()` is done to reject tags contianing such chars
- [x] Unicode normalization should be applied to tags so that e.g. U+00E9 (é) is considered the same as U+0065 U+0301 (e + combining acute accent)
  - is done by applying NFC normalization 
  - we initially put this into the `tag_name_validator` validation method, but apparently to be effective this has been added to the `tag_string_convert` method, as the validator itself indeed won't modify tags but only validate them  ( so when we can assume that `tag_name_validator` is always only called from within `tag_string_convert`, we could remove the redundant normalization call

gerbsen:
hey @wardi, is there anything we should/could do in order to get this merged? cheers, gerbsen

PR #8611, opened 2025-01-14
Replace btn default

Zharktas:
### Proposed fixes:
Bootstrap 4 renamed `.btn-default` to `.btn-secondary`  https://getbootstrap.com/docs/4.6/migration/#buttons

This just renames all of them as it wasn't done when migrating to bs5 as bs4 was skipped.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public-midnight-blue/base/css/main-rtl.css
+++ b/ckan/public-midnight-blue/base/css/main-rtl.css
@@ -3077,7 +3077,7 @@ textarea.form-control-lg {
   --bs-btn-disabled-border-color: #d43f3a;
 }
 
-.btn-light, .btn-default {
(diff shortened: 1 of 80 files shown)


Zharktas:
Failing cypress test doesn't seem to be related to this as it fails on other PRs as well.

amercader:
Is this worth backporting? It does look a bit scary to be honest and I'd imagine that users would have adapted their themes already.
What do you think?

Zharktas:
I was considering backporting it as it's clearly a bug, but fixing it would be a breaking change for extensions that customize the theme and we are trying to avoid those :). In the end, this is a minor issue and backporting this would make only my life easier and not probably worth the effort.

PR #8618, opened 2025-01-17
add 'en_AU' locale to DataTables

ThrawnCA:
### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datatablesview/assets/vendor/DataTables/i18n/en_AU.json
+++ b/ckanext/datatablesview/assets/vendor/DataTables/i18n/en_AU.json
@@ -1,0 +1,178 @@
+{
+    "emptyTable": "No data available in table",
+    "info": "Showing _START_ to _END_ of _TOTAL_ entries",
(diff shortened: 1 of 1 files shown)


tino097:
hey @ThrawnCA , if we update DataTables in future, this will get replaced and lost. Is this fix only for current version and if that's a case, im not sure if we can merge

ThrawnCA:
> hey @ThrawnCA , if we update DataTables in future, this will get replaced and lost.

Where is the right place to submit this, then?

wardi:
@ThrawnCA looks like we're supposed to use a plug-in: https://datatables.net/plug-ins/i18n/

ThrawnCA:
> @ThrawnCA Could we host the json file from the our datatables extension code instead of mixing it with the vendor files, and reference it as a plug-in https://datatables.net/plug-ins/i18n/ ?

I probably know less about it than you. I just dropped a file in alongside the other i18n files, with the same content as the `en_GB` one.

PR #8621, opened 2025-01-18
[Midnight blue] Account header

aleeexgreeen:
### Proposed fixes:
- Account masthead as dropdown split button

[Figma mockup](https://www.figma.com/design/oI3wDSdZ0cHZLh6CrfYMnb/CKAN-3.0-Design-System?node-id=567-3265&t=tsa0sAtl2a0Na8eM-4)

** I have changed a few things from the Figma.

### Features:

- [ ] includes API changes


amercader:
Looks great @aleeexgreeen , some comments:

* Question: what is preventing from using this (much more visually appealing) dropdown style to the [search form order by list](https://github.com/ckan/ckan/pull/8592#discussion_r1914704419)? Is it that the search form one is an actual select input and this is just a ul list?
* Can you fix these minor spacing issues:
   1. User name cut
   2. Increase space between icons and labels
   3. Add a bit of bottom padding to the counter to match the upper one
Screenshot from 2025-01-20 14-52-03
* There is an issue with the visibility of the notifications. If you have pending notifications highlighted you won't know unless you open the account menu (see below). This probably needs to be talked with the designer but would be good to see the notifications indicator without any interaction.

Screenshot from 2025-01-20 14-54-24



Aside: this notification indicator is a perfect candidate for htmx/async refactoring as it's currently done as part of the main request and slows things down a lot if you follow lots of stuff.

aleeexgreeen:
Hi @amercader, I could try changing the select of the search order by to a dropdown. I would just need to test the dropdown using a screen reader. Keyboard wise it seems to be accessible. One thing I noticed for the header dropdown, is that the dashboard list item removes the class from the link and the list item itself gets the class "notifications". Do you know where that code is located? I can't seem to find it. Try using your keyboard to navigate the dropdown options, it skips the dashboard option because of this.

For the username, we could truncate the name and display it in full in the tooltip and potentially add a spot for notifications? We could also expand the width of the button section and reduce the width of the nav bar list? I think the username length will cause issues with this button implementation, what do you think?

amercader:
Hi @aleeexgreeen see below:

> I could try changing the select of the search order by to a dropdown. I would just need to test the dropdown using a screen reader. Keyboard wise it seems to be accessible.

Cool, let's give it a go, it looks much nicer as a dropdown, but I suspect we will have to change the way the form is submitted. If it gets too complicated let's leave it for now

> One thing I noticed for the header dropdown, is that the dashboard list item removes the class from the link and the list item itself gets the class "notifications". Do you know where that code is located? I can't seem to find it. Try using your keyboard to navigate the dropdown options, it skips the dashboard option because of this.

The `activity` plugin is overriding this block [here](https://github.com/ckan/ckan/blob/1da2bc6bb4a08d2ece70a0131ac7cae08a00917f/ckanext/activity/templates/header.html#L5). Feel free to add any class needed as long as it doesn't break the standard theme. If it does we'll need to use a small helper.

> For the username, we could truncate the name and display it in full in the tooltip and potentially add a spot for notifications? We could also expand the width of the button section and reduce the width of the nav bar list? I think the username length will cause issues with this button implementation, what do you think?

I think so. Keep in mind that, if present, we will display the Display Name here, which will probably we even longer. I don't think we should truncate the user name as it is an important piece of information. I like having the user name wrapped in the button but if we can't expand it to make it fit whatever length the name is we should think about a different design

Screenshot 2025-01-30 at 15-24-55 Adrià Mercader - Users - CKAN

Thanks for fixing the spacing issues, the activities count is not properly aligned though:

amercader:
Hi @aleeexgreeen see below:

> I could try changing the select of the search order by to a dropdown. I would just need to test the dropdown using a screen reader. Keyboard wise it seems to be accessible.

Cool, let's give it a go, it looks much nicer as a dropdown, but I suspect we will have to change the way the form is submitted. If it gets too complicated let's leave it for now

> One thing I noticed for the header dropdown, is that the dashboard list item removes the class from the link and the list item itself gets the class "notifications". Do you know where that code is located? I can't seem to find it. Try using your keyboard to navigate the dropdown options, it skips the dashboard option because of this.

The `activity` plugin is overriding this block [here](https://github.com/ckan/ckan/blob/1da2bc6bb4a08d2ece70a0131ac7cae08a00917f/ckanext/activity/templates/header.html#L5). Feel free to add any class needed as long as it doesn't break the standard theme. If it does we'll need to use a small helper.

> For the username, we could truncate the name and display it in full in the tooltip and potentially add a spot for notifications? We could also expand the width of the button section and reduce the width of the nav bar list? I think the username length will cause issues with this button implementation, what do you think?

I think so. Keep in mind that, if present, we will display the Display Name here, which will probably we even longer. I don't think we should truncate the user name as it is an important piece of information. I like having the user name wrapped in the button but if we can't expand it to make it fit whatever length the name is we should think about a different design

Screenshot 2025-01-30 at 15-24-55 Adrià Mercader - Users - CKAN

Thanks for fixing the spacing issues, the activities count is not properly aligned though:

aleeexgreeen:
Okay, I'm going to pause development on this pull request until the designers and I can figure out a better suited alternative. But do not worry, there will be more PRs coming!

PR #8593, opened 2024-12-22
[Midnight blue] Components

aleeexgreeen:
### Proposed fixes:

- Pagination styling
- Popovers styling
- Nav-pills replace nav-tabs on `page.html`
- Minor header fix on tablet
Components not being changed:
- Tooltips
- Toasts
- Modals (for now)

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [X] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #8622, opened 2025-01-20
feat: allow plugins to patch user name

SyeKlu:
This pull request introduces changes that allow plugins to patch the username and email address using external authentication providers. Previously, only system administrators could update the username, but it was not possible to modify it via plugins. This functionality now enables plugins to update both the username and email, allowing for better integration with external authentication systems and more flexible user management.

### DIFF
--- a/changes/8622.feature
+++ b/changes/8622.feature
@@ -1,0 +1,1 @@
+Allows plugins to change the user name during ``user_patch`` or ``user_update`` by passing ``ignore_auth: True`` in the context. ``user_show`` also returns restricted properties like ``email`` or ``apikey`` when passing ``ignore_auth: True`` in the context.
--- a/ckan/tests/logic/action/test_patch.py
+++ b/ckan/tests/logic/action/test_patch.py
@@ -214,6 +214,33 @@ class TestPatch(object):
         assert user2["fullname"] == "Mr. Test User"
         assert user2["about"] == "somethingnew"
 
+    def test_extensions_successful_patch_updating_user_name(self):
+        user = factories.User()
+
+        updated_user = helpers.call_action(
+            "user_patch",
+            context={"user": user["name"], "ignore_auth": True},
+            id=user["id"],
+            name="somethingnew"
+        )
+
+        assert updated_user["name"] == "somethingnew"
+
+        user2 = helpers.call_action("user_show", id=user["id"])
+
+        assert user2["name"] == "somethingnew"
+
+    def test_extensions_failed_patch_updating_user_name(self):
+        user = factories.User()
+
+        with pytest.raises(ValidationError):
+            helpers.call_action(
+                "user_patch",
+                context={"user": user["name"], "ignore_auth": False},
+                id=user["id"],
+                name="somethingnew2"
+            )
+
     def test_package_patch_for_update(self):
 
         dataset = factories.Dataset()
--- a/ckan/tests/logic/action/test_update.py
+++ b/ckan/tests/logic/action/test_update.py
@@ -41,11 +41,12 @@ class TestUpdate(object):
 
         # 1. Setup.
         user = factories.User()
+        context = {"user": user["name"], "ignore_auth": False}
         user["name"] = "updated"
 
         # 2. Make assertions about the return value and/or side-effects.
         with pytest.raises(logic.ValidationError):
-            helpers.call_action("user_update", **user)
+            helpers.call_action("user_update", context=context, **user)
 
     # END-BEFORE
 
@@ -264,22 +265,6 @@ class TestUpdate(object):
 
         updated_user = helpers.call_action("user_update", **params)
         assert "password" not in updated_user
-
-    def test_user_update_does_not_return_apikey(self):
-        """The user dict that user_update returns should not include the user's
-        API key."""
-
-        user = factories.User()
-        params = {
-            "id": user["id"],
-            "fullname": "updated full name",
-            "about": "updated about",
-            "email": user["email"],
-            "password": factories.User.stub().password,
-        }
-
-        updated_user = helpers.call_action("user_update", **params)
-        assert "apikey" not in updated_user
 
     def test_user_update_does_not_return_reset_key(self):
         """The user dict that user_update returns should not include the user's
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -605,7 +605,7 @@ def user_dictize(
         result_dict['apikey'] = apikey
         result_dict['email'] = email
 
-    if authz.is_sysadmin(requester):
+    if authz.is_sysadmin(requester) or context.get("ignore_auth") is True:
         result_dict['apikey'] = apikey
         result_dict['email'] = email
 
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -702,7 +702,8 @@ def user_name_validator(key: FlattenKey, data: FlattenDataDict,
             errors[key].append(_('The username is already associated with another account. Please use a different username.'))
     elif user_obj_from_context:
         requester = context.get('auth_user_obj', None)
-        if requester and authz.is_sysadmin(requester.name):
+        ignore_auth = context.get('ignore_auth', None)
+        if ignore_auth or (requester and authz.is_sysadmin(requester.name)):
             return
         old_user = model.User.get(user_obj_from_context.id)
         if old_user is not None and old_user.state != model.State.PENDING:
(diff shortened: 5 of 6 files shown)


amercader:
@SyeKlu can you merge this change in https://github.com/ckan/ckan/pull/8593 and pack the two changes there?
Also the error looks related so we need to look into that

SyeKlu:
I tested it again, and it works so far. I've added an additional test, which is currently failing intentionally. The issue is that the `model_dictize` function also needs to include an `ignore_auth` parameter. However, the tests are running with `ignore_auth` set in the test setup, which is causing issues in other tests.

The line to be added is:

```python
if authz.is_sysadmin(requester) or context.get('ignore_auth', False):
```

Since PR #8593 was merged last week, how should I proceed from here?

ckan/tests/logic/action/test_patch.py
```diff
@@ -214,6 +214,31 @@ def test_user_patch_updating_single_field(self):
...
+
+        assert user2["name"] == "somethingnew"
+
+    def test_extensions_failed_patch_updating_user_name(self):
+        user = factories.User()
+
+        with pytest.raises(ValidationError):
+            helpers.call_action(
+            "user_patch",
+            context={"user": None, "ignore_auth": True},
```
amercader, line 238: Here we want to test the standard scenario, i.e. that user_patch will fail when trying to update the name (if ignore_auth is missing or False)

```suggestion
            context={"user": None, "ignore_auth": False},
```

ckan/logic/action/update.py
```diff
@@ -949,6 +949,7 @@ def user_update(context: Context, data_dict: DataDict) -> ActionResult.UserUpdat
     include_plugin_extras = False
     if author_obj:
         include_plugin_extras = author_obj.sysadmin and 'plugin_extras' in data
+    print(user, context, include_plugin_extras)
```
amercader, line 952: Please remove

ckan/tests/logic/action/test_update.py
```diff
@@ -41,11 +41,12 @@ def test_user_update_name(self):
...
         user = factories.User()
+        context = {"user": user["name"], "ignore_auth": False}
         user["name"] = "updated"
 
         # 2. Make assertions about the return value and/or side-effects.
         with pytest.raises(logic.ValidationError):
-            helpers.call_action("user_update", **user)
+            helpers.call_action("user_update", context=context ,**user)
 
     # END-BEFORE
```
amercader, line 51: We need a `test_user_update_name_by_extensions()` where the action succeeds by passing `ignore_auth: True` (we need to cover both patch and update)

ckan/tests/logic/action/test_patch.py
```diff
@@ -214,6 +214,31 @@ def test_user_patch_updating_single_field(self):
         assert user2["fullname"] == "Mr. Test User"
         assert user2["about"] == "somethingnew"
 
+    def test_extensions_successfull_patch_updating_user_name(self):
```
amercader, line 217:
```suggestion
    def test_extensions_successful_patch_updating_user_name(self):
```

amercader (commented):
@SyeKlu sorry, I meant #8609 which is now merged.
This looks good, we just need to add the change you suggest to make the `test_extensions_successful_patch_updating_user_name()` test pass. Please use this form as it is safer:

```diff
diff --git a/ckan/lib/dictization/model_dictize.py b/ckan/lib/dictization/model_dictize.py
index f4ac5a08a..e8c9a6a1c 100644
--- a/ckan/lib/dictization/model_dictize.py
+++ b/ckan/lib/dictization/model_dictize.py
@@ -606,7 +606,7 @@ def user_dictize(
         result_dict['apikey'] = apikey
         result_dict['email'] = email
 
-    if authz.is_sysadmin(requester):
+    if authz.is_sysadmin(requester) or context.get("ignore_auth") is True:
         result_dict['apikey'] = apikey
         result_dict['email'] = email
 ```

Also a couple minor changes and an additional test needed.

File: changes/8622.feature [hunk 5]
```diff
@@ -0,0 +1 @@
+Allows plugins to change the user name during user patch
```
amercader, line 1:
```suggestion
Allows plugins to change the user name during ``user_patch`` or ``user_update`` by passing ``ignore_auth: True`` in the context
```

ckan/tests/logic/action/test_update.py
SyeKlu: doesn't a patch always test updates at the same time as it logically creates a user with the patched changes and then transfers them to the update function?

changes/8622.feature
```diff
@@ -0,0 +1 @@
+Allows plugins to change the user name during ``user_patch`` or ``user_update`` by passing ``ignore_auth: True`` in the context
```
amercader, line 1:
```suggestion
Allows plugins to change the user name during ``user_patch`` or ``user_update`` by passing ``ignore_auth: True`` in the context. ``user_show`` also returns restricted properties like ``email`` or ``apikey`` when passing ``ignore_auth: True`` in the context.
```

amercader:
@SyeKlu [Test failures](https://app.circleci.com/pipelines/github/ckan/ckan/7794/workflows/14bd7e62-0ae8-49fe-8b1f-b42ab3a94f57/jobs/15817/tests) are related to these changes. As `helpers.call_action()` sets `ignore_auth: True` under the hood, there were some tests that checked the return of restricted properties that need to be updated (to make sure they use `ignore_auth: False`. 
- [ ] Apply the diff below to fix the circle CI tests and the linting errors
- [ ] Merge latest master to fix the type checking errors
- [ ] Add the changelog change I suggested, I think this needs to be mentioned

Thanks

---
```diff

diff --git a/ckan/tests/logic/action/test_get.py b/ckan/tests/logic/action/test_get.py
index d384a5b0d..912555bde 100644
--- a/ckan/tests/logic/action/test_get.py
+++ b/ckan/tests/logic/action/test_get.py
@@ -872,7 +872,12 @@ class TestUserList(object):
     def test_user_list_default_values(self):
         user = factories.User()
 
-        got_users = helpers.call_action("user_list")
+        got_users = helpers.call_action(
+            "user_list",
+            # call_action will set ignore_auth: True by default, which will
+            # include restricted fields like apikey
+            context={"ignore_auth": False}
+        )
 
         assert len(got_users) == 1
         got_user = got_users[0]
@@ -1049,7 +1054,13 @@ class TestUserShow(object):
 
         user = factories.User()
 
-        got_user = helpers.call_action("user_show", id=user["id"])
+        got_user = helpers.call_action(
+            "user_show",
+            id=user["id"],
+            # call_action will set ignore_auth: True by default, which will
+            # include restricted fields like apikey
+            context={"ignore_auth": False}
+        )
 
         assert got_user["id"] == user["id"]
         assert got_user["name"] == user["name"]
@@ -1071,7 +1082,14 @@ class TestUserShow(object):
         user = factories.User()
 
         got_user = helpers.call_action(
-            "user_show", context={"keep_email": True}, id=user["id"]
+            "user_show",
+            id=user["id"],
+            # call_action will set ignore_auth: True by default, which will
+            # include restricted fields like apikey
+            context={
+                "keep_email": True,
+                "ignore_auth": False
+            }
         )
 
         assert got_user["email"] == user["email"]
@@ -1084,7 +1102,14 @@ class TestUserShow(object):
         user = factories.User()
 
         got_user = helpers.call_action(
-            "user_show", context={"keep_apikey": True}, id=user["id"]
+            "user_show",
+            id=user["id"],
+            # call_action will set ignore_auth: True by default, which will
+            # include restricted fields like apikey
+            context={
+                "keep_apikey": True,
+                "ignore_auth": False
+            }
         )
 
         assert "email" not in got_user
diff --git a/ckan/tests/logic/action/test_patch.py b/ckan/tests/logic/action/test_patch.py
index f95549cdf..90d3844aa 100644
--- a/ckan/tests/logic/action/test_patch.py
+++ b/ckan/tests/logic/action/test_patch.py
@@ -219,7 +219,7 @@ class TestPatch(object):
 
         updated_user = helpers.call_action(
             "user_patch",
-            context={"user": None, "ignore_auth": True},
+            context={"user": user["name"], "ignore_auth": True},
             id=user["id"],
             name="somethingnew"
         )
@@ -235,11 +235,11 @@ class TestPatch(object):
 
         with pytest.raises(ValidationError):
             helpers.call_action(
-            "user_patch",
-            context={"user": None, "ignore_auth": False},
-            id=user["id"],
-            name="somethingnew"
-        )
+                "user_patch",
+                context={"user": user["name"], "ignore_auth": False},
+                id=user["id"],
+                name="somethingnew2"
+            )
 
     def test_package_patch_for_update(self):
 ``

SyeKlu:
@amercader i’ve made all the required changes. Unfortunately, I had to remove the test `test_user_update_does_not_return_apikey`, as it was based on the assumption that `ignore_auth` would not be triggered. Since the API key is now returned when queried by the user, sysadmins, or with `ignore_auth`, there's no way to trigger the test without the key being included. This is because one of these three conditions must be met to execute the user update.

PR #8625, opened 2025-01-24
remove ajax_snippets feature

### LINKED ISSUE #7619: Let's stop doing ajax_snippets
We've seen @pdelboca 's demo of HTMX that renders partial pages using our existing views with AJAX. HTMX has huge promise for our site performance.

ajax_snippets isn't that.

Files in `templates/ajax_snippets` are served by ckan without auth checks or parameter validation. All url parameters are passed unchecked through to the template found.

Let's stop doing ajax_snippets.

The users of ajax_snippets I see are:
- datastore api documentation pop-up (create a simple view instead)
- follow button that calls a helper that acts like a view which renders a different template (convert the view-like-helper into a view instead)
- JS test-supporting templates (use webassets to serve as static content?)

wardi:
Fixes #7619 

### Proposed fixes:
- remove `/api/1/snippet/...` a.k.a. `templates/ajax_snippets` a.k.a. `getTemplate()` a.k.a. `loadFixture()` feature for the reasons discussed in #7619

### Features:

- [X] includes API changes


ckan/views/util.py
```diff
@@ -29,9 +29,17 @@ def primer() -> str:
...
+        {'extras': [{'key': 'key', 'value': 'value'}], 'errors': {}}
+    )
 
 
 util.add_url_rule(
-    u'/util/redirect', view_func=internal_redirect, methods=(u'GET', u'POST',))
-util.add_url_rule(u'/testing/primer', view_func=primer)
+    '/util/redirect', view_func=internal_redirect, methods=('GET', 'POST',))
+util.add_url_rule('/testing/primer', view_func=primer)
+util.add_url_rule('/testing/custom_form_fields', view_func=custom_form_fields)
```
amercader, line 45: Why do we need this endpoint? it just shows the unstyled template. Isn't it better to include `custom_form_fields.html` in the `primer.html` template so they get displayed properly?

amercader (commented):
This looks good. Only one minor comment

ckan/views/util.py
wardi: AFAICT it's only used by `cypress/e2e/modules/custom-fields.cy.js`

wardi: @amercader it's just the way these tests were written. They load the site homepage then request and render a snippet into a new div in the body. For some reason.

I've tried to leave them mostly the way they were but I replaced the bit that calls `loadFixture` with code that just makes a get request and returns the content.

wardi:
Note: I was able to re-enable one previously disabled flaky dashboard test with a fun hack I found on cypress.io https://github.com/ckan/ckan/pull/8625/commits/5718c1e5434c1886f0b041efe64825b7988be007#diff-65a555a4cad92f84456a1e40b0835620d9686293c541c9d341cd146e5a44c059

Not really happy with adding a busy loop to the tests when it looks like there should be a much simpler way, but this might be a better fix for those other flaky tests too.

wardi:
@amercader I managed to find a simpler way. This latest replaces a buggy wrapper of a wrapper of a bad API around a get-request wrapper with native cypress get requests, and solves the test race condition at the same time.

PR #8626, opened 2025-01-27
feat: Case-insensitive login by email / Unique emails across all user states

smotornyuk:
This PR changes two places in the user's email-related logic.

The first suggestion: `User.by_email` should use a case-insensitive email search. 
In this way, the user can authenticate using any letter case for an email address, even if it is not the same as one used during registration. It isn't a common problem, but recently I heard a complaint from `Sergey@Gmail.Com` that he cannot log in as `sergey@gmail.com`. It does not look like a security measure and we can allow it.

Second suggestion: make states of users, that are checked for email uniqueness, configurable. Right now only active users are checked when a new email is registered in the system, which is intentional(to allow reusing emails of deleted user accounts). But it causes the problem if:

* you want to delete an account and prevent user from registering a new account with the same email
* you have something similar to the request-an-account feature, where users are created in a pending state initially and must be approved by an authorized person. Ideally, there should be no requests with an existing email.

With this PR the default behavior of email uniqueness check remains unchanged, but now it's possible to configure states of users across which email is checked using `ckan.user.unique_email_states = active pending deleted some-other-status` option.

### DIFF
--- a/changes/8626.feature
+++ b/changes/8626.feature
@@ -1,3 +1,5 @@
 :ref:`ckan.user.unique_email_states` can be used to specify statuses of user
 accounts that are used for checking uniqueness of the email during
-registration.
+registration. After changing the value of the option, run `ckan db
+duplicate_emails` CLI command to verify that all existing emails are still
+unique.
--- a/changes/8626.feature
+++ b/changes/8626.feature
@@ -1,0 +1,3 @@
+:ref:`ckan.user.unique_email_states` can be used to specify statuses of user
+accounts that are used for checking uniqueness of the email during
+registration.
--- a/changes/8626.bugfix
+++ b/changes/8626.bugfix
@@ -1,0 +1,1 @@
+Switch login by email to case-insensitive mode.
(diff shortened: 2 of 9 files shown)


changes/8626.feature
```diff
@@ -1,3 +1,5 @@
 :ref:`ckan.user.unique_email_states` can be used to specify statuses of user
 accounts that are used for checking uniqueness of the email during
-registration.
+registration. After changing the value of the option, run `ckan db
+duplicate_emails` CLI command to verify that all existing emails are still
+unique.
```
wardi, line 5: The most important part of this change is treating differing email cases as the same, so even if you haven't touched the new option you'll still want to run the duplicate_emails cli check, right?

wardi:
Just needs a couple tests (or maybe a small changes to an existing tests)

changes/8626.feature
smotornyuk: Yes, makes sense. I created a separate changelog record for case-insensitive emails and mentioned there recommendation to run the command

smotornyuk:
There were no tests for this command, but now we have 2 :fire:

PR #8630, opened 2025-01-28
chore: Normalize format of log strings

smotornyuk:
Replace [str.format](https://docs.astral.sh/ruff/rules/logging-string-format/) and [printf-format](https://docs.astral.sh/ruff/rules/logging-percent-format/) with the recommended way of passing variables into log messages.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           python-version: '3.9'
       - name: Install requirements
-        run: pip install flake8 pycodestyle
+        run: pip install flake8 pycodestyle flake8-logging-format
       - name: Check syntax
         # Stop the build if there are Python syntax errors or undefined names
         run: flake8 --count --statistics --show-source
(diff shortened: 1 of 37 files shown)


ckan/lib/plugins.py
```diff
@@ -181,8 +181,8 @@ def register_package_blueprints(app: 'CKANFlask') -> None:
                 u"resource", blueprint=resource_blueprint)
             app.register_blueprint(resource_blueprint)
             log.debug(
-                'Registered blueprints for custom dataset type \'{}\''.format(
-                    package_type))
+                'Registered blueprints for custom dataset type \'%s\'',
```
EricSoroos, line 184: use double quotes here?

EricSoroos (approved):
Thanks for doing this...

.github/workflows/flake8.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         with:
           python-version: '3.9'
       - name: Install requirements
-        run: pip install flake8 pycodestyle
+        run: pip install flake8 pycodestyle flake8-logging-format
```
wardi, line 16: just having this installed is enough to enforce the check?

PR #8635, opened 2025-01-29
fix: remove unnecessary BOM_UTF8 in group members CSV export

aleks-iv:
### Proposed fixes:
This PR removes the unnecessary BOM_UTF8 from the CSV export of group members. The current code writes BOM_UTF8 to output_stream, which expects a string but receives bytes, causing an error. Since the BOM is not required for UTF-8 encoding, removing it prevents this issue while keeping the CSV export behavior unchanged.

### DIFF
--- a/ckan/views/group.py
+++ b/ckan/views/group.py
@@ -8,7 +8,6 @@ from typing import Any, Optional, Union
 from urllib.parse import urlencode
 import csv
 from io import StringIO
-from codecs import BOM_UTF8
 
 import ckan.lib.base as base
 from ckan.lib.helpers import helper_functions as h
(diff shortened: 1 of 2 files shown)


ckan/views/group.py
```diff
@@ -604,7 +603,6 @@ def member_dump(id: str, group_type: str, is_organization: bool):
         ])
 
     output_stream = StringIO()
```
wardi, line 605: Let's keep the BOM (just fix the type) so that non-ASCII characters are rendered properly in Excel

```suggestion
    output_stream = StringIO()
    output_stream.write('\N{BOM}')  # for Excel handling of non-ASCII
```

PR #8636, opened 2025-01-29
fix: Fix(or rather extend) IAuthenticator interface

### LINKED ISSUE #7581: Flask-login break IAuthenticator identify interface

smotornyuk:
Fixes #7581 and partially replaces #7591 (which is still active, because Tome agreed to replace `g.user` references with `current_user` there instead of the original goal).

## The Problem

`IAuthenticator.identify` has been partially broken since v2.10. The part where it sets cookies or produces redirects is still working. But setting `g.user` has no effect anymore, because `current_user` from flask login has higher priority.

## Why this important

Broken functionality allows per-request user identification. CKAN itself does not rely on this ability, but here are some examples of features that can be built upon it:

* Perform a single web request on behalf of another user.
* Temporarily disable the API token for a user(without removing it).
* Login into a "nested session". For example, as a sysadmin, switch to another account, do something as a different user, and switch back to your original session.
* Implement an authenticator that identifies the user's every request, just like authentication via API tokens.

## Rejected ideas

* **call `login_user()` instead of setting `g.user`**. Even though it works, every `login_user()` regenerates the session cookie. Because every page emits additional requests to pull CSS/JS, so the session cookie is regenerated a dozen times for every rendered page. But even with a single request, replacing the session cookie every time doesn't sound right.
* **replicate internal behavior of `login_user()` without actually calling it, to preserve session cookie**. This can be done via the method suggested by @jonathansberry -`app.login_manager._update_request_context_with_user(desired_ckan_user_obj)`. The main disadvantage of this method is relying on internals - it can change in the future and it can cause bugs because flask-login expects that the user is logged in only with public methods.
* **call `IAuthenticator.identify` inside flask-login's hooks that load user from session\headers**. This is the best-rejected option. But if we go with it, we cannot modify responses and emit redirects inside `identify` anymore. Given the fact that it's the only functionality of this method that currently works, it will only cause more issues during the upgrade. And this method does not fit perfectly into flask-login's hooks anyway.

## Solution

First, functionality that works. Producing redirects and modifying the response. Instead of using `IAuthenticator.identify`, it's better to implement `IMiddleware.make_middleware` that adds `before_request` / `after_request` callbacks to the application, which do the same. Ideally `IAuthenticator.identify` should not be used and I marked as deprecated. Nevertheless, it's still called before each request and it's possible to modify the response from it.

Second, broken functionality. Because it doesn't work in v2.10 and v2.11, plugins do not rely on `identify`'s ability to override the current user. It means we can introduce a new method that will do the same, and nobody will get upset. Actually, I mentioned before, that `identify` cannot be integrated ideally into flask-login. So I added two new methods instead, exactly as flask-login recommends:

* `identify_session_user` - this method receives the user's ID/name from a session or remember cookie. If the method returns a user object, authentication happens. If the method cannot identify the user, flask-login checks the next method.
* `identify_request_user` - this method works as default for anonymous requests(without sessions and cookies) as well as a fallback for `identify_session_user`. It accepts current requests and can return user objects based on headers(api-token), custom cookies or IP address.

Historically, we did not tested login via session, but to verify this new functionality, I updated test client and now it's possible to login user for subsequent requests using the following code in tests:

```py
def test_something(app, user):
   app.get(url) # this is an anonymous request
 
   app.set_session_user(user["id"]) # user is logged in
   app.get(url) # this is an authenticated request

   app.set_session_user(None) # user is logged out
   app.get(url) # once again, this is an anonymous request
```

It's still possible to log in user via `Authorization` header, so tests do not require modification. This new method can be used if you like it, but I added this method only for testing.


changes/8636.bugfix
```diff
@@ -0,0 +1,45 @@
...
+      # this method get called only when flask-login locates an existing user
+      # session
+      username = request.headers.get("USERNAME")
+
+      # ignore ID from session and login using header value if you trust this header
+      if is_verified(username):
+          return model.User.get(username)
+
+      # otherwise, continue with default implementation
+      return None
```
wardi, line 45: Would it be simpler to have one method with an optional `user_id` parameter instead of two methods? At least for this example that would reduce some duplicated code.

ckan/config/middleware/__init__.py
```diff
@@ -33,6 +33,9 @@ def make_app(conf: Union[Config, CKANConfig]) -> CKANApp:
     # Set this internal test request context with the configured environment so
     # it can be used when calling url_for from tests
     global _internal_test_request_context
-    _internal_test_request_context = flask_app._wsgi_app.test_request_context()
+    _internal_test_request_context = getattr(
+        flask_app,
+        "_wsgi_app",
+    ).test_request_context()
```
wardi, line 39: What's getattr here for? isn't this just a slower version of `flask_app._wsgi_app.test_request_context()`?

ckan/config/middleware/flask_app.py
```diff
@@ -90,15 +95,47 @@ def _get_interface(self, app: CKANApp):
...
         self.app = app
 
     def __call__(self, environ: Any, start_response: Any):
-
         handle_i18n(environ)
         return self.app(environ, start_response)
 
 
+def handle_i18n(base_environ: dict[str, Any] | None) -> None:
+    u'''
```
wardi, line 107: Might as well clean up any `u` prefixes on code we're touching

ckan/config/middleware/flask_app.py
```diff
@@ -90,15 +95,47 @@ def _get_interface(self, app: CKANApp):
...
+    '''
+    environ = base_environ or request.environ
+    locale_list = get_locales_from_config()
+    default_locale = config.get(u'ckan.locale_default')
+
+    # We only update once for a request so we can keep
+    # the language and original url which helps with 404 pages etc
+    if u'CKAN_LANG' not in environ:
+        path_parts = environ[u'PATH_INFO'].split(u'/')
+        if len(path_parts) > 1 and path_parts[1] in locale_list:
```
wardi, line 124: `ckan.root_path` might have a more complicated structure, see the example in https://docs.ckan.org/en/2.11/maintaining/configuration.html#ckan-root-path

ckan/config/middleware/flask_app.py
```diff
@@ -251,55 +288,78 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+    """
+    This callback function is called whenever we need to reload from
+    the database the logged in user in the session (ie the cookie).
+
+    Site maintainers can choose to completely ignore cookie based
+    authentication for API calls, but that will break existing JS widgets
+    that rely on API calls so it should be used with caution.
+    """
+    for p in PluginImplementations(IAuthenticator):
+        if user := p.identify_session_user(user_id):
```
wardi, line 325: The walrus appears! :heart:

ckan/config/middleware/flask_app.py
```diff
@@ -251,55 +288,78 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+            return user
+
+    endpoint = request.endpoint or ""
+    is_api = endpoint.split(".")[0] == "api"
+    if is_api and not config.get("ckan.auth.enable_cookie_auth_in_api"):
+        return
```
wardi, line 331: This should be at the start of the function

ckan/config/middleware/flask_app.py
```diff
@@ -251,55 +288,78 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+    return model.User.get(user_id)
+
+
+def load_user_from_request(
+        request: Request
+) -> model.User | model.AnonymousUser | None:
+    """
+    This callback function is called whenever a user could not be
+    authenticated via the session cookie, so we fall back to the API token.
```
wardi, line 340: Is this the correct order? If a user passes an API key I would expect that to override any session information

ckan/config/middleware/flask_app.py
```diff
@@ -251,55 +288,78 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+def _get_user_for_apitoken() -> model.User | None:
+    apitoken_header_name = config["apitoken_header_name"]
+    apitoken: str = request.headers.get(apitoken_header_name, "")
+
+    if not apitoken:
+        return None
+    apitoken = str(apitoken)
+    log.debug('Received API Token: %s[...]', apitoken[:10])
+
+    return api_token.get_user_from_token(apitoken)
```
wardi, line 360: it'd be useful to log the user name (if one is found) at the same time as the token

ckan/config/middleware/flask_app.py
```diff
@@ -557,7 +615,7 @@ def filter(self, log_record: Any) -> bool:
 
 
 def _setup_webassets(app: CKANApp):
-    app.use_x_sendfile = config.get('ckan.webassets.use_x_sendfile')
+    setattr(app, "use_x_sendfile", config.get('ckan.webassets.use_x_sendfile'))
```
wardi, line 618: why setattr? If this is only to avoid a pyright error I'd say leave the original because it's clearer

ckan/plugins/interfaces.py
```diff
@@ -1688,45 +1688,112 @@ def organization_facets(
...
+    ) -> model.User | model.AnonymousUser | None:
+        """Load a user using identifier stored in the session.
 
-        If the user is identified then it should set:
+        When :py:func:`~ckan.plugins.toolkit.ckan.plugins.toolkit.login_user`
+        is called with a user object, user's ID is saved in the session. After
+        that, in the beginning of each request the same user's ID from the
+        session is used to get user details via
+        :py:meth:`.identify_session_user`. And when
+        :py:func:`~ckan.plugins.toolkit.ckan.plugins.toolkit.logout_user` is
```
wardi, line 1740:
```suggestion
        :py:func:`~ckan.plugins.toolkit.logout_user` is
```

smotornyuk: This is intentional. Somehow, sphinx builds these double-paths anchors in toolkit, look for example on [this anchor](https://docs.ckan.org/en/2.11/extensions/plugins-toolkit.html#ckan.plugins.toolkit.ckan.plugins.toolkit.ObjectNotFound). When duplication of `ckan.plugins.toolkit` is removed from it, you won't get to the correct place on the page. Another example: [blanket reference on the first line here](https://docs.ckan.org/en/2.11/maintaining/configuration.html#using-a-text-file-json-yaml-or-toml) leads to the correct place because it uses double-path, while [add_template_directory just above parameters description here](https://docs.ckan.org/en/2.11/maintaining/configuration.html#using-a-text-file-json-yaml-or-toml) does not work because it uses normal path.

I added this bug to my todo list, but inside this PR I decided not to investigate it to narrow down the scope of changes

ckan/plugins/interfaces.py
```diff
@@ -1688,45 +1688,112 @@ def organization_facets(
...
+            return response
+
+    Instead of using :py:meth:`.identify` in this role, it's recommended to use
+    :py:class:`~ckan.plugins.interfaces.IMiddleware` interfaces. Its
+    :py:meth:`~ckan.plugins.interfaces.IMiddleware.make_middleware` accpets
+    ``app`` object that can be supplied with before-request callback::
+
+        p.implements(IMiddleware, inherit=True)
+        def make_middleware(self, app):
+            app.before request(
```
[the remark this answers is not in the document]
smotornyuk: Certainly

ckan/config/middleware/flask_app.py
smotornyuk: Yes, that only added to suppress typechecker complaints, reverting it back

changes/8636.bugfix
smotornyuk: Let's try it. I'll update the documentation to explain, how this optional `user_id` marks different situations.

ckan/config/middleware/__init__.py
smotornyuk: The same as below, I did it to avoid complaints about undefined attribute - reverting it back, as it isn't any safer than direct access to the attribute

ckan/config/middleware/flask_app.py
smotornyuk: Yes, makes sense. But this is an unchanged code of the function that I moved into this module because it's used only here by I18n middleware. I'll probably fix it inside simplify-url-for PR as it's more related to locale, than the current PR with authentication fixes.

ckan/config/middleware/flask_app.py
smotornyuk: That's how it worked since v2.10. But now you mentioned it, I think that I'd also expect that api-token has higher priority. So I added similar logic and now api-token overrides session user if present.

ckan/config/middleware/flask_app.py
```diff
@@ -251,55 +288,78 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
...
+) -> model.User | model.AnonymousUser | None:
+    """
+    This callback function is called whenever a user could not be
+    authenticated via the session cookie, so we fall back to the API token.
+    """
+    for p in PluginImplementations(IAuthenticator):
+        if user := p.identify_request_user(request):
+            return user
+
+    g.login_via_auth_header = True
```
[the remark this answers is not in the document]
smotornyuk: That's also the situation we have since v2.10. I checked this variable and it's used only to disable CSRF for header authentication. There is no sense in disabling CSRF for anonymous user, so I moving this line inside the condition. It may turn 403/401 error into CSRF error, but I think this is a breaking change for good.

smotornyuk:
Because I'm setting `logged_via_header` only when user actually logged via header(before it was set whenever user failed to login via session), POST requests in test produce 400 error instead of 403. Instead of updating tests I'm disabling CSRF protection of post requests in tests. It can be enabled on individual test using decorator `@pytest.mark.ckan_config("WTF_CSRF_ENABLED", True)`

ckan/config/middleware/flask_app.py
smotornyuk: Turned out, it's not a mistake. Whatever root-path is, it never get's into CKAN - `/data/{{LANG}}` or `/x/y/{{LANG}}/z/f` - in both cases when user visits for example organization page in CKAN with such root path, application sees only part after the root-path, i.e `/organization`. 
It means, it's responsibility of the server to insert the language in the beginning of the path when proxifying the request. For example, using following nginx configuration for `/data/{{LANG}}/site` root path:
```ini
    location ~ \/data(\/\w+)?\/site\/(.*) {
        proxy_pass   http://127.0.0.1:5000$1/$2;
    }
```

wardi: I don't think that's the way we've used it in the past. Where is this documented?

smotornyuk: There is no documentation that explains how locale gets into CKAN, that's the behavior I see in the few last versions - CKAN does not understands the locale if it's not in the beginning of the URL that gots into CKAN.
Documentation of the [root_path](https://docs.ckan.org/en/2.11/maintaining/configuration.html#ckan-root-path) mentions only URL construction and says nothing about URL parsing, so it seems that that was the behaviour that no one knew about.

Zharktas:
I've been testing this with our extension where the original issue report comes from. For extension developers whom used to set g.user, this solution is elegant and the migration is just couple lines of code.

Zharktas:
@smotornyuk api tokens given on wrong format seem to result csrf error in the UI even if user tries to use just api calls.

ckan/config/middleware/flask_app.py
smotornyuk: @wardi , I just realized, that the code in question is not a part of this PR. It was already there, I just moved it from `views/__init__.py` to `flask_app.py`, because it's used only here.

smotornyuk:
@Zharktas, I found the problem. There is a `g.login_via_auth_header` that is used to disable CSRF protection. I misinterpreted its meaning and enabled it only after _succesfull_ logins.  But initially, this flag was enabled whenever a login via header was attempted, no matter whether is succesfull or not. So I restored the original behavior, and now it must be compatible with your PR

ckan/tests/helpers.py
```diff
@@ -198,39 +200,69 @@ def flask_app(self):
...
+        else:
+            self.client.delete_cookie(key, domain=domain)
+
+    def test_client(
+            self,
+            use_cookies: bool = True,
+            user: model.User | None = None
+    ):
+        return CKANTestClient(
+            self.flask_app,
```
amercader, line 234: @smotornyuk Here we are switching the WSGI app used by the test client. Up until now we were using `self.app`, i.e. the CKAN wsgi app, which includes stuff like the I18NMiddleware, and here you are using `self.flask_app` (the Flask WSGI app).

I think this is causing the test failures that we are seeing.

If we need to use the Flask app because you switched to `FlaskLoginClient`  I think the fix is what you hinted here:

```
# TODO: convert to app.before_request
    app: Any = I18nMiddleware(app)
```

I tried to create a before_request handler but as we are modifying URLs on the environ it needed to modify Flask internal mapper to dispatch things properly and it increased the complexity a bit. So I decided for a much simpler approach to wrap the Flask app with the middleware rather than the CKAN app:

https://github.com/ckan/ckan/commit/14006ec2669d7474c1db05b4e35741998bc42432

Testing it now in a separate branch to see if tests are happy:

https://github.com/ckan/ckan/actions/runs/23001635527

smotornyuk: Yes, this looks great. I did not realize that `before_request` would not work here because it's executed after the routing. Your solution looks like the best option for us.

I see that tests are all green in your branch. I'm trying to apply it to the current PR

amercader:
Thanks @smotornyuk , merging this

PR #8639, opened 2025-01-30
Fix datastore data dictionary page access

mutantsan:
The datastore data dictionary page doesn't have a proper auth check

This also will fix the table designer views (data dictionary, add, remove, update views), because they call the `DictionaryView` view.

### Proposed fixes:

Check for `datastore_upsert` 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -139,9 +139,9 @@ class DictionaryView(MethodView):
     def _prepare(self, id: str, resource_id: str) -> dict[str, Any]:
         try:
             check_access(
-                "datastore_upsert",
+                "datastore_create",
                 context={"user": current_user.name, "auth_user_obj": current_user},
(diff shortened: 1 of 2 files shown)


ckanext/datastore/blueprint.py
```diff
@@ -136,6 +138,12 @@ class DictionaryView(MethodView):
 
     def _prepare(self, id: str, resource_id: str) -> dict[str, Any]:
         try:
+            check_access(
+                "datastore_upsert",
+                context={"user": current_user.name, "auth_user_obj": current_user},
+                data_dict={"id": id, "resource_id": resource_id},
```
wardi, line 144: We should use the correct action that will be called
```suggestion
                "datastore_create",
                context={"user": current_user.name, "auth_user_obj": current_user},
                data_dict={"resource_id": resource_id},
```

mutantsan: you're right

PR #8640, opened 2025-01-31
Simplify url_for

smotornyuk:
Make `url_for` helper less custom.

Flask uses `SERVER_NAME` and `APPLICATION_ROOT` when building URLs. `SERVER_NAME` is used for external URLs and `APPLICATION_ROOT` serves a similar purpose to `ckan.root_path`. These options can be automatically set during initialization, allowing us to remove essential parts of the `url_for` custom logic. 

Breaking changes: 

* `url_for`, `url_for_static`, and `url_for_static_or_external` now accept only one positional parameter - endpoint in case of `url_for` and relative/absolute URL in case of the other two helpers. All other parameters are keyword-only
* `url_for` accepts only endpoint names. Previously it was allowed to pass into it relative or absolute URLs(can be easily replaced with `url_for_static_or_external`).
* Query parameters must be passed into `url_for_static` as named parameters. Previously it was allowed to pass relative URLs with `?...` part.

### DIFF
--- a/ckan/config/middleware/common_middleware.py
+++ b/ckan/config/middleware/common_middleware.py
@@ -14,27 +14,24 @@ from ckan.lib.redis import connect_to_redis
 
 
 class RootPathMiddleware(object):
-    '''
-    Prevents the SCRIPT_NAME server variable conflicting with the ckan.root_url
-    config. The routes package uses the SCRIPT_NAME variable and appends to the
-    path and ckan addes the root url causing a duplication of the root path.
-    This is a middleware to ensure that even redirects use this logic.
-    '''
+    """Set SERVER_NAME and SCRIPT_NAME.
+
+    Ideally, these variables must be set by web server, but adding it here is
+    more reliable and works even when server configuration is not synchronized
+    with CKAN configuration.
+    """
     def __init__(self, app: CKANApp):
         self.app = app
 
     def __call__(self, environ: Any, start_response: Any):
-        # Prevents the variable interfering with the root_path logic
-        parsed = urlparse(config["ckan.site_url"])
-        environ["SERVER_NAME"] = parsed.netloc
+        # SERVER_NAME is used for building external URLs.
+        environ["SERVER_NAME"] = config["SERVER_NAME"]
 
-        if not environ.get("SCRIPT_NAME"):
-            application_root = config["ckan.root_path"] or "/"
-            if not application_root.endswith("/"):
-                application_root += "/"
-            application_root = application_root.replace("/{{LANG}}", "")
-            environ["SCRIPT_NAME"] = application_root
-
+        # SCRIPT_NAME defines the prefix of the URL path. It's set to
+        # `ckan.root_path` without locale part. The locale is added manually
+        # inside `url_for` helper, which opens a possibility for building
+        # cross-locale links(mainly used by language selector).
+        environ["SCRIPT_NAME"] = config["APPLICATION_ROOT"]
 
         return self.app(environ, start_response)
 
--- a/ckan/config/middleware/common_middleware.py
+++ b/ckan/config/middleware/common_middleware.py
@@ -25,8 +25,16 @@ class RootPathMiddleware(object):
 
     def __call__(self, environ: Any, start_response: Any):
         # Prevents the variable interfering with the root_path logic
-        if 'SCRIPT_NAME' in environ:
-            environ['SCRIPT_NAME'] = ''
+        parsed = urlparse(config["ckan.site_url"])
+        environ["SERVER_NAME"] = parsed.netloc
+
+        if not environ.get("SCRIPT_NAME"):
+            application_root = config["ckan.root_path"] or "/"
+            if not application_root.endswith("/"):
+                application_root += "/"
+            application_root = application_root.replace("/{{LANG}}", "")
+            environ["SCRIPT_NAME"] = application_root
+
 
         return self.app(environ, start_response)
 
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -11,6 +11,7 @@ import logging
 
 from logging.handlers import SMTPHandler
 from typing import Any, Optional, Union, cast
+from urllib.parse import urlparse
 
 from flask import Blueprint, send_from_directory, current_app
 from flask.ctx import _AppCtxGlobals
@@ -131,6 +132,13 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     app.testing = testing
     app.template_folder = os.path.join(root, 'templates')
     app.app_ctx_globals_class = CKAN_AppCtxGlobals
+
+    application_root = config["ckan.root_path"] or "/"
+    if not application_root.endswith("/"):
+        application_root += "/"
+    application_root = application_root.replace("/{{LANG}}", "")
+    config["APPLICATION_ROOT"] = application_root
+    config["SERVER_NAME"] = urlparse(config["ckan.site_url"]).netloc
 
     # Update Flask config with the CKAN values. We use the common config
     # object as values might have been modified on `load_environment`
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -140,8 +140,6 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
         app.config.update(conf)
 
     # Do all the Flask-specific stuff before adding other middlewares
-
-    root_path = config.get('ckan.root_path')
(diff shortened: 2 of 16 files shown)


pdelboca:
+1 on simplifying `url_for` 🎉 

I will repeat come context for future readers. Our `url_for` has two main purposes:
 1) Helped with the Pylons -> Flask migration.
 2) Use `url_for` in tests. (See https://github.com/ckan/ckan/pull/7251)
 
If I understood correctly, [flask.url_for()](https://flask.palletsprojects.com/en/stable/api/#flask.Flask.url_for) now allows for it to be called outside request contexts and it will use ENV variables to build it right (like the emails example in the their documentation).
 
Will this allow us to finally kill the `_internal_test_request_context` hack? Any thoughts?

smotornyuk:
Sounds great. I'll try removing it and see how CKAN reacts.

smotornyuk:
Seems, it's not our case. What they mean is that **when inside application context(`app.app_context()` - `current_app` global variable is set)** you don't need to create nested request context(`app.request_context()` - pretend that user visited some specific endpoint). 

We still need application context to tell Flask, which exact application we are working with. Flask doesn't track all the application objects initialized, so it doesn't understand that we are going to use the application initialized by CLI command/test/etc. Without knowing which app to use, flask doesn't know which endpoints are connected to the application and how to build a URL for `home.index`, for example.

Instead of creating a request context, we can use app context - that sounds better, as we are not imitating the request. I'll test it to see, how many tests are relying on this fake-request details. But removing it completely is not an option, because it will force either to hardcode URLs without using url_for in tests(it sounds somehow interesting, because we are using url_for in functional tests, that imitate behavior or real users, so hardcoded URLs can highlight unexpected changes in URL-rules, while url_for hides this type of problem), or wrap any call of url_for into new app_context. But in both cases that's just a lot of work so we can discuss it separately and maybe suggest it with a new PR as a part of CKAN v3 innovations.

ckan/lib/webassets_tools.py
```diff
@@ -135,10 +135,14 @@ def include_asset(name: str) -> None:
...
-    for url in urls:
-        link = url.split("?")[0]
+    urls: list[str] = []
+
+    for url in bundle.urls():
+        parts = url.split("?", 1)
+        link = parts.pop(0)
+        query = parts.pop() if parts else ""
+        link = url_for_static_or_external(link)
+        urls.append("?".join([link, query]))
```
wardi, line 145: this would add a trailing `?` to urls without a query, right?

ckan/templates-midnight-blue/snippets/language_selector.html
```diff
@@ -1,17 +1,17 @@
 {% set current_lang = request.environ.CKAN_LANG %}
+{% set url_args = request.args.to_dict() %}
+{% do url_args.update(request.view_args) %}
```
wardi, line 3: mixing a MultiDict of request parameters with view arguments seems like a really bad practice.

The MultiDict may have multiple values for the same parameter that will be lost and may conflict with view args names (again, losing parameter values)

What about having a specific parameter for the MultiDict parameters to `url_for` instead to keep these different values separate?

ckan/config/middleware/common_middleware.py
```diff
@@ -14,19 +14,24 @@
...
-        if 'SCRIPT_NAME' in environ:
-            environ['SCRIPT_NAME'] = ''
+        # SERVER_NAME is used for building external URLs.
+        environ["SERVER_NAME"] = config["SERVER_NAME"]
+
+        # SCRIPT_NAME defines the prefix of the URL path. It's set to
+        # `ckan.root_path` without locale part. The locale is added manually
+        # inside `url_for` helper, which opens a possibility for building
+        # cross-locale links(mainly used by language selector).
+        environ["SCRIPT_NAME"] = config["APPLICATION_ROOT"]
```
wardi, line 34: I remember we needed this unset so that the application could see the whole URL and handle some of the root_path edge cases. Is that no longer required?

ckan/config/middleware/flask_app.py
```diff
@@ -132,6 +133,13 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     app.template_folder = os.path.join(root, 'templates')
     app.app_ctx_globals_class = CKAN_AppCtxGlobals
 
+    application_root = config["ckan.root_path"] or "/"
+    if not application_root.endswith("/"):
+        application_root += "/"
+    application_root = application_root.replace("/{{LANG}}", "")
+    config["APPLICATION_ROOT"] = application_root
```
wardi, line 140: this won't work if the `/{{LANG}}` part isn't at the end of `root_path`

ckan/lib/helpers.py
```diff
@@ -346,193 +312,91 @@ def url_for(*args: Any, **kw: Any) -> str:
...
+    if not args:
+        args = (f"{blueprint}.{view}",)
 
-    This function should not be called directly, use ``url_for`` instead
+    # pylons-style absolute URLs. It should be removed after explicit
+    # deprecation.
+    if kw.pop("qualified", False):
+        kw.setdefault("_external", True)
+    if not request:
+        kw.setdefault("_external", False)
```
wardi, line 331: So if we're calling `url_for` and there's no request, e.g. from tests or background job or from the cli then default to not including the full path? Is that the intent?

ckan/lib/helpers.py
```diff
@@ -346,193 +312,91 @@ def url_for(*args: Any, **kw: Any) -> str:
...
 
-        # Pylons
-        url_for(controller='api', action='action', ver=3, qualified=True)
+    _safe_context = _get_auto_flask_context() or contextlib.nullcontext()
+    with _safe_context:
+        url = flask.url_for(*args, **kw)
 
-        # Flask
-        url_for('api.action', ver=3, _external=True)
+    if locale != 'default':
```
wardi, line 339: locale will be None here if it wasn't included, shouldn't that be treated as the default?

ckan/lib/helpers.py
smotornyuk: Yes, but it's better to say that url_for without an explicit `_external` flag _always_ produces an internal URL in CKAN.
In flask, when you are calling url_for during the request(without _external), it produces an internal URL. When url_for is called outside the request(without the _external), it produces an external URL. This is mentioned in the last sentence of (_external's description](https://flask.palletsprojects.com/en/stable/api/#flask.Flask.url_for)
It looks like our tests are assuming that url_for always behaves in the same way. Probably it's a Pylons legacy, but I see no reasons to opt-oout of this fix - it will break a lot of tests in extensions without any visible gain.

ckan/config/middleware/flask_app.py
smotornyuk: I missed the point, what is wrong?

* `/data/{{lang}}` -> append / `/data/{{lang}}/` -> drop lang with preceding / `/data`
* `/data/{{lang}}/other` -> append / `/data/{{lang}}/other/` -> drop lang with preceding / `/data/other/`

ckan/config/middleware/common_middleware.py
smotornyuk: Not exactly. The application never sees the full URL. It always receives only the part after SCRIPT_NAME(thanks to werkzeug). When we are outside of the request context, APPLICATION_ROOT is used as a prefix for constructed URLS. But when we are inside the request context, APPLICATION_ROOT is ignored and flask relies on SCRIPT_NAME instead, to compute the prefix of the URL. That's why we need both - SCRIPT_NAME most of the time, and APPLICATION_ROOT for tests, cli, etc. 

Previously we cleaned SCRIPT_NAME, because some of webservers added it without an explicit configuration, while others didn't. As result, sometimes flask duplicated root_path, using this SCRIPT_NAME set by the server.

Without this line, SCRIPT_NAME can also be set by the web server. But to achieve deterministic results, it must be explicitly configured, which requires more work during the upgrade. So I just making sure this variable is set and has a value that matches CKAN configuration

ckan/lib/webassets_tools.py
smotornyuk: Yes, I didn't notice it.

amercader:
@smotornyuk there are a ton of tests failing, looks like it should be something easy to fix

smotornyuk:
Yes, that was imperfect `remove_url_param`. It seems we used it only on pages without URL wildcards before, and because of it, we didn't notice that it has small issues when the processed URL contains a dynamic part, like `<package_type>` or `<id>`. Now it's fixed

Another problem is related to the conditional registration of collaborator routes. Flask does not support blueprint modifications, so I changed the logic a bit. Now collaborator routes are registered all the time, but they will return 404 error if this feature is not enabled

PR #8657, opened 2025-02-10
IFormRedirect for custom dataset/resource editing workflows

### LINKED ISSUE #7778: Custom dataset editing redirects
To implement features like **upload-first workflow** and **table designer resources** we need a convenient way to override redirects when submitting dataset and resource forms.

e.g. to implement the following workflow:

1. upload resource
2. customize data dictionary
3. fill out resource metadata
4. fill out dataset metadata
5. done

<dl>
<dt>step 1:</dt>
<dd>can be accomplished with a custom view that accepts a file upload and creates a draft dataset+resource, waits for xloader/datapusher, then redirects to the data dictionary form.</dd>
<dt>steps 2-4:</dt>
<dd>require 90% of the existing ckan dataset/resource flask view code but need to alter the redirect at the end to continue to the correct page for the next step. We can only accomplish that now by copying 90% of the code into a new view with a custom redirect at the end.</dd>
</dl>

Proposal:
- have an easy way for a plugin to override the redirect url after successfully submitting dataset/resource/data dictionary forms

This could be a template helper function (we can already override and chain these) or a new interface of some kind.

Should this interface cover all forms that redirect on success, like user editing, group/org editing, resource view editing too?

wardi:
initial implementation and clean-up of dataset/resource create/edit views

Fixes #7778

### Proposed fixes:

IFormRedirect interface allows customizing the redirect at the end of form submissions to allow custom workflows for dataset/resource editing and creation without re-implementing/copying view code from ckan core.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/dataset.py
+++ b/ckan/views/dataset.py
@@ -27,7 +27,7 @@ from ckan.common import _, config, g, request
 from ckan.views.home import CACHE_PARAMETERS
 from ckan.lib.plugins import lookup_package_plugin
 from ckan.lib.search import (
-    SearchError, SearchQueryError, SearchIndexError, SolrConnectionError
+    SearchError, SearchQueryError, SolrConnectionError
 )
 from ckan.types import Context, Response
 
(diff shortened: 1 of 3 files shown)


wardi:
@amercader suggested moving these methods to IDatasetForm
- would be more elegant e.g. for handling different types differently and for ckanext-scheming to control the whole dataset form progression

@smotornyuk suggested forcing `inherit=True` for plugin interfaces to make it easier to add methods to existing interfaces
- likely better as a separate PR

amercader:
Perhaps the new `*_save_redirect()` methods are better placed inside of the existing `IDatasetForm` rather than in a separate new interface. This would allow more granularity if someone really wanted different redirect behaviours for different `IDatasetForm` implementations. And would avoid having to introduce a new interface for a pretty specific functionality.

ckan/views/dataset.py
```diff
@@ -510,137 +477,63 @@ def read(package_type: str, id: str) -> Union[Response, str]:
...
-                pkg_dict = get_action(u'package_show')(context, data_dict)
-                data_dict[u'state'] = pkg_dict[u'state']
-                return EditView().get(
-                    package_type,
-                    data_dict[u'id'],
-                    data_dict,
-                    errors,
-                    error_summary
-                )
-            data_dict[u'state'] = u'none'
```
amercader, line 643: Is this old unused code?

wardi: Right now the bulk of the CreateView and EditView code duplicate the same editing functionality. Here I've removed the editing functionality from CreateView so we can rely on the EditView implementation instead of having two slightly different copies of the same code.

PR #8662, opened 2025-02-11
Fix lint and typing errors in master

amercader:
@tino097 this should take care of the check failures in master (at least for the lint and typing workflows)


tino097 (approved):
LGTM

PR #8669, opened 2025-02-13
Pass `_ckan_phase` into Context & New Package Form Template Blocks

JVickery-TBS:
This is to support ckanext-scheming pages better.

### Features:

- [ ] includes API changes


wardi:
Not in favor of adding more variables to the context, especially if it's something that only an extension depends on and isn't covered by tests.

What's the larger problem you're trying to solve here, maybe there's a better approach?

wardi:
IIUC you're looking to implement something like https://github.com/ckan/ckanext-scheming/pull/418 and https://github.com/ckan/ckan/pull/8309

Those PRs works by checking if the state is draft and allows saving even if some required fields have not been provided yet, and updating the publish step so that it expects validation errors. This is a better approach than passing in an extra parameter that says "don't actually validate" because the validation is always consistent with the data stored.

JVickery-TBS:
@wardi with those PRs, would it just show all the validation errors at the final page of publishing? and then the user has to go find where those respective fields are throughout the pages?

wardi:
No, it shows the missing fields on the tabs with an red icon. When you hover over the icon it lists the required fields missing on that page.

JVickery-TBS:
@wardi hmmm okay yeah that is good then. I am just wondering what is better, to have no validation happen between the pages until you go to publish, or to show validation on each page as you try to save the page

JVickery-TBS:
@wardi oh yeah sorry, it is not that no validation happens in my scheming PR, it also just affects the `scheming_required` validator but instead of checking `state=draft` it checks for the page

PR #8673, opened 2025-02-17
Add `--create` option to `ckan sysadmin add` command 

amercader:
To automatically create the user if it doesn't exist. This simplifies initialization scripts by avoiding calling two commands (`ckan user add` and `ckan sysadmin add`) to avoid the confirmation prompt.

Also added tests for the whole `ckan sysadmin` command as there weren't any


wardi:
This is also possible in one command with
```bash
ckan user add ... sysadmin=true
```
using this fix: #8674

Zharktas (approved):

PR #8674, opened 2025-02-17
default_user_schema: add missing boolean_validator

wardi:
### Proposed fixes:

Add missing `boolean_validator` to user sysadmin flag.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8674.bugfix
+++ b/changes/8674.bugfix
@@ -1,0 +1,1 @@
+Add missing boolean_validator to sysadmin field in user schema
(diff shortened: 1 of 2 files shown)


amercader:
For reference, this prevented setting this field at all via the API as the value was passed all the way to sqlalchemy as a string:

```
ckanapi action user_create name=test-user-4 email=a4@example.org password=test1234 sysadmin=False

sqlalchemy.exc.StatementError: (builtins.TypeError) Not a boolean value: 'False'

```

So I think it's worth backporting

amercader:
Backported manually

PR #8675, opened 2025-02-17
[Midnight Blue] Dashboard pages

aleeexgreeen:
### Proposed fixes:
- Simple js module for scrolling to active nav item in mobile
- Dashboard secondary nav items
- Followee filter dropdown
- Add HTMX to dashboard news feed page. There is currently a bug when you press the newer or older activities button, it duplicates the page in the body instead of replacing the body content with the newer/older activities.
- Layout for dashboard pages

[Figma designs](https://www.figma.com/design/oI3wDSdZ0cHZLh6CrfYMnb/CKAN-3.0-Design-System?node-id=1037-8316)

### Screenshots

Desktop:
image

Mobile:
image

Screenshots taken from [preview site](https://ckan-design-preview.galv1.links.com.au/dashboard/), have to be logged in to see dashboard.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public-midnight-blue/base/css/main-rtl.css
+++ b/ckan/public-midnight-blue/base/css/main-rtl.css
@@ -12476,6 +12476,7 @@ a.tag:hover {
 @media (max-width: 767.98px) {
   .secondary .nav-items {
     background-color: transparent;
+    border-bottom: none;
     flex-wrap: nowrap;
     overflow-x: auto;
     overflow-y: scroll;
--- a/ckan/public-midnight-blue/base/css/main-rtl.css
+++ b/ckan/public-midnight-blue/base/css/main-rtl.css
@@ -12439,6 +12439,56 @@ a.tag:hover {
   color: #2f4b91;
 }
 
+.secondary .nav-tabs {
+  background-color: #f2f4f7;
+  border-radius: 8px;
+}
+.secondary .nav-tabs .active {
+  background-color: #153084;
+}
+.secondary .nav-tabs .active a {
+  color: #fff;
+}
+.secondary .nav-tabs .active:hover {
+  background-color: #153084;
+}
+.secondary .nav-tabs li {
+  border-radius: 8px;
+  padding: 10px 16px;
+  width: 100%;
+}
+.secondary .nav-tabs li a {
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row-reverse;
+}
+.secondary .nav-tabs li a i {
+  display: flex;
+  align-items: center;
+}
+.secondary .nav-tabs li a:hover {
+  text-decoration: none;
+}
+.secondary .nav-tabs li:hover {
+  background-color: #e4e7ec;
+}
+@media (max-width: 767.98px) {
+  .secondary .nav-tabs {
+    background-color: transparent;
+    flex-wrap: nowrap;
+    overflow-x: auto;
+    overflow-y: scroll;
+    white-space: nowrap;
+  }
+  .secondary .nav-tabs li {
+    margin-left: 8px;
+    background-color: #f2f4f7;
+  }
+  .secondary .nav-tabs li i {
+    margin-left: 1rem;
+  }
+}
+
 .nav-stacked > li {
   float: none;
 }
@@ -13950,11 +14000,7 @@ td.diff_header {
 
 @media (max-width: 767.98px) {
   .main .secondary {
-    -webkit-box-ordinal-group: 2;
-    -moz-box-ordinal-group: 2;
-    -webkit-order: 2;
-    order: 2;
-    margin-top: 1rem;
+    margin: 1rem 0;
   }
   .js .main .secondary .filters {
     display: none;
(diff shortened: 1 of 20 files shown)


wardi:
This looks promising but my dashboard "Activity from" drop down always says "you are not following anything" even though I've followed datasets and organizations.

Also, not sure if related but there are a few pixels between the dashboard and profile settings icons where the tool-tip flickers continuously:
[Screencast from 2025-02-24 11:54:29 AM.webm](https://github.com/user-attachments/assets/ca0475ad-6e3a-442c-abf7-253e163943b4)

amercader:
> Simple js module for scrolling to active nav item in mobile

I find the horizontal scroll for the nav items really confusing, isn't it better to be consistent with the tabs/pills in other pages and stack them vertically?

amercader:
I don't think htmx is working, selecting an item from the news feed dropdown refreshes the whole page. I don't see any htmx tags added to the templates so there might be some setup needed, I'll have a look

aleeexgreeen:
Hi @amercader I've taken care of most of the issues you've brought up. For `ckanext/activity` what is the best way to proceed? Changes will affect the default theme and there will be more changes to come with the activity stream.

The HTMX was added to the pagination buttons of the activity stream in #8160, the htmx additions were to fix a bug duplicating the page in the activity stream, no HTMX was added to the dropdown. After briefly investigating add HTMX to the dropdown, the dropdown items in `followee_dropdown.html` are being hidden in a form and I do not see where the visible ones are being added in the code, so it does not look to be an option right now.

amercader:
> For ckanext/activity what is the best way to proceed? Changes will affect the default theme and there will be more changes to come with the activity stream.
That's a tricky question, I'm not sure of the best approach for now, just thinking out loud.

I guess in an ideal world we would modify the existing templates in an extension in a way that is compatible with both themes (i.e. markup changes and class names) but that will get messy soon, specially with significant design changes.
I think we might need a separate folder for midnight blue templates in the activity extension *but* it will work differently than the core one.

Extensions templates directories are added to a stack of paths and when rendering a particular template file the rendering function goes in order over it so the first ones loaded can override blocks from the previous ones (or the other way around, I can't remember). This means that if we just register a `templates-midnight-blue` folder in addition to the `templates` one it will always render the midnight one on top of the default regardless of whether the theme is loaded or not. So we need to do something like:

```diff
diff --git a/ckanext/activity/plugin.py b/ckanext/activity/plugin.py
index fe610c040..98b8f48d5 100644
--- a/ckanext/activity/plugin.py
+++ b/ckanext/activity/plugin.py
@@ -22,6 +22,8 @@ class ActivityPlugin(p.SingletonPlugin):
     # IConfigurer
     def update_config(self, config: CKANConfig):
         tk.add_template_directory(config, "templates")
+        if config["ckan.base_templates_folder"] == "templates-midnight-blue":
+            tk.add_template_directory(config, "templates-midnight-blue")
         tk.add_public_directory(config, "public")
         tk.add_resource("assets", "ckanext-activity")

```

Once that's set up we have two options:

1. We do the same as in core and duplicate the whole `templates` folder, even if we just need to tweak certain files and blocks for the new theme
2. We only put files in `templates-midnight-blue` that need to be changed for the midnight blue theme changes.


While 2 involves less files now, if we ever want to remove the current theme or even make it not the default one that might cause issues so option 1 might be the safest. Let me check with the rest of the team and get back to you.

amercader:
> The HTMX was added to the pagination buttons of the activity stream in https://github.com/ckan/ckan/pull/8160, the htmx additions were to fix a bug duplicating the page in the activity stream, no HTMX was added to the dropdown. After briefly investigating add HTMX to the dropdown, the dropdown items in followee_dropdown.html are being hidden in a form and I do not see where the visible ones are being added in the code, so it does not look to be an option right now.

@pdelboca as you reviewed #8160 and are the sitting htmx expert would you mind looking at what's missing to get this working? thanks

amercader:
@aleeexgreeen after discussing in the dev call and have it thought through let's go with duplicated folders for now. It's a pain because this extension has a million little templates files but it will be safer this way. So copy the `ckanext/activity/templates` to `ckanext/activity/templates-midnight-blue` and only edit files in the latter.
Then add the following lines to plugin.py to load one or the other:

```diff
diff --git a/ckanext/activity/plugin.py b/ckanext/activity/plugin.py
index fe610c040..527a52934 100644
--- a/ckanext/activity/plugin.py
+++ b/ckanext/activity/plugin.py
@@ -21,7 +21,10 @@ class ActivityPlugin(p.SingletonPlugin):

     # IConfigurer
     def update_config(self, config: CKANConfig):
-        tk.add_template_directory(config, "templates")
+        if config["ckan.base_templates_folder"] == "templates-midnight-blue":
+            tk.add_template_directory(config, "templates-midnight-blue")
+        else:
+            tk.add_template_directory(config, "templates")
         tk.add_public_directory(config, "public")
         tk.add_resource("assets", "ckanext-activity")
```

Let's use the activity plugin as a test to see if how manageable this is and if we can reuse the approach can be used in other extensions.

BTW it looks like your changes to the SCSS files might also affect the existing theme so we might need to do the same with the assets (ie `tk.add_resource("assets", "ckanext-activity")` or `tk.add_resource("assets-midnight-blue", "ckanext-activity")` depending on the loaded theme)

Let me know if all this makes sense

aleeexgreeen:
@amercader This should be ready for re-review

amercader:
Sorry it took me a while to get to this @aleeexgreeen 
This looks good let's merge it as is and we can figure out the htmx later, it is not a priority.
---
Some investigation bits to implement the followee filter with htmx in the future (don't worry about this @aleeexgreeen ):

* To allow hx- attributes in the popover items they need to be added to the allowed attributes list:
```diff
diff --git a/ckanext/activity/assets-midnight-blue/dashboard.js b/ckanext/activity/assets-midnight-blue/dashboard.js
index 5f5dc89ea..4772759c4 100644
--- a/ckanext/activity/assets-midnight-blue/dashboard.js
+++ b/ckanext/activity/assets-midnight-blue/dashboard.js
@@ -31,7 +31,9 @@ this.ckan.module('dashboard', function ($) {
             return DOMPurify.sanitize(content, { ALLOWED_TAGS: [
               "form", "div", "input", "footer", "header", "h1", "h2", "h3", "h4",
               "small", "span", "strong", "i", 'a', 'li', 'ul','p'
-            ]});
+            ],
+            ADD_ATTR: ["hx-get", "hx-target"]
+            });
           },
           content: $('#followee-content').html()
         });

```
* This still doesn't trigger Htmx, I suspect because we are loading the popover from a [js module](https://github.com/ckan/ckan/blob/46cb7cfcf04b1c2f022a8147f20b3067eaea3869/ckanext/activity/assets/dashboard.js#L26)
* We can probably get rid of the dashboard.js module and use the contents of `followee_dropdown.html` directly as the popover content, that will simplify things and probably enable htmx

PR #8684, opened 2025-02-21
DataStore Actions Return Records

JVickery-TBS:
feat(logic): datastore returned records;

- Return records from SQL statements.

### Proposed fixes:

This will actually record the inserted/updated/deleted rows in the records returned in `datastore_create`, `datastore_upsert`, and `datastore_delete`

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1195,6 +1195,7 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
     records = data_dict['records']
     sql_columns = ", ".join(
         identifier(name) for name in field_names)
+    return_columns = "_id, " + sql_columns
     if not sql_columns:
         # insert w/ no columns is a postgres error
         return
@@ -1217,16 +1218,18 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
             rows.append(row)
 
         sql_string = '''INSERT INTO {res_id} ({columns})
-            VALUES ({values});'''.format(
+            VALUES ({values}) RETURNING {return_columns};'''.format(
             res_id=identifier(data_dict['resource_id']),
             columns=sql_columns,
+            return_columns=return_columns,
             values=', '.join([
                 f":val_{idx}" for idx in range(0, len(field_names))
             ])
         )
 
         try:
-            context['connection'].execute(sa.text(sql_string), rows)
+            results = context['connection'].execute(sa.text(sql_string), rows)
+            data_dict['records'] = [dict(r) for r in results.mappings().all()]
         except (DatabaseError, DataError) as err:
             raise ValidationError({
                 'records': [_programming_error_summary(err)],
@@ -1235,7 +1238,7 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
 
     elif method in [_UPDATE, _UPSERT]:
         unique_keys = _get_unique_key(context, data_dict)
-
+        updated_records = []
         for num, record in enumerate(records):
             if not unique_keys and '_id' not in record:
                 raise ValidationError({
@@ -1306,12 +1309,14 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
                 sql_string = u'''
                     UPDATE {res_id}
                     SET ({columns}, "_full_text") = ({values}, NULL)
-                    WHERE ({primary_key}) = ({primary_value});
+                    WHERE ({primary_key}) = ({primary_value})
+                    RETURNING {return_columns};
                 '''.format(
                     res_id=identifier(data_dict['resource_id']),
                     columns=u', '.join(
                         [identifier(field)
                          for field in used_field_names]),
+                    return_columns=return_columns,
                     values=u', '.join(values),
                     primary_key=pk_sql,
                     primary_value=pk_values_sql,
@@ -1320,6 +1325,7 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
                     results = context['connection'].execute(
                         sa.text(sql_string),
                         {**used_values, **unique_values})
+                    updated_records += [dict(r) for r in results.mappings().all()]
                 except DatabaseError as err:
                     raise ValidationError({
                         'records': [_programming_error_summary(err)],
@@ -1338,6 +1344,7 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
                     columns=u', '.join(
                         [identifier(field)
                          for field in used_field_names]),
+                    return_columns=return_columns,
                     values=u', '.join([
                         f'cast(:{p} as nested)'
                         if field['type'] == 'nested' else ":" + p
@@ -1358,19 +1365,23 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
                            SELECT {values}
                            WHERE NOT EXISTS (SELECT 1 FROM {res_id}
                                     WHERE ({primary_key}) = ({primary_value}))
+                           RETURNING {return_columns}
                 """.format(**format_params)
 
                 values = {**used_values, **unique_values}
                 try:
                     context['connection'].execute(
                         sa.text(update_string), values)
-                    context['connection'].execute(
+                    results = context['connection'].execute(
                         sa.text(insert_string), values)
+                    updated_records += [dict(r) for r in results.mappings().all()]
                 except DatabaseError as err:
                     raise ValidationError({
                         'records': [_programming_error_summary(err)],
                         'records_row': num,
                     })
+        if updated_records:
+            data_dict['records'] = updated_records
 
 
 def validate(context: Context, data_dict: dict[str, Any]):
@@ -1621,6 +1632,11 @@ def delete_data(context: Context, data_dict: dict[str, Any]):
     validate(context, data_dict)
     fields_types = _get_fields_types(
         context['connection'], data_dict['resource_id'])
+    fields = _get_fields(context['connection'], data_dict['resource_id'])
+    field_names = _pluck('id', fields)
+    sql_columns = ", ".join(
+        identifier(name) for name in field_names)
+    return_columns = "_id, " + sql_columns
 
     query_dict: dict[str, Any] = {
         'where': []
@@ -1631,12 +1647,14 @@ def delete_data(context: Context, data_dict: dict[str, Any]):
                                              fields_types, query_dict)
 
     where_clause, where_values = _where(query_dict['where'])
-    sql_string = u'DELETE FROM "{0}" {1}'.format(
+    sql_string = u'DELETE FROM "{0}" {1} RETURNING {2}'.format(
         data_dict['resource_id'],
-        where_clause
+        where_clause,
+        return_columns
     )
 
-    _execute_single_statement(context, sql_string, where_values)
+    results = _execute_single_statement(context, sql_string, where_values)
+    data_dict['deleted_records'] = [dict(r) for r in results.mappings().all()]
 
 
 def _create_triggers(connection: Any, resource_id: str,
--- a/ckanext/datastore/logic/action.py
+++ b/ckanext/datastore/logic/action.py
@@ -67,6 +67,9 @@ def datastore_create(context: Context, data_dict: dict[str, Any]):
     :param records: the data, eg: [{"dob": "2005", "some_stuff": ["a", "b"]}]
                     (optional)
     :type records: list of dictionaries
+    :param include_records: return the inserted records in the return dict
+                            (optional)
+    :type include_records: bool
     :param primary_key: fields that represent a unique key (optional)
     :type primary_key: list or comma separated string
     :param indexes: indexes on table (optional)
@@ -181,7 +184,8 @@ def datastore_create(context: Context, data_dict: dict[str, Any]):
 
     result.pop('id', None)
     result.pop('connection_url', None)
-    result.pop('records', None)
+    if not data_dict.pop('include_records', False):
+        result.pop('records', None)
     return result
 
 
@@ -435,7 +439,7 @@ def datastore_delete(context: Context, data_dict: dict[str, Any]):
 
     **Results:**
 
-    :returns: Original filters sent.
+    :returns: Original filters sent and list of deleted_records
     :rtype: dictionary
 
     '''
--- a/ckanext/datastore/logic/schema.py
+++ b/ckanext/datastore/logic/schema.py
@@ -134,6 +134,8 @@ def datastore_create_schema() -> Schema:
                 one_of([u'row'])],
             'function': [not_empty, unicode_only],
         },
+        'include_records': [ignore_missing, default(False),
+                            boolean_validator],
         'calculate_record_count': [ignore_missing, default(False),
                                    boolean_validator],
         '__junk': [empty],


ckanext/datastore/backend/postgres.py
```diff
@@ -1217,16 +1218,18 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
             rows.append(row)
 
         sql_string = '''INSERT INTO {res_id} ({columns})
-            VALUES ({values});'''.format(
+            VALUES ({values}) RETURNING {return_columns};'''.format(
             res_id=identifier(data_dict['resource_id']),
             columns=sql_columns,
+            return_columns=return_columns,
```
wardi, line 1224: we'd want to not return the values from the sql if they're going to be ignored (include_records is False)

wardi:
This would be really useful for ckanext-dsaudit (it's missing _id in cases where it's not passed and doesn't record the data after trigger modifications)

ckanext/datastore/backend/postgres.py
JVickery-TBS: just to save on the extra sql work? yeah, makes sense

wardi: the slow bit is probably parsing and looping over the values on the python side, but yeah

changes/8684.feature
```diff
@@ -0,0 +1 @@
+`datastore_create` now accepts `include_records` to return inserted records in the return dict. `datastore_create`, `datastore_upsert`, and `datastore_delete` will now include the SQL rows from `RETURNING` in the return dict's `records` instead of the user passed list.
```
wardi, line 1: indicate that the records will now *not* be returned unless include_records=True is passed and that the values that *used* to be returned were not necessarily the actual values stored, but are now.

wardi:
Needs some new or modified tests to cover the new parameter.

ckanext/datastore/backend/postgres.py
```diff
@@ -1217,16 +1218,18 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
...
             values=', '.join([
                 f":val_{idx}" for idx in range(0, len(field_names))
             ])
         )
 
         try:
-            context['connection'].execute(sa.text(sql_string), rows)
+            results = context['connection'].execute(sa.text(sql_string), rows)
+            data_dict['records'] = [dict(r) for r in results.mappings().all()]
```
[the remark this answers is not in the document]
JVickery-TBS: @wardi currently, all of the methods in `postgres.py` do not return things, they just modify the data_dict or add to it

ckanext/datastore/backend/postgres.py
JVickery-TBS: @wardi as discussed, these actions will all default to `"include_records": False` and the sql and python loop for `records` will only happen if include_records is True

changes/8684.feature
JVickery-TBS: changelog updated to be better. hopefully??

JVickery-TBS:
@wardi okay I made some changes here now for the `include_records` for all the actions. Will work on some new tests for this

changes/8684.feature
```diff
@@ -0,0 +1 @@
+`datastore_create`, `datastore_upsert`, and `datastore_delete` now accepts `include_records` to return inserted/deleted records in the return dict, this defaults to False and `datastore_upsert` no longer includes `records` in the return dict by default. `datastore_create`, `datastore_upsert`, and `datastore_delete` with `"include_records": True` will now include the SQL rows from a psql `RETURNING` statement in the return dict's `records` (`deleted_records` for `datastore_delete`) instead of the user passed list, the returned record dicts will now include the `_id` column and all field values post database tirggers.
```
wardi, line 1:
```suggestion
`datastore_create`, `datastore_upsert`, and `datastore_delete` now accepts `include_records` to return inserted/deleted records in the return dict, this defaults to False and `datastore_upsert` no longer includes `records` in the return dict by default. `datastore_create`, `datastore_upsert`, and `datastore_delete` with `"include_records": True` will now include the actual data inserted/updated in the return dict's `records` (`deleted_records` for `datastore_delete`). The returned records now include the `_id` column and all field values after processing by any database triggers. Formerly the records returned were only a copy of the values passed by the caller.
```

ckanext/datastore/backend/postgres.py
```diff
@@ -1217,16 +1218,21 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
             rows.append(row)
 
         sql_string = '''INSERT INTO {res_id} ({columns})
-            VALUES ({values});'''.format(
+            VALUES ({values}) {return_statement};'''.format(
             res_id=identifier(data_dict['resource_id']),
             columns=sql_columns,
+            return_statement='RETURNING {return_columns}'.format(
+                return_columns=return_columns) if data_dict.get(
+                    'include_records') else '',
```
wardi, line 1226: The validation will guarantee this value exists in data_dict so it's more defensive to use `[]`.

Would be happier if `include_records` was a normal parameter, but I guess this is part of a published interface so it's harder to change?

```suggestion
                return_columns=return_columns) if data_dict[
                    'include_records'] else '',
```

ckanext/datastore/backend/postgres.py
```diff
@@ -1621,6 +1642,11 @@ def delete_data(context: Context, data_dict: dict[str, Any]):
     validate(context, data_dict)
     fields_types = _get_fields_types(
         context['connection'], data_dict['resource_id'])
+    fields = _get_fields(context['connection'], data_dict['resource_id'])
+    field_names = _pluck('id', fields)
+    sql_columns = ", ".join(
+        identifier(name) for name in field_names)
```
wardi, line 1648: `_pluck` shouldn't exist, please use something like this instead.

```suggestion
    sql_columns = ", ".join(
        identifier(f['id']) for f in fields)
```

ckanext/datastore/backend/postgres.py
```diff
@@ -1631,12 +1657,17 @@ def delete_data(context: Context, data_dict: dict[str, Any]):
                                              fields_types, query_dict)
 
     where_clause, where_values = _where(query_dict['where'])
-    sql_string = u'DELETE FROM "{0}" {1}'.format(
+    sql_string = u'DELETE FROM "{0}" {1} {2}'.format(
         data_dict['resource_id'],
```
wardi, line 1661: While we're here let's make this safer
```suggestion
    sql_string = u'DELETE FROM {0} {1} {2}'.format(
        identifier(data_dict['resource_id']),
```

ckanext/datastore/logic/action.py
```diff
@@ -67,6 +67,9 @@ def datastore_create(context: Context, data_dict: dict[str, Any]):
     :param records: the data, eg: [{"dob": "2005", "some_stuff": ["a", "b"]}]
                     (optional)
     :type records: list of dictionaries
+    :param include_records: return the inserted records in the return dict
```
wardi, line 70:
```suggestion
    :param include_records: return the full values of inserted records
```

JVickery-TBS:
@wardi all that feedback is done now, if you want to take another gander?

wardi:
With some test coverage this looks great. I'll wait until those are done and discuss this change at the next dev call before merging.

JVickery-TBS:
@wardi okay tests are in now (plus a small fix to upsert "method" and added `include_records` to the `datastore_records_delete` schema).

I THINK the one failed test is unrelated? I cannot seem to recreate the test failure locally.

EricSoroos:
Is this always going to return the final inserted version of the data -- e.g. upsert with two records containing the same primary key.

JVickery-TBS:
Nice catch @EricSoroos 
```
{
    "resource_id": "c69611a8-0514-47f4-a6f7-bc42ba1569f4",
    "records": [
        {"prim_id": "1", "example_text_field": "WIZZZZZARD"},
        {"prim_id": "1", "example_text_field": "WIZZZZZARD"}
    ],
    "method": "upsert"
}
```
Returns:
```
{
    "help": "http://127.0.0.1:5009/en/api/3/action/help_show?name=datastore_upsert",
    "success": true,
    "result": {
        "method": "upsert",
        "resource_id": "c69611a8-0514-47f4-a6f7-bc42ba1569f4",
        "records": [
            {
                "_id": 69,
                "example_text_field": "WIZZZZZARD",
                "prim_id": "1"
            },
            {
                "_id": 69,
                "example_text_field": "WIZZZZZARD",
                "prim_id": "1"
            }
        ]
    }
}
```
which is not really what I want here haha. okay will add some check to the `_id` in `upsert_data` method. I think we can just rely on the _id field as it is the true primary key

JVickery-TBS:
oh right...we have to just take the LAST upsert/update record as that is the one that will be the final DB value. Will add some new tests to cover this as well

EricSoroos:
One option for the upsert (which would greatly improve the performance) would be to do:

```
insert into table (fields) values (values)
on conflict (unique_fields_list) do 
update field=excluded.field, field2=excluded.field2, ...
```

Then if that was an insert returning, you'd get back all the info in one shot. You would need to know the unique columns on the table though, since you need that for the on conflict clause.

JVickery-TBS:
@EricSoroos okay that should be fixed now, I did it in a very very bad way. But then imagined what Ian's comment on my 20 nested loops would be, so now it is done in a good-ish pythonian way.

ohhhh so with the `INSERT INTO ... ON CONFLICT ...` would replace the 2 separate sql statements that are there now? so it would try insertion first, then update on conflict? would that work without primary index keys, or would we just use `_id` if no custom primary keys?

EricSoroos:
`INSERT ... ON CONFLICT` is postgresql's version of upsert, and it's been in since 9.4 or so.  So yes, one command. but you do need to specify which columns you're looking at for the conflict, and what is the resolution. (You can actually do things like on conflict(field) do nothing, or other options, but that's not really useful for us)

So you'd have to get the fields with a unique key, which can be done with a query against the pg metadata tables (and is currently used in the datastore pg backend for getting unique indexes, though you can't quite use that code)

```
            # check to see if there are any unique indexes
            cur.execute(
                sql.SQL("""select att.attname
                              from pg_attribute att,
                                pg_class t,
                                pg_index idx
                              where
                                t.oid = idx.indrelid
                                AND t.relkind = 'r'
                                AND idx.indisunique = 't'
                                AND t.relname = %s
                                and att.attnum in (select unnest(idx.indkey))
                                and att.attrelid=idx.indrelid"""), (resource_id,))
            unique_columns_list = cur.fetchall()
```

ckanext/datastore/backend/postgres.py
```diff
@@ -1360,33 +1360,19 @@ def upsert_data(context: Context, data_dict: dict[str, Any]):
...
                     primary_key=pk_sql,
-                    primary_value=pk_values_sql,
                 )
 
-                update_string = """
-                    UPDATE {res_id}
+                sql_string = """
+                    INSERT INTO {res_id} ({columns}) VALUES ({values})
+                    ON CONFLICT ({primary_key}) DO UPDATE
                     SET ({columns}, "_full_text") = ({values}, NULL)
```
EricSoroos, line 1368: This part here can be:
`set column=EXCLUDED.column, ... `, so you don't have to repeat the values statement, and you can do a execute_many call. Though, I guess because we're doing one insert per, that's an optimization that's really not helpful here.

JVickery-TBS: @EricSoroos for upsert/update in datastore_upsert, you are not allowed to execute it if you do not either pass `_id` or all of the unique keys, the `primary_key` in the replacement here being the sql text list of those unique keys (or just `_id`).

the `EXCLUDED` would have to be expanded for each column right? e.g. `SET (my_textfield, my_otherfield) = (et column=EXCLUDED.my_textfield, EXCLUDED.my_otherfield)`?? and this would allow it so a user does not have to pass all of the values to update a record? they could just pass the primary/unique keys and a handful of values?

JVickery-TBS: @EricSoroos okay that makes sense, I will try the above out.

Interesting issue now is coming from the failed tests, is that before you could pass `{"_id": "1"}` (as a string not int) and it would do the sql query with there `WHERE` just fine, but now I guess the `ON CONFLICT` is specific with data types(?). do you think we should just sql typecast the `_id`? and do you see any issues with custom unique constraint data types having this issue too?

JVickery-TBS: @EricSoroos okay I figured out the `EXCLUDED` stuff, as an example this is the resulting SQL statement:
```
INSERT INTO "c69611a8-0514-47f4-a6f7-bc42ba1569f4" ("example_text_field", "prim_id") VALUES (:val_0, :val_1)
                    ON CONFLICT ("prim_id") DO UPDATE
                    SET "example_text_field"=EXCLUDED."example_text_field", "prim_id"=EXCLUDED."prim_id"
                    WHERE ("c69611a8-0514-47f4-a6f7-bc42ba1569f4"."prim_id") = (:val_2)
                    RETURNING _id, "example_text_field", "prim_id"

```
with the pg values:
```
{'val_0': 'HOWDY', 'val_1': '77', 'val_2': '77'}
```
I found that we do not want to do `_id=EXCLUDED._id` as that would insert a new record when _id is not passed and another unique key is passed instead.

JVickery-TBS: @EricSoroos sorry more advanced database stuff is my downfall. I cannot seem to get this to work. Here is the newest example of the generate SQL string:
```
INSERT INTO "c69611a8-0514-47f4-a6f7-bc42ba1569f4" ("example_text_field") VALUES (:val_0)
                    ON CONFLICT ("_id") DO UPDATE
                    SET "example_text_field"=EXCLUDED."example_text_field", "prim_id"=EXCLUDED."prim_id"
                    WHERE ("c69611a8-0514-47f4-a6f7-bc42ba1569f4"."_id") = (:val_1)
                    RETURNING _id, "example_text_field", "prim_id"
```
with the sql values:
```
{'val_0': 'BLAHAHAHA', 'val_1': 103}
```
the issue is that this inserts the record:
```
{
    "_id": 128,
    "example_text_field": "BLAHAHAHA",
    "prim_id": null
}
```
which is clearly not _id=103 and the prim_id which is defined as a primary key is null...which seemingly did not get raised as a `DatabaseError -> ValidationError`

EricSoroos:
```
INSERT INTO "c69611a8-0514-47f4-a6f7-bc42ba1569f4" ("example_text_field") VALUES (:val_0)
                    ON CONFLICT ("_id") DO UPDATE
                    SET "example_text_field"=EXCLUDED."example_text_field", "prim_id"=EXCLUDED."prim_id"
                    WHERE ("c69611a8-0514-47f4-a6f7-bc42ba1569f4"."_id") = (:val_1)
                    RETURNING _id, "example_text_field", "prim_id"
```

You don't need the where -- that's on conflict (_id) part of the insert - on conflict. And I suspect that the where in this case is part of the on conflict, rather than the insert. 

Looking at this in pieces:

```INSERT INTO "c69611a8-0514-47f4-a6f7-bc42ba1569f4" ("example_text_field") VALUES (:val_0)```

This is going to always succeed, assuming a non-unique text field), because the _id is auto increment. 
so the expected result is going to be `{_id: nextval(_id_seq), example_text_field: "BLAHAHAHA", prim_id: null}`. It's never going to trigger the on conflict, so that clause is never used.

If `prim_id` has a not null constraint (as opposed to just a unique constraint), then the insert will fail, but not because of the conflict. 

So if the code was:
```
INSERT INTO "c69611a8-0514-47f4-a6f7-bc42ba1569f4" (_id, "example_text_field") VALUES (:val_1, :val_0)
                    ON CONFLICT ("_id") DO UPDATE
                    SET _id=excluded.id, 
                            "example_text_field"=EXCLUDED."example_text_field", 
                            "prim_id"=EXCLUDED."prim_id"
                    RETURNING _id, "example_text_field", "prim_id"
```

then I'd expect the following, with prim_id always null:
* _id: null, always succeeding, new row. 
* _id: value, potentially colliding, update or insert.

If prim_id was set, and supposed to be unique, then sometimes this would fail. 

If the code was:
```
INSERT INTO "c69611a8-0514-47f4-a6f7-bc42ba1569f4" (_id, "example_text_field", prim_id) VALUES (:val_1, :val_0, :val_2)
                    ON CONFLICT ("_id", prim_id) DO UPDATE
                    SET _id=excluded.id, 
                            "example_text_field"=EXCLUDED."example_text_field", 
                            "prim_id"=EXCLUDED."prim_id"
                    RETURNING _id, "example_text_field", "prim_id"
```

I'd expect:
* _id: null, prim_id=no conflict: ordinary insert
* _id: null, prim_id=conflict:  update an existing row, changing the example_text_field and setting the _id to the next value in the sequence.
* _id: value, prim_id=no conflict: update existing row _id.
* _id: value, prim_id=conflict: potentially a db error here -- if the _id and prim_id are associated with different rows, I don't know if the conflict will set one or both to the value. I actually suspect that it will error out here, and that's probably a good thing. 

I've got code that does this, but I was dealing with doing bulk uploads, so _id was never defined and I was making a temp table, copying into that, and then inserting into the final table. So similar but different use case, and _far_ faster than the individual insert/update mode.

I've got a bit of a philosophical issue here with the synthetic _id primary key -- from a db design point of view it's not really required, and if we've specified other keys, then it's actively getting in the way.  It's the worst for people who care about their schema and have already defined it properly, since it is always visible and gets in the way of some operations.

Comment (JVickery-TBS) [reply ckanext/datastore/backend/postgres.py]: @EricSoroos okkkaaayyyy I think I got it now. Code has been updated, so yeah I was missing the `_id` insert if that was provided. The code already states that if `_id` is provided, it is going to use that as the unique key and not any of the other unique keys. So I added in some conditions to add `_id` into the INSERT if that is passed. This works as expected.

I have excluded the `_id=EXCLUDED._id` from the DO UPDATE as when updating with just a passed other unique key, I don't think the _id should get updated to a new value, so the true _id primary key will stay the same in the DO UPDATE scenario. This was the old behavior so just keeping it the same.

As for passing an _id and a unique key (e.g. prim_id), a `duplicate key value violates unique constraint` error will still be raised as expected.

And I agree, when working with a fully fleshed out DataStore schema, the _id true primary key is confusing and annoying to work with. That being said, all of the other unique keys are only on the index, so not sure what the best scenario is. We have some plugins and modifications which generally exclude or workaround the _id column.

As for doing bulk update in one big SQL statement, I WAS going to state that it might be better to just do one by one as you could get partial insertions, but the DB commit does not happen unless all the records are valid, otherwise it just rolls it back. So not sure if we can manage to just make on SQL statement and if that would be an improvement at all as well???

Comment (JVickery-TBS) [reply ckanext/datastore/backend/postgres.py]: @EricSoroos so from the failing tests and local debugging, it seems that the `column=EXCLUDED.column` is going to null anything not passed? is that correct? I don't think that is the desire for the DO UPDATE

Comment (JVickery-TBS) [reply ckanext/datastore/backend/postgres.py]: I think this was a me issue, was using all the columns in the DO UPDATE SET statement instead of just the user passed ones. Will see how the pytests go now.

Comment (EricSoroos) [reply ckanext/datastore/backend/postgres.py]: I'd argue that an upsert should be a record replacement rather than a partial update. If you're inserting, you're not going to get the extra bits.

I'm not clear what this means:
> The code already states that if _id is provided, it is going to use that as the unique key and not any of the other unique keys.

If you've got a unique index on a second column, then with the table:
```
1, a, value1
2, b, value2
``` 
What should happen if I ask for an upsert for a row of:
```
1, b, value3
```

**Comment by wardi** [comment]
Hi @JVickery-TBS I let this slip, do you mind resolving the merge conflicts on this?

@EricSoroos how do you feel about these changes in their current state?

**Comment by JVickery-TBS** [comment]
@wardi conflicts resolved. Still an issue with the deleted records return for large datasets do you think? (https://github.com/open-data/ckan/pull/200)

**Comment by wardi** [comment]
Yes we'll want a hard limit on the records returned from delete because it will consume an arbitrary amount of memory, or time out and fail the operation.

We should also be consistent about returning the records as `records` from all actions, or change the delete parameter to `include_deleted_records` if you prefer returning them as `deleted_records`.

**Comment by JVickery-TBS** [comment]
@wardi okay I changed `include_records` -> `include_deleted_records` for datastore_delete and datastore_records_delete actions and schemas.

I also updated the code in the postgres.py for delete_data for the LIMIT in the returning statement. I just used the definition of `ckan.datastore.search.rows_max` but lemme know if it should just be a hard coded limit (or a new, different configurable option)

**Review by wardi — CHANGES_REQUESTED** [review]
Please resolve the conflict and I will merge

**Review by wardi — CHANGES_REQUESTED** [review]
I've resolved the conflicts but it looks like there are still some tests to update

**Comment by JVickery-TBS** [comment]
@wardi and that is why you need to keep telling me to write tests for everything. When I changed the key in the records_delete schema, I did not add a pop to the cloned dict in the psql validation method. So that is added now

**Comment by EricSoroos** [comment]
Unfortunately, this errors with dsaudit -- setting the include_records to default to false means that by default, anything passed through `datastore_upsert` is just going to be a bare response for success, and not include the records in the dictionary. 

`dsaudit/actions.py` unconditionally checks the records parameter here: https://github.com/ckan/ckanext-dsaudit/blob/master/ckanext/dsaudit/actions.py#L116
 
``` 
activity_data = {
        'fields': [
            f for f in srval['fields']
            if any(f['id'] in r for r in rval['records'])
        ],
        'records': rval['records'],
        'method': rval.get('method', 'upsert'),
        'resource_id': res.id,
    }
```

There are a couple of options:
1) dsaudit can force return values (if this is merged??) and then drop them from the response if they would otherwise be false or undefined.
2) we can just skip the dsaudit if the response isn't returned. 
3) or this patch can return [] for the records to preserve the return type.

wardi:
@EricSoroos I've added a fix to dsaudit that should address the incompatibility. If that looks good to you please go ahead and merge these changes.

PR #8698, opened 2025-02-28
Fix language code in html lang global attribute

### LINKED ISSUE #8697: Incorrect language format in global lang attribute (en_GB not en-GB)
### Discussed in https://github.com/ckan/ckan/discussions/8663

<div type='discussions-op-text'>

<sup>Originally posted by **daveytyke** February 11, 2025</sup>
Hi all,

I'm doing some accessibility work on our CKAN instance and we're getting an error on the language setup.

in our .ini file we have set -

ckan.locale_default = en_GB

This results in each web page having the following set up - 

<html lang="en_GB class= "js">

However, it should be <html lang="en-GB class= "js"> a small but significant change.

How can we make this correct?

(Changing to ckan.locale_default = en-GB crashes the service).
</div>

amercader:
Fixes #8697

Language codes in the `<html>` global tag must follow RFC 5646 ( also known as BCP 47). Region subtags are separated with a hyphen rather than an underscore which is what we use elsewhere in the code.

This just changes this particular usage as other places like the DataTables view rely on the underscore format returned by `h.lang()`


henryuzuegbu:
name tops  53357312          Top 500 hk.                  4am    Date 30/3/295.           Fps

amercader:
@EricSoroos can you check this one to see if it can be included in the next patch release?

wardi:
I'd like to see this as something like `h.unix_locale_to_bcp47(h.lang())` with this replace as an initial implementation and a fixme to do a proper conversion in the future. At least this way for users with locales that need to be mapped differently they can override the helper.

PR #8702, opened 2025-03-03
refactor: Remove model from context

smotornyuk:
* Stop using the model from context and switch to the corresponding module.
* Remove model from context's type definition
* deprecate `get_domain_object` that accepts model as an argument(used only in tests)
* Keep the model item in context for backward compatibility.


ckan/types/__init__.py
```diff
@@ -69,13 +70,12 @@ class Context(TypedDict, total=False):
...
     user: str
-    model: Model
     session: AlchemySession
 
     __auth_user_obj_checked: bool
     __auth_audit: list[tuple[str, int]]
-    auth_user_obj: Union["Model.User", "Model.AnonymousUser", None]
-    user_obj: "Model.User"
+    auth_user_obj: model.User | model.AnonymousUser | None
+    user_obj: model.User
```
pdelboca, line 78:
```suggestion
    user_obj: model.User
```

Shouldn't user_obj be as well: `model.User | model.AnonymousUser | None?

pdelboca (approved):
I'm +1 on this change. It cleans a lot of noise in our codebase, so unless @amercader has a comment on it I'm happy to merge it!

ckan/types/__init__.py
smotornyuk: No, seems like this never happens. `user_obj` is not a popular option and seems that only few actions are using it as some [sort of cache](https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L920-L923)

```py
    user_obj = model.User.get(id)
    if user_obj is None:
        raise NotFound('User was not found.')
    context['user_obj'] = user_obj
```

And as it never takes value of `current_user`, it never can be an `AnonymousUser` or `None`. At least not in existing CKAN code and there is no reason to allow abusing this property by extending it's type range.

As for `auth_user_obj`, it often [set by views](https://github.com/ckan/ckan/blob/master/ckan/views/dataset.py#L251) and takes value of the `current_user`, that's why it has to accept all corresponding types:

```py
    context: Context = {
        u'user': current_user.name,
        u'for_view': True,
        u'auth_user_obj': current_user
    }
```

amercader:
This looks great, can you fix the conflicts @smotornyuk ?

smotornyuk:
Done, and tests are green

PR #8705, opened 2025-03-05
fix: Revert deleted method get_site_statistics()

### LINKED ISSUE #8522: get_site_statistics() removed from 2.11.0
## Describe the bug

The new release of 2.11.0 removes [get_site_statistics()](https://github.com/ckan/ckan/blob/66abe426634625b9cad7a67728e481f554436412/ckan/lib/helpers.py#L2480) from the core helper. There is no explanation for why this helper was removed. Some extensions/templates use this helper to get the number of datasets/organizations.  The last release with this helper was 2.10.5. https://docs.ckan.org/en/2.10/theming/template-helper-functions.html#ckan.lib.helpers.get_site_statistics

KaterinaSmilevska:
Fixes #8522

### Features:

- [ ] includes API changes

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,11 @@ Migration notes
   the string ``*`` as value in both options (this is dangerous and
   **not** recommended).
 
+Minor changes
+-------------
+
+- Remove helper ``get_site_statistics()`` (#8705)
+
 v.2.11.2 2025-02-05
 ===================
 
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2548,15 +2548,6 @@ def featured_group_org(items: list[str], get_action: str, list_action: str,
 
     return groups_data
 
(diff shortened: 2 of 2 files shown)


amercader:
hi @KaterinaSmilevska , thanks for your PR, in https://github.com/ckan/ckan/issues/8522#issuecomment-2671691821 we decided to re-add this helper to  2.11 and update the changelog in master to mention the helper was removed. So we will need two PRs:

1. One targeting the `dev-v2.11` branch where you add back the helper like you are doing on this PR, plus a changelog entry that says something like "Re-adding xxx helper that was removed without mention in the changelog" (Changelog entries go to the `changes` folder and are named <pr number>.<change type>, in this case it can be `.minor`)
2. Another PR targeting master with just an update in the CHANGELOG.rst file in 2.12.0 -> Migration notes that says that the helper was finally removed

Let me know if this makes sense

PR #8666, opened 2025-02-12
[#8596] Fix fonts encoding after frontend build

amercader:
Fixes #8596 

There was a change in gulp 5 that defaulted the stream encoding to UTF8. This caused the font files to be corrupted.
To preserve binary files like fonts you need to pass `{encoding: false}`

Reference:

https://github.com/gulpjs/gulp/issues/2797

PR #8723, opened 2025-03-14
upgrade cypress to 14.2.0, fix intermittent failures

wardi:
`cypress run` with cypress version 13.7.2 would not launch in my local Ubuntu 22.04 environment

### Proposed fixes:
- Upgrade cypress
- add workarounds for two intermittent failures in the autocomplete cypress tests

### Features:

- [ ] includes API changes


amercader:
To avoid changing (and breaking) the fonts, changes in https://github.com/ckan/ckan/pull/8666 need to be applied

wardi:
@amercader I reverted the font changes and applied the `{"encoding": false}` fix, does that do it?

PR #8727, opened 2025-03-18
Show/Hide Resource Side Nav

JVickery-TBS:
feat(templates): resource list limit;

- Added show more/hide to resource side bar nav.

Fixes #

### Proposed fixes:

The side bar/nav list of resources can make the page extremely long if there are a lot of resources in the dataset. This just puts them into a bootstrap collapse.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2884,3 +2884,8 @@ def make_login_url(
 @core_helper
 def csrf_input():
     return snippet('snippets/csrf_input.html')
+
+
+@core_helper
+def resource_list_view_limit():
+    return int(config.get('ckan.resource_list_view_limit', 8))
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1194,6 +1194,13 @@ groups:
         description: Custom CSS directives to include on all CKAN pages.
         editable: true
 
+      - key: ckan.resource_list_view_limit
+        type: int
+        example: 15
+        default: 8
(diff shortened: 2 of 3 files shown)


amercader:
@aleeexgreeen what are your thoughts on hiding elements via collapse items from an accessibility point of view? Is it better to a add a vertical overflow (scroll bar) in the container element?

aleeexgreeen:
@amercader I would say that show more is more accessible than scrolling in this instance, as the vertical scroll on medium & big screens force users to be in a smaller area to use it, leading to accidentally scrolling the entire page rather than the section. A vertical scroll may not be obvious as well, while the button/link itself is a clear focusable element and is easily identified by assistive technologies. The collapse is also more mobile friendly because it gives user a bigger contact area rather than trying to scroll the page and being stuck inside a section or missing the hidden resources entirely.

amercader:
Thanks @aleeexgreeen , that's great feedback

ckan/lib/helpers.py
```diff
@@ -2884,3 +2884,8 @@ def make_login_url(
 @core_helper
 def csrf_input():
     return snippet('snippets/csrf_input.html')
+
+
+@core_helper
+def resource_list_view_limit():
+    return int(config.get('ckan.resource_list_view_limit', 8))
```
amercader, line 2891: The config declaration will take care of ensuring the type and assigning a default value

```suggestion
    return config['ckan.resource_list_view_limit']
```

PR #8728, opened 2025-03-19
Improve API token creation UI

avdata99:
image

image

After a user creates a API token the same page is reloaded
The form to create a new token is displayed again an it's confusing for some users.

It'll be better to remove the form and display a clearer page

### Features:

- [ ] includes API changes


avdata99:
Add this also to the new template at `ckan/templates-midnight-blue`

wardi:
Thanks @avdata99 

One problem with hiding API token page content is that it's not obvious how to go back to the list of tokens. The page can't be reloaded because it's a POST request, instead users need to click the "Create a New API Token" button (even if that's not their intent) or navigate back to the profile page then to the API token page again.

The API token is also still being presented as a flash message outside of the normal page content, which doesn't feel very elegant.

I think we have an opportunity here for a really small bit of HTMX that could solve both issues:
1. user fills out a token name and clicks "Create API Token"
2. the new API token is added to the table below the form with HTMX along with the temporarily displayed token value
3. the focus is moved to that new token (in case the window needs to be scrolled to show it) and the form is cleared

This way the new token would appear, without a page refresh, along with all the existing tokens where a user would expect it.

If the user needs more than one token they can fill out a new name and click "Create API Token" again. All the just-created tokens will be visible until the user navigates away.

What do you think?

PR #8731, opened 2025-03-21
Enable pytest on GitHub Actions and other CICD improvements

### LINKED ISSUE #8730: CICD Improvement discussion
CKAN uses a mix of GitHub actions and CircleCI.  One of the major missing elements is pytests inside GitHub actions. Without this when work is worked on a fork the that repo outside of ckan org only gets 6,000 build minutes of cicd testing and current ckan takes over 40ish minutes so giving about 150 builds a month on the free plan.

Proposal:
* Allow github to run pytests in similar fashion
* Improve workflow actions, only generate 1 per commit/pr instead of n workflows.

Possibility:
Auto generate doc's and publish to http://ckan.github.io/ckan/ for master (optional)

duttonw:
Fixes #8730

### Proposed fixes:
* Merge all testing under one parent workflow
* Pytest now runs and is split into 16 to run in under 3 mins (vs 10min for only 4 workers)
  * Allow dynamic splits to be passed in
* Allow individual workflows to still be callable (manually)
* Allow pytest to run in single instance mode to generate test duration time data for improved splits

### Features:

- [ ] includes API changes

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -41,7 +41,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: github/codeql-action/init@v2
       with:
         languages: ${{ matrix.language }}
         config-file: ./.github/codeql/codeql-config.yml
@@ -54,7 +54,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@v1
+      uses: github/codeql-action/autobuild@v2
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 https://git.io/JvXDl
@@ -68,4 +68,4 @@ jobs:
     #   make release
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v1
+      uses: github/codeql-action/analyze@v2
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,5 +1,8 @@
 name: Cypress
-on: [pull_request]
+on:
+#  pull_request:
+  workflow_call:
+  workflow_dispatch:
 env:
   NODE_VERSION: '16'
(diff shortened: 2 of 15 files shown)


duttonw:
Unsure why but https://github.com/qld-gov-au/ckan/actions/runs/13982496360 is getting green but this run is getting a failure. Is it possible to 'retry' the workflow on that file job.

wardi:
> Unsure why but https://github.com/qld-gov-au/ckan/actions/runs/13982496360 is getting green but this run is getting a failure. Is it possible to 'retry' the workflow on that file job.

I think this is just a flaky test; it failed in circle CI on another PR I was worked when I reran it.

duttonw:
With my discussions around enterprise github as well as my own org usage where we are forced to move a repo private. 

When a Repo is private, that is when things start to 'cost'. When public/public. At this time its a ok. If GitHub does change their model for Public/Open Source repo's. Then we should hopefully have another CICD supplier coming around to drop onto.

Speaking of such items. With how Docker Hub has been cheeky with their container download limits which was suppose to start march 1st but was rapidly reversed. https://news.ycombinator.com/item?id=43125089  We should also look at other places to retrieve the ckan container images. 

GitHub has such availability but you do need to use an auth token to download from their package management platform.

duttonw:
Hi @amercader @wardi,

I've polished this alot more and got it to now auto generate the pytest timing file per individual split run which is then merged and stored for future runs. We should 'ad-hoc' download this file and include back into the repo (gzipped) as the fallback.

This also now includes the pytest-retry and i've marked the 'flaking' test to give it a couple more retries before failing that segment.

With this work, we now have a single entry on a commit/PR instead of multiples independant workflows you need to look at individual. If at a later date we want to 'enforce' Flake8 and PyRight before running the full list, we can (saying that flake8 is a good contender as its super fast but PyRight takes nearly as long as pytest to complete). 

12 is a good split for speed as it now finishes pytest testing nearly at the same time as cypress.

Zharktas:
Towncrier check has seemingly got confused as the changelog entry was there before force push and then it was added again.

duttonw:
raised https://github.com/ckan/ckan/pull/8744 to fix towncrier bug.

duttonw:
I'm closing this PR and made a 'fresh' PR to fix up the 'confused' towncrier. See https://github.com/ckan/ckan/pull/8744

PR #8740, opened 2025-03-26
datastore_search: use LazyJSONObject for api view only

### LINKED ISSUE #8739: LazyJSONObject leaks
## CKAN version
master, 2.11, 2.10

## Describe the bug
LazyJSONObject is a class that mimics a dict or list but can be serialized by simplejson without decoding. It's used to speed up the `datastore_search` API by having postgres generate JSON that is returned as part of the API response without decoding and re-encoding the record data in python.

@ThHanke mentioned in his ckan monthly live presentation last week that he ran into problems with this class when extending ckan and calling `datastore_search` from python.

Python doesn't support true proxy objects so there's no way to make LazyJSONObject indistinguishable from a normal list or dict. Instead we should *only* return a LazyJSONObject when the caller is the API view and we know the result is intended to be serialized as JSON.

### Steps to reproduce
Call the `datastore_search` action from python.

### Expected behavior
The records returned should be a normal list. The LazyJSONObject performance optimization shouldn't negatively impact other users.

### Additional details

wardi:
Fixes #8739

### Proposed fixes:

Modify `datastore_search` to only use LazyJSONObject when called from the api view.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8739.bugfix
+++ b/changes/8739.bugfix
@@ -1,0 +1,1 @@
+datastore_search: return records as LazyJSONObject only when called from api view.
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -28,6 +28,7 @@ from urllib.parse import (
     urlencode, urlunparse, parse_qsl, urlparse
 )
 from io import StringIO
+import msgspec
 
 import ckan.plugins as p
 import ckan.plugins.toolkit as toolkit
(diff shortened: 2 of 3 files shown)


wardi:
@amercader I forgot the backport labels, do you think this is worth backporting?

amercader:
msgspec is a 2.11 requirement but not a 2.10 one, so I guess we can backport it to 2.11 unless you think is really worth the new req on 2.10 (or an alternative implementation)

PR #8744, opened 2025-03-26
Enable pytest on GitHub Actions and other CICD improvements

### LINKED ISSUE #8730: CICD Improvement discussion

duttonw:
Fixes #8730 #8744

### Proposed fixes:
* Merge all testing under one parent workflow
* Pytest now runs and is split into 16 to run in under 3 mins (vs 10min for only 4 workers)
  * Allow dynamic splits to be passed in
* Allow individual workflows to still be callable (manually)
* Allow pytest to run in single instance mode to generate test duration time data for improved splits

### Features:

- [ ] includes API changes

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -41,7 +41,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: github/codeql-action/init@v2
       with:
         languages: ${{ matrix.language }}
         config-file: ./.github/codeql/codeql-config.yml
@@ -54,7 +54,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@v1
+      uses: github/codeql-action/autobuild@v2
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 https://git.io/JvXDl
@@ -68,4 +68,4 @@ jobs:
     #   make release
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v1
+      uses: github/codeql-action/analyze@v2
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,5 +1,8 @@
 name: Cypress
-on: [pull_request]
+on:
+#  pull_request:
+  workflow_call:
+  workflow_dispatch:
 env:
   NODE_VERSION: '16'
   PYTHON_VERSION: '3.9'
@@ -11,8 +14,8 @@ jobs:
   cypress:
     runs-on: ubuntu-latest
     services:
-      ckan-postgres:
-        image: postgres:12
+      ckan_postgres:
+        image: postgres:15
         ports:
           - 5432:5432
         options: >-
@@ -25,11 +28,11 @@ jobs:
           POSTGRES_PASSWORD: pass
           POSTGRES_DB: ckan_test
 
-      ckan-redis:
+      ckan_redis:
         image: redis
         ports:
           - 6379:6379
-      ckan-solr:
+      ckan_solr:
         image: ckan/ckan-solr:master
         ports:
           - 8983:8983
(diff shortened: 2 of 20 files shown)


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,402 @@
...
+    needs: define-matrix
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        split: ${{ fromJSON(needs.define-matrix.outputs.SPLIT_LIST) }}
+
+    services:
+      ckan_postgres:
+        image: postgres:12 # should be postgres:15
```
duttonw, line 107: I've left this as what circleci users, but it should be on a lts supported version. 
12 is already end of support Nov 2024 
We should be on 13 (only gives us to Nov 2025). Postgres has a 5 year support window per Major Version.

https://www.postgresql.org/support/versioning/

doc/conf.py
```diff
@@ -133,12 +133,20 @@
...
+        return False
+    else:
+        return True
 
 def get_release_tags():
     git_tags = subprocess.check_output(
         ['git', 'tag', '-l'], stderr=subprocess.STDOUT).decode('utf8')
     git_tags = git_tags.split()
-    release_tags_ = [tag for tag in git_tags if tag.startswith('ckan-')]
+    release_tags_ = [tag for tag in git_tags if extract_version(tag)]
```
duttonw, line 149: This was done to fix forks which make have version data that does not follow the normal pattern. This will remove it from being in the list causing docs to not be built.

.github/workflows/cypress.yml
```diff
@@ -40,25 +43,32 @@ jobs:
...
-          ckan -c test-core-cypress.ini db init
+          ckan -c test-core-github-actions.ini db init
+          #cypress testing requires activity plugin to be active
+          ckan config-tool test-core-github-actions.ini "ckan.plugins = activity"
 
       - name: Run Cypress
         uses: cypress-io/github-action@v6
         with:
-          start: ckan -c test-core-cypress.ini run
+          start: ckan -c test-core-github-actions.ini run
```
duttonw, line 71: Updated to use github actions ini so that cypress ini can be left alone and used with docker-compose etc.

.github/workflows/codeql-analysis.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: github/codeql-action/init@v2
```
duttonw, line 44: remove deprecation warnings that were being generated.

test-infrastructure/install_deps.sh
```diff
@@ -3,6 +3,7 @@
 # OS Dependencies
 apt update
 apt install -y postgresql-client
+## MaxOS  ``brew install postgresql libmagic``
```
ThrawnCA, line 6: Misspelled "MacOS"

ThrawnCA (approved):

Zharktas:
Do we really need all those artifacts the workflow uploads ? For example do we actually have a use for merged test report as html, codecov doesn't require merging coverage reports and failed test should be visible in UI instead of downloading separate artifact to view what has failed.

duttonw:
With the merged data, you can go into the action and get the console view of the combined results. https://github.com/ckan/ckan/actions/runs/14120820459/job/39560729168 

image

image


This should also translate to console + artefacts available via ``act``  https://nektosact.com/usage/index.html  local running of the workflow. (html after its run sometimes is nicer to look at then doom scrolling logs)

Also, just in case codecov glitches, you can still get at the reports per build and make it fail locally if you indeed want a set level of coverage for the suite that was run parallel:   i.e. ``coverage report --fail-under=80``

Also this came up in my research which seems like we don't want to be connecting too closely: https://hynek.me/articles/ditch-codecov-python/


I've also not tested if the coverage for uploading the split of 12 to see if it gives the same result as the combined .coverage (sqllite dbs) and upload in one go. Your welcome to fork this branch and give it a run with the updated workflow to upload per matrix instead of at the end.

Also for all of the 'merge' data, they do expire after 1 day since its already consumed for the initial upload and final result artifact generations. (which stick around for only 30-90 days)

Zharktas:
All in all these seems quite good improvements over the old Circle CI setup, but I'll do some testing how it reports failing tests and such before merging.

Zharktas:
https://github.com/ckan/ckan/actions/runs/14195306642/attempts/1#summary-39768987303 This is expectionally nice which probably should be included in our various workflow running pytests.

duttonw:
> https://github.com/ckan/ckan/actions/runs/14195306642/attempts/1#summary-39768987303 This is expectionally nice which probably should be included in our various workflow running pytests.

The problem is that if you have too many tests fail, it won't be added to github summary. (reason why i have it create a test md file also)

I've also noticed that github won't show all summaries if there is more than 3. the rest are hidden and need to be individual clicked on.

duttonw:
Ok, I've incorporated your suggestions. 

CodeCov is called at end of each part.  

I've canned the xml generation of coverage on rollup which reduces the html generation rollup from 1min 30sec (44 combine +14 sec codecov upload )to less than 50 secs (25s combine, 0 sec no codecov upload). 

I've added the option to disable the rollup, with the above cleanup, its now an extra 50sec's for cleanup. (well 43 sinces 7 seconds is required for timing data for next run combine and upload).

To disable it is 
Key: ``IS_PYTEST_COMBINE_TEST_RESULTS_DISABLED`` Value: ``True``

This can also be done on a manual run on pytest direct.

Generally, you already know that the system is green once all pytest's pass, so you don't need to wait for the final post cleanup. 

I do believe we should come to a consensus on failing builds if test coverage drops below a certain level, say 80% it fails the build without needing us to have that part outsourced to codcov.

Zharktas (approved):

EricSoroos:
Noting that this has now forked the test-infrastructure so that we don't have local testing that tracks the CI testing environment.

duttonw:
@EricSoroos , I was hoping at least a bit of time for this to bed in and work out the way forward with local testing prior to circleci being pulled.

PR #8745, opened 2025-03-27
Files: phase 1

smotornyuk:
This is the first tiny step in a long journey of replacing file management logic. This PR contains only the minimal changes required to preserve standard behavior, to simplify the review process. I expect that the same reviewers will check all stages, and I'm trying to split changes, gradually increasing their complexity

Goals of phase 1:

* add interface for registering new storage adapters
* add base classes that can be extended when new adapter is defined
* integration with config declarations
* switch image and resource uploaders to hardcoded storages that re-use existing config options

Notes:

* The structure may change in the future phases, but I'll try to avoid big changes
* I'll add more tests and docstrings in future phases


smotornyuk:
I updated documentation, try building it locally and check changes on `maintaining/filestore.html`

Overview of all phases is described in #8920

smotornyuk:
Documentation of [file-keeper is finally published](https://datashades.github.io/file-keeper/)

ckan/config/environment.py
```diff
@@ -146,6 +152,10 @@ def update_config() -> None:
         msg = "Invalid configuration values provided:\n" + msg
         raise CkanConfigurationException(msg)
 
+    # storages rely only on valid configuration
```
amercader: Does this mean that config options from storages (e.g FsStorage's `path` or `create_path` won't be validated as validation happened a few lines before?

ckan/lib/files/base.py
```diff
@@ -0,0 +1,267 @@
...
+MultipartData: TypeAlias = fk.MultipartData
+
+
+def is_supported_type(content_type: str, supported: Iterable[str]) -> bool:
+    """Check whether content_type matches supported types.
+
+    :param content_type: tested type
+    :param supported: collection of supported types
+    """
+    maintype, subtype = content_type.split("/")
```
amercader: Is it worth adding some error handling or checks on `content_type` here (to ensure it follows  "type/format") or this has been handled before by file-keeper?

ckan/lib/files/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+                name,
+                {
+                    "type": "ckan:fs",
+                    "path": os.path.join(path, "storage", "uploads", object_type),
+                    "create_path": True,
+                    "max_size": config["ckan.max_image_size"] * 1024 * 1024,
+                },
+            )
+
+    return result
```
amercader, line 212: Shouldn't `result` include the storages created from the config options (registered with `storages.register()`), and conversely, the `resources`, `user` etc storages also registered with `storages.register()`)?

Basically it looks like the "new" and legacy storages are not tracked in the same registry.

ckan/lib/files/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # first, group config options by the storage name
+    for k, v in config.items():
+        if not k.startswith(prefix):
+            continue
+
+        # when `flat` flag is enabled, nested keys `a.b.c` kept as a string:
+        # `{"a.b.c": ...}`. When it's disabled, transform such keys into nested
+        # dictionaries `{"a": {"b": {"c": ...}}}`
+        name, *path = k[prefix_len:].split(".", 1 if flat else -1)
```
amercader, line 151: This looks like a can of worms. What's the use case for nested dicts in config settings?

ckan/lib/files/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+    config: Mapping[str, Any], prefix: str = STORAGE_PREFIX, /, flat: bool = False
+):
+    """Return settings of every storage located in the config.
+
+    Args:
+        config: mapping with configuration
+        prefix: common prefix for storage options
+
+    Keyword Args:
+        flat: do not transfrom nested keys into dictionaries
```
amercader:
```suggestion
        flat: do not transform nested keys into dictionaries
```

ckan/lib/uploader.py
```diff
@@ -98,44 +74,30 @@ def get_max_resource_size() -> int:
...
-                    raise
         self.object_type = object_type
         self.old_filename = old_filename
-        if old_filename:
-            self.old_filepath = os.path.join(self.storage_path, old_filename)
+
+        storage_name = f"{object_type}_uploads"
+        self.storage = files.storages.get(storage_name)
+        if not self.storage:
+            log.debug(
```
amercader: If the storage for this type is not configured and something is trying to upload stuff perhaps we should use `log.warning()` or higher?

ckan/config/environment.py
smotornyuk: Yes, you are right. If ConfigurationError is not raised by this moment, it means all storage settings are valid. Of course, this also means that settings **must be declared** with appropriate validators.

And, as a consequence, starting from this line it's safe to initialize storages because we can guarantee that their settings are valid according to declarations.

ckan/lib/files/base.py
smotornyuk: No, format is not verified elsewhere and you can use absolutely any string as a `content_type`. Good catch

ckan/lib/files/__init__.py
smotornyuk: I was rewriting this function multiple times and didn't notice this discrepancy. In this PR, I'll update online to register everything in the same dictionary, but `resources`, `user`, ... will override configured storages with the same name. 
In the second phase, this function is further improved, and it will check the existing configuration before overriding storage with hardcoded versions.

ckan/lib/files/__init__.py
smotornyuk: For storage, it's necessary because some of 3rd third-party adapters expect dictionaries. For example, while FS can be initialized via flat structure `{path: ..., override_existing: ...}`, azure-blob-storage expects more sophisticated input. Yes, azure actually will work even without nested dictionaries, but is you have really unique setup, you may want to configure client/transport options:
```
{
   url: str,
   key: str,
   secret: str,
   client_options: dict[str, Any],
   transport_options: dict[str, Any]
}
```
 
Additionally, indirect adapters, such as libcloud, fsspec, object_storage, or opendal, often expect the name of the driver (e.g., fs, cloud, s3, Dropbox) as a string and parameters for this driver as a nested dictionary.

That's why nested dictionaries are absolutely required. 
There is an alternative to allow a single option with JSON(which can be used in the current implementation without code changes, as long as you specify a validator that parses JSON), but I like the first style more.

Due to config declarations per option, we can validate every individual parameter before attempting to initialize the storage. With JSON as a value, we have to pass this blackbox to the storage adapter and hope that it will report incorrect settings during startup, not in runtime.

Finally, user doesn't required to keep an eye on JSON syntax:

```ini
# nested dictionary
ckan.files.storage.cloud.type = ckan:libcloud
ckan.files.storage.cloud.driver = S3
ckan.files.storage.cloud.param.host = https://aws.com
ckan.files.storage.cloud.param.key = aaa
ckan.files.storage.cloud.param.secret = bbb

# json version
ckan.files.storage.cloud.type = ckan:libcloud
ckan.files.storage.cloud.driver = S3
ckan.files.storage.cloud.param = {"host": "https://aws.com", "key": "aaa", "secret": "bbb"}
```


I actually thought that other extensions can also use this style. For example, scheming can add options per schema name, DCAT - per DCAT profile. I extracted this logic as a method of the config object in the second or third phase, but if you don't like it, I can keep it exclusive to files.

smotornyuk: An additional perspective, which cannot be used to justify current implementations, but still may be considered: with nested dictionaries(and with another ~50 lines of code), we can allow using JSON/YAML/TOML as an input format for CKAN configuration. 
Imagine that when the config file is parsed, if it came from a non-INI file, we add a new function that will `flatten` the configuration so that everything in CKAN will not notice the change. And on the config generation level, we can add support for the `--format` option to produce an initial sample in the expected format. Or even add a command that will transform one format into another.

But don't focus on this idea much, as it's pretty contradictory and has no relationship to the current PR.

ckan/lib/uploader.py
smotornyuk: Yes, agree. I used debug because previously, the uploader silently returned from this method when the path was not specified, but it's a good method to become more transparent.

ckan/lib/files/__init__.py
amercader: > For storage, it's necessary because some of 3rd third-party adapters expect dictionaries.

That makes sense . But one concern is that anything under `params` is not being validated, and I believe can not be declared in advanced right?

ckan/config/environment.py
amercader: Sorry, I'm confused, if I'm right, config options are not validated but you also say it should be safe to initialize storages because we can guarantee that their settings are valid according to declarations. So what will happen if e.g. `create_path` is not a boolean?

ckan/tests/pytest_ckan/fixtures.py
```diff
@@ -26,6 +26,7 @@
 """
 from __future__ import annotations
 
+import pathlib
```
amercader:
```suggestion
```

not used

doc/maintaining/filestore.rst
```diff
@@ -16,6 +16,9 @@ organization.
    :ref:`filestore_21_to_22_migration` for details on how to migrate). This is
    to give CKAN more control over the files and make access control possible.
 
+.. versionchanged:: 2.12
+   Add support of configurable storages.
```
amercader:
```suggestion
   Add support for configurable storages.
```

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
...
+.. _using-configured-storages:
+
+-------------------------
+Using configured storages
+-------------------------
+
+CKAN uses `file-keeper`_ as an abstraction layer for low-level interaction with
+the file storages. It exposes classes that provide standard storage interface
+disregarding of the type of underlying system. As result, saving files into
+local filesystem, cloud or DB looks exactly the same from code perspective.
```
amercader:
```suggestion
the file storages. It exposes classes that provide a standard storage interface
regardless of the underlying system. As a result, saving files into
the local files ystem, a cloud provider or a database looks exactly the same from the 
code perspective.
```

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
...
+
+Storages are initialized during application startup and must be configured in
+advance. The exact settings depend on the type of the storage, but general form
+is following::
```
amercader:
```suggestion
Storages are initialized during application startup and must be configured in
advance. The exact settings depend on the type of the storage, but in general they 
look like this::
```

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
...
+
+  ckan.files.storage.my_storage.type = ckan:fs
+  ckan.files.storage.my_storage.path = /tmp/my_storage
+  ckan.files.storage.my_storage.create_path = true
+
+Any option that starts with ``ckan.files.storage.`` is a storage
+configuration. After the prefix fillows the name of the storage,
```
amercader:
```suggestion
configuration. After the prefix follows the name of the storage,
```

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
...
+``my_storage``, and everything after the name is an option that will be
+consumed by the storage.
+
+In the example above, storage ``my_storage`` is detected with configuration
+``{"type": "ckan:fs", "path": "/tmp/my_storage", "create_path":
+true}``. Configuration for storages is grouped by the name, and that's how
+multiple storages can be configured::
```
amercader:
```suggestion
true}``. Configuration for storages is grouped by the name, and that allows
multiple storages to be configured at the same time::
```

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
...
+When the object with file details is not available, usually it can be created
+manually using location of the file and ``ckan.lib.files.FileData`` class::
+
+  path = "path/to/file/inside/the/storage.txt"
+  info = FileData(path)
+  content = storage.content(info)
+
+
+Additional information about storage functionality is available inside
+`file-keeper`_ documentation.
```
amercader:
```suggestion
To get the instance of the storage, use the``ckan.lib.files.get_storage``
function::

  storage = get_storage("my_storage")

To create a new file in the storage use its ``upload`` method and
the ``ckan.lib.files.make_upload`` function, which can transform a variety of objects into an
uploadable structure::

  upload = make_upload(b"hello world")
  info = storage.upload("file.txt", upload)

When the storage instance uploads the file, it returns an object with the file details, namely
its location, size, content type and content hash. This information is required
to read the file back from the storage::

  content = storage.content(info)

When the object with the file details is not available, it can usually be created
manually using the location of the file and the ``ckan.lib.files.FileData`` class::

  path = "path/to/file/inside/the/storage.txt"
  info = FileData(path)
  content = storage.content(info)


Additional information about storage functionality is available in the
`file-keeper`_ documentation.
```

pyproject.toml
```diff
@@ -152,6 +152,7 @@ module = [
 ]
 
 [tool.ruff]
+target-version = "py39"
```
amercader: CKAN 2.12 is >= 3.10

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
...
 .. _mimetypes: https://docs.python.org/3/library/mimetypes.html
+
+.. _file-keeper: https://pypi.org/project/file-keeper/
+
+-------------
+Storage types
+-------------
+
+Storage configuration requires ``type`` of the storage. Out of the box,
+following storage types are available:
```
amercader:
```suggestion
Configuring a storage requires defining its ``type`` of the storage. CKAN provides the following
built-in storage types:
```

doc/maintaining/filestore.rst
```diff
@@ -169,5 +174,181 @@ custom types get registered on startup::
             # ...
 
 
+.. _using-configured-storages:
+
+-------------------------
+Using configured storages
```
amercader, line 180: Maybe this is addressed on the next phases PRs but I think we should describe what an "storage" is, i.e. there's one for resources, one for user images, one for your custom thing etc

ckan/lib/files/__init__.py
smotornyuk: Basically, yes. But the real answer is longer.

 This will be cleaner in the third phase. I'm not validating the content of `params` and just verifying that `isinstance(params, dict)`. [Example](https://github.com/ckan/ckan/pull/9026/files#diff-6637cfaa8993b198e7bbff2d6960ffeb63a12dc4c949fe9d2ddc7b934a7f28c8R36-R38) from opendal adapter
```py
declaration.declare(key.params).set_description(
   "JSON object with parameters passed directly to OpenDAL operator.",
).set_validators("default({}) convert_to_json_if_string dict_only")
```

We can instead declare subkeys of `params` as
```py
decl.declare_int(key.params.timeout)
decl.declare_bool(key.params.turbo_mode)
```

to validate certain options, while the rest of params(undeclared ones) will be kept as un-validated string.

But as OpenDAL itself provides access to a number of different engines that use different `params`, I'm afraid that diferent underlying adapters may expect different type of value inside params with the same name. I.e, imagine that one adapter expects `certificate` with a path to SSL certificate, another one - the content of the certificate, and the third adapter expects true/false to enable/disable some kind of internal certificate. 

Maybe in future, when I can guarantee internal structure and expectations for the nested parameters, I'll add validation for them. But for now we cannot (or rather must not) declare it too agressively.

ckan/config/environment.py
smotornyuk: I misread your question. In the first message you wrote

> validation happened a few lines before

That's true. Validation of all config options, **including options of storages**, happens a few lines before. That's why starting from line 155, we don't need additional explicit validation of storages  -they were already validated by this moment.

As a result, if you provide the wrong option, you'll never get to line 155, because of the ConfigurationException a few lines above.

UPD. Maybe my comment was confusing. As I understand, you read it as "storage relies on valid configuration and we HOPE that it's valid", but I meant rather "CKAN guarantees that configuration is valid now and storage does need anything else apart from configuration". It's a hint for further developers if they want to move storage initialization up or down - they will know that storage can be initialized anywhere **after** the config validation, but not before.

ckan/lib/files/__init__.py
amercader: Cool, thanks for clearing that up. It's good to know that we have a way of tightening validation for these in the future

ckan/config/environment.py
amercader: All clear, thanks. I found the comment confusing, what about something like:

```suggestion
    # Declared storage config options are valid at this point, storages can be configured
```

amercader:
Overall the changes introduced in this PR look good to me.

Let's wait if @wardi or @EricSoroos want to have a look as well, but I think for the next review I'll jump straight to phase 3, keeping in mind phase 2 for context if needed, but it seems easier looking at the final thing to understand it better

doc/maintaining/filestore.rst
smotornyuk: I added few more sentences as an introduction, but yes, this is further changed in the next phases, so let's continue there.

smotornyuk:
Closing this as work is focused in #9026

PR #8747, opened 2025-03-28
chore: Update old HTTP documentations links

smotornyuk:
Replace a number of `http://` links with `https://` versions. Certain links are no longer available - I replaced them with the closest alternative


wardi:
@amercader for these to be pulled into 2.10 and 2.11 docs do we backport them the same way or do we need to update the `2.10` and `2.11` branches?

Issue #8749, opened 2025-03-28
command "ckan tracking update" runs out of memory


FuhuXia:
## CKAN version
2.11.2

## Describe the bug
app crashes when running `ckan tracking update` with message:
```
Exit status 137 (out of memory)
```

### Steps to reproduce
Run command `ckan tracking update` on catalog.data.gov. 
Our daily hits to `/_tracking` is 200+k. 

### Additional details
No such an issue on CKAN core 2.10.5

FuhuXia:
Looking at the postgresql log, it sends this query to the database:
```
SELECT tracking_summary.url, tracking_summary.package_id, tracking_summary.tracking_type, tracking_summary.count, tracking_summary.running_total, tracking_summary.recent_views, tracking_summary.tracking_date 
    FROM tracking_summary ORDER BY tracking_summary.tracking_date DESC
```
It should be fine for a small CKAN app, but it runs out of memory on our app with 20 million records in  table `tracking_summary.` 

Rewriting the query by adding the "limit 1" solved the issue:
- https://github.com/ckan/ckan/pull/8757

PR #8757, opened 2025-04-01
rewrite query to fetch one row instead of whole table

FuhuXia:
Fixes 
- #8749

### Proposed fixes:

Before the fix, the query fetches the whole table into the result set then gets the first row.
After the fix, the new query added `limit 1` and fetches one row only.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/tracking/cli/tracking.py
+++ b/ckanext/tracking/cli/tracking.py
@@ -53,8 +53,7 @@ def update_all(start_date: Optional[str] = None):
         # No date given. See when we last have data for and get data
         # from 2 days before then in case new data is available.
         # If no date here then use 2011-01-01 as the start date
-        stmt = select(ts).order_by(ts.tracking_date.desc())
-        result = session.scalars(stmt).first()
+        result = session.query(ts).order_by(desc(ts.tracking_date)).first()
         if result:
             date = result.tracking_date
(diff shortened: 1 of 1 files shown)


ckanext/tracking/cli/tracking.py
```diff
@@ -53,8 +53,7 @@ def update_all(start_date: Optional[str] = None):
         # No date given. See when we last have data for and get data
         # from 2 days before then in case new data is available.
         # If no date here then use 2011-01-01 as the start date
-        stmt = select(ts).order_by(ts.tracking_date.desc())
-        result = session.scalars(stmt).first()
+        result = session.query(ts).order_by(desc(ts.tracking_date)).first()
```
duttonw, line 56: 👍 

Since we throw away the scalars list of objects there no point in pulling it into memory when we only care about the first record. 

I'd say there is heaps of things like this since its very easy to mess up in ORM systems when it works with tiny table sizes. It should also help in reducing Postgres cpu/disk scan's getting all records which are just thrown on the floor.

amercader:
@tino097  can you check this one to see if it can be included in the next patch release?

tino097 (approved):
LGTM

tino097:
This cannot be backported in 2.10

PR #8759, opened 2025-04-02
Cancel running tests when pushed on branch

Zharktas:
### Proposed fixes:

Running tests on every commit on branch is not needed, this cancels test workflow when there is a push on a branch. On master tests are run on every commit to catch failures.

Cancel is seen here https://github.com/ckan/ckan/actions/runs/14214481840

### Features:

- [ ] includes API changes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ on:
   workflow_call:
 
 concurrency:
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }} # workflow limits it to this workflow, ref includes a pushed branch or on pull request merge branch
   cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
 
 permissions:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
   workflow_call:
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
+
 permissions:
   contents: read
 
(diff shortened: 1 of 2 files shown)


amercader:
Why don't we need to run the tests on branches? This is one of the main benefits of CI tests, to tests if changes in a working branch break existing tests. Unless I'm missing something else?

.github/workflows/test.yml
```diff
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
   workflow_call:
 
+concurrency:
+  group: ${{ github.ref }}
```
duttonw, line 9: group on branch name.

.github/workflows/test.yml
```diff
@@ -5,6 +5,10 @@ on:
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
```
duttonw, line 10: if a newer commits occurs on branch, cancel last job in progress.  (would only occur if commits are faster than 4 minutes)

duttonw (commented):
I'd suggest you also include the 2.(9|10|11|12?)-dev branches also on not cancelling.

Zharktas:
@amercader newest commit on the branch will always run. If you push new commit before the previous CI has finished, this will cancel the previous one from the branch. Building and testing every single commit just uses a lot of CI time more or less in vain. Our Circle CI builds have the same as a setting: 
image

.github/workflows/test.yml
```diff
@@ -6,7 +6,7 @@ on:
   workflow_call:
 
 concurrency:
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }} # workflow limits it to this workflow, ref includes a pushed branch or on pull request merge branch
```
Zharktas, line 9: As per docs https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

PR #8762, opened 2025-04-02
Psql connections

avdata99:
I'm seeing multiple unexpected old and idle psql connections.
After a short investigation i realize that CKAN usually do not close close connections with something like `model.Session.close()`.
This is probably OK considering that SQLAlchemy will eventually recycle them but the default value for `sqlalchemy.pool_recycle` is -1 meaning that the connections are never recycled. This can lead to problems with long-lived connections.

The PSQL pool could be improved in order to avoid unnecessary connections.  It'll be good to define `sqlalchemy.pool_recycle = 3600` as default to recycle connections after 1 hour.  

Also, this PR proposes to delete a non-required `Session.commit()` after a read-only query.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -363,6 +363,13 @@ groups:
         type: bool
         default: true
 
+      - key: sqlalchemy.pool_recycle
+        type: int
+        default: 3600
+        description: |
+          The number of seconds after which a connection is recycled. The
+          default is -1 and means that connections are never recycled.
+
       - key: sqlalchemy.<OPTION>
         type: dynamic
         description: |
--- a/ckan/model/system_info.py
+++ b/ckan/model/system_info.py
@@ -57,8 +57,6 @@ def get_system_info(key: str, default: Optional[str]=None) -> Optional[str]:
             return obj.value
     except ProgrammingError:
         meta.Session.rollback()
-    finally:
-        meta.Session.close()  # Always close
     return default
 
 
--- a/ckan/model/system_info.py
+++ b/ckan/model/system_info.py
@@ -53,11 +53,12 @@ def get_system_info(key: str, default: Optional[str]=None) -> Optional[str]:
     ''' get data from system_info table '''
     try:
         obj = meta.Session.query(SystemInfo).filter_by(key=key).first()
-        meta.Session.commit()
         if obj:
             return obj.value
     except ProgrammingError:
         meta.Session.rollback()
+    finally:
+        meta.Session.close()  # Always close
     return default
 
 
--- a/changes/8762.misc
+++ b/changes/8762.misc
@@ -1,0 +1,6 @@
+The PSQL pool could be improved in order to avoid unnecessary connections.  
+The default value for `sqlalchemy.pool_recycle` is -1 meaning that the
+connections are never recycled. This can lead to problems with long-lived
+connections. It'll be good to define `sqlalchemy.pool_recycle = 3600` as
+default to recycle connections after 1 hour.  
+Also, there is a commit after a read-only query that looks like an error.  


ckan/model/system_info.py
```diff
@@ -53,11 +53,12 @@ def get_system_info(key: str, default: Optional[str]=None) -> Optional[str]:
     ''' get data from system_info table '''
     try:
         obj = meta.Session.query(SystemInfo).filter_by(key=key).first()
-        meta.Session.commit()
         if obj:
             return obj.value
     except ProgrammingError:
         meta.Session.rollback()
+    finally:
```
duttonw, line 60: Closing session may be failing the db update in unexpected manner, tho it may also due to not having the hooks triggered if there is no ``commit()`` called at any times doing a standard get request. 
i.e. the test that is failing is about checking that 'last updated' on a user field is updated in db when doing api calls (i.e. a live key).

be aware that for normal flask operations the ``ckan/config/middleware/flask_app.py`` 
has Session cleanup so closing the session prematurely could be doing it but only switching out the logic will tell  (and now with have pytest's running on all commits when GitHub Actions is enabled, this is not as blind as it use to be)
```
def ckan_after_request(response: Response) -> Response:
    u'''Common handler executed after all Flask requests'''

    # Dispose of the SQLALchemy session
    model.Session.remove()
```


There is in the meta a ``create_local_session`` which does not have scoped sessions and won't auto clean up. The only plugin I've seen so far that uses this is datapusher and xloader and with this type of 'session' you MUST remove(which also closes). 

Then there is datastore which uses raw engine/connections outside of scope.

duttonw (commented):
What other plugins do you run when you noticed this? 

I do think the default pool_recycle is useful to keep, unsure about the full system_info alterations.

ckan/model/system_info.py
avdata99: Thanks @duttonw, I removed the session close considering it's not neccesary.

avdata99:
> What other plugins do you run when you noticed this?
> 
> I do think the default pool_recycle is useful to keep, unsure about the full system_info alterations.

These is the list of plugins @duttonw:
 
 - https://github.com/ckan/ckanext-pdfview
 - https://github.com/ckan/ckanext-scheming
 - https://github.com/ckan/ckanext-geoview
 - https://github.com/okfn/ckanext-comments
 - https://github.com/okfn/ckanext-dashboard
 - https://github.com/okfn/ckanext-saml2auth.git@v1.3.5
 - https://github.com/okfn/datapusher-plus.git@okfn_tmp
 - https://github.com/NorwegianRefugeeCouncil/ckanext-api-tracking
 - https://github.com/okfn/ckanext-announcements
 - https://github.com/unckan/ckanext-push-errors

We are going to test this setting and see if it fixes our problem.

ckan/model/system_info.py
```diff
@@ -53,7 +53,6 @@ def get_system_info(key: str, default: Optional[str]=None) -> Optional[str]:
     ''' get data from system_info table '''
     try:
         obj = meta.Session.query(SystemInfo).filter_by(key=key).first()
-        meta.Session.commit()
```
avdata99, line 56: @pdelboca it look like you add this at #6932 
Do you think it is worth keeping it?

ckan/config/config_declaration.yaml
```diff
@@ -363,6 +363,13 @@ groups:
         type: bool
         default: true
 
+      - key: sqlalchemy.pool_recycle
+        type: int
+        default: 3600
+        description: |
+          The number of seconds after which a connection is recycled. The
+          default is -1 and means that connections are never recycled.
```
wardi, line 371: So after an hour will sqlalchemy automatically close a connection leaving the web process in a broken state that will cause it to fail the next time it accesses the db?

avdata99: @wardi I'm assuming `sqlalchemy.pool_recycle` only affects idle connections (not in use).
The `sqlalchemy.pool_pre_ping` is defined as `True` (just above this proposal) so we test connections before using them.
I'm testing this is a dev environment and so far we stopped having an increasing number of idle connections.
I'm not sure who is leaving connections open (it's probably an extension) but this is probably a good idea.

duttonw: https://docs.sqlalchemy.org/en/20/core/pooling.html#setting-pool-recycle

avdata99: And what do you think about it @duttonw? 
The CKAN project in which we implemented this, we are better now.
I'm not sure if user must take care of this in their implementations or we need to define a good default value

duttonw: My production system which spikes up to 6 front of house containers and up to 2 batch servers has 260 idle (60ish in datastore, 200ish with ckan), majority are just an hour old due to wsgi rotation (i.e. processes are rotated every hour). 

i.e. In my context, each of our uwsgi's starts with min 2 processes with [16 threads](https://github.com/qld-gov-au/opswx-ckan-cookbook/blob/8232a2450b311270b234ef8005318bdce75656ef/files/default/ckan-uwsgi.ini#L27)

Based on this data and reading the docs and other places,  **SQLAlchemy pools are per-process, not shared across threads or processes.** And due to the life of the process/thread can be up for a very very long time. Having the pool-recycle would allow the system to reduce down after a peak for systems which don't impliement ``max-worker-lifetime = 3600`` in uwsgi configuration and having this set ``sqlalchemy.pool_recycle`` won't affect normal behaviour.

Other options to consider or put in our config without a default with nicer instructions then (go look at sqlachemey) are:
``sqlalchemy.pool_size`` (defaults to 5) { keep up to these always in the pool }
``sqlalchemy.max_overflow`` (defaults to 10) { allow overuse for short periods but remove once not needed }

I would suggest that https://docs.sqlalchemy.org/en/20/core/pooling.html#using-fifo-vs-lifo is not enabled. (last in first out) so idle connections can be cleaned up.

avdata99:
Closing this because each implementation will define its own database settings and probably CKAN shouldn't define this.

PR #8767, opened 2025-04-06
[Backport dev-v2.11] Enable pytest on GitHub Actions and other CICD improvements

duttonw:
Details
Original pull request	#8744 (Enable pytest on GitHub Actions and other CICD improvements)
Original pull request	#8759 (Cancel running tests when pushed on branch)
Original author	@duttonw  @amercader 
Target branch	dev-v2.11

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,6 +8,8 @@ exclude =
 
 per-file-ignores =
                  test_*.py:E501
+                 ckan/lib/helpers.py:F824
+                 ckan/lib/jobs.py:F824
 extend-exclude =
                ckan/lib/app_globals.py
                ckan/lib/cli.py
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -41,7 +41,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: github/codeql-action/init@v2
       with:
         languages: ${{ matrix.language }}
         config-file: ./.github/codeql/codeql-config.yml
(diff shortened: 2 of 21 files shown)


amercader:
@duttonw I'd like to finish tweaking the new workflows in https://github.com/ckan/ckan/pull/8783 before backporting them to 2.10 and 2.11

duttonw:
All good, we have been using this on our qld-gov-au org fork which includes changes not in upstream just yet and it works really well.  I've also tried to backport to 2.10 and ... seems there is far more 'bugs' unsure if its in upstream or in our 'changes' than i like. I'll be raising a PR to fix an invalid test, https://github.com/duttonw/ckan/commit/5531d20e7d020618d7a8f02040faaee733d28ff0 picked up on our fork since we only allow PDF's to be inlined in the browser and all other files are 'forced' to be downloaded. 


Do check the build time for this tweak, i do hope we don't get 'much' slowdown from pulling in the extra container for running vs using the 'secret sauce' system.

But good news from this change you are doing is that we won't need the 'mega' ACT local runner image as we now use our own mini container for building etc. And on that note: ACT local runner is also a bit of a interesting beast due to how it has usually pulls containers (unless explicitly told not to) and then also force pulls all pypi packages again between runs (if testing passes etc) I've not tinkered enough on setting environ variables to force it to use local docker/pypi proxy cache's to solve this.

I'm also on the fence on fully containerising with how docker hub pulled a limit as well as 'slow down' on not logged in users.
https://news.ycombinator.com/item?id=43125089
Sometimes also having instructions on how to run things locally via a VENV as well as other apps locally installed is nice also.

amercader:
Closed in favour of https://github.com/ckan/ckan/pull/8910

PR #8771, opened 2025-04-08
New blocks for the home page site search template

avdata99:
### Proposed fixes:

Just adding more blocks to the site search form template

### Features:

- [ ] includes API changes


wardi:
This is a small snippet, instead of adding new blocks it might be better to override the whole thing if a theme needs customization here.

PR #8774, opened 2025-04-09
New home page blocks for template

avdata99:
### Proposed fixes:

Just home page template blocks

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8774.misc
+++ b/changes/8774.misc
@@ -1,0 +1,1 @@
+Home page new temṕate blocks.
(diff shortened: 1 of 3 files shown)


wardi:
As discussed at the dev meeting there are too many blocks added here. We can insert contents at the end or beginning of blocks by using `{{ super() }}`. Also the purpose of the new blocks are unclear since they aren't being used anywhere within ckan.

Any new blocks we add we need to maintain or break compatibility so there should be a good reason when we add them. Closing for now but happy to consider a new PR.

PR #8781, opened 2025-04-09
feature: Enable etag and better cache control header settings

### LINKED ISSUE #8778: Better Cache control config's and etags

duttonw:
Fixes #8778

### Proposed fixes:

new config:
```
      - key: ckan.cache.expires
        default: 3600
        legacy_key: ckan.cache_expires
        type: int
        example: 2592000
        description: |

        default: 7200
        type: int
        example: 7200
        description: |

        type: int
        default: 60
        example: 3600
        description: |

        legacy_key: ckan.cache_enabled
        default: True
        type: bool
        example: "true"
        description: |
          public`` header will be added. For all other requests see ``ckan.cache.private.enabled``
          config option.

        default: true
        type: bool
        example: "true"
        description: |
          else ``Cache-Control: no-cache max-age=0``
          will be set on all requests except for sensitive responses where no-store
          will also be set.

        default: 0
        type: int
        example: 3600
        description: |
          3600 = 1 hour to serve stale content while being background revalidated

        default: 86400
        type: int
        example: 7200
        description: |

        type: bool
        default: false
        example: "true"
        description: |

      - key: ckan.etags.enabled
        type: bool
        default: true
        example: "true"
        description: |
```

new core helpers:
* cache_level
   get ``g.cache_type``
* limit_cache_for_page
* set_limit_cache_for_page
  set ``g.limit_cache_for_page
* set_cache_level
* etag_append
* set_etag_replace
* set_etag_modified_time

Fix:
* pass through to csrf system from ckan config:
  * WTF_CSRF_ENABLED
  * WTF_CSRF_TIME_LIMIT
  does not seem to be occurring and only works on ENV setup externally.

Updates:
* Vary now set on api as well as on pages, surface global ``g`` flag for extensions to use

### Features:

- [X] includes API changes


changes/8781.feature
ThrawnCA, line 1: Ordinarily the changelog ID should be an _issue_ number, not a pull request number.

ckan/tests/views/test_cache.py
```diff
@@ -0,0 +1,193 @@
...
+    with app.flask_app.request_context(env):  # only works if you have app.flask_app
+        updated_response = views.set_cache_control_headers_for_response(response)
+    assert updated_response.cache_control.public is True, updated_response
+    assert updated_response.cache_control.max_age == 3600, updated_response
+    assert updated_response.cache_control.must_revalidate is True, updated_response
+
+
+@pytest.mark.ckan_config("ckan.cache_expires", 3600)
+def test_sets_cache_control_headers_with__no_private_cache__set(app: CKANTestApp):
+    """Test that cache control headers are set correctly when caching is allowed."""
```
ThrawnCA, line 38: This comment should be different from the previous test.

ckan/views/__init__.py
```diff
@@ -43,25 +45,97 @@ def set_cors_headers_for_response(response: Response) -> Response:
...
 
 
+def set_etag_and_fast_304_response_if_unchanged(response: Response) -> Response:
+    """Set ETag and return 304 if content is unchanged."""
+
+    enable_etags = config.get(u'ckan.cache_etags', True)
+    enable_etags_not_modified = config.get(u'ckan.cache_etags_notModified', True)
+    etag = None
+
+    allowed_status_codes = {200, 404}
```
ThrawnCA, line 55: Are constants available for these codes?

Also, are there any other codes in the 2xx range that we should recognise?

And why would there be ETags on 404?

ThrawnCA (changes requested):

ckan/views/__init__.py
```diff
@@ -43,25 +46,93 @@ def set_cors_headers_for_response(response: Response) -> Response:
...
+    if response.status_code in allowed_status_codes and enable_etags:
+        if 'etag' not in response.headers:
+            # s3 etag uses md5, so using it here also
+
+            content_type = response.mimetype or ''
+            allowed_types = {'text/plain', 'text/css', 'text/html', 'application/json'}
+
+            try:
+                if content_type not in allowed_types:
+                    data_to_hash = response.get_data()
```
wardi, line 70: IIUC this will break any streaming responses e.g. `/datastore/dump/<resource_id>`

wardi:
Help me understand what's going on with etags: Is the benefit here to prevent the server only from transferring the data if the client already has a cached version based on the hash of that content? i.e. the server still generates the response every time a page is requested, it just doesn't send the full response to the client when the hash matches?

I see the code strips out specifically formatted CSRF tokens with a regex before computing the hash. How do we prevent pages that are cached from rendering with stale tokens that lead to form errors? Could we do this without a regex across the whole page content somehow?

duttonw:
Since we already send out on all pages usually with some sane defaults like ``cache-control: public, max-age=1800, must-revalidate`` or ``cache-control: private`` and if you place cdn or varnish cache in front. It will be cached for up to 30 minutes on certain pages, thankfully there is the 'must-revalidate' tag.

But the issue is that we don't provide other headers details to allow the browser nor cdn's on when it should go back for doing the staleness check.

I have also come across certain situations where the 'vary' flag in current google chrome is not honoured if there is dynamic cookies on each request (i.e. AWS ALB sticky session's). As well as modern browser updated cache strategy that skips intermediate pages and returns a full response with cookies from a session that was 'logged out' which was super fun when the next page you visit shows your 'logged out' due to old session was purged from server.


There is also not much detail about etag generation, one camp as it must be a exact byte comparable hash, on the other you are allowed to omit things which change and still keep the same etag.

I've gone with removing it, but it may still need more refinement, such as disabling for:
* e-tags for logged in users
* Known pages login/register/forgot password  pages  (these should also be on the force no cache)
* If a form is found on the page and has a csrf hidden value (via regex)
* on 'streaming' responses like ``/datastore/dump/<resource_id>``


In my research in building this, I came up with:

| Field	 | In Body? | In Headers? | Strip Before Hashing? |  Why?|
| ------------- | ------------- | ------------- | ------------- | ------------- |
| CSRF Token |	✅ Yes |	✅ Maybe |	✅  Yes |	CSRF tokens change per session/user. Including them would break deterministic | hashing. |
| Set-Cookie |	❌ No |	✅ Yes |	✅ Yes |	Session-based — changes per user. |
| Date |	❌ No |	✅ Yes |	✅ Yes |	Time-volatile — breaks caching. |
| Cache-Control, Content-Type, etc. |	❌ No |	✅ Yes |	❌|	 Usually No	These are typically constant per resource and can be  |

🛠️ Recommendations
1. Strip all user/session-specific headers and dynamic tokens before hashing.
2. If you're hashing the body only:
 * Remove CSRF tokens (e.g., hidden inputs) if they are injected dynamically.
3. If you're hashing the entire response:
 * Exclude headers like Set-Cookie, ETag, X-CSRF-Token, and Date.

References:
https://en.wikipedia.org/wiki/HTTP_ETag
https://http.dev/conditional-requests
https://http.dev/caching
https://http.dev/etag
https://github.com/rails/rails/issues/29889
https://github.com/makandra/rack-steady_etag/blob/master/lib/rack/steady_etag.rb#L20

ckan/views/__init__.py
duttonw: good to know, I'll get that addressed.

wardi:
Instead of generating etags in a python middeware that creates a copy of the response, does some regex magic replacement, decodes the result into a second copy then computes the md5sum, what about letting nginx generate it for us from its already decoded content that it's about to send?

We could replace the dynamic CSRF form field with a static HTMX tag that AJAX-es in the CSRF token after the page is rendered. This tag could also automatically refresh before the token is out of date, or instead do nothing on page load and only request the token right before forms are submitted.

This should be more efficient (no duplicate copies of the response in memory, string processing nor md5sum calculations in python) and doesn't involve any magic regexes that we have to maintain. As a bonus it would fix a separate problem we currently have with CSRF tokens going stale when a user leaves a form open for too long.

duttonw:
Seems nginx etag generation is even worse. Its just a hash of the last modified time stamp and content length. 

https://github.com/nginx/nginx/blob/stable-1.26/src/http/ngx_http_core_module.c#L1701
https://stackoverflow.com/questions/55860358/why-is-the-nginx-etag-created-from-last-modified-time-and-content-length

Looking over our Spring Framework/Spring Boot does its CSRF protection, they just use a random hash that's also stored server side in the session for the user.
https://docs.spring.io/spring-security/reference/features/exploits/csrf.html
https://www.baeldung.com/spring-security-csrf

I've not dug into our ckan does its csrf generation but i do find it odd we add the csrf token meta data even if no form is on the page that needs it.

duttonw:
Hi @wardi ,

Sadly all pages are affected due to how the footer contains the ``language_selector.html`` snippet and csrf form posts and its been like that since 2012 or longer.


Reviewing the code of that module, it should be 'GET' as its side effect free, or migrated to use javascript to do the url switch repost


https://github.com/ckan/ckan/blob/master/ckan/templates/snippets/language_selector.html
https://github.com/ckan/ckan/blob/master/ckan/templates/footer.html#L27
``
util.add_url_rule(
    '/util/redirect', view_func=internal_redirect, methods=('GET', 'POST',))
``

duttonw:
ok, i've gotten the best of both worlds until we tidy up unneeded posts on public non-logged in users.

I've included an environ flag to force non altered content checking. 

```
               etag_super_strong = u'__etag_super_strong__' in request.environ

                try:
                    # anon/public get ignored csrf
                    if (etag_super_strong
                            or content_type not in allowed_types
                            or current_user.is_authenticated or g.user):
                        data_to_hash = response.get_data()
```

as well as allow an override on skipping the make_conditional payload
```
            etag_not_conditional = u'__etag_not_conditional__' not in request.environ
            # Allow legacy behaviour if config is set
            if enable_etags_not_modified and etag_not_conditional:
                # Use built-in function now that we have an eTag
                response.make_conditional(request.environ)
```

i've made a common function to allow more direct control of the cache system and enforced the user login/reset/register to 'never cache' with 'byte' correct etags. 

I'll add a couple more tests for this as well as a tweak to private datasets which are surfaced to log in (which currently is cached as public/private) causing all sorts of ``fun`` with browser's remembering the 302 skipping the first page load and just renders the login page again in what i've seen occur on www.data.qld.gov.au


@wardi , Should this function be extended to allow max age/must revalidate flags options also? Second question, should i wrap this in a try/catch just in case its added to cli/batch code paths that don't have a valid request or this is not needed as those 'generate' a 'fake' request/response?
```
def overide_cache(etag_super_strong: bool = True,
                  no_private_cache: bool = True,
                  no_cache: bool = True):
    request.environ['__etag_super_strong__'] = etag_super_strong
    request.environ['__no_private_cache__'] = no_private_cache
    request.environ['__no_cache__'] = no_cache
```

duttonw:
> Thanks, I didn't know that nginx only applies etags for static files served (and only based on the length and last modified time) not dynamic content.

Just an FYI @wardi , with the push for 'containerisation', there is no Nginx in the stack, its uwsgi direct for both development + production mode. 

https://github.com/ckan/ckan-docker-base/blob/main/ckan-2.11/setup/start_ckan.sh#L48

If nginx was also placed after this, I'm not certain that it would add nginx anyway due to uwsgi is serving the static files also (plus dynamic building js files)

I've also not seen and documentation on ckan for nginx etag module enabling.

duttonw:
I'll need to rejig parts of this since coming across:

* https://developer.chrome.com/docs/web-platform/bfcache-ccns
* https://datatracker.ietf.org/doc/html/rfc9111 which obsoletes https://datatracker.ietf.org/doc/html/rfc7234 


Changes that need to be done
* Sensitive: ``CacheControl: no_store`` Chrome assumes the page should never be reused, even in memory, dev tools may only hold it for up to 1min before purge (if that). - Very useful for where 'cookies' are set/altered
* Shared Max Age: ``s-maxage``  - For Public cache usage so you can have a shorter maxage for browser and a slightly longer for shared cache (this is dependant on cookies being set on the requests etc)
* May also need to include ``Authorisation`` in the ``Vary`` response 
  - https://datatracker.ietf.org/doc/html/rfc9111#caching.authenticated.responses 
* Review how to not have private cache on backend edit pages and how to force revalidation on the parent page (on submit/redirect include timestamp of last modified  on query string for authors?)
* When in uwsgi only mode, for assets js/css/img/png/etc to be cached longer/independently to html pages (could use last modified?), also ensure we don't affect dynamic js/css where csrf tokens could be injected.


Then there is the nuclear option now available (Except for sarfari)
* https://www.w3.org/TR/clear-site-data/ 
     response header: ``Clear-Site-Data: "cache", "cookies", "storage", "executionContexts"``


--

For html pages, with the movement for [CSP](https://www.w3.org/TR/CSP3/) which is now mandatory on PCIDSS websites and a growing push to include it everywhere to limit malware extensions we will still need to run the html files through regex to ignore the nonce fields which are injected for dynamic pages (for public users) or enforce injection of md5 hash's of js/css thats included on pages (nonce is required for the likes of google tag manager daisy chaining other scripts from marketing teams etc)

duttonw:
Hi @wardi, @smotornyuk and @amercader,

So that we are testing properly, I've set some sane (limited to 5min/30min) for cache and now included private cache to have a revalidation period.

This is still not complete due to needing to possible dealing with disabling said cache on loopback.  Some tests are still broken since enabling public/private cache.
I also want to include the 'nuclear' header flag that majority of browsers have on 'logout' so users don't need to learn about force cache refresh commands. 

We may also need to look at query params for cache busting for authors editors so they see the 'latest' copy not up to 5 minutes old private cache on redirect. (maybe a last modified query param?)

Now that we provide etag's the browsers can now do verification without data bandwidth charges. I've also included the shared cache option. ``s-maxage`` This also works between shared cached (ala CloudFront / Cloudflare) and the backend.

---
Unsure when it crept in but sessions now occur on all requests due to 'accessing' the session even when its anon and nothing was added.  I've included a rudimentary check if we are on a 'sensitive' request, flash messages set etc..

Looking through csrf you don't use the session so unless there is a direct change occurring from public, i.e. login/register/reset OR a 'flash message'  then the anon session needs to be purged prior to giving back to wusgi/flask processing.  We should look at having the csrf verification 'lockout/timeout' be set to the max age or shared cache max age which would then not make those requests break from user (getting a coffee/lunch) distractions prior to submission.

---

I've also figured out how to mark public web assets which already hash themselves to not include the 'vary' header and to also use the 'built-in' etag generator. In 2.11 this was being set to no-cache on all responses :/. web assets also skip standard cache controls since they are handled by their own function.

Running this locally in a docker container, majority of requests back to a standard browser are 304 not modified when etags match and yes there may be more cpu overhead since no data is being passed other than response headers its really really fast.  

I've also fixed up the language selector which was making all pages have csrf tokens in forms to be a ``GET`` request.

---

@wardi  Do note that the csrf meta is not conditional and is included on all html templated pages. I've made etags not strip form hidden csrf's, but I have left in the csrf meta include stripping. (comments left in to do both if we with).
 Also for CSPv3 we will need to continue to regex strip dynamic values like nonce tokens which is now a super simple way of including any javascript on your page that you know about. (instead of needing to SHA-256 integrity hash value on all css/js ) - This may be my next thing I work on to uplift our product. 


https://content-security-policy.com/hash/

ckan/config/config_declaration.yaml
```diff
@@ -409,19 +409,60 @@ groups:
...
+          else ``Cache-Control: no-cache no-store must-revalidate max-age=0``, ``Expires: 0``
+
+      - key: ckan.cache_etags
+        type: bool
+        default: true
+        example: "true"
+        description: |
+          Add etag response header on all payloads
+
+      - key: ckan.cache_etags_notModified
```
wardi, line 459: let's stick to snake_case

ckan/config/middleware/flask_app.py
```diff
@@ -394,10 +395,23 @@ def ckan_after_request(response: Response) -> Response:
...
     response = set_cache_control_headers_for_response(response)
 
+    # Set ETag headers and unchanged response if config allows it
+    response = set_etag_and_fast_304_response_if_unchanged(response)
+
     r_time = time.time() - g.__timer
     url = request.environ['PATH_INFO']
     status_code = response.status_code
 
+    is_sensitive = u'__is_sensitive__' in request.environ
```
wardi, line 405: `__dunder__` names indicate "has internal python meaning" which doesn't apply here.  Something like `'cache_middleware_is_sensitive'` would give users a better idea of what this variable is intended for.

ckan/config/middleware/flask_app.py
```diff
@@ -561,8 +575,21 @@ def _setup_webassets(app: CKANApp):
...
+        if cache_expire == 0:
+            # If set, ``Cache-Control`` will be ``public``, otherwise
+            #         it will be ``no-cache``
+            cache_expire = None
+        response = send_from_directory(webassets_folder, path,
+                                       etag=True, max_age=cache_expire)
+
+        shared_cache_expire = config.get(u'ckan.shared_cache_expires', 0)
+        if cache_expire is not None and cache_expire != 9999999:
+            response.cache_control.s_maxage = shared_cache_expire
```
wardi, line 591: wouldn't this be better set as a `max_age` argument on `send_from_directory`?

ckan/config/middleware/flask_app.py
```diff
@@ -561,8 +575,21 @@ def _setup_webassets(app: CKANApp):
...
+        cache_expire = config.get(u'ckan.cache_expires', 0)
+        if cache_expire == 0:
+            # If set, ``Cache-Control`` will be ``public``, otherwise
+            #         it will be ``no-cache``
+            cache_expire = None
+        response = send_from_directory(webassets_folder, path,
+                                       etag=True, max_age=cache_expire)
+
+        shared_cache_expire = config.get(u'ckan.shared_cache_expires', 0)
+        if cache_expire is not None and cache_expire != 9999999:
```
wardi, line 590: where does `9999999` come from?

ckan/lib/base.py
```diff
@@ -124,12 +127,17 @@ def _allow_caching(cache_force: Optional[bool] = None):
     elif request.args.get('__no_cache__'):
         allow_cache = False
     # Don't cache if caching is not enabled in config
-    elif not config.get('ckan.cache_enabled'):
+    elif not h.asbool(config.get('ckan.cache_enabled', True)):
```
wardi, line 130: config declaration enforces the type of this value, there's no need for `asbool`

ckan/templates-midnight-blue/snippets/language_selector.html
```diff
@@ -1,5 +1,5 @@
 {% set current_lang = request.environ.CKAN_LANG %}
-<form class="lang-select" action="{% url_for 'util.internal_redirect' %}" data-module="select-switch" method="POST">
+<form class="lang-select" action="{% url_for 'util.internal_redirect' %}" data-module="select-switch" method="GET">
   {{ h.csrf_input() }}
```
wardi, line 3: we can remove this now, right?

ckan/views/__init__.py
```diff
@@ -13,8 +16,12 @@
 import ckan.plugins as p
 
 import logging
+
 log = logging.getLogger(__name__)
 
+# For more details about caching, please read the spec located at
+# https://datatracker.ietf.org/doc/html/rfc7234
```
wardi, line 23:
```suggestion
# https://datatracker.ietf.org/doc/html/rfc9111
```

ckan/views/__init__.py
```diff
@@ -43,27 +50,142 @@ def set_cors_headers_for_response(response: Response) -> Response:
     return response
 
 
+# ETag's are very useful since it allows conditional freshness checks
+# without bandwidth costs by the browser/cdn.
+# Since html pages from ckan are always generated on the fly it
+# can't use last modified and content length our Nginx does it
+def set_etag_and_fast_304_response_if_unchanged(response: Response) -> Response:
```
wardi, line 57: This middleware would be better as a separate optional plugin.

As is it assumes that bandwidth is expensive and memory/processing time on the web server are cheap. This won't be true for all users. It also adds a significant amount of code to the handling of every request.

ckan/views/__init__.py
```diff
@@ -43,27 +50,142 @@ def set_cors_headers_for_response(response: Response) -> Response:
...
+
 def set_cache_control_headers_for_response(response: Response) -> Response:
+    """ This uses the presents of request environ's: '__no_cache__',
+    '__no_private_cache__', '__limit_cache_by_cookie__' as well
+    as config variables to control cache response headers"""
+
+    is_webasset = u'__webasset__' in request.environ
+    if is_webasset:
+        # Don't alter web assets as flask_app has handled cache control for us
+        return response
```
wardi, line 129: instead of a webasset-specific check can we instead check that the response already has some cache-control headers set? That would be more generic and allow custom views to be more intelligent.

ckan/views/user.py
```diff
@@ -295,6 +296,7 @@ def api_token_revoke(id: str, jti: str) -> Response:
 
 class EditView(MethodView):
     def _prepare(self, id: Optional[str]) -> tuple[Context, str]:
+        request.environ[u'__is_sensitive__'] = True
```
wardi, line 299: passing values to middleware with request environment settings feels fragile and unrelated to response headers that will be added. Is this the best interface?

We'll also have to think about backwards compatibility for existing extensions, what will extension authors need to know about upgrading to support the next version of ckan with these changes?

ckan/lib/base.py
duttonw: will cleanup

ckan/templates-midnight-blue/snippets/language_selector.html
duttonw: prefer to leave it explicit as its against the grain of most other forms.
also missed removing the {{ h.csrf_input() }} on blue.

ckan/views/user.py
duttonw: for now, i'll leave it out and come back to this when tests are green again.

duttonw (commented):
More work will be updated shortly, moving away from environ to g space

duttonw:
@wardi , I know your busy just want you to check on the wip. (tests need a full rewrite and tweaking still needs to be completed)
https://github.com/duttonw/ckan/compare/feature/cache_tags...duttonw:d8360b43aa101d540c37c883e3a54d33bd05971f

Full diff: https://github.com/ckan/ckan/compare/master...duttonw:d8360b43aa101d540c37c883e3a54d33bd05971f 

---

I do note that we had WTF_CSRF_TIME_LIMIT in our CKAN config but it was not being configured into the csrf plugin which meant that all CSRF timeouts was defaulting to 30 minutes even if set shorter or longer. (could have worked but by indirection instead of explicitly set and we already have enough magic that makes this hard for juniors to learn)
```
app.config['WTF_CSRF_TIME_LIMIT'] = config.get('WTF_CSRF_TIME_LIMIT')
```
This should also be set to max shared/browser cache + suitable session time (say 30mins longer) for users to use the form on the page. (could be the same but the poor person who loads this page on the last second won't be able to log in etc)

---

per byte md5 plus certian csrf/nonce regex removal has now moved to its own plugin (tests still need to be written). 

Middleware interface updated to include overrides for post ckan functions (cors/cache/etag). @wardi  should I allow all plugins to run who have an interface or just the first that returns a response? ATM it allows all and if one does return a response the default function is skipped.

For ckan default etag for templates its similar to how ``send_from_directory`` does it's etag (like nginx) except that modified time is always now when generated, this still works when a cdn is in front as etag hits on shared cache can be longer than browser.
```
# s3 etag uses md5 if you want that, load etag plugin, this is weak etag
                check = (adler32(request.environ['PATH_INFO'].encode('utf-8'))
                         & 0xFFFFFFFF)
                mtime = time.time()
                size = response.content_length
                response.set_etag(f"{mtime}-{size}-{check}")
```
As we store in the page the csrf token that 'should' expire after the cache by a suitable margin and that CSPv3 nonce cache controls when a not-modified response is received does not change the headers+body in said cache. The user who has a cache page is none the wiser.

The only downside is that if private cache max age is short say 1 minute, then all revalidation will always fetch a new page even if the content (other than nonce/csrf) is the same.

----

## Cache breaking (it needs to be dicussed)

For internal call loopbacks that goes to the external url which do a 'post' then a 'get' to get the resource/package/user we will need to add to the helper functions a no_cache=timestamp where needed.

This already exists in home.py CACHE_PARAMETERS (now moved to common) 
```
This is what can be passed into the query args that is removed from dicts
CACHE_PARAMETERS = frozenset([u'__cache', u'__no_cache__'])
```

This is already in majority of views
```
data_dict = logic.clean_dict(
                dict_fns.unflatten(
                    logic.tuplize_dict(
                        logic.parse_params(req,
                                           ignore_keys=CACHE_PARAMETERS))))
```

Since this is already included, using key ``__no_cache__`` with a timestamp on calls will always provide the latest db response (instead of cdn/varnish cached object).


We also need to see if the user who has logged in is an 'editor/admin/sysadmin' on said page/view if they are able to 'alter' we should just force them to cacheType no_cache on said user so they can always see the results.  This would mean that the general public who can log in will have a cached page which will expire in due time.  (Will need to follow up testing on the 'follow'/'unfollow' sub system and how this plays with pages for logged in public.


Alternatively:
Any edit posts that redirect back to the 'view' page should also with a query string addition to get 'latest' view instead of the cached copy. 

### Brute force

There is always ways to tell certain cache/cdn providers to purge url paths. For varnish cache, this is free, but for CloudFront and others its only free up to a certain point. i.e. CloudFront is 1000 per month free then $0.005 per path(regex) submitted.

---

End of the day, we have the option for public cache and private and its never been properly tested causing my editors/authors no ends of trouble. I hope with these improvements when completed will help the more `vocal` users who know who to contact to be happier in the day to day workloads.

ckan/config/middleware/flask_app.py
wardi: @smotornyuk do we have a common way to allow different types for config options, or would we need to use a separate option for this?

ckan/views/__init__.py
wardi: if it's in the `g` object then it can be set or unset anywhere, including e.g. from some template snippet that was included in a theme.

Maybe that's useful? e.g. a theme might add a feature to a page that is dynamic and shouldn't be cached.

If it can be set anywhere should we prevent loosening the cache setting once it's set for a particular request in case it's modified in more than one place? @smotornyuk what do you think?

wardi:
Cache invalidation is notoriously one of the "two hard things in computer science"

I believe this is what we should be shooting for in core ckan:

#### webassets should be cached as "public" until the end of time
  - browsers never have to request them twice
  - caching proxies can eliminate these requests from the web server almost completely
  - any time these change a new url is generated so no cache-busting is necessary

#### static files that might change but keep the same URL use modified time/size etags
  - e.g. local file uploads, generated JS translations, JSON translation files for JS libs like datatables
  - these are fast to compute and save bandwidth between the web server and client
  - It looks like you've added this change above, that looks great

#### pages viewed by anonymous users *may optionally* be cached as "public"
  - recommend a short timeout, only for sites with heavy traffic
  - only if it's deemed acceptable for some users to be served stale pages
  - default to *off*
  - I assume this is the new `ckan.shared_cache_expires` configuration option

#### specific pages with *smart etags*
  - pages that always render the same given some initial condition can use etags to share that condition and avoid extra work
  - e.g. the dataset search page might use the package ids + `metadata_modified` values of the solr response to generate an etag
  - this gives the view an early out to return a `304 Not Modified` after performing a solr search but *before doing any work rendering the response*
  - server start time (or configuration reload time) should always be included in the etag in case a configuration change affects the output
  - extensions that modify the templates might add dynamic content that won't be part of these smart etags
    - backwards compatibility is difficult: we need to allow themes to disable or reject some smart etags
    - should we allow extensions to extend the smart etag generation and checking for existing pages?

#### everything else should not be cached at all

I want to never have to think about adding `__no_cache__` query parameters because once I do I know that the canonical URLs for pages on the site can't be trusted. That's not a happy place to be.

Let's also make sure to include the HTMX header in the Vary header because more views will support rendering just part of the page with HTMX in the future.

duttonw:
so: https://htmx.org/docs/#caching, https://htmx.org/headers/hx-trigger/

### cache overides
Templates can change the type to any in the class ``CacheType``
``g.cacheType = CacheType.NO_CACHE``
It may be useful to allow  ``g.privateMaxAge`` ``g.publicMaxAge``,  ``g.publicSharedMaxAge`` to override the ckan default config. I'e on [Open Data Qld](https://www.data.qld.gov.au/datarequest) with a datarequest public section. We may want to be public cached but on a 30 second limit and on private to be 0 on those pages.

If a template overrides that its sensitive and another puts it back to private, we may want a helper function to disallow setting to a more public cache level unless an override is given. (i.e. you can make it no_cache but cant then change it back to private) ``h.setCacheLevel(cacheType: CacheType: force: bool)``
 
```
class CacheType(Enum):
    """ This enum is to simplify cache control's since you can only have one type set
    If response cache_controls are manually configured, use context OVERRIDDEN to
    not allow default behaviour to alter your settings"""
    PUBLIC = 'public'
    PRIVATE = 'private'
    NO_CACHE = 'no-cache'
    SENSITIVE = 'no-store'
    OVERRIDDEN = 'overridden'
``` 

### Etag additions/override by g scope
how does ``g.etagAppend`` (which just adds to the end of the hash) and ``g.etagReplace`` which if found uses that like how if we find an etag on the ``response.getETag()``  we don't regenerate it.

We my may also want to use the current etag structure but change one of its setting like ``mtime`` instead of being now to being the time of last modified from the db (very useful when dealing with vanilla ckan that is not that dynamic) say that could be ``g.etagMTime``


### File freshness 
For file retrieved via s3/cloudfront with super aggressive caching even ignoring cookies when its a public file, we had had to install cache busting on the query param to allow xloader to get the 'freshest' file possible on download. As well as provide different download links when clicked. 

The way we did this was to include the etag in the url for the file, in s3's cache it was included in the metadata.
https://github.com/qld-gov-au/ckanext-s3filestore/blob/e6e2116dcb950e92e3524422b6a42d5c50e3d31b/ckanext/s3filestore/uploader.py#L282

For private asssets, we set the s3 file to private and on first redis/cache miss, store in redis (for 30min) the signed url with a 60min timelimit  to reduce calc/lookup times, with redis in the mix it goes from about 1sec to 20ms or less for private package owner datasets.

Should we look at including the computed etag on file upload to our meta data on the resource to allow similar behaviour for default file storage? i.e. download includes a hash, which allows it to go through to the server, the server ignores the etag in the param as its cache assist just in case a user or antivirus pulled the file. 

I think this would be available to us since we already are getting the file on upload and just insert the calc. We can't really use last modified as it is updated all over the place when extra metadata is inserted.

duttonw:
Hi @wardi , 

Still battling failing tests due to CSRF adding data to each request setting making 'public' cache controls difficult to test. (since if session is modified, we default to ``sensitive`` level to ensure cookies are not left in any caches, then if 'accessed' we set to ``private`` cache) 

This is very close to being completed unless you or others have input on improvements.

I really think we need to rejig how csrf is done since it should **NOT** be enabled unless a csrf input form is put through the jinja templating system. As for JS, you already have a PR open to allow HTMX to ajax request a new csrf and add to the page, we could do that via native js instead to collect a meta id if not found on the page only when interacted with which should make more pages public/public cach-able. 


Regards,

@duttonw

duttonw:
Since this branch is over 4 weeks old, i've rebased/squashed to keep it clean. @wardi  please review https://github.com/ckan/ckan/pull/8938 

I am no quite pleased with how it now allow full public cache until you visit any pages with csrf token in the form or right after a xhr call collects a csrf token for post call usage.

Alot of the 'cache private' checks are redundant since csrf trumps if user is logged in now.

.github/workflows/pytest.yml
```diff
@@ -111,12 +110,13 @@ jobs:
...
+      image: python:3.10-bookworm
 
     services:
-      ckan_postgres:
-        image: postgres:12 # should be postgres:15
-        ports:
-          - 5432:5432
+      ckan-postgres:
+
+        image: postgres:12
```
ThrawnCA: PostgreSQL 12 is out of support.

ThrawnCA (changes requested):
This seems to be incorporating a lot of unrelated changes, can that be cleaned up?

duttonw:
> This seems to be incorporating a lot of unrelated changes, can that be cleaned up?


Just pulled the debugger into its own module, i'll clean this out of this and the squashed pr in due time: https://pypi.org/project/ckanext-pycharm-debugger/

duttonw:
CSRF pulled out into https://github.com/ckan/ckan/pull/8952

duttonw:
Hi @wardi , I'm closing this PR as rework can now be found at https://github.com/ckan/ckan/pull/8954 as well as CSRF uplift in PR https://github.com/ckan/ckan/pull/8952

PR #8782, opened 2025-04-10
Fix tabledesigner integration with datatables, fix datatables localisation issues

mutantsan:
I've encountered few issues with the table designer and data tables. I understand, that those are multiple issues, but they are tightly connected.

1. Previous approach with the comparing strings doesn't include the `ckan.root_path`. This leads to a language file load error. 
2. Moving everything from `public` to `assets` seems erroneous, because we still want to expose data tables language files. 
3. Comparing string for traditional and simplified Chinese is a flawed approach as well, so the language file path determination was moved to the helper. Also, as I can see, the same error may happen with the Norwegian Bokmål locale, altough no one has reported it yet.
4. Move table designer button initialization to the end of the `InitComplete` hook as per the data tables documentation https://datatables.net/reference/option/language.url. Otherwise, the buttons won't appear if the translation is applied.
5. Add a default button class for table designer buttons, Because previously these buttons had different classes - `btn-secondary` and `btn-default`. Bow they all the same and `btn-secondary` looks bad for table designer (no highlight, bad looking disabled). The classes `btn btn-secondary` are hardcoded in the datatable.js, so I can't fully rewrite it (if I can't change the `datatable.js`). I'll leave it for someone, who wants to restyle these buttons in their projects.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8782.bugfix
+++ b/changes/8782.bugfix
@@ -1,0 +1,1 @@
+Fix tabledesigner integration with datatables and the way datatables work with the i18n files
--- a/ckanext/datatablesview/assets/datatablesview.js
+++ b/ckanext/datatablesview/assets/datatablesview.js
@@ -381,17 +381,6 @@ this.ckan.module('datatables_view', function (jQuery) {
       gtableSearchText = that._('TABLE FILTER')
       gcolFilterText = that._('COLUMN FILTER/S')
 
-      let activelanguage = languagefile
-      // en is the default language, no need to load i18n file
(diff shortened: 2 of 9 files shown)


wardi:
@mutantsan just need a changelog file for this and I'll merge

mutantsan:
@wardi, thanks for the review. I've added the changelog file.
Yes, we might want to backport it — I'll create a separate PR.

PR #8783, opened 2025-04-10
GitHub Actions tests workflow improvements

amercader:
Modify the test workflow to align it with the test-infrastructure scripts, namely:

* Run on the `python:3.10-bookworm` image instead of the github runner with a custom setup, the same used in the local test docker compose and the official Docker images
* Use the existing `install_deps.sh` and `init_environment.sh` scripts to have a consistent setup
* Use consistent service names and adapt the ini file (`test-core-github-actions.ini`) so it can be used in both environments

Other than that, cleanup the workflow file, specially around env vars set, and update the docs to reflect the Circle CI -> GitHub Actions migration


amercader:
I've switched off the Circle CI builds so hopefully that will stop the checks being run

duttonw (changes requested):
I've provided comments, some are optional (comment update etc)

The reason you are having major issues is that the coverage system is ``designed`` to be run on the same 'ecosystem' that generated the sqlite db file from pytest.

Since you are mixing container runtime and root GitHub Actions, they have different fully qualified paths and as such it unable to find the source code that generated the coverage artifact.

tl;dr: Use the same container in merging as what made the coverage database. 

Possible **Please don't do this** via trial and error change the working directory to the same path as whats in the container used for git checkout (this sadly may be a fools errand since you are not even remotely in the same path structure and it could be rejected due to clashes with other container runners)/

.github/workflows/pytest.yml
```diff
@@ -125,26 +126,13 @@ jobs:
...
-          POSTGRES_DB: ckan
-          PGDATA: "/data"
-
-      ckan_redis:
-        image: redis:7
-        ports:
-          - 6379:6379
-      ckan_solr:
+      ckan-solr:
         image: ckan/ckan-solr:master
```
duttonw, line 130: Its been on my mind that we use 'vanilla' containers for all but solr. Is it at all possible to inject via a shell script or a customer container in this repo the real 'master' def so that you are not fighting the cicd if you want to add additional config to the solr search system?

kowh-ai:
AFAICT this looks solid...will merge

PR #8784, opened 2025-04-10
[#8775] Add migration to populate empty permission labels

### LINKED ISSUE #8775: Activities missing in `package_activity_list` after 2.11 upgrade
## CKAN version

CKAN 2.11.0+

## Describe the bug
Calling  `package_activity_list` does not show any activities that were created before migrating to 2.11

### Steps to reproduce
* Create a dataset on CKAN 2.10
* Call `package_activity_list` with the dataset id: Activities are returned
* Migrate the database to CKAN 2.11
* Call `package_activity_list` with the dataset id: No activities are returned

### Additional details

In [#7885](https://github.com/ckan/ckan/pull/7885), when activities for private datasets were introduced, the `activity` table got a new column for the permission labels. These permission labels are "public" for public datasets and things like "member-{org_id}" or "collaborator-{dataset_id}" for private datasets.

#7885 included a [migration](https://github.com/ckan/ckan/pull/7885/files#diff-d1c988aeee1e4b84ede29708c556f9e5c0401f5793642940583e022ec5e25036) to add the `permission_labels` column to the `activity` table, but it didn't include any command to populate the existing records. This means that existing activities before the 2.11 migration are not matched when [filtering](https://github.com/ckan/ckan/blob/ec08bf845218dac26c38018343e242538aa0cf8e/ckanext/activity/model/activity.py#L845) the activities by the user permission labels (because they don't match "public" ).

We need to add a new migration that populates the `permission_labels` column for package activities that are null. 

If that had been done with the initial migration that would have been easy, just setting it to "public" on all cases, as all the existing activities were created on public datasets. Now that it has been a while a dataset might have changed from public to private so it is safer to compute the dataset labels based on the current dataset visibility
Something like:

```
for activity in activities:
     if not activity["activity_type"].endswith("package"):
         continue
     if activity["permission_labels"] is not null:
         continue
     pkg = model.Package.get(activity["object_id"])
     activity["permission_labels"] = get_permission_labels().get_dataset_labels(pkg)

```

Update: after discussion with @wardi, as activities don't change their permission labels when the dataset they are associated with changes its visibility and all activities pre-2.11 were created on private datasets we can follow the simpler approach of setting all dataset activities with null permission_labels to "public"

amercader:
Fixes #8775

Add the default "public" permission_label on dataset related activity records where the permission labels are null.

This will make existing activities created before a 2.11 migration to show up again in `package_activity_list` calls.

### DIFF
--- a/changes/8784.bugfix
+++ b/changes/8784.bugfix
@@ -1,0 +1,4 @@
+Add activity migration to fix existing activities created before a 2.11 migration not showing up in `package_activity_list` calls.
+It requires a database migration: 
+
+    ckan db upgrade -p activity
--- a/ckanext/activity/migration/activity/versions/fab3bfdcf830_populate_permission_labels_column.py
+++ b/ckanext/activity/migration/activity/versions/fab3bfdcf830_populate_permission_labels_column.py
@@ -1,0 +1,37 @@
+"""Populate permission_labels column
+
+Revision ID: fab3bfdcf830
+Revises: 71713a055d5c
+Create Date: 2025-04-10 12:55:05.336000
+
+"""
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "fab3bfdcf830"
+down_revision = "71713a055d5c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Set up the default "public" permission_label for dataset related activity
+    # records where the permission labels are null.
+    # See https://github.com/ckan/ckan/issues/8775
+    op.execute(
+        """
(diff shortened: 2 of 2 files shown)


changes/8784.bugfix
```diff
@@ -0,0 +1,4 @@
+Add activity migration to fix existing activities created before a 2.11 migration not showing up in `package_activity_list` calls.
+It requires a database migration: 
+
+    ckan db upgrade -p activity
```
wardi, line 4: `db upgrade` defaults to also migrate all plugins, doesn't it?

PR #8787, opened 2025-04-11
GitHub Actions tests workflow improvments - input

duttonw:
Hi @amercader ,

Here's some info you may want to take as a whole or cherry-pick parts from.

I'm happy for you to tweak/edit any of this (like the solr image on ghcr.io that is hash based for cicd testing).
Be aware that ghcr.io is org/user based and this is designed to be fork compatible (which is the reason to have dangling tag removal which is still wip since you can slowly manually cleanup via gui or via others gist scripts) 

Fixes #8783

### Proposed fixes:
* Build solr container so schema changes can be tested atomically these are stored in GitHub ghcr.io package system
  * only update if schema (hash) and solr version are different
* fix rollup of reports
* we build docs, keep them for a day for review
* extracted most bash scripts out of yaml
* reintroduced pip cache for speedup
* move most pytest options into toml file
* retrys now 0, not 1 so flaky tests are more visible. 
* .test_durations.gz updated with latest master run

### Features:

- [ ] includes API changes


pyproject.toml
```diff
@@ -53,7 +54,8 @@ markers = [
 ]
 
 testpaths = ["ckan", "ckanext"]
-addopts = "--strict-markers --pdbcls=IPython.terminal.debugger:TerminalPdb --ckan-ini=test-core.ini"
+# Set sensible defaults, cli will overide these (except for flag options
+addopts = "--strict-markers --pdbcls=IPython.terminal.debugger:TerminalPdb --ckan-ini=test-core.ini --cov=ckan --cov=ckanext --cov-branch --cov-report=term-missing --junitxml=./junit/result/junit.xml   -o junit_family=legacy --store-durations --clean-durations"
```
smotornyuk, line 58: `--cov=ckan --cov=ckanext` are great when you run all tests, but locally it can be irritating because output is bloated when you execute just a single test(if you forget to add --no-cov or --cov-reset, or empty --cov-report).
What do you think about moving `--cov` to workflow file. All other coverage related options can remain here, they are good defaults

duttonw: have moved the --cov items back to workflow COMMON. 

If i get some time, i'll see which options in cov can be left in if there is no -cov items listed.

.github/workflows/pytest.yml
```diff
@@ -15,13 +17,92 @@ on:
...
+          # We care about 2 things the solr version and the hash of the schema, this allows reuse
+          # i.e. solr9.8-2a86c4ca719150edb97bafd4dde30a60ac417d1f
+          HASH=`sha256sum ckan/config/solr/schema.xml | cut -c1-16`
+          TAG=solr${SOLR_VERSION}-${HASH}
+
+          if [[ "${IS_PR}" == "true" && "${PR_REPO}" != "${GITHUB_REPOSITORY}" ]]; then
+            ACTOR_TYPE=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            https://api.github.com/users/${GITHUB_ACTOR} | jq -r .type)
+            if [[ "$ACTOR_TYPE" == "Organization" ]]; then
+              ACTOR_API_PATH=orgs/${GITHUB_ACTOR}
```
duttonw, line 57: Security wise, PR's cross org are not allowed to deploy to ghcr.io package system unless that user is a 'member' of said org.


This could be improved to use local org ckan if the tag matches (usually it would) but if there is schema changes we won't have a tag and would need to reference the container made by the incoming user if possible. (there is some risk on this that its been 'poisoned' by a bad actor but unsure what it could do if anything)

Since schema changes are so rare we would be referencing the exact tagged version from repo owner majority of the time.

(unsure if a pr rasied by an actor would be encoded into the github_token to allow a push { it could but needs further testing} )


This is now working for PR from user to org and should also do org to org. 

It will need more testing on creating containers inside the ckan org (as well as possible settings if not enabled in the org/repo level)

Its also a pita that they made a differentiation between user and orgs for package api.

This could could be moved into a shell script and have ENV_ARGS passed through but I've spent enough time today on this and need to give it a rest.


----

I also investigated having the docker container locally built and not user a container manager but due to running the build in a container image (not root gh) it would not be possible to add it to auto dns resolution and would then need some hackery with /etc/hosts and local port forwarding to 'emulate' it being on ``ckan-solr``

PR #8792, opened 2025-04-14
fix datatables i18n file managing

mutantsan:
This related to this PR https://github.com/ckan/ckan/pull/8782

Those are the changes related to datatables view and could be backported to the version where we don't have tabledesigner. 

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8792.bugfix
+++ b/changes/8792.bugfix
@@ -1,0 +1,1 @@
+Fix the way datatables work with the i18n files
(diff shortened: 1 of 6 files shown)


wardi:
@mutantsan my mistake, the move from public to assets was only part of the 2.11 release. We can close this and backport the whole of #8782 to 2.11

PR #8804, opened 2025-04-15
Longer default CSRF token timeout setting

### LINKED ISSUE #8803: The CSRF token is invalid.
## CKAN version
master, 2.11, 2.10

## Describe the bug
`WTF_CSRF_TIME_LIMIT` defaults to 60 minutes. If you load a CKAN form then go to a nice long lunch or an uncomfortably long meeting you will get this error when you try to submit the form on your return.

### Steps to reproduce
set WTF_CSRF_TIME_LIMIT to a small value, load a page with a form like the dataset edit page, wait that number of seconds then submit the form.

### Expected behavior
CKAN should be meeting and lunch-break friendly

### Additional details

wardi:
Fixes #8803

### Proposed fixes:
- ~wrap csrf input tag in a div that uses HTMX to refresh the token just before it expires~
- render directly from `h.csrf_input()` instead of using the slow `snippet` helper function.
- default CSRF token timeout to 1 year

~If users leave a page with a form open with the default configuration this change will result in one additional (tiny) request per hour to the server. If someone more clever knows how to intercept form submissions to replace stale tokens only just before they are needed instead of every WTF_CSRF_TIME_LIMIT, please let me know.~

### Features:

- [ ] includes API changes


ckan/templates/snippets/csrf_input.html
```diff
@@ -1,2 +1,8 @@
 {# Adds a csrf_token to the forms as a hidden input. #}
+{%- if refresh_time -%}
+  <div hx-get="{{ h.url_for("util.csrf_input") }}" hx-trigger="load delay:{{ refresh_time }}s" hx-swap="outerHTML">
+{%- endif -%}
 <input type="hidden" name="{{ g.csrf_field_name }}" value="{{ csrf_token() }}"/>
```
duttonw, line 5: We could look at placing a 'not-after' meta on this field instead of a delay timer so if cached it will still get a fresh token for usage.

> Should Timestamps be Included in CSRF Tokens for Expiration?
>
>It's a common misconception to include timestamps as a value to specify the CSRF token expiration time. A CSRF >Token is not an access token. They are used to verify the authenticity of requests throughout a session, using >session information. A new session should generate a new token ([1](https://stackoverflow.com/a/30539335)).

https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html

ckan/views/util.py
```diff
@@ -39,7 +39,19 @@ def custom_form_fields() -> str:
     )
 
 
+def csrf_input() -> str:
+    '''
+    Render a hidden CSRF input field that refreshes automatically
+    so that users won't get an expired token error.
+    '''
+    # refresh just before expiration (enforce min in case of misconfiguration)
+    return base.render('snippets/csrf_input.html', {
```
duttonw, line 48: When calling this function. 

Can we either, add either;
1. a nonce param so its always unique
2. set to sensitive ``no-cache`` response cache level
3. change to a POST and ensure csrf protections are not on said function.

smotornyuk:
We discussed a bit different approach on the meeting, to reduce number of synchronization calls to server. You need to initialize periodic function either embedding JS into the page via `script` tag, or by appending code to existing file, for example `htmx-csrf.js`.

This periodic function must do 2 things:
* update value of csrf inputs(you already doing this)
* update value of csrf meta tag inside `head` tag. this tag is used for AJAX requests which will break as well when token expires.

As you need only the value of the new CSRF token, it would be better to either render just the token via new flask endpoint or maybe add another API action that returns the token. In either case, you'll need the following code for the periodic function:

```js
setInterval(
  // anonymous function that will not pollute global namespace
  async () => {
    // use sandbox to build URLs that are aware of root path
    const sandbox = ckan.sandbox();

    // get the fresh token from new flask endpoint
    const resp = await fetch(sandbox.url("page/that/renders/token/only"));
    const token = await resp.text();

    // or get it from the new API action
    const data = await new Promise((done, fail) =>
      sandbox.client.call("GET", "get_csrf_token_action", "", done, fail),
    );
    const token = data["csrf_token"];

    // get the name of the CSRF field
    const fieldName = document.head
      .querySelector("meta[name=csrf_field_name]")
      .getAttribute("content");

    // replace value of the meta tag used by AJAX requests. This tag has the
    // same name as any CSRF field
    document.querySelector(`meta[name=${name}]`).setAttribute("content", token);

    // update value of all CSRF field
    document
      .querySelectorAll(`input[name=${name}]`)
      .forEach((el) => (el.value = token));
  },

  // interval is measured in milliseconds
  1000 * 3600,
);
``` 

This is a global code and we cannot use variables to configure timeout. So we can go further and wrap it into CKAN module:

```js
ckan.module("refresh-csrf", function () {
  return {
    options: {
      timeout: 3600,
    }
    initialize() {
      setInterval(..., 1000 * this.options.timeout)
    },
  };
});
```

but you need to add this module to an element inside the body of the page to initialize it. Appending `<div data-module="refresh-csrf" data-module-timeout="{{ refresh_timeout }}"></div>` anywhere inside the body tag must do the trick.

wardi:
I tried setting the default timeout to None as indicated in https://flask-wtf.readthedocs.io/en/1.2.x/config/ but that's not actually implemented in flask-wtf so I went with 1 year instead.

I've left the changes that skip `h.render_snippet` because `render` is simpler for this one-line template and I removed two tests that depended on `h.render_snippet` being called somewhere in the rendering of the home page (`{% snippet %}` is recommended over `{{ h.render_snippet }}` for performance/caching reasons).

duttonw:
Can you also include in the  ``ckan/config/middleware/flask_app.py``
```
    wtf_key = "WTF_CSRF_SECRET_KEY"
    if not app.config.get(wtf_key):
        config[wtf_key] = app.config[wtf_key] = app.config["SECRET_KEY"]
    app.config["WTF_CSRF_FIELD_NAME"] = config.get('WTF_CSRF_FIELD_NAME')
+    app.config["WTF_CSRF_ENABLED"] = config.get('WTF_CSRF_ENABLED')
+    app.config['WTF_CSRF_TIME_LIMIT'] = config.get('WTF_CSRF_TIME_LIMIT')
    csrf.init_app(app)
```

As it looks like only python Env Vars variables are being pulled in only and not ckan's config items.

duttonw:
> I've left the changes that skip `h.render_snippet` because `render` is simpler for this one-line template and I removed two tests that depended on `h.render_snippet` being called somewhere in the rendering of the home page (`{% snippet %}` is recommended over `{{ h.render_snippet }}` for performance/caching reasons).

h.render_snippet is used in testing only, unsure if any plugins use it externally but its not referenced in the ckan core normally.

wardi:
@duttonw don

> Can you also include in the `ckan/config/middleware/flask_app.py`
> 
> ```
>     wtf_key = "WTF_CSRF_SECRET_KEY"
>     if not app.config.get(wtf_key):
>         config[wtf_key] = app.config[wtf_key] = app.config["SECRET_KEY"]
>     app.config["WTF_CSRF_FIELD_NAME"] = config.get('WTF_CSRF_FIELD_NAME')
> +    app.config["WTF_CSRF_ENABLED"] = config.get('WTF_CSRF_ENABLED')
> +    app.config['WTF_CSRF_TIME_LIMIT'] = config.get('WTF_CSRF_TIME_LIMIT')
>     csrf.init_app(app)
> ```
> 
> As it looks like only python Env Vars variables are being pulled in only and not ckan's config items.

done

smotornyuk:
Looks good. @wardi , errors from GH actions seems unrelated to your changes, right?

duttonw:
> Looks good. @wardi , errors from GH actions seems unrelated to your changes, right?

2025-04-29T07:54:17.2350894Z ##[group]Run actions/setup-python@v2

Should be on v5 now. V2 is not compatible with the latest cache and has major issues as you saw here. You could cherry pick or merge current master if you wish but, based on what I've looked at and tested in my own forks this will not cause any issues.

wardi:
@smotornyuk all green after merge from master

PR #8808, opened 2025-04-17
Manage API tokens without page refresh or flash messages

wardi:
Fixes #8728 

### Proposed fixes:
This is an alternate approach for improving the API token UI that uses HTMX to update the page and display new tokens instead of storing a token in a flash message, redirecting and reloading the page to display it to the user. It's now also possible to revoke multiple tokens at the same time.

[Screencast from 2025-04-22 04:36:56 PM.webm](https://github.com/user-attachments/assets/4f77742b-358d-49f6-9e30-16b2c4750775)

After clicking "Create API Token" one response from the server clears the form and adds the new API token to the table along with a the text that used to be a flash message at the top of the page.

This is a simple demonstration of the HTMX "Out of Band" swapping feature that lets us update multiple page elements with a single HTML response. The same out of band swap feature is used for revoking tokens: a single response sends a list of rows to delete from the table.

Also part of this PR:
- ckan now automatically initializes ckan JS modules for new content swapped in with HTMX
- `confirm-actions` ckan JS module now works with HTMX forms (see revoke dialog above)
- error responses from HTMX requests are shown in a toast message instead of being ignored
- API tokens are now returned in a consistent order from the API (oldest to newest)
- API token list now includes the created time for each token, showing the newest on top
- the error summary was removed from the Create API Token form because there's only one input field

### Features:

- [ ] includes API changes


wardi:
htmx upgrade will disappear from this PR once #8837 is merged

wardi:
@mutantsan you mentioned a fix like https://github.com/DataShades/ckanext-collection/blob/master/ckanext/collection/assets/js/htmx-init-ckan-modules.js but that doesn't re-initialize elements that have already been initialized once. What's the link for that?

mutantsan:
Hey, @wardi. This is [an example ](https://github.com/DataShades/ckanext-admin-panel/blob/master/ckanext/ap_main/assets/js/ap-htmx.js)of such script. After an initialization, we're adding a custom attribute on an element, marking it as initialized. If there was no extra swap with HTMX, the attribute will remain there and tell us, that we don't need to initialize the CKAN module again.

As about your PR:
1. I'd suggest hiding the `Revoke tokens` button, when we don't have any tokens. And a few things about how you use it:
    a. It kinda looks, like it might revoke all your token
    b. When you click it without selecting anything, just nothing happens. I don't get any response. Unlike the create button, where I know, that I have to give token a name. 
3. I can't say I really like placing the newly created token inside the table.
4. And when you submit an empty form, you got an extra error message just above the input, is this an indented behavior? For me, it's kinda self-explanatory, as we already add an error under the input.

mutantsan:
While I've been thinking about the 1b, where you'd like to get some response from a server... Maybe we need to rework the notification system in CKAN a bit? 

For example, use something like this - https://marcelodolza.github.io/iziToast/

wardi:
> Hey, @wardi. This is [an example ](https://github.com/DataShades/ckanext-admin-panel/blob/master/ckanext/ap_main/assets/js/ap-htmx.js)of such script. After an initialization, we're adding a custom attribute on an element, marking it as initialized. If there was no extra swap with HTMX, the attribute will remain there and tell us, that we don't need to initialize the CKAN module again.

Thank you that's perfect. I've added this here https://github.com/ckan/ckan/pull/8808/files#diff-e142270f118d5903f9f1793439f0fac16469a99cc8af699ef95bc31283583120R18-R35

> As about your PR:
> 
>     1. I'd suggest hiding the `Revoke tokens` button, when we don't have any tokens. And a few things about how you use it:
>        a. It kinda looks, like it might revoke all your token
>        b. When you click it without selecting anything, just nothing happens. I don't get any response. Unlike the create button, where I know, that I have to give token a name.

The revoke tokens button is now disabled until at least one token is selected.

>        c. I'd like to see a confirmation window before deleting the token.

I've fixed the confirm-actions module so it works with HTMX forms and added it back in.

>     2. I can't say I really like placing the newly created token inside the table.

This is the best approach I could think of because:
1. I want users to be able to create multiple tokens
2. I want users to be able to revoke tokens they've just created
3. When a user refreshes the page their new tokens do appear in the table so this seems like the most natural place for them

I've swapped the order so that newer tokens appear at the top right below the form (see the clip above). This way moving the focus to the new token typically doesn't have to scroll the page at all. I also think it makes more sense to see the recently created ones first.

>     3. And when you submit an empty form, you got an extra error message just above the input, is this an indented behavior? For me, it's kinda self-explanatory, as we already add an error under the input.

Agreed. I've removed the error summary from the form.

mutantsan:
@wardi looks perfect now

wardi:
Bumping this back to a draft because I want to add some generic HTMX error display code to ckan as part of the same PR

wardi:
@amercader I've updated the error display with your feedback. Now when a standard error page is returned (e.g. via a `base.abort(400, 'this was a naughty request')`) users will see just the message extracted from the page:

Screenshot from 2025-04-24 16-38-43

If the message can't be found in the response we'll display the message from HTMX which includes the request URL that might be useful for users to report to the site maintainers:

Screenshot from 2025-04-24 16-37-37

amercader:
@wardi this looks good, we can merge once the typing error is fixed.

I think the toasts design needs to be polished a bit but it can be done in a later PR.

I have a big monitor and it wasn't very apparent at first. Perhaps we can increase the delay before they disappear.

@aleeexgreeen just a heads up that we are now using this new UI element (just for errors coming from AJAX calls for now)

wardi:
Increasing the delay is a good idea. I did notice that they stay up if you hover or tap on them which is a nice touch.

PR #8823, opened 2025-04-17
Unify secondary buttons

Zharktas:
Probably due search and replace mistake during bootstrap migration, some buttons were changed from btn-default to btn-light, this PR unifies all of them to btn-secondary.

This has the following change on midnight blue templates:

https://github.com/user-attachments/assets/52d72453-db68-41e0-84ec-5c191ea802a5

https://github.com/user-attachments/assets/2dbb8440-3426-45ef-aa97-3c575fd39f54

I didn't find these actually designed in figma templates so it probably doesn't matter.

### Proposed fixes:
Unifies secondary buttons to use btn-secondary class. Makes it easier to customize the theme as the developer can define primary and secondary colors instead of primary, secondary and light colors. I'll make separate PR for 2.10 and 2.11 to change incorrectly migrated btn-light's to btn-default as btn-secondary is used only on master.

### Features:

- [ ] includes API changes


mutantsan:
Verified

PR #8906, opened 2025-04-22
prevent layout shifting in header when logo loads

### LINKED ISSUE #8905: CKAN logo in header causes layout shifting
## CKAN version
2.10, 2.11, master

## Describe the bug
When CKAN logo loads into the page the header expands shifting all page content down

https://github.com/user-attachments/assets/55029a40-787d-475c-b5b4-da5924897f5c

### Steps to reproduce
Refresh any page

### Expected behavior
No layout shifting

wardi:
Fixes #8905

### Proposed fixes:
In the spirit of https://xkcd.com/386/ here is a fix that works but is clearly wrong in the hope that someone that knows what they're doing will suggest a better fix.

~You don't want this *clearly wrong* fix in ckan master and backported to 2.10 and 2.11 do you? **Do you?**~

### Features:

- [ ] includes API changes


amercader:
Should it be `min-height` rather than `height`? Otherwise it will affect the actual image loaded (which can have different sizes). And added in the scss files (that can later be compiled with `ckan sass`)

```diff
diff --git a/ckan/public/base/scss/_masthead.scss b/ckan/public/base/scss/_masthead.scss
index 39075ec78..c95267a95 100644
--- a/ckan/public/base/scss/_masthead.scss
+++ b/ckan/public/base/scss/_masthead.scss
@@ -119,6 +119,7 @@ $notificationsBg: #C9403A;
             display: inline-block;
             img {
                 max-height: $logoMaxHeight;
+                min-height: 35px;
             }
         }
     }

```

@aleeexgreeen what do you think, is there a better approach?

wardi:
min-height won't work because the shifting is caused by the logo being taller than 35px

wardi:
The default logo is 55px and the midnight-blue one is 48px:
```
./ckan/public/base/images/ckan-logo.png PNG 122x55 122x55+0+0 8-bit sRGB 626B 0.000u 0:00.000
./ckan/public-midnight-blue/base/images/ckan-logo.png PNG 159x48 159x48+0+0 8-bit sRGB 6220B 0.000u 0:00.000
```
So do I add the min-height values to each of the `_masthead.css` files, or add configuration settings for them?

What's the best way for these images to not look blurry on high-DPI displays? Should we be using higher resolution images and force the size down to the original intended size with css?

Zharktas:
> What's the best way for these images to not look blurry on high-DPI displays? Should we be using higher resolution images and force the size down to the original intended size with css?

The proper way would be using svg file for the logo, as the vector graphics scale with the resolution. Using bigger image will fix the blurryness, but it will also use more bandwidth and more memory in the browser. I doubt the old logo exists anywhere else than those low resolution files in the repository, the new logo might exist in figma where it can be exported as svg.

wardi:
@mutantsan I've added min-height settings to the scss corresponding to the default logo sizes, how does this look?

ckan/public-midnight-blue/base/scss/_variables.scss
```diff
@@ -113,8 +114,3 @@ $iconWhiteSpritePath: "../images/bootstrap-glyphicons-halflings-white.png";
 
 // Navigation
 $zindex-navbar: 1;
-
-// Icons
-$chevronActiveIcon: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='#{$white}'><path fill-rule='evenodd' d='M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708z'/></svg>");
-$removeStateIcon: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none'><path d='M8 12H16M22 12C22 17.5228 17.5228 22 12 22C6.47715 22 2 17.5228 2 12C2 6.47715 6.47715 2 12 2C17.5228 2 22 6.47715 22 12Z' stroke='white' stroke-linecap='round' stroke-linejoin='round'/></svg>");
-$addStateIcon: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none'><path d='M12 8V16M8 12H16M22 12C22 17.5228 17.5228 22 12 22C6.47715 22 2 17.5228 2 12C2 6.47715 6.47715 2 12 2C17.5228 2 22 6.47715 22 12Z' stroke='black' stroke-linecap='round' stroke-linejoin='round'/></svg>");
```
aleeexgreeen, line 120: Can we add this back in? These were being used for the new ui facet icons.

mutantsan:
@wardi this looks fine if it helps with the default logo. I might be wrong, but forcing the logo size is okay. If someone wants to change it, because it doesn’t fit they logo, they have few ways to do it - add custom css via the  webassets, add custom css via admin, or even change the template.

wardi:
thank you for the review @mutantsan

ckan/public-midnight-blue/base/scss/_variables.scss
wardi: Sorry I didn't see this. My fault with merging from upstream I'll fix this now

wardi: Fixed in [eff406a](https://github.com/ckan/ckan/commit/eff406a98d259fde00908e418240c1a6b6905c28)

PR #8907, opened 2025-04-23
Updated datastore_info input to resource_id matching documentation

### LINKED ISSUE #8770: Documentation / Consistency Error with datastore_info
## CKAN version
2.11.2

## Describe the bug
The documentation for datastore info states that the input parameter should be 'resource_id'  which is consistent with the rest of the datastore API. The current code however uses just 'id'.

### Expected behavior
Code to match documentation

pdekraker-epa:
Fixes #8770 

### Proposed fixes:

Updates datastore_info to use resource_id as input parameter name. Preserves backwards compatibility, accepting id instead.

The fix was implemented using a __before rename like the other datastore methods

The tests were updated to use resource_id, there are not tests of backwards compatibility

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8770.misc
+++ b/changes/8770.misc
@@ -1,0 +1,1 @@
+Switch 'datastore_info' to use 'resource_id' as input argument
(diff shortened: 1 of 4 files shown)


amercader:
@wardi I created a simpler 2.11 version of this: https://github.com/ckan/ckan/pull/8916

PR #8923, opened 2025-05-03
[Midnight blue] Organization and groups pages layout

aleeexgreeen:
### Proposed fixes:
- Organization and Group pages only have one H1 for each page (within main content of page).
- Organization, Group, and Dataset search pages' H1 is in pre-primary
- UI changes to org & group pages side bar
- Org & group search helper information and add/edit member helper is in an accordion
- Removes duplicate heading in organization `bulk_process.html`
- Consistent bookmarks and content action across all org and group sub pages

### Screenshots:

image
Mobile: 
image

[Organizations](https://ckan-design-preview.galv1.links.com.au/organization/)
[Groups](https://ckan-design-preview.galv1.links.com.au/group/)

### Features:

- [ ] includes API changes


ckan/templates-midnight-blue/organization/snippets/helper.html
```diff
@@ -9,19 +9,27 @@
...
-  </h2>
-  <div class="module-content">
-    <p>
-      {% trans %}
-        CKAN Organizations are used to create, manage and publish collections
-        of datasets. Users can have different roles within an Organization,
-        depending on their level of authorisation to create, edit and publish.
-      {% endtrans %}
-    </p>
+<div class="module module-narrow module-shallow accordion">
```
wardi, line 12: since this covers all the content in the sidebar what do we gain with the accordion here? Is this just for mobile users?

On desktop I feel that if we're closing the content of the sidebar the whole bar should collapse or something, instead of creating more white space.

aleeexgreeen: I think the accordions were initially to keep the sidebar consistent across the pages, but now we have two element designs we are using in the sidebar. We could go with the sidebar boxes in #8965 or I could go back to the designers get their rationale and see if it needs to be redesigned altogether.

amercader:
Two minor comments from me:
* I don't like the accordion in the org info heading either, they are confusing with the search filters accordions. I think sidebar boxes in #8965 look fine if this is what you mean:

Screenshot 2025-05-30 at 12-00-18 Organizations - CKAN

* Regarding this:

     > Organization, Group, and Dataset search pages' H1 is in pre-primary

    I think they are a bit redundant , and I can't find them in the [Figma designs](https://www.figma.com/design/oI3wDSdZ0cHZLh6CrfYMnb/CKAN-3.0-Design-System?node-id=2397-1504&p=f&t=SjBDUQIMHg4W7SqR-0) (or it's likely that I'm looking at the wrong place in Figma). What's the rationale behind them? AFAICT they are only present in the top level Dataset/org/group pages.
   If they really need to be there perhaps we can add a bit more spacing below the header to match the one above?
Screenshot 2025-05-30 at 11-56-03 Dataset - CKAN


Other than these, this looks good to me

aleeexgreeen:
I have switched the H1s to be inside the main content. Which one do you (& the community) prefer @wardi @amercader ?

Logged in users:
image


Not logged in user:
image


The H1 can go underneath the primary action like this as well, if preferred:
image

wardi:
The H1s inside the main content does look better.

In the last example the primary action below the H1 makes more sense to me. Maybe the tabs should be below as well. The "Test Organization" tabs are displaying an aspect of the organization so it's strange for them to be above the organization title. Also the "Manage" button is most tightly related to the organization so those should be closer together.

aleeexgreeen:
I opted to move the primary action to the right side, but while talking to the designers we agreed the tabs should stay at the top of the page since it is navigation. 

image
image

ckan/views/group.py
```diff
@@ -353,6 +353,10 @@ def pager_url(q: Any = None, page: Optional[int] = None):
 def _update_facet_titles(
         facets: 'OrderedDict[str, str]',
         group_type: str) -> 'OrderedDict[str, str]':
+    if group_type == "group":
```
wardi: we should be checking the `group.is_organization` value instead because custom groups and orgs may have different type values.

ckan/templates-midnight-blue/group/index.html
```diff
@@ -7,39 +7,45 @@
...
-    {% endif %}
-  {% endblock %}
-  {% block page_pagination %}
-    {{ page.pager(q=q or '', sort=sort_by_selected or '') }}
-  {% endblock %}
+{% block primary_content %}
+  <article class="module">
+    <div class="module-content">
+      <div class="row">
+        <div {% if h.check_access('organization_create') %}class="col-md-9"{% endif %}>
```
wardi: group_create?

ckan/templates-midnight-blue/group/index.html
```diff
@@ -7,39 +7,45 @@
...
+          <h1>{% block page_heading %}{{ h.humanize_entity_type('group', group_type, 'page title') or _('Groups') }}{% endblock %}</h1>
+        </div>
+        <div class="page_primary_action col-md-3">
+          {% block page_primary_action %}
+            {% if h.check_access('organization_create') %}
```
wardi: here too

ckan/templates-midnight-blue/group/index.html
wardi: and keep the page_primary_action block in case someone is overriding it in a theme

aleeexgreeen:
I'm now wondering since we are moving the "Add new resource" button to the top of the dataset page beside the "Manage" button, whether we should consider doing the same for the "Add dataset" button on the groups and organization pages?
image
The design would be consistent, but the designs for the org, group and dataset search pages would need to change from having a primary action to having a content action.
image
@wardi wdyt?

wardi:
I like the idea. It puts all the buttons for making changes to things together.

PR #8929, opened 2025-05-08
Fix multithreaded JSON decoding issue by adding json_deserializer at engine level

TomeCirun:
# Problem Description:

In a multithreaded environment, when executing SQL queries such as:

```python
scalar(sa.text(f"EXPLAIN (VERBOSE, FORMAT JSON) {sql}"))
```
the results sometimes return as already decoded Python objects (e.g., lists), rather than as raw JSON strings. This inconsistency led to issues where the query results were not being processed in a consistent manner, especially when the data needed to be handled as JSON.

Previously, the `register_default_json` function was used to apply a JSON patch at the connection level, but this only affected a single psycopg2 connection instance, not the connections from the pool. As a result, other threads using different connections did not receive the correct JSON deserialization, leading to inconsistent results when querying the database in a multithreaded context.

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -125,26 +125,25 @@ def get_write_engine():
 
 
 def _get_engine_from_url(connection_url: str, **kwargs: Any) -> Engine:
-    '''Get either read or write engine.'''
+    """Get either read or write engine."""
     engine = _engines.get(connection_url)
-    if not engine:
-        extras = {'url': connection_url}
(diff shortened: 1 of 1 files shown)


duttonw:
For those wondering why you can't change the ``globally=False`` to true. Its due to the scope being all connections even that is in the python environment, not just this connection pool. 

We already expecting strings where this code is used so its fine to enforce on that thread pool.

https://www.psycopg.org/docs/extras.html#json-adaptation

PR #8932, opened 2025-05-09
Use empty string alt text on decorative images

toothstone:
The ["German centre for accessible reading"](https://www.dzblesen.de/auftraege/digitales) performed an accessibility audit on [our CKAN instance](https://opendata.leipzig.de), and one of the findings was that there are a lot of unnecessary alt-texts on group and organization "profile" pictures. User pages were not checked but they have the same issue imho.

[Images of limited informational value should have an empty string for their alt text](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt), to avoid clutter in assistive output.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1431,7 +1431,7 @@ def gravatar(email_hash: str,
         default = quote(default, safe='')
 
     return literal('''<img src="//gravatar.com/avatar/%s?s=%d&amp;d=%s"
-        class="user-image" width="%s" height="%s" alt="Gravatar" />'''
+        class="user-image" width="%s" height="%s" alt="" />'''
                    % (email_hash, size, default, size, size)
                    )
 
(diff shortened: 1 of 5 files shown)


amercader:
Reading the linked Mozilla docs, they mention "Icons or logos":

> Logos (such as corporate or brand logos) and informational icons should use the corresponding text in their alt strings. That is, if an image is a corporate logo, the alt text should be the name of the company

This is quite close to Group and Organization images in my opinion. I think showing the group/org name is a good replacement. Same for users, although "Gravatar" might not be the best one.

What do you think @mgifford (sorry to ping you, but your a11y expertise has been very valuable in the past)

toothstone:
> Reading the linked Mozilla docs, they mention "Icons or logos":
> 
> > Logos (such as corporate or brand logos) and informational icons should use the corresponding text in their alt strings. That is, if an image is a corporate logo, the alt text should be the name of the company
> 
> This is quite close to Group and Organization images in my opinion. I think showing the group/org name is a good replacement. Same for users, although "Gravatar" might not be the best one.

I think context needs to be taken into consideration here, if you have an interface only consisting of brand icons it of course makes sense to put the brand name in the alt text.
However, in the templates/snippets I propose changing, the name/title is already present in text form, so there is no information gain at all if we repeat them in the alt text.

Besides, there is no guarantee that users will use the profile image that way. A lot of our [organizations ](https://opendata.leipzig.de/organization/) use stock photos or other related images, with no relation to the org name.

PR #8938, opened 2025-05-12
feature: Enable etag and better cache control header settings (rebased/squashed)

### LINKED ISSUE #8778: Better Cache control config's and etags

duttonw:
Fixes #8778
Squashed from: #8781

### Proposed fixes:

new config:
```
       - key: ckan.cache.expires
        default: 3600
        legacy_key: ckan.cache_expires
        type: int
        example: 2592000
        description: |
        default: 7200
        type: int
        example: 7200
        description: |
        type: int
        default: 60
        example: 3600
        description: |
        legacy_key: ckan.cache_enabled
        default: True
        type: bool
        example: "true"
        description: |
          public`` header will be added. For all other requests see ``ckan.cache.private.enabled``
          config option.
        default: true
        type: bool
        example: "true"
        description: |
          else ``Cache-Control: no-cache max-age=0``
          will be set on all requests except for sensitive responses where no-store
          will also be set.
        default: 0
        type: int
        example: 3600
        description: |
          3600 = 1 hour to serve stale content while being background revalidated
        default: 86400
        type: int
        example: 7200
        description: |
        type: bool
        default: false
        example: "true"
        description: |
      - key: ckan.etags.enabled
        type: bool
        default: true
        example: "true"
        description: |
```

new core helpers:
* cache_level
   get ``g.cache_type``
* limit_cache_for_page
* set_limit_cache_for_page
  set ``g.limit_cache_for_page
* set_cache_level
* etag_append
* set_etag_replace
* set_etag_modified_time

Updates:
* Vary now set on api as well as on pages, surface global ``g`` flag for extensions to use

### Features:

- [X] includes API changes


duttonw:
Closing as i'm splitting this into smaller chunks for review

PR #8949, opened 2025-05-15
Ensure ckan.storage_path is always absolute.

TkTech:
Prevents common issues like webassets and the dev server using different directories ($PWD/storage and $PWD/ckan/storage).

### DIFF
--- a/ckan/cli/__init__.py
+++ b/ckan/cli/__init__.py
@@ -6,6 +6,7 @@ from typing import Any, Optional
 
 import click
 import logging
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


amercader:
This is a great idea, but it would be better implemented as a config validator for `ckan.storage_path`, i.e. adding a `validators: path_exists path_is_absolute` here:

https://github.com/ckan/ckan/blob/38c918b2a2f6ec4970439baedd346891424098d3/ckan/config/config_declaration.yaml#L1305-L1310

This way the check will be done whenever config is initialized. Validators can be added to `ckan/logic/validators/py`

PR #8952, opened 2025-05-16
Feature/CSRF enhancement only activate when used or in session

### LINKED ISSUE #8951: Enhancement: CSRF head meta only activated when CSRF session token is available
Since CSRF protection is enabled it is injected on all html pages, CSRF Meta. 

This forces a session to be created for all user's which is not required unless XHR/Fetch uses a non GET call. e.g. POST.

duttonw:
Fixes #8951

### Proposed fixes:

* JS Util library, ckan.fetchCsrfToken() to collect XHR/Fetch CSRF Tokens from page meta or ajax request get call (CORS protected)
  * client.js updated to use new util library to inject CSRF on POST requests
  * htmx updated to use new util library to inject CSRF on unsafe requests i.e. not GET, HEAD, OPTIONS.
  * ``<meta name="csrf_header_name" content="{{ g.csrf_header_name }}" />`` added to HEAD to ensure we are compatible
* Util.csrf_input added for Csrf Token collection (cors protected)
* Core helper function ``csrf_session_enabled()`` which provides a helper to jinja for dynamic csrf meta inclusion
* global variables  ``g.csrf_header_name`` ``g.csrf_enabled`` added for templates to reference.
* new snippet ``snippets/csrf_meta.html`` to contain csrf logic for XHR/Fetch 
* ``WTF_CSRF_ENABLED`` flag now fully disables generation of CSRF Tokens, not just disable verify. Stopping session creation on page requests.
* Ckan config ``WTF_CSRF_HEADERS`` now copied into app

Please note that this follows best practices on https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html  
The addition of a COR's protected endpoint for XHR/Fetch calls to collect form does not diminish said CSRF protection. For uplifted/enhanced [security CSP](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html) integrity checks on included javascript would go along way in protecting Author's sessions for XSS.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/tests/controllers/test_user.py
+++ b/ckan/tests/controllers/test_user.py
@@ -879,8 +879,63 @@ class TestUserImage(object):
 
 @pytest.mark.usefixtures("clean_db")
 class TestCSRFToken:
+
+    def test_csrf_token_get_rest_endpoint_options_header(self, app: helpers.CKANTestApp):
+        response = app.options(url_for("util.csrf_input"), headers={"Origin": "http://test.ckan.net",
+                                                                    "Accept": "application/json"})
+
+        assert "Access-Control-Allow-Origin" in response.headers
+        assert "Access-Control-Allow-Methods" in response.headers
+        assert "Access-Control-Allow-Headers" in response.headers
+        assert "Access-Control-Allow-Credentials" in response.headers
+        assert response.headers["Access-Control-Allow-Methods"] == "GET, OPTIONS"
+        assert response.headers["Access-Control-Allow-Headers"] == "Content-Type"
+        assert response.headers["Access-Control-Allow-Credentials"] == "true"
+        assert response.headers["Access-Control-Allow-Origin"] == 'http://test.ckan.net'
+
+    @pytest.mark.ckan_config("DEBUG", "true")
+    def test_csrf_token_get_rest_endpoint_options_debug_mode(self, app: helpers.CKANTestApp):
+        response = app.options(url_for("util.csrf_input"), headers={"Origin": "http://api.test.ckan.net",
+                                                                    "Accept": "application/json"})
+
+        assert "Access-Control-Allow-Origin" in response.headers
+        assert "Access-Control-Allow-Methods" in response.headers
+        assert "Access-Control-Allow-Headers" in response.headers
+        assert "Access-Control-Allow-Credentials" in response.headers
+        assert response.headers["Access-Control-Allow-Methods"] == "GET, OPTIONS"
+        assert response.headers["Access-Control-Allow-Headers"] == "Content-Type"
+        assert response.headers["Access-Control-Allow-Credentials"] == "true"
+        assert response.headers["Access-Control-Allow-Origin"] == 'http://test.ckan.net'
+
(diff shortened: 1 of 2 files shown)


ckan/public-midnight-blue/base/javascript/client.js
```diff
@@ -48,14 +48,17 @@
...
       };
-      if (type == 'POST') {
-        options.type = 'POST';
-        options.data = JSON.stringify(data);
-        var csrf_field = $('meta[name=csrf_field_name]').attr('content');
-        var csrf_token = $('meta[name='+ csrf_field +']').attr('content');
-        options.headers = {
-          'X-CSRFToken': csrf_token
-        }
+      if (type === 'POST') {
```
duttonw, line 51: I need a second set of eyes to review if this is valid promise and not skip the first try? or if it needs to be rejigged so its not ignored since its gone async style.

duttonw: Also missed: https://github.com/ckan/ckan/blob/master/ckan/public/base/javascript/htmx-ckan.js

duttonw:
I have successfully manually tested:
* The htmx CSRF injection used on ``/user/admin/api-tokens`` 'Create Api Token' sub-subsystem
* Jquery ``ckan.sandbox().client.call('POST', 'user_show', {} , function(json) { console.log(json) })``

duttonw:
@smotornyuk  @amercader , can I get any comments on this PR please.

Regards,

@duttonw

ckan/config/middleware/flask_app.py
```diff
@@ -387,8 +388,12 @@ def ckan_before_request() -> Optional[Response]:
             dest = f"{view.__module__}.{view.__name__}"
             csrf.exempt(dest)
 
-    # Set the csrf_field_name so we can use it in our templates
+    # Set the csrf_field_name, so we can use it in our templates
     g.csrf_field_name = config.get("WTF_CSRF_FIELD_NAME")
+    g.csrf_header_name = config.get("WTF_CSRF_HEADERS")[0] \
```
smotornyuk, line 393: Headers are declared as list, so there should be no situation when `isinstance` returns false. 
And if we have a list all the time, accessing the 0th element causes an index error when the list is empty.

ckan/lib/helpers.py
```diff
@@ -2908,3 +2908,16 @@ def csrf_input():
...
+
+
+@core_helper
+def csrf_session_enabled() -> bool:
+    """ Returns true if the session is enabled for CSRF protection """
+    log.debug("WTF_CSRF_FIELD_NAME: %r", config.get('WTF_CSRF_FIELD_NAME'))
+    log.debug("session: %r", session)
+
+    session_accessed = session.accessed
+    is_enabled = config.get('WTF_CSRF_FIELD_NAME') in session
```
smotornyuk, line 2920: You are already relying on flask-wtf internals to check whether token was already accessed, so how about testing `g.<WTF_CSRF_FIELD_NAME>` instead? In this way you won't need to patch session's access time

PR #8954, opened 2025-05-19
feature: Enable etag and improved cache control options

### LINKED ISSUE #8778: Better Cache control config's and etags

duttonw:
Fixes #8778

Depends-on #8952  (CSRF session creation when required)

### Proposed fixes:

For more details about caching, please read the spec located at https://datatracker.ietf.org/doc/html/rfc9111 as well as RFC 7234 + RFC 5861 + RFC 9111

Public pages will stay public until session is configured for the user, swapping to private. `Cache-Controls` request headers now honoured on response allowing 'cache busting' (dependant on other modules/cdn's being correctly configured)

Any page request which set cookies or the session is modified (cookie based session) will be forced to 'not cache' to ensure we do not pollute cache's with invalid cookies.

This also allows browser ``If-None-Match`` ability to query the server if it is still valid reducing bandwidth transfers and when in combinations with cdn/varnish cache's in the picture will also reduce cpu.

Web assets now configured with etags and set to not include 'vary: cookie' and is public cached and this is 100% fine since python web assets already include hash sum in its filenames.

new config:
```
      - key: ckan.cache.expires
        default: 3600
        legacy_key: ckan.cache_expires  # since v.2.12.0
        type: int
        example: 2592000
        description: |

        default: 7200
        type: int
        example: 7200
        description: |

        type: int
        default: 60
        example: 3600
        description: |

        legacy_key: ckan.cache_enabled  # since v.2.12.0
        default: True
        type: bool
        example: "true"
        description: |
          public`` header will be added. For all other requests see `
          `ckan.cache.private.enabled`` config option.

        default: true
        type: bool
        example: "true"
        description: |
          else ``Cache-Control: no-cache max-age=0`` will be set on all
          requests except for sensitive responses where no-store
          will also be set.

        default: 3600
        type: int
        example: 3600
        description: |
          In Seconds, config used to serve stale content while being background revalidated

        default: 86400
        type: int
        example: 7200
        description: |
          Note: When both ``stale_while_revalidates`` and ``stale_if_error`` are set to 0, ``must-revalidate`` is used instead.

        type: bool
        default: false
        example: "true"
        description: |

      - key: ckan.etags.enabled
        type: bool
        default: true
        example: "true"
        description: |
```

new core helpers:
* cache_level
   get ``g.cache_type``
* limit_cache_for_page
* set_limit_cache_for_page
  set ``g.limit_cache_for_page
* set_cache_level
* etag_append
* set_etag_replace
* set_etag_modified_time

Updates:
* Vary now set on api as well as on pages, surface flask global ``g`` flag for extensions to use

Tests:
Over 1000 lines of test's added covering all new features and then some.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -11,10 +11,13 @@ from __future__ import annotations
 
 import logging
 from collections.abc import MutableMapping, Iterable
+from enum import Enum
 
 from typing import (
     Any, Optional, TYPE_CHECKING,
     TypeVar, cast, overload, Union)
+
+from flask.sessions import SessionMixin
 from typing_extensions import Literal
 
 import flask
@@ -30,7 +33,6 @@ import simplejson as json  # type: ignore # noqa
 import ckan.lib.maintain as maintain
 from ckan.config.declaration import Declaration
 from ckan.types import Request
-
 
 if TYPE_CHECKING:
     MutableMapping = MutableMapping[str, Any]
@@ -216,11 +218,16 @@ class CKANRequest(LocalProxy[Request]):
         return self.args
 
 
-def _get_c():
+def _get_g():
     return flask.g
 
 
-def _get_session():
+# Deprecated, use _get_g()
+def _get_c():  # pyright: ignore[reportUnusedFunction]
+    return _get_g()
+
+
+def _get_session() -> SessionMixin:
     return flask.session
 
 
@@ -330,9 +337,46 @@ config_declaration = local.config_declaration = Declaration()
 # Proxies to already thread-local safe objects
 request = CKANRequest(_get_request)
 # Provide a `c`  alias for `g` for backwards compatibility
-g: Any = LocalProxy(_get_c)
+g = LocalProxy(_get_g)  # flask.ctx._AppCtxGlobals
 c = g
-session: Any = LocalProxy(_get_session)
+session = LocalProxy(_get_session)
 
 truthy = frozenset([u'true', u'yes', u'on', u'y', u't', u'1'])
 falsy = frozenset([u'false', u'no', u'off', u'n', u'f', u'0'])
+
+
+class CacheType(Enum):
+    """ This enum is to simplify cache controls since you can only have one type set
+    If response cache_controls are manually configured, use context OVERRIDDEN to
+    not allow default behaviour to alter your settings"""
+
+    priority: int
+
+    def __new__(cls, value: str, priority: int):
+        obj = object.__new__(cls)
+        obj._value_ = value
+        obj.priority = priority
+        return obj
+
+    PUBLIC = ('public', 1)
+    PRIVATE = ('private', 2)
+    NO_CACHE = ('no-cache', 3)
+    SENSITIVE = ('no-store', 4)
+    OVERRIDDEN = ('overridden', 99)
+
+    @classmethod
+    def can_override(cls, current_in: 'str|CacheType', new_in: 'str|CacheType') -> bool:
+        """Returns True if `new` can override `current`"""
+        if isinstance(current_in, str):
+            current: CacheType = cls(current_in)
+        else:
+            current = current_in
+        if isinstance(new_in, str):
+            new: CacheType = cls(new_in)
+        else:
+            new = new_in
+        return new.priority >= current.priority
+
+
+# This is what can be passed into the query args that is removed from dicts
+CACHE_PARAMETERS = frozenset([u'__cache', u'__no_cache__'])
--- a/ckan/common.py
+++ b/ckan/common.py
@@ -5,7 +5,7 @@
 # but at the same time making it easy to change for example the json lib
 # used.
 #
-# NOTE:  This file is specificaly created for
+# NOTE:  This file is specifically created for
 # from ckan.common import x, y, z to be allowed
 from __future__ import annotations
 
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1075,7 +1075,7 @@ groups:
 
       - key: ckan.site_logo
         default: /base/images/ckan-logo.png
-        example: /images/ckan_logo_fullname_long.png
+        example: /base/images/ckan-logo.png
         description: This sets the logo used in the title bar.
         editable: true
 
@@ -1185,7 +1185,7 @@ groups:
       - key: ckan.site_custom_css
         description: Custom CSS directives to include on all CKAN pages.
         editable: true
-      
+
       - key: ckan.default_collapse_facets
         type: bool
         default: False
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -408,20 +408,88 @@ groups:
           token. This is useful in some scenarios where using the default ``Authorization`` one
           causes problems.
 
-      - key: ckan.cache_expires
-        default: 0
+      - key: ckan.cache.expires
+        default: 3600
+        legacy_key: ckan.cache_expires  # since v.2.12.0
         type: int
         example: 2592000
-        description: This sets ``Cache-Control`` header's max-age value.
+        description: |
+          This sets ``Cache-Control`` header's max-age value on public cache.
+          3600 = 1 hour before revalidate with shared cache
 
-      - key: ckan.cache_enabled
+      - key: ckan.cache.shared.expires
+        default: 7200
+        type: int
+        example: 7200
+        description: |
+          This sets ``Cache-Control`` header's s-maxage value on public cache.
+          7200 = 2 hour before shared cache revalidates
+
+      - key: ckan.cache.private.expires
+        type: int
+        default: 60
+        example: 3600
+        description: |
+          This sets ``Cache-Control`` header's max-age value on private cache.
+
+      - key: ckan.cache.public.enabled
+        legacy_key: ckan.cache_enabled  # since v.2.12.0
+        default: True
         type: bool
         example: "true"
         description: |
           This enables cache control headers on all requests. If the user is
           not logged in and there is no session data a ``Cache-Control:
-          public`` header will be added. For all other requests the
-          ``Cache-control: private`` header will be added.
+          public`` header will be added. For all other requests see `
+          `ckan.cache.private.enabled`` config option.
+
+      - key: ckan.cache.private.enabled
+        default: true
+        type: bool
+        example: "true"
+        description: |
+          When this is set to true and cache type requested is private,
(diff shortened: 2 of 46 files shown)


ckan/common.py
```diff
@@ -330,9 +337,46 @@ def repr_untrusted(danger: Any):
...
+    If response cache_controls are manually configured, use context OVERRIDDEN to
+    not allow default behaviour to alter your settings"""
+
+    priority: int
+
+    def __new__(cls, value: str, priority: int):
+        obj = object.__new__(cls)
+        obj._value_ = value
+        obj.priority = priority
+        return obj
```
wardi, line 359: instead of overriding `__new__` you could use a dataclass, e.g.

```python
@dataclass
class CacheTypeData:
    value: str
    priority: int

class CacheType(CacheTypeData, Enum):
    PUBLIC = ('public', 1)
    ...
```

ckan/config/config_declaration.yaml
```diff
@@ -408,20 +408,88 @@ groups:
           token. This is useful in some scenarios where using the default ``Authorization`` one
           causes problems.
 
-      - key: ckan.cache_expires
-        default: 0
+      - key: ckan.cache.expires
+        default: 3600
```
wardi, line 412: Should we change the default? Dynamic content a site owner has added with any extension would be broken if cached, right?

ckan/config/config_declaration.yaml
```diff
@@ -1075,7 +1144,7 @@ groups:
 
       - key: ckan.site_logo
         default: /base/images/ckan-logo.png
-        example: /images/ckan_logo_fullname_long.png
+        example: /base/images/ckan-logo.png
```
wardi: typically examples are used to show a value the user might use to override the default

ckan/config/config_declaration.yaml
```diff
@@ -1248,7 +1317,7 @@ groups:
         example: public
         description: |
           This config option is used to configure the base folder for static files used
-          by CKAN core. Starting CKAN 2.11 it only accepts: ``public`` as a value.
+          by CKAN core. Starting CKAN 2.11 it only accepts: ``public`` or ``public-midenight-blue`` as a value.
```
wardi:
```suggestion
          by CKAN core. Starting CKAN 2.12 it only accepts: ``public`` or ``public-midnight-blue`` as a value.
```

ckan/lib/base.py
```diff
@@ -105,33 +105,53 @@ def render(template_name: str,
...
-    allow_cache = True
-    # Force cache or not if explicit.
+    cache_force is deprecated, use 'h.set_cache_level' function instead
+    """
     if cache_force is not None:
-        allow_cache = cache_force
+        log.error("cache_force is deprecated use "
+                  "h.set_cache_level(CacheType.PUBLIC, True)")
+
+    # Any rendered template will have Very header added, unless OVERRIDDEN flag is found
```
wardi:
```suggestion
    # Any rendered template will have Vary header added, unless OVERRIDDEN flag is found
```

ckan/lib/base.py
```diff
@@ -105,33 +105,53 @@ def render(template_name: str,
...
+    if h.cache_level() != CacheType.OVERRIDDEN:
+        if h.limit_cache_for_page() is None:
+            g.limit_cache_for_page = True
+
+    if h.cache_level():
+        # log.error("Cache Level found:%r, skipping cache config", h.cache_level)
```
wardi:
```suggestion
```

ckan/lib/base.py
```diff
@@ -105,33 +105,53 @@ def render(template_name: str,
...
-        allow_cache = False
-
-    # Any rendered template will have a login-sensitive header
-    request.environ['__limit_cache_by_cookie__'] = True
-    if not allow_cache:
-        # Prevent any further rendering from being cached.
-        request.environ['__no_cache__'] = True
+        # deprecated, use header Cache-Control: no-cache
+        log.warning("request.args['__no_cache__'] is deprecated, "
+                    "use h.set_cache_level(CacheType.NO_CACHE)")
```
wardi: A site owner can't prevent a user from adding `__no_cache__` to a url, why issue a warning?

ckan/lib/helpers.py
```diff
@@ -2908,3 +2908,78 @@ def csrf_input():
...
+
+
+@core_helper
+def limit_cache_for_page() -> Optional[bool]:
+    return getattr(g, 'limit_cache_for_page', None)
+
+
+@core_helper
+def set_limit_cache_for_page(limit: bool) -> None:
+    g.limit_cache_for_page = limit
```
wardi, line 2938: should this only allow setting to True? If it's called once with True any time during rendering the limit_cache_for_page should be set to True, right?

ckan/lib/helpers.py
```diff
@@ -2908,3 +2908,78 @@ def csrf_input():
...
+            return None
+
+    currentCacheType = cache_level()
+
+    if currentCacheType and cache_type:
+        if CacheType.can_override(currentCacheType, cache_type) or force:
+            g.cache_type = cache_type
+    else:
+        g.cache_type = cache_type
+    # log.debug('cacheType set to %r', cache_type)
```
wardi:
```suggestion
```

ckan/lib/helpers.py
```diff
@@ -2908,3 +2908,78 @@ def csrf_input():
...
+    return g.cache_type
+
+
+@core_helper
+def etag_append(append_value: str) -> None:
+    """ Adds additional etag uniqueness, can be called multiple times"""
+    current_value = getattr(g, 'etag_append', "")
+    g.etag_append = current_value + append_value
```
wardi, line 2969: appending is unsafe for collisions, we can use an HMAC-like approach and store the hash after appending so that
```
h.etag_append('Smith')
h.etag_append('anson')
```
will produce a different etag than
```
h.etag_append('Smit')
h.etag_append('hanson')
```

ckan/views/__init__.py
```diff
@@ -43,31 +52,202 @@ def set_cors_headers_for_response(response: Response) -> Response:
...
+# can't use last modified and content length our Nginx does it
+# This is a simple etag which changes on each request
+def set_etag_for_response(response: Response) -> Response:
+    """Set ETag and return 304 if content is unchanged."""
+
+    # skip streaming content
+    if response.is_streamed:
+        return response
+
+    enable_etags = config.get(u'ckan.etags.enabled', True)
```
wardi: we can rely on config options existing since they're in the config_declaration, and we don't want to repeat default values in the code
```suggestion
    enable_etags = config['ckan.etags.enabled']
```

ckan/views/__init__.py
```diff
@@ -43,31 +52,202 @@ def set_cors_headers_for_response(response: Response) -> Response:
...
+        if 'etag' not in response.headers:
+            etag_replace = getattr(g, 'etag_replace', None)
+            if etag_replace:
+                response.set_etag(etag_replace)
+            else:
+                # s3 etag uses md5 if you want that, load `etag` plugin, this is fast
+                # hash etag which h.set_etag_modified_time(str) can be used to set
+                # correct modified time as well as extended by plugins via
+                # `h.etag_append(str)` for their uniqueness constraints
+                mtime = getattr(g, 'etag_modified_time', time.time())
```
wardi, line 87: We need to set the modified time even if we've already provided unique data in etag_append or etag_replace?

ckan/common.py
duttonw: Thanks for pointing out https://docs.python.org/3/library/dataclasses.html

ckan/config/config_declaration.yaml
duttonw: I'm on the fence on this, we kinda want caching enabled, but ckan's default stance was that it was disabled 'ckan.cache.public.enabled'  / 'ckan.cache_enabled' and the core was never tested with caching enabled.

It may need to be included as a migration note that if you wish to have default behaviour set ``ckan.cache.public.enabled`` to false. Which then ignores this variable.

ckan/config/config_declaration.yaml
```diff
@@ -408,20 +408,88 @@ groups:
...
+        example: "true"
+        description: |
+          When this is set to true and cache type requested is private,
+          then ``Cache-control: private`` header will be added,
+          else ``Cache-Control: no-cache max-age=0`` will be set on all
+          requests except for sensitive responses where no-store
+          will also be set.
+
+      - key: ckan.cache.stale_while_revalidates
+        default: 3600
```
duttonw, line 458: @wardi  we may also want to put in migration notes to enable existing behaviour to set both ``ckan.cache.stale_while_revalidates`` and ``ckan.cache.stale_if_error`` to 0 so ``must-revalidate`` is sent like it does currently. 

Enabling this options allows pages to be background refresh and when loss of internet allows pages to not be 'nuked' if cache has expired and can't be retrieved.

ckan/lib/base.py
duttonw: I'll remove this warning then.

ckan/lib/helpers.py
duttonw: I've left this open so that extensions who don't wish to vary by cookie can force the flag to false at the right time.

ckan/lib/helpers.py
duttonw: The append will add to the current hamc like hash we have. it allows multiple extensions to add their own 'hash' . so if you call 
``h.etag_append('Smith'); h.etag_append('anson')``
then the etag would now have ``{mtime}-{size}-{check}{etag_append} `` and etag_append would be ``Smithanson``

this does not change the etag completely.  That is done by function ``set_etag_replace`` which if set append won't be used.

ckan/views/__init__.py
duttonw: etag_append is for extensions to add their own uniquness hash.
etag_replace just blats everything with whatever was set (nuclear approach) only for extensions to use (with caution)

etag_modified_time can be used in the future when extensions are supporting the etag_append feature to allow us to just use the date in the 'db' for when the record was last modified on certain pages/functions.

duttonw (commented):
Just come back from 3 weeks leave. I'll get onto the code changes requested in the PR comments shortly.  Attached is responses to some of the comments also.

ckan/common.py
duttonw: I've looked into this, 

* since Its an enum, uses 'new' to set itself up
* ``@dataclass`` uses init so they won't mix well. 

I've looked into doing the @dataclass extra object but it's cumbersome and adds far more indirection to the system.

PR #8959, opened 2025-05-22
Skip changelog check if 'no changelog' label is given

Zharktas:
### Proposed fixes:
Towncrier checks on pull requests are good on enforcing that features and bug fixes are added to the changelog, but it also adds a lot of bloat in the changelog when internal changes which are not related to the use of CKAN itself gets added in there just to get all checks green.

This modification on the workflow would skip the towncrier check if the pull request has `no changelog` label added into it so it would be requested by the pull request creator or reviewer themselves.

### Features:

- [ ] includes API changes


.github/workflows/towncrier.yml
```diff
@@ -7,6 +7,7 @@ permissions:
 jobs:
   towncrier:
     runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no changelog') }}
```
amercader, line 10:
```suggestion
    if: ${{ !contains(github.event.pull_request.labels.*.name, 'No changelog') }}
```

PR #8960, opened 2025-05-22
Hide and show simply

avdata99:
Related to #7900

### Proposed fixes:

I'm having troubles to hide the loading dots in the datatables view.
This fixed it.
The proposal is to move to the simple way to hide/show in jQuery

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datatablesview/assets/datatablesview.js
+++ b/ckanext/datatablesview/assets/datatablesview.js
@@ -484,6 +484,7 @@ this.ckan.module('datatables_view', function (jQuery) {
         // show loading indicator when first painting data into the table.
         // useful with very large resources that take long to load
         $('body.dt-view').css('visibility', 'visible');
+        $('#dtprv_processing').addClass('pre-init');
         $('#dtprv_processing').show();
       });
       datatable = $('#dtprv').DataTable({
@@ -606,6 +607,7 @@ this.ckan.module('datatables_view', function (jQuery) {
           }
 
           // hide the pre-loading indicator background so the table is interactive when loading
+          $('#dtprv_processing').removeClass('pre-init');
           $('#dtprv_processing').hide();
 
           // restore selected rows from state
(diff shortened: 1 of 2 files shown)


avdata99:
You created #7900 @JVickery-TBS 
What do you think about this?

ckanext/datatablesview/assets/datatablesview.js
```diff
@@ -606,7 +606,7 @@ this.ckan.module('datatables_view', function (jQuery) {
           }
 
           // hide the pre-loading indicator background so the table is interactive when loading
-          $('#dtprv_processing').removeClass('pre-init');
+          $('#dtprv_processing').hide();
```
JVickery-TBS, line 609:
```
$('#dtprv_processing').removeClass('pre-init').hide();
```

JVickery-TBS (changes requested):

JVickery-TBS:
@avdata99 the CSS class I wrote is just to have a forced white foreground which covers the table when it is initializing, because the table does a bunch of various painting which doesn't look great. But this is not entirely necessary.

It is the datatables shipped vendor JS script with `processing: true` that is SUPPOSED to handle the show/hiding of the loader,

The dt processing not being removed was an old issue with some jQuery versions or something like that. To support various versions of jQuery, the show/hide here should work fine. But if you want to keep the white splash screen for the initialization to make it look better, just leave in the add/remove class.

avdata99:
I just re-add the `pre-init` class add/remove
Thanks @JVickery-TBS !

JVickery-TBS (approved):

mutantsan:
Honestly, I'm having a hard time understanding what this fixes.

avdata99:
Cleaning, too old

JVickery-TBS:
agreed, I have recoded and upgraded distros for datatables here: https://github.com/ckan/ckan/pull/9226

PR #8964, opened 2025-05-23
Replace flake8 with ruff

### LINKED ISSUE #7325: Replace flake8 with ruff
Many bigger projects are replacing flake8 with [ruff](https://github.com/charliermarsh/ruff), like FastAPI, Jupyter and pandas. This change would have to be made in CI, CI-templates, various extension CIs and so on.

If there is an interest in this, I could do some testing on linting if there are any differences.

Zharktas:
Fixes #7325 

### Proposed fixes:

Adds ruff workflow and moves various configuration from [.flake8](https://github.com/ckan/ckan/blob/master/.flake8) file to `pyproject.toml`

Does not include checking for warnings in linting like the `flake8` workflow does, but those warnings were largely ignored. Instead adds files with linting errors to `pyproject.toml', from which they will hopefully be removed in the future once linting has been fixed. 

Updates [tool docs](https://docs.ckan.org/en/2.11/contributing/python.html#some-helpful-tools-for-python-code-quality) which had really old links to various places.

Does not yet remove flake8 workflow or configuration, I'll remove them once these has been approved.

### Features:

- [ ] includes API changes


Zharktas:
Noticed that there was black configuration in `pyproject.toml`, removed it as ruff will replace black as well and the configuration seemed to be different to flake8, we should only have one configuration to linting.

pyproject.toml
```diff
@@ -36,10 +36,6 @@ directory = 'removal'
 name = 'Removals and deprecations'
 showcontent = true
 
-[tool.black]
-line-length = 79
-preview = true
```
wardi, line 41: unrelated change?

Zharktas: See above comment https://github.com/ckan/ckan/pull/8964#issuecomment-2915604613 :)

PR #8967, opened 2025-05-27
Upgrade select2 to 4.0.13

### LINKED ISSUE #5697: Upgrade to Select2 4.x

wardi:
Fixes #5697

### Proposed fixes:
- add select2 to gulpfile for easier future updates
- replace select2 in both public and public-midnight-blue
- use select2.full.js instead of select2.js for backwards compatibility with 3.x

### Features:

- [ ] includes API changes


wardi:
@smotornyuk 4.1.0-rc.0 also appears to work, would you like me to switch to that version instead?

PR #8977, opened 2025-05-30
Fix critical error if storage path is missing by adding a new config option

### LINKED ISSUE #8966: ckan.storage_path should not log critical errors when configuration is missing
## CKAN version

2.10, 2.11, master

## Describe the bug

Docs say that maintainer of ckan should add configuration for `ckan.storage_path` if they want to enable file uploads. https://docs.ckan.org/en/2.11/maintaining/filestore.html#setup-file-uploads. 

If the maintainer does not want to enable file uploads, CKAN checks if the uploads are enabled via helper:

https://github.com/ckan/ckan/blob/ce12bfcc19a203a4c48e8b454d420dccb461d120/ckan/lib/helpers.py#L2485-L2489

Which calls `get_storage_path`:

https://github.com/ckan/ckan/blob/ce12bfcc19a203a4c48e8b454d420dccb461d120/ckan/lib/uploader.py#L82-L89

Which logs a message on critical level, which in log level terms is worse than error.

### Steps to reproduce

Run CKAN without defining `ckan.storage_path`.

### Expected behavior

Sometimes it desired behavior not to allow file uploads, it should be at least lower level log message and clearer message that this is optional.

Zharktas:
Fixes #8966 

### Proposed fixes:

Adds a new config option `ckan.uploads_enabled` which only affects `uploads_enabled` helper.  If uploader instances call `get_storage_path` they still get the error.

Does not affect translations or webassets directories as they use the config directly as seen on below:

https://github.com/ckan/ckan/blob/f2cc28e8b71608d70361cf2488ea536447e102a5/ckan/lib/i18n.py#L79

https://github.com/ckan/ckan/blob/f2cc28e8b71608d70361cf2488ea536447e102a5/ckan/lib/webassets_tools.py#L209-L212  

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8977.bugfix
+++ b/changes/8977.bugfix
@@ -1,0 +1,4 @@
+Added a new config option :ref:`ckan.uploads_enabled` to fix critical error logs about missing storage_path.
+
+You must set this to true, if you want to enable file uploads in the UI.
+
--- a/changes/8977.bugfix
+++ b/changes/8977.bugfix
@@ -1,4 +1,4 @@
 Added a new config option :ref:`ckan.uploads_enabled` to fix critical error logs about missing storage_path.
 
-You must set this to true, if you want to enable file uploads in the UI.
+If not enabled, you either must have IUploader plugin configured or have defined storage_path to have uploads in the UI.
 
--- a/ckan/config/config_declaration.yaml
+++ b/ckan/config/config_declaration.yaml
@@ -1185,7 +1185,7 @@ groups:
       - key: ckan.site_custom_css
         description: Custom CSS directives to include on all CKAN pages.
         editable: true
-      
+
       - key: ckan.default_collapse_facets
         type: bool
         default: False
@@ -1304,10 +1304,22 @@ groups:
 
   - annotation: Storage Settings
     options:
+      - key: ckan.uploads_enabled
+        type: bool
+        default: false
+        example: false
+        description: |
+          Defines file uploads are enabled or not.
+
+          If enabled also set :ref:`ckan.storage_path` if you want to use internal storage.
+
       - key: ckan.storage_path
         placeholder: /var/lib/ckan/default
         example: /var/lib/ckan/default
-        description: This defines the location of where CKAN will store all uploaded data.
+        description: |
+          This defines the location of where CKAN will store all uploaded data.
+
+          If set to some directory, also set :ref:`ckan.uploads_enabled` to true to enable uploads in the UI.
 
       - key: ckan.max_resource_size
         type: int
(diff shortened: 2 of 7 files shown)


wardi:
If we add an `ckan.uploads_enabled` configuration option I'd like it to be optional for backwards compatibility.

`h.uploads_enabled()` could then:
- return value of `ckan.uploads_enabled` if it's set
- otherwise return True if any IUploader plugin is loaded or if the storage path is set (without logging an error)

When @smotornyuk 's replacement interface for IUploader is introduced we can probably get rid of `h.uploads_enabled()`  and switch to `h.check_access()` calls on the new upload action.

smotornyuk:
Yes, I experimented with different combinations of default values for configuration and see that @Zharktas 's solution is the best we can get in the current state. 

Even after implementing the upload API for different upload types, this new option may be maintained as a global switch for uploads. For example, if you'd like to disable uploads for a day due to maintenance, but keep all existing files available, you can simply turn this option off. By that time, we'll change the declaration of the option to `bool` and set the default value to `true`, eliminating this `none` value that confuses me in boolean flags.

amercader:
@Zharktas @smotornyuk I'm reviewing this while updating the CHANGELOG for CKAN 2.11 and 2.10, two questions:

> Added a new config option :ref:`ckan.uploads_enabled` to fix critical error logs about missing storage_path.

> If not enabled, you either must have IUploader plugin configured or have defined storage_path to have uploads in the UI.

Shouldn't this be "if enabled" on the second sentence?

Second question is, do existing sites that define `ckan.storage_path` need to additionally add `ckan.uploads_enabled` to their configs to have functioning uploads? I assume the answer is no otherwise that would be a major breaking change, so in that case [this](https://github.com/ckan/ckan/pull/8977/files#diff-5752be3d55839c292941807dff4cb1943222e276a2d2a64888e23f5b49104628R1321) config documentation is misleading as `ckan.uploads_enabled` is not required right?

> If set to some directory, also set :ref:`ckan.uploads_enabled` to true to enable uploads in the UI.

If I understood the change correctly what do you think about this changelog entry in the Minor Changes section (not the bugfixes):

> Added a new config option :ref:`ckan.uploads_enabled`, which you can set to avoid a critical error logs about missing storage_path during startup.
>
> If set to True, you either must have set :ref:`ckan.storage_path` or have an ``IUploader`` plugin configured to have uploads  enabled in the UI.

wardi:
Do we have a problem with the JS translations if a storage path isn't configured? The current code will use a temporary directory if there's no storage path but it will also no longer auto-regenerate the translations on every cli invocation so if the temp directory was removed we'll be serving ckan with no translation files, right?

We could document that if you don't have a storage path configured you'll need to generate the js translations each time before starting the server.

There's a similar problem with webassets, but webassets requires another configuration option `ckan.webassets.path` if `ckan.storage_path` is not defined.

PR #8984, opened 2025-06-04
New config value for query timeout

avdata99:
### Proposed fixes:

This is a proposal to allow users to define the default query timeouts for the datastore

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/8984.misc
+++ b/changes/8984.misc
@@ -1,0 +1,2 @@
+Add a new config value for Datastore query
+timeout ckan.datastore.default_query_timeout
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -59,8 +59,6 @@ _type_names: Set[str] = set()
 _engines: Dict[str, Engine] = {}
 WhereClauses: TypeAlias = "list[tuple[str, dict[str, Any]] | tuple[str]]"
 
-_TIMEOUT = 60000  # milliseconds
-
 # See http://www.postgresql.org/docs/9.2/static/errcodes-appendix.html
 _PG_ERR_CODE = {
     'unique_violation': '23505',
@@ -571,6 +569,9 @@ def _build_query_and_rank_statements(
         rank_field, query_alias)
     return statement, rank_statement
 
+
+def _get_timeout() -> int:
+    return config.get('ckan.datastore.default_query_timeout')
 
 def _fts_lang(lang: Optional[str] = None) -> str:
     return lang or config.get('ckan.datastore.default_fts_lang')
@@ -1729,7 +1730,8 @@ def upsert(context: Context, data_dict: dict[str, Any]):
     backend = DatastorePostgresqlBackend.get_active_backend()
     engine = backend._get_write_engine()  # type: ignore
     context['connection'] = engine.connect()
-    timeout = context.get('query_timeout', _TIMEOUT)
+    default_timeout = _get_timeout()
+    timeout = context.get('query_timeout', default_timeout)
 
     trans: Any = context['connection'].begin()
     try:
@@ -1778,7 +1780,8 @@ def search(context: Context, data_dict: dict[str, Any]):
     engine = backend._get_read_engine()  # type: ignore
     _cache_types(engine)
     context['connection'] = engine.connect()
-    timeout = context.get('query_timeout', _TIMEOUT)
+    default_timeout = _get_timeout()
+    timeout = context.get('query_timeout', default_timeout)
 
     try:
         context['connection'].execute(sa.text(
@@ -1808,7 +1811,8 @@ def search_sql(context: Context, data_dict: dict[str, Any]):
     _cache_types(engine)
 
     context['connection'] = engine.connect()
-    timeout = context.get('query_timeout', _TIMEOUT)
+    default_timeout = _get_timeout()
+    timeout = context.get('query_timeout', default_timeout)
 
     sql = data_dict['sql']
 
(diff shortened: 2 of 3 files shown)


ckanext/datastore/backend/postgres.py
```diff
@@ -572,6 +570,9 @@ def _build_query_and_rank_statements(
     return statement, rank_statement
 
 
+def _get_timeout() -> int:
+    return config.get('ckan.datastore.default_query_timeout', 60000)
```
wardi: The default is set in config_declaration so we don't need to repeat it here.

ckanext/datastore/config_declaration.yaml
```diff
@@ -92,6 +92,16 @@ groups:
       them. It can be overwritten by the user by passing the "lang" parameter to
       "datastore_search" and "datastore_create".
 
+  - default: 60_000
```
wardi: let's set the default to a much smaller value (like 1s) because on busy sites one request being able to fully load the db server for 60s is really bad.

ckanext/datastore/config_declaration.yaml
```diff
@@ -92,6 +92,16 @@ groups:
...
+    key: ckan.datastore.default_query_timeout
+    type: int
```
wardi, line 97: I can see wanting to set this to a fraction of a second

ckanext/datastore/config_declaration.yaml
```diff
@@ -92,6 +92,16 @@ groups:
       them. It can be overwritten by the user by passing the "lang" parameter to
       "datastore_search" and "datastore_create".
 
+  - default: 60_000
+    key: ckan.datastore.default_query_timeout
```
wardi: I might even add `_ms` to this name because without reading the docs users won't know the units.

Alternatively make this a float value in seconds because all the other config values are in seconds.

avdata99: Done here https://github.com/avdata99/ckan/commit/679ee36a52853809d5d7311b57424775869244a3

ckanext/datastore/config_declaration.yaml
avdata99: Done here https://github.com/avdata99/ckan/commit/2d235309d02f363c6932f5493594ff17aafffd01

ckanext/datastore/backend/postgres.py
avdata99: Done here https://github.com/avdata99/ckan/commit/3d99079f7a8c5cab4efcb98a05d96ee6bf95f1b4

ckanext/datastore/cli.py
```diff
@@ -19,10 +19,12 @@
     get_read_engine,
     get_write_engine,
     _get_raw_field_info,
-    _TIMEOUT,
+    _get_timeout,
 )
 from ckanext.datastore.blueprint import DUMP_FORMATS, dump_to
 
+
+_TIMEOUT: int = _get_timeout()
```
wardi: the timeout in the cli for the `fts-index` command can be much longer, maybe leave this at 60s. No need for a separate configuration option because it can be changed with a cli option.

ckanext/datastore/config_declaration.yaml
```diff
@@ -92,6 +92,16 @@ groups:
       them. It can be overwritten by the user by passing the "lang" parameter to
       "datastore_search" and "datastore_create".
 
+  - default: 1_000
+    key: ckan.datastore.default_query_timeout_ms
```
wardi: Better to call this `query_timeout`? It's a value that can't be changed by users so it's not a default.

ckanext/datastore/config_declaration.yaml
```diff
@@ -92,6 +92,16 @@ groups:
...
+    type: int
+    example: 180_000
+    description: >
+      The default query timeout (in milliseconds) for the datastore queries.
+      This is used when executing queries against the datastore.
+      "search"/"search_sql", "upsert", and "create" actions will use this
```
wardi, line 102: It's worth having a longer timeout for `upsert` and `create`. Would you add config options for these as well? These can default to 60s.

I might have been too aggressive with my 1s default suggestion for the search queries. To prevent too many reports of failures for large sites let's set that to 5s.

changes/8984.misc
```diff
@@ -0,0 +1,2 @@
+Add a new config value for Datastore query
+timeout ckan.datastore.default_query_timeout_ms
```
wardi: We need to tell users what the new timeout configuration options and default values are so they can change them if needed.

wardi (changes requested):
thank you @avdata99 just a couple more small changes and this is good to go

ckanext/datastore/cli.py
avdata99: Done here @wardi: https://github.com/avdata99/ckan/commit/b9b83f14363ee0d0ad682dd0fcdecc4ed57f0583

ckanext/datastore/config_declaration.yaml
avdata99: Done here @wardi: https://github.com/avdata99/ckan/commit/185e32275f3de145f20f08fb8cdf46edb2410941

ckanext/datastore/config_declaration.yaml
avdata99: Fixed here @wardi: https://github.com/avdata99/ckan/commit/9dee07417ceee7b444d543ff540dd9175b0f2dca

changes/8984.misc
avdata99: Done in the latest commit

PR #8985, opened 2025-06-04
new option: ckan jobs worker --with-scheduler

### LINKED ISSUE #8980: datastore-first write operations should update resource last_modified value
## CKAN version
all

## Describe the bug
For datastore-first resources modifying the datastore with `datastore_create`, `datastore_upsert`, `datastore_records_delete` or `datastore_delete` does not update the corresponding resource `last_modified` value.

Users must separately call e.g. `resource_patch` to update the `last_modified` value.

This is inconsistent with uploading a file to a resource which does update the `last_modified` value automatically.

### Steps to reproduce
Create a resource with `datastore_create` or use table designer.

### Expected behavior
Datastore changes made with the API or table designer UI should cause the resource's `last_modified` value to be updated.

### Additional details
1. Don't update `last_modified` on datastore changes for resources that are not datastore-first (e.g. tables created with datapusher or xloader) because the datastore is only a copy of the data.

2. If a datastore-first write operation does not change the table or its data the `last_modified` value should also not change, similar to how `metadata_modified` at the package level does not change unless there is an actual change made to the metadata ( since #8421 ). This can be implemented building on #8684

3. Updating `last_modified` on every datastore-first write operation will cause performance issue when loading large tables into datasets with many resources.

   ~We need a new parameter on these datastore endpoints:~
    * ~`update_last_modified = "exact"` set the last_modified value to the exact current time~
    * ~`update_last_modified = "minute"` set the last_modified value to the nearest minute (DEFAULT)~
    * ~`update_last_modified = "unchanged"` do not set the last_modified value~

   ~By default updating the value to the nearest minute repeated calls to `datastore_upsert` won't slow down significantly since the metadata will only be updated once or twice.~

   ~Tools like datapusher and xloader can be modified to pass `update_last_modified = "unchanged"` for all their initial datastore API calls, then `update_last_modified = "exact"` on the last. This is similar to how they should already be using the `calculate_record_count = True` parameter on completion to update the table statistics on the last `datastore_upsert` call.~

   Update: the fix in #8985 is much better than this suggestion and would also be a great approach for the high water mark workflow described by @EricSoroos

wardi:
Fixes #8980 

### Proposed fixes:

- Add new option: `ckan jobs worker --with-scheduler` to schedule jobs as well as run jobs in the queue
- `datastore_create`, `datastore_upsert` and `datastore_delete` now schedule a job to patch the corresponding resource's `last_modified` value for datastore-first resources. A scheduled job is used to reduce duplicated metadata updates that would slow down these operations.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/cli/jobs.py
+++ b/ckan/cli/jobs.py
@@ -18,8 +18,10 @@ def jobs():
 @click.option(u"--max-idle-time", default=None, type=click.INT,
               help=u"Max seconds for worker to be idle. "
               "Defaults to None (never stops idling).")
+@click.option("--with-scheduler", is_flag=True,
+              help="Schedule jobs from this worker.")
 @click.argument(u"queues", nargs=-1)
-def worker(burst: bool, max_idle_time: int, queues: list[str]):
+def worker(burst: bool, max_idle_time: int, queues: list[str], with_scheduler: bool):
     """Start a worker that fetches jobs from queues and executes them. If
     no queue names are given then the worker listens to the default
     queue, this is equivalent to
@@ -42,7 +44,9 @@ def worker(burst: bool, max_idle_time: int, queues: list[str]):
     If the `--max-idle-time` option is given then the worker will exit
     after it has been idle for the number of seconds specified.
     """
-    bg_jobs.Worker(queues).work(burst=burst, max_idle_time=max_idle_time)
+    bg_jobs.Worker(queues).work(
+        burst=burst, max_idle_time=max_idle_time, with_scheduler=with_scheduler,
+    )
 
 
 @jobs.command(name=u"list", short_help=u"List jobs.")
--- a/changes/8980.feature
+++ b/changes/8980.feature
@@ -1,0 +1,8 @@
+New option `ckan jobs worker --with-scheduler` will schedule jobs
+as well as run jobs in the queue.
+
+`datastore_create`, `datastore_upsert` and `datastore_delete` now
+schedule a job to patch the corresponding resource's `last_modified`
+value for datastore-first resources. A scheduled job is used to
+reduce duplicated metadata updates that would slow down these
+operations.
--- a/ckanext/datastore/logic/action.py
+++ b/ckanext/datastore/logic/action.py
@@ -5,6 +5,7 @@ from ckan.types import Context
 import logging
 from typing import Any
 from contextlib import contextmanager
+from datetime import datetime, timezone, timedelta
 
 import sqlalchemy
 import sqlalchemy.exc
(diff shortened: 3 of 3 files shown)


ckan/cli/jobs.py
```diff
@@ -18,8 +18,10 @@ def jobs():
 @click.option(u"--max-idle-time", default=None, type=click.INT,
               help=u"Max seconds for worker to be idle. "
               "Defaults to None (never stops idling).")
+@click.option("--with-scheduler", is_flag=True,
+              help="Schedule jobs from this worker.")
 @click.argument(u"queues", nargs=-1)
-def worker(burst: bool, max_idle_time: int, queues: list[str]):
+def worker(burst: bool, max_idle_time: int, queues: list[str], with_scheduler: bool):
```
EricSoroos: So, this just passes through a scheduler option to the backend, allowing for something similar (but differently implemented) from ckanext-harvests method of scheduling for some point in the future. 

I also note that the docstring for this method mentions paster -- not sure if that's still relevant to the post python2 world.

wardi: Thanks, fixed docstring.

Yes, ckanext-harvest predates rq in ckan by a fair bit. This PR exposes the same option that is available for running vanilla rq workers with the scheduler https://python-rq.org/docs/scheduling/

EricSoroos: If this isn't the default, will these datastore jobs just quietly stack up and not get run?

wardi: From the docs https://python-rq.org/docs/scheduling/#running-the-scheduler :

> Only a single scheduler can run for a specific queue at any one time. If you run multiple workers with scheduler enabled, only one scheduler will be actively working for a given queue.
>
> Active schedulers are responsible for enqueueing scheduled jobs. Active schedulers will check for scheduled jobs once every second.
>
> Idle schedulers will periodically (every 15 minutes) check whether the queues they’re responsible for have active schedulers. If they don’t, one of the idle schedulers will start working. This way, if a worker with active scheduler dies, the scheduling work will be picked up by other workers with the scheduling component enabled.

We could change the default and go in a different direction than the vanilla `rq worker` cli and add a `--no-scheduler` option instead for all the secondary workers on a queue so they don't wake every 15 minutes for no reason.

amercader:
The code looks great, some things that I think are missing:

* Docs:
  * Mention in the scheduler option somewhere in the [Background jobs](https://docs.ckan.org/en/latest/maintaining/background-tasks.html) page
  * Mention the `--with-scheduler` option in the [jobs CLI docs](https://docs.ckan.org/en/latest/maintaining/cli.html#run-a-background-job-worker)
  * Mention the handling of `last_modified` in the `datastore_*` actions docs (and what happens if you are not running a worker)
  
* Tests: some tests to show that at least the job is called when calling the datastore actions

wardi:
@amercader lots of changes since your last review but this should address everything you raised. Workers now default to running the scheduler (there's a `--no-scheduler` option now instead), tests are improved, docs are improved, unnecessary scheduled jobs are skipped when resource_patch is already being called to update the datastore_active flag.

PR #8992, opened 2025-06-10
Replace usage of pkg_resources.iter_entry_points, update exception message

amercader:
`pkg_resources` is deprecated and to be removed soon, replace it with `importlib.metadata` instead. :

```
UserWarning: pkg_resources is deprecated as an API. See
https://setuptools.pypa.io/en/latest/pkg_resources.html. The
pkg_resources package is slated for removal as early as 2025-11-30.
Refrain from using this package or pin to Setuptools<81.
```

While working on that code, I never liked this exception message when trying to load an non-existent plugin:

```
  File "/home/adria/dev/projects/ckan-py310/ckan/ckan/plugins/core.py", line 250, in _get_service
    raise PluginNotFoundException(plugin_name)
ckan.plugins.base.PluginNotFoundException: Interface dcat does not exist
```

So I updated the message:

```
  File "/home/adria/dev/projects/ckan-py310/ckan/ckan/plugins/core.py", line 250, in _get_service
    raise PluginNotFoundException(plugin_name)
ckan.plugins.base.PluginNotFoundException: Plugin 'dcat' not found.

Did you install the 'ckanext-*' Python package that contains the plugin?
```


amercader:
The changes are not compatible with Python 3.9. Waiting on https://github.com/ckan/ckan/issues/8994 to update the workflows

amercader:
Typing errors will be fixed when https://github.com/ckan/ckan/pull/8995 is merged

ckan/plugins/core.py
```diff
@@ -240,13 +240,10 @@ def _get_service(plugin_name: str) -> Plugin:
...
-        iterator = iter_entry_points(
-            group=group,
-            name=plugin_name
-        )
-        plugin = next(iterator, None)
-        if plugin:
-            return plugin.load()(name=plugin_name)
+        eps = entry_points(group=group, name=plugin_name)   # type:ignore
+        if len(eps.names):
+            plugin_ep = eps.pop()
```
wardi: This retrieves the last item but the original code retrieves the first. Is that intended?

ckan/plugins/core.py
```diff
@@ -240,10 +240,11 @@ def _get_service(plugin_name: str) -> Plugin:
...
     """
     for group in GROUPS:
-        iterator = iter_entry_points(group=group,name=plugin_name)    # type: ignore
-        plugin = next(iterator, None)
-        if plugin:
-            return plugin.load()(name=plugin_name)
+        eps = entry_points(group=group, name=plugin_name)   # type:ignore
+        if len(eps.names):
+            plugin_ep = eps.pop(0)
+            return plugin_ep.load()(name=plugin_name)
```
wardi: looking at the EntryPoints class definition treating it as a list seems to be deprecated. Something like this might be more future-proof:

```suggestion
        if ep:
            return ep[plugin_name].load()(name=plugin_name)
```

amercader:
@wardi Not sure what codecov is going on about, but this is ready to go

amercader:
2.11 version is here: https://github.com/ckan/ckan/pull/9013. On 2.10 the plugins code was too different to apply the patches.

PR #9008, opened 2025-06-27
[CLI] Export migration command

smotornyuk:
Register 2 new `ckan db` subcommands:

### `history  [from:to] [-p plugin]`

Show all the migrations. Core by default, `-p plugin` switches to a specific plugin. Optionally, the range of migrations can be restricted using slice-like syntax with revision hashes instead of numeric offset.

```sh
$ ckan db history -p activity
<base> -> 71713a055d5c, Add permission labels in activity table
71713a055d5c -> fab3bfdcf830 (head), Populate permission_labels column
``` 

### `export-migration [from:to] [-p plugin] [--downgrade]`

Instead of applying migrations, print SQL that can be executed to achieve the same result. Due to alembic restrictions, some parts cannot be exported in this way, which is mentioned in the documentation. For example, if data is pulled into Python, processed there, and then ingested back into the DB, Alembic cannot export it.

```sh
$ ckan db export-migration -p activity
BEGIN;

CREATE TABLE activity_alembic_version (
    version_num VARCHAR(32) NOT NULL,
    CONSTRAINT activity_alembic_version_pkc PRIMARY KEY (version_num)
);

-- Running upgrade  -> 71713a055d5c

ALTER TABLE activity ADD COLUMN permission_labels TEXT[];

INSERT INTO activity_alembic_version (version_num) VALUES ('71713a055d5c') RETURNING activity_alembic_version.version_num;

-- Running upgrade 71713a055d5c -> fab3bfdcf830

UPDATE activity
        SET permission_labels = '{"public"}'
        WHERE (
            activity_type LIKE '%% package' AND
            permission_labels IS NULL
        );

UPDATE activity_alembic_version SET version_num='fab3bfdcf830' WHERE activity_alembic_version.version_num = '71713a055d5c';

COMMIT;
```


ckan/cli/db.py
```diff
@@ -265,3 +272,72 @@ def _repo_for_plugin(plugin: str):
...
+        repo.setup_migration_version_control()
+        current = (repo.current_version() or "base").removesuffix(" (head)")
+        versions = versions.replace("current", current)
+
+        try:
+            func(repo.alembic_config, versions, True)
+        except CommandError as err:
+            error_shout(err)
+        except AssertionError:
+            error_shout(f"Cannot compute revision {versions}")
```
wardi, line 325: when does this happen? Why is it an assertion error?

smotornyuk: Alembic widely uses assertions instead of raising specific exceptions, like [here](https://github.com/sqlalchemy/alembic/blob/cd4f404358f101b2b930013c609c074baca61468/alembic/script/base.py#L919). Usually it just means that you are using it's tools in a wrong way

PR #9012, opened 2025-07-02
Attributtes in assets

amercader:
Attributes for [`link`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link#attributes) and [`script`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script#attributes) are really useful to control a wide range of features. For instance attributes like `defer` or `async` can have important performance implications, `type="module` can enable the use of [JS modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules), `crossorigin` might be useful when dealing with CORS, etc

Sadly our webassets rendering code didn't allow any of these but the good news is that with a relatively simple change we can allow arbitrary attributes on the generated HTML tags.

For instance:

```
my_asset:
  output: base/%(version)s_my_asset.js
  extra:
    attrs:
      async:
  contents:
    - ...
```

Will generate:

```
<script src="{url}" async type="text/javascript"></script>
```

Another example:
```
my_asset:
  output: base/%(version)s_my_asset.js
  extra:
    attrs:
      type: module
  contents:
    - ...
```

Will generate:

```
<script src="{url}" type="module"></script>
```

Scripts loaded using `preload` will be correctly renders as `<link>` tags.

```
my_asset:
  output: base/%(version)s_my_asset.js
  extra:
    attrs:
      rel: "preload"
      as: "script"
  contents:
    - ...
```

Will generate:

```
<link href="{url}" rel="preload" as="script"/>
```
The same functionality is available for CSS styles (`<link rel="stylesheet">` tags). 
Includes docs and tests

### DIFF
--- a/ckan/lib/webassets_tools.py
+++ b/ckan/lib/webassets_tools.py
@@ -163,7 +163,7 @@ def _to_tag(asset_info: Tuple[str, dict[str, str]], type_: AssetType) -> str:
 
     if type_ == "style" and "rel" not in attrs:
         attrs["rel"] = "stylesheet"
-    elif type_ == "script" and "type" not in attrs:
+    elif type_ == "script" and "type" not in attrs and not is_preload:
         attrs["type"] = "text/javascript"
 
     attr_items = []
--- a/ckan/lib/webassets_tools.py
+++ b/ckan/lib/webassets_tools.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 import os
-from typing import Any
+from typing import Any, Tuple
 from typing_extensions import Literal, TypedDict, assert_never
 
 from markupsafe import Markup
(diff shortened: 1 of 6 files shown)


ckan/lib/webassets_tools.py
```diff
@@ -146,19 +148,37 @@ def include_asset(name: str) -> None:
...
+        attrs["rel"] = "stylesheet"
+    elif type_ == "script" and "type" not in attrs and not is_preload:
+        attrs["type"] = "text/javascript"
+
+    attr_items = []
+    for key, value in attrs.items():
+        if not value:
+            attr_items.append(key)
+        else:
+            attr_items.append(f'{key}="{value}"')
```
wardi, line 174: The key/value could include `>`, `"`, etc. so we should escape them to prevent generating invalid html.

ckan/lib/webassets_tools.py
```diff
@@ -127,6 +127,7 @@ def include_asset(name: str) -> None:
 
     bundle: Any = env[name]
     deps: list[str] = bundle.extra.get("preload", [])
+    attrs: dict[str, str] = bundle.extra.get("attrs", {})
```
wardi, line 130: yaml could include almost any type of object. Force str key and values here or raise an exception if the types are wrong?

tino097:
hey @amercader can you please address Ian's comment and we can merge this

PR #9013, opened 2025-07-03
Replace usage of pkg_resources.iter_entry_points [2.11 version]

amercader:
2.11 version of #8992 
Slightly more complex as the entry_point functions have different signatures on Python 3.9

### DIFF
--- a/ckan/plugins/core.py
+++ b/ckan/plugins/core.py
@@ -7,7 +7,7 @@ import logging
 from contextlib import contextmanager
 from typing import Generic, Iterator, TypeVar
 from typing_extensions import TypeGuard
-from pkg_resources import iter_entry_points
+from importlib.metadata import entry_points
 
 
 from ckan.common import config, aslist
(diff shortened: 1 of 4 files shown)


ckan/plugins/core.py
```diff
@@ -242,9 +242,16 @@ def find_system_plugins() -> list[str]:
...
 
     eps = []
-    for ep in iter_entry_points(group=SYSTEM_PLUGINS_ENTRY_POINT_GROUP):
-        ep.load()
-        eps.append(ep.name)
+    try:
+        for ep in entry_points(group=SYSTEM_PLUGINS_ENTRY_POINT_GROUP):     # type: ignore
+            ep.load()
+            eps.append(ep.name)
+    except TypeError:
```
wardi, line 246: does the TypeError happen in the entry_points call? If so it would be better to not accidentally mask a TypeError from the load call. Same for the code below.

wardi (changes requested):
Just the possible overly broad TypeError to fix and this is good to go.

amercader:
@wardi made the changes you requested here

amercader:
@wardi ping. Would be good to merge this to get rid of the ugly warning on the next 2.11 patch release

PR #9016, opened 2025-07-05
feat: add toast.js module

mutantsan:
This PR introduces a reusable `ckan.toast` module that displays Bootstrap 5 toast notifications across CKAN. It provides a flexible way to show temporary status messages to users with support for different positions, styles, icons, and progress indicators.

- Displays Bootstrap 5 toast messages with customizable options
- Supports message types (success, danger, info, etc.) with bootstrap styles
- Allows optional title, subtitle, icon and progress bar
- Configurable 9 positions (e.g., top-right, bottom-center)
- Stacking control to prevent overlapping toasts

image

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/public-midnight-blue/base/css/main-rtl.css
+++ b/ckan/public-midnight-blue/base/css/main-rtl.css
@@ -35,7 +35,7 @@
   --bs-success: #3A833A;
   --bs-info: #0dcaf0;
   --bs-warning: #fd7e14;
-  --bs-danger: #d43f3a;
+  --bs-danger: rgb(212.4719626168, 62.5046728972, 58.0280373832);
   --bs-light: #f2f4f7;
   --bs-dark: #101828;
   --bs-primary-rgb: 21, 48, 132;
@@ -46,28 +46,28 @@
   --bs-danger-rgb: 212, 63, 58;
   --bs-light-rgb: 242, 244, 247;
   --bs-dark-rgb: 16, 24, 40;
-  --bs-primary-text-emphasis: #081335;
-  --bs-secondary-text-emphasis: #292d35;
-  --bs-success-text-emphasis: #173417;
-  --bs-info-text-emphasis: #055160;
-  --bs-warning-text-emphasis: #653208;
-  --bs-danger-text-emphasis: #551917;
+  --bs-primary-text-emphasis: rgb(8.4, 19.2, 52.8);
+  --bs-secondary-text-emphasis: rgb(40.8, 44.8, 53.2);
+  --bs-success-text-emphasis: rgb(23.2, 52.4, 23.2);
+  --bs-info-text-emphasis: rgb(5.2, 80.8, 96);
+  --bs-warning-text-emphasis: rgb(101.2, 50.4, 8);
+  --bs-danger-text-emphasis: rgb(84.9887850467, 25.0018691589, 23.2112149533);
   --bs-light-text-emphasis: #344054;
   --bs-dark-text-emphasis: #344054;
-  --bs-primary-bg-subtle: #d0d6e6;
-  --bs-secondary-bg-subtle: #e0e2e7;
-  --bs-success-bg-subtle: #d8e6d8;
-  --bs-info-bg-subtle: #cff4fc;
-  --bs-warning-bg-subtle: #ffe5d0;
-  --bs-danger-bg-subtle: #f6d9d8;
-  --bs-light-bg-subtle: #f9fafb;
+  --bs-primary-bg-subtle: rgb(208.2, 213.6, 230.4);
+  --bs-secondary-bg-subtle: rgb(224.4, 226.4, 230.6);
+  --bs-success-bg-subtle: rgb(215.6, 230.2, 215.6);
+  --bs-info-bg-subtle: rgb(206.6, 244.4, 252);
+  --bs-warning-bg-subtle: rgb(254.6, 229.2, 208);
+  --bs-danger-bg-subtle: rgb(246.4943925234, 216.5009345794, 215.6056074766);
+  --bs-light-bg-subtle: rgb(248.5, 249.5, 251);
   --bs-dark-bg-subtle: #98a2b3;
-  --bs-primary-border-subtle: #a1acce;
-  --bs-secondary-border-subtle: #c2c6ce;
-  --bs-success-border-subtle: #b0cdb0;
-  --bs-info-border-subtle: #9eeaf9;
-  --bs-warning-border-subtle: #fecba1;
-  --bs-danger-border-subtle: #eeb2b0;
+  --bs-primary-border-subtle: rgb(161.4, 172.2, 205.8);
+  --bs-secondary-border-subtle: rgb(193.8, 197.8, 206.2);
+  --bs-success-border-subtle: rgb(176.2, 205.4, 176.2);
+  --bs-info-border-subtle: rgb(158.2, 233.8, 249);
+  --bs-warning-border-subtle: rgb(254.2, 203.4, 161);
+  --bs-danger-border-subtle: rgb(237.9887850467, 178.0018691589, 176.2112149533);
   --bs-light-border-subtle: #e4e7ec;
   --bs-dark-border-subtle: #667085;
   --bs-white-rgb: 255, 255, 255;
@@ -93,11 +93,11 @@
   --bs-tertiary-color-rgb: 16, 24, 40;
   --bs-tertiary-bg: #f2f4f7;
   --bs-tertiary-bg-rgb: 242, 244, 247;
-  --bs-heading-color: #081647;
+  --bs-heading-color: rgb(7.5480769231, 21.8894230769, 70.9519230769);
   --bs-link-color: #153084;
   --bs-link-color-rgb: 21, 48, 132;
   --bs-link-decoration: none;
-  --bs-link-hover-color: #11266a;
+  --bs-link-hover-color: rgb(16.8, 38.4, 105.6);
   --bs-link-hover-color-rgb: 17, 38, 106;
   --bs-link-hover-decoration: underline;
   --bs-code-color: #d63384;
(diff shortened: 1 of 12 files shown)


wardi:
definitely nice to move this out of htmx-ckan and make the feature reusable

mutantsan:
@wardi I'm sorry, how is this not reusable? This is how to use it `ckan.toast({message: "Toast message"})`

And why do we need to move it out of the `htmx-ckan`? We need to show an error with the toast message on HTMX error.

wardi:
Sorry for being unclear, I should have said:
> It's definitely nice that you have moved this feature out of htmx-ckan and made it reusable

mutantsan:
> Sorry for being unclear, I should have said:
> 
> > It's definitely nice that you have moved this feature out of htmx-ckan and made it reusable

No worries, my English isn’t perfect so I got it wrong

ThrawnCA:
How easily are the notifications managed via keyboard?

ThrawnCA:
How easily are the notifications managed via keyboard?

How easily would automated testing like Selenium interact with them?

mutantsan:
> How easily are the notifications managed via keyboard?
> 
> How easily would automated testing like Selenium interact with them?

As it’s just a wrapper over the bootstrap toasts, you can read about it [here](https://getbootstrap.com/docs/5.1/components/toasts/#accessibility).

mutantsan:
@wardi can we proceed with the review? 

I've added a feature to add an optional toast message after a successful HTMX request. This is **an example** of how it can be used:

```jinja2
{% if following or am_following %}
  <a class="btn btn-danger" hx-toast-type="default" hx-toast-message="{{ _('You have unfollowed the {}'.format(obj_type)) }}" hx-post="{{ h.url_for(obj_type + '.unfollow', id=obj_id) }}" hx-target="#{{ anchor_type }}-info"><i class="fa-solid fa-circle-minus"></i> Unfollow</a>
{% else %}
  <a class="btn btn-success" hx-toast-type="default" hx-toast-message="{{ _('You have followed the {}'.format(obj_type)) }}" hx-post="{{ h.url_for(obj_type + '.follow', id=obj_id) }}" hx-target="#{{ anchor_type }}-info"><i class="fa-solid fa-circle-plus"></i> Follow</a>
{% endif %}
```

image

Lately, I plan to rework a group add/delete feature where I can use this. Also, sysadmin demote/promote feature

ckan/public-midnight-blue/base/javascript/htmx-ckan.js
```diff
@@ -40,41 +40,94 @@ function htmx_initialize_ckan_modules(event) {
...
+  }
+});
+
+class ToastHandler {
+  constructor(element, options = {}) {
+    this.defaultToastOptions = {
+      type: "success",
+      title: ckan.i18n._("Notification"),
+    };
+    this.toastAttributePrefix = options.toastAttributePrefix || "hx-toast-";
```
wardi: does naming this `hx-toast` imply that this is an htmx feature? Would `ckan-hx-toast` or similar be preferred?

ckan/public-midnight-blue/base/javascript/htmx-ckan.js
```diff
@@ -40,41 +40,94 @@ function htmx_initialize_ckan_modules(event) {
...
+      .replace(this.toastAttributePrefix, "")
+      .replace(/-([a-z])/g, (_, letter) => letter.toUpperCase());
+  }
+
+  parseAttributeValue(value) {
+    if (['true', 'yes', 't', 'y', '1'].includes(value.toLowerCase())) return true;
+    if (['false', 'no', 'f', 'n', '0'].includes(value.toLowerCase())) return false;
+    if (!isNaN(value) && value.trim() !== "") return Number(value);
+    return value;
+  }
```
wardi, line 119: This is a lot of magic. How many toast options are there and will automatically converting a set of boolean-looking values potentially cause problems?

ckan/public-midnight-blue/base/javascript/toast.js
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+    if (hasDelay && opts.showProgress) {
+      const progressEl = toast._createProgressBar(opts, style);
+
+      // reset progress bar to 100% on mouse enter
+      toastEl.addEventListener("mouseenter", function () {
+        progressEl.style.animation = "none";
+        progressEl.style.width = "100%";
+      });
```
wardi, line 92: mobile users can tap to keep a toast visible?

ckan/public-midnight-blue/base/javascript/toast.js
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    const opts = {
+      type: "default",
+      title: "",
+      icon: "",
+      subtitle: "",
+      delay: 3000,
+      position: "bottom-right",
+      showProgress: true,
```
wardi, line 44: is enabling a progress bar too visually busy as a default?

Please attach a capture of how this looks so we can discuss.

mutantsan: @wardi I don't have a strong opinion about it. One can say, that it might be "visually busy", but that it's better from the accessibility point of view. You want to know how much time you have to read the message before it disappear. 

toast-example

ckan/public-midnight-blue/base/javascript/htmx-ckan.js
mutantsan: What kind of problems are we talking about? I just need a way to pass something besides strings, specifically Booleans (for `showProgress` and `stacking`) and numbers (for `delay`). I see no magic here, it's similar to `tk.asbool`. I can delete everything except for `true` and `false` from the arrays, but what for?

ckan/public-midnight-blue/base/javascript/htmx-ckan.js
mutantsan: Hmm, I think you’re right. Maybe we [should use](https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes?utm_source=chatgpt.com) data- prefix attribute?

ckan/public-midnight-blue/base/javascript/htmx-ckan.js
wardi: If we have a numeric parameter this will convert 1's and 0's to true and false, potentially breaking them.

We could use a data attribute that contains JSON paramers and use JSON parsing instead of hand-rolling something.

ckan/public-midnight-blue/base/javascript/toast.js
wardi: @aleeexgreeen any thoughts about this kind of UI for toasts?

aleeexgreeen: Having the progress bar is helpful for users and the moving component can draw attention to the toast, which I think is a good thing. The main A11y/UI concern would be whether the users can pause/extend the timing of the toast and have the progress bar be paused as well. In your implementation, @mutantsan the toasts go back to full progress on mouseenter, and they don’t begin to count down until mouseleave, correct? (If so, I would just make sure mobile users have this same functionality)

Just an A11y note, because the toasts are stackable and each set to role alert, some screen reader toast content may be skipped/cut off due to new toasts popping up. Role status may be a better fit for stackable toasts.

ckan/public-midnight-blue/base/javascript/htmx-ckan.js
mutantsan: Oh, good catch with '1' and '0'. Thanks.

As about the data attribute, you mean one attribute to pass all the info? Like this?
```html
<span ckan-hx-toast='{"message": "Hello world", "delay": 1000, "showProgress": false}'></span>
```
Isn't it going to look heavy in HTML?

ckan/public-midnight-blue/base/javascript/toast.js
mutantsan: Yes, it works, I tested it on my iPad. My code just adds a progress bar and manages its animation. The bootstrap toast has their events to prevent toast from hiding.

ckan/public-midnight-blue/base/javascript/toast.js
mutantsan: Yes, it works, I tested it on my iPad. My code just adds a progress bar and manages its animation. The bootstrap toast has their events to prevent toast from hiding.

mutantsan:
@wardi   I've updated the `ToastHandler` to use `data-hx-toast` attribute, that expects a JSON-like string. Also, the `midnight-blue` theme was missing `htmx:confirm` event code, so I've added it.

ckan/public/base/javascript/htmx-ckan.js
```diff
@@ -70,41 +70,54 @@ document.body.addEventListener("htmx:afterSwap", function (event) {
...
-        const key = this.convertAttributeNameToKey(attr.name);
-        const value = this.parseAttributeValue(attr.value);
-        options[key] = value;
-      }
+    const attrValue = element.getAttribute(this.attrKey);
+    if (!attrValue) return this.defaultToastOptions;
+
+    try {
+      const parsed = JSON.parse(attrValue);
+      console.log(parsed);
```
wardi: forgotten debugging?

mutantsan: Oops, fixed ._.

wardi:
@mutantsan please add a changelog file describing the new feature and this is good to go

ThrawnCA (approved):

mutantsan:
@wardi done 👀

wardi:
@mutantsan it looks like your generated css makes a number of changes that aren't related like using `rgb(…)` instead of `#…` for colors, would you make sure you're generating the css with the same config/version of scss and resolve the conflict with master?

mutantsan:
@wardi. Done, I recompiled the CSS with the old `sass@1.74.1`. We updated it [here](https://github.com/ckan/ckan/commit/2300cf08562820f1cee28cd12d0f38682ee75297#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L28) 3 months ago.

Anyway, at some point we will have to embrace the new output and merge it. Also, other people should do `npm ci` to update the dependency when they're doing PRs.

mutantsan:
@wardi, hi, it would be nice if you could check it again 🙂

PR #9018, opened 2025-07-07
feat: add confirm.js module

mutantsan:
This PR introduces a reusable `ckan.confirm` module that displays Bootstrap 5 confirmation modal windows across CKAN. It provides a consistent and customizable way to request user confirmation before performing critical actions.

- Displays Bootstrap 5 modal confirmation dialogs with configurable options
- Supports different modal styles (primary, danger, success, etc.) with Bootstrap classes
- Allows optional title, message, icon, and custom button labels
-  Provides callback functions for both confirm and cancel actions (close button works as cancel action)
- Includes options for centering, scrollable content, full-screen mode, backdrop control, and keyboard behavior

I understand, that we don't have anything that will use it (yet). And we already have a `confirm-action.js` module, but it works only with the form and have a different approach. I still find it useful as we're trying to make the CKAN app more modern with `htmx` and all the things. I also want to update the CKAN admin a bit, and this and `ckan.toast` modules will be useful. 

I could use the `sweetalert2.js` for this, but didn't want to bring an extra dependency, as it seems, that we don't need much code to write a wrapper around bootstrap 5. 

image

Note: the screenshots were deleted, because I think, that they shouldn't be committed at all. It seems, that all of them reported an error during cypress tests run and only should be used by the developer for the debugging. 

### Features:

- [ ] includes API changes


wardi:
Please remove unrelated changes and see if confirm-action can be modified to use this new feature so that it doesn't become dead code

mutantsan:
@wardi I've restored the failed cypress tests screenshots and refactored the `confirm-action.js` to use `ckan.confirm`.

ThrawnCA:
How easily are the dialogs managed via keyboard?

How easily would automated testing like Selenium interact with them?

mutantsan:
> How easily are the dialogs managed via keyboard?
> 
> How easily would automated testing like Selenium interact with them?

As it’s just a wrapper over the bootstrap modals (that we already use everywhere), you can read about it [here](https://getbootstrap.com/docs/5.1/components/modal/).

mutantsan:
@wardi can we proceed with the review? 

I've also replaced a [default alert() ](https://htmx.org/examples/confirm/#vanilla-js-hx-confirm)for `htmx-confirm` with the ckan.confirm. This is an example (just an example, I'm not suggesting to add a confirmation window for following process, we should add it only on some "important" actions)

image

image

ckan/public-midnight-blue/base/javascript/confirm.js
```diff
@@ -0,0 +1,140 @@
...
+   *      console.log("User cancelled.");
+   *    }
+   * });
+   *
+   * Options:
+   * - message     (string)  [required] Message shown in modal body.
+   * - title       (string)  Optional.  Title in modal header.
+   * - icon        (string)  Optional.  HTML for icon before title.
+   * - confirmText (string)  Optional.  Confirm button label. Default: "Yes".
+   * - cancelText  (string)  Optional.  Cancel button label. Default: "No".
```
wardi: these aren't the defaults

ckan/public-midnight-blue/base/javascript/confirm.js
```diff
@@ -0,0 +1,140 @@
...
+    primary: { btnClose: "btn-close-white", main: "bg-primary text-white", confirm: "btn-primary" },
+    success: { btnClose: "btn-close-white", main: "bg-success text-white", confirm: "btn-success" },
+    warning: { btnClose: "btn-close-white", main: "bg-warning text-dark", confirm: "btn-warning" },
+    danger: { btnClose: "btn-close-white", main: "bg-danger text-white", confirm: "btn-danger" },
+    default: { btnClose: "", main: "", confirm: "btn-primary" }
+  };
+
+  confirm.selectors = {
+    modal: "#ckan-confirm-modal",
+    yesBtn: "#ckan-confirm-yes",
```
wardi: switch to `confirmBtn` to be consistent with `cancelBtn`?

ckan/public-midnight-blue/base/javascript/confirm.js
```diff
@@ -0,0 +1,140 @@
...
+    };
+
+    const style = confirm.styles[opts.type] || confirm.styles.primary;
+    const select = confirm.selectors;
+
+    if (document.querySelector(select.modal)) {
+      return console.error("Confirm: Modal already exists.");
+    }
+
+    const modalHtml = `
```
wardi: prefer the uppercase HTML in names because it's an acronym not a word and for consistency

mutantsan:
@wardi appreciate the review, fixed the issues

wardi:
Thanks, @mutantsan please also add a changes file describing the new feature to fix the last check.

mutantsan:
> Thanks, @mutantsan please also add a changes file describing the new feature to fix the last check.

Done

crstn-aiw:
Hello, we have recently installed CKAN 2.11 package and having the 'CSRF Token is missing' issue on most delete buttons (resource, dataset)- is this patch already applied or when it is coming? Thank you

image

PR #9024, opened 2025-07-16
Default exceptions handlers for non-debug mode

smotornyuk:
Any NotFound, ValidationError, or NotAuthorized raised from the view produces an "Internal server error" message in non-debug mode. While API calls are _usually_ wrapped into the try/except block, there a plenty of lines where action is called with an assumption that it won't fail. And sometimes such assumptions are wrong, especially when extension chains action and modify their validation or authorization rules.

In debug mode, such unhandled exceptions are shown with a stack trace, which is perfect. But in non-debug mode, "Internal server error" is often not acceptable.

This PR registers default error handlers for NotFound, ValidationError, and NotAuthorized. These handlers are activated only when debug is disabled, so the stack trace is still available. And these handlers are used only for uncaught exceptions - as long as you are using `base.abort`, you will never see them.

The table below highlights the problem with. Only the cell with the red icon is fixed, showing the normal error page. All other cells remain unchanged.

| | Uncaught exception | `base.abort` |
|-|-|-|
|debug = false | :red_circle: Internal server error | :green_circle: Normal error page |
|debug = true  | :green_circle: stack trace | :green_circle: Normal error page |

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -8,6 +8,7 @@ import importlib
 import inspect
 import pkgutil
 import logging
+import pprint
 
 from logging.handlers import SMTPHandler
 from typing import Any, Optional, Union, cast
@@ -29,7 +30,7 @@ from flask_babel import Babel
 
 from flask_login import LoginManager
 from flask_wtf.csrf import CSRFProtect
-from ckan.common import CKANConfig, asbool, session, current_user
+from ckan.common import CKANConfig, asbool, session, current_user, _
 
 import ckan.model as model
 from ckan.lib import base
@@ -59,6 +60,8 @@ from ckan.views import (identify_user,
                         _get_user_for_apitoken,
                         )
 from ckan.types import CKANApp, Config, Response
+from ckan.logic import NotFound, NotAuthorized, ValidationError
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


wardi:
If this hides the error from the logs maybe we could put logic in the 500 error handler that renders different pages based on the exception instead.

smotornyuk:
Yes, this sounds much better. Even though I focused on 3 common CKAN exceptions, there is no guarantees, that something else will be raised from the view. It will be much better to always show normal error page to client and at the same time, print all exception details to logs, no matter what exactly caused the problem

ckan/config/middleware/flask_app.py
```diff
@@ -489,78 +490,69 @@ def is_blueprint(mm: Any):
...
+            with tags.ul() as details:  # pyright: ignore[reportGeneralTypeIssues]
+                for field, text in e.error_summary.items():
+                    tags.li(f"{field}: {text}")
 
-def _not_valid_handler(error: ValidationError) -> tuple[str, int]:
-    """Generic handler for NotFound exception."""
-    message = _("Validation error:")
-    detail = h.clean_html(pprint.pformat(error.error_dict))
-    content = h.literal(f"{message} <pre><code>{detail}</code></pre>")
+            content = h.literal(f"{message}{details}")
```
wardi, line 539: use of literal here is unsafe, right?

wardi: my mistake, I'm unfamiliar with how dominate escapes string parameters.

Still, wouldn't it be better to pass this list to the template and let it render the tags?

wardi:
I'm uncomfortable with how this might leak information unintentionally e.g. if there's an internal call that triggers a validation error while passing parameters that the user isn't supposed to see.

Changing the "Server Error" message is one thing but rendering content from an exception during a failure to end users seems like a bad idea.

ckan/config/middleware/flask_app.py
smotornyuk: Generally, I like this option, but I don't know if it can cause any problems. Previously, `content` was a string, so any customization of the error template would use it and the whole message will be shown. 
If details are passed as dictionary - customized template will require an update. But it may be even better, because it's a new information.. yes, let's pass these details separately

smotornyuk:
It's a valid point. How about just showing `400 bad request` heading with `Validation error` message then? And leave current details for not found and not authorized - I assume these exceptions should not expose undesired information as they are expecting only plain strings. And I hardly can imagine anyone inserting sensitive information into the message for authorization error.

wardi:
Thanks @smotornyuk that sounds better to me.

PR #9025, opened 2025-07-16
IRender interface 

smotornyuk:
New interface with a method that can

* Change the name of the rendered template
* Modify data passed into the template

Use cases:

1. Add data to the template. Without the interface, developers are required to use non-idiomatic calls to data-producing helpers inside the template.
2. Change the template. Without the interface, developers are required to register a custom view on an existing endpoint and copy the code from the original view, replacing only the last line, where the name of the template is specified. There are alternatives, but they are also not entirely correct, so I won't mention them.
3. Conditionally render different template. Imagine that I want to implement HTMX shortcut for sorting on the organizations page. Without interface I have to override the whole endpoint. With the interface I can do something like
   ```py
   def prepare_render(name, data):
       if is_htmx():
            return name + ".htmx", data
       return name, data
   ```
   and register the corresponding HTMX template.

### DIFF
--- a/ckan/lib/base.py
+++ b/ckan/lib/base.py
@@ -85,22 +85,23 @@ def render_snippet(*template_names: str, **kw: Any) -> str:
         raise last_exc or TemplatesNotFound(template_names)
 
 
-def render(template_name: str,
-           extra_vars: Optional[dict[str, Any]] = None) -> str:
-    '''Render a template and return the output.
+def render(template_name: str, extra_vars: Optional[dict[str, Any]] = None) -> str:
+    """Render a template and return the output.
 
     This is CKAN's main template rendering function.
 
-    :params template_name: relative path to template inside registered tpl_dir
-    :type template_name: str
-    :params extra_vars: additional variables available in template
-    :type extra_vars: dict
+    :param template_name: relative path to template inside registered tpl_dir
+    :param extra_vars: additional variables available in template
 
-    '''
+    """
     if extra_vars is None:
         extra_vars = {}
 
     _allow_caching()
+
+    for plugin in p.PluginImplementations(p.IRender):
+        template_name, extra_vars = plugin.prepare_render(template_name, extra_vars)
+
     return flask_render_template(template_name, **extra_vars)
 
 
--- a/ckan/plugins/interfaces.py
+++ b/ckan/plugins/interfaces.py
@@ -32,37 +32,38 @@ if TYPE_CHECKING:
 
 
 __all__ = [
-    u'Interface',
-    u'IMiddleware',
-    u'IAuthFunctions',
-    u'IDomainObjectModification',
-    u'IFeed',
-    u'IGroupController',
-    u'IOrganizationController',
-    u'IPackageController',
-    u'IPluginObserver',
-    u'IConfigurable',
-    u'IConfigDeclaration',
-    u'IConfigurer',
-    u'IActions',
-    u'IResourceUrlChange',
-    u'IDatasetForm',
-    u'IValidators',
-    u'IResourceView',
-    u'IResourceController',
-    u'IGroupForm',
-    u'ITagController',
-    u'ITemplateHelpers',
-    u'IFacets',
-    u'IAuthenticator',
-    u'ITranslation',
-    u'IUploader',
-    u'IBlueprint',
-    u'IPermissionLabels',
-    u'IForkObserver',
-    u'IApiToken',
-    u'IClick',
-    u'ISignal',
+    "Interface",
+    "IMiddleware",
+    "IAuthFunctions",
+    "IDomainObjectModification",
+    "IFeed",
+    "IGroupController",
+    "IOrganizationController",
+    "IPackageController",
+    "IPluginObserver",
+    "IConfigurable",
+    "IConfigDeclaration",
+    "IConfigurer",
+    "IActions",
+    "IResourceUrlChange",
+    "IDatasetForm",
+    "IValidators",
+    "IResourceView",
+    "IResourceController",
+    "IGroupForm",
+    "ITagController",
+    "ITemplateHelpers",
+    "IFacets",
+    "IAuthenticator",
+    "ITranslation",
+    "IUploader",
+    "IBlueprint",
+    "IPermissionLabels",
+    "IForkObserver",
+    "IApiToken",
+    "IClick",
+    "ISignal",
+    "IRender",
 ]
 
 
@@ -2265,3 +2266,33 @@ class ISignal(Interface):
 
         """
         return {}
+
+
+class IRender(Interface):
+    """Control template rendering."""
+
+    def prepare_render(
+        self, template: str, extra_vars: dict[str, Any]
+    ) -> tuple[str, dict[str, Any]]:
+        """Modify rendering parameters.
+
+        This method receives and returns the name of the rendered template and
+        all the variables passed to it. Replace the name of the template to
+        render a different page, and patch variables when page requires more
+        data than view provides by default.
+
+        Example::
+
+            def prepare_render(self, template, extra_vars):
+                if tk.get_endpoint() == ("home", "about"):
+                    template = "home/better_about.html"
+                    extra_vars["additional_content"] = ...
+
+                return template, extra_vars
+
+        :param template: name of the rendered template
+        :param extra_vars: variables used in the template
+        :returns: name of the template and its additional variables
+
+        """
+        return template, extra_vars


amercader:
I really like this. In many cases you need to override a whole blueprint method just to pass an extra var to the template, and the htmx use case it's interesting too.

Modifying the response further to return a redirect, stream it, return json, etc is also useful sometimes, but I think finding a single place to hook into might be a bit more difficult (perhaps a Flask before response hook exists), and the plugins might not get all the input params passed to the view function. 

In any case if that exists it could be added as method to the same interface and users can choose whichever they need, eg:

```
class IViewRender()

     def prepare_render(self, template_name, extra_vars):
          pass

     def prepare_response(self, original_response):
          pass
     
```

If the second hook is too difficult to implement I think the one implemented here is worth merging. 
I'd rename the interface to `IViewRender` or `ITemplateRender` though

smotornyuk:
Thank you for the idea.

CKAN already has this place where page rendering time is recorded and CORS headers set on response. I added `prepare_response` hook there.

And, as this new interface affects templates and responses, I choose `IViewRender` name

ckan/config/middleware/flask_app.py
```diff
@@ -418,6 +418,9 @@ def ckan_after_request(response: Response) -> Response:
     # Set Cache Control headers
     response = set_cache_control_headers_for_response(response)
 
+    for plugin in PluginImplementations(IViewRender):
+        response = plugin.prepare_response(response)
```
wardi, line 422: `prepare_response` suggests something that happens before the response is generated. Here we're modifying an already-generated response.

If the intention is to completely replace a response it should be possible for a plugin to skip all other rendering code (rendering pages is often the slowest part of generating the response).

If we're replacing the existing response this way these plugins should be called in reverse plugins order to allow the earlier plugins to override settings from the later ones for consistency with template loading order etc.

wardi: I'd rather see something like the chained action interface: each plugin gets a chance to call the `up_func` (the next plugin in plugins order or the original render call) to generate the response

ckan/lib/base.py
```diff
@@ -85,22 +85,23 @@ def render_snippet(*template_names: str, **kw: Any) -> str:
...
 
-    '''
+    """
     if extra_vars is None:
         extra_vars = {}
 
     _allow_caching()
+
+    for plugin in p.PluginImplementations(p.IViewRender):
+        template_name, extra_vars = plugin.prepare_render(template_name, extra_vars)
```
wardi, line 103: How about `plugin.render_template_vars` or `plugin.render_parameters` since this method only modifies the parameters to the actual render call?

Again reverse order is needed so earlier plugins override later ones.

Is this the right place for the plugin? We only know the template name that was rendered and its parameters. We don't know which view was called and don't have access to other objects that could be useful from the view method.

It's uncomfortable to have a bunch of potentially fragile extra logic run inside a `render` method that looks like it should be fairly simple, and this function gets called all over the place in ckan. Do we really want every snippet to need to loop through a bunch of plugins as part of the rendering process?

We could make a new method that our core views would use instead of `render`, something like:
```python
def plugin_render_response(
    template_name: str,
    extra_vars: Optional[dict[str, Any]] = None,
    plugin_context: Optional[dict[str, Any]] = None,
) -> Any:
```

`plugin_context` could include anything that was computed in the view that might be useful for plugins when extending or overriding the response, like models and validated parameters passed, so that the plugins won't have to re-compute them.

Knowing which specific view was called might be useful too so that plugins can be conditional on the view not just the template name.

smotornyuk: The view/blueprint can be identified via `toolkit.get_endpoint()`. I don't expect someone to build conditions on the template name, especially since it can already be replaced due to this new interface.

I want to keep views as black boxes to a certain degree. Although exposing all internal variables may help in overoptimizing computations, it also introduces a contract regarding the implementation of the view, and any code that relies on internal variables becomes unstable when the view code is changed. 
Access to all intermediate variables used by view code is likely more fragile than accessing only variables from the template. Data that is passed to templates tends to remain approximately the same even when templates are changed or views are refactored, because we are trying not to break all the extensions that are extending templates and rely on the presence of certain variables. At the same time, the logic of view is encapsulated, and there are no restrictions regarding the changes in logic - view can be completely overridden as long as its output has the same form. 
So there must be balance between exposing just enough data to allow customizations, but not too much, to prevent tight coupling.


As for snippets, I agree, it doesn't have much value to hook into snippet rendering, because it's invoked from template, so it's the same as using custom helpers or overriding the template.
But instead of introducing a new method that must be used by views to be "hookable" and which most likely will be forgotten by extension developers, let's use the old `render` function. And instead, change snippet-rendering, creating a custom function that will bypass hook and plugin iteration and just normally render the snippet

ckan/config/middleware/flask_app.py
smotornyuk: The whole thing with responses can be done via middleware. Using `before_request` the whole view execution and template rendering, `after_request` can replace the final response. That's why I'm not completely sure, if this method worth adding. The only thing that is possible here and not possible in middleware - this hook is called before any of the middleware, even middleware from CKAN itself, while custom middleware rely on plugin order and always will be called after native CKAN middleware.

I like the idea of chaining, but if you see the better place for this hook or think it's better not to add it at all, I don't mind either

ckan/lib/base.py
wardi: Cool, I didn't know about `get_endpoint()`

I fully agree regarding not passing `locals()` from a view to plugins. I was suggesting being able to pass only specific things that are relevant but not passed to the template.

I'd like to know what @ckan/core thinks about adding this to our existing `render` function. Maybe there are other places that call render and don't expect a plugin to be able to inject itself into the process.

Also our `render` method returns a str so it can't do more interesting things like send redirects, stream a response or trigger a sendfile response and it would be useful for a new interface to support those things.

smotornyuk:
Instead of changing the existing function, this change can be implemented as an opt-in feature. The downside is that a lot of developers can ignore it and prevent further customizations of the view. But it's a question of good practices, and we can accept that not everyone wants to be flexible.

The good thing, if a new feature relies on new functions, they can do more. For example, imagine that views are using `render_response` instead of `render`. This new function returns a response, but we have no other expectations about its behavior. Its initial implementations are as follows:

```python
def render_response(template, data) -> ResponseLikeObject:
   # A) hook to modify template name
   # B) hook to modify variables
   # C) ...
   return render(template, data)
```

In this way, no extra work is performed if the developer does not want it. It returns a string at the moment, but the contract only guarantees something that can be treated as a Flask response, so no one will use this function to produce strings(for this purpose, we have good'n'old `render`)

--- 

Then, `render_response` is extended, computing ETag based on variables and returning `304 Not Modified` if a client already has the page with the same ETag.

```python
def render_response(template, data, /, etag = None):
   ...
   # C) hook that computes etag based on variables, without the actual rendering
   # D) ....
   if etag_match():
       return Response(304)

   return render(template, data)
```

This `render_response` initially was created as a function returning a response-like object. So we are allowed to return either the rendered template if the client does not have the page, or skip rendering if we predict that the template will be the same as the cached template from the client. When new  plugins come, they play, like `activity` that adds a notification counter, they will 

1. Use a hook to add variables [instead of computing them inside the template](https://github.com/ckan/ckan/blob/master/ckanext/activity/templates/header.html#L4)
2. Use the second hook to modify etag based on the changes that are introduced by new variables.

This etag-hook can be pretty simple. First, the base etag is computed by the view itself and passed into `render_response`. Then it goes through hook and if plugin introduces changes to the output, it will use the base etag and extra variables to compute new etag, i.e `compute_etag(base_etag + hash_from_changes)`

```python
def etag_hook(base_etag, template, data):
   if has_no_effect_on_content(data, endpoint, template, ...):
       return etag

   new_etag = md5(base_etag)
   # information that produces new blocks in the template
   new_element_key = data['custom_variable']['id']

   # if the user visits the same page that is customized by the plugin in the same way,
   # i.e same additional entity is rendered somewhere,
   # The new ETag will match, and we'll be able to skip the second render
   new_etag.update(new_element_key)
   return new_etag
```

---
    
Now we can add more hooks either before the response/render is produced to avoid computations, or after, to modify the result with extra headers, etc.

```python
def render_response(template, data, /, etag = None):
   ...
   # D) call hook with etag, data, template name that can emit custom response, like redirect


   result = ... # 304_if_etag_match else render template

   # E) hook that modifies the response  and adds more headers
   return result 
```


---

Finally, as @wardi suggested, we can pass an additional mapping with data that is not used in the template, but was used to compute variables. For example, if view pulls all the groups of the user and then just reduces this list into the list of unique roles that the user has, we can keep the list of groups inside `computation_context`, reducing the number of queries required for computing additional variables.

The fact that computation_context is not fixed and can be changed at any moment must be clearly expressed. Anyone who relies on variables from `computation_context` will provide a fallback logic to get all the data if it's missing. But in the best-case scenario, we'll improve performance even more, while in the worst case, we'll just do a normal amount of work.

```python
def render_response(template, data, /, etag = None, **computation_context):
   ...
```
---

After this is done, in a similar manner, we can add `redirect_response` and `file_response` that will reuse parts of `render_response`. Generally speaking, this `XXX_response` function will reflect the initial intention of the view, but all of them will be used as a flag for people who want to extend logic: "Here my plugin/CKAN core produces XXX-like response, but you are allowed to replace/modify it as you wish."

PR #9028, opened 2025-07-18
[#9027] improve CSV download using Datastore dump

### LINKED ISSUE #9027: Datastore dump is slow while downloading more then 1mln records
Postgres backend is using "offset" implementation and as a results, the iteration between items getting slower and slower due to "offset" trying to cut off previous Records.

Need to update the SQL queries to have much faster solution.

Engerrs:
Fixes: #9027 

## advanced datastore filters + fast datastore_search pagination

- `datastore_search` and `datastore_delete` `filters` now accept
  range values and nested AND and OR operations
- `datastore_search` now returns a `next_page` value with
  filters that can be used for fast pagination when records
  are returned sorted by their `_id` field
- datastore dump endpoint now uses fast pagination to more
  quickly stream large datasets as CSV, JSON, etc.


ckanext/datastore/backend/postgres.py
```diff
@@ -1470,13 +1472,19 @@ def search_data(context: Context, data_dict: dict[str, Any]):
         sort_clause = ''
 
     records_format = data_dict['records_format']
+    if last_id:
+        # Keyset Pagination
+        final_statement = 'WHERE _id > {last_id} {sort} LIMIT {limit}'
```
wardi: This won't work when a different sort order is given

ckanext/datastore/backend/postgres.py
```diff
@@ -1536,6 +1548,17 @@ def search_data(context: Context, data_dict: dict[str, Any]):
             records = LazyJSONObject(v)
         else:
             records = msgspec.json.decode(v)
+
+    # Currently apply Keyset Pagination only to CSV format
+    if records_format == u'csv':
+        m_csv_string = records.rstrip('\n').rsplit('\n', 1)
```
wardi: This is checking for more than one line in the records?

Engerrs:
Hi @wardi , thanks for noticing the sorting issue. I updated the logic a bit to take thins into account.

Keyset Pagination sorting is its issue, when there complex sorting from more then 1 sorting element, in such cases requires to modify the "WHERE" condition into tuples, which makes things more complex and I want to avoid this for now, as we will need to re-write the "where_clause" logic. There also an scenario in which the query becomes even more complex, is where the sorting elements have different sorting order (e.g. one asc and the second desc), this can require splitting the query with "OR"s.

Due to the fact the Keyset Pagination initially was added to support quicker pure download for datastore dump, I would leave for now the logic to be able to support at least basic scenarios of sorting (e.g. asc and desc for _id) and conditioning. For all other more complex filters and sorting, it would use "offset", keeping in mind, that after filtering the amount of results becomes much lower.

wardi:
Thanks @Engerrs 

Instead of adding a new `last_id` parameter to the `datastore_search` API let's implement a special case of the [upcoming generic search filters](https://github.com/ckan/ckanext-search/blob/main/doc/filters.md) feature.

e.g. If `datastore_search` parameters contain:
```
{
  "filters": {
    "_id": {"gt": 123},
    …
  },
  …
}
```
Then treat "123" as the `last_id` value.

This way once the new generic search feature is implemented this will stop being a special case and users won't need to update their API usage. Document it as a special case for efficient pagination in `datastore_search` as part of this PR.

It's fine if it doesn't help with paginating other sort orders since we're being clear that we only want `_id` values greater than the number passed.

In this PR the same filter should work for all record formats, not just CSV.

ckanext/datastore/backend/postgres.py
```diff
@@ -1536,6 +1558,25 @@ def search_data(context: Context, data_dict: dict[str, Any]):
...
+            if m_csv_string:
+                try:
+                    m_csv_string = m_csv_string[-1]
+
+                    # Turn to int to make sure no broken value
+                    last_id = int(m_csv_string.split(',')[0].strip())
+                    data_dict['last_id'] = last_id
+                except (ValueError, IndexError, AttributeError):
+                    pass
+
```
wardi: let's avoid csv parsing by having the `datastore_search` endpoint return the last id in its response

Engerrs:
Hi @wardi , your last request required full rebuild of the initial implementation, which was a bit challenging to integrate it with the original logic and not break things.

The main updates:
1. Added constant for the operator logic + updated _where_clause to take into account the dict values, which have the operator and value.
2. There is no option of getting the "last_id" in the response, as different types (e.g. xml, json, csv) provide different output of the results, which will require similar parsing as done previously for CSV. Due to this, I added additional query for getting the last_id independently from the chosen type. I checked the performance and there no change in it after the additional query was added.
3. Additionally I had to add "last_id_operator" in the response, as the new approach of using filters with "operator": "value" approach requires the "operator" to make it work. That's why  "last_id_operator" provides this information in the "stream_dump" for the next iteration.
4. Updated the docstring + the .rst file for Datastore. I'm not that good at documentations, so feel free to modify it if needed.

ckanext/datastore/backend/postgres.py
```diff
@@ -85,6 +85,14 @@
 _UPSERT = 'upsert'
 _UPDATE = 'update'
 
+_OPERATORS = {
+    'eq': '=',
+    'gt': '>',
+    'gte': '>=',
+    'lt': '<',
+    'lte': '<=',
+    'in': 'in',
```
wardi: let's drop `'in'` because it's not required for pagination and we don't have tests covering its edge cases here

ckanext/datastore/backend/postgres.py
```diff
@@ -488,6 +499,16 @@ def _where_clauses(
     return clauses
 
 
+def _prepare_where_operator_and_value(value: dict[str, Any]) -> tuple[str, Any]:
+    try:
+        key = next(iter(value))
+        val = value[key]
+    except (StopIteration, KeyError):
+        return '=', value
```
wardi: we can ignore the multiple-value dict case too with:
```suggestion
    try:
        [(key, val)] = value.items()
    except ValueError:
        return '=', value
```

Engerrs:
Hi @wardi , done according to your request.

ckanext/datastore/backend/postgres.py
EricSoroos: This server-errors if you pass in a dictionary of:
```
field: {lt: val1, gt: val2}
```
as it winds up passing field = '{"lt": val1, "gt": val2}' to the database, and this isn't castable to any numeric format. 

Suggest something like: 
```
def _prepare_where_operator_and_value(value: dict[str, Any]) -> list[tuple[str, Any]]:
    try:
        return [(_OPERATORS[key],val) for key, val in value.items()]
    except KeyError as key:
        raise ValidationError(_("{0} is not a valid operator").format(key)
```

And then, since we're changing that return type to a list of tuples:
```
        else:
            field_filters = [('=', value)]
            if isinstance(value, dict):
                field_filters = _prepare_where_operator_and_value(value)

            for operator, value in field_filters:
                if fields_types[field] == 'text':
                    # pSQL can do int_field = "10"
                    # but cannot do text_field = 10
                    # this fixes parity there.
                    value = str(value)
                placeholder = f"value_{next(idx_gen)}"
                clause: tuple[Any, ...] = (
                    f'{sa.column(field)} {operator} :{placeholder}',
                    {placeholder: value}
                )
                clauses.append(clause)
```

To be fair, this isn't required for the specific use case of improving datastore dump, and this really just adds a backdoor feature for requesting ranges against the datastore. However, the error that we're

EricSoroos:
I'm seeing about a 2x speedup on my system with a 1.5M row dataset.  (In this case, 5 vs 10 sec)

wardi:
@Engerrs @EricSoroos in [5893dc5](https://github.com/ckan/ckan/pull/9028/commits/5893dc5627c086b94d69b76a9e7365c3f8a943e6) we're now getting the next id value as part of the same query that retrieves the data when possible, and the `next_page` return value can now be added directly to the filters parameter of the next call.

Let me know what you think.

wardi:
I was hoping to fix the issue of duplicated/missing rows that can happen when a rows are added/removed while a dump is running. Getting the next id as part of the same query solves this but for csv/tsv there's still a race condition where the next_id returned won't align with the records generated.

I could solve this by creating a temporary table, but if the temporary space configured in postgres is too small this could result in useless disk writes

I could calculate the `next_id` *first* then modify the query that generates the records to stop at the row with that id instead of using a `limit` but this could mean generating a page that is larger or smaller than what was requested. In the worst case it could even generate an empty set of results before the end of the actual records are released.

It might also be possible to generate csv/tsv data without `copy to` but that would be a much more complicated query (darn you postgres syntax for not being composable)

EricSoroos:
> I was hoping to fix the issue of duplicated/missing rows that can happen when a rows are added/removed while a dump is running. 

The only reliable way to do this is to rely on the database concurrency management to ensure that we're running in a single transaction or single query. 

> I could solve this by creating a temporary table, but if the temporary space configured in postgres is too small this could result in useless disk writes

I think this would be very bad for (large) dump tables where we're getting hit by multiple concurrent downloads.

Expanding that a bit -- we'd be trading (potentially cached) disk read pressure for read + write traffic. You might not see this at any given table size, but as the table size increases, you're more and more limited by disk access, and then you'd kick over into writing temporary tables and blowing your cache.

wardi:
CSV dump time for a real 13M record dataset:

before | after | change
--- | --- | ---
30:09 | 2:08 | 15× faster

Engerrs:
@wardi this is great!

ckanext/datastore/blueprint.py
```diff
@@ -141,11 +148,7 @@ class DictionaryView(MethodView):
 
     def _prepare(self, id: str, resource_id: str) -> dict[str, Any]:
         try:
-            check_access(
-                "datastore_create",
-                context={"user": current_user.name, "auth_user_obj": current_user},
-                data_dict={"resource_id": resource_id},
-            )
+            h.check_access("datastore_create", {"resource_id": resource_id})
```
EricSoroos: This is incorrect.  Calling h.check_access and not checking the return boolean essentially bypasses the check.

wardi: reverted. thanks

wardi:
@EricSoroos fixed my earlier fix, let me know how this looks

EricSoroos (approved):

EricSoroos:
(what this doesn't help is That One Guy downloading an entire 700k element dataset using datastore_search and limit offset pagination, by 100 records at a time)

wardi:
@EricSoroos I'm in favor of putting a hard limit on the accepted `offset` values to *encourage* use of more efficient pagination

EricSoroos:
I've done that recently on a non-ckan project.

amercader:
@wardi it looks like merging this broke master (and 2.12 as I cherry picked the changes):

https://github.com/ckan/ckan/actions/runs/25790338655/job/75754184258

The branch wasn't up to date with master and there are conflicts with https://github.com/ckan/ckan/pull/9298. Basically `WHITELISTED_RESOURCES` is gone from datastore/logic/action.py but is still imported and used from the datastore blueprint. Would you mind updating the blueprint to the new logic against latest master?

PR #9030, opened 2025-07-22
Hide users from stats

avdata99:
### Proposed fixes:

Hide users data from `stats` if the `ckan.auth.public_user_details` setting is set to `False`.
This ensures that user details are not exposed in the statistics when public user details are disabled.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/stats/blueprint.py
+++ b/ckanext/stats/blueprint.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from typing import Any
 from flask import Blueprint
 
-from ckan.common import config
 from ckan.plugins.toolkit import render, h
 import ckanext.stats.stats as stats_lib
 
--- a/ckanext/stats/blueprint.py
+++ b/ckanext/stats/blueprint.py
@@ -1,9 +1,9 @@
 # encoding: utf-8
 from __future__ import annotations
 from typing import Any
-
 from flask import Blueprint
 
+from ckan.common import _, config
 from ckan.plugins.toolkit import render, h
 import ckanext.stats.stats as stats_lib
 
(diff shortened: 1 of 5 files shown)


ckanext/stats/stats.py
```diff
@@ -116,6 +117,13 @@ def top_tags(cls, limit: int = 10,
 
     @classmethod
     def top_package_creators(cls, limit: int = 10) -> list[tuple[Optional[model.User], int]]:
+        if not config.get('ckan.auth.public_user_details'):
+            log.warning(
```
pdelboca:
```suggestion
            log.debug(
```

@avdata99 let's change the log verbosity to `debug`. This isn't a malfunction or something going wrong, it's just a configuration value so `debug` should be enough.

ckanext/stats/blueprint.py
```diff
@@ -14,17 +14,23 @@
...
         'top_tags': stats.top_tags(),
-        'top_package_creators': stats.top_package_creators(),
         'most_edited_packages': stats.most_edited_packages(),
         'new_packages_by_week': stats.get_by_week('new_packages'),
         'deleted_packages_by_week': stats.get_by_week('deleted_packages'),
         'num_packages_by_week': stats.get_num_packages_by_week(),
-        'package_revisions_by_week': stats.get_by_week('package_revisions')
+        'package_revisions_by_week': stats.get_by_week('package_revisions'),
+        # Stats with users info
+        'top_package_creators': None,
```
pdelboca: To keep the style of the function let's keep calling `stats.top_package_creators()`  directly here instead of initializing it with `None`. 

All the logic for checking the config is encapsulated in the method so it is safe. We can avoid an extra `if` statement.

pdelboca (commented):
@avdata99 looks good! Just some minor comments :)

ckanext/stats/blueprint.py
avdata99: You're right, done here https://github.com/avdata99/ckan/commit/b6324d96709192a4dbaf49e121cf7be9ad6384cd

pdelboca:
Thanks @avdata99 !

PR #9033, opened 2025-07-23
Fix header header site navigation place

### LINKED ISSUE #9032: header_site_navigation_tabs block is misplaced, so it makes it impossible to rewrite.
## CKAN version
Latest version (e.g. CKAN 2.11.3)

## Describe the bug
In **header.html** there an block **header_site_navigation**, which theoretically should wrap the **ul** element, so its possible to overwrite it in custom template that extends **header.html**.
But **header_site_navigation** currently is within the **ul**, which makes the overwrite almost impossible. 

### Expected behavior
**header_site_navigation** wraps the whole **ul** element.

Engerrs:
Fixes #9032

### DIFF
--- a/ckan/templates/header.html
+++ b/ckan/templates/header.html
@@ -94,8 +94,8 @@
       </button>
 
       <div class="main-navbar collapse navbar-collapse" id="main-navigation-toggle">
+      {% block header_site_navigation %}
         <ul class="navbar-nav ms-auto mb-2 mb-lg-0">
-          {% block header_site_navigation %}
             {% block header_site_navigation_tabs %}
               {% set org_type = h.default_group_type('organization') %}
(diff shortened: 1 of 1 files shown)


EricSoroos:
It looks like there's a bug in the existing template where the ul is opened outside the block but closed inside. 

Can you clean up the commit so that it doesn't contain unrelated whitespace changes?

avdata99:
This was already fixed for the `midnight-blue` template at `ckan/templates-midnight-blue/header.html`
+1 to the change

Engerrs:
@EricSoroos , updated the PR.

EricSoroos:
We need a changelog entry, but otherwise LGTM.

EricSoroos (approved):

Issue #8789, opened 2025-04-12
demo.ckan.org still on 2.9.9 (May 24, 2023)


duttonw:
Its nearly 2 years from current 2.11.x.

Should this site be automated to be on last major release patch or even master (maybe containerize as thats now more mature).

https://demo.ckan.org/api/action/status_show
```
{"help": "https://demo.ckan.org/api/3/action/help_show?name=status_show", "success": true, 
"result": 
{"site_title": "CKAN Demo", 
"site_description": "", 
"site_url": "https://demo.ckan.org", 
"ckan_version": "2.9.9",
 "error_emails_to": null, "locale_default": "en",
 "extensions": ["envvars", "s3filestore", "image_view", "text_view", "recline_view", "recline_grid", "recline_graph", "recline_map", "datastore", "datapusher", "pdf_view", "geo_view", "resource_proxy", "stats"]}}
```

wardi:
Stay tuned, @EricSoroos mentioned working on setting up a replacement demo.ckan.org, it should be up soon.

PR #9037, opened 2025-07-23
chore: update API docs to show modern Python usage (ckanapi instead of urllib2)

### LINKED ISSUE #9000: API docs are stale and should use more modern Python code
I see the API docs are largely 13 years old. They suggest making Python API calls with `urllib2` but this is legacy code from Python 2. In Python 3, people are using `urllib`, `urllib3`, or most commonly `requests` which wraps `urllib3`. See https://stackoverflow.com/questions/2018026/what-are-the-differences-between-the-urllib-urllib2-urllib3-and-requests-modul

I am happy to swap in code that uses `requests` instead of `urllib2` if that's a desired contribution.

sfirke:
Fixes #9000 

### Features:

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -81,19 +81,29 @@ in a terminal::
 
     curl https://demo.ckan.org/api/3/action/group_list
 
+Alternatively, for Python users, we recommend using the `ckanapi <https://github.com/ckan/ckanapi>`_ 
+library which provides a simpler and more robust way to interact with CKAN APIs. 
+To install it, run::
+
+    pip install ckanapi
+
+Then you can make the same API call with just a few lines of Python code::
+
+from ckanapi import RemoteCKAN
+
+ckan = RemoteCKAN('https://demo.ckan.org')
+result = ckan.action.group_list()
+print(result)
+
+The ckanapi library automatically handles JSON encoding/decoding, error handling, and 
+provides convenient shortcuts for all CKAN API actions.
+
 The response from CKAN will look like this::
 
     {
-        "help": "...",
-        "result": [
-            "data-explorer",
-            "department-of-ricky",
-            "geo-examples",
-            "geothermal-data",
-            "reykjavik",
-            "skeenawild-conservation-trust"
-        ],
-        "success": true
+        "help": "https://demo.ckan.org/api/3/action/help_show?name=group_list",
+        "success": true,
+        "result": ["david", "roger"]
     }
 
 The response is a JSON dictionary with three keys:
@@ -117,7 +127,7 @@ The response is a JSON dictionary with three keys:
 2. ``"result"``: the returned result from the function you called. The type
    and value of the result depend on which function you called. In the case of
    the ``group_list`` function it's a list of strings, the names of all the
-   datasets that belong to the group.
+   groups on the site.
 
    If there was an error responding to your request, the dictionary will
    contain an ``"error"`` key with details of the error instead of the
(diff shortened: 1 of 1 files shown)


sfirke:
I wasn't sure what to do about showing the response coming from `demo.ckan.org`. Clearly the result in the API docs was wrong, so I put the result I am getting now. It seems like maybe there's an update coming to the demo site soon per https://github.com/ckan/ckan/issues/8789#issuecomment-2807199093, in which case maybe this example can be updated again after that.

doc/api/index.rst
```diff
@@ -81,19 +81,27 @@ in a terminal::
...
-            "department-of-ricky",
-            "geo-examples",
-            "geothermal-data",
-            "reykjavik",
-            "skeenawild-conservation-trust"
-        ],
-        "success": true
+        "help": "https://demo.ckan.org/api/3/action/help_show?name=group_list",
+        "success": true,
+        "result": ["david", "roger"]
```
wardi, line 104: ckanapi unwraps results so the response you get will be simply:
```
['david', 'roger']
```

doc/api/index.rst
```diff
@@ -382,6 +419,18 @@ request, with curl you can do this using the ``@file.csv``::
...
+The same operation can be done with ``ckanapi``::
+
+    from ckanapi import RemoteCKAN
+    
+    ckan = RemoteCKAN('https://demo.ckan.org', apikey='XXXX')
+    
+    with open('updated_file.csv', 'rb') as f:
+        result = ckan.action.resource_patch(
+            id='<resource_id>',
+            upload=('updated_file.csv', f)
```
wardi, line 431: does passing a tuple like this work with ckanapi?

wardi:
There was some concern at the dev meeting about using ckanapi in our examples instead of requests directly.

Our documentation should encourage users to use the best tools for the task, and I believe ckanapi is such a tool for a number of reasons.

However ckanapi does hide some of the details of calling our APIs that it might be useful for users to be aware of, in particular if they are using another language to make requests that doesn't come with a ckan API wrapper library:

- our responses include an envelope that indicate whether a call was successful
  - ckanapi automatically unwraps the envelope and converts errors to appropriate exceptions so they aren't silently ignored

- our API accepts POST with a JSON body if there are no files as part of the request

- when there are files the API request must be sent as a multipart POST request, with each top level parameter being passed as a string
  - we lose JSON types for parameters so nested parameters are typically not supported when combined with file uploads
  - there is no automatic parsing of JSON parameters sent in multipart POST requests
  - as a workaround some parameters will accept a comma-separated string value so they can be parsed as a string

- "side-effect-free" or "read-only" requests can also be sent as GET requests using URL parameters
  - URL parameters are treated as strings with no automatic parsing of JSON
  - as a result endpoints like datastore_search with a filters parameter can't be made using GET requests

We could raise some or all of these points in the documentation, possibly with examples using curl or other languages. IMHO we should use ckanapi for python examples, but if we do use requests for some examples let's make sure each example includes proper error handling so that we don't encourage bad practices.

sfirke:
I agree with the desire to show some examples with requests. A colleague is working in a Python environment where they can't install `ckanapi` and so must use `requests`.

I can indicate how using `requests` requires additional error handling that comes along with `ckanapi`.

I am inclined to lead off with a section that shows how you can interact with the API in a similar way using curl, R, and python (with requests). Then show the wrapper packages:
- Python has ckanapi
- R has [ckanr](https://cran.r-project.org/web/packages/ckanr/index.html)

Then we can keep the `ckanapi` and `ckanr` examples to a minimum and instead point to those packages' respective documentation. Is https://github.com/ckan/ckanapi the best documentation link for ckanapi?

wardi:
Sounds good.

Yes the readme on https://github.com/ckan/ckanapi is the only documentation at the moment

wardi (changes requested):
@sfirke please add the requests examples as discussed and I will merge

wardi:
This is an improvement over what we have right now so I'll merge as-is. If someone else would like to expand the examples using requests or another library (I've seen requests being criticized as haven frozen development for a few years, maybe there's a better option) that can be done in a separate PR.

PR #9041, opened 2025-07-27
[Midnight blue] Resource Management

aleeexgreeen:
Fixes:
[Discussion comment](https://github.com/ckan/ckan/discussions/8188#discussioncomment-13414804)

### Proposed fixes:
- Replaces 'edit resource' and 'views' buttons with one 'manage' button
- Adds button to 'Add new resource' on dataset page
- Removes 'Add new resource' link from sidebar of new resource pages
- Moves resource buttons to left of 'manage' button

### Features:

- [ ] includes API changes


wardi:
One result of this change that is a little odd is that from resource 1 (unnamed resource) I can still navigate directly to edit the views on resource 2 (another one) but not the views on resource 1.

Screenshot from 2025-07-31 16-48-44

Also when the DataStore is active a "Data Dictionary" button normally appears here next to the manage and views buttons, and when xloader/datapusher is active a "DataStore" button normally appears here too.

If we decided that this manage button should be a dropdown like the ones in the sidebar and include "Views", "DataStore", "Data Dictionary" then this might be fine.

aleeexgreeen:
Essentially, if we go with a dropdown button for manage there would be “views”, “datastore”, and the “data dictionary” options and those should also be present in the resources dropdown button in the sidebar? If we don’t go with a dropdown button for manage then we should remove the datastore and data dictionary buttons and the edit views option from the sidebar?

amercader:
Maybe a better way forward is for the small wrench icons in the sidebar to only point to "Manage" like the big Manage button on top. In some circumstances like if someone is editing just the views in all resources it might mean some more clicks but I think is better to have consistency, e.g. than to not show DataStore in the wrench icons as it happens now.

wardi:
I do like having direct navigation resource pages like the data dictionary from any resource. The resource editing pages can render very slowly and adding an extra click actually adds many seconds to the navigation.

amercader:
OK, so then we need the wrench on all sidebar resources (including the active one), and that should be a dropdown with all items in the Manage top bar, right?

@aleeexgreeen The items on the dropdown are added by plugins by extending the relevant blocks. E.g. the "DataStore" button on the top bar is added [here](https://github.com/ckan/ckanext-xloader/blob/594891a69c3be40c1beda02be83ec20fd937135a/ckanext/xloader/templates/package/resource_edit_base.html#L3) and the "DataStore" item in the sidebar dropdown [here](https://github.com/ckan/ckanext-xloader/blob/master/ckanext/xloader/templates/package/snippets/resource_item.html). As long as the blocks don't change they should still come up in the new theme

aleeexgreeen:
@amercader I just wanted to confirm whether the datastore button should still be visible in the button list of the resource page?

amercader:
> I just wanted to confirm whether the datastore button should still be visible in the button list of the resource page?

Sorry @aleeexgreeen I can't remember the exact context for this, but if the `datastore` plugin is active, the DataStore button should be visible in both the Manage page top bar and the sidebar wrench dropdown (like the rest of manage items like Views, Data Dictionary etc). Does this help?

amercader:
@aleeexgreeen I was writing a long comment with screenshots listing all UI elements that were not showing up in the resource page and then I realized it was that no templates from the datastore plugin were loaded at all.
The reason are these lines in the datastore plugin:


They must have been added in the past to load two different sets of templates like we are doing with the activity plugin, but as there is no `templates-midnight-blue` folder, when `ckan.base_templates_folder` is set to that value an non-existing templates folder is registered.

Let's just change it back to load just the `templates` folder:

```diff
diff --git a/ckanext/datastore/plugin.py b/ckanext/datastore/plugin.py
index da5c291c4..fec72e0b3 100644
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -78,9 +78,7 @@ class DatastorePlugin(p.SingletonPlugin):
         DatastoreBackend.register_backends()
         DatastoreBackend.set_active_backend(config)

-        templates_base = config.get('ckan.base_templates_folder')
-
-        p.toolkit.add_template_directory(config, templates_base)
+        p.toolkit.add_template_directory(config, 'templates')
         p.toolkit.add_resource('assets', 'ckanext_datastore')
         self.backend = DatastoreBackend.get_active_backend()
```

With this change the datastore UI elements show up [*]:

Screenshot 2025-10-02 at 13-28-13 Test new files - API 1 - CKAN

The green "Data API" button should be shown to both logged in and non-logged in users, so that's good (actually it would be great if you could do a pass on the styling of the popup that opens when clicking it)

But the "Data Dictionary" is a more interesting one. According to the description, manage-like buttons (i.e. those that require permissions) have all moved to the Manage section:

> Replaces 'edit resource' and 'views' buttons with one 'manage' button

So what do we do, with the Datastore's "Data Dictionary" button and more in general with other extensions adding buttons to the action section:
* We could be strict about it and remove the `action_manage_inner` block in the midnight theme to ensure no one adds a button there (current theme will still work). This sounds extreme and goes against the goal of making CKAN customizable, not mentioning that it will break existing extensions
* We selectively remove it on the datastore templates if on the midnight blue templates. This means a new set of templates that gets loaded like in the activity plugin, which sounds overkill for just one button, or a custom helper (which is ugly, as will have the theme name hardcoded)
* We leave the button as is
* * We leave the button as is and bring back all the other buttons removed (Edit resource and Views)

@aleeexgreeen @wardi any thoughts?

---

[*] If you want to test this locally
1. [Set up the DataStore](https://docs.ckan.org/en/2.11/maintaining/datastore.html#setting-up-the-datastore)
2. Run the following command to create a Datastore-only resource:
```
curl -X POST -H "Authorization: <YOUR_TOKEN_HERE>" http://localhost:5000/api/action/datastore_create  -H 'Content-Type:application/json' --data-raw '{"resource": {"name": "API 1", "package_id": "<YOUR_DATASET_NAME_HERE>"},"fields": [{"id": "id", "type": "numeric"}, {"id": "name", "type": "text"}], "primary_key": "id", "records":[{"id": 1, "name": "cat"}, {"id": 2, "name": "dog"}, {"id": 3, "name": "rabbit"}]}'
```

aleeexgreeen:
@amercader I think it defeats the purpose of removing the "Edit resource" and "Views" buttons if we keep the "Data dictionary" button, because the data dictionary is one of the "tabs" when editing the resource. Seems counterintuitive to keep both, so I would vote against keeping the current layout. I also think the old way of having separate buttons for each of the functions of editing a resource a bit redundant, but given the options I would say that is the best one for this situation.

wardi:
If all the navigation drop-downs are shown on the left (including for the selected resource) then that would be the place we want links to the datastore, data dictionary etc. tabs to appear now instead of along the top, right?

amercader:
Removing all action buttons except the Manage one from the top is the best approach, the issue is how to implement it on extensions that are adding currently buttons themselves and must still work on the current theme, i.e. they need to show the button on the now default theme but not on the midnight blue one.

Here for instance, "DataStore" is added by ckanext-xloader and "Data Dictionary" by ckanext/datastore:

Screenshot 2025-10-06 at 15-36-39 Test new files - libraries csv - CKAN

The "cleanest" way would to have two full different sets of templates for each theme and each extension but that seems overkill. So perhaps a helper might be enough. It's not super elegant but at least we can update it in the future if the default theme changes and xloader or others can use the same helper.

what do you think?

```diff

diff --git a/ckanext/datastore/helpers.py b/ckanext/datastore/helpers.py
index 308bd43b6..f92e6ecfc 100644
--- a/ckanext/datastore/helpers.py
+++ b/ckanext/datastore/helpers.py
@@ -249,3 +249,11 @@ def datastore_rw_resource_url_types() -> list[str]:
     datastore_delete
     """
     return ["datastore"]
+
+
+def datastore_show_resource_actions():
+    """
+    Extensions should not show action buttons (i.e.) next to the Manage / Data API core ones
+    """
+
+    return "midnight-blue" not in tk.config.get("ckan.base_templates_folder")
diff --git a/ckanext/datastore/plugin.py b/ckanext/datastore/plugin.py
index 2251d9f5a..289ac4cb9 100644
--- a/ckanext/datastore/plugin.py
+++ b/ckanext/datastore/plugin.py
@@ -258,11 +258,13 @@ class DatastorePlugin(p.SingletonPlugin):
         conf_dictionary = datastore_helpers.datastore_dictionary
         conf_sql_enabled = datastore_helpers.datastore_search_sql_enabled
         rw_url_types = datastore_helpers.datastore_rw_resource_url_types
+        datastore_show_resource_actions = datastore_helpers.datastore_show_resource_actions

         return {
             'datastore_dictionary': conf_dictionary,
             'datastore_search_sql_enabled': conf_sql_enabled,
             'datastore_rw_resource_url_types': rw_url_types,
+            'datastore_show_resource_actions': datastore_show_resource_actions,
         }

     # IForkObserver
diff --git a/ckanext/datastore/templates/package/resource_read.html b/ckanext/datastore/templates/package/resource_read.html
index f1247f52f..3381f96ea 100644
--- a/ckanext/datastore/templates/package/resource_read.html
+++ b/ckanext/datastore/templates/package/resource_read.html
@@ -43,7 +43,7 @@

 {% block action_manage_inner %}
   {{ super() }}
-  {% if res.datastore_active %}
+  {% if res.datastore_active and h.datastore_show_resource_actions() %}
     <li>{% link_for _('Data Dictionary'), named_route='datastore.dictionary', id=pkg.name, resource_id=res.id, class_='btn btn-secondary', icon='code' %}</li>
   {% endif %}
 {% endblock %}
``

PR #9049, opened 2025-07-30
CSRF token session improvements

amercader:
1. The CSRF token in the session was only cleared in the logout view function (i.e. when hitting the /logout endpoint), but logouts can happen programmatically calling the `logout_user()` function from the toolkit. Moved the logic to a single wrapper in lib/common.py
2. The `rotate_token` function only generated a new CSRF token if one was already present in the session. This was the case for a normal UI request, but not always if calling `rotate_token` directly. Moved back one indentation level to always generate the token.

Added a test to make sure the token is cleared from the session on logout.

### DIFF
--- a/changes/9049.misc
+++ b/changes/9049.misc
@@ -1,0 +1,1 @@
+Clear CSRF token from session when logging out programmatically.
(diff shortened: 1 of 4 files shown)


amercader:
I worked on this while investigating whether to clear the whole session on logout. I opted for not doing as `logout_user()` (the Flask-login one) will remove the user identifier keys from the session, but if we decide to do it now we have a single place to do it.

amercader:
@EricSoroos this might be worth getting in 2.12

PR #9056, opened 2025-08-04
Replace Legacy Munge Filename with Munge Filename

### LINKED ISSUE #9051: Deprecated `munge_filename_legacy()` functions is used
According to the [docstring](https://github.com/ckan/ckan/blob/23b6599f148679277eedc3bcb10d861376e151e0/ckan/lib/munge.py#L125) the function is deprecated and not working correctly but it is still used in `lib/uploader.py` and `lib/dictization/model_dictize.py`.

It should be replaced with `munge_filename()` and removed.

diaslui:
Fixes #9051 

### Proposed fixes:

Only the deprecated legacy function has been replaced by the current one.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9018.feature
+++ b/changes/9018.feature
@@ -1,0 +1,1 @@
+Add reusable `ckan.confirm` module that displays Bootstrap 5 confirmation modal windows across CKAN
--- a/ckan/public-midnight-blue/base/javascript/confirm.js
+++ b/ckan/public-midnight-blue/base/javascript/confirm.js
@@ -1,0 +1,140 @@
+(function (ckan, $) {
+
+  /**
+   * Displays a Bootstrap 5 confirmation modal window
+   *
+   * Usage:
+   *
+   * ckan.confirm({
+   *    message: "Are you sure you want to delete this item?",
+   *    title: "Confirm Deletion",
+   *    confirmText: "Yes, Delete",
+   *    cancelText: "Cancel",
+   *    icon: "<i class='fa fa-trash me-2'></i>",
+   *    type: "danger",
+   *    onConfirm: function () {
+   *      console.log("User confirmed.");
+   *    },
+   *    onCancel: function () {
+   *      console.log("User cancelled.");
+   *    }
+   * });
+   *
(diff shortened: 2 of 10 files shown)


amercader:
The reason this function was left in place was that if some image (group/org or user) was uploaded using `munge_filename_legacy` and we switch to `munge_filename` and there is a difference between the two outputs, then the existing images might be broken. For most file names this won't be an issue but in some edge cases the output is different, e.g.

| input | `munge_filename()` | `munge_filename_legacy()` |
| --- | --- |--- |
| some ( org ) image.png | some-org-image.png | some--org--image.png |

Looking at the code, the part where there is an issue with backwards compatibility is in [model_dictize](https://github.com/ckan/ckan/blob/f2e70acc1ab2f5f80e7cfa4ce0382d0e35a2b2ef/ckan/lib/dictization/model_dictize.py#L460) (in lib/uploader.py the function can be safely replaced as it only affects new files).

In the model_dicitze functions, the munge function will process either:
1. A new `image_url` value provided via the API, this should not be an issue if we replace the function as it is a new value
2. An existing ìmage_url` value. For uploads, this is the result of calling the legacy munge function in lib/uploader. Here's where we can find an issue. For instance, if the original file name was "some ( org ) image.png", the `munge_filename_legacy()" function would have stored the file and `image_url` as `some--org--image.png`. After replacing the function, calling the dictazation functions would return `some-org-image.png` and the image URLs would be broken.

I think is better to err in the side of caution and provide some kind of migration mechanism. Any mechanism (CLI command, db migration...) will have to:

1. Check all `image_url` values in the `group` and `user` tables which are not actual URLs to see if the output of `munge_filename()` and `munge_filename_legacy()` are different
2. If so, check if the actual uploaded file still exists and if so, rename it to the output of `munge_filename()`
3. Update the actual `image_url` value to the output of `munge_filename()`

Aside: we discussed in the past using group or user ids as file names to avoid issues like this, I don't remember the details but I think the way we handle uploads made it difficult to provide an id beforehand

PR #9057, opened 2025-08-04
fix: pass proper context to tabledesigner _create_table_and_view

mutantsan:
`_create_table_and_view` function was using an empty context for calling `resource_view_list` and `resource_view_create`. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/tabledesigner/datastore.py
+++ b/ckanext/tabledesigner/datastore.py
@@ -20,7 +20,9 @@ END;
 '''
 
 
-def create_table(resource_id: str, fields: List[dict[str, Any]], context: Context | None = None) -> None:
+def create_table(
+    resource_id: str, fields: List[dict[str, Any]], context: Context | None = None
+) -> None:
     '''
     Set up datastore table + validation
     '''


ckanext/tabledesigner/datastore.py
```diff
@@ -50,9 +51,7 @@ def create_table(resource_id: str, fields: List[dict[str, Any]]):
             validate_rules=''.join(validate_rules),
         )
         get_action('datastore_function_create')(
-            {
-                'ignore_auth': True,
-            },
```
wardi, line 55: This one should be fine, no? `create_table` is a function for internal use, not tied to a user or session.

ckanext/tabledesigner/datastore.py
```diff
@@ -62,7 +61,7 @@ def create_table(resource_id: str, fields: List[dict[str, Any]]):
         )
 
     get_action('datastore_create')(
-        {},
```
wardi, line 65: Let's use `fresh_context` for these to make sure unrelated context variables are not passed and that no context modifications leak to the callers.

mutantsan:
Updated

changes/9057.bugfix
```diff
@@ -0,0 +1 @@
+Pass proper context to tabledesigner `_create_table_and_view`
```
wardi: please indicate what kind of problem this solves in the changelog

ckanext/tabledesigner/datastore.py
```diff
@@ -19,7 +20,7 @@
 '''
 
 
-def create_table(resource_id: str, fields: List[dict[str, Any]]):
```
wardi, line 22: I would like to backport this fix so please add the new parameter to the end of the list and set a default (None should work in this case) to prevent breaking some other code that might be using this function.

ckanext/tabledesigner/logic/action.py
```diff
@@ -25,23 +30,26 @@ def resource_update(
         context: Context,
         data_dict: dict[str, Any]) -> dict[str, Any]:
     res = original_action(context, data_dict)
-    _create_table_and_view(res)
+    _create_table_and_view(context, res)
     return res
 
 
-def _create_table_and_view(res: dict[str, Any]) -> None:
```
wardi, line 32: this function is named in a way that indicates it is `_private` so context at the beginning is fine here.

mutantsan:
@wardi hey, can you merge it it’s okay?

wardi (changes requested):
@mutantsan please resolve the CI error and this is good to go. Thanks!

PR #9059, opened 2025-08-06
Datastore dump interface

avdata99:
In progress

### Proposed fixes:

Allow extensions adding datastore dump formats

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/DUMP_FORMATS.md
+++ b/ckanext/datastore/DUMP_FORMATS.md
@@ -1,0 +1,138 @@
+# Extensible Datastore Dump Formats
+
+This document describes the new extensible dump format system for CKAN's datastore plugin.
+
+## Overview
+
+The datastore plugin now supports configurable and extensible dump formats through the `IDatastoreDump` interface. This allows:
+
+1. **Configurable formats**: Dump formats are no longer hardcoded
+2. **Overrideable defaults**: Default formats (CSV, TSV, JSON, XML) can be overridden by plugins
+3. **Extensible formats**: External plugins can add new dump formats (e.g., XLSX, ODS, etc.)
+
+## Architecture
+
+### IDatastoreDump Interface
+
+The new `IDatastoreDump` interface allows plugins to register custom dump formats:
+
+```python
+from ckanext.datastore.interfaces import IDatastoreDump
+
+class MyPlugin(p.SingletonPlugin):
+    p.implements(IDatastoreDump)
+
+    def register_dump_formats(self) -> dict[str, dict[str, Any]]:
+        return {
+            'xlsx': {
+                'writer_factory': xlsx_writer,
+                'records_format': 'objects',
+                'content_type': b'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+                'file_extension': 'xlsx'
+            }
+        }
+```
+
+### Format Configuration
+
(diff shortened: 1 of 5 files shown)


avdata99:
Discussion started about AI generated code here https://github.com/ckan/ckan/discussions/9060

EricSoroos:
Looking at using this to additionally decorate some information about the JSON dump, and I think it would be useful if there was an option to pass additional information into the writer. In my case, I need something out of the resource dictionary, so passing a resource id is enough for my purposes.

I'm using a patch like this, just putting **kwargs on the existing writers. An alternate version would be putting this in a data_dict type of parameter, but they're both just two ways of piping arbitrary non-schema data through the interface.

```
2 files changed, 5 insertions(+), 5 deletions(-)
ckanext/datastore/blueprint.py | 2 +-
ckanext/datastore/writer.py    | 8 ++++----

modified   ckanext/datastore/blueprint.py
@@ -287,7 +287,7 @@ def dump_to(
     bom = options.get('bom', False)
 
     def start_stream_writer(fields: list[dict[str, Any]]):
-        return writer_factory(fields, bom=bom)
+        return writer_factory(fields, bom=bom, resource_id=resource_id)
 
     def stream_result_page(offs: int, lim: Union[None, int]):
         return get_action('datastore_search')(
modified   ckanext/datastore/writer.py
@@ -18,7 +18,7 @@ BOM = "\N{bom}"
 
 
 @contextmanager
-def csv_writer(fields: list[dict[str, Any]], bom: bool = False):
+def csv_writer(fields: list[dict[str, Any]], bom: bool = False, **kwargs):
     '''Context manager for writing UTF-8 CSV data to file
 
     :param fields: list of datastore fields
@@ -35,7 +35,7 @@ def csv_writer(fields: list[dict[str, Any]], bom: bool = False):
 
 
 @contextmanager
-def tsv_writer(fields: list[dict[str, Any]], bom: bool = False):
+def tsv_writer(fields: list[dict[str, Any]], bom: bool = False, **kwargs):
     '''Context manager for writing UTF-8 TSV data to file
 
     :param fields: list of datastore fields
@@ -71,7 +71,7 @@ class TextWriter(object):
 
 
 @contextmanager
-def json_writer(fields: list[dict[str, Any]], bom: bool = False):
+def json_writer(fields: list[dict[str, Any]], bom: bool = False, **kwargs):
     '''Context manager for writing UTF-8 JSON data to file
 
     :param fields: list of datastore fields
@@ -115,7 +115,7 @@ class JSONWriter(object):
 
 
 @contextmanager
-def xml_writer(fields: list[dict[str, Any]], bom: bool = False):
+def xml_writer(fields: list[dict[str, Any]], bom: bool = False, **kwargs):
     '''Context manager for writing UTF-8 XML data to file
 
     :param fields: list of datastore fields

```

avdata99:
@EricSoroos, the customer who requested this feature has removed its priority.
I cannot continue with it, so if you find it useful, feel free to continue.
Note: This change request was initiated by an AI, and I have not finished reviewing it completely.

EricSoroos:
I'm happy enough to push this -- I've reviewed all except for the example XLS writer plugin. The comment above is mainly extending this to fit the use case I have, and if we're adding an interface we may as well make it as future proof as possible.

PR #9065, opened 2025-08-21
Add Provenance section to PR template

wardi:
Suggested text for discussion.

Also included text specifically stating that the changes can be released under AGPL, not sure if this is needed or just noise.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,15 @@ Fixes #
 
 ### Proposed fixes:
 
+-
 
+### Provenance:
+
+A majority of these changes were
+- [ ] created using automatic code or documentation generation tools
+- [ ] created myself
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,7 +2,15 @@ Fixes #
...
 
+-
 
+### Provenance:
+
+A majority of these changes were
+- [ ] created using automatic code or documentation generation tools
+- [ ] created myself
+
+These changes may be released under [CKAN's AGPL license](https://github.com/ckan/ckan/blob/master/LICENSE.txt)
```
wardi, line 13: suggestion from @EricSoroos 
```suggestion
These changes will be released under [CKAN's AGPL license](https://github.com/ckan/ckan/blob/master/LICENSE.txt)
```

amercader:
Django's similar PR: https://github.com/django/django/pull/19594

They are much more explicit in requesting disclosure, I don't think we need that level of detail as things stand now

PR #9068, opened 2025-08-23
Control configuration options shared with flask

smotornyuk:
Currently, Flask has access to [all](https://github.com/ckan/ckan/blob/4b509103203563ed310273ade03befc68a203137/ckan/config/middleware/flask_app.py#L164) CKAN's config options. It doesn't hurt, but there are chances to unintentionally introduce a custom CKAN option that matches Flask's option and causes unexpected behavior.

I suggest using the `flask: true` flag in config declarations, marking options that must be shared with Flask. Any other option, without this flag, will remain inside CKAN, and Flask won't see it, even if it has an option with the same name.

This change also means that we can add some sort of `flask_rename` parameter that will rename options before sharing them with Flask. For example, instead of using Flask's uppercase-style `REMEMBER_COOKIE_SAMESITE`, we can start using more CKANish style `remember_cookie.samesite`. Note, this is an idea - there are no such features in the current PR.

Small additional changes slightly related to the current PR:
* Drop authentication by `REMOTE_USER` in tests, which was deprecated 3 years ago. During preparation for v2.11 extensions, already migrated to authentication via the `Authorization` header
* Drop `conf` parameter to `make_flask_stack` and predictably use common config object.


ckan/config/config_declaration.yaml
```diff
@@ -47,74 +44,109 @@ groups:
...
+        flask: true
         ignored: true
       - key: BABEL_DOMAIN
+        flask: true
         ignored: true
       - key: BABEL_TRANSLATION_DIRECTORIES
+        flask: true
         ignored: true
       - key: CKAN_INI
+        flask: true
```
wardi: flask needs this one?

ckan/config/middleware/flask_app.py
```diff
@@ -125,15 +125,15 @@ def __call__(self, environ: Any, start_response: Any):
         return self.app(environ, start_response)
 
 
-def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
+def make_flask_stack() -> CKANApp:
     """ This has to pass the flask app through all the same middleware that
     Pylons used """
```
wardi: update docstring while we're here?

ckan/config/middleware/flask_app.py
```diff
@@ -160,10 +160,11 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
...
     # Update Flask config with the CKAN values. We use the common config
     # object as values might have been modified on `load_environment`
-    if config:
-        app.config.update(config)
-    else:
-        app.config.update(conf)
+    app.config.update({
+        k: config[k]
+        for key in config_declaration.iter_options()
+        if config_declaration[key].has_flag(Flag.flask) and (k := str(key))
```
wardi: Trying to understand `and (k := str(key)` here because it looks funny.

Can a config key ever not be a string? Can it ever be `""`?

changes/9068.removal
```diff
@@ -0,0 +1,2 @@
+Tests do not support authentication by ``REMOTE_USER`` environment
+variable. Use ``Authorization`` header with user's API Token instead.
```
wardi:
```suggestion
Tests no longer support authentication by ``REMOTE_USER`` environment
variable. Use an ``Authorization`` header with user's API Token instead.
```
and maybe include a link to an example

ckan/config/middleware/flask_app.py
smotornyuk: Here I'm iterating over `config_declaartions`, where `key` has type `ckan.config.declaration.Key`(which is similar to string with few additions). I need to convert `Key` into `str` and I have three options:

1. the one I used
2. `str(key): config[str(key)]` which is cleaner but converts twice
3. rewrite it completely to few expressions:
```py
patch = {}

for key in config_declaration.iter_options():
    if config_declaration[key].has_flag(Flag.flask):
       option_name = str(key)
       patch[option_name] = config[option_name]

app.config.update(patch)
```

ckan/config/config_declaration.yaml
smotornyuk: I did mass edit here and did not notice this option in a wrong place. I'm moving it to a different block with CKAN internal options. It does not change anything, just organizing things

ckan/config/middleware/flask_app.py
smotornyuk: I'll switch it to double-conversion - it's just takes few nanoseconds but makes things much more readable

PR #9070, opened 2025-08-26
Fix select2 autocomplete

amercader:
After the upgrade to Select2 4.0 (https://github.com/ckan/ckan/pull/8930, https://github.com/ckan/ckan/pull/8979) all select2 fields that were using autocomplete were broken. The logic to define the source endpoint and how the results are processed changed in Select2 4.0:

https://select2.org/upgrading/migrating-from-35#custom-data-adapters-instead-of-query

`createSearchChoice` has been renamed to `createTag`:

https://select2.org/upgrading/migrating-from-35#renamed-createsearchchoice

Also the methods for rendering the results have changed significantly.

TODO:

- [x] Minimum input length not working
- [x] Tags option not working
- [x] Review other module options
- [x] Autocomplete not working on midnight blue theme
- [x] Tests failing


mutantsan:
@amercader should it be a part of the `public-midnight-blue` as well?

amercader:
@mutantsan , yes, wherever select2 is used

smotornyuk:
Visually, this fix works on the classic theme, so I applied identical changes to the midnight-blue. I tested it on new-member page and it did suggested me options

Zharktas:
@smotornyuk this is still totally broken with tags, I'm looking this today what still needs to be changed.

Zharktas:
This should be now finished. I'm not entirely happy about tests and stubbing select2 from, but it seemed difficult to make them pass without it.

ckan/public/base/javascript/modules/autocomplete.js
amercader, line 119: @Zharktas one minor thing, you can not add a tag by pressing the `,` key like in 2.11. I *think* it is because `select2.search` which referenced the input no longer exists. Do you know if we can replace it with something else?

Zharktas: I'll take look today, select2 should still support it.

Zharktas: @amercader should work now, select2 is being silly since various options do not apply when using custom data adapter https://github.com/select2/select2/blob/595494a72fee67b0a61c64701cbb72e3121f97b9/src/js/select2/defaults.js#L63-L103

Also restored containerClass attribute since I noticed that its just under different name in the new select2.

amercader: Thanks @Zharktas !

PR #9075, opened 2025-08-27
#9069: Fix ITemplateHelpers plugin order

### LINKED ISSUE #9069: ITemplateHelpers are handled in the wrong order
## CKAN version
2.10, 2.11, probaly master as well.

## Describe the bug

When multiple plugins have a template helper with the same name, they are handled in reversed order compared to templates itself.

This was changed in https://github.com/ckan/ckan/pull/5337 and is seemingly undocumented change. 

### Steps to reproduce
I made example extensions in https://github.com/Zharktas/plugin_order, both extensions have `base.html` which calls `h.helper()`

The rendered template displays the following when plugin order is `first second`:

Image

The following is rendered when plugins are configured in order `second first`:

Image

### Expected behavior
Templates and template helpers should be handles in the same order.

Zharktas:
Fixes #9069 

### Proposed fixes:

Adds suggested fix from the issue and tests.

### Features:

- [ ] includes API changes


changes/9069.bugfix
```diff
@@ -0,0 +1 @@
+Reversed handling of plugin order for ITemplateHelpers.
```
wardi:
```suggestion
Restore handling of plugin order for ITemplateHelpers to align with
add_template_directory precedence again (first plugin wins).
```

wardi (changes requested):
Just a minor suggestion for the changelog, otherwise looks great

wardi:
the commit comments reference the wrong issue, git rebase -i to correct them?

Zharktas:
@wardi good catch, with slight dyslexia I thought i checked them all, but now they are correct.

Zharktas:
Noticed that 2.10, does not have the attribute, it was added in 2.11. So 2.10 backport requires restoring the reversion on the plugin loop itself.

wardi:
Thanks @Zharktas would you send a separate PR for 2.10?

PR #9085, opened 2025-09-01
[Midnight blue] Footer

aleeexgreeen:
Fixes #

### Proposed fixes:
- Attribution and language dropdown to black section of footer
- Changes colour of text in "Open Data" button to black, for accessibility.
- Customize the language dropdown container
- Make "Language" text sr-only
- Fixes bug where changing language causes 500 error because of `nav_link` macro. 

### Screenshots/Figma:
[Figma link](https://www.figma.com/design/oI3wDSdZ0cHZLh6CrfYMnb/CKAN-3.0-Design-System?node-id=1356-6511&t=dNVF1eYeNjIKmWx1-4)
 
image
image

### Features:

- [ ] includes API changes


amercader:
Unfortunately the locale issue is not as simple as that as CKAN can be configured to be served from a path like /data/{{Lang}}/ckan . Leave that with me and I'll send a separate PR that implements proper `remove_locale()` and `remove_root_path()` functions ( we need to go from `/ca/dataset` or `data/ca/dataset` to `/dataset` to be able to get the `dataset.search` endpoint)

Besides tha,t the only comment I have is that the blue section of the footer feels a bit unbalanced with all the links crammed to one side and the other empty. Maybe we can increase the spacing between the links a bit. I guess each site can decide what to put in the right hand side but if you can think of something generic we could put it there as a placeholder.

aleeexgreeen:
@amercader okay sounds good! I have removed the changelog entry and the code added to helpers.py. I've also added a description to the footer, let me know if the content should be changed at all. I've updated the screenshots in the PR.

PR #9093, opened 2025-09-08
Add helpers for removing locale and root path,  fix page_is_active

amercader:
The `page_is_active()` helper is used to determine if a provided URL matches the page the user is currently in (i.e the requested endpoint) This is used in the header menu to highlight e.g. the "Datasets" menu item when the user is in `/dataset` or in `/dataset/some-dataset`.

The current implementation (introduced as part of the new theme, so 2.12 only) failed both when the user was requesting a URL with a locale (e.g. `/ca/dataset`) or with a root path (e.g. `/my/site/dataset`).

Added two new helpers that dealt exclusively with removing these bits from URLs, currently only used by `endpoint_from_url()` but that might be useful in other contexts. Includes numerous tests.

### Features:

### DIFF
--- a/ckan/lib/search/__init__.py
+++ b/ckan/lib/search/__init__.py
@@ -4,8 +4,6 @@ from __future__ import annotations
 
 import logging
 import sys
-import cgitb
-import warnings
 import traceback
 
 import xml.dom.minidom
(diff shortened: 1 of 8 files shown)


ckan/lib/helpers.py
```diff
@@ -748,6 +752,67 @@ def endpoint_from_url(url: str) -> str:
     return endpoint
 
 
+@core_helper
+def remove_root_path_from_url(url: str) -> str:
+    """
+    Remove the root path (i.e. config["ckan.root_path"]) from a url.
+
+    Note: the
```
Zharktas: This comment might have some unwritten point.

amercader: Done

PR #9096, opened 2025-09-15
Notify API user that they are using old removed api keys

Zharktas:
### Proposed fixes:

When older CKAN instance is upgraded and users have had API keys, they might still try to use them for API authorization if they haven't been notified that they should be using API tokens instead. Only mention that this might be happening is in the logs where there is a message `Cannot decode JWT token: Not enough segments`

This PR adds a check to the API view that if Authorization header contains uuid4 key, the API would return bad request and tell the user to use API token instead.

Currently checks only Authorization header, but perhaps it should check headers based on config since the api key itself might have been configured to other headers as well.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/views/api.py
+++ b/ckan/views/api.py
@@ -236,9 +236,12 @@ def action(logic_function: str, ver: int = API_DEFAULT_VERSION) -> Response:
     authorization_header = request.headers.get("Authorization")
 
     if authorization_header:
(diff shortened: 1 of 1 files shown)


Zharktas:
This does not seem to work with https://github.com/ckan/ckan/pull/8636, as api tokens are handled on flask stack and failing token just produces a csrf error mentioned in the comments.

PR #9132, opened 2025-10-27
Disable retries on any failed test

### LINKED ISSUE #9039: tests are retried even if not marked flaky
## CKAN version
master

## Describe the bug
All failing tests are retried 3 times but only those tests marked flaky should be retried, e.g.
https://github.com/ckan/ckan/actions/runs/16506730732/job/46679159824?pr=8684

### Steps to reproduce
run a broken test not marked flaky

### Expected behavior
test fails once

### Additional details
I looked for an option to change this behavior but couldn't find one. Should we drop this pytest extension?

wardi:
Fixes #9039

### Proposed fixes:

retry/reruntestfailures don't seem to be configurable to not apply to every test.

Let's remove these pytest extensions and fix flaky tests instead for the sake of my sanity.

### Features:

- [ ] includes API changes


ckanext/activity/tests/test_views.py
```diff
@@ -747,7 +748,6 @@ def test_read_dataset_as_it_used_to_be_after_deleting_resource(self, app):
         assert helpers.body_contains(response, "Dataset title")
         assert helpers.body_contains(response, resource["name"])
 
-    @pytest.mark.flaky(retries=3, delay=1)
```
EricSoroos, line 750: Does it make sense to xfail(flakey) instead of internally rerunning?

wardi: IIUC these tests aren't expected to fail they just might take some time to settle to the expected response. So what I've done here is put the sleep() inside the test just at the stage of checking the result instead of in between reruns of the full test like the retry was doing.

ckan/tests/pytest_ckan/test_fixtures.py
```diff
@@ -65,8 +65,13 @@ def test_fake_plugin_does_not_exist(self,):
...
         """Fake plugin can be loaded when provided via mark."""
-        plugin = plugins.load("fake_plugin")
-        assert isinstance(plugin, _FakePlugin)
+        try:
+            plugin = plugins.load("fake_plugin")
+            assert isinstance(plugin, _FakePlugin)
+        finally:
+            # manually loaded plugins need to be cleaned up
+            # or they will affect other tests
+            plugins.unload_all()
```
wardi, line 74: automatic retries have been hiding this broken test for a while now

wardi:
conflict resolved

PR #9026, opened 2025-07-18
Files: phase 3

smotornyuk:
Phase 3 of #8920

This PR consolidates DB models and API actions into 4,000 lines of code.

## Improvements of features introduced in phase 2

* config option to switch between the original unchanged Uploader, which does not use files at all, and the new Uploader that relies on configurable storage. This option is for extensions that extend and heavily override the original uploader and can break after even the slightest change in its logic.

## API 

I want to try a different layout of logic separation inside `ckan.logic`. As new logic belongs to the "file" domain, instead of separating actions between `get`, `delete`, and `update`, I created new files:

* `logic/action/file.py`: file actions
* `logic/auth/file.py`: file auth functions
* `logic/schema/file.py`: file schemas

I added just the most generic actions, like create, delete, etc. Other operations, like resumable uploads and the ability to override a file, may sound tempting, but I see a lot of problems and cannot make a good generic solution for them. For now, I recommend adding custom actions over the existing generic functionality.

And there is a "hidden" `file_search` action. It can be called normally, I just excluded it from documentation, that's why it's "hidden". I'm not sure if searching for files should be available via API(for now, I locked it behind sysadmin permission). And as @amercader is working on new search specs, the syntax of search parameters is also unstable. Even though I'm not going to index the file via a search engine, using similar search parameters is natural. I diverged slightly from the current specification, and here are  the rules:

Soring is controlled by `sort`. It can be either string `sort: "name"`, which means sort by a single column ASC. Alternatively, it can be a list of strings, `sort: ["name", "size"]` to sort by multiple columns ASC simultaneously. Finally, it can be listed with nested 2-element lists, `sort: [["name", "asc"], ["size", "desc"]]`, to control direction. The second and third styles can be mixed.  There is no string processing and `"name desc"` won't be split into column and direction - you must provide list with two elements if you need more control.

Filters are similar to specification from `ckanext-search`, but I decided to use `$` prefix for every operator, not only for `and` / `or`. In addition, `$and` / `$or`, must be specified on the top level, which is similar to the way one builds queries in SQLAlchemy. For example, if you want to find a file with the name `hello` or `world` (and you do not want to use `$in`), you must use the first payload, not the second

```py
# correct
filters = {
  "$or": [
    {"name": "hello"},
    {"name": "world"},
  ]
}

# INCORRECT
filters = {
  "name": {
    "$or": ["hello", "world"] # `$or` is not allowed as field-level operator
  }
}

```

[...]

PR #9138, opened 2025-10-30
add draft API action to chain API calls together, discussion #9112

ThrawnCA:
As per discussion #9112 this adds a new API allowing the user to submit a batch of (potentially interdependent) API calls, with one request. Each call is individually access-checked, and the results of one call may optionally be saved and used in future calls.

The envisioned use case is for setting up integration test data, but it also has potential applications for making bulk processing more efficient in general.

Needs updated docs, but the code as written does appear to work, so it may provide at least a starting point.

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -1460,11 +1460,11 @@ def bulk_api_call(context: Context, data_dict: DataDict):
             for name, path in save_results.items():
                 value_to_save = result
                 for path_element in path.split('.'):
-                    if isinstance(value_to_save, dict):
-                        value_to_save = value_to_save.get(path_element)
-                    else:
+                    if not isinstance(value_to_save, dict) \
+                            or path_element not in value_to_save:
                         raise ValidationError(
                             {'save_results': 'Path not found'}
                         )
+                    value_to_save = value_to_save.get(path_element)
                 saved_results[name] = value_to_save
     return saved_results
--- a/ckan/tests/logic/action/test_update.py
+++ b/ckan/tests/logic/action/test_update.py
@@ -7,14 +7,11 @@ import unittest.mock as mock
 import pytest
 import sqlalchemy as sa
 
-import ckan.lib.app_globals as app_globals
-import ckan.logic as logic
-from ckan.logic.action.get import package_show as core_package_show
-import ckan.plugins as p
-import ckan.tests.factories as factories
-import ckan.tests.helpers as helpers
-from ckan import model
+from ckan import logic, model, plugins as p
+from ckan.lib import app_globals
 from ckan.lib.navl.dictization_functions import DataError
+from ckan.logic.action.get import package_show as core_package_show
+from ckan.tests import factories, helpers
 from freezegun import freeze_time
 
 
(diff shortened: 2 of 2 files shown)


ThrawnCA:
`update.py` is perhaps not the best home as it can delegate to any type of call, but I was hesitant to make a whole new module.

wardi:
> `update.py` is perhaps not the best home as it can delegate to any type of call, but I was hesitant to make a whole new module.

This kind of thing *would* fit better in its own module since it's orthogonal to the other types of actions. There's already precedent (sort of) with `patch.py` being added for patch actions and the upcoming `file.py` module https://github.com/ckan/ckan/pull/9026/files#diff-2bf122bc390382357ed059c417a1f6d9ba71cfdb6793de881aa8d5c35193108f for file operations

Dividing actions by their CRUD classification means common logic gets spread across different modules and mixed with unrelated code. Changes in `update.py` almost always also come with changes to `create.py` and these 4 modules grow over time. The magic side effect of applying `side_effect_free` decorator for actions in `get.py` isn't a very strong argument for keeping this pattern.

ckan/logic/auth/update.py
```diff
@@ -285,3 +285,10 @@ def config_option_update(context: Context, data_dict: DataDict) -> AuthResult:
        Only sysadmins can do it
     '''
     return {'success': False}
+
+
+def bulk_api_call(context: Context, data_dict: DataDict) -> AuthResult:
+    '''
+    Anyone can call this API, each delegated call is checked separately.
+    '''
+    return {'success': True}
```
wardi, line 294: `bulk_api_call` as is would be an attractive DoS target.

Limiting the number of calls (with config options) that can be sent in a single request might resolve this by allowing for throttling of abusive requests e.g. by IP address in nginx or a reverse proxy. We could have one call limit for unauthenticated users, another for authenticated users, WDYT?

ckan/logic/action/update.py
```diff
@@ -1355,3 +1355,116 @@ def config_option_update(
...
+
+    :return A dictionary of the saved return values, if any.
+    '''
+    _check_access('bulk_api_call', context, data_dict)
+    saved_results: dict[str, Any] = {}
+    calls = cast("list[dict[str, Any]]", data_dict.get('calls', []))
+    for call in calls:
+        action_name: str = call.get('action', '')
+        if not action_name:
+            raise ValidationError({'action': _('Missing value')})
```
wardi, line 1445: It would be better to raise ValidationErrors that we can detect early, before executing any of the calls.

When there are errors that occur in the calls we should capture them and report back to the user how many calls actually succeeded before the error, possibly including the intermediate results.

ckan/logic/auth/update.py
ThrawnCA: Fair call, some type of throttling would be valuable.

ckan/logic/action/update.py
ThrawnCA: Fail-fast is a good idea.

Can we include a ValidationError _and_ a list of outcomes, in the same response?

wardi: This API is breaking new ground so we'll need to decide if we return `"success": true`, `"success": false` or something else entirely.

I could see an argument for always returning `"success": true` then including a report on which actions succeeded/failed and results that need to be parsed by the caller.

`"success": true` would make sense especially if the API had a way to schedule the work in a background job as an option (the scheduling would be expected to always succeed).

Another idea is not using the action API for bulk actions. We could instead use a streaming response from the server to send progress/interim results/failures for each action call and not be tied to a single request/response.

Or maybe we don't need this to be a core feature at first? Any of these options could be explored as ckan extensions and we can later adopt the one that has the most support.

PR #9156, opened 2025-11-12
[#9155] Enable using overridden helpers in model_dictize.py

### LINKED ISSUE #9155: Helpers used in `model_dictize.py` cannot be overridden
## CKAN version
2.11.4

## Describe the bug
If a plugin implementing `ITemplateHelpers` overrides a core helper function, and that helper is called within [‎ckan/lib/dictization/model_dictize.py](https://github.com/ckan/ckan/commit/85bd352e17da79af5f9b5e23f93162f62a511754#diff-479ec62a04a4a945f75aa6d8644b5b609ddc678b4ee8f56328b50b9d748420b9), only the original helper is called, not the overridden one.

### Steps to reproduce
1. In a plugin, override the helper `strxfrm`. This helper is used as a sort key to sort the list of groups when dictizing a package: https://github.com/ckan/ckan/commit/85bd352e17da79af5f9b5e23f93162f62a511754. Make sure that the plugin implements `ITemplateHelpers` and includes the helper in its `get_helpers` method.

```python
# plugin.py

@tk.chained_helper
def strxfrm(next_helper, s):
    """
    Reverse the string and pass it on to the built-in strxfrm helper.
    """
    s = s[::-1]

    return next_helper(s)


class MyPlugin(plugins.SingletonPlugin):
    plugins.implements(plugins.ITemplateHelpers)

    def get_helpers(self):
        return {
            "strxfrm": strxfrm,
        }
```

2. Create a dataset and add it to several groups.
3. While logged in as a sysadmin, show the dataset in the API: http://example.org/api/3/action/package_show?id=my-dataset&include_plugin_data=true. (`include_plugin_data=true` stops the `package_show` action using the cached validated data dict of the dataset, ensuring that it dictizes the package here.)
4. See that the package's groups are sorted normally by their `display_name`, not using the custom helper.

### Expected behavior
I would expect the overridden helper in my plugin to be used as the sort key for groups in the group list.

Overriding a plugin like this works in templates and controllers, and it's a stated expectation in the CKAN code that it will work in `model_dictize.py`:

```python
@core_helper
def strxfrm(s: str) -> str:
    '''
    Transform a string to one that can be used in locale-aware comparisons.
    Override this helper if you have different text sorting needs.
    '''
    return unicodedata.normalize('NFD', s).lower()
```

bellisk:
Otherwise, if a helper function that is used here is overridden by an extension, the core helper will still be used and the overridden version ignored.

Fixes #9155

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -1781,11 +1781,6 @@ def dump_json(obj: Any, **kw: Any) -> str:
 
 
 @core_helper
-def load_json(obj: str, **kw: Any) -> dict[Any, Any]:
-    return json.loads(obj, **kw)
-
-
-@core_helper
 def snippet(template_name: str, **kw: Any) -> str:
     '''
     Use {% snippet %} tag instead for better performance.


wardi:
@bellisk I'm interested to know the use case for overriding json decoding from an extension, this kind of a change could have surprising side effects.

bellisk:
@wardi oh, I totally see how this PR would give that impression, but that's not actually something we're using or want to do!

When I made the first commit here, several tests started failing because `model_dictize` is calling `h.json.loads` and that didn't work any more, because `h.json` doesn't get added to the `helper_functions`.

I tried adding `core_helper(json)` to `helpers.py`, but that didn't work either, because `json` here is ultimately the `simplejson` module and `core_helper()` only works on classes or functions. Looking a bit further, I saw that there's already the `dump_json` core helper and added a `load_json` as an analogue to it.

If there's a better way of handling this, I'm happy to do that instead. :)

bellisk:
By the way, the only failing check on this PR is about the code coverage for the patch itself, not for the codebase as a whole. At the risk of being a bit cheeky, how important is that really? I can add a test for the new helper, but I really don't have time to do it at the moment, so it might have to wait for a while.

ckan/lib/dictization/model_dictize.py
```diff
@@ -29,14 +29,14 @@
...
 import ckan.logic as logic
 import ckan.plugins as plugins
-import ckan.lib.helpers as h
 import ckan.lib.dictization as d
 import ckan.authz as authz
 import ckan.lib.search as search
 import ckan.lib.munge as munge
 import ckan.model as model
 from ckan.types import Context
 from ckan.common import config
```
wardi:
```suggestion
from ckan.common import config, json
```

ckan/lib/dictization/model_dictize.py
```diff
@@ -516,7 +516,7 @@ def tag_dictize(tag: model.Tag, context: Context,
         q: dict[str, Any] = {
             'q': tag_query, 'fl': 'data_dict', 'wt': 'json', 'rows': 1000}
 
-        package_dicts = [h.json.loads(result['data_dict'])
+        package_dicts = [h.load_json(result['data_dict'])
```
wardi:
```suggestion
        package_dicts = [json.loads(result['data_dict'])
```

ckan/lib/helpers.py
```diff
@@ -1780,6 +1780,11 @@ def dump_json(obj: Any, **kw: Any) -> str:
     return json.dumps(obj, **kw)
 
 
+@core_helper
+def load_json(obj: str, **kw: Any) -> dict[Any, Any]:
+    return json.loads(obj, **kw)
+
+
```
wardi:
```suggestion
```

wardi:
@bellisk the coverage failure can be ignored.

I've suggested some smaller changes instead of creating a new load_json helper that's only used in one place. This would bring the change more in line with the rest of the code base

bellisk:
@wardi thanks for the suggestions! I've accepted them :)

PR #9166, opened 2025-11-18
[L10n] use full sentence for split messages, v2.10-only changes

### LINKED ISSUE #9149: L10n: Strings should consist of full sentence instead of breaking into multiple strings
## CKAN version
master / 2.10 / 2.11

## Describe the bug
In the Activity extension we have messages[1] broken into multiple strings for localization, however the broken strings are not syntax-valid in many languages. 

We should have the complete line of message concatenated into one string, and have variable/placeholders injected into it, so localizers may translate in the grammar of the language transalted into.  [2] is an example of a good string.

Any templates consists of a standalone 'from' or 'to' message, might be affected, and is not localizable into zh-TW.

### Additional details
[1] https://github.com/ckan/ckan/blob/fd5068dd76df2c0ab57655ccd83a046c0690964c/ckanext/activity/templates/snippets/changes/author_email.html#L6
Or refer to the zh-TW CKAN 2.10/2.11 strings missing translation on transifex.

[2]
https://github.com/ckan/ckan/blob/fd5068dd76df2c0ab57655ccd83a046c0690964c/ckanext/activity/templates/snippets/changes/author.html#L5

petercpg:
Fixes #9149 

### Proposed fixes:
* Concatenated split strings, placeholders, punctuation of a message into one string
* i18n POT file NOT updated to avoid conflict

This MR is opened toward dev-v2.10 branch as the template file ckan/templates/home/snippets/stats.html was already removed in 2.11 and master.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/templates/home/snippets/stats.html
+++ b/ckan/templates/home/snippets/stats.html
@@ -2,25 +2,40 @@
 
 <div class="module-stats">
   <div class="card box">
-    <h3>{{ _('{0} statistics').format(g.site_title) }}</h3>
+    {# TRANSLATORS: site_title is the name of the CKAN site #}
+    <h3>{{ _('{site_title} statistics').format(site_title=g.site_title) }}</h3>
     <ul>
       {% block stats_group %}
+      {% set dataset_num = h.SI_number_span(stats.dataset_count) %}
       <li>
         <a href="{{ h.url_for('dataset.search') }}">
-          <strong>{{ h.SI_number_span(stats.dataset_count) }}</strong>
-          {{ _('dataset') if stats.dataset_count == 1 else _('datasets') }}
+          {% trans count=stats.dataset_count, num=dataset_num %}
+            <strong>{{ num }}</strong> dataset
+          {% pluralize count %}
+            <strong>{{ num }}</strong> datasets
+          {% endtrans %}
         </a>
       </li>
+
+      {% set org_num = h.SI_number_span(stats.organization_count) %}
       <li>
         <a href="{{ h.url_for('organization.index') }}">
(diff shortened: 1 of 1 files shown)


ckan/templates/home/snippets/stats.html
```diff
@@ -2,25 +2,40 @@
...
         </a>
       </li>
+
+      {% set org_num = h.SI_number_span(stats.organization_count) %}
       <li>
         <a href="{{ h.url_for('organization.index') }}">
-          <strong>{{ h.SI_number_span(stats.organization_count) }}</strong>
-          {{ _('organization') if stats.organization_count == 1 else _('organizations') }}
+          {% trans count=stats.organization_count, num=org_num %}
+            <strong>{{ num }}</strong> organization
```
wardi, line 24: org_num?

wardi:
@amercader would we normally consider this kind of change for a 2.10 patch release?

ckan/templates/home/snippets/stats.html
petercpg: Since the amounts are injected when rendering, I think `{{ num }} organization` should be fine? But I can change these placeholders into `dataset_num`, `org_num`, and `grp_num` for more clarity :)

wardi:
I believe changes to translation strings are not typically backported to point releases to avoid breaking any production site's existing custom translations, and IIUC our workflow for requesting that translators update languages in transifex is tied to normal releases.

The changes here *are* fixing a real problem for sites looking to implement new translations, so I can appreciate the argument for applying them. Hopefully we can get them released in 2.12 soon.

PR #9178, opened 2025-11-28
Fix Unique User Email Index for Case

JVickery-TBS:
fix(db): user email unique index;

- Case insensitivity for user email index.

Fixes #

### Proposed fixes:

The unique index for user emails does not handle case sensitivity. This makes it so it does.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/migration/versions/108_f7b64c701a10_fix_conditional_index_on_user.py
+++ b/ckan/migration/versions/108_f7b64c701a10_fix_conditional_index_on_user.py
@@ -21,7 +21,7 @@ depends_on = None
 def upgrade():
     op.create_index(
         "idx_only_one_active_email_no_case", "user",
-        [sa.func.lower(sa.Column('email')), "state"],
+        [sa.func.lower(sa.Column('email'))],
         unique=True, postgresql_where=sa.text('"user".state=\'active\''))
     print('Created "idx_only_one_active_email_no_case" index')
     op.drop_index("idx_only_one_active_email")
--- a/ckan/model/user.py
+++ b/ckan/model/user.py
@@ -62,7 +62,7 @@ user_table = Table('user', meta.metadata,
         Index('idx_user_id', 'id'),
         Index('idx_user_name', 'name'),
         Index('idx_only_one_active_email_no_case', func.lower(Column('email')),
-              'state', unique=True, postgresql_where="(state = 'active'::text)"),
+              unique=True, postgresql_where="(state = 'active'::text)"),
         )
 
 


JVickery-TBS:
@wardi as discussed here is the upstream PR. there is already the `ckan db duplicate_emails` command so no work needed there?

ckan/migration/versions/108_f7b64c701a10_fix_conditional_index_on_user.py
```diff
@@ -0,0 +1,37 @@
...
+revision = 'f7b64c701a10'
+down_revision = '4eaa5fcf3092'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_index(
+        "idx_only_one_active_email_no_case", "user",
+        [sa.func.lower(sa.Column('email')), "state"],
```
wardi: does it work to create this index without storing the state?

wardi:
> @wardi as discussed here is the upstream PR. there is already the `ckan db duplicate_emails` command so no work needed there?

It should be updated to ignore case too though (and maybe let postgres do the case matching because technically lowercasing may be different based on the postgres locale settings)

JVickery-TBS:
the original index used state, I think the idea was to allow people to keep deleted user accounts in case they needed to extract info from the database? otherwise you would have to purge all duplicate users from the tables @wardi

JVickery-TBS:
@wardi I think the `duplicate_emails` command is already doing case, I think it does `.lower()` in the loop: https://github.com/ckan/ckan/blob/master/ckan/cli/db.py#L206

wardi:
Thanks, I was checking against an older version where it didn't

JVickery-TBS:
@wardi tests updated and change log added. Lemme know about the `user.state="active"` thing. maybe need to bring up in next core dev chat?

wardi (changes requested):
The index change should appear in the user model as well. We can create a db much faster from the models than from the migrations.

JVickery-TBS:
@wardi index has been updated to the newly named one with the func.lower in the user model now

wardi:
this reminds me that we're missing a test that creates a db with the migrations then checks that `ckan generate migration --autogenerate` doesn't generate anything

ckan/migration/versions/108_f7b64c701a10_fix_conditional_index_on_user.py
wardi: I mean to say: since this is a conditional index that only indexes records with state=active, and the index itself doesn't need to include state.

JVickery-TBS: aahhhhh okok that should be done now then

JVickery-TBS:
@wardi this should be good to go now hopefully

PR #9203, opened 2026-01-03
fix resource form publish error handling

### LINKED ISSUE #9202: Dataset publish: problems with state=active validation errors
## CKAN version
master

## Describe the bug
When dataset validation errors are triggered when publishing a dataset from the resource create page those errors are not shown and the resource form is incorrectly filled with data that has already been saved (likely resulting in a duplicate resource being created when submitted again)

### Steps to reproduce
Use a ckanext-scheming schema with [draft_fields_required](https://github.com/ckan/ckanext-scheming?tab=readme-ov-file#draft_fields_required) = false. Create a dataset and leave a required dataset field blank. Attempt to publish the dataset from the create resource step.

### Expected behavior
Error should be shown indicating what has prevented publishing the resource, and if a resource was created in the same step the resource form should now be blank.

wardi:
Fixes #9202

### Proposed fixes:

On validation errors in the resource form triggered by setting state=active
- pass all dataset errors to form (not only errors from the last resource)
- leave resource form blank since the resource was created successfully

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9202.bugfix
+++ b/changes/9202.bugfix
@@ -1,0 +1,2 @@
+display correct errors and leave resource form empty when dataset validation errors
+are triggered by publishing a dataset (setting state=active)
--- a/ckan/views/resource.py
+++ b/ckan/views/resource.py
@@ -228,10 +228,9 @@ class CreateView(MethodView):
                     _(u'The dataset {id} could not be found.').format(id=id)
                 )
             except ValidationError as e:
-                errors = cast(
-                    "list[ErrorDict]", e.error_dict.get('resources', [{}]))[-1]
+                errors = cast("list[ErrorDict]", e.error_dict)
                 error_summary = e.error_summary
-                return self.get(package_type, id, data, errors, error_summary)
+                return self.get(package_type, id, {}, errors, error_summary)
             return h.redirect_to(u'{}.read'.format(package_type), id=id)
 
         data[u'package_id'] = id
--- a/ckan/views/resource.py
+++ b/ckan/views/resource.py
@@ -228,9 +228,8 @@ class CreateView(MethodView):
                     _(u'The dataset {id} could not be found.').format(id=id)
                 )
             except ValidationError as e:
-                errors = cast("list[ErrorDict]", e.error_dict)
                 error_summary = e.error_summary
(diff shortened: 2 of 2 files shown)


smotornyuk:
This is not the case for the core functionality, because [resource_form.html](https://github.com/ckan/ckan/blob/master/ckan/templates/package/snippets/resource_form.html#L17) relies on `error_summary` that contains all errors already. I usually see the following when trying to publish invalid draft dataset from the resource creation page

image

Nevertheless, I'd like to have all errors available in template, so generally this direction is good. But you may have modified a wrong place. You are saying "resource was created successfully", but look on the condition at the begining of block with modification 
image

it says `if not data_provided ...`, i.e. this block handles the submission only when resource form was empty. The actual resource creation happens only 5 lines below

image

amercader:
@smotornyuk do you have a suggestion for an alternative approach? Should this be handled in the ValidationError except below the resource update/creation?

wardi:
@smotornyuk In this PR all I'm doing is updating the call to `self.get` to not pass the data values to the form if a validation error happens when setting `state=active` on the package, after creating or updating the resource successfully. These are the exact try blocks:

no-data case (just publish):

https://github.com/ckan/ckan/blob/dfb62a90eb855091dcc238294d3e0fc428f462b7/ckan/views/resource.py#L218-L222

with-data case (publish after create/update):

https://github.com/ckan/ckan/blob/dfb62a90eb855091dcc238294d3e0fc428f462b7/ckan/views/resource.py#L257-L261


The `error_summary` is still being passed so I don't see how this removes any error information:

no-data case:

https://github.com/ckan/ckan/blob/dfb62a90eb855091dcc238294d3e0fc428f462b7/ckan/views/resource.py#L232

with-data case:

https://github.com/ckan/ckan/blob/dfb62a90eb855091dcc238294d3e0fc428f462b7/ckan/views/resource.py#L264

Are you saying that there's a different problem with the code already on master that should be resolved at the same time?

smotornyuk:
No, sorry, my bad. Previously, I didn't see the "with-data case (publish after create/update):" change in the diff tab of GitHub and thought you'd only handled the "no-data case". 
Now, when I see both modifications, that makes sense and I have no other comments

PR #9208, opened 2026-01-09
[2.10] Upgrade security requirements December 2025 (2.10)

amercader:
Related to #9197 
This is just for 2.10. The urllib3 jump is a big one, but according to the [upgrade notes](https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html):

> We’re maintaining functional API compatibility for most users to make the migration an easy choice for almost everyone. Most changes are either to default configurations, supported Python versions, or internal implementation details. So unless you’re in a specific situation you should notice no changes! 

And the tests pass (failures are linting errors)

### DIFF
--- a/ckan/config/middleware/flask_app.py
+++ b/ckan/config/middleware/flask_app.py
@@ -276,7 +276,7 @@ def make_flask_stack(conf: Union[Config, CKANConfig]) -> CKANApp:
     _register_plugins_blueprints(app)
 
     if config.get("ckan.csrf_protection.ignore_extensions"):
-        log.warn(csrf_warn_extensions)
+        log.warning(csrf_warn_extensions)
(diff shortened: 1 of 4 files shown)


pdelboca:
@amercader I'm a little bit afraid of this one. The documentations says:

>  That said, urllib3 1.26.x still sees significant download numbers, mainly because the botocore package still requires urllib3 1.26.x for Python 3.9 and earlier.

`botocore` is for AWS connect and CKAN 2.10 supports 3.9. Could it cause conflict with Cloud Storage extensions? 🤔

amercader:
You are right, better to not risk it. Sites on CKAN 2.10 can manually update urllib3 if they are concerned about the vulnerability reports. I've reverted the upgrade.
Please ignore the lint/typecheck fails, I'll fix separately

PR #9216, opened 2026-01-13
Backport faster datastore_info: include_meta, include_fields_schema parameters

wardi:
Backport #8590 to dev v2.11 as requested by https://github.com/ckan/ckan/issues/8589#issuecomment-3741880368

This is a performance improvement for the resource and data dictionary views. The default values for the new parameters maintain backwards compatibility.

### DIFF
--- a/changes/8770.misc
+++ b/changes/8770.misc
@@ -1,0 +1,1 @@
+Switch 'datastore_info' to use 'resource_id' as input argument


wardi:
I've pulled in #8907 as well to fix failing tests instead of writing an alternate version for 2.11

wardi:
@ThrawnCA for your review

ckanext/datastore/backend/postgres.py
```diff
@@ -2249,15 +2249,17 @@ def resource_plugin_data(self, id: str) -> dict[str, Any]:
...
             return plugin_data
 
-    def resource_fields(self, id: str) -> dict[str, Any]:
+    def resource_fields(
+            self, id: str, include_meta: bool = True,
+            include_fields_schema: bool = True) -> dict[str, Any]:
 
-        info: dict[str, Any] = {'meta': {}, 'fields': []}
-
-        try:
```
ThrawnCA, line 2256: Why was the `try` dropped?

ckanext/datastore/backend/postgres.py
```diff
@@ -2314,19 +2316,23 @@ def resource_fields(self, id: str) -> dict[str, Any]:
...
-            # get the data dictionary for the resource
-            with engine.connect() as conn:
-                data_dictionary = _result_fields(
-                    _get_fields_types(conn, id),
-                    _get_field_info(conn, id),
-                    None
-                )
+        # get the data dictionary for the resource
+        with engine.connect() as conn:
+            data_dictionary = _result_fields(
```
ThrawnCA, line 2326: Should this have a type declaration?

ckanext/datastore/logic/action.py
```diff
@@ -380,11 +384,12 @@ def datastore_info(context: Context, data_dict: dict[str, Any]
 
     '''
     backend = DatastoreBackend.get_active_backend()
+    schema = dsschema.datastore_info_schema()
+    data_dict, errors = _validate(data_dict, schema, context)
+    if errors:
+        raise p.toolkit.ValidationError(errors)
 
-    resource_id = data_dict.get("resource_id", data_dict.get("id"))
-    if not resource_id:
```
ThrawnCA, line 385: Why was this check dropped? The schema can be overridden, yes?

ckanext/datastore/backend/postgres.py
wardi: this was a try with a bare `except Exception: pass` that will quietly hide real problems in the code.

ckanext/datastore/backend/postgres.py
wardi: the return value type declaration on `_result_fields` seems to be enough to keep the type checker happy.

ckanext/datastore/logic/action.py
wardi: IMHO if someone has overridden the schema it's up to them to make sure they haven't broken the core implementation, not the other way around.

ckanext/datastore/logic/action.py
```diff
@@ -380,11 +384,12 @@ def datastore_info(context: Context, data_dict: dict[str, Any]
...
-        raise ValidationError({"resource_id": ["Missing value"]})
-
+    resource_id = data_dict['resource_id']
```
amercader, line 392: Requiring `resource_id` instead of `id` is a breaking change, can we put a fallback to `data_dict["id"]` in place instead of enforce it via validation? Or use [`if_empty_same_as`
](https://github.com/ckan/ckan/blob/5d6d3a4b4755b9f2d504354d35503a74d97b901a/ckan/lib/navl/validators.py#L78)

wardi: The same was accomplished with a `rename` in the `__before` validator https://github.com/ckan/ckan/pull/9216/files#diff-0689e619a39a5dd863ff5bbc36c714d67882e335a1fa2a117803b84cfd6f1c94R250
```python
def datastore_info_schema() -> Schema:
    return {
        'resource_id': [not_missing, not_empty, unicode_safe],
        'id': [ignore_missing],
        'include_meta': [default(True), boolean_validator],
        'include_fields_schema': [default(True), boolean_validator],
        '__before': [rename('id', 'resource_id')]
    }
```

wardi: Still, if this is too much of a change for a backport I'm fine with closing the PR

Issue #9219, opened 2026-01-19
Multiple h1 elements on same page


ThrawnCA:
## CKAN version

2.11.4

## Describe the bug
Many pages contain multiple `h1` elements, even going so far as to have different content in them. This is [poor accessibility practice](https://www.boia.org/blog/multiple-h1-tags-are-bad-for-accessibility-and-seo) as it can confuse screen readers and search engines.

### Steps to reproduce

1) Go to https://demo.ckan.org/dataset/my-org--pcns
2) Examine the page source and search for `<h1`. There are three hits, two of them with copies of the dataset name, and one for the parent organisation name.

### Expected behavior

A page should have only one `h1` element. It should definitely not be in the sidebar.

wardi:
on master we'll have to check both the new theme and the current one

wardi:
please reopen or create a new ticket if this is not fully addressed by #9223

PR #9223, opened 2026-01-20
Replace sidebar 'h1' with 'h2'

ThrawnCA:
Possible solution for #9219

### Proposed fixes:

Use `h2` rather than `h1` in sidebar blocks, as the page should only have one `h1` for compatibility with screen readers and search crawlers.

Update the stylesheet so that the "heading" class is applied on h2 as well as h1.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9219.bugfix
+++ b/changes/9219.bugfix
@@ -1,0 +1,1 @@
+Use 'h2' in sidebar instead of 'h1' as pages should only have one h1


wardi:
Thanks @ThrawnCA

It looks like the new theme doesn't have the same issue.

PR #9230, opened 2026-01-26
Site Lockdown Mode

JVickery-TBS:
feat(auth): site lockdown mode;

- Created a site lockdown for non sysadmins.

Fixes #

### Proposed fixes:

Puts the site into a sort of "read only" mode. Non sysadmins won't be able to execute non side effect free action methods.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9230.feature
+++ b/changes/9230.feature
@@ -1,0 +1,5 @@
+Added a `ckan.site_lockdown` config option which disables actions causing side effects,
+such as `*_create`, `*_update`, and `*_delete`, for non-sysadmin users.
+
+This setting does not prevent updates to the database from sysadmin users or updates
+that skip the action API, such as collecting page view tracking data.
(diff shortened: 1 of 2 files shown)


wardi:
There were some concerns raised about the negative connotations of "lockdown". We could go with Read-only but make sure the documentation makes it clear that it's not 100% read-only.

Other terms we could use include "quiet mode" or "restricted mode". Any other suggestions?

JVickery-TBS:
So the options are:

- `ckan.unwritable_mode`
- `ckan.lockdown_mode` (VETO'd)
- `ckan.readonly_mode` (what we have on Canada fork rn)
- `ckan.quiet_mode`
- `ckan.restricted_mode`
- `ckan.sysadmin_mode` (this one is fairly descriptive of what is actually happening?)

@wardi @pdelboca lemme know what u think! I am currently leaning towards `ckan.sysadmin_mode`

pdelboca:
I like the `readonly_mode` name for it. I say it on other softwares and it was quite forward to understand what was happening.

PR #9232, opened 2026-01-28
[post 2.12] group/org_show: include_users=False default

wardi:
`group_show` and `organization_show` set the default value of `include_users` based on `ckan.auth.public_user_details`.

Other actions simply default to False for a faster response time by avoiding work in the common case.

Let's do the same for `group_show` and `organization_show` after the 2.12 release and after giving users notice of the change. For users that have already configured `ckan.auth.public_user_details = false` there will be no difference.

### Proposed fixes:

default `include_users` to False consistently across group and org actions.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/logic/action/test_get.py
+++ b/ckan/tests/logic/action/test_get.py
@@ -2059,7 +2059,7 @@ class TestUserAutocomplete(object):
         result = helpers.call_action("user_autocomplete", q="compl", limit=1)
         assert len(result) == 1
 
-    def test_autcomplete_email(self):
+    def test_autocomplete_email(self):
         user = factories.Sysadmin()
         context = {"user": user["name"]}
         factories.User(name="user1234", email="joe@doe.com")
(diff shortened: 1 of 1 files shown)


wardi:
Thanks @smotornyuk . I intended this change to be merged after we branched off for the 2.12 release because it's possible this will break existing workflows and I'd like us to have a release warning users about the change.

@amercader WDYT shall we start the 2.12 branch now off the parent commit or should this be reverted on master?

smotornyuk:
Sorry, it mixed it my head and I thought we were going to make it available after 2.**11**

smotornyuk:
I think that I will merge a revert-commit, and then create another revert-revert with these changes inside reopened PR. Any objections?

wardi:
instead of the revert-revert commit I'll rebase these commits to after the revert in a new PR

smotornyuk:
Ok, I reverted this merge inside 457a37e2e3 commit

Issue #9107, opened 2025-09-16
cgitb module not present in Python 3.13


jgrocha:
## CKAN version

Current master

## Describe the bug

Installing from source on a Python 3.13 environment is not working.

The cgitb was deprecated in Python 3.11 and removed in 3.13 ([deprecated notice](https://docs.python.org/3/library/cgitb.html)).

The last version of Python that provided the cgitb module was [Python 3.12](https://docs.python.org/3.12/library/cgitb.html).

### Steps to reproduce

```bash
$ python3 --version
Python 3.13.3
$ python3 -m venv /usr/lib/ckan/default
$ . /usr/lib/ckan/default/bin/activate
$ pip install -e 'git+https://github.com/ckan/ckan.git#egg=ckan[requirements,dev]'
$ ckan
Traceback (most recent call last):
  File "/usr/lib/ckan/default/bin/ckan", line 3, in <module>
    from ckan.cli.cli import ckan
  File "/home/jgr/ckan/lib/default/src/ckan/ckan/cli/cli.py", line 14, in <module>
    from ckan.config.middleware import make_app
  File "/home/jgr/ckan/lib/default/src/ckan/ckan/config/middleware/__init__.py", line 11, in <module>
    from ckan.config.environment import load_environment
  File "/home/jgr/ckan/lib/default/src/ckan/ckan/config/environment.py", line 21, in <module>
    import ckan.lib.search as search
  File "/home/jgr/ckan/lib/default/src/ckan/ckan/lib/search/__init__.py", line 7, in <module>
    import cgitb
ModuleNotFoundError: No module named 'cgitb'
```

jgrocha:
The alternative is to continue to use 3.11, for example, to run ckan:

```bash
sudo apt install python3.11 python3.11-venv python3.11-dev
/usr/bin/python3.11 -m venv /usr/lib/ckan/default
pip install -e 'git+https://github.com/ckan/ckan.git@ckan-2.11.3#egg=ckan[requirements]'
```

Zharktas:
`cgitb` probably can be removed, it's only used for logging errors from solr, `cgi` was removed in Python 3.13 as well, so backwards compatible `cgi.FieldStorage` file handling need to be removed as well.

Zharktas:
Whomever implements this, probably should try to target python 3.14 which was released yesterday.

jmckenna:
@jgrocha trying using the replacement [ legacy-cgi ](https://pypi.org/project/legacy-cgi/)

PR #9236, opened 2026-01-30
[#9107] Python 3.13 and 3.14 support

amercader:
Changes to make CKAN master (2.12) work on Python 3.13 and 3.14:

* Remove cgitb module: This module was deprected on Python 3.11 and removed on Python 3.13. Its usage has been refactored to use standard library modules
* Remove usage of cgi module]: The cgi module, removed on Python 3.13, was used to identify files uploaded in the old Pylons web framework. We can safely focus on the Werkzeug/Flask ones now

Part of the cross-python version support tracked [here](https://github.com/ckan/ckan-python-monitor)

Note that CKAN 2.10 and 2.11 can't be installed in Python 3.12+ and 3.13+ respectively because of incompatible lxml versions (TODO separate issue)

### DIFF
--- a/ckan/lib/uploader.py
+++ b/ckan/lib/uploader.py
@@ -18,7 +18,6 @@ import ckan.plugins as plugins
 from ckan.common import config
 from ckan.types import ErrorDict, PUploader, PResourceUploader
 
-ALLOWED_UPLOAD_TYPES = (FlaskFileStorage,)
 MB = 1 << 20
 
 log = logging.getLogger(__name__)
@@ -38,10 +37,6 @@ def _copy_file(input_file: IO[bytes],
         output_file.write(data)
         if current_size > max_size:
             raise logic.ValidationError({'upload': ['File upload too large']})
-
-
-def _get_underlying_file(wrapper: FlaskFileStorage):
-    return wrapper.stream
 
 
 def get_uploader(upload_to: str,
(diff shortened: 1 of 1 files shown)


Zharktas:
It would be nice to run tests on CI on newer python, currently its still 3.10 https://github.com/ckan/ckan/blob/1902693db45f27bd6a4b93fb0f16b83bba5f79b8/.github/workflows/pytest.yml#L114

amercader:
@Zharktas We generally run the test suite in the minimum Python version supported, if we decide to bump the minimum version for 2.12 then we can definitely increase it. Let's discuss on the dev call

Python 3.10 EOL is [next October](https://devguide.python.org/versions/)

ckan/lib/uploader.py
```diff
@@ -41,10 +40,8 @@ def _copy_file(input_file: IO[bytes],
             raise logic.ValidationError({'upload': ['File upload too large']})
 
 
-def _get_underlying_file(wrapper: Union[FlaskFileStorage, cgi.FieldStorage]):
-    if isinstance(wrapper, FlaskFileStorage):
-        return wrapper.stream
-    return wrapper.file
+def _get_underlying_file(wrapper: FlaskFileStorage):
+    return wrapper.stream
```
Zharktas: Should this function be removed completely, since it does nothing expect return a stream. Its called only in two places in the same file.

ckan/lib/uploader.py
```diff
@@ -19,7 +18,7 @@
 from ckan.common import config
 from ckan.types import ErrorDict, PUploader, PResourceUploader
 
-ALLOWED_UPLOAD_TYPES = (cgi.FieldStorage, FlaskFileStorage)
+ALLOWED_UPLOAD_TYPES = (FlaskFileStorage,)
```
Zharktas: No need  to check allowed upload types when we only allow one type.

Issue #9237, opened 2026-01-30
CKAN 2.10 and 2.11 don't work on latest Python versions


amercader:
Part of the cross-python version support tracked [here](https://github.com/ckan/ckan-python-monitor)

After fixing issues like https://github.com/ckan/ckan/pull/9235 installing CKAN 2.10 and 2.11 still fails in the following versions:

* CKAN 2.10 : Python 3.12+
* CKAN 2.11 : Python 3.13+

The reason is that the lxml versions pinned in requirements.txt (4.9.1 and 5.2.2 respectively) are not compatible with these Python versions.

Lxml is a transitive requirement introduced by feedgen, an (unmaintained) requirement used to generate the Atom feeds in `ckan/views/feeds.py`. When manually upgrading lxml to the latest version (or at least a newer one) the relevant tests still pass (`ckan/tests/controllers/test_feed.py`).

* In 2.11 we could pin lxml to the relevant version in requirements.in to ensure it install a Python 3.12+ compatible version (same in 2.10 if we think it's worth it)
* Alternatively we can just document on the changelog that users targeting newer Python versions need to manually upgrade lxml
* Going forward (2.12 or 3.0) it would be good to split the feeds into their own separate extension, and refactor them to not rely on the unmaintained feedgen. I think it would be great that CKAN core did not have the lxml requirement, as it adds complexity to deployments.

Zharktas:
Apparently lxml is not the only one needing upgrades based on https://github.com/ckan/ckan-python-monitor/actions/runs/21704962634/job/62593663303 

<img width="571" height="64" alt="Image" src="https://github.com/user-attachments/assets/bd987736-9904-4d87-92e3-a3a8771bace8" />

greenlet is transitive via sqlalchemy, others are in requirements.in.

amercader:
Bumping requirements like pysolr in CKAN 2.11 to support Python 3.14 means dropping support for Python 3.9.

It is a breaking change for existing instances of course, but there is precedent e.g. in [2.10.5](https://docs.ckan.org/en/2.11/changelog.html#v-2-10-5-2024-08-21)

PR #9239, opened 2026-02-04
[2.11][#9237] Include lxml as a main requirement to support Python 3.13+

amercader:
For CKAN 2.11

The version pinned in requirements.txt (introduced via feedgen) is too old to support Python 3.13+, so we need to include it in our main requirements to bump it to a modern version.

Ref #9327


amercader:
@Zharktas as discussed in the last dev meeting (and #9238)

PR #9251, opened 2026-02-10
Resource Position Unique Index

JVickery-TBS:
feat(model): res position unique index;

- Create a unique index for active resource positions.

Fixes #

### Proposed fixes:

With all the metadata code changes with the index and position code changes, it could be beneficial to have an actual unique constraint at the model level to prevent any funny position behavior.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/model/resource.py
+++ b/ckan/model/resource.py
@@ -199,6 +199,7 @@ meta.registry.map_imperatively(Resource, resource_table, properties={
                 'position',
                 # type_ignore_reason: incomplete typing
                 ordering_func=_get_stately_resource_position),  # type: ignore
+            order_by=resource_table.c.position,
             cascade='all, delete'
         ),
     )
--- a/changes/9093.misc
+++ b/changes/9093.misc
@@ -1,0 +1,1 @@
+Added `remove_locale_from_url()` and `remove_root_path_from_url()` template helpers
--- a/changes/9232.misc
+++ b/changes/9232.misc
@@ -1,0 +1,3 @@
+After CKAN 2.12 the include_users parameter to organization_show and group_show
+actions will default to False regardless of the ckan.auth.public_user_details
+configuration setting.
--- a/changes/9265.bugfix
+++ b/changes/9265.bugfix
@@ -1,0 +1,1 @@
+Fixed an issue with Resources not uploading new files during `package_update` some of the time.
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -739,6 +739,10 @@ def get_flashed_messages(**kwargs: Any):
 
 @core_helper
 def endpoint_from_url(url: str) -> str:
+
+    url = remove_locale_from_url(url)
+    url = remove_root_path_from_url(url)
+
     try:
         urls = current_app.url_map.bind("")
         match = urls.match(url)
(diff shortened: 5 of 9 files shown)


JVickery-TBS:
hmmm so unique index isn't gunna work in this case right. cuz we need to be able to defer it to the end of the transaction cuz its gunna reorder positions and that will fail. But I don't think we can put a UniqueConstraint in its place as we need to have state=active

and to my knowledge unique constraints can't do where clauses or column literals either...hmmm. If we do the UniqueConstraint, the issue would be with deleted resources all getting a 0 position. So in THEORY we can have them go into negative values? (that kinda seems crazy to do in a database though just to fix in a unique constraint)

JVickery-TBS:
RIGHT we can just make deleted resources have a position of null in the DB. that is easy enough

ckan/migration/versions/109_e12e991f586b_create_res_position_constraint.py
```diff
@@ -0,0 +1,41 @@
...
+down_revision = 'f7b64c701a10'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("""
+        UPDATE resource SET position=null WHERE state='deleted';
+    """)
+    print('Set position=null for deleted resources.')
```
wardi: We should print nothing on success so these can be skipped

ckan/model/resource.py
```diff
@@ -171,15 +174,33 @@ def related_packages(self) -> list[Package]:
...
+def _get_stately_resource_position(
+        index: int, resources: List[Resource]) -> Optional[int]:
+    """
+    Give state='deleted' resources null positions.
+
+    Required for con_package_resource_unique_position
+    unique constraint when deleting resources.
+    """
+    if resources[index].state != 'deleted':
+        return index
```
wardi, line 186: could this leave gaps in the positions when there are deleted resources?

ckan/migration/versions/109_e12e991f586b_create_res_position_constraint.py
JVickery-TBS: k will remove.

ckan/model/resource.py
JVickery-TBS: hmm good question. I don't THINK so, but I will see if I an put some more test coverage in for it.

JVickery-TBS: @wardi it does NOT, I actually have test coverage for mass resource soft deletes and mass resource purgings during package patch/update in our canada plugin for this constraint. So I will bring some more tests in to cover that stuff. But basically the backref collection for position is always gunna start at zero when the package_dictize_save calls the resource_list_save function, when it adds the whole list onto the backref thiny, it will count 0,1,2,3,etc from the python list.

The deleted ones are added one by one in a python loop afterwards, those would all go through the backref collection thingy one at a time, starting at 0 too. Which is why deleted resources always get position 0. But now, they will get null.

But yeah I will throw a bunch more test coverage in here

JVickery-TBS: @wardi okay I added a bunch more test coverage to cover deleting and purging resources from a dataset, and the recalculation of the resource positions.

I also added in a query to the migration script to set the correct resource positions, in case there are duplicates already in the db

JVickery-TBS:
i think codecov is just screaming cuz of all the code additions in the migration script

ckan/migration/versions/109_e12e991f586b_create_res_position_constraint.py
```diff
@@ -0,0 +1,58 @@
...
+                    PARTITION BY package_id
+                    ORDER BY position
+                ) - 1 AS new_position
+            FROM resource WHERE state='active'
+        )
+        UPDATE resource
+            SET position=numbered.new_position
+            FROM numbered
+            WHERE resource.id = numbered.id
+            AND resource.position != numbered.new_position;
```
wardi, line 45: shouldn't we only need to clear the position of existing resources, not renumber all existing ones?

JVickery-TBS: Yeah this query would only change active resources that do not have a correct position due to the `AND resource.position != numbered.new_position;`

So basically if you had a resource list with positions like: 1, 2, 4, 5, 5

This would be able to sort by position for active resources of that dataset, and then just use the psql ROW_NUMBER counter, in the above case would be 0, 1, 2, 3, 4 and then change the above resource positions to the correct numbers.

This is mainly just a scenario for encase you have super super old packages that were running on like CKAN 2.8 or something, as it would fail when trying to make the constraint in this migration. Cuz a package could have multiple of the same positions. We ran into this with active resources. But not sure if that was just a wildly specific bug in our fork?

If you don't think this should really happen in normal CKAN installs, we can omit this query?

wardi: It's ok to keep this but please add a comment that this resolves an issue with bad numbering that might occur in existing databases.

PR #9252, opened 2026-02-11
[post 2.12] Faster datastore_search row counts (estimate version)

### LINKED ISSUE #9234: [post-2.12] faster datastore row counts
Row counts are a major performance drain on `datastore_search` and `datastore_info` calls, especially for large tables (delays measured in seconds).

In other PR's we've reduced the amount of work required for datatables and datastore dump, views but for normal users of the datastore APIs this performance issue is real and very noticeable.

Lets:
- schedule a background job to compute the number of rows after each datastore table modification
- rely on the computed totals by default for calls to datastore_search, datastore_info (when no filters applied)
- return data faster, make users happier

wardi:
Fixes #9234

### Proposed fixes:
- schedule a background job to compute the number of rows after each datastore table modification
- rely on the computed totals by default for tables with a noticeable row count time (>20k rows) for calls to datastore_search when no filters are applied

This is intended for after the dev-2.12 branch is created. A separate PR will warn about the upcoming new datastore API default values once this PR is approved.

#### Timings

Example calls to `datastore_search` with no filters, returning default 100 records

table rows | before (s) | after (s) | total | change
---| --- | --- | --- | ---
13M | 1.760 | 0.041 | estimate | 43x
2M | 0.515 | 0.041 | estimate | 13x
0.1M | 0.075 | 0.040 | estimate | 1.9x
0.02M | 0.059 | 0.059 | exact | 1x

### Discussion:

This implementation uses the postgres table statistics to generate approximate row counts instead of caching exact row counts, but is simple and adds no additional storage to the datastore or application database.

1. Should we store exact counts instead of using the available approximate table statistics?
2. If we store exact counts or other table statistics, should they live in the datastore database, redis, or the application database (e.g. as part of the resource metadata)

Note: Some types of useful table statistics, such as frequency tables, can be very large and would significantly bloat resource metadata and impact performance if they are stored there.

### Features:

- [ ] includes API changes

TODO: add tests

### DIFF
--- a/ckanext/datastore/logic/schema.py
+++ b/ckanext/datastore/logic/schema.py
@@ -209,7 +209,10 @@ def datastore_search_schema() -> Schema:
         'sort': [ignore_missing, list_of_strings_or_string],
         'distinct': [ignore_missing, boolean_validator],
         'include_total': [default(True), boolean_validator],
-        'total_estimation_threshold': [default(None), int_validator],
+        'total_estimation_threshold': [
+            configured_default('ckan.datastore.default_total_estimation_threshold',
+                               20000),
+            int_validator],
         'records_format': [
             default(u'objects'),
             one_of([u'objects', u'lists', u'csv', u'tsv'])],


wardi:
@EricSoroos I've added some points for discussion in the description above

wardi:
Where I fall on these questions:
1. Yes we should store exact values for row counts but this could be implemented as a follow-up PR
2. Table statistics should live in the datastore database for more efficient updates (all SQL, no marshaling through Python) some statistics may be surfaced as resource metadata if desired through before_index plugins and custom solr schemas.

wardi:
closing in favor of #9260

PR #9260, opened 2026-02-13
Faster datastore_search row counts (cache version)

### LINKED ISSUE #9234: [post-2.12] faster datastore row counts

wardi:
Fixes #9234

### Proposed fixes:

This is an alternate version of #9252 that computes the actual row counts and caches the number in a `_table_stats` table in the datastore database. No estimated row counts will be returned, in fact, the whole estimate rows feature is removed as part of this PR.

The existing estimated-row-counts-from-the-query-planner feature has been removed because that estimate was only ever used when this cached value can be returned instead.

If a background worker is not running or busy, this version will calculate the exact row count and cache it on the next call to `datastore_search` resulting in a single slow response but not a stale count value.

This `_table_stats` table can also be used by other extensions to cache statistics like frequency tables, histograms etc.

#### Timings

Example calls to `datastore_search` with no filters, returning default 100 records

table rows | before | after | change
--: | --: | --: | --:
13,000k | 1.760s | 0.041s | 42.9x
2,000k | 0.515s | 0.041s | 12.6x
100k | 0.075s | 0.040s | 1.9x
20k | 0.059s | 0.041s | 1.4x

### Features:

- [ ] includes API changes


EricSoroos:
Looking a bit at the index-only count(*) scans, and it appears that at some point index only scans started working on postgres. 

The restrictions are: 
* There has to be a single covering index for all the columns of the query
* The data in each page of the table needs to be older than all outstanding open transactions, so that the page flag for visiblity map needs to be set, so then it's a heap visibility scan which scans _far_ less data than a data scan. 

from https://www.postgresql.org/docs/current/indexes-index-only-scans.html

> In short, while an index-only scan is possible given the two fundamental requirements, it will be a win only if a significant fraction of the table's heap pages have their all-visible map bits set. But tables in which a large fraction of the rows are unchanging are common enough to make this type of scan very useful in practice.

We've got an index on _id that works as a covering index, so the only issue that we'd get is that we're trying to do the counts just after an update, so we'd be getting the slow path. 

What I'm seeing on a small test system (running pg17) with ~5 minute updates, about half of the tables are showing that I can do a fully indexed count(*), and half are showing that it's a table scan. 


e.g. two ~ 130k row tables with different plans.
``` 
explain analyze select count(*) from "54044dc6-8ea7-42af-89ec-230270e60647";

                                                                                                 QUERY PLAN                                                                                                  
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=4644.05..4644.06 rows=1 width=8) (actual time=50.600..50.601 rows=1 loops=1)
   ->  Index Only Scan using "54044dc6-8ea7-42af-89ec-230270e60647_pkey" on "54044dc6-8ea7-42af-89ec-230270e60647"  (cost=0.29..4297.09 rows=138786 width=0) (actual time=0.045..37.866 rows=135382 loops=1)
         Heap Fetches: 45279
 Planning Time: 0.115 ms
 Execution Time: 50.656 ms
(5 rows)

explain analyze select count(*) from "2bb017bc-bbb5-4a66-8fa1-e5e98cdd2e8f";
                                                                     QUERY PLAN                                                                     
----------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=5623.95..5623.96 rows=1 width=8) (actual time=49.449..49.451 rows=1 loops=1)
   ->  Seq Scan on "2bb017bc-bbb5-4a66-8fa1-e5e98cdd2e8f"  (cost=0.00..5369.96 rows=101596 width=0) (actual time=0.042..38.532 rows=101596 loops=1)
 Planning Time: 1.191 ms
 Execution Time: 49.532 ms
(4 rows)
```

These are quite similar for performance which is a little surprising, but for another install with much bigger tables, I'm seeing roughly 80msec count(*) calls on a 1.5M row table on an index only scan.

wardi:
@EricSoroos that's very cool, I'll need to upgrade to try it out.

With the cache we can get the row count in less than 1ms so I think a feature like this for the datastore is still useful. Hitting the slow path in a background job doesn't seem like a problem to me, does having the index-only scan support change how this could be implemented in ckan?

ckanext/datastore/set_permissions.sql
```diff
@@ -106,3 +106,53 @@ DO $body$
...
+CREATE OR REPLACE FUNCTION update_table_stats(table_name name)
+    RETURNS BOOL
+AS $body$
+    BEGIN
+        EXECUTE format('ANALYZE public.%I', table_name);
+        PERFORM cached_table_row_count(table_name);
+        RETURN TRUE;
+    END;
+$body$ LANGUAGE plpgsql;
+ALTER FUNCTION update_table_stats() OWNER TO {writeuser};
```
wardi: @EricSoroos this function can be overridden by site owners to populate the `_table_stats` (or any other) table with information like min/max, frequencies, cardinality, histograms etc.

It's called from a scheduled background worker that is automatically rescheduled if another update/delete occurs on the same table. It's allowed to take longer than just calculating the row count, which is why I've got an ANALYZE in here.

wardi:
@EricSoroos I've merged master, applied another small fix and added tests. Please review again.

ckanext/datastore/set_permissions.sql
```diff
@@ -106,3 +107,53 @@ DO $body$
     END;
 $body$;
 
+-- _table_stats for caching row counts and other useful table statistics
+CREATE UNLOGGED TABLE IF NOT EXISTS "_table_stats" (
```
EricSoroos: I don't like that this is created unlogged. 

1) There's not heavy write load on this table, vs the general write load against the datastore, as writes are coalesced per external table. It's a very small set of stats to write. It's also very unlikely to conflict, because it's a write per resource.
2) The IO load for this is more likely to be the tablescans to read all the tuples in the target table rather than the one row write against the table.
3) I'm not sure if docker containers typically shut down the database, or if they just do a crash like thing to quit. I know this is a cache, but crash recovery and having to refill the cache at the same time might make things just that much worse.
4) If people do have a more advanced system than just a single datastore db, unlogged tables don't replicate or backup.
5) It's easy to go from a logged table to an unlogged table if it's a performance issue. It's not easy to go the other way if you want durability.

wardi:
@EricSoroos reverted unlogged. The one failing test is unrelated to this feature

EricSoroos (approved):

amercader:
@wardi can you look at that conflict so I can merge? cheers

wardi:
@amercader merged and applied fixes related to other PRs. The schema test failure is unrelated

wardi:
@amercader this includes a similar fix to #9353

Issue #9264, opened 2026-02-17
Resource Update File Upload Fails (Flaky)


JVickery-TBS:
## CKAN version

master branch

## Describe the bug

So this is based off of the changes from https://github.com/ckan/ckan/pull/8421 for faster metadata updates. Along with https://github.com/ckan/ckan/commit/d3fff92ebefa1109dfe7cf9c0cad072eaba2bc0b

Basically we see that package_dict_save saves the Resource objects with their positions as they were added in the order of the python list. However, the return of the Package model object from package_dict_save does NOT return the Resource objects in this position. This is very problematic as to how all the new code works from PR#8421 as it loops based on position values and not on resource indices anymore. However, this piece of code to actually call the Uploader's upload method is still based on the resource index of the Package model object.

This seems to be a flaky situation, dependent on how the db returns the resources into the Package object. The issue is that some users might not even know that their file was not uploaded if the new file has the same name. As the framework will still be able to resolve the old filename w/ resource ID for the file download. It is clearer when uploading a file to an existing resource with a new filename.

### Steps to reproduce

- Have a dataset with some resources;
- Reorder the resources;
- Update a resource with a new file (easier to see if it has a new filename, as using the same filename will just still show to old file);
- The upload will "silently" not get called and the file won't be uploaded.

### Expected behavior

Updated resources with new files actually get uploaded.

### Additional details

Specifically, the package_dict_save should return the Resource objects in the order of position, much like the package_dictize method does during package_show action. I looked into the core code quickly, and it does not seem like there are any other places affected. But I could assume, depending on plugin code, the `IPackageController.edit(pkg)` could be affected if any plugin code relied on the correct position ordering.

I am currently debugging still and will open a PR soon. Hopefully this morning. cc: @wardi

JVickery-TBS:
Fortunately this is a super easy fix, just need to add `order_by=resource_table.c.position` to the `resources_all` backref. The harder part is just making up the pytests as this is a flaky issue, the pytests might be a bit funny looking

JVickery-TBS:
Should be fixed in: https://github.com/ckan/ckan/pull/9265

PR #9265, opened 2026-02-17
Fix Flaky Resource File Uploads during `package_update`

JVickery-TBS:
fix(model): resource flaky uploads;

- Fix a major issue silently failing resource uploads.

Fixes #

### Proposed fixes:

Fixes: https://github.com/ckan/ckan/issues/9264

Flaky issue when updating a resource with an upload.

PS - sorry about the crazy test coverage for this one, its like a flaky issue and the only way to get it is in different db sessions, so removing and re-initializing db sessions in the test cover it. That new test consistently fails if the backref does not have that new `order_by`. If all the resources are added in a single db session, they are returned in the correct order in which they were added into the database. its with different db sessions when stuff gets crazy.

Wondering if it is better to do a negative test? And somehow patch the backref to remove the order_by.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/tests/logic/action/test_update.py
+++ b/ckan/tests/logic/action/test_update.py
@@ -1374,6 +1374,8 @@ class TestResourceUpdate(object):
             assert current_file_data != content_1
             assert current_file_data == content_2
 
+        helpers.call_action('package_delete', id=dataset.id)
+
     def test_extras(self):
         user = factories.User()
         dataset = factories.Dataset(


ckan/model/resource.py
```diff
@@ -178,6 +178,7 @@ def related_packages(self) -> list[Package]:
         # formally package_resources_all
         backref=orm.backref('resources_all',
                             collection_class=ordering_list('position'),
+                            order_by=resource_table.c.position,
```
wardi, line 181: I like this change because it's a sensible default, but it might be better to also explicitly set the order_by on the actual queries that rely on the resource ordering.

JVickery-TBS: Yeah for sure, its mainly cuz its on the Package model object, in its `resources` property, which just accesses the `resources_all` backref:

```
    @property
    def resources(self) -> list["Resource"]:
        return [resource for resource in
                self.resources_all
                if resource.state != 'deleted']
```
if the backref has the order_by, then we would not need a python sorted in here eh?

wardi: never mind I thought we got back a query object here not an actual list

PR #9269, opened 2026-02-19
Tabledesigner -- Allow user to delete last field

EricSoroos:
The tabledesigner view for the data dictionary enforces that fields is a list, however when deleting the last field, this is empty, and therefore null. This results in a confusing 400 validation error to the user. 

This traps it to prevent the 400. 

@wardi  I'm not sure if this is breaking a validation, because now a bare post to this endpoint will delete all the fields in the table. It might be a better approach in the long term to have tombstones for deleted fields. 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/tabledesigner/views.py
+++ b/ckanext/tabledesigner/views.py
@@ -45,6 +45,9 @@ class _TableDesignerDictionary(MethodView):
         if not isinstance(info, list):
             info = []
         custom = data.get('fields')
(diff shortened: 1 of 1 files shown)


wardi:
@EricSoroos I'm fine with this. Please add a regression test and a changelog entry.

Tombstoning is a great idea and would be a real improvement over the current code, but this is a good immediate fix.

PR #9276, opened 2026-02-26
8189 clean activity history action

stefina:
Fixes [#8189](https://github.com/ckan/ckan/issues/8189)

### Proposed fixes:
As we would benefit from this functionality, too, I checked out the branch by https://github.com/TomeCirun referenced in this PR https://github.com/ckan/ckan/pull/8322 and fixed the merge-conflicts.
I also added deletion of the related activity_detail-tables before deleting the activities itself.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/activity/logic/action.py
+++ b/ckanext/activity/logic/action.py
@@ -790,7 +790,8 @@ def _delete_activities_with_keep(
     """Delete activities in range but keep N most recent per object_id.
     """
 
-    # Order activities by timestamp within each object_id and rank them (i.e number them).
+    # Order activities by timestamp within each object_id and rank them
+    # (i.e. number them).
     ranked = (
         query
         .with_entities(


stefina:
@wardi not sure if the self-assign was automated, but while the build is still failing, ~I think this is good to review now.~
If you have any pointers on how to fix it and of course, if there are any changes you like to see in general, kindly let me know.

And of course, thank you to @TomeCirun for your great work!

EDIT: I already tested the CLI locally, but realized that the view of activities in the admin-trash-view does not work.
I will have a look at this tomorrow.

stefina:
The build is now succeeding. I also added a "Purge" button per day in the UI.
So far you can either delete all activities or all activities of one day via the UI. While this works, I feel like this could confuse the average portal admin but was the best I can do with the time I had. 
There are multiple ways to approach this but for the time being it should be good enough and a good basis to build upon.
Please review 👍 

Screenshot 2026-03-01 at 14 35 57

ckanext/activity/logic/action.py
```diff
@@ -663,3 +660,97 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
         "activities": activities,
     }
+
+
+@tk.validate(schema.default_activity_delete_schema)
+def activity_delete(context: Context, data_dict: DataDict) -> dict[str, Any]:
+    """
+    Deletes activities from the database based on a specified date range or
+    offset days.
+
```
amercader, line 699: Add `id`  to the supported params, and mention that you need to provide id, start/end date or an offset

ckanext/activity/logic/action.py
```diff
@@ -663,3 +660,97 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
+    :param start_date: The start date in 'YYYY-MM-DD' format.
```
amercader: Dates are great, but arbitrary time stamps are even better! Users might want to delete activities between two hours or within the last ten minutes if they have been testing.
We can allow values following ISO 8601 here besides just dates so hours and minutes can also be defined. You can just use the existing `isodate` validator instead of the custom `convert_yyyy_mm_dd_format` one for these two fields.
It should support both `YYYY-MM-DD` and `YYYY-MM-DDTHH:mm` so it shouldn't break any existing logic in the UI or CLI

ckanext/activity/logic/action.py
```diff
@@ -663,3 +660,97 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
+    :type start_date: str
+    :param end_date: The end date in 'YYYY-MM-DD' format.
+    :type end_date: str
+    :param offset_days: Number of days from today. Activities older
```
amercader, line 910: To support more granular offsets than days we can support ISO 8601 [durations](https://en.wikipedia.org/wiki/ISO_8601#Durations). For the day use case they are defined as `P1D` (1 day), `P3D` (3 days) etc. If we wanted to clear activities from the last 2 hours we would use `PT2H`, for 10 minutes `PT10M`... (of course these can be combined)

Using the builtin `isodate` module we can easily parse and validate these within a CKAN validator that returns the timedelta object that we need:

```python
import isodate

delta = isodate.parse_duration("PT2H")  
# datetime.timedelta(seconds=7200)
# Will raise an exception if wrong input

```

ckan/templates/admin/trash.html
```diff
@@ -16,10 +16,12 @@
   </div>
```
amercader, line 16: All changes in `templates/admin` need to be replicated in `templates-midnight-blue/admin` otherwise you get an exception using the new theme.

You can test this using the following in your config file:

```
ckan.base_public_folder = public-midnight-blue
ckan.base_templates_folder = templates-midnight-blue
```

ckanext/activity/templates/admin/snippets/data_type.html
```diff
@@ -0,0 +1,62 @@
+{% ckan_extends %}
```
amercader: This snippet needs to be copied to `ckanext/activity/templates-midnight-blue/admin`

ckanext/activity/templates/admin/snippets/data_type.html
```diff
@@ -0,0 +1,62 @@
...
+    {% if ent_type == 'activity' %}
+        <i class="fa fa-tasks p-2"></i>
+    {% else %}
+        {{ super() }}
+    {% endif %}
+{% endblock accordion_icon %}
+
+{% block accordion_title %}
+    {% if ent_type == 'activity' %}
+        {{ _('Deleted activities') }}
```
amercader: I know the rest of entities are actually deleted, but it's probably less confusing to just say "Activities" here as they are not yet deleted.

ckanext/activity/cli.py
```diff
@@ -0,0 +1,83 @@
...
+from ckan import model, logic, types
+
+
+@clean.command(
+    "activities",
+    help="Deletes activities based on a specified date range or offset days.",
+)
+@click.option(
+    "--start_date",
+    type=click.DateTime(formats=["%Y-%m-%d"]),
```
amercader: Update CLI params based on the action params changes

ckanext/activity/cli.py
```diff
@@ -0,0 +1,83 @@
...
+    help="The end date in 'YYYY-MM-DD' format.",
+)
+@click.option(
+    "--offset_days",
+    type=click.INT,
+    help="Number of days from today. Activities older than this will "
+    "be deleted",
+)
+@click.option(
```
amercader, line 42: Other `ckan` commands use this to skip confirmation:

```
@click.option(
     "-f", "--force", is_flag=True, help="Do not ask for confirmation."
)
```

ckanext/activity/cli.py
```diff
@@ -0,0 +1,83 @@
...
+from ckan.cli.clean import clean
+from ckan import model, logic, types
+
+
+@clean.command(
+    "activities",
+    help="Deletes activities based on a specified date range or offset days.",
+)
+@click.option(
+    "--start_date",
```
amercader: `ckan` params options use hyphens rather than underscores (e.g. `--start-date`)

ckanext/activity/views.py
```diff
@@ -995,3 +997,115 @@ def dashboard_testing() -> str:
...
+
+    def _get_deleted_activities(self) -> List[dict[str, Any]]:
+        """Return a flat list of dicts (id, type, title, name, date_str)
+        for display and purge.
+        date_str is the date of the activity in the format YYYY-MM-DD.
+        """
+        activities = (
+            model.Session.query(Activity)
+            .order_by(Activity.activity_type, sa.desc(Activity.timestamp))
+            .all()
```
amercader: This might be really expensive on really big sites. Maybe we can limit it to display only the 1000 more recent activities (or 10000, whatever makes sense). Users can still purge *all* activities from the UI but only the most recent ones are displayed.

We can add a notice somewhere in the list and when purging all activities.

ckanext/activity/templates/admin/snippets/data_type.html
stefina: Thank you very much for your review and your feedback! Already working on the other points, hope I can squeeze that in.
Thinking about this a bit more, I actually think, it might make more sense, to give "Activity Cleanup" its own tab in the Sysadmin-Panel as it is really something different than trash. What do you think? @amercader

amercader: That makes a lot of sense and will simplify the template logic. Also we can expand the UI further if needed in the future

ckan/templates/admin/trash.html
stefina: This is obsolete now, that activity-management was moved to its own tab.

ckanext/activity/templates/admin/snippets/data_type.html
stefina: Should also be obsolete, as activity management is moved to its own tab.

ckanext/activity/logic/action.py
stefina: That would be super helpful. I will look into combining this with the `offset-days`-parameter (e.g. delete 1 day worth of activities a week ago)

ckanext/activity/views.py
stefina: I am actually wondering how useful it is, to display the activities at all. Sure, we can display a part of them, but maybe it is more useful anyways to just display the number of activities per day/month/year if at all (but that might be just as expensive?) 
I think to provide the most basic commands would probably more valuable, e.g.
- "purge all activities"
- deleting the oldest activities -> "purge activities older than 1 day, 1 month, 1 year"
- for deleting most recent activities -> "purge activities of duration 1 hour, duration 1 day, duration 1 month"

WDYT @amercader ?

amercader: Yes, I agree that showing the actual activities is not very helpful. Counts might be helpful to get a sense of the size of the table (you could show total activities / from last year / last month), but again this is just a nice to have.

I think we can keep the UI simple and offer just three options:

* Delete all
* Delete activities older than 1 day, 1 month, 1 year
* Delete between two dates for all other cases

How does that sound?

stefina:
@amercader please review again.
I have made the requested changes:
For the UI I kept the functionality very simple and did not add an option to choose a date. I think as soon as you need more elaborate functionality you can still use the CLI.
Screenshot 2026-03-18 at 16 39 17

I also added:
- batched processing
- optional keep-argument for the cli

One cypress-test is still failing but I think that it is not this PR that introduced that error, but please let me know if I should have another look.

ckanext/activity/views.py
```diff
@@ -984,14 +983,62 @@ def display_name(followee: dict[str, Any]) -> Optional[str]:
     }
 
 
-@bp.route("/testing/dashboard")
```
amercader, line 987: @stefina the Cypress test is failing because you deleted this view function used in the tests. If you re-add it they should be green again

stefina: OMG, thank you for pointing that out, lost track after so many different changes 😅 
Please review @amercader

ckanext/activity/logic/action.py
```diff
@@ -663,3 +662,256 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
         "activities": activities,
     }
+
+
+def activity_delete_counts(context: Context, data_dict: DataDict) -> dict[str, Any]:
+    """
+    Returns counts of activities for the predefined delete options in the admin UI :
+    older than 1 day, 30 days, 365 days, and total.
+    """
+    tk.check_access("activity_delete", context, data_dict)
```
amercader: It's best if each action has a dedicated auth function, even if it replicates the logic of the `activity_delete` one

ckanext/activity/templates/admin/snippets/data_type.html
amercader: Do we need this template at all now that we have a dedicated tab?

ckanext/activity/logic/schema.py
```diff
@@ -98,3 +97,23 @@ def default_activity_list_schema(
...
+    int_validator: Validator,
+) -> Schema:
+    return {
+        "__before": [ensure_id_or_date_criteria_provided],
+        "id": [ignore_missing, ignore_empty, unicode_safe],
+        "start_date": [ignore_empty, isodate],
+        "end_date": [ignore_empty, isodate],
+        "offset_days": [ignore_empty, int_validator],
+        "keep": [ignore_empty, int_validator],
+        "batch_size": [ignore_empty, int_validator],
```
amercader: Is best to use `natural_number_validator` (if value can be 0 or higher) or `is_positive_integer` (if value can be 1 or higher), as `int_validator allows negative values

ckanext/activity/logic/action.py
```diff
@@ -663,3 +662,256 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
+    session = context["session"]
+    now = datetime.datetime.now()
```
amercader: The Activity model (and all other CKAN models) use [`utcnow()` ](https://github.com/ckan/ckan/blob/ec49c8b0bfaf242453140a6b9b05bc8cdab815a4/ckanext/activity/model/activity.py#L77) so make sure to use this to avoid mismatches depending on the server timezone

ckanext/activity/logic/action.py
```diff
@@ -663,3 +662,256 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
+    start_date: Optional[datetime.datetime],
+    end_date: Optional[datetime.datetime],
+    offset_days: Optional[int],
+):  # -> tuple[Query, None] | tuple[None, dict]
+    """
+    Build query for range/offset delete.
+    Returns (query, None) or (None, error_dict).
+    """
+    if offset_days:
+        threshold_date = datetime.datetime.now() - datetime.timedelta(
```
amercader:
```suggestion
        threshold_date = datetime.datetime.utcnow() - datetime.timedelta(
```

ckanext/activity/logic/action.py
```diff
@@ -663,3 +662,256 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
+    session: Any,
+    start_date: Optional[datetime.datetime],
+    end_date: Optional[datetime.datetime],
+    offset_days: Optional[int],
+):  # -> tuple[Query, None] | tuple[None, dict]
+    """
+    Build query for range/offset delete.
+    Returns (query, None) or (None, error_dict).
+    """
+    if offset_days:
```
amercader:
```suggestion
    if offset_days is not None:
```

Otherwise offset_days=0 is not matched

ckanext/activity/cli.py
```diff
@@ -0,0 +1,99 @@
...
+            ):
+                click.secho("Operation cancelled.", fg="yellow")
+                return
+
+        site_user = logic.get_action("get_site_user")(
+            {"ignore_auth": True}, {}
+        )
+        context: types.Context = {
+            "user": site_user["name"],
+            "defer_commit": not force,
```
amercader: Why does `defer_commit` depend on force? Shouldn't it always be `False` in this case?

ckanext/activity/views.py
```diff
@@ -984,6 +983,66 @@ def display_name(followee: dict[str, Any]) -> Optional[str]:
...
+    }
+    tk.check_access("sysadmin", context)
+
+
+# Predefined delete options: form action value -> offset_days (-1 = all)
+_ADMIN_DELETE_OPTIONS = {
+    "older_than_1_day": 1,
+    "older_than_30_days": 30,
+    "older_than_365_days": 365,
+    "all": -1,  # threshold = now + 1 day so every activity is older
```
amercader: This feels a bit like too much magic :) Better to have an explicit `delete_all` action in the form

ckanext/activity/logic/validators.py
```diff
@@ -98,3 +99,56 @@ def object_id_validator(
...
+    context: Context,
+) -> Any:
+    """Pass if id is provided, otherwise require date range or offset_days."""
+    activity_id = data.get(("id",))
+    if activity_id and str(activity_id).strip():
+        return
+    ensure_date_range_or_offset_provided(key, data, errors, context)
+
+
+def convert_yyyy_mm_dd_format(value: Any, context: Context) -> Any:
```
amercader: No longer needed

ckanext/activity/logic/action.py
```diff
@@ -663,3 +662,256 @@ def activity_diff(context: Context, data_dict: DataDict) -> dict[str, Any]:
...
+
+def _delete_activities_with_keep(
+    session: Any,
+    query: Any,
+    keep: int,
+    detail_table: Any,
+    commit: bool,
+) -> dict[str, Any]:
+    """Delete activities in range but keep N most recent per object_id."""
+    matching = query.all()
```
amercader: This will load all activities in Python, and for big instances this might mean running out of memory if the ranges are big. Also later on we are constructing an IN clause and we might hit limits there.

I believe the right approach is to let Postgres do all the work rather than Python using a [window function](https://en.wikipedia.org/wiki/Window_function_(SQL)), so we:

1. Rank all activities within each object_id ordered by timestamp (i.e. assign a number to it 1,2 ,3,...), 
2. We get the ids of the ones that have a higher rank than our keep value (.e.g if `keep` is 5, activities from position 6 onward)
3. We delete these records


@wardi does this sound right? I might need another pair of eyes here as you probably have more experience here


```diff
diff --git a/ckanext/activity/logic/action.py b/ckanext/activity/logic/action.py
index c39ffed835..97aa57db76 100644
--- a/ckanext/activity/logic/action.py
+++ b/ckanext/activity/logic/action.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import logging
 import datetime
 import json
-from collections import defaultdict
 from typing import Any, Optional
 
 import sqlalchemy as sa
@@ -766,28 +765,42 @@ def _delete_activities_with_keep(
     detail_table: Any,
     commit: bool,
 ) -> dict[str, Any]:
-    """Delete activities in range but keep N most recent per object_id."""
-    matching = query.all()
-    by_object: defaultdict[str, list[tuple[str, datetime.datetime]]] = (
-        defaultdict(list)
+    """Delete activities in range but keep N most recent per object_id.
+    """
+
+    # Order activities by timestamp within each object_id and rank them (i.e number them).
+    ranked = (
+        query
+        .with_entities(
+            model_activity.Activity.id,
+            sa.func.row_number()
+            .over(
+                partition_by=model_activity.Activity.object_id,
+                order_by=model_activity.Activity.timestamp.desc(),
+            )
+            .label("rank"),
+        )
+        .subquery("ranked")
     )
-    for match in matching:
-        timestamp = match.timestamp or datetime.datetime.min
-        by_object[match.object_id or ""].append((match.id, timestamp))
-    to_delete_ids: list[str] = []
-    for _object_id, act_list in by_object.items():
-        act_list.sort(key=lambda x: x[1], reverse=True)
-        to_delete_ids.extend(id_ for id_, _ts in act_list[keep:])
-    if not to_delete_ids:
-        return {"message": tk._("No activities found matching the specified criteria.")}
+
+    # Get the ids of activities where rank is bigger than keep
+    # (.e.g if `keep` is 5, activities from position 6 onward)
+    ids_to_delete = (
+        session.query(ranked.c.id)
+        .filter(ranked.c.rank > keep)
+    )
+
+    # Delete acivities
     session.query(model_activity.ActivityDetail).filter(
-        detail_table.c.activity_id.in_(to_delete_ids)
+        detail_table.c.activity_id.in_(ids_to_delete)
     ).delete(synchronize_session=False)
+
     deleted_count = (
         session.query(model_activity.Activity)
-        .filter(model_activity.Activity.id.in_(to_delete_ids))
+        .filter(model_activity.Activity.id.in_(ids_to_delete))
         .delete(synchronize_session=False)
     )
+
     if commit:
         session.commit()
     msg = tk._("Deleted {amount} rows from the activity table.").format(


```

ckanext/activity/templates/admin/snippets/data_type.html
stefina: You are right, I will delete this.

ckanext/activity/logic/action.py
amercader: cc @EricSoroos

wardi: ~Instead of using an `in_` filter on the delete call IIUC it's better to use the actual query that generates the ids~

Nevermind, that's not possible with window functions. @amercader what you're suggesting here LGTM

ckanext/activity/cli.py
stefina: You are right, this is wrong at this point so I will remove it.

ckanext/activity/logic/action.py
stefina: Thank you very much for your suggestion @amercader and your input @wardi 
I added the suggested changes, but added an early return in case nothing would be deleted (to keep the same response as before) and used coalesce on timestamp in the window ordering so NULLs are still treated as oldest, matching the previous implementation.
Let me know if you see any problems with that.

wardi: On second thought, a query like this could perform better:

```sql
with boundary as (
    select object_id as bid, nth_timestamp
    from (
        select distinct object_id, nth_value(timestamp, ?) over(
            partition by object_id order by timestamp desc) as nth_timestamp
        from activity
    ) t
    where nth_timestamp is not null
)
delete from activity where timestamp < (
    select nth_timestamp from boundary where object_id = bid
);
```

This way we avoid doing an `in` on a potentially huge temporary table of ids.

I'm not used to optimizing window functions, there might be an even simpler way to compute this `boundary` table.

stefina: Neither am I, unfortunately, but thanks a lot for both your reviews and suggestions. Very much appreciated!
Though, is it maybe possible to merge this PR as is and raise this with somebody who has more experience with this? Though I'd love to optimize it, I already spent more time on this than anticipated 😅

EricSoroos: Ok, so the goal is to keep the last n values for each object, and our schema looks like this:
``` 
ckan=> \d activity
                             Table "public.activity"
      Column       |            Type             | Collation | Nullable | Default 
-------------------+-----------------------------+-----------+----------+---------
 id                | text                        |           | not null | 
 timestamp         | timestamp without time zone |           |          | 
 user_id           | text                        |           |          | 
 object_id         | text                        |           |          | 
 revision_id       | text                        |           |          | 
 activity_type     | text                        |           |          | 
 data              | text                        |           |          | 
 permission_labels | text[]                      |           |          | 
Indexes:
    "activity_pkey" PRIMARY KEY, btree (id)
    "idx_activity_object_id" btree (object_id, "timestamp")
    "idx_activity_user_id" btree (user_id, "timestamp")
Referenced by:
    TABLE "activity_detail" CONSTRAINT "activity_detail_activity_id_fkey" FOREIGN KEY (activity_id) REFERENCES activity(id)
```

So the first thing to do is get an ordering of the activities, using timestamps per object. We've got a 2 column index there, so that should be already ordered. (on the other hand, I'm not seeing that in my current install, it's doing a quicksort)  Then we need the id and the row number from the first step, and delete anything with a row number higher than our cutoff. 

```
with _windowed as (
   select id, row_number() over (partition by object_id order by timestamp desc) as per_obj_order
   from activity
)
delete from activity 
  using _windowed
  where activity.id = _windowed.id
  and per_obj_order > ?
```

Using a select version of this query on a site with 33k activities, I get:

```
ckan=> explain analyze with _windowed as (
   select id, row_number() over (partition by object_id order by timestamp desc) as per_obj_order
   from activity
)
select * from activity inner join _windowed using(id) 
  where per_obj_order > 5

;
                                                                      QUERY PLAN                                                                       
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Join  (cost=5891.02..7978.21 rows=12043 width=368) (actual time=37.234..47.085 rows=35636 loops=1)
   Hash Cond: (activity.id = _windowed.id)
   ->  Seq Scan on activity  (cost=0.00..1831.28 rows=36128 width=360) (actual time=0.008..1.910 rows=36254 loops=1)
   ->  Hash  (cost=5740.48..5740.48 rows=12043 width=45) (actual time=37.209..37.210 rows=35636 loops=1)
         Buckets: 65536 (originally 16384)  Batches: 1 (originally 1)  Memory Usage: 3192kB
         ->  Subquery Scan on _windowed  (cost=4566.34..5740.48 rows=12043 width=45) (actual time=22.689..33.530 rows=35636 loops=1)
               Filter: (_windowed.per_obj_order > 5)
               Rows Removed by Filter: 618
               ->  WindowAgg  (cost=4566.34..5288.88 rows=36128 width=90) (actual time=22.683..31.337 rows=36254 loops=1)
                     ->  Sort  (cost=4566.32..4656.64 rows=36128 width=82) (actual time=22.679..25.018 rows=36254 loops=1)
                           Sort Key: activity_1.object_id, activity_1."timestamp" DESC
                           Sort Method: external merge  Disk: 3272kB
                           ->  Seq Scan on activity activity_1  (cost=0.00..1831.28 rows=36128 width=82) (actual time=0.005..4.661 rows=36254 loops=1)
 Planning Time: 0.271 ms
 Execution Time: 48.635 ms
(15 rows)
```

This compares with 33 ms for the inner, CTE query:

```
ckan=> explain analyze    select id, row_number() over (partition by object_id order by timestamp ) as per_obj_order
   from activity
;
                                                        QUERY PLAN                                                        
--------------------------------------------------------------------------------------------------------------------------
 WindowAgg  (cost=4566.34..5288.88 rows=36128 width=90) (actual time=23.737..31.808 rows=36254 loops=1)
   ->  Sort  (cost=4566.32..4656.64 rows=36128 width=82) (actual time=23.732..26.113 rows=36254 loops=1)
         Sort Key: object_id, "timestamp"
         Sort Method: external merge  Disk: 3272kB
         ->  Seq Scan on activity  (cost=0.00..1831.28 rows=36128 width=82) (actual time=0.048..4.931 rows=36254 loops=1)
 Planning Time: 0.094 ms
 Execution Time: 33.473 ms
(7 rows)
```

Running explain analyze on @wardi's query  I get a 60 second run time (56518.314 ms)  Now, I know that the select version is probably different than the delete as I'm reading in all of the table, but that's kind of crazy.

```
ckan=> explain analyze with boundary as (
    select object_id as bid, nth_timestamp
    from (
        select distinct object_id, nth_value(timestamp, 5) over(
            partition by object_id order by timestamp desc) as nth_timestamp
        from activity
    ) t
    where nth_timestamp is not null
)
select * from activity where timestamp < (
    select nth_timestamp from boundary where object_id = bid
);

                                                                                           QUERY PLAN                                                                                           
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on activity  (cost=0.00..4117362.29 rows=12043 width=360) (actual time=75.552..56514.775 rows=35636 loops=1)
   Filter: ("timestamp" < (SubPlan 1))
   Rows Removed by Filter: 618
   SubPlan 1
     ->  Subquery Scan on t  (cost=108.93..113.91 rows=248 width=8) (actual time=1.556..1.556 rows=1 loops=36254)
           Filter: (t.nth_timestamp IS NOT NULL)
           Rows Removed by Filter: 1
           ->  HashAggregate  (cost=108.93..111.42 rows=249 width=53) (actual time=1.555..1.556 rows=2 loops=36254)
                 Group Key: nth_value(activity_1."timestamp", 5) OVER (?)
                 Batches: 1  Memory Usage: 37kB
                 ->  WindowAgg  (cost=0.85..108.31 rows=249 width=53) (actual time=0.005..1.266 rows=4179 loops=36254)
                       ->  Index Only Scan Backward using idx_activity_object_id on activity activity_1  (cost=0.41..104.58 rows=249 width=45) (actual time=0.004..0.357 rows=4179 loops=36254)
                             Index Cond: (object_id = activity.object_id)
                             Heap Fetches: 10896269
 Planning Time: 0.144 ms
 JIT:
   Functions: 19
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 0.699 ms (Deform 0.222 ms), Inlining 8.511 ms, Optimization 39.738 ms, Emission 27.211 ms, Total 76.158 ms
 Execution Time: 56518.314 ms
(20 rows)
```

I don't normally see 3 order of magnitude speedups, but when I do, it's _always_ sql.

EricSoroos: And the query plan if I actually do delete is very similar to the select one: 
```
                                                                         QUERY PLAN                                                                          
-------------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on activity  (cost=5891.02..7978.21 rows=0 width=0) (actual time=94.953..94.956 rows=0 loops=1)
   ->  Hash Join  (cost=5891.02..7978.21 rows=12043 width=67) (actual time=42.567..54.094 rows=35636 loops=1)
         Hash Cond: (activity.id = _windowed.id)
         ->  Seq Scan on activity  (cost=0.00..1831.28 rows=36128 width=43) (actual time=0.035..3.904 rows=36254 loops=1)
         ->  Hash  (cost=5740.48..5740.48 rows=12043 width=98) (actual time=42.513..42.515 rows=35636 loops=1)
               Buckets: 65536 (originally 16384)  Batches: 1 (originally 1)  Memory Usage: 5315kB
               ->  Subquery Scan on _windowed  (cost=4566.34..5740.48 rows=12043 width=98) (actual time=25.271..38.541 rows=35636 loops=1)
                     Filter: (_windowed.per_obj_order > 5)
                     Rows Removed by Filter: 618
                     ->  WindowAgg  (cost=4566.34..5288.88 rows=36128 width=90) (actual time=25.261..34.340 rows=36254 loops=1)
                           ->  Sort  (cost=4566.32..4656.64 rows=36128 width=82) (actual time=25.256..27.817 rows=36254 loops=1)
                                 Sort Key: activity_1.object_id, activity_1."timestamp" DESC
                                 Sort Method: external merge  Disk: 3272kB
                                 ->  Seq Scan on activity activity_1  (cost=0.00..1831.28 rows=36128 width=82) (actual time=0.003..6.269 rows=36254 loops=1)
 Planning Time: 0.387 ms
 Trigger for constraint activity_detail_activity_id_fkey: time=254.942 calls=35636
 Execution Time: 351.875 ms
(17 rows)

```
Where the fkey constraint is actually the bulk of the difference in execution time.

wardi: Simpler, faster and easier to read too. Love it.

amercader: Many thanks for your time @stefina , I can take it from here, although I think we are mostly done.

@wardi @EricSoroos I believe I captured your suggestion to use a CTE in SQLAlchemy correctly in https://github.com/ckan/ckan/commit/fbfa0201c1 (I switched to SQLAlchemy 2.0 style `sa.select()`to make it easier to debug the generated statements). This is the generated SQL:

```
WITH windowed AS (
     SELECT filtered.id AS id, row_number() 
     OVER (
          PARTITION BY filtered.object_id
          ORDER BY coalesce(filtered.timestamp, :param_1) DESC
       ) AS per_obj_order
FROM (
     SELECT activity.id AS id, [...] 
     FROM activity
     WHERE activity.timestamp < :timestamp_1) AS filtered
)

DELETE FROM activity WHERE activity.id IN (
     SELECT windowed.id 
     FROM windowed
     WHERE windowed.per_obj_order > :per_obj_order_1
)
```

Does this look good to you?

wardi: @amercader does the EXPLAIN on your query look like Eric's even though it's using an IN and a subquery instead of a DELETE .. USING?

It's always an option to write this as plain SQL if the CTE is cramping your style. Although it seems like DELETE .. USING is supported https://docs.sqlalchemy.org/en/21/tutorial/data_update.html#multiple-table-deletes

wardi: @amercader switching to DELETE .. USING doesn't need to hold up this PR, what you're suggesting here LGTM

amercader:
@wardi @EricSoroos I finally used the raw sql statement which is much simpler and easier to read (https://github.com/ckan/ckan/commit/9333b373410af0c76d65e8ca822f1dfe025f4d4d)

Thanks again @stefina for your great contribution

Issue #9277, opened 2026-02-28
Organization List Page and Organization List Action Improvements


Engerrs:
Versions effected:
All (as the original code is pretty old)

While checking page loading time using FDT Profiler on the project, we noticed that Organization List Page loads pretty slow comparing to other pages.
After an investigation, found the actual issue within **get_packages_for_this_group** takes really the most of the time to actually retrieve Dataset counts using **package_search** action.

I prepared an solution that increases the Organization List loading time by **~33%** and Organization List Action **~50%**.

Here some tests that I've run to get those numbers
For Org List Page ~ 1000 Datasets and 41 Orgs
Before
~300-400ms

After
~200-300ms

For org list api ~ 1000 Datasets and 41 Orgs
Before
/api/action/organization_list render time 0.210 seconds

After 
/api/action/organization_list render time 0.095 seconds

I also increased the amount of Datasets to 5k and the situation haven't changed.

As for our project, this is even more faster, the loading of Organization List page dropped from ~800-900ms to ~430-550ms, which is an huge increase.

Need to take into account that for custom projects, the performance can be even higher, as for our project, due to the fact the **package_search** hooks are fired each time we call it, while the rows are 0, its still steal time, the more extensions have this hook, the slower it will be.

I will add PR to this issue with solution.

Engerrs:
Hi @wardi , I reworked the the approach to use permission labels to get the data and store them in context, so it will be taken from there instead of calling get_packages_for_this_group. Please review this one.

I can see even better performance then the originally mentioned in the original description.
For Org List Page ~ 1000 Datasets and 41 Orgs
Before
~300-400ms

After
~140-250ms

For org list api ~ 1000 Datasets and 41 Orgs
Before
/api/action/organization_list render time 0.210 seconds

After
/api/action/organization_list render time 0.075 seconds

Zharktas:
I've been dealing with problems concerning `group_list` and `organization_list` actions as well lately and I've been thinking that the whole action needs a rework when `all_fields` are called. The #9278 improves it, but as this bit of code https://github.com/ckan/ckan/blob/7a60264189f543fba04043d2193b09f545a6ef32/ckan/logic/action/get.py#L433-L440 just calls action functions in a loop, it really slows it down as it does auth checks and dictization for each group or organization individually. 

Since @Engerrs you have been going through this bit of code lately, any thoughts on this ?

wardi:
[...]

PR #9278, opened 2026-02-28
[#9277] Improve Organization List Action performance for Org List pages

Engerrs:
Fixes #

### Proposed fixes:
Change the logic, where the Dataset count logic happens by moving it from package_search into actual DB call.

### Features:

- [ x] includes user-visible changes
- [x ] includes API changes

### DIFF
--- a/changes/9278.misc
+++ b/changes/9278.misc
@@ -1,0 +1,1 @@
+Improved organization_list/group_list API performance by ~50% when returning all_fields, resulting in ~30% faster load times for Organization/Group List Pages.


ckan/logic/action/get.py
```diff
@@ -414,6 +416,31 @@ def _group_or_org_list(
...
+                    .filter(model.Member.table_id == user_obj.id) \
+                    .filter(model.Member.group_id.in_(group_ids)) \
+                    .filter(model.Member.state == 'active') \
+
+                memberships = [
+                    'member-' + membership.group_id for membership in memberships_query.all()]
+                labels = labels + memberships
+
+        counts = model_dictize.get_group_dataset_counts(
+            fq='entity_type:package', permissions_labels=labels)
```
wardi, line 426: Let's get the permission labels in an extension-compatible way:
https://github.com/ckan/ckan/blob/09ca35b6c7ed4290c0bc2dcbd55eacf8a5819eb2/ckan/logic/action/get.py#L1856-L1861

ckan/logic/action/get.py
```diff
@@ -414,6 +416,31 @@ def _group_or_org_list(
...
+
+        # won't call get_packages_for_this_group in group_dictize
+        context['dataset_counts'] = counts
```
wardi: modifying the passed context will leak this value to the caller. Something like this would avoid that problem:

```suggestion
        show_context = dict(context, dataset_counts=counts)
```

wardi:
Thanks @Engerrs, please add a changes file to document the performance improvement and I'll merge

Engerrs:
@wardi , added one, hope its correct, I wasn't sure on which number to add the issue 9277 or 9278 of the PR.

PR #9285, opened 2026-03-03
Make sure JS module system loads before the actual modules

smotornyuk:
Webassets with all JS modules have no dependency on webassets that define the module system. It means it's possible to break JS with a heavily customized script block that affects asset order.

By providing this dependency, we can guarantee that whenever any of the JS modules loads, the module system will be there already.

Additionally, the initialization logic of Bootstrap tooltips was moved to the section that depends on Bootstrap, to avoid a similar problem.


amercader:
If a view plugin or other extension JS module relies on bootstrap, this means that if targeting CKAN >= 2.12 you need to explicitly load the bootstrap bundle in your webassets definition, as described here:

https://github.com/ckan/ckan/wiki/CKAN-2.11-to-2.12-migration-tips#bootstrap-bundling

PR #9297, opened 2026-03-27
Escape entities when generating links in `link_to()` helper

amercader:
Remove `raw_dom_tags()` usage from the `link_to()` helper, which could led to potential XSS attacks.

> [!NOTE] 
> This vulnerability was not released in a public CKAN version, it was present only in the master branch

This involved two changes to keep previous behaviour:

* Refactor `_link_to` passed markup to be used as the body of the <a> tags generated by `link_to`. This was a problem
because the markup got (correctly) escaped and didn't render properly. Now the markup is generated using dominate tags in _link_to directly.
* Update `render_markdown()` tests to reflect that quotes are outputted as html entities now

Many thanks to @Shirshaw64p for responsively reporting the original vulnerability in the master branch.

To report security related issues please use security@ckan.org

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -864,24 +864,40 @@ def _link_to(text: str, *args: Any, **kwargs: Any) -> Markup:
             active = ''
         return kwargs.pop('class_', '') + active or None
 
-    def _create_link_text(text: str, **kwargs: Any):
-        ''' Update link text to add a icon or span if specified in the
-        kwargs '''
-        if kwargs.pop('inner_span', None):
-            text = literal('<span>') + text + literal('</span>')
-        if icon:
-            text = literal('<i class="fa fa-%s"></i> ' % icon) + text
-        return text
-
     icon = kwargs.pop('icon', None)
+    inner_span = kwargs.pop('inner_span', None)
     cls = _link_class(kwargs)
     title = kwargs.pop('title', kwargs.pop('title_', None))
-    return link_to(
-        _create_link_text(text, **kwargs),
-        url_for(*args, **kwargs),
-        cls=cls,
-        title=title
-    )
+
+    label = text
+    if inner_span:
+        label = dom_tags.span(text)
+    elif icon:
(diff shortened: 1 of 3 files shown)


ckan/lib/helpers.py
```diff
@@ -864,24 +863,40 @@ def _link_class(kwargs: dict[str, Any]):
...
+
+        out.add(icon, label)  # type: ignore
+
+    else:
+        out = dom_tags.a(
+            label,
+            href=url_for(*args, **kwargs),
+            cls=cls,
+            title=title
+        )
```
smotornyuk, line 897: Can simplify it to

```py
link = dom_tags.a(href=url_for(*args, **kwargs), cls=cls, title=title)

if icon:
    link.add(dom_tags.i(cls=f"fa fa-{icon}"))

link.add(dom_tags.span(text) if inner_span else f" {text}")
return literal(link)
```

ckan/tests/lib/test_helpers.py
```diff
@@ -21,6 +22,24 @@
...
+        <li>
+            <span>Test</span>
+        </li>
+    </ul>
+
+    becomes
+
+    <ul><li><span>Test</span></li></ul>
+    """
+    return re.sub(r'\n\s+', '', text)
```
wardi: This function is a little odd because it doesn't match HTML whitespace collapsing rules, e.g.
```python
_remove_whitespace("Hello\nworld") == "Helloworld"
_remove_whitespace("Hello    \nworld") == "Hello    world"
```
We could either remove this function and put the `re.sub` in the one test that uses it (I like this), or if we really need a reusable function at least make it follow the HTML collapsing rules, i.e.

```suggestion
    return re.sub(r'\s+', ' ', text)
```
where any amount of whitespace is reduced to one space
```python
_remove_whitespace("Hello    \n  world") == "Hello world"
```

amercader:
@Zharktas I've applied @smotornyuk and @wardi suggestions, this is ready for a final look

PR #9307, opened 2026-04-05
Run plugin migrations with disabled plugin

### LINKED ISSUE #9047: CKAN 2.11: Handling migrations for a plugin that extends another plugin
Is there a way to somehow tell Alembic to run the migrations for the parent of a plugin?

I am currently working on migrating a CKAN project to CKAN 2.11, and ran into this issue.

I have added [Alembic migrations to ckanext-subscribe](https://github.com/bellisk/ckanext-subscribe/tree/master/ckanext/subscribe/migration/subscribe). Upgrading the database with the plugin `subscribe` enabled works:

```
ckan-docker$ docker compose exec ckan ckan -c /srv/app/ckan.ini db upgrade
...
2 unapplied migrations for activity
2 unapplied migrations for harvest
1 unapplied migrations for subscribe
Upgrading DB: SUCCESS
```

However, in our project, we actually use a custom plugin that inherits from `subscribe`, `ogdch_subscribe` (https://github.com/opendata-swiss/ckanext-switzerland-ng/blob/553868d17d445cd876d4260c7d9fac4e042019b9/ckanext/switzerland/plugins.py#L647). With `ogdch_subscribe` enabled, the migrations for `subscribe` are not run:

```
ckan-docker$ docker compose exec ckan ckan -c /srv/app/ckan.ini db upgrade
...
2 unapplied migrations for activity
2 unapplied migrations for harvest
Upgrading DB: SUCCESS
```

Trying to run the migrations for `subscribe` (with `ckan db upgrade -p subscribe`) doesn't work, because it's not enabled.

Trying to run the migrations for `ogdch_subscribe` gets an Alembic error, because there are no Alembic migrations in the `ckanext-switzerland-ng` repo:

```
alembic.util.exc.CommandError: No 'script_location' key found in configuration.
```

The issue is also preventing running our tests. We have this fixture:

```
@pytest.fixture
def clean_db(reset_db, migrate_db_for):
    reset_db()
    migrate_db_for("subscribe")
    migrate_db_for("activity")
```

But, as above:
- Using the fixture in a test with the plugin `subscribe` enables runs the migrations, but the tests fail because they are testing functionality in `ogdch_subscribe`, not `subscribe`
- Using the fixture in a test with the with the plugin `ogdch_subscribe` enabled fails, because `subscribe` is not enabled.
- Changing the fixture to `migrate_db_for("ogdch_subscribe")` fails, because that plugin has no migrations.
- Using the fixture in a test with both `ogdch_subscribe` and `subscribe` enabled fails, because the two plugins clash.

For now I think I will just copy the migrations from `ckanext-subscribe` into `ckanext-switzerland-ng`, but this feels like a hack. If there's a better way to handle this I'd love to learn it!

smotornyuk:
Fixes #9047

Allow applying DB migrations of disabled plugins via providing `--alembic-ini path/to/alembic/configuration/of/plugin`.


wardi:
consider using the plugin name instead of path to alembic ini

smotornyuk:
Instead of `--alembic-ini`, command accepts `--load-plugin` flag. It must be used only with disabled plugins - in this case they are temporarily loaded before applying migrations.

PR #9308, opened 2026-04-05
Add `nullable` flag to config declarations

### LINKED ISSUE #8997: Config with type bool will always be true or false, even if its left out of configuration
## CKAN version

master

## Describe the bug

While implementing tests on https://github.com/ckan/ckan/pull/8977, I stumbled on that if you define config with type bool and without default, it will always be false if its not actually configured.

### Steps to reproduce

Create config declaration with type bool and without defaut.
Do not define it in local .ini-file
Fetch the variable in code with `config.get(some_variable)`, it will be false.

### Expected behavior
If the default is not defined, it should be None if its missing from .ini-file

smotornyuk:
Fixes #8997

Config declaration can use the `nullable: true` flag. When it's enabled, if option **is not specified in the config file at all** it will have the value `None`, no matter what type it has. If the option is present in the config, just have empty value, it still runs through the whole validation chain.


wardi:
let's add a `null_value: null` as a default that can be overridden on a per-config option basis

also, please no nullable text fields

smotornyuk:
Added configurable `null_value`. 
As there is no `str` type for options yet(we support `int`, `bool` and `list` at the moment and everything else keeps existing type which is usually string but heavily depends on validators), I'm not denying nullable options of other type. This may be addressed in future, when I finally prepare PR with more option types, like `str`, `date`, `enum`, etc

Zharktas (approved):

PR #9318, opened 2026-04-20
Support keep_email in user_list_dictize

### LINKED ISSUE #9317: user_list_dictize does not support keep_email
## CKAN version
2.11.4

## Describe the bug
`user_dictize` supports a context variable `keep_email`, but `user_list_dictize` does not, removing email addresses that may have been retained by `user_dictize`.

### Steps to reproduce
1. Call `user_list_dictize` with `keep_email=True` in context

### Expected behavior
Users in the result have email addresses just as when calling `user_dictize`

### Additional details
This is needed for example when building an organization member view that should show member email addresses to avoid querying each user individually

bzar:
Fixes #9317

### Proposed fixes:
- Add support for `keep_email` context value in `user_list_dictize`

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9317.bugfix
+++ b/changes/9317.bugfix
@@ -1,0 +1,1 @@
+Add `keep_email` support to `user_list_dictize`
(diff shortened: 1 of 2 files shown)


bzar:
There are no tests to improve for internal APIs like `user_list_dictize` and this change won't be visible in any public API as far as I can tell, only extensions overriding default functionality.

How should I approach testing this?

PR #9324, opened 2026-04-24
Null chars

avdata99:
We get 500 error in the `user.reset` URL from user using the NUL byte `\x00`.
SQLAlchemy/psycopg2 rejects this character because PSQL cannot store them.

<pre>
File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
ValueError: <b>A string literal cannot contain NUL (0x00) characters.</b>
</pre>

With this test we captured a 500 error in the `user.reset` URL

<pre>
====================================================================================== FAILURES ======================================================================================
__________________________________________________________________ TestUser.test_request_reset_with_nul_in_username __________________________________________________________________

self = ckan.tests.controllers.test_user.TestUser object at 0x70be441133d0, send_reset_link = MagicMock name='send_reset_link' id='123962482955408'
app = ckan.tests.helpers.CKANTestApp object at 0x70be46313210

    @mock.patch("ckan.lib.mailer.send_reset_link")
    def test_request_reset_with_nul_in_username(self, send_reset_link, app):
        """Regression: a NUL byte (U+0000) in the ``user`` form field must
        not 500. PostgreSQL rejects NUL in text columns and psycopg2 raises
        ValueError before the query is sent, so unguarded input used to
        crash the reset form with an Internal Server Error.
    
        Malformed input is surfaced with a flash error and a redirect back
        to the form (same pattern the view uses for empty input or a bad
        captcha), not a 500 and not a silent success.
        """
        offset = url_for("user.request_reset")
        response = app.post(offset, data=dict(user="foo\x00bar"))
    
>       assert response.status_code == 200
E       assert 500 == 200
E        +  where 500 = WrapperTestResponse streamed [500 INTERNAL SERVER ERROR].status_code

ckan/tests/controllers/test_user.py:677: AssertionError
</pre>

After applying the validator the test is green.

We probably will need to implement this validator in other fields.
It look like we have something similar for SOLR at the `_illegal_xml_chars_re` regex at ckan/lib/search/index.py to avoid invalid solr characters

Also: a _too complex_ error were found so we needed to split the user view function

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -367,27 +367,6 @@ def group_id_or_name_exists(reference: str, context: Context) -> Any:
     return reference
 
 
-def no_nul_byte(value: Any, context: Context) -> Any:
-    '''Reject strings containing the NUL character (U+0000).
-
-    PostgreSQL ``text`` columns cannot store the NUL byte and ``psycopg2``
-    raises ``ValueError`` when binding a NUL-bearing string as a query
-    parameter. Without this validator, a stray NUL in user-supplied input
-    bubbles up as a 500 instead of a normal validation error.
-
-    Other illegal-XML characters (most C0 controls, Unicode surrogates,
-    non-characters) are accepted by PostgreSQL and are stripped at Solr
-    indexing time only -- see ``ckan.lib.search.index._illegal_xml_chars_re``.
-
-    :raises ckan.lib.navl.dictization_functions.Invalid: if ``value``
-        contains the NUL character
-
-    '''
-    if isinstance(value, str) and '\x00' in value:
-        raise Invalid(_('Nul bytes are not allowed.'))
-    return value
-
-
 name_match = re.compile(r'[a-z0-9_\-]*$')
 def name_validator(value: Any, context: Context) -> Any:
     '''Return the given value if it's a valid name, otherwise raise Invalid.
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -367,6 +367,27 @@ def group_id_or_name_exists(reference: str, context: Context) -> Any:
     return reference
 
 
+def no_nul_characters(value: Any, context: Context) -> Any:
+    '''Reject strings containing the NUL character (U+0000).
+
+    PostgreSQL ``text`` columns cannot store the NUL byte and ``psycopg2``
+    raises ``ValueError`` when binding a NUL-bearing string as a query
+    parameter. Without this validator, a stray NUL in user-supplied input
+    bubbles up as a 500 instead of a normal validation error.
+
+    Other illegal-XML characters (most C0 controls, Unicode surrogates,
+    non-characters) are accepted by PostgreSQL and are stripped at Solr
+    indexing time only -- see ``ckan.lib.search.index._illegal_xml_chars_re``.
+
+    :raises ckan.lib.navl.dictization_functions.Invalid: if ``value``
+        contains the NUL character
+
+    '''
+    if isinstance(value, str) and '\x00' in value:
+        raise Invalid(_('Null characters are not allowed.'))
+    return value
+
+
 name_match = re.compile(r'[a-z0-9_\-]*$')
 def name_validator(value: Any, context: Context) -> Any:
     '''Return the given value if it's a valid name, otherwise raise Invalid.
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -384,7 +384,7 @@ def no_nul_byte(value: Any, context: Context) -> Any:
 
     '''
     if isinstance(value, str) and '\x00' in value:
-        raise Invalid(_('Null characters are not allowed.'))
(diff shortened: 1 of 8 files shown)


ckan/views/user.py
```diff
@@ -670,26 +674,60 @@ def post(self) -> Response:
             h.flash_error(error_msg)
             return h.redirect_to(u'user.request_reset')
 
-        id = request.form.get(u'user', '')
-        if id in (None, u''):
-            h.flash_error(_(u'Email is required'))
-            return h.redirect_to(u'user.request_reset')
-        log.info('Password reset requested for user %s', repr_untrusted(id))
+        id_ = request.form.get(u'user', '')
```
wardi: if we're cleaning up unrelated code, might as well remove `u` prefixes from literal strings as well

ckan/views/user.py
```diff
@@ -49,6 +49,10 @@ def _perform_reset_form_to_db_schema() -> Schema:
     return schema.user_perform_reset_form_schema()
 
 
+def _request_reset_form_schema() -> Schema:
+    return schema.request_reset_form_schema()
```
wardi, line 53: does this function need to exist?

ckan/tests/logic/test_validators.py
```diff
@@ -356,6 +356,54 @@ def call_validator(*args, **kwargs):
...
+        "emoji 🎉",
+    ]
+
+    for valid_value in valid_values:
+
+        @returns_arg
+        def call_validator(*args, **kwargs):
+            return validators.no_nul_characters(*args, **kwargs)
+
+        call_validator(valid_value)
```
wardi:
```suggestion
        returns_arg(validators.no_nul_characters)(valid_value)
```

ckan/tests/logic/test_validators.py
```diff
@@ -356,6 +356,54 @@ def call_validator(*args, **kwargs):
...
+        "\x00\x00\x00",
+    ]
+
+    for invalid_value in invalid_values:
+
+        @raises_invalid
+        def call_validator(*args, **kwargs):
+            return validators.no_nul_characters(*args, **kwargs)
+
+        call_validator(invalid_value, context={})
```
wardi:
```suggestion
        raises_invalid(validators.no_nul_characters)(invalid_value, context={})
```

ckan/logic/validators.py
```diff
@@ -367,6 +367,27 @@ def group_id_or_name_exists(reference: str, context: Context) -> Any:
     return reference
 
 
+def no_nul_characters(value: Any, context: Context) -> Any:
```
wardi: It's confusing that the error calls them "Null characters" the validator is named "no_nul_characters" and the docstring calls them NUL bytes. Can we settle on one notation?

wardi (changes requested):
See my comments above, and I'm wondering if we need a more general fix.

@ckan/core 
There are lots of other user-supplied strings that could trigger a server error for the same reason. Should we:

1. sprinkle this new validator across all the schemas,
2. create a new validator for string values stored as text in the db, or
3. add this checking to an existing one like `unicode_safe`

EricSoroos:
Null chars aren't unicode safe, so I'd say that's a good place for it.

ckan/views/user.py
avdata99: You're right. Done here: https://github.com/avdata99/ckan/commit/3f815d759f8701d98ff1bbf5776f824062d4ed1c

ckan/views/user.py
avdata99: We are just following old code style for this
image

ckan/logic/validators.py
avdata99: All changed to nul byte
https://github.com/ckan/ckan/pull/9324/changes/558dbcd305e20e8af95183c7597eeef0f940eae5 + https://github.com/ckan/ckan/pull/9324/changes/26acd3d8453c8fb7e5ed87c5dcffd73dd8c63ebb

wardi (changes requested):
Thanks @avdata99 , as @EricSoroos agreed the better place for this change is in `unicode_safe` instead of as a new validator.

In `unicode_safe` we can silently strip all NUL bytes and then this fix will automatically apply to every schema that already uses that validator.

avdata99:
Thanks, is there any other PR with this change for `unicode_safe`?
Should I update this PR to include this validation `unicode_safe`?

wardi:
@avdata99 there is no other PR. Yes, in this PR I would like you to remove the new validator and instead strip NUL bytes from all strings that go through the `unicode_safe` validator.

avdata99:
This proposal was updated to move the invalid chars to the `unicode_safe` validator @wardi

PR #9329, opened 2026-05-03
Update CodeQL actions

amercader:
* Ignore files in the cookiecutter template as syntax errors are raised
  when trying to analyze them
* Bump action versions and add security param to checkout
* Test release branches


amercader:
Part of #9332

PR #9330, opened 2026-05-03
Idea: Page layout

smotornyuk:
Add an intermediate template between `page.html` and the final template rendered by the view. This new template can be used to modify page layout like this:

```django
{% block main_content %}
 
    {%- if _layout == "no-sidebar" -%}
      {{ self.primary() }}

    {%- elif _layout == "sidebar-end" -%}
      {{ self.primary() }}
      {{ self.secondary() }}

    {%- else -%} # the default layout if `_layout` is not specified or has unknown value
      {{ self.secondary() }}
      {{ self.primary() }}

    {%- endif %}

{% endblock %}
```

This approach significantly simplifies swapping the sidebar and main content area, moving the sidebar to the top/bottom of the page, or completely hiding it. Further, it can be used to add banners on a specific group of pages(for example, a banner with the organization name and logo for every page showing datasets of a specific organization).

The new intermediate template is required because Jinja2 does not allow the creation of block duplicates or condition branches. For example, the code snippet above will not work if you rewrite it like this:

```django
{% block main_content %}
 
    {%- if _layout == "no-sidebar" -%}
      {%block primary%}{%endblock%}

    {%- elif _layout == "sidebar-end" -%}
      {%block primary%}{%endblock%}
      {%block secondary%}{%endblock%}

    {%- else -%} # the default layout if `_layout` is not specified or has unknown value
      {%block secondary%}{%endblock%}
      {%block primary%}{%endblock%}

    {%- endif %}

{% endblock %}
```

Because of Jinja2 restrictions, the base template must define all blocks, while its descendant moves **existing** blocks around.

Current PR is compatible with existing templates - by default, it keeps existing behavior. When it sees `_layout` variables inside the template, it starts changing the template: try adding `{% set _layout = 'no-sidebar' %}` (or `sidebar-end`) in the `package/search.html` to see the difference. Implemented layouts are extremely straightforward, and mainly they serve as an example for extensions.

To keep this new feature compatible with existing code, I moved `page.html` into the new `layout.html` and replaced the content of `page.html` with a layout management example. This updated `page.html` extends the new `layout.html`, so any further extensions of the template and block overrides will work, and there will be absolutely no difference. Because of reusing `page.html`, I don't have to update all existing templates that extend `page.html`.

---

There are a few CSS modifications: I removed custom CSS and replaced it with native Bootstrap 5 functionality.

### DIFF
--- a/ckan/public-midnight-blue/base/scss/_ckan-rtl.scss
+++ b/ckan/public-midnight-blue/base/scss/_ckan-rtl.scss
@@ -1,14 +1,4 @@
 @use "sass:math";
-.wrapper {
-  @include media-breakpoint-up(sm) {
-    &:before {
-      left: initial;
-      right: 0;
-      width: 25%;
-      border-left: 1px solid $genericBorderColor;
-    }
-  }
-}
 
 .simple-input .field .btn-search {
   right: initial;
--- a/ckan/public-midnight-blue/base/scss/_layout.scss
+++ b/ckan/public-midnight-blue/base/scss/_layout.scss
@@ -1,25 +1,11 @@
-[role=main],
-.main {
-    position: relative;
-}
-
 .main {
     padding: 20px 0;
     background-color: $white;
 }
 
-.main:after,
-[role=main]:after {
-    bottom: 0;
-    border-top-width: 1px;
-}
-
 .main .primary {
     border-radius: 20px;
     background-color: $layoutPrimaryBackgroundColor;
-    position: relative;
-    float: right;
-    margin-left: 0; // Remove grid margin.
 }
 
 .main .secondary {
--- a/ckan/public-midnight-blue/base/scss/_module.scss
+++ b/ckan/public-midnight-blue/base/scss/_module.scss
@@ -218,10 +218,6 @@
     padding-bottom: 8px;
 }
 
-.no-nav .module:last-child {
-    margin-top: 0;
-}
-
 .module-image {
     float: left;
     width: 50px;
--- a/ckan/public-midnight-blue/base/scss/_toolbar.scss
+++ b/ckan/public-midnight-blue/base/scss/_toolbar.scss
@@ -110,10 +110,6 @@
     justify-content: flex-end;
 }
 
-.no-nav .page-header {
-    @include border-radius(3px 3px 0 0);
-}
-
 .nav-tabs-plain {
     padding: 0 $gutterX;
     &>.active>a,
--- a/ckan/public/base/scss/_layout.scss
+++ b/ckan/public/base/scss/_layout.scss
@@ -5,41 +5,10 @@
     min-height: 300px;
     background-color: #fff;
     @include media-breakpoint-up(md) {
-        &:before {
-            content: '';
-            display: block;
-            position: absolute;
-            top: 0;
-            bottom: 0;
-            left: 0;
-            width: 25%;
-            border-right: 1px solid $genericBorderColor;
-            z-index: 1; // Fixes overlapping .secondary on the Resource view page
-        }
-        &.no-nav {
-            &:before {
-                display: none;
-            }
-            .module {
-                >.page-header {
-                    margin-top: 0;
-                }
-            }
+        .primary + .secondary, .secondary + .primary {
+            border-inline-start: 1px solid $genericBorderColor;
         }
     }
-}
-
-// @media (min-width: $screen-md-min) {
-//   .wrapper {
-//     background-position:0px 0px;
-//   }
-// }
-// .wrapper.no-nav {
-//     background-image: none;
-// }
-[role=main],
-.main {
-    position: relative;
 }
 
 .main {
@@ -47,22 +16,9 @@
     background: $layoutBackgroundColor url("#{$bgPath}");
 }
 
-.main:after,
-[role=main]:after {
-    bottom: 0;
-    border-top-width: 1px;
-}
-
-.main .primary {
-    position: relative;
-    float: right;
-    margin-left: 0; // Remove grid margin.
-}
-
 .main .secondary {
     padding: 0;
     z-index: 1;
-    padding-right: 1px;
 }
 
 
--- a/ckan/templates-midnight-blue/layout.html
+++ b/ckan/templates-midnight-blue/layout.html
@@ -1,0 +1,134 @@
+{% extends "base.html" %}
+
+{%- block page -%}
+
+  {% block skip %}
+    <div class="visually-hidden-focusable"><a href="#content">{{ _('Skip to main content') }}</a></div>
+  {% endblock %}
+
+  {#
+  Override the header on a page by page basis by extending this block. If
+  making sitewide header changes it is preferable to override the header.html
+  file.
+  #}
+  {%- block header %}
+    {% include "header.html" %}
+  {% endblock -%}
+
+  {# The content block allows you to replace the content of the page if needed #}
+  {%- block content %}
+    {% block maintag %}<div class="main">{% endblock %}
+      <div id="content" class="container">
+        {% block main_content %}
+          {% block flash %}
+            <div class="flash-messages">
+              {% block flash_inner %}
+                {% for category, message in h.get_flashed_messages(with_categories=true) %}
+                  <div class="alert alert-dismissible fade show {{ category }}" role="alert">
+                    {{ h.literal(message) }}
+                  </div>
+                {% endfor %}
+              {% endblock %}
+            </div>
+          {% endblock %}
+
+          {% block toolbar %}
+            <div class="toolbar" role="navigation" aria-label="{{ _('Breadcrumb') }}">
+              {% block breadcrumb %}
+                {% if self.breadcrumb_content() | trim %}
+                  <ol class="breadcrumb">
+                    {% snippet 'snippets/home_breadcrumb_item.html' %}
+                    {% block breadcrumb_content %}{% endblock %}
+                  </ol>
+                {% endif %}
+              {% endblock %}
+            </div>
+          {% endblock %}
+
+          <div class="row{% block wrapper_class %}{% endblock %}">
+            {#
+            The pre_primary block can be used to add content to before the
+            rendering of the main content columns of the page.
+            #}
+            {% block pre_primary %}
+              {% if self.content_action() | trim %}
+                <div class="content_action gap-1 mb-md-3">
+                  {% block content_action %}{% endblock %}
+                </div>
+              {% endif %}
+            {% endblock %}
+
+            {% block secondary %}
+              <aside class="secondary col-md-4 col-lg-3">
+                {#
+                The secondary_content block can be used to add content to the
+                sidebar of the page. This is the main block that is likely to be
+                used within a template.
+
+                Example:
+
+                  {% block secondary_content %}
+                    <h2>A sidebar item</h2>
+                    <p>Some content for the item</p>
+                  {% endblock %}
+                #}
+                {% block secondary_content %}{% endblock %}
+              </aside>
+            {% endblock %}
+
+            {% block primary %}
+              <div class="primary col" role="main">
+                {#
+                The primary_content block can be used to add content to the page.
+                This is the main block that is likely to be used within a template.
+
+                Example:
+
+                  {% block primary_content %}
+                    <h1>My page content</h1>
+                    <p>Some content for the page</p>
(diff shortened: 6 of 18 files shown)


ckan/public-midnight-blue/base/scss/_ckan-rtl.scss
```diff
@@ -1,14 +1,4 @@
...
-.wrapper {
-  @include media-breakpoint-up(sm) {
-    &:before {
-      left: initial;
-      right: 0;
-      width: 25%;
-      border-left: 1px solid $genericBorderColor;
-    }
-  }
-}
```
smotornyuk, line 11: This is not required for RTL languages, because I'm using the `border-inline-start` CSS property, which works as `border-left` for LtR, and as `border-right` for RtL

ckan/public-midnight-blue/base/scss/_layout.scss
```diff
@@ -1,25 +1,11 @@
-[role=main],
-.main {
-    position: relative;
-}
-
```
smotornyuk, line 5: Do not see how this position affects the template, looks like leftower from BS3

ckan/public-midnight-blue/base/scss/_layout.scss
```diff
@@ -1,25 +1,11 @@
...
 .main {
     padding: 20px 0;
     background-color: $white;
 }
 
-.main:after,
-[role=main]:after {
-    bottom: 0;
-    border-top-width: 1px;
-}
```
smotornyuk, line 15: It has no effect in the current theme, most likely leftovers from the BS3

ckan/public-midnight-blue/base/scss/_layout.scss
```diff
@@ -1,25 +1,11 @@
...
-
 .main .primary {
     border-radius: 20px;
     background-color: $layoutPrimaryBackgroundColor;
-    position: relative;
-    float: right;
-    margin-left: 0; // Remove grid margin.
```
smotornyuk, line 22: We are using BS5 grid here, which relies on flexbox. Using `float` only complicates things.

ckan/public-midnight-blue/base/scss/_module.scss
```diff
@@ -218,10 +218,6 @@
     padding-bottom: 8px;
 }
 
-.no-nav .module:last-child {
-    margin-top: 0;
-}
```
smotornyuk, line 223: Do not see real use of this in templates, looks like it wasn't deleted during migration from BS3

ckan/public-midnight-blue/base/scss/_toolbar.scss
```diff
@@ -110,10 +110,6 @@
     justify-content: flex-end;
 }
 
-.no-nav .page-header {
-    @include border-radius(3px 3px 0 0);
-}
-
```
smotornyuk, line 116: In midnight-blue it has no sense, in classic theme I updated selectior so that it does not rely on `.no-nav` class

ckan/public/base/scss/_layout.scss
```diff
@@ -5,64 +5,20 @@
...
-            width: 25%;
-            border-right: 1px solid $genericBorderColor;
-            z-index: 1; // Fixes overlapping .secondary on the Resource view page
-        }
-        &.no-nav {
-            &:before {
-                display: none;
-            }
-            .module {
-                >.page-header {
```
smotornyuk, line 24: This already has the same value set in _module.scss

ckan/public/base/scss/_layout.scss
```diff
@@ -5,64 +5,20 @@
...
-            position: absolute;
-            top: 0;
-            bottom: 0;
-            left: 0;
-            width: 25%;
-            border-right: 1px solid $genericBorderColor;
-            z-index: 1; // Fixes overlapping .secondary on the Resource view page
-        }
-        &.no-nav {
-            &:before {
```
smotornyuk, line 20: No need in this after switching to real border.

ckan/public/base/scss/_layout.scss
```diff
@@ -5,64 +5,20 @@
     min-height: 300px;
     background-color: #fff;
     @include media-breakpoint-up(md) {
-        &:before {
```
smotornyuk, line 8: Instead of using pseudo-selector, I added real border between `.primary` and `.secondary`

ckan/public/base/scss/_layout.scss
```diff
@@ -5,64 +5,20 @@
...
-.main .primary {
-    position: relative;
-    float: right;
-    margin-left: 0; // Remove grid margin.
-}
-
 .main .secondary {
     padding: 0;
     z-index: 1;
-    padding-right: 1px;
```
smotornyuk, line 65: This was required by `:before`, now with real border between elements, it has no sense

ckan/templates-midnight-blue/layout.html
```diff
@@ -0,0 +1,134 @@
+{% extends "base.html" %}
```
smotornyuk, line 1: This file is almost the same as an old `page.html`. I added comments below for lines that were modified

ckan/templates-midnight-blue/layout.html
```diff
@@ -0,0 +1,134 @@
...
+                  <ol class="breadcrumb">
+                    {% snippet 'snippets/home_breadcrumb_item.html' %}
+                    {% block breadcrumb_content %}{% endblock %}
+                  </ol>
+                {% endif %}
+              {% endblock %}
+            </div>
+          {% endblock %}
+
+          <div class="row {% block wrapper_class %}{% endblock %}">
```
smotornyuk, line 48: logic for `no-nav` class is updated and moved to the new `page.html`.

ckan/templates-midnight-blue/layout.html
```diff
@@ -0,0 +1,134 @@
...
+                    <h2>A sidebar item</h2>
+                    <p>Some content for the item</p>
+                  {% endblock %}
+                #}
+                {% block secondary_content %}{% endblock %}
+              </aside>
+            {% endblock %}
+
+            {% block primary %}
+              <div class="primary col" role="main">
```
smotornyuk, line 80: `col-md-8 col-lg-9 col-xs-12` is replaced with a single `col` class. BS5 uses the flexbox model, and this block automatically adapts to the width of `.secondary`

ckan/templates-midnight-blue/page.html
```diff
@@ -1,134 +1,24 @@
-{% extends "base.html" %}
+{% extends "layout.html" %}
```
smotornyuk, line 1: Content of this file was moved into `layout.html` and now we extend it. In this way every template that extends `page.html` will see no difference with old implementation.

ckan/templates-midnight-blue/page.html
```diff
@@ -1,134 +1,24 @@
...
+{%- set _layout = _layout | default("sidebar-start" if self.secondary()|trim else "no-sidebar") -%}
```
smotornyuk, line 2: To allow changing layout in descendant templates, always use this form 
```django
{% set _layout = _layout | default(MY_PREFERED_LAYOUT) %}
```

As long as you do this, any descendant of the `page.html` will be able to change the layout. Because of the template processing order, the following snippet **does not work**  and hardcodes the layout for every descendant.  
```django
{% set _layout = MY_PREFERED_LAYOUT %}
```

ckan/templates-midnight-blue/page.html
```diff
@@ -1,134 +1,24 @@
...
 
-{%- block page -%}
+{% block main_content %}
+  {{ self.flash() }}
+  {{ self.toolbar() }}
```
smotornyuk, line 6: We can use `block toolbar %} {{super()}}`, but it will cause problems if different layouts need to show the same block in different places. `self.BLOCK_NAME()` form has no restrictions and does not breaks inheritance, so it's much better choice here, as long as we only want to move blocks around, without any changes in their content.

ckan/templates-midnight-blue/page.html
```diff
@@ -1,134 +1,24 @@
...
 
-  {% block skip %}
-    <div class="visually-hidden-focusable"><a href="#content">{{ _('Skip to main content') }}</a></div>
-  {% endblock %}
+  <div class="row ckan-layout--{{ _layout }} {{ self.wrapper_class() }}">
```
smotornyuk, line 8: Add CSS class with chosen layout. From my experience, wrapper classes can be omitted almost all the time. But if you are not working with CSS frequently, such a class can simplify your life.

ckan/templates-midnight-blue/page.html
```diff
@@ -1,134 +1,24 @@
...
 
-  {#
-  Override the header on a page by page basis by extending this block. If
-  making sitewide header changes it is preferable to override the header.html
-  file.
-  #}
-  {%- block header %}
-    {% include "header.html" %}
-  {% endblock -%}
+    {%- if _layout == "no-sidebar" -%}
```
smotornyuk, line 11: These conditions are primarily serving as an example of layout definition. No logic in CKAN relies on these values. In the future, we can use layouts to avoid checking `secondary() | trim` when deciding whether we need a sidebar or not. For now, I'm not changing it to keep compatibility with existing customizations.

ckan/templates-midnight-blue/page.html
```diff
@@ -1,134 +1,24 @@
...
-              {% block flash_inner %}
-                {% for category, message in h.get_flashed_messages(with_categories=true) %}
-                  <div class="alert alert-dismissible fade show {{ category }}" role="alert">
-                    {{ h.literal(message) }}
-                  </div>
-                {% endfor %}
-              {% endblock %}
-            </div>
-          {% endblock %}
+    {%- elif _layout == "sidebar-end" -%}
```
smotornyuk, line 14: `sidebar-end`/`sidebar-start` are better names than `sidebar-right`/`sidebar-left`, because UI is mirrored for RTL languages.

smotornyuk:
Initial suggestion relies on `_layout` variable in defined in the template. This may be tricky because of the template inheritance order. Imagine you have `package/base.html` that sets layout to `A` and it's extended by `package/search.html` that sets layout to `B`. Jinja2 moves from child to parent when rendering templates, so initially, the layout will be set to `B` and then overridden in the parent template to `A`. Because of this, with variables, we must always use the `default` filter and set the template **only if it has no value yet**. In this way, when a child template specifies a layout, the parent template will not override it:
```django
# parent.html
  set layout only if it was not yet defined by the child
{% set _layout = _layout | default('no-sidebar') %}

# child
  We also have to use `default` here in case somebody extends the child template
{% set _layout = _layout | default('sidebar-start') %}

```

---


Instead of a variable, we can use a helper call. This does not look right, but it's an option:
```django
{% do h.set_page_layout('no-sidebar') %}
```

This approach has the same problem - templates are processed from child to parent. But assuming that the helper maintains the state of the layout via a global variable, it internally can set the layout only if it wasn't set already(presumably, by a child template)
```py
def set_page_layout(layout: str):
    if g._layout:
        return
    g._layout = layout
```

---

The other option is to use blocks. It may be more verbose than variables, but it fits naturally into the child-parent relationship:

```django
# parent
{% block layout %}no-sidebar{% endblock %}

# child
{% block layout %}sidebar-start{% endblock %}
```

In this case, `page.html` can set the value of a variable via set-block. It may look unusual, but it's correct Jinja2 construction:

```django
# page.html
  Because the value of the variable set via the block, it can be replaced in child templates

{% set _layout | trim %}
  {% block layout %}default-layout{% endblock %}
{% endset %}
```

---

Alternatively, we can modify `extends` and `ckan_extends` expressions and pass the layout there:

```
{% ckan_extends layout="no-sidebar" %}

{% extends "page.html", layout="sidebar-start" %}
```

But this approach has a risk for long template chains. If any of the templates in the chain do not pass the layout value to the extended template, layout customization will fail.

wardi:
I like the block because it's clearly communicating with a parent template.

ckan/templates-midnight-blue/page.html
wardi: Consider `sidebar-first` / `sidebar-second`, somehow `start` and `end` make me think of the top and bottom of the page

ckan/templates-midnight-blue/page.html
wardi: If we store the `secondary()` value in layout.html then we wouldn't have to render it twice, right?

PR #9333, opened 2026-05-04
Fixes for setuptools>=82

### LINKED ISSUE #9246: CKAN broken with latest setuptools version (setuptools>=82)
This affects all versions (CKAN 2.10, 2.11 and master/2.12)

Image

The writing has been on the wall (logs) for a while:

```
ckan
/home/adria/.pyenv/versions/3.9.21/envs/ckan-211/lib/python3.9/site-packages/pysolr.py:14: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as
 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import DistributionNotFound, get_distribution, parse_version
```

Using setuptools 82 (latest version published to pypi) prevents CKAN to start. The actual exception varies across CKAN versions but the bottom line is that the `pkg_resources` module is no longer available and something tries to access it, either CKAN core, a CKAN extension or a third-party package. 

```
   ckanext/stats/__init__.py:3: in <module>
      __import__('pkg_resources').declare_namespace(__name__)
  E   ModuleNotFoundError: No module named 'pkg_resources'
```

We've replaced some of the `pkg_resources` functionality regarding entry points in #8992 but the one that remains is a more tricky one, namespace registration (see https://github.com/ckan/ckan/issues/8465), e.g.

https://github.com/ckan/ckan/blob/457a37e2e329413ea636e76e95190d9be70d93d0/ckanext/__init__.py#L3-L6

* Messing with this in CKAN 2.10 and 2.11 seems really risky, so I'd suggest pinning setuptools < 82 in dev-v2.11 / dev-v2.10
* Docker images also need to be updated to add an upper limit to setuptools:
https://github.com/ckan/ckan-docker-base/blob/9722f7809ccc3fec43069e471443040b671c006a/ckan-2.11/Dockerfile#L71
* On master / 2.12 I'd suggest keep the setuptools<82 pin for now, but maybe it's the kick in the butt we needed to decide on a strategy for namespace packages

wardi:
Fixes #9246 

### Proposed fixes:

- Remove unnecessary declare_namespace in ckanext/stats/\_\_init\_\_.py

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9246.bugfix
+++ b/changes/9246.bugfix
@@ -1,0 +1,2 @@
+Remove unnecessary declare_namespace in ckanext/stats/__init__.py
+to resolve compatibility with setuptools>=82
--- a/ckanext/stats/__init__.py
+++ b/ckanext/stats/__init__.py
@@ -1,3 +1,0 @@
-# encoding: utf-8
-
-__import__('pkg_resources').declare_namespace(__name__)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools<82"]
+requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
 
--- a/test-infrastructure/install_deps.sh
+++ b/test-infrastructure/install_deps.sh
@@ -9,7 +9,7 @@ apt-get install -y postgresql-client
 
 #Python Dependencies
 pip install -U pip
-pip install -U setuptools==81
+pip install -U setuptools
 pip install -r requirements.txt
 pip install -r dev-requirements.txt
 pip install -e .


wardi:
@amercader upgrading to setuptools 82.0.1 and relying on the existing `pkgutil.extend_path` in https://github.com/ckan/ckan/blob/96e4eb0e84813859cb21c9b50db20306fa3eb59f/ckanext/__init__.py#L7-L9

Seems to work after reinstalling all the extensions in my local environment. Are there other problems that you encountered when upgrading setuptools?

PR #9343, opened 2026-05-08
chore: Align model definition with DB tables

smotornyuk:
There are a few differences in model definition compared to the real tables from the DB. It does not change the functionality, but adds noise to `ckan generate migration --autogenerate`


ckan/model/user.py
```diff
@@ -63,7 +64,10 @@ def set_api_key() -> Optional[str]:
         Index('idx_user_name', 'name'),
         Index('idx_only_one_active_email_no_case', func.lower(Column('email')),
               unique=True, postgresql_where="(state = 'active'::text)"),
-        )
+        Index('idx_user_name_index', sa.text(
+            "(\nCASE\n    WHEN fullname IS NULL OR fullname = ''::text"
+            " THEN name\n    ELSE fullname\nEND)"
+        )))
```
smotornyuk, line 73: Index must match existing definition, including every space and newline character, or else Alembic's autogenerate will make an attempt to drop the existing index and create a new one.

wardi:
@smotornyuk I forgot to ask that we add a regression test to verify that an upgraded db generates nothing when we call `ckan generate migration --autogenerate`

smotornyuk:
Good idea, I'll make a PR with this test

PR #9344, opened 2026-05-08
Use JSONB instead of customized text for nested data

smotornyuk:
The following columns are still using a text column that is serialized/deserialized upon access to the DB:

* Resource extras
* ResourceView config
* Activity data
* ActivityDetail data (this is a deprecated model, and it will be removed in the future)

This PR alters given columns to JSONB

### DIFF
--- a/ckan/migration/versions/110_a6da0c623f10_use_jsonb_type_instead_of_customized_string.py
+++ b/ckan/migration/versions/110_a6da0c623f10_use_jsonb_type_instead_of_customized_string.py
@@ -1,0 +1,82 @@
+"""Replace custom JSON type with JSONB column.
+
+Revision ID: a6da0c623f10
+Revises: 9445ce34fc23
+Create Date: 2026-05-08 13:34:31.431283
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "a6da0c623f10"
+down_revision = "9445ce34fc23"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column(
+        "activity",
+        "data",
+        existing_type=sa.TEXT(),
+        type_=postgresql.JSONB(astext_type=sa.Text()),
+        postgresql_using="data::jsonb",
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "activity_detail",
+        "data",
+        existing_type=sa.TEXT(),
+        type_=postgresql.JSONB(astext_type=sa.Text()),
+        postgresql_using="data::jsonb",
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "resource",
+        "extras",
+        existing_type=sa.TEXT(),
+        type_=postgresql.JSONB(astext_type=sa.Text()),
+        postgresql_using="extras::jsonb",
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "resource_view",
+        "config",
+        existing_type=sa.TEXT(),
+        type_=postgresql.JSONB(astext_type=sa.Text()),
+        postgresql_using="config::jsonb",
+        existing_nullable=True,
+    )
+
+
+def downgrade():
+    op.alter_column(
+        "resource_view",
+        "config",
+        existing_type=postgresql.JSONB(astext_type=sa.Text()),
+        type_=sa.TEXT(),
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "resource",
+        "extras",
+        existing_type=postgresql.JSONB(astext_type=sa.Text()),
+        type_=sa.TEXT(),
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "activity_detail",
+        "data",
+        existing_type=postgresql.JSONB(astext_type=sa.Text()),
+        type_=sa.TEXT(),
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "activity",
+        "data",
+        existing_type=postgresql.JSONB(astext_type=sa.Text()),
+        type_=sa.TEXT(),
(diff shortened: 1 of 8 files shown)


ckanext/activity/model/activity.py
```diff
@@ -47,7 +48,7 @@ class Activity(domain_object.DomainObject, BaseModel):
         Column("user_id", types.UnicodeText),
         Column("object_id", types.UnicodeText),
         Column("activity_type", types.UnicodeText),
-        Column("data", _types.JsonDictType),
+        Column("data", JSONB),
```
smotornyuk, line 51: Activities are immutable from the API point of view, and there is no reason to track data changes via `MutableDict` here

wardi:
@smotornyuk looks good just some types and a test fix remain

smotornyuk:
I almost sure that the failing test that reports an incompatible Solr version 2.12 is not related to my changes:) Everything else is fixed

ckan/migration/versions/110_a6da0c623f10_use_jsonb_type_instead_of_customized_string.py
```diff
@@ -0,0 +1,82 @@
...
+        existing_nullable=True,
+    )
+    op.alter_column(
+        "activity_detail",
+        "data",
+        existing_type=sa.TEXT(),
+        type_=postgresql.JSONB(astext_type=sa.Text()),
+        postgresql_using="data::jsonb",
+        existing_nullable=True,
+    )
```
wardi, line 35: The changes don't mention this migration. If we're planning to remove this table eventually we could just leave it out.

ckan/model/types.py
```diff
@@ -1,71 +1,7 @@
...
-class JsonDictType(JsonType):
-
-    impl = types.UnicodeText
-
-    cache_ok = False
-
-    def process_bind_param(self, value: Any, dialect: Any):
-        # ensure we stores nulls in db not json "null"
-        if value is None or value == {}:
-            return None
```
wardi, line 63: we currently special case our custom json type to store real nulls, with the new code will be storing `'null'::jsonb` instead?

wardi:
This could use a couple tests for json storage edge cases for actions working with these fields, or if you think we're already covered by existing tests please point them out.

ckan/migration/versions/110_a6da0c623f10_use_jsonb_type_instead_of_customized_string.py
```diff
@@ -0,0 +1,82 @@
...
+
+def upgrade():
+    op.alter_column(
+        "activity",
+        "data",
+        existing_type=sa.TEXT(),
+        type_=postgresql.JSONB(astext_type=sa.Text()),
+        postgresql_using="data::jsonb",
+        existing_nullable=True,
+    )
```
wardi, line 27: activity migrations have moved to the ckanext/activity/migration directory

ckan/migration/versions/110_a6da0c623f10_use_jsonb_type_instead_of_customized_string.py
wardi: By leave it out I mean we could switch to a normal text field so no migration is created.

Another alternative, since this PR is intended for 2.13 or 3.0, is to actually remove this table if it's empty or abandon it otherwise (with a warning message).

smotornyuk: The ActivityDetail model was reported as deprecated in 2.9.0. There also [a notice that the model will be removed](https://docs.ckan.org/en/2.11/changelog.html#id725) in the next(2.10?) CKAN release. Also, in [2.11  changelog](https://docs.ckan.org/en/2.11/changelog.html#removals-and-deprecations) we mentioned "the last chance to run the migrate_package_activity.py script", so it should be safe to finally drop it.

ckan/model/types.py
smotornyuk: I can do it, but given our use of the extras column, this is a questionable choice. JSONB columns are used everywhere as a *dictionary* with additional data. All such columns are nullable, so it's already either SQL NULL or a dictionary. Introducing a third option and saving JSON `null` there probably will make things more complex. I'd prefer either to use SQL NULL to check if extras are present(`extras IS NOT NULL`), or make the column non-nullable and store an empty dictionary by default, so that extras are always there. The latter feels more natural, but I'm not sure whether PostgreSQL works better when a JSONB column contains an empty dict or NULL.

In the current codebase, I see no code that relies on this old JSON null fallback. As I understand, it would require `extras = 'null'::jsonb` SQL filter instead of `extras IS NULL` or `extras = '{}'`, which in turn requires a bit more  complex construction of SQLAlchemy query on python side

ckan/model/resource.py
```diff
@@ -54,7 +56,7 @@
     Column('cache_url', types.UnicodeText),
     Column('cache_last_updated', types.DateTime),
     Column('url_type', types.UnicodeText),
-    Column('extras', _types.JsonDictType),
+    Column('extras', MutableDict.as_mutable(JSONB)),
```
wardi, line 59: Here's what I did with the package extras field, in this case I wanted to make sure that only a dict with strings as values can be stored in the field for compatibility:
https://github.com/ckan/ckan/blob/caad0c4d4eb620e83e7d8cb6b53e02f686f4a32a/ckan/model/package.py#L78-L83

For resources IIUC we have always supported other types stored as values but it would be good to enforce that the top level is an object, at least.

wardi: I think it's worth doing the same for the other JsonDictType replacements. Validators can be overridden, this makes sure we don't store completely unexpected things by triggering a server error when trying to update the value instead of when someone later tries to view a page or access the data with the API.

PR #9226, opened 2026-01-21
Datatables View Functional JS & Distro Upgrades

JVickery-TBS:
Feature/datatables functional js upgrade

As discussed in https://github.com/ckan/ckan/pull/9218 - this PR focuses on upgrading the DataTables distros. The upgraded DataTables version fixes basically all of the redrawing issues that DataTables in CKAN has had in the past, so we can now get rid of all of the many many redraw calls.

Furthermore, all of the CKAN JS module code is being rewritten for this PR into functional JS code. This will allow a lot of the objects to just stay in memory on the page, allowing for more robust things (like not needed page reloads for compact/table mode changes).

I have added in the ColumnControl plugin which will help with future improvements coming to the DataStore soon (search with "contains" or "starts with" etc.)

It will continue to use the Bootstrap5 theme for DataTables.

Also, we are improving the language object, it is now passed to the jinja2 template via the setup_template_variables, just loaded via python json. This solves major performance issues using full MutationObservers in JS to just change languages awaiting the async loading of the language JSON files on the front-end.

cc: @wardi I am still working on this one, just in Draft PR for now.

### Features:

- [ ] includes tests covering changes
- [ ] includes updated documentation
- [X] includes user-visible changes
- [ ] includes API changes
- [ ] includes bugfix for possible backport

Please [X] all the boxes above that apply

PR #9345, opened 2026-05-08
Idatastore formats

avdata99:
### Proposed fixes:

Adding a new `IDatastoreDump` interface for the datastore so users can add, remove or modify current dump formats.
Now the list of file types to downloads from datastore is fixed (csv, tsv, xml, etc)

### Features:

- [ ] includes API changes

### DIFF
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -17,7 +17,7 @@ from ckan.plugins.toolkit import (
     ObjectNotFound, NotAuthorized, get_action, get_validator, _, request,
     abort, render, g, h, ValidationError, asbool, check_access, config,
 )
-from ckan.types import Schema, ValidatorFactory
+from ckan.types import Context, Schema, ValidatorFactory
 
 from ckanext.datastore.interfaces import IDatastoreDump
 from ckanext.datastore.logic.schema import (
@@ -116,7 +116,7 @@ def build_dump_context(
     sp = search_params or {}
     # If the caller already provides total and fields, we can skip the DB query here
     if total is None or fields is None:
-        ds_context = {'user': user} if user is not None else {}
+        ds_context = cast(Context, {'user': user} if user is not None else {})
         result = get_action('datastore_search')(
             ds_context,
             dict(sp, resource_id=resource_id, limit=0, include_total=True),
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -62,28 +62,28 @@ def get_dump_format_configs() -> dict[str, dict[str, Any]]:
             "label": "CSV",
             "writer_factory": csv_writer,
             "records_format": "csv",
-            "content_type": b"text/csv; charset=utf-8",
+            "content_type": "text/csv; charset=utf-8",
             "file_extension": "csv",
         },
         "tsv": {
             "label": "TSV",
             "writer_factory": tsv_writer,
             "records_format": "tsv",
-            "content_type": b"text/tab-separated-values; charset=utf-8",
+            "content_type": "text/tab-separated-values; charset=utf-8",
             "file_extension": "tsv",
         },
         "json": {
             "label": "JSON",
             "writer_factory": json_writer,
             "records_format": "lists",
-            "content_type": b"application/json; charset=utf-8",
+            "content_type": "application/json; charset=utf-8",
             "file_extension": "json",
         },
         "xml": {
             "label": "XML",
             "writer_factory": xml_writer,
             "records_format": "objects",
-            "content_type": b"text/xml; charset=utf-8",
+            "content_type": "text/xml; charset=utf-8",
             "file_extension": "xml",
         },
     }
@@ -188,7 +188,6 @@ def dump(resource_id: str):
                                 sort=sort,
                                 search_params=search_params,
                                 user=user_context),
-                        mimetype='application/octet-stream',
                         headers=headers)
     except ObjectNotFound:
         abort(404, _('DataStore resource not found'))
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -153,7 +153,7 @@ def dump(resource_id: str):
     dump_formats = get_dump_format_configs()
 
     if fmt not in dump_formats:
-        abort(404, _("Unsupported format"))
+        abort(400, _("Unsupported format"))
 
     format_config = dump_formats[fmt]
     file_extension = format_config["file_extension"]
@@ -298,7 +298,7 @@ def dump_to(
     dump_formats = get_dump_format_configs()
 
     if fmt not in dump_formats:
-        assert False, "Unsupported format"
+        abort(400, _("Unsupported format"))
 
     format_config = dump_formats[fmt]
     writer_factory = format_config["writer_factory"]
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -59,28 +59,28 @@ def get_dump_format_configs() -> dict[str, dict[str, Any]]:
     """Get dump format configurations from all plugins"""
     all_formats = {
         "csv": {
-            "label": "CSV",
+            "label": _("CSV"),
             "writer_factory": csv_writer,
             "records_format": "csv",
             "content_type": "text/csv; charset=utf-8",
             "file_extension": "csv",
         },
         "tsv": {
-            "label": "TSV",
+            "label": _("TSV"),
             "writer_factory": tsv_writer,
             "records_format": "tsv",
             "content_type": "text/tab-separated-values; charset=utf-8",
             "file_extension": "tsv",
         },
         "json": {
-            "label": "JSON",
+            "label": _("JSON"),
             "writer_factory": json_writer,
             "records_format": "lists",
             "content_type": "application/json; charset=utf-8",
             "file_extension": "json",
         },
         "xml": {
-            "label": "XML",
+            "label": _("XML"),
             "writer_factory": xml_writer,
             "records_format": "objects",
             "content_type": "text/xml; charset=utf-8",
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -169,6 +169,16 @@ def dump(resource_id: str):
     file_extension = format_config["file_extension"]
     content_type = format_config["content_type"]
 
+    # If the format declares a validator, run it before we start
+    # streaming. A 400 here is the only way to surface a reason
+    # cleanly; once the chunked response starts, dropping the
+    # connection is the best we can do.
+    validate = format_config.get("validate")
+    if validate is not None:
+        reason = validate(resource_id)
+        if reason:
+            abort(400, reason)
+
     content_disposition = 'attachment; filename="{name}.{ext}"'.format(
         name=resource_id, ext=file_extension
     )
--- a/ckanext/datastore/blueprint.py
+++ b/ckanext/datastore/blueprint.py
@@ -78,9 +78,8 @@ def get_dump_format_configs() -> dict[str, dict[str, Any]]:
         },
     }
 
-    # Allow plugins to add, override, or remove formats.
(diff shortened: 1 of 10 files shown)


avdata99:
A extension to download as XLS is working here: https://github.com/okfn/ckanext-ddfxls/

image

avdata99:
#9353 was started because tests fails here

ckanext/datastore/tests/test_dump.py
```diff
@@ -911,3 +911,75 @@ def get_csv_record_values(response_body):
...
+    @pytest.mark.ckan_config("ckan.plugins", "datastore")
+    @pytest.mark.usefixtures("clean_datastore", "with_plugins")
+    def test_unknown_format_rejected(self, app):
+        # Sanity check: the schema's one_of validator rejects formats
+        # not in the registry, regardless of any plugin.
+        resource = factories.Resource()
+        helpers.call_action(
+            "datastore_create",
+            resource_id=resource["id"],
+            force=True,
```
wardi: instead of `force=True` use `url_type='datastore'` in the Resource factory.

ckanext/datastore/interfaces.py
```diff
@@ -232,3 +232,60 @@ def update_datastore_info_field(
         in the data dictionary page.
         """
         return field
+
+
+class IDatastoreDump(interfaces.Interface):
+    """
+    Allow plugins to register custom dump formats and writers for datastore
+    exports
+    """
```
wardi, line 241: use `_reverse_iteration_order = True` so that earlier plugins can override values from later plugins

ckanext/datastore/interfaces.py
```diff
@@ -232,3 +232,60 @@ def update_datastore_info_field(
...
+        - 'records_format': The format for records ('csv', 'tsv',
+          'lists', 'objects')
+        - 'content_type': The MIME type for the response (str)
+        - 'file_extension': The file extension for downloads
+
+        Example: add ``xlsx`` and remove ``xml``::
+
+            {
+                'xlsx': {
+                    'label': 'Excel',
```
wardi: are label translations supported?

wardi:
@avdata99 Note that formats like Excel have hard limits on the number of rows per worksheet:
https://support.microsoft.com/en-us/office/excel-specifications-and-limits-1672b34d-7043-467e-8e27-269d656771c3

So we would need a way to return a list of formats supported based on the number of records being exported.

ckanext/datastore/tests/test_dump.py
avdata99: Updated here: https://github.com/ckan/ckan/commit/d5b212a1842501eef96258168e488fc6e22377cb

ckanext/datastore/interfaces.py
avdata99: Done her https://github.com/ckan/ckan/commit/868ab5ed280b48561912b44f43fab06da6caee18

ckanext/datastore/interfaces.py
avdata99: They are supported. Here is maybe clearer: https://github.com/ckan/ckan/commit/f26600396c350fdacd9700a0dea82709759bd6fc
In the extension: https://github.com/okfn/ckanext-ddfxls/commit/d81ad127daa3c13a42855c64e517bb2106536c6f

avdata99:
> @avdata99 Note that formats like Excel have hard limits on the number of rows per worksheet: https://support.microsoft.com/en-us/office/excel-specifications-and-limits-1672b34d-7043-467e-8e27-269d656771c3
> 
> So we would need a way to return a list of formats supported based on the number of records being exported.

@wardi here we allow extension to define validators: https://github.com/ckan/ckan/commit/62d33d632f9af6e79abe0adaf1ce37bd4b785050
We were worried about counting rows on each request but it looks like `datastore_search ... include_total=True` will not be expensive. What do you think? It look like this is a new feature.
Also, another extensions may require different things, not only row or fields count so we think extension should handle validation their own.
We implement the validator in the extension here: https://github.com/okfn/ckanext-ddfxls/commit/5cc3f6b67b39fc9fe3372be4c1fb497949df5b06

ckanext/datastore/templates/package/resource_read.html
```diff
@@ -11,22 +11,18 @@
...
-          <a class="dropdown-item" href="{{ h.url_for('datastore.dump', resource_id=res.id, format='xml') }}"
-          target="_blank" rel="noreferrer"><span>XML</span></a>
+        {% if fmt.available %}
+        <a class="dropdown-item" href="{{ h.url_for('datastore.dump', resource_id=res.id, format=fmt.name, bom=True) }}"
+          target="_blank" rel="noreferrer"><span>{{ fmt.label }}</span></a>
+        {% else %}
+        <span class="dropdown-item disabled" title="{{ fmt.reason }}" aria-disabled="true">
+          {{ fmt.label }} <small class="text-muted">({{ _('unavailable') }})</small>
+        </span>
+        {% endif %}
```
wardi, line 25:
```suggestion
        {% if fmt.available %}
          <a class="dropdown-item" href="{{ h.url_for('datastore.dump', resource_id=res.id, format=fmt.name, bom=True) }}"
            target="_blank" rel="noreferrer"><span>{{ fmt.label }}</span></a>
        {% else %}
          <span class="dropdown-item disabled" title="{{ fmt.reason }}" aria-disabled="true">
            {% trans format_label=fmt.label %}
              {{ format_label }} <small class="text-muted">(unavailable)</small>
            {% endtrans %}
          </span>
        {% endif %}
```
It's better to provide the whole text block to gettext in case some translators prefer to reorder things.

ckanext/datastore/tests/sample_dump_plugin.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def invalid_validate(resource_id):
+    """Validator that always rejects the format.
+
+    Used by tests to confirm that a ``validate`` callable on a
+    registered format flows through to the helper, the blueprint, and
+    the resource-read template.
+    """
+    return ALWAYS_INVALID_REASON
```
wardi, line 13: This solution is too general because the only reason we need to show (unavailable) is when there are too many rows/columns for an export format. There aren't other cases that need to be handled based on resource metadata, and starting from a resource id means there will be extra work and wasted time fetching the information required e.g. by a call to `datastore_info`.

This is also too specific because it assumes that we can determine the number of rows/columns from the resource id, but the records may be filtered by the user so the total number of records/columns won't be the same as the number of records being exported.

Let's instead let IDatastoreDump plugins define a maximum number of rows and columns for a format and make the download resource button aware of the number of rows filtered and columns displayed by the user so it can display the available formats.

This will be compatible with future code that allows filtering without a page refresh with JS or HTMX.

ckanext/datastore/blueprint.py
```diff
@@ -37,17 +40,73 @@
...
 
 
+def get_dump_formats() -> list[str]:
+    """Get available dump formats from plugins.
+
+    Returns a list
+    List > Tuple when we have one element only
+    """
+    dump_format_configs = get_dump_format_configs()
+    return list(dump_format_configs.keys())
```
wardi:
```suggestion
```
This function doesn't need to exist. It's equivalent to:
```python
list(get_dump_format_configs())
```

ckanext/datastore/blueprint.py
```diff
@@ -244,20 +304,15 @@ def dump_to(
...
-    elif fmt == 'xml':
-        writer_factory = xml_writer
-        records_format = 'objects'
-    else:
-        assert False, 'Unsupported format'
+    # Get format configuration from plugin registry
+    dump_formats = get_dump_format_configs()
+
+    if fmt not in dump_formats:
+        assert False, "Unsupported format"
```
wardi: while we're here let's send a 400 error instead of a server error.

ckanext/datastore/blueprint.py
avdata99: TIL:
```python
data_dict = {"a": 1, "b": 2}
print(list(data_dict))
['a', 'b']
```
Updated here: https://github.com/ckan/ckan/commit/1717b48d7119ab1bb61970ec8de07465419f9915

ckanext/datastore/blueprint.py
avdata99: Updated here: https://github.com/ckan/ckan/commit/ed2824f078f5d0543ce4658d5936fdf6fcc7f015
Also another response updated (it was 404)

ckanext/datastore/tests/sample_dump_plugin.py
avdata99: Thanks, we reworked the proposal around your points, and agree with most of it.
"Starting from a resource id is wasted work": agreed. The validator no longer starts from a resource_id or fetches on its own. We resolve the export scope once per request (a single datastore_search with limit=0, include_total=True; for the unfiltered case) and passes it as a context to every format. No per-format datastore_info calls.

"Records may be filtered": we were missing this one. Availability is now evaluated against the filtered export scope (post-filter row count + the actually-selected columns), not the resource totals. So a user who filters a 2M-row resource down to 50k correctly still gets XLSX offered. The dropdown helper and the dump endpoint both take the query params, and build_dump_context also accepts caller-supplied total/fields.  So a future JS/HTMX "filter without reload" flow can pass the counts it already knows and skip the query entirely. It's forward-compatible with the no-refresh filtering you mentioned.

"The only reason is too many rows/columns": here we'd push back a little. Rows/columns are the common case and probably what most people hit, so we made them first-class: a format just declares max_rows/max_columns and the framework does the comparison and writes the message, no validator callable needed. But they aren't the only case. For example a GeoJSON format should only be offered when the resource actually has geometry columns (lat/lng, or a geom/PostGIS column), that's a schema check, not a count. So we kept an optional validate(context) for constraints counts can't express.

Resolution order is max_columns -> max_rows -> validate, first rejection wins. For vanilla CKAN (which ships none of these controls) the dropdown render and the dump endpoint do no extra DB work. The scope is only resolved when an installed format opts in.

We've already adopted this in our XLSX extension and dropped its custom validator entirely, it's now just two declared limits:
https://github.com/okfn/ckanext-ddfxls/commit/035768db69dafa57a518123b7b7c2a1a8f285027

Latest changes here:
Prepare new validation for max column and rows + We preserve custom validation fn for extension
https://github.com/ckan/ckan/commit/5405ad92bb903e53490dcc7b4f307226d398d9b7

Update interface docs
https://github.com/ckan/ckan/commit/1fec3abe570ff89d5a499bdadddd1aeb234c210e

Update sample plugin and test
https://github.com/ckan/ckan/commit/5454bfaa07a23f2ae4c52cc3d80abb9579c79bec

wardi: Feels like I'm talking to an LLM here.

@avdata99 if you're using claude or something to make these changes please make sure you edit the code it produces, remove excessive comments and rephrase github replies to make sure you understand what you're posting and that everything makes sense.

wardi:
I would like to see this feature integrated with the datatables changes in #9226 . That way we can actually exercise the code that checks the totals and columns based on filters the user choses interactively.

avdata99:
I understand what you're saying @wardi. I always keep in mind that a pull request should be as easy to review as possible and not overloaded.
My response was as brief and explanatory as possible, considering all the issues I should address.
I cleaned up the comments a bit here: https://github.com/avdata99/ckan/commit/bd32358d0204bd047018402d6718be2714795631

I suppose that as a reviewer you have to deal with this all the time. I'm aware of this, and while I took it into account, it seems it wasn't enough. I'll try to be briefer and more concise in the future. Now that I reread it, I think I could have been.

When I submitted this pull request, I felt I understood what it changes. Some doors have been left open so that if the data is filtered, the extensions can validate whether each format is available or not. Ensuring this covers everything necessary and testing compatibility with #9226 is pending. if #9226 is merged, I can pull and do more testing on master.
If this PR is closer to being mixed, perhaps Jesse can also review it once this change is applied.

wardi:
@JVickery-TBS if you have time please take a look at this with an eye to how it can be combined with #9226

PR #9352, opened 2026-05-13
chore: cloud storages has duplicated path in download links

smotornyuk:
I noticed a bug in file-keeper while experimenting with cloud storage. When `path` is specified in cloud storage(i.e, files uploaded to a subfolder of the bucket), download links are generated with a duplicated path section(something like `resource/resource/file.txt`). 

The fix is added to [new file-keeper version](https://github.com/DataShades/file-keeper/releases/tag/v0.2.3)


smotornyuk:
Backported this manually

PR #9354, opened 2026-05-14
fix: Active state of facets is not computed properly

smotornyuk:
`active` state of facets is computed against `request.args.items(...)` that [returns generator](https://werkzeug.palletsprojects.com/en/stable/datastructures/#werkzeug.datastructures.MultiDict.items). Because of it, not all active facets are marked as such.

This issue appears only in v2.12 and does not require backporting to older versions.


smotornyuk:
Backported this manually

PR #9365, opened 2026-05-28
Datastore Select distinct fix

EricSoroos:
Select distinct, seen in the filters for the datastore, is returning incorrect data after #9028 (master/2.12)

The sql query that's generated is this:
```
datastore-1  | 	            WITH records_page AS (
datastore-1  | 	                SELECT * FROM "beaa4a4e-04ec-455f-a79a-08f201ce6296"   ORDER BY "bidding_zone" asc LIMIT 20 OFFSET 0
datastore-1  | 	            ), select_cols AS (
datastore-1  | 	                SELECT DISTINCT "bidding_zone" as "bidding_zone" FROM records_page
datastore-1  | 	            ), json_cols AS (
datastore-1  | 	                SELECT array_to_json(array_agg(select_cols))::text j
datastore-1  | 	                FROM select_cols
datastore-1  | 	            ) SELECT json_cols.j FROM json_cols
```

This is selecting the first 20 records by id, and then running select distinct on the bidding_zone column, however the limit/offset needs do be done in the distinct section, otherwise we're only sampling that first 20 rows. 

### Proposed fixes:

```
            WITH records_page AS (
                SELECT * FROM {resource} {ts_query} {final_statement}
            ), select_cols AS (
                SELECT {distinct} {select} FROM records_page
            ), json_cols AS (
                SELECT array_to_json(array_agg(select_cols))::text j
                FROM select_cols
            )
```

This splits the final_statement into final_statement and distinct_final_statement, so that the sort/order/limit can be done in the distinct segment if required.

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9365.bugfix
+++ b/changes/9365.bugfix
@@ -1,0 +1,1 @@
+Fix datastore distinct queries with limit/offset
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -1458,6 +1458,7 @@ def search_data(context: Context, data_dict: dict[str, Any]):
     limit = query_dict['limit']
     offset = query_dict['offset']
 
+    distinct_final_statement = ''
     if query_dict.get('distinct'):
         distinct = 'DISTINCT'
     else:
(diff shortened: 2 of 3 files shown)


wardi:
@EricSoroos I planning to merge without a changelog because it's fixing a feature that's not released yet. I think that's what we usually do.

PR #9370, opened 2026-06-03
fix: Case-sensitive reset-link by email

smotornyuk:
Password reset relies on a case-sensitive email search. If the user is registered with `Sergey@gmailcom`, he won't be able to reset his password using `sergey@gmail.com`.

This PR is similar to #9178 (case-insensitive duplicate check) and #8626 (case-insensitive `User.by_email`). But because pasword reset endpoint relies on `user_list` action, which in turn builds complex query and instead of calling `User.by_email` it calls `query.filter(email==email)`, previous fixes are not applied to the current problem.

### DIFF
--- a/changes/9370.bugfix
+++ b/changes/9370.bugfix
@@ -1,0 +1,1 @@
+Use case-insensitive email search when requesting password reset.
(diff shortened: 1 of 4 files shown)


wardi:
backported manually

PR #9373, opened 2026-06-04
Add csrf_input to tabledesigner row forms

### LINKED ISSUE #9372: tabledesigner row forms (add/edit/delete) are missing csrf_input()
## CKAN version
`master` (verified at `caad0c4d4e`); present since the Table Designer feature landed (#7882) and in all releases that include it (2.10+ / 2.11.x).

## Describe the bug
The three state-changing forms in `ckanext.tabledesigner` do not include the CSRF token (`{{ h.csrf_input() }}`):

- `ckanext/tabledesigner/templates/tabledesigner/add_row.html:12` — `<form method="post" action="{{ action }}" >`
- `ckanext/tabledesigner/templates/tabledesigner/edit_row.html:12` — `<form method="post" action="{{ action }}?_id={{ row['_id'] }}" >`
- `ckanext/tabledesigner/templates/tabledesigner/delete_rows.html:27` — `<form method="post" >`

The token is absent from these templates and any snippets they include (no `csrf_input` anywhere under `ckanext/tabledesigner/templates/`), and the forms are submitted as plain HTML POSTs (no `X-CSRF-Token` header via JS).

This is inconsistent with the other bundled extensions, which **do** emit the token: `datastore` (`templates/datastore/dictionary.html`), `datapusher` (`templates/datapusher/resource_data.html`), `datatablesview` (`templates/datatables/datatables_view.html` + the `X-CSRF-Token` header in `datatablesview.js`).

Note: the `csrf_input()` helper (#7029, 2.10) predates the Table Designer feature (#7882), so the helper was available when these forms were written.

## Why it matters
`tabledesigner` registers an `IBlueprint`, so by default (`ckan.csrf_protection.ignore_extensions = true`) its blueprint is exempted and the missing token goes unnoticed. But:

1. **CSRF exposure when enabled:** with the extension active, `POST` to add/edit/delete DataStore rows is reachable with no CSRF protection — a forged cross-site request from a logged-in editor's browser can insert/modify/delete rows.
2. **Incompatible with the recommended hardening:** the docs recommend `ckan.csrf_protection.ignore_extensions = false` to enforce CSRF on extensions (and it's slated to become the default — the option is annotated *"will be deprecated in future versions"*). With that set, tabledesigner's **own** forms break with HTTP 400, because they don't supply a token. So an admin following the hardening guidance disables row editing in tabledesigner.

## Steps to reproduce
1. Enable `tabledesigner` (+ `datastore`, `datatables_view`) and set `ckan.csrf_protection.ignore_extensions = false`.
2. Open a DataStore-backed resource's Table Designer and try to add/edit/delete a row.
3. The POST is rejected with **400** (CSRF token missing).

(With the default `ignore_extensions = true`, the same POST is instead accepted **without any token**, i.e. unprotected.)

## Expected behaviour
The add/edit/delete-row forms should include `{{ h.csrf_input() }}` so they're protected when extension CSRF is enforced and submit successfully — consistent with the other bundled extensions.

## Suggested fix
Add `{{ h.csrf_input() }}` immediately inside each `<form method="post" …>` in `add_row.html`, `edit_row.html`, and `delete_rows.html`. Happy to open a PR.

graingert-coef:
## Fixes #9372

The Table Designer add-row, edit-row and delete-rows forms were missing `{{ h.csrf_input() }}`, so these state-changing POSTs carried no CSRF token:

- `ckanext/tabledesigner/templates/tabledesigner/add_row.html`
- `ckanext/tabledesigner/templates/tabledesigner/edit_row.html`
- `ckanext/tabledesigner/templates/tabledesigner/delete_rows.html`

Because `tabledesigner` registers an `IBlueprint`, its routes are exempt from CSRF by default (`ckan.csrf_protection.ignore_extensions = true`), so this went unnoticed. But:

- when the extension is enabled, add/edit/delete-row POSTs are reachable with no CSRF protection; and
- under enforced extension CSRF (`ckan.csrf_protection.ignore_extensions = false`, which the docs recommend and which is slated to become the default) tabledesigner's own forms break with HTTP 400.

This adds the token inside each `<form method="post">`, consistent with the other bundled extensions (`datastore`, `datapusher`, `datatablesview`), which already emit it.

A changelog fragment is included (`changes/9372.bugfix`).

See #9372 for the full analysis and reproduction.

### DIFF
--- a/changes/9372.bugfix
+++ b/changes/9372.bugfix
@@ -1,0 +1,3 @@
+Add the missing CSRF token (``h.csrf_input()``) to the Table Designer add-row,
+edit-row and delete-rows forms, so they are protected and continue to work when
+extension CSRF protection is enforced (``ckan.csrf_protection.ignore_extensions = false``).
--- a/ckanext/tabledesigner/templates/tabledesigner/add_row.html
+++ b/ckanext/tabledesigner/templates/tabledesigner/add_row.html
@@ -10,6 +10,7 @@
   <h1>{{ _('Add row') }}</h1>
 
   <form method="post" action="{{ action }}" >
+    {{ h.csrf_input() }}
     {% block add_row %}
       {% for field in fields %}
         {% set ct = h.tabledesigner_column_type(field) %}
--- a/ckanext/tabledesigner/templates/tabledesigner/delete_rows.html
+++ b/ckanext/tabledesigner/templates/tabledesigner/delete_rows.html
@@ -25,6 +25,7 @@
   </table>
 
   <form method="post" >
+    {{ h.csrf_input() }}
     <button class="btn btn-primary" name="save" type="submit">
       <i class="fa fa-alert"></i> {{ _('Delete') }}
     </button>
--- a/ckanext/tabledesigner/templates/tabledesigner/edit_row.html
+++ b/ckanext/tabledesigner/templates/tabledesigner/edit_row.html
@@ -10,6 +10,7 @@
   <h1>{{ _('Edit row') }}</h1>
 
   <form method="post" action="{{ action }}?_id={{ row['_id'] }}" >
+    {{ h.csrf_input() }}
     {% block add_row %}
       {% for field in fields %}
         {% set ct = h.tabledesigner_column_type(field) %}


munzzyy:
Verified this against current master (e1c9baa, tabledesigner views.py) - there are exactly four POST handlers in the extension: the dictionary view (delegates to datastore's DictionaryView/dictionary.html, which already has csrf_input at line 15) plus add_row, edit_row, and delete_rows, which this PR covers. That's the full set of state-changing forms in tabledesigner; nothing submits via JS/fetch here (the only fetch() calls in the extension are in the datastore API doc examples, unrelated to these forms), so a header-based token isn't needed.

Placement matches the existing pattern in datastore/dictionary.html (csrf_input as the first line inside the form tag). CI is green across the board and the changelog fragment is in the right format.

This looks complete and correct to me - covers the whole vulnerable surface, not just part of it.

PR #9379, opened 2026-06-10
feat: separate webassets debug mode

smotornyuk:
### Enable remote dependencies for webassets

Asset definition looks like:

```yaml
scripts:
  contents:
      - vendor/hljs.js
      - javascript/ui.js
```

If you want to add a library to the assets, you need to download and include it in the source code first. It generally works, but the webassets library actually supports CDN links for external libraries. This PR modifies the code to enable the following definition:

```yaml
scripts:
  contents:
      - https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.11.1/highlight.min.js
      - javascript/ui.js
```

When assets are served in debug mode, the CDN URL will be included as is into the page source. When assets are served in production mode, the remote library will be downloaded upon first access and concatenated with the rest of the local assets.

### Separate debug mode for webasseets

At the moment, when CKAN's debug mode is enabled, webassets are not aggregated, and every source file is served in its original form, separately. Depending on the number of files in the web assets group, it may give us 50-200 additional requests for separate files when there are ~10-15 extensions enabled. 

In production mode, source files are merged, and every unique `{% asset ... %}` tag produces approximately 1 additional request, so for 15 extensions, there will be ~15 additional requests, which is reasonable. And, because merged files are stored in a single webassets folder, we can further configure the web server to serve these files directly, bypassing CKAN. In the case of the debug mode, it's not possible because every file is served from its original location inside the extension.

Unless you are debugging webasset issues, most likely you'll prefer production-mode behavior for webassets, even if CKAN's debug mode is enabled. This PR moves webassets debug mode to a new config option `ckan.webassets.debug`, which is turned off by default so that it won't cause unexpected performance penalty before you start debugging webassets.


EricSoroos:
CDN usage should be able to use a content security hash.

wardi:
@smotornyuk @EricSoroos instead of relying on the node package lock format for integrity maybe we can use https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Subresource_Integrity for hashing CDN resources, and allow extensions to use CDN links instead of downloading and bundling on the server side.

wardi:
An ideal end state would support any of the following:
- site operators/extension developers wanting JS served from CDNs safely
- site operators wanting all JS from all enabled extensions verified, bundled and served from CKAN
- sites that require bundling JS before deployment because their web VMs are firewalled and can't reach out to download anything.

smotornyuk:
I will work on CDN assets separately, as the desired scope it far from built in capabilities of webassets. 

In this PR I keep only config option to control whether assets must be merged or not, to reduce server load. We still can include it into v2.12, while CDN assets will take more time for proper implementation

ckan/config/config_declaration.yaml
```diff
@@ -1605,6 +1605,14 @@ groups:
...
           requires that the web server (eg Nginx) supports the ``X-Sendfile`` header. See :ref:`x-sendfile` for more information.
 
+      - key: ckan.webassets.debug
+        type: bool
+        example: true
+        description: |
+          When this setting is ``True``, webassets will not combine and minify
+          the assets, and will instead serve them as separate files. This may
+          be useful for debugging purposes, but it is not recommended for
+          production environments.
```
wardi, line 1616: @smotornyuk shouldn't we set a default here?

smotornyuk: Added it as `false` by default. Because it not depends on the global debug mode anymore and will apply to sites in production mode, it must be disabled.

PR #9386, opened 2026-06-16
[2.11][#9380] Fix organization patch on orgs with extras

### LINKED ISSUE #9380: [2.11] `organization_patch` fails for non-sysadmin users if org has extras
## CKAN version

2.11.x

## Describe the bug
A non-sysadmin user can not use `group_patch` or `organization_patch` on a group / org with `extras`. It returns a validation error regardless of the input.

### Steps to reproduce

```python 
    def test_organization_patch_with_extras():
        user = factories.User()
        organization = factories.Organization(
            description="some test now", user=user,
            extras=[
                {"key": "k1", "value": "v1"},
                {"key": "k2", "value": "v2"},
            ]
        )

        organization = helpers.call_action(
            "organization_patch",
            id=organization["id"],
            description="somethingnew",
            context={
                "user": user["name"],
                "ignore_auth": False,
            },
        )
```
Fails with:

```
E           ckan.logic.ValidationError: None - {'extras': [{'id': ['The input field id was not expected.']}, {'id': ['The input field id was not expected.']}]}
```

### Additional details
The issue is the show schema used for groups and orgs plus the validators on `extras["id"]`

Back in #8069 we added validators to ignore ids unless you were a sysadmin:

https://github.com/ckan/ckan/blob/28c414b43a9803ae4855854f318f9fdae48dedfb/ckan/logic/schema.py#L386

The patch functions call `organization_show` / `group_show` to fill the rest of properties not provided. When calling the show function, the group/show schema does not use the [`default_show_extras_schema`](https://github.com/ckan/ckan/blob/dev-v2.11/ckan/logic/schema.py#L397), which hides the `id` field.

https://github.com/ckan/ckan/blob/28c414b43a9803ae4855854f318f9fdae48dedfb/ckan/logic/schema.py#L401

So the `organization_patch` action, calls `organization_show` and passes the output to `organization_update`. If the org has extras, these have an `id`, and unless you are a sysadmin, this makes validation fail.

This is not an issue in 2.12/master because the dictize functions only return `key`, `value` for extras after https://github.com/ckan/ckan/pull/8288

### Potential fixes

The proper fix would probably to make `default_show_group_schema` use `default_show_extras_schema` but that will change the API output on 2.11 sites (hiding the `extras["id"]` field, which admitely should not be there in the first place). 
A safer fix is to remove the `id` field from `extras` within the patch functions before passing the dict to the update function.

amercader:
Fixes #9380

Removing the `id` field from extras before passing the data dict to `organization_update`. Rather than changing the schema this seems like a safer fix to backport

### DIFF
--- a/ckan/logic/action/patch.py
+++ b/ckan/logic/action/patch.py
@@ -132,9 +132,6 @@ def organization_patch(
     patched = dict(organization_dict)
     patched.pop('display_name', None)
     patched.update(data_dict)
-    if patched.get("extras"):
-        for extra in patched["extras"]:
-            extra.pop("id", None)
 
     patch_context = context.copy()
     patch_context['allow_partial_update'] = True
--- a/ckan/logic/action/patch.py
+++ b/ckan/logic/action/patch.py
@@ -132,6 +132,9 @@ def organization_patch(
     patched = dict(organization_dict)
     patched.pop('display_name', None)
     patched.update(data_dict)
+    if patched.get("extras"):
+        for extra in patched["extras"]:
+            extra.pop("id", None)
 
     patch_context = context.copy()
     patch_context['allow_partial_update'] = True
--- a/ckan/logic/schema.py
+++ b/ckan/logic/schema.py
@@ -383,7 +383,7 @@ def default_extras_schema(ignore: Validator, not_empty: Validator,
                           ignore_missing: Validator, uuid_validator: Validator,
                           ignore_not_sysadmin: Validator) -> Schema:
     return {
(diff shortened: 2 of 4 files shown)


wardi:
Could this instead be fixed by replacing `empty_if_not_sysadmin` with `ignore_not_sysadmin` here:

https://github.com/ckan/ckan/blob/28c414b43a9803ae4855854f318f9fdae48dedfb/ckan/logic/schema.py#L386

That would fix the issue for `organization_update` and `group_update` too, which are used by `ckanapi load groups | organizations`

PR #9403, opened 2026-07-04
Fix R datastore API examples

### LINKED ISSUE #9371: Error in R api example code
## CKAN version
ckan-2.11.5

## Describe the bug

There is an error in the R code api example usage at: 
https://github.com/ckan/ckan/blob/caad0c4d4eb620e83e7d8cb6b53e02f686f4a32a/ckanext/datastore/templates/datastore/api_examples/r.html#L17
which does not provide a complete example request as the last two functions are missing their brackets and a connecting pipe. 

The same error occurs at lines 34 and 48.  

### Expected behavior
A correction to the example would be: 
 
```
req %&gt;% 
    req_headers(Authorization = API_TOKEN) %>% 
    req_body_json(list(
        resource_id='{{ resource_id }}', 
        filters = list(
            subject = list("watershed", "survey"), 
            stage = "active"))) %>%
    req_perform() %>% 
    resp_body_json()
```

wenxiaojie1:
## Summary
- add the missing pipes after `req_body_json(...)` in the R datastore API examples
- call `req_perform()` and `resp_body_json()` as functions so the examples are complete httr2 pipelines

Fixes #9371.

## Validation
- Ran `git diff --check`.
- Searched the edited R example template for remaining `req_perform` / `resp_body_json` calls without parentheses.

### DIFF
--- a/changes/9371.bugfix
+++ b/changes/9371.bugfix
@@ -1,0 +1,1 @@
+Fix the R datastore API examples so the httr2 request pipelines are complete.
--- a/ckanext/datastore/templates/datastore/api_examples/r.html
+++ b/ckanext/datastore/templates/datastore/api_examples/r.html
@@ -13,9 +13,9 @@ result &lt;- req %&gt;%
     req_body_json(list(
         resource_id = '{{ resource_id }}',
         limit = 5,
-        q = 'jones'))
-    req_perform %&gt;% 
-    resp_body_json</code></pre>
+        q = 'jones')) %&gt;%
+    req_perform() %&gt;%
+    resp_body_json()</code></pre>
 {% endcall %}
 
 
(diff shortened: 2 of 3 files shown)


wardi:
@wenxiaojie1 similar changes may be required in https://github.com/ckan/ckan/blob/master/ckanext/tabledesigner/templates/datastore/api_examples/r.html

wenxiaojie1:
Thanks for the pointer. I updated the tabledesigner R examples with the same pipe fixes, and also cleaned up the two remaining unescaped `%>%` pipes in the datastore R examples.

Validation:
- `git diff --check`
- checked that no `req_perform` / `resp_body_json` calls remain without `()`
- checked that no `Authorization = API_TOKEN) %>%` pipes remain unescaped in the R examples

PR #9432, opened 2026-07-21
fix: gracefully handle missing context fields in API call

ThrawnCA:
When calling an API internally and setting ignore_auth, we should not need to supply a user object.

### Proposed fixes:

Use `context.get('user')` instead of `context['user']` so we can gracefully handle absent keys.

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/action/patch.py
+++ b/ckan/logic/action/patch.py
@@ -37,9 +37,9 @@ def package_patch(
     _check_access('package_patch', context, data_dict)
 
     show_context: Context = {
-        'session': context['session'],
-        'user': context['user'],
-        'auth_user_obj': context['auth_user_obj'],
+        'session': context.get('session'),
+        'user': context.get('user'),
+        'auth_user_obj': context.get('auth_user_obj'),
         'ignore_auth': context.get('ignore_auth', False),
         'for_update': True
     }


smotornyuk:
Any internal api call must be made via [get_action](https://github.com/ckan/ckan/blob/acd9fff726105b8cd46d2b04788f837d2d19f5cc/ckan/logic/__init__.py#L448). This function always invokes [_prepopulate_context](https://github.com/ckan/ckan/blob/acd9fff726105b8cd46d2b04788f837d2d19f5cc/ckan/logic/__init__.py#L570) that is responsible for [setting the user](https://github.com/ckan/ckan/blob/acd9fff726105b8cd46d2b04788f837d2d19f5cc/ckan/logic/__init__.py#L327).

If you call the action in this way, there should be no problem. Every other action makes the same assumption and there is no reason to make an exception for package_patch

ThrawnCA:
> Any internal api call must be made via [get_action](https://github.com/ckan/ckan/blob/acd9fff726105b8cd46d2b04788f837d2d19f5cc/ckan/logic/__init__.py#L448). This function always invokes [_prepopulate_context](https://github.com/ckan/ckan/blob/acd9fff726105b8cd46d2b04788f837d2d19f5cc/ckan/logic/__init__.py#L570) that is responsible for [setting the user](https://github.com/ckan/ckan/blob/acd9fff726105b8cd46d2b04788f837d2d19f5cc/ckan/logic/__init__.py#L327).
> 
> If you call the action in this way, there should be no problem. Every other action makes the same assumption and there is no reason to make an exception for package_patch

My testing of ckanext-ytp-comments was encountering KeyError when I didn't manually pass a `user`. Maybe there's something about the latest `dev-v2.12`?

ckan/logic/action/patch.py
```diff
@@ -37,9 +37,9 @@ def package_patch(
     _check_access('package_patch', context, data_dict)
 
     show_context: Context = {
```
caiotheodoro, line 39: bug: `Context` declares `user`, `auth_user_obj` etc. as required (non-Optional), so `.get()` returns a type this dict literal can't satisfy - that's why PyRight is failing on this exact line, not a flake. `logic.fresh_context` already solves this the right way a few lines below in the same module tree: `{k: context[k] for k in (...) if k in context}` builds the dict from only the keys that actually exist instead of defaulting missing ones to `None`. Use that pattern here instead of `.get()`.

caiotheodoro (commented):
CI is red for real reasons, not flakiness: typecheck fails because of the `.get()` change (see inline), and towncrier fails because there's no fragment in `changes/`.

ckan/logic/action/patch.py
ThrawnCA: Thanks for the feedback. Does this look better?

Issue #4897, opened 2019-07-10
RFE: Extend paster 'search-index rebuild' to support multiple inputs


ThrawnCA:
### CKAN Version if known (or site URL)

2.8.2

### Please describe the expected behaviour

The paster `search-index rebuild` command should ideally be able to accept multiple input IDs, perhaps a filename or reading STDIN.

Eg

`paster --plugin=ckan search-index rebuild -f /tmp/package-ids.txt -c /path/to/config.ini`

where `/tmp/package-ids.txt` contains:

    aaaaaaaa-0000-1111-222222222222
    bbbbbbbb-3333-4444-555555555555
    cccccccc-6666-7777-888888888888

Or to read STDIN:

`cat /tmp/package-ids.txt | paster --plugin=ckan search-index rebuild -f - -c /path/to/config.ini`

### Please describe the actual behaviour

`search-index rebuild` can process either one dataset or every dataset.

davidread:
Sounds reasonable to me.

wardi:
How about a sysadmin-only API action to rebuild the search index for a dataset? That way we can build any sort of system to feed dataset ids that we want, and we don't require running a command directly on the server.

ThrawnCA:
Can that already be achieved with some kind of a no-op package update?

davidread:
@ThrawnCA Possibly. There is a check_real_change method that would prevent a revision being created, but maybe it is still reindexed.

wardi:
forcing a re-index is a slightly different thing. If there was really no update ideally `package_update` will do absolutely nothing (and we could even return an indication to the user indicating that)

The times when you need to force a reindex you know you don't want to change anything you just want the indexing code to be re-run (eg. to update "hotness" indication on a dataset or cause some other plugin to attach some updated data via `before_index`) so I feel like they make sense as separate actions.

Also it's really hard to extend `package_update` with a "index_only" option or similar since all values passed are assumed to be part of the metadata you'd like updated.

ThrawnCA:
As a superior alternative to multiple inputs, it would be nice if the `rebuild` action could make use of search filters.

PR #9440, opened 2026-07-22
Add new `package_reindex` API action

ThrawnCA:
### Proposed fixes:

Add a `package_reindex` API action to trigger an index rebuild for a package without making changes.

This is useful when eg plugins have registered `before_dataset_index` hooks that they want to manually run. See https://github.com/ckan/ckan/pull/8396#discussion_r3626992080

### Features:

- [X] includes API changes

### DIFF
--- a/ckan/logic/action/update.py
+++ b/ckan/logic/action/update.py
@@ -186,6 +186,7 @@ def resource_view_update(
         model.repo.commit()
     return model_dictize.resource_view_dictize(resource_view, context)
 
+
 def resource_view_reorder(
         context: Context, data_dict: DataDict) -> ActionResult.ResourceViewReorder:
     '''Reorder resource views.
@@ -301,8 +302,8 @@ def package_update(
         dataset_changed = dict(
             original_package, metadata_modified=None, resources=None
             ) != dict(data_dict, metadata_modified=None, resources=None)
-    if not dataset_changed and original_package.get('resources'
-            ) == data_dict.get('resources'):
+    if not dataset_changed and original_package.get('resources') \
+            == data_dict.get('resources'):
         return pkg.id if return_id_only else original_package
 
     res_deps = []
@@ -851,6 +852,7 @@ def group_update(context: Context, data_dict: DataDict) -> ActionResult.GroupUpd
     '''
     return _group_or_org_update(context, data_dict)
 
+
 def organization_update(
         context: Context, data_dict: DataDict) -> ActionResult.OrganizationUpdate:
     '''Update an organization.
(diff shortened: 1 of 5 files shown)


ThrawnCA:
@wardi How does this look?

ThrawnCA:
I'm not sure why Codecov is reporting lower coverage. As far as I can see, I've covered each branch.

duttonw (approved):

ckan/logic/action/update.py
```diff
@@ -231,6 +232,32 @@ def resource_view_reorder(
     return {'id': id, 'order': new_order}
 
 
+def package_reindex(
+        context: Context, data_dict: DataDict) -> None:
+    '''Rebuild the search index for a dataset (package).
+
+    You must be authorized to edit the dataset
+    and the group(s) it belongs to.
```
wardi, line 240: Pretty sure you don't need to have permission on the groups

ckan/logic/auth/update.py
```diff
@@ -11,6 +11,11 @@
 from ckan.types import Context, DataDict, AuthResult
 
 
+@logic.auth_allow_anonymous_access
+def package_reindex(context: Context, data_dict: DataDict) -> AuthResult:
+    return package_update(context, data_dict)
```
wardi: Use this so that overridden auth functions are called
```suggestion
    return authz.is_authorized('package_update', context, data_dict)
```

ckan/logic/action/update.py
```diff
@@ -231,6 +232,32 @@ def resource_view_reorder(
...
+    :type id: string
+    '''
+    name_or_id = data_dict.get('id') or data_dict.get('name')
+    if name_or_id is None:
+        raise ValidationError({'id': _('Missing value')})
+
+    pkg = model.Package.get(name_or_id)
+    if pkg is None:
+        raise NotFound(_('Package was not found.'))
+    context["package"] = pkg
```
wardi: There's no need to put the package in the context for the auth function (it's no longer used) or to return it to the caller (relying on side effects leads to hard-to-debug code)
```suggestion
```

ckan/logic/action/update.py
```diff
@@ -231,6 +232,32 @@ def resource_view_reorder(
...
+
+    data_dict["id"] = pkg.id
```
wardi: We're not validating the other parameters, so as written, arbitrary parameters will be passed to a potentially customized auth function. Let's pass through only the parameters we've checked just to be safe.
```suggestion
    data_dict = {"id": pkg.id}
```

ckan/logic/action/update.py
ThrawnCA: This is copied from the description of package_update, whose auth we're reusing. Does it need to be changed there?

ThrawnCA:
This was actually proposed as https://github.com/ckan/ckan/issues/4897

PR #9446, opened 2026-07-22
Fix creating values in single-value autocomplete fields

TomeCirun:
Generic single-value autocomplete fields previously allowed users to select a value that was not returned by the autocomplete API. After the Select2 4 migration, CKAN still defines createTag, but its custom data adapter is only decorated with Select2’s Tags adapter when the field uses the multi-value tags option. As a result, createTag is never called for single-value autocomplete fields.

Resource formats are single-value autocomplete fields, so enabling tags would incorrectly turn them into multi-value fields. This change installs the Tags data adapter whenever createtags is enabled, while keeping multiple-value and tokenization behavior exclusive to fields using tags.

### Features:

- [ ] includes API changes


TomeCirun:
I’m fixing the failing Cypress test

PR #9451, opened 2026-07-24
refactor: mark unused helpers as deprecated

mutantsan:
Mark unused helpers as deprecated.


amercader:
I think this is good to merge into 2.12 but we need an entry in the CHANGELOG with the deprecated helpers

mutantsan:
@amercader done

Issue #7188, opened 2022-11-09
Download Button needs discernible text


mgifford:
**CKAN version**
https://demo.ckan.org/dataset/sample-dataset-1/resource/e687245d-7835-44b0-8ed3-0827de123895

**Describe the bug**
Title: WCAG 4.1.2: Ensures buttons have discernible text (.dropdown-toggle)
Tags: Accessibility, WCAG 4.1.2, button-name

Issue: Ensures buttons have discernible text (button-name - https://accessibilityinsights.io/info-examples/web/button-name)

Target application: Sample Dataset - sample-linked.csv - CKAN Demo - https://demo.ckan.org/dataset/sample-dataset-1/resource/e687245d-7835-44b0-8ed3-0827de123895

Element path: .dropdown-toggle

Snippet: <button class="btn btn-primary dropdown-toggle" data-toggle="dropdown"> <span class="caret"></span> </button>

How to fix: 
Fix any of the following:
  Element does not have inner text that is visible to screen readers
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
  Element has no title attribute
  Element's default semantics were not overridden with role="none" or role="presentation"

Environment: Microsoft Edge version 107.0.1418.35

====

This accessibility issue was found using Accessibility Insights for Web 2.35.0 (axe-core 4.4.1), a tool that helps find and fix accessibility issues. Get more information & download this tool at http://aka.ms/AccessibilityInsights.

PR #9471, opened 2026-08-11
a11y: add accessible names to icon-only search and copy buttons (refs #7188)

opensource-joe:
Refs #7188 (buttons need discernible text).

## What

Two icon-only buttons have no accessible name, so screen readers announce them only as "button" (WCAG 4.1.2 Name, Role, Value):

- `ckan/templates/user/snippets/user_search.html` — the search submit button (`fa-search`). The search *input* has a label, but the submit button does not.
- `ckan/templates/user/snippets/api_token_create_form.html` — the "copy to clipboard" button shown for a newly created API token (`fa-copy`).

## Change

Add a translatable `aria-label` to each button and mark the decorative icons `aria-hidden="true"`.

This continues the button-name work referenced in #7188 (see also #9469, which fixed the resource-actions toggle). Includes a changelog fragment. Non-visual change; no layout or styling impact.

### DIFF
--- a/changes/9471.bugfix
+++ b/changes/9471.bugfix
@@ -1,0 +1,1 @@
+Add accessible names to the icon-only user-search submit button and the API-token "copy to clipboard" button so screen readers announce them with discernible text (WCAG 4.1.2).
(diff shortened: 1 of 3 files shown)


amercader:
Thanks for this @opensource-joe . In the future, PRs that are closely related like these three can be combined into one with separate commits. cheers

PR #9472, opened 2026-08-12
Replace bleach with nh3

smotornyuk:
[Bleach is no longer maintained](https://github.com/mozilla/bleach/issues/698). 

There is one library that resembles bleach API, [nh3](https://nh3.readthedocs.io/en/latest/). There are some distinctions, but generally it fits into CKAN as we are using bleach sparingly.

The main risk is the fact that nh3 is not written in python, and instead it provides bindings for the rust library and is built with maturin. From my experience, maturin provides decent support for platforms and looks like there are enough wheels available on pypi to work without extra steps on popular Linux distros such as ubuntu, centos, AWS linux, arch, debian, etc (and MacOS + Windows, btw). 
Users of unsupported distors may be required to install rust compiler in order to build this library, but I was unable to identify any popular image of linux that would require such step, so it should not be a frequent scenario.

Apart from that, there is a number of differences, that will be observed unless we deliberately replicate behavior of bleach. Here's the changes that I identified:

### nh3 allows more tags by default

Bleach allows 12 tags: `['a', 'abbr', 'acronym', 'b', 'blockquote', 'code', 'em', 'i', 'li', 'ol', 'strong', 'ul']`

nh3 allows 75 tags: `['a', 'abbr', 'acronym', 'area', 'article', 'aside', 'b', 'bdi', 'bdo', 'blockquote', 'br', 'caption', 'center', 'cite', 'code', 'col', 'colgroup', 'data', 'dd', 'del', 'details', 'dfn', 'div', 'dl', 'dt', 'em', 'figcaption', 'figure', 'footer', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'header', 'hgroup', 'hr', 'i', 'img', 'ins', 'kbd', 'li', 'map', 'mark', 'nav', 'ol', 'p', 'pre', 'q', 'rp', 'rt', 'rtc', 'ruby', 's', 'samp', 'small', 'span', 'strike', 'strong', 'sub', 'summary', 'sup', 'table', 'tbody', 'td', 'th', 'thead', 'time', 'tr', 'tt', 'u', 'ul', 'var', 'wbr']`

All tags from bleach are included into this list and there are 63 more tags. They look safe enough: table, cite, dl, etc. - looks like they provide reacher output without any unsafe attributes.

### Allowed attributes are different

Bleach allows just few attributes: `{'a': ['href', 'title'], 'abbr': ['title'], 'acronym': ['title']}`

nh3 uses much wider list:
```py
{'tbody': {'align', 'char', 'charoff'},
 'blockquote': {'cite'},
 'del': {'cite', 'datetime'},
 'table': {'align', 'char', 'charoff', 'summary'},
 'thead': {'align', 'char', 'charoff'},
 'ol': {'start'},
 'colgroup': {'align', 'char', 'charoff', 'span'},
 'tr': {'align', 'char', 'charoff'},
 'col': {'align', 'char', 'charoff', 'span'},
 'bdo': {'dir'},
 'hr': {'align', 'size', 'width'},
 'img': {'align', 'alt', 'height', 'src', 'width'},
 'q': {'cite'},
 'tfoot': {'align', 'char', 'charoff'},
 'ins': {'cite', 'datetime'},
 'td': {'align', 'char', 'charoff', 'colspan', 'headers', 'rowspan'},
 'th': {'align', 'char', 'charoff', 'colspan', 'headers', 'rowspan', 'scope'},
 'a': {'href', 'hreflang'}}
```

I've added `title` to the allowed attributes of `a` and `img` tags; everything else is kept as is.

### Signature of the `clean` function

Bleach clean:
```py
  clean(text, tags=ALLOWED_TAGS, attributes=ALLOWED_ATTRIBUTES,
        protocols=ALLOWED_PROTOCOLS, strip=False, strip_comments=True,
        css_sanitizer=None)
```

nh3 clean:
```py 
 clean(html, tags=None, clean_content_tags=None, attributes=None,
        attribute_filter=None, strip_comments=True, link_rel='noopener noreferrer',
        generic_attribute_prefixes=None, tag_attribute_values=None,
        set_tag_attribute_values=None, url_schemes=None, allowed_classes=None,
        filter_style_properties=None, url_relative=None)
```

they are different,  but we are only using `tags`( :heavy_check_mark: ), `attributes`( :heavy_check_mark: ), and `strip`( :exclamation: ) arguments of `bleach.clean`. The latter is responsible for keeping HTML-escaped tags(i.e., `<xxx>` is replaced with `&lt;xxx&gt;`) - nh3 has no such thing, and it always strips disallowed tags. Mainly, we also expecting that bleach would strip such tags, but for some reason, DataPusher's logs snippet keeps escaped html instead of stripping it completely. I don't see any other place in core that can be affected by this difference.

### other differences

nh3 does additional work apart from cleaning. Most noticeable, it adds `rel='nofollow noopener'` to links. Here's the same input that is processed differently by bleach and nh3:

Input: `<a href="/test" onclick="alert(1)">Link</a>`   
Bleach: `<a href="/test">Link</a>`
NH3:      
```html
<a href="/test" rel="noopener noreferrer">Link</a> 
```

I kept this behavior as it seems appropriate to mark potentially unsafe links with this no-sniff option.

Also, nh3 removes `script` and `style` tags alongside with their content, while bleach removed the wrapping tag, but kept the content. It's possible to restore such behavior, but I don't see value in keeping scripts and styles as plain-text.

### DIFF
--- a/changes/9472.misc
+++ b/changes/9472.misc
@@ -1,0 +1,1 @@
+HTML cleaning is done with ``nh3`` instead of ``bleach`` from now on. More tags and attributes are allowed by default and ``clean_html`` now strips unknown tags instead of escaping HTML-entities.
--- a/changes/9472.misc
+++ b/changes/9472.misc
@@ -1,1 +1,3 @@
-HTML cleaning is done with ``nh3`` instead of ``bleach`` from now on. More tags and attributes are allowed by default and ``clean_html`` now strips unknown tags instead of escaping HTML-entities.
+HTML cleaning is done with ``nh3`` instead of ``bleach`` from now on. 
+More tags and attributes are allowed by default and ``clean_html`` now strips unknown tags instead of escaping HTML-entities.
+``render_markdown`` with ``allow_html=False`` now strips ``script`` and ``style`` tags alongside with  their content; previously, only the tag was removed, while content was kept.
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -29,7 +29,7 @@ from typing import (
 
 import dominate.tags as dom_tags
 from markdown import markdown
-from bleach import clean as bleach_clean, ALLOWED_TAGS, ALLOWED_ATTRIBUTES
+from nh3 import clean as nh3_clean, ALLOWED_TAGS, ALLOWED_ATTRIBUTES
 from ckan.common import asbool, config, current_user
 from flask import flash, has_request_context, current_app
 from flask import get_flashed_messages as _flask_get_flashed_messages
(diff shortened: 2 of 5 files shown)


Zharktas:
Did we want this in 2.12 as well or not ?

smotornyuk:
/backport

Yes, we need this in 2.12 or else we'll be locked with bleach till the 2.13/3.0

Issue #9473, opened 2026-08-13
JS translations failing for BCP 47 locale codes


mutantsan:
Recently we did this [change](https://github.com/ckan/ckan/pull/8698), but it seems that it caused a bug. 

<img width="1038" height="160" alt="Image" src="https://github.com/user-attachments/assets/8b57ed9f-8ab5-4da3-acdf-641e99c7c842" />

The lang code on an `html` tag looks like that `lang="uk-UA"`, but this lang code is used by the JS. Inside the `ckan.initialize` we call `client.getLocaleData(locale)`

```js
getLocaleData: function (locale, success, error) {
  var url = this.url('/api/i18n/' + (locale || ''));
  return jQuery.getJSON(url).then(success, error);
}
```

This will call the `i18n_js_translations` function, which should return translations for the given lang. But the files are stored with underscores, not hyphens. 

```py
def i18n_js_translations(
        lang: str,
        ver: int = API_REST_DEFAULT_VERSION) -> Union[str, Response]:

    if lang not in get_locales_from_config():
        return _finish_bad_request('Unknown locale: {}'.format(lang))

    js_translations_folder = get_js_translations_dir()

    source = os.path.join(js_translations_folder, f"{lang}.js")
    if not os.path.exists(source):
        return "{}"
    translations = io.open(source, "r", encoding="utf-8").read()
    return _finish_ok(LazyJSONObject(translations))
```
An obvious solution would be to do something like `lang = lang.replace('-', '_')` in here to revert the changes of `unix_locale_to_bcp47` helper. But I don't really know what those lang code shenanigans are :)

PR #9474, opened 2026-08-13
fix: fix js translations failing for bcp 47 locales

mutantsan:
Fix for https://github.com/ckan/ckan/issues/9473

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9473.bugfix
+++ b/changes/9473.bugfix
@@ -1,0 +1,1 @@
+Fix JS translations failing for BCP 47 locale codes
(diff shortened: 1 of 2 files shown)


wardi:
A more robust fix (maybe just for master) would be to store the JS translations as the BCP-47 language codes instead of unix locales. This would mean one less `replace` hack and would continue to work if a site overrides the new `unix_locale_to_bcp47` helper. The helper was created specifically so that sites could override it.

PR #9475, opened 2026-08-13
Use pglast instead of sqlparse for datastore_search_sql

bluepython508:
`pglast` uses the same underlying parser as PostgreSQL itself, so this should reduce the risk of parser differences leading to SQL injections.


bluepython508:
The remaining failures should be down to my not having updated `requirements.txt`, because I couldn't get pip-compile to behave

amercader:
@bluepython508 add the following to your branch:

```diff
diff --git a/requirements.txt b/requirements.txt
index 4532fecb30..ac2a54ee6d 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -88,6 +88,8 @@ packaging==26.2
     # via -r requirements.in
 passlib==1.7.4
     # via -r requirements.in
+pglast==8.4
+    # via -r requirements.in
 pluggy==1.6.0
     # via file-keeper
 polib==1.2.0
@@ -133,8 +135,6 @@ sqlalchemy==2.0.51
     # via
     #   -r requirements.in
     #   alembic
-sqlparse==0.5.5
-    # via -r requirements.in
 tomli==2.4.1
     # via alembic
 typing-extensions==4.16.0

```

To add a new requirement to requirements.txt:

```
pip install pip-tools
```

and

```
pip-compile -P pglast requirements.in
```

ckanext/datastore/backend/postgres.py
```diff
@@ -1890,6 +1908,101 @@ def search(context: Context, data_dict: dict[str, Any]):
...
+    # and we're unlikely to have a value for it anyway
+    result = context['connection'].execute(sa.text('''
+        SELECT table_catalog, table_schema, table_name
+        FROM information_schema.view_table_usage
+        WHERE view_schema = :schema and view_name = :name
+    ''').bindparams(schema= schema or 'public', name=name))
+
+    return {
+        (table.table_schema, table.table_name) for table in result
+    } or {(schema, name)}
```
wardi: This function should accept multiple table names and return corresponding resource ids. As `sanitize_sql` is written we're making a round trip to the DB for every time a table appears in a node, even if it's the same table multiple times in the same query.

Ideally we make a single round-trip to collect all the resource ids corresponding to a set of table names.

This is useful enough that it could even be its own action in the API, e.g.
```
$ ckanapi action datastore_aliases 'names:["foo", "bar"]'
{
  "foo": "bdbb26cc-c621-47df-859f-9aa6cbf2407d",
  "bar": "afb95ad1-1e88-45a1-829a-5593a7ffe689",
}
```

ckanext/datastore/backend/postgres.py
```diff
@@ -1890,6 +1908,101 @@ def search(context: Context, data_dict: dict[str, Any]):
...
+    set[tuple[str, ...]]
+]:
+    """Ensure SQL is a single safe query, returning tables and functions used
+
+    Parses the query using the upstream PostgreSQL parser, checks
+    the AST nodes provided against an allow list of safe types, and
+    reserializes the query.
+
+    This doesn't touch the database, so it cannot resolve views into
+    their underlying tables or resolve equivalent names.
```
wardi: this is no longer true

ckanext/datastore/tests/test_db.py
```diff
@@ -249,3 +252,103 @@ def test_worker_datastore_access(self, app):
...
+            "CREATE VIEW test_e AS SELECT * from test_a INNER JOIN test_b ON id_a = id_b"
+        ]
+        for create_table_sql in create_tables:
+            session.execute(sa.text(create_table_sql))
+
+        test_cases = [
+            (u"SELECT * FROM test_a", {(None, "test_a")}),
+            (u"SELECT * FROM public.test_a", {("public", "test_a")}),
+            (u'SELECT * FROM "TEST_C"', {(None, "TEST_C")}),
+            (u'SELECT * FROM public."TEST_C"', {("public", "TEST_C")}),
```
wardi: don't need `u` prefixes on strings anymore

ckanext/datastore/tests/test_helpers.py
```diff
@@ -33,13 +28,13 @@ def test_is_single_statement(self):
             'SELECT * FROM "bartable";',
             'SELECT * FROM "bart;able";',
             "select 'foo'||chr(59)||'bar'",
+            r"""SELECT ' \'' as A, * FROM "foo"; SELECT * FROM "abc" --'""",
         ]
 
         multiples = [
-            "SELECT * FROM abc; SET LOCAL statement_timeout to"
+            "SELECT * FROM abc; SET LOCAL statement_timeout to",
```
wardi, line 35: good catch

ckanext/datastore/tests/test_search.py
```diff
@@ -1471,7 +1471,7 @@ def test_works_with_semicolons_inside_strings(self):
             "records": [{"author": "bob"}, {"author": "jane"}],
         }
         helpers.call_action("datastore_create", **data)
-        sql = 'SELECT * FROM public."{0}" WHERE "author" = \'foo; bar\''.format(
+        sql = 'SELECT * FROM "{0}" WHERE "author" = \'foo; bar\''.format(
```
wardi: is removing `public` from these tests still necessary?

ckanext/datastore/tests/test_search.py
```diff
@@ -1541,7 +1541,7 @@ def test_quoted_allowed_functions_are_case_sensitive(self):
         )
         helpers.call_action("datastore_search_sql", sql=sql)
 
-        sql = 'SELECT CoUnT(*) from "{}"'.format(
+        sql = 'SELECT "CoUnT"(*) from "{}"'.format(
```
wardi, line 1544: why is quoting `count` necessary now?

wardi (dismissed):

ckanext/datastore/backend/postgres.py
```diff
@@ -1890,6 +1908,101 @@ def search(context: Context, data_dict: dict[str, Any]):
...
+        return
+    yield ast
+    for value in ast.values():
+        if isinstance(value, (tuple, dict)):
+            for node in flatten_tuples(value):
+                if node is None:
+                    continue
+                yield from nodes(node)
+
+def referenced_tables(context: Context, schema: str | None, name: str):
```
wardi: if we're introducing a new function that only needs a connection from the context, why not make the connection a parameter instead of passing the whole context?

wardi: (same comment about sanitize_sql below)

ckanext/datastore/backend/postgres.py
```diff
@@ -1890,6 +1908,101 @@ def search(context: Context, data_dict: dict[str, Any]):
...
+            )
+        elif node['@'] == 'FuncCall':
+            funcname = (node['funcname'], ) \
+                if isinstance(node['funcname'], dict) else node['funcname']
+            if not all(part['@'] == 'String' for part in funcname):
+                raise toolkit.ValidationError(f'Invalid function name: {funcname}')
+            functions.add(tuple(n['sval'] for n in funcname))
+
+    serialized = pglast.stream.RawStream()(stmts[0])
+
```
EricSoroos:
```suggestion
            tables.update(
                (node['schemaname'], node['relname'])
            )
        elif node['@'] == 'FuncCall':
            funcname = (node['funcname'], ) \
                if isinstance(node['funcname'], dict) else node['funcname']
            if not all(part['@'] == 'String' for part in funcname):
                raise toolkit.ValidationError(f'Invalid function name: {funcname}')
            functions.add(tuple(n['sval'] for n in funcname))

    serialized = pglast.stream.RawStream()(stmts[0])
    _referenced_tables = set(referenced_tables(context, schema, relname) for (schema,relname) in tables)
    return serialized, _referenced_tables, functions
```

ckanext/datastore/backend/postgres.py
EricSoroos: If we're potentially calling this multiple times with the same tables, then it should be deduped in set form prior to the call, or potentially memoized. It's possible to do this in one shot, but it's not nearly as clear of a query if we're allowing for anything other than a specific schema. 

I don't think we're creating a new connection each time, so this should be single digit milliseconds per query.

(I think we'd have to create a table of schema/relname pairs, and join that to the view_table_usage, then coalesce the original values back out.  Otherwise we'd risk returning public.foo if we were resolving private.foo and public.bar.)

EricSoroos: It's also not guaranteed that the resulting tables will be resource ids. I'd want to be very careful about exposing this, as the information leakage from it will probably bite us at some point.

ckanext/datastore/tests/test_search.py
EricSoroos: The semantics are different.  Pglast parses `CoUnT` as `count`, as postgres would, where the previous solution was a text processing regex.

ckanext/datastore/backend/postgres.py
wardi: resource ids and things appearing in `ckan.datastore.public_table_search` are the only things we can check permissions against, right?

wardi: we can pass the list of names as strings and use unnest, e.g.:
```sql
select al, t.name
from unnest(array['foo', 'baz']) al
left join _table_metadata t on t.alias_of = al;
 al  | name 
-----+------
 foo | bar
 baz | 
```

ckanext/datastore/tests/test_search.py
bluepython508: No; I've readded them

ckanext/datastore/backend/postgres.py
```diff
@@ -1890,6 +1909,102 @@ def search(context: Context, data_dict: dict[str, Any]):
...
+        return
+    yield ast
+    for value in ast.values():
+        if isinstance(value, (tuple, dict)):
+            for node in flatten_tuples(value):
+                if node is None:
+                    continue
+                yield from nodes(node)
+
+def referenced_tables(connection: Any, schema: str | None, name: str):
```
wardi: this should be a `sqlalchemy.engine.Connection`

ckanext/datastore/backend/postgres.py
wardi: I realized we don't need a new action API for resolving aliases, since we can already:
```bash
$ ckanapi action datastore_search resource_id=_table_metadata \
   fields=name,alias_of filters:'{"name":["foo","bar"]}'
...
  "records": [
    {
      "alias_of": "foo",
      "name": "bar"
    },
    {
      "alias_of": null,
      "name": "foo"
    }
...
```

wardi:
@amercader I've implemented the last change I was looking for on this PR. Would you like to take a look before it's merged?

PR #9491, opened 2026-08-18
fix: markdown_extract returns only '...' when a description starts with a long token

### LINKED ISSUE #9488: markdown_extract truncates on the escaped text, so descriptions containing & or < lose a third of the extract
## CKAN version

2.11.5 (verified against the installed package, not a checkout). `markdown 3.10.2`, `bleach 6.4.0`, matching `requirements.txt`.

## Describe the bug

`h.markdown_extract()` measures its truncation width against the **HTML-escaped** text, so any description containing `&`, `<` or `>` is cut far shorter than the requested length.

`ckan/lib/helpers.py:1402`:

```python
plain = bleach_clean(markdown(text), tags=(), strip=True)
if not extract_length or len(plain) < extract_length:
    return literal(plain)
return literal(str(shorten(plain, width=extract_length, placeholder='...')))
```

`bleach` leaves entities in the result, so `R&D` is five characters longer than it looks: `R&amp;D` costs 7 against the budget and displays as 3. `shorten` then measures the escaped string.

The visible effect is on dataset search listings, where `templates/snippets/package_item.html:14` calls it with `extract_length=180`.

### Steps to reproduce

```python
from ckan.lib.helpers import markdown_extract

prefix = "Ozone and NO2 columns measured at Uccle. "
with_amp    = prefix + "R&D " * 40
without_amp = prefix + "RxD " * 40

print(len(str(markdown_extract(with_amp, 180)).replace("&amp;", "&")))  # 111
print(len(str(markdown_extract(without_amp, 180))))                     # 179
```

Two descriptions of the same length, differing only in whether a character needs escaping.

### Expected behavior

Both should keep roughly the same number of **visible** characters, near the 180 requested. The one containing ampersands keeps **111**, losing about a third of its extract.

The more entity-dense the text, the worse it gets. This affects any description using `&` in an organisation name, `<` or `>` in a value range, or quoted attributes.

### Additional details

The fix presumably involves measuring the unescaped text, or unescaping before `shorten` and re-escaping after. I have not opened a PR because the choice affects what "length" means in the API and that seems worth a maintainer's view first. Happy to implement whichever direction you prefer.

Found while investigating #9366, which is a different problem. This one is separate and reproducible on its own.

### LINKED ISSUE #9489: markdown_extract returns only '...' when a description starts with a token longer than the extract length
## CKAN version

2.11.5 (verified against the installed package, not a checkout). `markdown 3.10.2`, `bleach 6.4.0`, matching `requirements.txt`.

## Describe the bug

If a dataset description begins with a single token longer than the extract length, `h.markdown_extract()` returns **nothing but the placeholder**, so the dataset appears in search listings with no description at all.

`ckan/lib/helpers.py:1402` uses `textwrap.shorten`, which drops any word that does not fit rather than cutting it. When the first word alone exceeds the width there is nothing left to keep.

A long URL, a DOI, a filename or a generated identifier at the start of a description is enough to trigger it.

### Steps to reproduce

```python
from ckan.lib.helpers import markdown_extract

print(repr(str(markdown_extract("B" * 400, 180))))            # '...'
print(repr(str(markdown_extract("B" * 400 + " tail", 180))))  # '...'
```

A trailing word does not help, because the first token still cannot fit.

### Expected behavior

Something readable. Either fall back to a hard character cut when the leading word is longer than the width, or break the token, rather than returning only `...`.

At present the dataset looks as though it has no description, which is misleading in a search listing where the extract is the only signal about content.

### Additional details

`templates/snippets/package_item.html:14` is the visible caller, with `extract_length=180`.

`textwrap.shorten` is documented to behave this way, so this is about whether it is the right tool here rather than a bug in the standard library.

I have not opened a PR because the sensible fallback is a judgement call: a hard cut is simplest, but changes behaviour for everyone, and it interacts with the entity-width problem I have raised separately. Glad to implement whichever you prefer.

Found while investigating #9366, which is a different problem.

opensource-joe:
Fixes #9489.

`textwrap.shorten` drops any word wider than the requested width rather than cutting it, so a description opening with a long URL, DOI or generated identifier comes back as nothing but the `...` placeholder and the dataset reads as having no description. When `shorten` returns only the placeholder, cut the token instead.

On master (`0731a7a4ab`) such a description returns exactly `'...'`, three characters. With this change it returns 190 ending in `...`. Descriptions that fit, and those starting with a normal word, are unchanged.

This PR used to fix #9488 as well. `html.unescape` landed in `64e999c8` and covers that, so only #9489 is left. I rebuilt the branch on master rather than rebasing, because master also dropped the `literal()` wrapper and a rebase would have put it back.

`ckan/tests/lib/test_helpers.py`: 184 passed, 1 skipped. Reverting the change alone turns the new test red.


opensource-joe:
The red job is `Test (Pytest) / pytest (5)`, failing on `TestPackageQuery::test_overall` with `assert 1 == 0` at the `title:peace` line. I do not think it is related to this change. I think that test is flaky by construction.

`factories.Dataset` sets `title = factory.LazyFunction(lambda: fake.sentence(nb_words=5))`, and faker's `en_US` lorem list is 971 words, one of which is `peace`. The `warandpeace` dataset in that test takes its title from that draw, so every so often it gets a title containing the word, and `title:peace` then matches 1 instead of 0. Roughly one run in two hundred by my arithmetic. The query has a colon in it, so it is a genuine fielded search on `title` rather than dismax.

I checked this against unpatched master: creating the same two datasets and pinning that one title to `Reach whom fill peace bring.` makes `title:peace` return 1, with nothing else changed.

Nothing here touches search. `markdown_extract` is a template helper and does not appear anywhere under `ckan/lib/search/`.

Would you mind re-running that one job? I am happy to open a separate issue or PR to pin the title in `test_overall` if you would like it made deterministic, but it seemed wrong to fold an unrelated test change into this one.

opensource-joe:
Rewritten on current master and narrowed to #9489.

`html.unescape` landed in `64e999c8` and fixes #9488, so that half of this is no longer needed. I rebuilt the branch rather than rebasing, because master also dropped the `literal()` wrapper and a rebase would have reintroduced it.

#9489 still reproduces on master: a description starting with a long URL or DOI returns exactly `'...'`. Test and changelog entry included.

PR #9495, opened 2026-08-25
fix: more "compatible" uploads_enabled helper

### LINKED ISSUE #9490: CKAN 2.12 makes non-backward-compatible change to 'uploads_enabled'
## CKAN version

2.12

## Describe the bug

The signature of the 'uploads_enabled' core helper is changed in 2.12 in such a way that templates cannot be compatible with both old and new, except by checking versions and making different calls. This seems like unnecessary complication, breaking eg the 'scheming' plugin. Many sites will not care about the distinction between different upload types, especially if they are using a custom storage plugin.

Wouldn't it make more sense for an `uploads_enabled` call with no parameters to instead go straight to checking for an IUploader plugin, similarly to the pre-2.12 behaviour?

### Steps to reproduce

* Install the `dev-v2.12` CKAN branch.
* Install the latest ckanext-scheming release, 3.1.0.
* Browse to a resource and find that the upload button is disabled.

### Expected behavior

Calls to `uploads_enabled` with no parameters should retain their existing behaviour.

smotornyuk:
Fixes #9490

Initially, I insisted on the strict rejection of the `h.uploads_enabled` call without specific upload types. In this way, we have controlled and predictable behavior of the helper.

But @ThrawnCA highlighted a valid point: some extensions are using a custom uploader for a custom endpoint. They will handle one specific upload type on a completely new page, and they don't want to make changes like `uploads_enabled(type) if ckan_version(2.12) else uploads_enabled()`.

Even though I still think that for such extensions it's a good idea to migrate now so that everything likely remains functional when v3.0 is out, we can make this immediate migration optional. For this, I made the following changes in `h.uploads_enabled`:

* When `object_type` is specified(`h.uploads_enabled(object_type)`) - it remains unchanged. In this way, CKAN core and changes we made for scheming and other plugins work exactly as planned.
* When `object_type` is not provied(`h.uploads_enabled()`), logic is changed. Previously, it immediately returned `False`. Now it returns `True` when:
  * `ckan.storage_path` is specified. In this case, even if there is no dedicated storage, `get_uploader`/`get_resource_uploader` will [return the classic uploader](https://github.com/ckan/ckan/blob/master/ckan/lib/uploader.py#L87-L88) class that handles uploads as it was done in the past.
  * A custom IUploader implementation exists. In this case, this [custom implementation will handle the upload](https://github.com/ckan/ckan/blob/master/ckan/lib/uploader.py#L78-L81). This case is fragile, but at least it gives the extension a chance to remain functional without any changes. It's up to the maintainer to test whether functionality is broken after the v2.12 upgrade. Previously, we enforced migration to `uploads_enabled` with object type; now we leave the final decision regarding this migration to the maintainer.

### DIFF
--- a/ckan/lib/helpers.py
+++ b/ckan/lib/helpers.py
@@ -2518,12 +2518,19 @@ def uploads_enabled(object_type: str | None = None) -> bool:
         group, user or admin
     :type object_type: string
     """
-    if not object_type:
-        log.warning("h.uploads_enabled call without object_type is non supported")
-        return False
-
     if not config["ckan.uploads_enabled"]:
         return False
+
+    has_classic_uploader = bool(
+        config["ckan.storage_path"]
+        or any(plugin for plugin in p.PluginImplementations(p.IUploader))
+    )
+
+    if not object_type:
+        log.warning(
+            "h.uploads_enabled must be called with object_type. Switch to legacy check"
+        )
+        return has_classic_uploader
 
     storage_name = config.get(f"ckan.files.default_storages.{object_type}")
     if not storage_name:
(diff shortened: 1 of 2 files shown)


wardi:
@smotornyuk let's use `storage_name` to be consistent with the new upload code, otherwise LGTM

smotornyuk:
Done + mentioned planned changes in log.warning.

PR #9500, opened 2026-08-25
Fs storage improvements

smotornyuk:
Fixes for FS adapter

* If there is any public storage that includes another private storage, stop CKAN and report an error during startup
* Add new validator `starts_with` and use it to ensure that FS storages uses only absolute path. I decided to use it instead of preserving Windows compatibility, because `starts_with` sounds more usful than `is_path_absolute`
* add new config option `follow_symlinks` to FS storage(disabled by default)
* Update file-keeper
* Few tests to cover changes

> [!NOTE] 
> This PR addressed vulnerabilities that were not released in a public CKAN version, only in the master branch. Thanks to @phuongmai1212 and @AuQuangDuc for reporting them

### DIFF
--- a/ckan/lib/files/__init__.py
+++ b/ckan/lib/files/__init__.py
@@ -212,9 +212,9 @@ def collect_storages() -> dict[str, fk.Storage]:  # noqa: C901
         if not path.endswith(os.path.sep):
             path += os.path.sep
 
-        grouped[storage.settings.type][storage.settings.public].add(path)
+        storages_grouped_by_privacy[storage.settings.type][storage.settings.public].add(path)
 
-    for adapter, group in grouped.items():
+    for adapter, group in storages_grouped_by_privacy.items():
         if not group[False] or not group[True]:
             continue
 
--- a/ckan/lib/files/__init__.py
+++ b/ckan/lib/files/__init__.py
@@ -125,7 +125,7 @@ def _storages_from_config(
     return result
 
 
-def collect_storages() -> dict[str, fk.Storage]:
+def collect_storages() -> dict[str, fk.Storage]:  # noqa: C901
     """Initialize storages.
 
     :returns: mapping with storages
@@ -204,7 +204,7 @@ def collect_storages() -> dict[str, fk.Storage]:
             # functionality and we are won't interfere
             continue
 
-        path = storage.settings.path or ""
+        path = str(storage.settings.path or "")
 
         # for simplicity, append slash to the path. It allows using
         # `str.startswith` instead of `os.path.commonpath` which will complain
--- a/ckan/lib/files/__init__.py
+++ b/ckan/lib/files/__init__.py
@@ -223,7 +223,7 @@ def collect_storages() -> dict[str, fk.Storage]:  # noqa: C901
                 msg = (
                     f"{adapter} storage with path {private} is included into another"
                     + " storage with public access. Make sure paths of public"
-                    + " and private storages do not intersect."
+                    + " and private storages do not overlap."
                 )
                 raise CkanConfigurationException(msg)
 
--- a/ckan/lib/files/__init__.py
+++ b/ckan/lib/files/__init__.py
@@ -194,7 +194,7 @@ def collect_storages() -> dict[str, fk.Storage]:  # noqa: C901
     # check that there are no public storages that point to the parent folder
     # of another private storage, unintentionally exposing private files in
     # this way
-    grouped: dict[str, dict[bool, set[str]]] = defaultdict(
+    storages_grouped_by_privacy: dict[str, dict[bool, set[str]]] = defaultdict(
         lambda: {True: set(), False: set()}
     )
     for storage in result.values():
--- a/ckan/lib/files/__init__.py
+++ b/ckan/lib/files/__init__.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
-
 import copy
 import logging
 import os
+from collections import defaultdict
 from typing import cast, Any
+
 
 import file_keeper as fk
 from file_keeper import Registry, Upload, adapters, exc, ext, make_storage, make_upload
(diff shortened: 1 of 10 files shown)


smotornyuk:
@wardi , @amercader , I made changes that we discussed today. [symlinks](https://github.com/DataShades/file-keeper/blob/master/src/file_keeper/default/adapters/fs.py#L450-L453) and [common prefix](https://github.com/DataShades/file-keeper/blob/master/src/file_keeper/core/storage.py#L740-L753) are handled inside file-keeper, everything else is done on CKAN side

ckan/lib/files/default/fs.py
```diff
@@ -41,12 +41,25 @@ class FsStorage(base.Storage, fs.FsStorage):
...
+
+        declaration[key.path].required().append_validators(
+            'starts_with("/")'
+        ).set_description(
+            "Absolute path to the storage's root."
+            "\nWill cause a configuration error if not absolute."
+        )
+        declaration.declare_bool(key.follow_symlinks).set_description(
+            "Whether to follow symlinks when accessing a file."
+        )
```
amercader, line 61: Just checking that this is setting the config option correctly, I get this when inspecting a fs storage `settings` value:

```
pp storages.members["resources"].settings
Settings(type='ckan:fs',
         name='resources',
         overwrite_existing=True,
         path='/home/adria/dev/projects/ckan-py310/data2/resources',
         location_transformers=[],
         disabled_capabilities=[],
         initialize=True,
         skip_in_place_move=True,
         skip_in_place_copy=True,
         hashing_algorithm='md5',
         _extra_settings={'follow_symlinks': False},
         supported_types=[],
         max_size=10485760,
         public=False)
```

Is it in `_extra_settings` because it's a specific `fs` storage config?

ckan/lib/files/__init__.py
```diff
@@ -182,12 +183,50 @@ def collect_storages() -> dict[str, fk.Storage]:
...
         and config["ckan.storage_path"]
     ):
         log.warning(
             "All uploads are handled by storages. `ckan.storage_path` has no effect"
         )
 
+    # check that there are no public storages that point to the parent folder
+    # of another private storage, unintentionally exposing private files in
+    # this way
+    grouped: dict[str, dict[bool, set[str]]] = defaultdict(
```
amercader: Took me a while to wrap my head around this, perhaps a more descriptive name helps others
```suggestion
    storages_grouped_by_privacy: dict[str, dict[bool, set[str]]] = defaultdict(
```

ckan/lib/files/__init__.py
```diff
@@ -182,12 +183,50 @@ def collect_storages() -> dict[str, fk.Storage]:
...
+    for adapter, group in grouped.items():
+        if not group[False] or not group[True]:
+            continue
+
+        for private in group[False]:
+            if private.startswith(tuple(group[True])):
+                msg = (
+                    f"{adapter} storage with path {private} is included into another"
+                    + " storage with public access. Make sure paths of public"
+                    + " and private storages do not intersect."
```
amercader:
```suggestion
                    + " and private storages do not overlap."
```

doc/maintaining/filestore.rst
```diff
@@ -720,6 +720,33 @@ Use location and storage name to generate URL for the ``public_download`` view:
...
+.. warning:: Marking storage as ``public`` makes **all** files in the storage
+   accessible to everyone. This includes any file from any sub-folder of the
+   storage. Make sure you do not have another non-public storage inside the
+   same folder, or else files from that storage will also be publicly
+   accessible.
+
+   The following example shows the problem. Storage for ``resources`` is
+   private. But ``shared`` storage points at folder one level above and marked
+   as public, making all the files from ``resources`` storage accessible to
+   everyone, even though ``resources`` storage is not public::
```
amercader, line 732:
```suggestion
   The following example shows the problem. Storage for ``resources`` is
   private. But the ``shared`` storage points at a folder one level above and it is marked
   as public, making all the files from the ``resources`` storage accessible to
   everyone, even though ``resources`` storage is not public::
```

ckan/tests/lib/test_files.py
```diff
@@ -108,3 +109,29 @@ def test_transformers(self, faker: Faker):
...
+    monkeypatch.setitem(ckan_config, f"{prefix}.type", "ckan:fs")
+    monkeypatch.setitem(ckan_config, f"{prefix}.initialize", True)
+    monkeypatch.setitem(ckan_config, f"{prefix}.path", tmp_path + "/folder")
+
+    with pytest.raises(CkanConfigurationException):
+        reset_storages()
+
+    # storages with different backends do not overlap
+    monkeypatch.setitem(ckan_config, f"{prefix}.type", "ckan:memory")
+    reset_storages()
```
amercader, line 137: A couple more to cover more changes added in this PR:

```suggestion
    reset_storages()
    
    
def test_public_storages_cannot_include_private_storages_startup(
    monkeypatch: pytest.MonkeyPatch,
    ckan_config: dict[str, Any],
    tmp_path: Any,
):
    """Test that if public storages include private storages in their configuration
        CKAN will fail on startup."""
    tmp_path = str(tmp_path)
    prefix = "ckan.files.storage.test"
    monkeypatch.setitem(ckan_config, f"{prefix}.type", "ckan:fs")
    monkeypatch.setitem(ckan_config, f"{prefix}.path", tmp_path)
    monkeypatch.setitem(ckan_config, f"{prefix}.public", True)
    
    prefix = "ckan.files.storage.another"
    monkeypatch.setitem(ckan_config, f"{prefix}.type", "ckan:fs")
    monkeypatch.setitem(ckan_config, f"{prefix}.initialize", True)
    monkeypatch.setitem(ckan_config, f"{prefix}.path", tmp_path + "/folder")

    with pytest.raises(CkanConfigurationException) as e:
        files.collect_storages()


def test_file_system_storage_path_must_be_absolute(
    reset_storages: Any,
    monkeypatch: pytest.MonkeyPatch,
    ckan_config: dict[str, Any],
    tmp_path: Any,
):
   """Test that public storages cannot include private storages in their configuration."""
    prefix = "ckan.files.storage.test"
    monkeypatch.setitem(ckan_config, f"{prefix}.type", "ckan:fs")
    monkeypatch.setitem(ckan_config, f"{prefix}.path", "some_path")
    with pytest.raises(CkanConfigurationException) as e:
        files.collect_storages()

    tmp_path = str(tmp_path)
    prefix = "ckan.files.storage.another"
    monkeypatch.setitem(ckan_config, f"{prefix}.type", "ckan:fs")
    monkeypatch.setitem(ckan_config, f"{prefix}.initialize", True)
    monkeypatch.setitem(ckan_config, f"{prefix}.path", tmp_path + "/folder")    
```

amercader (commented):
@smotornyuk Looks good, just some minor tweaks

ckan/lib/files/default/fs.py
smotornyuk: Most likely, you did not run `pip install -r requirements.txt`. All unknown options are stored inside _extra_settings, and before _yesterday_ there was no official `follow_symlinks` option for fs storage. I added it in v0.2.6, and after upgrading file-keeper, you will see it on the top level

amercader: That was it, thanks

PR #9505, opened 2026-08-26
Don't use 'literal_binds' in executed SQL — it's a debugging option

bluepython508:
`literal_binds` is documented as generally being for logging or debugging purposes, and the result shouldn't be passed to databases.

We can use psycopg2's provided mechanism to do this.

### DIFF
--- a/ckanext/datastore/backend/postgres.py
+++ b/ckanext/datastore/backend/postgres.py
@@ -99,22 +99,26 @@ else:
         def __init__(self, error_dict: ErrorDict):
             pprint.pprint(error_dict)
 
-def is_single_statement(sql: str, params: dict[str, Any] | None = None) -> bool:
-    # Most calls to this have sqlalchemy `:abc` parameters
-    # pglast doesn't support these, since they're not the postgres format
-    # Compile the sqlachemy query to a string with psycopg2-style specifiers,
-    # then get psycopg2 to inline the parameters
-    # See also https://docs.sqlalchemy.org/en/21/faq/sqlexpressions.html#faq-sql-expression-string
-    #
-    # This requires a connection for the cursor, but shouldn't actually execute anything
-    engine = get_read_engine()
-    compiled = sa.text(sql).bindparams(**params or {}).compile(engine)
-    with engine.connect() as conn:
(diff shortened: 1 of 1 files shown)


ckanext/datastore/backend/postgres.py
```diff
@@ -99,22 +99,26 @@ class ValidationError(Exception):
...
-    # This requires a connection for the cursor, but shouldn't actually execute anything
-    engine = get_read_engine()
-    compiled = sa.text(sql).bindparams(**params or {}).compile(engine)
-    with engine.connect() as conn:
-        sql = conn.connection.cursor().mogrify(str(compiled), compiled.params).decode()
-    try:
-        return len(pglast.parser.parse_sql(sql)) == 1
-    except pglast.parser.ParseError:
-        return False
+ def inline_sql_parameters(sql: str, params: dict) -> str:
```
caiotheodoro, line 102: bug: mixed leading whitespace from here through `is_single_statement` - ruff reports `invalid-syntax: unindent does not match any outer indentation level` starting at this line. This is the whole reason Ruff/PyRight/pytest/Cypress are all red, not a real regression in the approach.

ckanext/datastore/backend/postgres.py
```diff
@@ -1689,16 +1693,15 @@ def _execute_single_statement_copy_to(
     for chunk in where_values:
         params.update(chunk)
 
-    if not is_single_statement(sql_string, params):
+    clause = inline_sql_parameters(sql_string, params)
+
+    if not is_single_statement(clause):
```
caiotheodoro, line 1698: risk: `clause` here already has every bind param inlined as a literal. `is_single_statement` calls `inline_sql_parameters` again internally, which runs `sa.text(clause)` - `sa.text` treats any `:word` substring as a bind parameter. A literal value containing a real colon (a time literal like `'12:30:00'`, a JSON string, a cast written with a single colon) would either blow up compiling with no matching bindparam or get silently mis-parsed. `is_single_statement` should take the sql string as-is and hand it straight to `pglast`, not re-run it through `sa.text`.

caiotheodoro (commented):
File doesn't parse as-is (indentation is broken from `inline_sql_parameters` onward), which is why 19/21 checks are failing. One design concern once that's fixed, inline.

PR #9508, opened 2026-08-27
bcp47 language codes for js translation files

wardi:
### Proposed fixes:
Instead of using a string replace in base.js let's store and serve the JS translations with BCP47 language codes as their names, as described in
https://github.com/ckan/ckan/pull/9474#issuecomment-5297647011

### Features:

- [ ] includes API changes

### DIFF
--- a/changes/9508.misc
+++ b/changes/9508.misc
@@ -1,0 +1,1 @@
+store JS translation files with BCP47 names instead of unix locale names
--- a/ckan/lib/i18n.py
+++ b/ckan/lib/i18n.py
@@ -327,6 +327,8 @@ def build_js_translations() -> None:
     ``ckan.i18n_directory``. These include only those translation
     strings that are actually used in JS files.
     '''
+    from ckan.lib.helpers import helper_functions as h
+
     log.debug(u'Generating JavaScript translations')
     ckan_i18n_dir = get_ckan_i18n_dir()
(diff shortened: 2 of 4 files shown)


u10313335:
Removing the `locale = locale.replace('-', '_')` from `ckan/public/base/javascript/client.js` will cause the  [get_locales_from_config()](https://github.com/ckan/ckan/blob/master/ckan/views/api.py#L472-L473) check to fail.

Maybe we should convert the language codes from `get_locales_from_config()` to BCP47 also:

```diff
diff --git a/ckan/views/api.py b/ckan/views/api.py
index 5530766784..6b0bbab60b 100644
--- a/ckan/views/api.py
+++ b/ckan/views/api.py
@@ -468,8 +468,11 @@ def organization_autocomplete(ver: int = API_REST_DEFAULT_VERSION) -> Response:
 def i18n_js_translations(
         lang: str,
         ver: int = API_REST_DEFAULT_VERSION) -> Union[str, Response]:
+    from ckan.lib.helpers import helper_functions as h

-    if lang not in get_locales_from_config():
+    langs = [h.unix_locale_to_bcp47(lang) for lang in get_locales_from_config()]
+
+    if lang not in langs:
         return _finish_bad_request('Unknown locale: {}'.format(lang))

     js_translations_folder = get_js_translations_dir()
```

PR #9510, opened 2026-08-27
replace theme with midnight-blue

wardi:
### Proposed fixes:
- replace ckan's default theme with new midnight-blue theme

### Features:

- [ ] includes API changes


smotornyuk:
@wardi, as discussed, I removed the `vendor/bootstrap/scss` folder. Instead, I added the `node_modules/` folder to the load paths for the SCSS imports. Because of it following became possible:
image

Previously, we imported Bootstrap sources using relative paths leading to the files included in the CKAN source.
Now, because of the updated load path configuration, when styles are compiled and `bootstrap/` is not found in the current directory, Gulp looks for it inside `node_modules/`.

Compiled CSS is still tracked by git, so this change will not affect the end user. As for theme-developers: they need to run `npm ci` or `npm install` at least once to get the sass compiler and all the dependencies(or else they won't be able to compile the theme). During this step, Bootstrap will be installed into `node_modules/` because it's listed as an explicit dependency inside `package.json`. In this way, we can guarantee that for theme developers there will be no difference.

wardi:
@smotornyuk I would appreciate some help with the failing webassets and cypress tests if you have time

smotornyuk:
CLI test: there was a `draft-package` asset definition that did not have a corresponding JS file, and it caused the command that removes compiled webassets to fail. The test was updated to report the reason for failure, and this unused asset definition was removed

As for the e2e test, it was failing because the midnight-blue version of the slug-preview module did not use translation for the `Edit` button; it looks like this problem was always there, but we did not notice it because tests were run with the classic theme.

The last failing tests from cypress, that I did not fix, is caused by insufficient contrast between normal text and links inside it - probably we should choose a brighter color for links?
image

wardi:
@smotornyuk I'm +1 on adjusting the CSS to make links just barely light enough to pass the test

(also it annoys me that the links on the bottom of the page include `https://` and only some of them have a trailing slash)

Zharktas:
@wardi @smotornyuk 

https://dequeuniversity.com/rules/axe/4.13/link-in-text-block

We either change the color of links slightly or add underlining to the links within the text which would also solve the issue. Even without low vision, the links within text are quite hard to see. Below I've attached a screenshot how it looks with underlining.

image

wardi:
I'm concerned that underlining links could negatively impact other pages.

Aside from the A11Y issue we should probably remove the last paragraph in this about section. I don't think we need a paragraph that includes links to specific portals.

Zharktas:
The accessibility issue is also on other pages like datasets with links in description and such, api and documentation links in the bottom of datasets page, we just don't have tests for those as of yet.

The link color needs to be really bright to pass the test. Below are some examples:
Link color #4747FF
image

Link color: #3333FF, surrounding text color #020204
image

wardi:
The tests are green now that the site links have been removed, @Zharktas WDYT about opening a separate ticket for the link color?

Zharktas:
Yeah lets create separate issue for that, the fix needs to backported to 2.12 anyway.

Zharktas:
@wardi should the changelog be something else than minor change since the more than a decade old theme is removed completely ? It will break all the instances that are built on the old theme and haven't updated to midnight blue. Otherwise this change is fine.

Issue #2357, opened 2015-03-20
Datapusher should have a timeout on post requests


rossjones:
When datapusher_submit is called, it makes a POST request to the datapusher service without a timeout.  As the datapusher has a limited number of workers (by default 1 process 15 threads) it is possible that post requests can take much longer than they should (stuck perhaps in the listening socket's backlog).  

When doing bulk-create/updates I get about 1 update/create a second - when disabling datapusher I get 4/s.

Fix should be to add timeout=somevalue to the requests.post call, and handle the timeout that is raised with a reasonable error message.

amercader:
Sounds sensible

davidread:
ckan-service-provider should simply queue each job up asynchronously and return. It sounds like that is not working correctly.

Vanuan:
Another issue I have is that datapusher goes through nginx, so api rate limit is applied. This results in 502 as there's no delay. Workaround is to disable nginx rate limit which is not acceptable.

Possible solutions:

1. Make datapusher go directly to python ckan server instead of public URL (#4013)
2. Modify datapusher cli to rate limit all requests:  https://github.com/ckan/datapusher/blob/master/datapusher/jobs.py#L6

Under normal conditions this is possible using ratelimit decorator: https://pypi.python.org/pypi/ratelimit/1.1.0
But since datapusher creates a new process for each job, this rate limit only applies per process.
So there should be some message broker which delays responses.

Vanuan:
Ok, now datapusher completely hangs. I need to restart to unblock it

Vanuan:
Maybe datapusher threads fail with exception and never finish

andtheWings:
### Current assessment

The original issue is mostly resolved in the current codebase:

- `datapusher_submit()` now passes `timeout=config.get("ckan.requests.timeout")` to the DataPusher POST request.
- `ckan.requests.timeout` is declared with a 5-second default.
- Connection failures are recorded on the task and reported as validation errors.

There is still a narrower edge case: `requests.exceptions.ReadTimeout` is not caught. A read timeout can leave the task in `submitting` and surface an unhandled exception. The existing tests also do not cover timeout handling.

The 2018 comments about nginx rate limiting and DataPusher backpressure are broader service/infrastructure concerns and are not addressed by this client timeout.

Recommendation: close this issue as fixed, or open a focused follow-up for handling all `requests.exceptions.Timeout` cases and adding regression coverage.

PR #9511, opened 2026-08-28
fix: handle DataPusher timeout exceptions

andtheWings:
## Summary

- Handle all DataPusher request timeout exceptions.
- Persist the task as `error` instead of leaving it in `submitting`.
- Add regression coverage for `ReadTimeout`.

Closes the remaining edge case identified in #2357.

## Testing

- `ckanext/datapusher/tests/test_action.py`: 10 passed


pdelboca (approved):
Thanks for this fix @andtheWings !

PR #9516, opened 2026-09-07
Commit unterminated tag text on blur in tags-mode autocomplete fields

### LINKED ISSUE #9515: Pasting a tag list without a trailing separator silently drops the last tag
### Summary

In any tags-mode `autocomplete` field (a `select2`-based tag input, e.g. the dataset "Tags" field), pasting a comma-separated list of tags that does **not** end with a trailing comma leaves the last tag sitting as plain, uncommitted text in the search box. If the form is then submitted without the user noticing, that last tag is silently lost — it was never added to the underlying `<select>`, so nothing warns the user it's missing.

### CKAN version

2.11 (reproduced and root-caused against 2.11.6's shipped JS).

### Steps to reproduce

1. Go to "Add Dataset" and focus the Tags field.
2. Paste: `delta, epsilon, zeta` (no trailing comma — this is the normal result of copying a comma-separated list from, e.g., a spreadsheet or another CKAN tag field).
3. Click elsewhere on the page (blur the field) without pressing Enter or typing a trailing comma.
4. Inspect the tag field: `delta` and `epsilon` became tags (chips); `zeta` is gone — not a chip, and not left as visible text either.

### Expected behavior

All three comma-separated values become tags, matching what happens when the pasted text does end with a separator (`"delta, epsilon, zeta,"` correctly produces three tags today).

### Actual behavior

The last, unterminated segment is dropped without any indication to the user.

### Root cause

`ckan/public/base/javascript/modules/autocomplete.js` sets select2's `tokenSeparators` (via the module's `tokensep` option, default `","`) to split pasted text into tags. select2's own tag-input handling only commits a segment once it sees a following separator — the last segment, having none, stays as unconverted search text. This is a known, long-standing limitation of `select2` 3.x's tag mode when combined with `tokenSeparators` (see select2/select2#5076), which is unlikely to ever be fixed upstream in `select2`, since that major version is no longer maintained. CKAN's own `autocomplete.js` module already works around a related quirk of the same mechanism for a typed comma (`_onKeydown`, added in #4421), so a wrapper-level fix here follows the same, already-accepted pattern.

### Suggested fix

Commit any leftover, unterminated text as a final tag when the field loses focus (blur), reusing the same "simulate pressing Enter" trick `_onKeydown` already uses for a typed comma. I have a minimal patch and a regression test ready and will open a pull request shortly.

bjornhagstrom:
Fixes #9515

### What this fixes

Pasting a comma-separated list of tags into a tags-mode `autocomplete` field without a trailing separator (e.g. `"delta, epsilon, zeta"`, the normal shape of text copied from a spreadsheet or another tag field) left the last entry (`zeta`) as uncommitted text in the search box. If the form was then saved, that tag was silently lost.

### Root cause

select2's `tokenSeparators` handling (used here via the module's `tokensep` option) only converts a pasted segment into a tag once it sees a *following* separator, so the final, unterminated segment is never committed. This is a known limitation of select2 3.x in tag mode
(see [select2/select2#5076](https://github.com/select2/select2/issues/5076), open and unlikely to be addressed given that major version's maintenance status).

### The fix

Two small, related changes to `ckan/public/base/javascript/modules/autocomplete.js`:

1. Extend the default `tokensep` from `","` to `",\n"`, so a newline-separated paste (e.g. one tag per line) is also split correctly — a smaller, related gap in the same mechanism.
2. Add a `blur` handler on the tag field's search box that commits any non-empty leftover text as a final tag, by firing the same synthetic Enter keydown `_onKeydown` already uses for a typed comma. This mirrors the approach already accepted in #4421 for a different edge case of the same input.

### Testing

- Extended `cypress/e2e/modules/autocomplete.cy.js`: updated the `.setupAutoComplete()` `tokenSeparators` expectations for the new default, and added a `._onBlur(event)` describe block covering: committing on blur when text is left over, trimming before deciding whether to commit, and doing nothing for empty/whitespace-only content.
- Ran the full `autocomplete.cy.js` spec locally against a CKAN dev instance (`test-infrastructure` docker setup): 36 passing, 2 pending (pre-existing, unrelated placeholder tests), 0 failing.
- Manually verified in a local CKAN instance: pasted `"delta, epsilon, zeta"` into a dataset's Tags field, clicked away, and confirmed all three became tags before saving.

### Checklist

- [x] New/changed tests included
- [x] Existing tests unaffected (no other file touched)
- [ ] Documentation update (none needed — internal implementation detail)

### Note

This PR was prepared with the assistance of an AI coding tool (Claude), under my direction and review — I tested the fix locally (see Testing above) and read through the diff myself before submitting. I'm new to contributing to CKAN (and to open source in general), so apologies in advance for any missteps in process or convention — happy to fix anything that's off.

### DIFF
--- a/ckan/public/base/javascript/modules/autocomplete.js
+++ b/ckan/public/base/javascript/modules/autocomplete.js
@@ -10,7 +10,10 @@
  * label    - A string of the label you want to appear within the dropdown for
  *            returned results
  * tokensep - A string that contains characters which will be interpreted
- *            as separators for tags when typed or pasted (default ",").
+ *            as separators for tags when typed or pasted (default ",\n" --
+ *            a comma or a newline, so pasting a newline-separated list of
+ *            tags -- such as text copied from another tag field -- is
+ *            split into separate tags too).
  * Examples
  *
  *   // <input name="tags" data-module="autocomplete" data-module-source="http://" />
@@ -25,7 +28,7 @@ this.ckan.module('autocomplete', function (jQuery) {
       key: false,
       label: false,
       source: null,
-      tokensep: ',',
+      tokensep: ',\n',
       interval: 300,
       dropdownClass: '',
       containerClass: '',
@@ -128,6 +131,18 @@ this.ckan.module('autocomplete', function (jQuery) {
         // find the "fake" input created by select2 and add the keypress event.
         // This is not part of the plugins API and so may break at any time.
         select2.search.on('keydown', this._onKeydown);
+
+        // Typing a comma (see _onKeydown below) commits the current search
+        // text as a tag, and select2's own tokenSeparators handling does
+        // the same for text arriving via paste -- but only up to the last
+        // separator. Pasting "a, b, c" with no trailing comma (the usual
+        // case when copying from a spreadsheet or a document) leaves "c"
+        // sitting unconverted in the search box. If the form is submitted
+        // like that, the trailing tag is silently dropped: it was never
+        // added to the underlying select, so nothing warns the user.
+        // Commit any such leftover text as a final tag when the field
+        // loses focus, the same way a typed separator would.
+        select2.search.on('blur', this._onBlur);
       }
 
       // This prevents Internet Explorer from causing a window.onbeforeunload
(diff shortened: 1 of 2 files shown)


Zharktas:
The newline character `\n` added to token separators has nothing to do with the described issue. The AI generated comments are wildly cumbersome to read, the change is 7 lines of code and AI adds almost 30 lines of comments, can you reduce those?

bjornhagstrom:
Thanks for the feedback, @Zharktas.

**Newline separator:** related edge case of the same `tokenSeparators` mechanism — pasting a list with one tag per line (e.g. from a spreadsheet column) doesn't get split without it. Happy to drop it into a follow-up if you'd rather keep this PR scoped to #9515 as reported.

**AI-generated comments:** agreed, will trim those to match the surrounding style.

**More importantly:** testing this against CKAN 2.12, we found the binding condition this PR (and the existing `keydown`/comma binding above it) relies on never actually fires — regardless of `data-module-tags`.

CKAN's `autocomplete.js` gates both bindings on `select2.search`:
```js
if (this.options.tags && select2 && select2.search) {
```
That property existed in Select2 3.x but not in 4.x (CKAN now pins `^4.1.0-rc.0`) — confirmed against a live instance (`'search' in select2` → `false`) and in Select2's source, where the search input moved to `select2.selection.$search`. So `select2.search` is always falsy on `master`, and neither binding ever attaches — likely the deeper cause of #9515, not just the no-trailing-separator case here. The Cypress tests here call `_onBlur()` directly, so they don't exercise this init path.

Happy to send a revised fix rebinding to `select2.selection.$search`, trimmed comments, and the newline question resolved either way — let us know what you'd prefer.

(As before: prepared with AI assistance, Claude, under our review.)

PR #9524, opened 2026-09-09
fix: KeyError in default_extras_schema when validating extras without…

### LINKED ISSUE #9523: KeyError in default_extras_schema when validating extras without an explicit id
## CKAN version
2.11.6

## Describe the bug
Since CKAN 2.11.6, `package_show`/search-indexing validation crashes with
`KeyError: ('extras', 0, 'id')` for any package whose `extras` list contains
an entry with no `id` key — which is the normal case for extras created via
`convert_to_extras` (no code sets an explicit `id` on them).

### Steps to reproduce

1. Create/update a package (in our case it is a ckanext-showcase showcase but should be reproducable with any package type whose plugin's `show_package_schema()` doesn't
   override `extras`) that has at least one plain key/value extra with no
   `id` field.
2. Trigger validation of that package's extras against
   `default_extras_schema()` in a context with no `user` and no
   `ignore_auth` — e.g. CKAN's own synchronous search indexer
   (`ckan.lib.search.index.PackageSearchIndex.index_package`, which calls
   `lib_plugins.plugin_validate(..., pkg_dict, schema, 'package_show')`
   with `context = {'model': model, 'session': model.Session}`).
3. Observe:

  File "ckan/lib/search/index.py", line 117, in index_package
    validated_pkg_dict, _errors = lib_plugins.plugin_validate(...)
  File "ckan/logic/validators.py", line 597, in ignore_not_sysadmin
    data.pop(key)
  KeyError: ('extras', 0, 'id')

Because `ckan.model.modification.DomainObjectModificationExtension.notify()`
only re-raises `SearchIndexError` the
enclosing `package_create`/`package_update` succeeds but the package is not added to the search index. No error is visible to the user who created it.

stefina:
Fixes #9523 

### Features:

- [ ] includes API changes

### DIFF
--- a/ckan/logic/validators.py
+++ b/ckan/logic/validators.py
@@ -581,7 +581,7 @@ def ignore_not_sysadmin(key: FlattenKey, data: FlattenDataDict,
     if ignore_auth or (user and authz.is_sysadmin(user)):
         return
 
-    data.pop(key)
+    data.pop(key, None)
 
 
 def limit_sysadmin_update(key: FlattenKey, data: FlattenDataDict,
--- a/changes/9523.bugfix
+++ b/changes/9523.bugfix
@@ -1,0 +1,1 @@
+fallback to None to prevent KeyError in default_extras_schema when validating extras without an explicit id


ckan/logic/validators.py
```diff
@@ -581,7 +581,7 @@ def ignore_not_sysadmin(key: FlattenKey, data: FlattenDataDict,
     if ignore_auth or (user and authz.is_sysadmin(user)):
         return
 
-    data.pop(key)
+    data.pop(key, None)
```
caiotheodoro, line 584: risk: on `master`, `default_extras_schema`'s `id` field is `[ignore_missing, empty_if_not_sysadmin, uuid_validator]` - `ignore_not_sysadmin` isn't in that chain at all, so this fix doesn't touch the path that's actually crashing for the reporter (2.11.6). On `2.11`, `id` is still `[ignore_not_sysadmin, ignore_missing, uuid_validator]` (from 134462b39f, the #9380 fix), so `ignore_not_sysadmin` runs first there and this same one-line change needs to land on `2.11` too or the reported crash keeps happening.

caiotheodoro, line 584: nit: no test covers this. I tried reproducing with a plain `ckan.lib.navl.dictization_functions.validate({'extras': [{'key': 'x', 'value': 'y'}]}, {'extras': default_extras_schema()}, {'model': model, 'session': model.Session})` and it didn't crash - `augment_data` seeds absent schema keys with the `missing` sentinel before validators run, so `data.pop(key)` finds something to pop either way. Whatever schema shape actually hits this (sounds like a plugin's `show_package_schema()` that leaves `extras` on `default_extras_schema()` unmodified, per the issue) is worth pinning down in a test so this can't regress silently again.

caiotheodoro (commented):
Fix itself is correct and safe on its own. Two things worth checking before merge: this doesn't actually touch the crash path on `2.11` (the branch the report is against), and there's no regression test.

ckan/logic/validators.py
stefina: @caiotheodoro thank you for that feedback, you are right. 
Can you change the target-branch of this PR to point to `dev-v2.11` as I don't have permissions to do that or do you prefer me opening a new PR?

stefina:
I created a new PR, generally containing the same changes, but addressing the feedback already given by @caiotheodoro 
So, closing this PR, as it is being replaced by this one: https://github.com/ckan/ckan/pull/9528 Hoping that I branched of the correct branch and targeting the correct branch, too 🤞

